PR #1, opened 2018-01-26
Deletes minikube-related content.

coollog:
Moves 'crepecake' directory to root and removes all 'minikube-build-tools-for-java' related files.


coollog:
https://github.com/google/jib/issues/6 will be fixed in a followup PR.

PR #36, opened 2018-01-31
Adds support for parsing a base image reference into components.

### LINKED ISSUE #28: Add support for parsing a base image reference into components.
Parse a base image reference into the registry, repository, and tag.

coollog:
Fixes #28 

Adds `from` parameter to plugin to replace the separate `baseImageRegistry`, `baseImageRepository`, and `baseImageTag` configuration parameters.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2018 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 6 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -0,0 +1,139 @@
...
+
+    if (registry == null) {
+      registry = DEFAULT_REGISTRY;
+    }
+
+    if (repository == null) {
+      throw new InvalidImageReferenceException(reference);
+    }
+    // For Docker Hub, if the repository is only one component, then it should be prefixed with
+    // 'library/'.
```
loosebazooka: why? maybe a link?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -86,6 +86,7 @@ public static ImageReference parse(String reference) throws InvalidImageReferenc
     String tag = matcher.group(3);
     String digest = matcher.group(4);
 
+    // If no registry was matched, use Docker Hub by default.
```
loosebazooka: I wonder if since we're behaving a certain way only if `DEFAULT_REGISTRY` is docker hub that we should just rename `DEFAULT_REGISTRY` to indicate that it is actually dockerhub: like `DOCKER_HUB` or something?

loosebazooka: Also since changing default without looking down at this section of code would break it, since `library` is specific to dockerhub?

PR #41, opened 2018-02-01
Fixes Kokoro build script.

coollog:
The `jib-maven-plugin` test needs `jib-core` to be installed into the local Maven repo.

### DIFF
--- a/kokoro/continuous.bat
+++ b/kokoro/continuous.bat
@@ -6,3 +6,4 @@ cd jib-core && call gradlew.bat clean build publishToMavenLocal --info
 cd ../jib-maven-plugin && call mvnw.cmd clean install cobertura:cobertura -B -U -X
 
 exit /b %ERRORLEVEL%
+g
--- a/kokoro/continuous.bat
+++ b/kokoro/continuous.bat
@@ -2,7 +2,7 @@
 
 cd github/jib
 
-cd jib-core && call gradlew.bat clean build --info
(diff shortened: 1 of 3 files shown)


kokoro/continuous.bat
```diff
@@ -2,7 +2,7 @@
 
 cd github/jib
 
-cd jib-core && call gradlew.bat clean build --info
+cd jib-core && call gradlew.bat clean build publishToMavenLocal --info
```
loosebazooka: I think I remember the problem with using the "publish" plugin vs "maven" plugin. It's that the publish plugin didn't behave when trying to publish to maven central, but it's now irrelevant since we'll be using kokoro to build and deploy.

kokoro/continuous.sh
```diff
@@ -5,5 +5,5 @@ set -x
 
 cd github/jib
 
-(cd jib-core; ./gradlew clean build --info)
+(cd jib-core; ./gradlew clean build publishToMavenLocal--info)
```
loosebazooka: I think you need a space between `publishToMavenLocal` and `--info`

coollog: Yea haha, fixed.

kokoro/continuous.bat
coollog: Yeah, this is only for jib-core. Is there another way for jib-maven-plugin to use jib-core as a dependency but not have it be used as a Maven artifact dependency?

PR #34, opened 2018-01-30
Reorganize build steps for parallelization.

coollog:
Fixes #8 
Sort of fixes #30 

Most of the code just wraps each build step in a `ListenableFuture`. The inputs to each build step then become the `ListenableFuture`s for their dependencies (the steps they depend on). 

The code should be refactored to reduce the mess of `ListenableFuture`s everywhere.

To check it out, run on `jib-core`:

`./gradlew publishToMavenLocal`

Then on `jib-maven-plugin`:

`./mvnw install`

Then on a test project, configure like such:

```
      <plugin>
        <groupId>com.google.cloud.tools</groupId>
        <artifactId>jib-maven-plugin</artifactId>
        <version>0.1.0-SNAPSHOT</version>
        <configuration>
          <credentialHelperName>gcr</credentialHelperName>-->
          <registry>gcr.io</registry>-->
          <repository>gcp-project/my-app</repository>
          <mainClass>com.test.HelloWorld</mainClass>
        </configuration>
      </plugin>
```

And run with `mvn compile jib:build`.

PR #46, opened 2018-02-03
Fixes swallowed ExecutionException.

coollog:
#34 introduced a regression that swallowed up exceptions from the build process. This fixes that.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -243,7 +279,11 @@ private ImageReference getImageReference() throws MojoFailureException {
     }
   }
 
-  private <T extends Exception> void provideSuggestionForException(
+  /**
+   * Wraps an exception in a {@link MojoExecutionException} and provides a suggestion on how to fix
+   * the error.
+   */
+  private <T extends Throwable> void provideSuggestionForException(
```
loosebazooka: Does this name still fit this method? Could we make it something along the lines of `throwMojoExceptionWithHelpMessage(...)`, or just anything that indicates this method is throwing something deliberately?

PR #54, opened 2018-02-05
Adds prepare_release.sh script.

coollog:
Forked from https://github.com/GoogleCloudPlatform/endpoints-framework-maven-plugin/blob/master/scripts/prepare_release.sh


scripts/prepare_release.sh
```diff
@@ -0,0 +1,77 @@
...
+# For example, when releasing 1.5.7, the next snapshot version would be 1.5.8-SNAPSHOT.
+NEXT_SNAPSHOT=${NEXT_VERSION}-SNAPSHOT
+mvn versions:set versions:commit -DnewVersion=${NEXT_SNAPSHOT}
+
+# Commits this next snapshot version.
+git commit -am "${NEXT_SNAPSHOT}"
+
+# Pushes the tag and release branch to Github.
+git push origin v${VERSION}
+git push --set-upstream origin ${VERSION}
```
loosebazooka: I always wondered why we need --set-upstream line? We don't really care that this branch is tracking with the upstream one do we?

coollog: This is for the branch to push initially. Pushing without `--set-upstream` gives a fatal error like `fatal: The current branch i-dont-exist-upstream has no upstream branch.` I believe.

coollog: Nvm that is for `git push`

PR #60, opened 2018-02-06
Update README.md

parkjam4:
Here are my suggested changes.
Let me know if you have any questions. Thanks!


README.md
```diff
@@ -8,11 +8,11 @@
 
 ## What is Jib?
 
-Jib is a tool for building container images for your Java applications.
+Jib is a tool for building container images for your Maven-based Java applications.
```
coollog: I think we can leave it as is since we will be adding a Gradle plugin as well. Or do you think it might be confusing to users to find that it is only for Maven currently?

README.md
```diff
@@ -38,9 +38,9 @@ In your Maven Java project, add the plugin to your `pom.xml`:
 
 ### Configuration
 
-Configure the plugin by changing `registry`, `repository`, and `credentialHelperName` accordingly.
+Configure the plugin by changing `registry`, `repository`, and `credentialHelperName` in the registry of your choice. For example:
```
coollog: "in the registry of your choice" might not be the right wording here, since `repository` is not specific to a registry. Perhaps something like "...changing `registry`, `repository`, and `credentialHelperName`. (See [Extended Usage]() for details.)" might fit better?

README.md
```diff
@@ -38,9 +38,9 @@ In your Maven Java project, add the plugin to your `pom.xml`:
...
 
-#### I am using Google Container Registry (GCR)
+Using Google Container Registry (GCR)...
```
coollog: Should this still be in header4?

README.md
parkjam4: Will you be able to add Gradle support relatively soon? If not, I think it's important to let them know that it only works with Maven as soon as possible.

README.md
parkjam4: What if we said something like:
Configure the plugin by changing `registry`, `repository`, and `credentialHelperName` in your registry.

README.md
parkjam4: I prefer to minimize the use of headers, but we can keep them if you prefer. Up to you.

README.md
coollog: Okay, sounds good. Gradle plugin won't be added until the Maven plugin has been worked out more. How does changing "a tool" to "a Maven plugin" sound?

README.md
coollog: I like headers mostly because they can be easily linked to, like https://github.com/google/jib#i-am-using-google-container-registry-gcr

README.md
parkjam4: So, the new description would be:
Jib is a Maven plugin for building container images for your Java applications.

Is that correct?

README.md
parkjam4: Ok, that makes sense.
But, I'm not crazy about leading with i am. 
Are you ok with :
### Using Google Container Registry (GCR)

coollog: Yep, that looks good!

README.md
coollog: Yep!

parkjam4:
Looks like I can't squash and merge.

PR #62, opened 2018-02-06
Adds example project using jib-maven-plugin.

### LINKED ISSUE #61: Add example project with jib-maven-plugin.
Add example `hello-world` project with `jib-maven-plugin` added.

coollog:
Fixes #61 

This project builds a simple HelloWorld application that uses one resource file and one dependency.


coollog:
@https://github.com/mattmoor This may be handy as an example to play around with.

examples/helloworld/src/main/java/example/HelloWorld.java
```diff
@@ -0,0 +1,36 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package example;
+
+import com.google.common.io.Resources;
```
elharo: For examples like this it's good if we can keep dependencies to an absolute minimum, not even Guava.

coollog: This example was intended to show how Jib can pack dependencies, resources, and classes into separate layers and that running the final image runs the application accordingly. I included Guava as an example dependency. Should I remove it and leave it as just a class and a resource instead?

PR #63, opened 2018-02-06
Adds exception handling to ManifestPusher.

### LINKED ISSUE #40: Pushing image with invalid tag should not succeed
For example, pushing an image with a tag that contains a colon.

Solutions:

1) Match configuration against regex.
2) Better error handling in `ManifestPusher` for `404`.

coollog:
Fixes #40 .

Specifically, it lets the user know what might have caused a 404 (a bad tag is most likely). Note that there is no validation for the configuration here.


jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -58,6 +60,24 @@ public Void handleResponse(Response response) {
...
+      throws HttpResponseException, RegistryErrorException {
+    if (httpResponseException.getStatusCode() == HttpStatusCodes.STATUS_CODE_NOT_FOUND) {
+      /*
+       * This error could happen if the repository doesn't exist (which should fail at BLOB-push).
+       * The more likely reason is that the image is pushing to a tag with backslashes in it, in
+       * which part of the tag becomes part of the repository.
+       */
+      throw new RegistryErrorExceptionBuilder(getActionDescription(), httpResponseException)
+          .addReason(
+              "repository name not known to registry (make sure that you are using a valid tag - tags cannot contain backslashes)")
```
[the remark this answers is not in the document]
coollog: Changed the wording to "...perhaps you are using an invalid tag..." - does that sound better?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryErrorExceptionBuilder.java
```diff
@@ -90,6 +92,9 @@ RegistryErrorExceptionBuilder addReason(String reason) {
   }
 
   RegistryErrorException build() {
+    // Provides a feedback channel.
+    errorMessageBuilder.append(
+        " | If this is a bug, please file an issue at https://github.com/google/jib/issues/new");
```
loosebazooka: Can you put `github.com/google/jib....` in a ProjectInfo file or something static that we reference everywhere?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -58,6 +60,24 @@ public Void handleResponse(Response response) {
...
+      throws HttpResponseException, RegistryErrorException {
+    if (httpResponseException.getStatusCode() == HttpStatusCodes.STATUS_CODE_NOT_FOUND) {
+      /*
+       * This error could happen if the repository doesn't exist (which should fail at BLOB-push).
+       * The more likely reason is that the image is pushing to a tag with backslashes in it, in
+       * which part of the tag becomes part of the repository.
+       */
+      throw new RegistryErrorExceptionBuilder(getActionDescription(), httpResponseException)
+          .addReason(
+              "repository name not known to registry (perhaps you are using an invalid tag - tags cannot contain backslashes)")
```
elharo: Is the repository name available here? If so it could be in the error message and you could tell for sure if it contains slashes.

coollog: I've tried this on GCR and it produces a correct error message already (about using an invalid tag - this error message will show up as part of the wrapped HttpResponseException), but on ECR, it simply gives a 404 and not in the expected Docker registry 404 format.

loosebazooka: If we know their tag here already on our own, from `imageTag`, can we just tell them that their tag contains backslashes if it does?

loosebazooka: and if it doesn't, then we print the registry from `registryEndpointProperties.getImageName()` ?

coollog: If their `imageTag` contains backslashes, we might want to just tell them that before entering the build process. Maybe we should just add a warning in the Maven plugin if the tag configuration contains backslashes?

loosebazooka: If tags with `\` are illegal, perhaps manifest pusher might want to throw an illegal arugment exception in the constructor?

coollog: I think it might be good to keep `ManifestPusher` decoupled from any logic about the formatting of the image tag, since that is specific to the implementation of different Docker registries. Also, an `IllegalArgumentException` would mean an additional logic in `BuildImageMojo` to catch that exception and give the user a more meaningful error message. I added some validation in the `BuildImageMojo` that validates its configuration - currently it just logs some errors but doesn't fail, but perhaps it should fail? I can also remove this custom error message in `ManifestPusher` since the one in the Mojo is more helpful.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -233,24 +233,31 @@ void buildImage(BuildImageSteps buildImageSteps) throws MojoExecutionException {
...
 
   /** Checks validity of plugin parameters. */
-  private void validateParameters() {
+  private void validateParameters() throws MojoFailureException {
+    boolean shouldFail = false;
+
     // 'tag' must not contain backslashes.
     if (tag.indexOf('/') >= 0) {
       getLog().error("'tag' cannot contain backslashes");
+      shouldFail = true;
```
loosebazooka: this tag check can probably just go at the end, and we can remove the shouldFail flag.

PR #73, opened 2018-02-07
Removes unnecessary check that actually breaks legitimate builds.

### LINKED ISSUE #72: Error when building with no resources
Building with no resources gives error: `Source files for application layer cannot be empty`

coollog:
Fixes #72

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
@@ -43,10 +43,6 @@ class LayerMetadata {
   private final FileTime lastModifiedTime;
 
   LayerMetadata(List<String> sourceFiles, FileTime lastModifiedTime) {
-    if (sourceFiles.isEmpty()) {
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
```diff
@@ -43,10 +43,6 @@ static LayerMetadata from(List<Path> sourceFiles, FileTime lastModifiedTime) {
   private final FileTime lastModifiedTime;
 
   LayerMetadata(List<String> sourceFiles, FileTime lastModifiedTime) {
-    if (sourceFiles.isEmpty()) {
```
loosebazooka: Would allow possibly invalid apps through? All resources, no source? I mean I guess we don't need to care if the user is being insane.

coollog: Yeah, it would allow that, though I think our guarantee could be that the application runs in the container like it would locally.

PR #84, opened 2018-02-09
Warns the user if an inferred Docker credential helper could not be found.

### LINKED ISSUE #82: Warn the user if a Docker credential helper could not be found.
Since we are inferring Docker credential helpers, we should not fail the build if one is not found, but rather just provide a warning.

coollog:
Requires #81 
Fixes #82 

1) Moves credential helper inference to `RetrieveRegistryCredentialsStep`.
2) The new flow for retrieving credentials is:
  i) Checks each configured `credHelper`. If any is not found, the build fails.
  ii) Checks the Maven settings.
  iii) Tries to infer a credential helper. If any is not found, a warning is given.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/RetrieveRegistryCredentialsStep.java
```diff
@@ -65,6 +66,24 @@ public Authorization call() throws IOException, NonexistentDockerCredentialHelpe
         return buildConfiguration.getKnownRegistryCredentials().getAuthorization(registry);
       }
 
+      // Tries to infer common credential helpers for known registries.
+      for (String registrySuffix : COMMON_CREDENTIAL_HELPERS.keySet()) {
+        if (registry.endsWith(registrySuffix)) {
+          try {
+            Authorization authorization =
+                retrieveFromCredentialHelper(COMMON_CREDENTIAL_HELPERS.get(registrySuffix));
```
loosebazooka: This method (`retreiveFromCredentialHelper`) seems to be catching the NonexistantDockerCredentialHelperException on line 110 below?

coollog: Yea, since we are inferring, the build should not break if the inferred credential helper was not found. For example, the user could be pulling a public image on gcr.io that does not require credentials.

loosebazooka: oops, okay I read the names wrong.

PR #86, opened 2018-02-12
Adds upcoming features to top of README.

### LINKED ISSUE #69: readme should make it clear that this is for non-war use-cases at the very top

coollog:
Fixes #69 

Named this `Upcoming Features`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,8 @@ We currently do not support building to a local Docker daemon. However, this fea
 
 You can still [`docker pull`](https://docs.docker.com/engine/reference/commandline/pull/) the image built with `jib-maven-plugin` to have it available in your local Docker daemon.
 
+You can also [run a local Docker registry](https://docs.docker.com/registry/deploying/) and point Jib to push to the local registry.
+
 ### How do I enable debugging?
 
 *TODO: Provide solution.*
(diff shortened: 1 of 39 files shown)


README.md
```diff
@@ -17,6 +17,15 @@ Jib is a Maven plugin for building container images for your Java applications.
 
 * **Native** - Reduce your CLI dependencies. Build your Docker image from within Maven <!--or Gradle--> and push to any registry of your choice. No more writing Dockerfiles and calling docker build/push.
 
+## Upcoming Features
```
loosebazooka: I would put this in later, maybe even at the end?

loosebazooka:
@parkjam4 hey jimmy, what do you think of this format? We want to tell users that only `JAR`s  are supported now and not `WAR`s so we're including this section in here to tell everyone what we don't support yet (but in a nice positive coming-soon way :smile:).

README.md
```diff
@@ -17,6 +17,15 @@ Jib is a Maven plugin for building container images for your Java applications.
...
+
+These features will be added in later releases.
```
parkjam4: I know what you're trying to do here, but the wording feels a little strange to me.
What if you were more explicit that these features are not supported?
Like changing this line to:
These features are currently not supported but will be coming in later releases.

coollog: Okay added the `currently not supported` phrase!

PR #88, opened 2018-02-12
Enables Kokoro integration testing for jib-core.

coollog:


coollog:
Not enabling for Kokoro Windows as `docker` is not working on that. Will need to figure this out.

CONTRIBUTING.md
```diff
@@ -25,6 +25,8 @@ again.
    use your corporate email address here, not your personal address.
 2. Fork the repository into your own Github account.
 3. Please include unit tests (and integration tests if applicable) for all new code.
-4. Make sure all existing tests pass. (./gradlew goJF build integrationTest)
+4. Make sure all existing tests pass. 
+  a. In `jib-core`, run `./gradlew clean goJF build integrationTest`
+  b. In `jib-maven-plugin`, run `./mvnw clean verify`
```
loosebazooka: users can run fmt:format to format the maven code?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildImageStepsIntegrationTest.java
```diff
@@ -35,6 +35,8 @@
 
   private static final Logger logger = LoggerFactory.getLogger(TestBuildLogger.class);
 
+  //  private static void assertImage
```
loosebazooka: remove this?

coollog:
Filed https://github.com/google/jib/issues/99

CONTRIBUTING.md
coollog: I actually attached fmt:format to run during the validate lifecycle, which now that I think about it, is probably not a good idea because Kokoro won't pick up formatting issues.

PR #94, opened 2018-02-13
Adds note about 'mvn clean' in README.

coollog:
This lets users know that if they just run `mvn clean compile jib:build` each time, they don't get the cache speed-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Build your container image with:
 mvn compile jib:build
 ```
 
-Subsequent builds are much faster than the initial build.
+Subsequent builds are much faster than the initial build. `mvn clean` cleans Jib's build cache and forces Jib to re-pull the base image and re-build the application layers.
 
 *Having trouble? Let us know by [submitting an issue](/../../issues/new).*
 
--- a/README.md
+++ b/README.md
@@ -84,7 +84,13 @@ Build your container image with:
 mvn compile jib:build
 ```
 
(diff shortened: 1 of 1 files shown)


coollog:
@parkjam4 How does this sound?

README.md
```diff
@@ -75,7 +75,7 @@ Build your container image with:
 mvn compile jib:build
 ```
 
-Subsequent builds are much faster than the initial build.
+Subsequent builds are much faster than the initial build. `mvn clean` cleans Jib's build cache and forces Jib to re-pull the base image and re-build the application layers.
```
parkjam4: This new sentence seems disconnected from the rest of the section. 
Are you trying to say that `mvn compile jib:build` should be the command you run to build your image and not `mvn clean compile jib:build`?

coollog: Yes, the idea is that `mvn compile jib:build` is the command to run in most cases, and if you want to clear Jib's cache, `mvn clean` would do that.

parkjam4: Ok, what if you worded it like this:

Build your container image with:
 
 ```commandline
 mvn compile jib:build
 ```
Subsequent builds are much faster than the initial build. 

If you want to clear Jib's build cache and force it to re-pull the base image and re-build the application layers, run:

 ```commandline
 mvn clean compile jib:build
 ```

coollog: That sounds much clearer! Will update with this

Issue #11, opened 2018-01-29
Run integration tests on commits to master with proper environment.


coollog:
Some blockers against this right now include:

- no credential helper on test machine
- no Docker on test machine

PR #97, opened 2018-02-14
Moves image running logic to core and uses jib-integration-testing GCP account.

coollog:
Refactors common Docker image running logic in integration tests for `jib-core` and `jib-maven-plugin`. Also changes the GCP account for `jib-maven-plugin` integration testing to `jib-integration-testing` rather than my own sandbox account.

This prepares for #11


jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/DockerImageRunner.java
```diff
@@ -0,0 +1,55 @@
...
+  /** Runs a command with naive tokenization by whitespace. */
+  private static String runCommand(String... command) throws IOException, InterruptedException {
+    Process process = new ProcessBuilder(Arrays.asList(command)).start();
+
+    try (InputStreamReader inputStreamReader =
+        new InputStreamReader(process.getInputStream(), StandardCharsets.UTF_8)) {
+      String output = CharStreams.toString(inputStreamReader);
+
+      if (process.waitFor() != 0) {
+        throw new IOException("Command '" + String.join(" ", command) + "' failed");
```
loosebazooka: Do we ever want to catch this? Is this actually the right exception type? Could it just be runtimeException? Are we trying to fail the test?

coollog: Yea, the intent is to fail the test.

PR #98, opened 2018-02-14
Make layer builder create reproducible layers.

loosebazooka:
- Strip out non-reproducible properties from tarArchiveEntry.
- Allow the user to toggle this feature through build config.

This is a first run, there still some work to do. Decisions in this PR made but I'm not sure about:
- default for reproducibility = false
- reproducibility is configurable. 
- the defaults for file permission overwrites (as described in #7)

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
@@ -49,6 +49,9 @@ public class BuildConfiguration {
     /** Known registry credentials to fallback on. */
     KNOWN_REGISTRY_CREDENTIALS(false),
 
+    /** Tell the layer build to use reproducibility features. */
+    ENABLE_REPRODUCIBLE_BUILDS(false),
+
     /** The main class to use when running the application. */
     MAIN_CLASS(true),
 
@@ -83,6 +86,7 @@ public class BuildConfiguration {
             put(Fields.TARGET_REPOSITORY, "target image name");
             put(Fields.TARGET_TAG, "target image tag");
             put(Fields.CREDENTIAL_HELPER_NAMES, "credential helper names");
+            put(Fields.ENABLE_REPRODUCIBLE_BUILDS, "enable reproducible layer builds");
             put(Fields.MAIN_CLASS, "main class");
             put(Fields.JVM_FLAGS, "JVM flags");
             put(Fields.ENVIRONMENT, "environment variables");
@@ -101,6 +105,7 @@ public class BuildConfiguration {
       values.put(Fields.KNOWN_REGISTRY_CREDENTIALS, RegistryCredentials.none());
       values.put(Fields.JVM_FLAGS, Collections.emptyList());
       values.put(Fields.ENVIRONMENT, Collections.emptyMap());
+      values.put(Fields.ENABLE_REPRODUCIBLE_BUILDS, false);
     }
 
     public Builder setBuildLogger(BuildLogger buildLogger) {
@@ -149,6 +154,11 @@ public class BuildConfiguration {
       if (knownRegistryCredentials != null) {
         values.put(Fields.KNOWN_REGISTRY_CREDENTIALS, knownRegistryCredentials);
       }
+      return this;
+    }
+
+    public Builder setEnableReproducibleBuilds(boolean enable) {
+      values.put(Fields.ENABLE_REPRODUCIBLE_BUILDS, enable);
       return this;
     }
 
@@ -260,6 +270,10 @@ public class BuildConfiguration {
 
   public List<String> getCredentialHelperNames() {
     return getFieldValue(Fields.CREDENTIAL_HELPER_NAMES);
+  }
+
+  public boolean getEnableReproducibleBuilds() {
+    return getFieldValue(Fields.ENABLE_REPRODUCIBLE_BUILDS);
   }
 
   public String getMainClass() {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushBlobStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushBlobStep.java
@@ -61,6 +61,9 @@ class PushBlobStep implements Callable<Void> {
               .setTimer(timer);
 
       if (registryClient.checkBlob(layerDigest) != null) {
+        buildConfiguration
+            .getBuildLogger()
+            .info("BLOB : " + layerDigest + " already exists on registry");
         return null;
       }
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerBuilder.java
@@ -22,6 +22,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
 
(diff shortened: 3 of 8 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/image/LayerBuilderTest.java
```diff
@@ -96,19 +119,54 @@ public void testBuild() throws URISyntaxException, IOException {
...
+    Assert.assertThat(layerHash, CoreMatchers.not(notReproducedHash));
+  }
 
-      Assert.assertEquals(expectedFileString, extractedFileString);
-    }
+  private Path createFile(Path root, String filename, String content, long lastModifiedTime)
+      throws IOException {
+    // TODO : add file permissions and other stuff, to make sure we're stripping them.
+    return Files.setLastModifiedTime(
+        Files.write(root.resolve(filename), content.getBytes(), StandardOpenOption.CREATE_NEW),
```
coollog: s/getBytes()/getBytes(StandardCharsets.UTF_8)

jib-core/src/test/java/com/google/cloud/tools/jib/image/LayerBuilderTest.java
```diff
@@ -96,19 +119,54 @@ public void testBuild() throws URISyntaxException, IOException {
...
+
+    Assert.assertThat(layerHash, CoreMatchers.is(reproducedHash));
+    Assert.assertThat(layerHash, CoreMatchers.not(notReproducedHash));
+  }
 
-      Assert.assertEquals(expectedFileString, extractedFileString);
-    }
+  private Path createFile(Path root, String filename, String content, long lastModifiedTime)
+      throws IOException {
+    // TODO : add file permissions and other stuff, to make sure we're stripping them.
```
coollog: File permissions probably shouldn't be stripped right? They probably shouldn't vary between builds on different machine.

jib-core/src/test/java/com/google/cloud/tools/jib/image/LayerBuilderTest.java
```diff
@@ -96,19 +119,54 @@ public void testBuild() throws URISyntaxException, IOException {
...
+    Blob layer =
+        new LayerBuilder(Arrays.asList(fileA1, fileB1), extractionPath, true).build().getBlob();
+    Blob reproduced =
+        new LayerBuilder(Arrays.asList(fileB2, fileA2), extractionPath, true).build().getBlob();
+
+    // this is a control layer that is the same as 'layerReproduced' without reproducibility on
+    Blob notReproduced =
+        new LayerBuilder(Arrays.asList(fileB2, fileA2), extractionPath, false).build().getBlob();
+
+    MessageDigest sha256 = MessageDigest.getInstance("SHA-256");
```
coollog: There's the CountingDigestOutputStream that is used in other tests for generating layer hashes.

For example:

```java
CountingDigestOutputStream countingDigestOutputStream = new CountingDigestOutputStream(ByteStreams.nullOutputStream())
layer.writeTo(countingDigestOutputStream);
BlobDescriptor layerBlobDescriptor = countingDigestOutputStream.toBlobDescriptor();
```

Although, here it seems that it might be simpler to just check the equality of `Blobs.writeToString()`?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -39,6 +40,10 @@
   /** Writes each entry in the filesystem to the tarball archive stream. */
   private static void writeEntriesAsTarArchive(
       List<TarArchiveEntry> entries, OutputStream tarByteStream) throws IOException {
+
+    List<TarArchiveEntry> sorted = new ArrayList<>(entries);
```
coollog: Since `LayerBuilder` already sorts the entries, is this necessary?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerBuilder.java
```diff
@@ -64,28 +70,42 @@ public UnwrittenLayer build() throws IOException {
                   for (Path sourceFileRelativePathComponent : sourceFileRelativePath) {
                     subExtractionPath.append('/').append(sourceFileRelativePathComponent);
                   }
-                  filesystemEntries.add(
-                      new TarArchiveEntry(path.toFile(), subExtractionPath.toString()));
+                  TarArchiveEntry tarArchiveEntry =
+                      new TarArchiveEntry(path.toFile(), subExtractionPath.toString());
+                  stripNonReproducibleElements(tarArchiveEntry);
```
coollog: This should be under the `enableReproducibleBuilds` flag as well?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerBuilder.java
```diff
@@ -64,28 +70,42 @@ public UnwrittenLayer build() throws IOException {
...
+                  filesystemEntries.add(tarArchiveEntry);
                 });
         continue;
       }
 
       TarArchiveEntry tarArchiveEntry =
           new TarArchiveEntry(sourceFile.toFile(), extractionPath + "/" + sourceFile.getFileName());
-      tarArchiveEntry.setModTime(0);
+      if (enableReproducibleBuilds) {
```
coollog: Perhaps L81-86 can be moved to a separate method that also replaces L73-76.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerBuilder.java
```diff
@@ -64,28 +70,42 @@ public UnwrittenLayer build() throws IOException {
...
     // Adds all the files to a tar stream.
+    TarStreamBuilder tarStreamBuilder = new TarStreamBuilder();
     for (TarArchiveEntry entry : filesystemEntries) {
       tarStreamBuilder.addEntry(entry);
     }
 
     return new UnwrittenLayer(tarStreamBuilder.toBlob());
   }
 
+  private void stripNonReproducibleElements(TarArchiveEntry tarArchiveEntry) {
```
coollog: This could be static, or be a part of the common non-static method for making a `TarArchiveEntry` as mentioned in the previous comment.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushBlobStep.java
```diff
@@ -61,6 +61,7 @@ public Void call()
               .setTimer(timer);
 
       if (registryClient.checkBlob(layerDigest) != null) {
+        buildConfiguration.getBuildLogger().info("SKIPPING");
```
coollog: Perhaps a message like `<layerDigest> exists, skipping` may be more informative

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -49,6 +49,9 @@
     /** Known registry credentials to fallback on. */
     KNOWN_REGISTRY_CREDENTIALS(false),
 
+    /** Tell the layer build to use reproducibility features. */
+    REPRODUCIBLE_BUILDS(false),
```
coollog: nit: Perhaps `ENABLE_REPRODUCIBLE_BUILDS` (and the getter/setter) to keep it consistent with the usage in `LayerBuilder`?

jib-core/src/test/java/com/google/cloud/tools/jib/image/LayerBuilderTest.java
coollog: Actually, it might not be completely safe to write the tar gzip BLOB to a string and check that. Perhaps it'd be better to `blob.writeTo` some byte array and then directly compare those?

jib-core/src/test/java/com/google/cloud/tools/jib/image/LayerBuilderTest.java
loosebazooka: Oops, I meant ownership not permissions.

coollog: Ah okay, what should be the stripped version of ownership? Just like root?

Issue #96, opened 2018-02-14
Support build OCI image.


coollog:
- [x] Add support for OCI images. (#108)
- [ ] Add configuration to build/push as OCI format. (#117)

PR #108, opened 2018-02-20
Adds support for OCI images.

coollog:
Part of #96

- Adds `OCIManifestTemplate` to be like `V22ManifestTemplate`, except with OCI-specific media types.
- Adds `BuildableManifestTemplate` to encompass both `OCIManifestTemplate` and `V22ManifestTemplate`.

Still need to:
- Add configuration to build/push as OCI format.


loosebazooka:
The structure is a little weird to me, feels like if they are the exact same template, then it should just be one file, and if there are different (OCI has more params?) then it should be two separate classes? If V22 is just a subset of OCI, then it should be able to be handled by the OCI class?

Would this kind of structure work here?

```
interface BuildableManifest {
  getSchemaVersion
  getMediaType
  getConfig
  getLayers
}

refactor out the common classes -- layers, config

class V22Manfiest {
  schemaversion
  mediatype
  config
  layers
}

class OCIManfiest {
  schemaversion
  mediatype
  config
  layers
  ... anything else?
}
```

PR #109, opened 2018-02-20
Fixes CacheReaderTest.

coollog:
`CacheReaderTest` needed to use a temporary cache directory that copies over the contents of the test resource cache directory because it is overwriting the metadata JSON file in there.


loosebazooka:
looks like you need to run the formatter on this.

PR #114, opened 2018-02-21
Refactors BuildConfiguration.

### LINKED ISSUE #103: Refactor BuildConfiguration to be more extensible.
The enum in BuildConfiguration should be replaced with member variables and validation.

coollog:
Fixes #103 

- Each configuration parameter now extends `ConfigurationParameter` with custom logic.
- Validation is handled by a `ConfigurationParameterValidator` class.
- Adding a new parameter now involves:
  1. Create a new {@link ConfigurationParameter} subclass.
  1. Add the new subclass to this set.
  1. Add the appropriate setter to {@link Builder}.
  1. Add the appropriate getter to {@link BuildConfiguration}.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/configuration/BuildConfiguration.java
```diff
@@ -0,0 +1,215 @@
...
+import com.google.cloud.tools.jib.builder.BuildLogger;
+import com.google.cloud.tools.jib.image.ImageReference;
+import com.google.cloud.tools.jib.registry.credentials.RegistryCredentials;
+import com.google.common.collect.ImmutableSet;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Immutable configuration options for the builder process. */
+public class BuildConfiguration {
```
loosebazooka: I was thinking this is more just a simple value/builder class?

Is there a reason we can't just use something like

```
public class BuildConfiguration {
   ImageReference imageReference;
   ... getImageReference()

   builder {
     build() {
        // validate all required params
        if (imageReference == null) -> then print out message?
    }

```

coollog: Looks like I overcomplicated this haha.

PR #117, opened 2018-02-22
Adds configuration for building and pushing as OCI image format.

### LINKED ISSUE #96: Support build OCI image.

coollog:
Fixes #96 

The `jib-maven-plugin` configuration is called `imageFormat` and can take `Docker` (default) or `OCI`.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -90,6 +109,9 @@
   @Parameter(defaultValue = "true", required = true)
   private boolean enableReproducibleBuilds;
 
+  @Parameter(defaultValue = "Docker", required = true)
+  private ImageFormat imageFormat;
```
loosebazooka: this doesn't allow variations in capitalizations, right? It must be `Docker` and not `docker`?

coollog: Yea, it is case sensitive right now. Should we have it be insensitive?

PR #118, opened 2018-02-22
Removes fmt:format from validate phase

### LINKED ISSUE #100: Remove fmt:format from the validate phase.
Having `fmt:format` run automatically means that if there is a formatting error, the `kokoro/continuous.*` build scripts will still pass.

`fmt` is the `fmt-maven-plugin` in `jib-maven-plugin/pom.xml`. It is currently attached to the `validate` phase.

`CONTRIBUTING.md` will need to be updated with adding `fmt:format` to the `jib-maven-plugin` verify command.

TadCordle:
Fixes #100.

### DIFF
--- a/jib-maven-plugin/pom.xml
+++ b/jib-maven-plugin/pom.xml
@@ -271,10 +271,9 @@
         </dependencies>
         <executions>
           <execution>
-            <id>default-cli</id>
             <phase>validate</phase>
             <goals>
-              <goal>format</goal>
+              <goal>check</goal>
             </goals>
           </execution>
         </executions>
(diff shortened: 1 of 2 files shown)


jib-maven-plugin/pom.xml
```diff
@@ -271,10 +271,9 @@
         </dependencies>
         <executions>
           <execution>
-            <id>default-cli</id>
             <phase>validate</phase>
             <goals>
-              <goal>format</goal>
+              <goal>check</goal>
```
coollog: The `maven-checkstyle-plugin` below handles the google style check, so I believe we could just remove the entire `<executions>` block here.

TadCordle: Fixed

coollog: Looks like we actually should have fmt:check be run in validate as the checkstyle doesn't catch everything haha.

PR #121, opened 2018-02-23
Adds jib:dockercontext to export a Docker context.

### LINKED ISSUE #49: Support export into a Docker context and generate Dockerfile.
Generate Dockerfile and all necessary files in a folder that can be built into a Docker image with `docker build .`.

coollog:
Fixes #49 

Takes additional configuration `targetDir` as the output directory.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/EntrypointBuilder.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/** Builds an image entrypoint for the Java application. */
+public class EntrypointBuilder {
+
+  /**
+   * Builds the container entrypoint.
+   *
+   * <p>The entrypoint is {@code java -cp [classpaths] [main class]}.
```
loosebazooka: is it `java [jvmFlags] -cp [classpaths] [main class]`

coollog: Yes

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/DockerContextMojo.java
```diff
@@ -0,0 +1,223 @@
...
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.apache.maven.project.MavenProject;
+
+/** Exports to a Docker context. */
+@Mojo(name = "dockercontext", requiresDependencyResolution = ResolutionScope.RUNTIME_PLUS_SYSTEM)
+public class DockerContextMojo extends AbstractMojo {
```
loosebazooka: So this doesn't really call into jib much to actually build the context does it? It's mostly handled by the plugin?

coollog: Yea, this doesn't call into Jib's image builder at all. It just copies files around.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/DirectoryWalker.java
```diff
@@ -0,0 +1,58 @@
...
+package com.google.cloud.tools.jib.filesystem;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+/** Recursively applies a function to each file in a directory. */
+public class DirectoryWalker {
```
loosebazooka: I feel like we're re-implementing something from java.nio here

coollog: Yea, this is mostly just `Files#walk`, but is able to account for `IOException`s in the lambda so we don't have to have the Unchecked to checked conversion everywhere.

loosebazooka: Hah, that just sounds like maybe we should take another look at it?

coollog: Hmm, I've looked into how others handle exceptions in Streams and the only solution I've found so far is to throw an unchecked exception. Perhaps we should not use Files.walk?

Example:
https://www.oreilly.com/ideas/handling-checked-exceptions-in-java-streams

loosebazooka: Yeah, I don't know why, but their solution just seems so hacky. Creating this whole uncheckedIOException and then catching that unchecked exception and throwing the cause, it gives me a gross feeling. But I guess it is what it is.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/DockerContextMojo.java
loosebazooka: Is there an experimental or incubating maven annotation we can throw on this. Since it's not core jib, it'd be nice if we didn't commit to it.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/DirectoryWalker.java
coollog: Yea, it is definitely not an elegant way to do it. Another way is to do what Guava does for `ListenableFuture`s, which I believe is wrap all exceptions in an `ExecutionException` to be caught and handled by the user.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/DockerContextMojo.java
coollog: I'm not able yet to find an experimental/incubating annotation for Maven specifically like there is for Gradle. I added a comment though.

coollog:
@loosebazooka I am planning to replace other places that do `Files.walk` with use of `DirectoryWalker`, like in `CacheReader`, so that we avoid having to do the `UncheckedIOException` hack in multiple places. (https://github.com/google/jib/issues/141)

PR #124, opened 2018-02-26
Updates README for OCI support.

coollog:
After releasing `v0.1.4` (#129).


coollog:
Will actually update this to be for `v0.1.4`, so people use that version over `v0.1.3`, which has the #125 and #126 bugs.

Issue #29, opened 2018-01-29
Separate cache into two: base image layers and application layers.


coollog:
This allows for base image layers cache to be shared between projects. Also, doing `mvn clean` would only clear the application layer cache.

- [x] Use `$HOME/.jib-cache` as the default base image layers cache directory (#133)
- [x] Use separate `Caches` caches for caching base image layers and application layers.
- [x] Don't split caches with `jib-maven-plugin` configuration flag `useOnlyProjectCache` (#134)

PR #133, opened 2018-02-28
Adds default base image layers cache directory for Caches.

coollog:
Part of #29


jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
```diff
@@ -16,24 +16,57 @@
...
-  private final Cache applicationCache;
+public class Caches implements Closeable {
 
   /** Initializes a {@link Caches} with directory paths. */
   public static class Initializer {
 
-    private Path baseCacheDirectory;
+    /** The default directory for caching the base image layers, in {@code $HOME/.jib-cache/}. */
+    private static final Path DEFAULT_BASE_CACHE_DIRECTORY =
+        Paths.get(System.getProperty("user.home")).resolve(".jib-cache");
```
loosebazooka: This might be relevant here : https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/534

coollog: Ah okay, I will change this to use that method.

coollog: Filed a separate PR for that: https://github.com/google/jib/pull/154

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
```diff
@@ -48,11 +81,17 @@ public Initializer setApplicationCacheDirectory(Path applicationCacheDirectory)
...
     }
 
-    public Caches init() throws CacheMetadataCorruptedException, NotDirectoryException {
-      if (baseCacheDirectory == null || applicationCacheDirectory == null) {
+    public Caches init()
+        throws CacheMetadataCorruptedException, IOException, CacheDirectoryNotOwnedException {
+      if (applicationCacheDirectory == null) {
         throw new IllegalStateException(
-            "Must initialize cache with both base image layer cache directory and application image layer cache directory");
+            "Must initialize cache with an application image layer cache directory");
```
loosebazooka: I realize I don't know what this error means. It's a error on our side, right?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
```diff
@@ -48,11 +81,17 @@ public Initializer setApplicationCacheDirectory(Path applicationCacheDirectory)
...
+      }
+
+      if (DEFAULT_BASE_CACHE_DIRECTORY.equals(baseCacheDirectory)) {
+        ensureOwnership(DEFAULT_BASE_CACHE_DIRECTORY);
```
loosebazooka: How does the user deal with this, if this does indeed happen? They must go delete the cache, or something?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
coollog: Yeah, this ensures that we must have called `setApplicationCacheDirectory` before calling `init`.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
coollog: There are a few options:

1) Tell the user to delete that cache directory
2) Provide the user with a configuration option to set the base cache directory
3) Default to using the same cache directory as the application cache directory

(These options can also be combined.)

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
loosebazooka: I think we can ensure applicationCacheDirectory is set by forcing it as a parameter down on line 99?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
```diff
@@ -61,12 +100,21 @@ public static Initializer initializer() {
     return new Initializer();
```
loosebazooka: maybe initializer() -> newInitializer()

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
```diff
@@ -48,25 +83,41 @@ public Initializer setApplicationCacheDirectory(Path applicationCacheDirectory)
...
 
-  public static Initializer initializer() {
-    return new Initializer();
+  /**
+   * @param applicationCacheDirectory Cache for the application image layers - should be local to
+   *     the application project
+   * @return a new {@link Initializer} to initialize the caches.
+   */
+  public static Initializer newInitializer(Path applicationCacheDirectory) {
+    return new Initializer(applicationCacheDirectory);
```
loosebazooka: I think we need `null` check here or in constructor?

coollog: I've been using a convention of if the parameter is not annotated with `@Nullable` then it must be non-null to void writing the `@NonNull` annotation. Do you think this is a good convention?

loosebazooka: oh sure, does error prone catch this? I was thinking we would need to do a preconditions.notNull?

coollog: Filed https://github.com/google/jib/pull/156

coollog:
How does using `[user cache home]/google-cloud-tools-java/jib` as the base layers cache directory sound?

PR #134, opened 2018-02-28
Adds useOnlyProjectCache jib-maven-plugin configuration.

### LINKED ISSUE #29: Separate cache into two: base image layers and application layers.

coollog:
Fixes #29


jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
```diff
@@ -107,22 +106,31 @@ public static Initializer newInitializer(Path applicationCacheDirectory) {
...
+      throws CacheMetadataCorruptedException, IOException {
     applicationCache = Cache.init(applicationCacheDirectory);
-  }
 
-  @Override
-  public void close() throws IOException {
-    baseCache.close();
-    applicationCache.close();
+    // Ensures that only one Cache is initialized if using the same directory.
+    if (Files.isSameFile(baseCacheDirectory, applicationCacheDirectory)) {
```
loosebazooka: just curious, does this need to be syncrhonized or made threadsafe somehow?

coollog: Yes, most things in this code are designed to be thread-local and/or immutable but the `CacheMetadata` in a Cache is shared between threads. The only layer that mutates the `CacheMetadata` state is `CacheMetadata#addLayer`, which is `synchronized`.

PR #142, opened 2018-03-01
Adds error-prone to compile checks.

coollog:
https://github.com/google/error-prone
http://errorprone.info/docs/installation

Gives us better compile-time checks.

This PR also fixes everything that error-prone caught.

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSourceFilesConfiguration.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSourceFilesConfiguration.java
@@ -26,7 +26,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Stream;
-
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.project.MavenProject;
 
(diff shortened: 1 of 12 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSourceFilesConfiguration.java
```diff
@@ -60,22 +62,23 @@ private static Path replaceClassExtensionWithJava(Path path) {
...
-                  classesSourceDir.resolve(classesOutputDir.relativize(classesSourceDirFile));
-              if (Files.exists(correspondingSourceDirFile)) {
-                // Adds the file as a classes file since it is in the source directory.
-                classesFiles.add(classesOutputDirFile);
-              } else {
-                // Adds the file as a resource since it is not in the source directory.
-                resourcesFiles.add(classesOutputDirFile);
-              }
-            });
+    try (Stream<Path> classesOutputDirFileStream = Files.list(classesOutputDir)) {
```
elharo: neat

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSourceFilesConfiguration.java
```diff
@@ -60,22 +62,23 @@ private static Path replaceClassExtensionWithJava(Path path) {
...
+      classesOutputDirFileStream.forEach(
+          classesOutputDirFile -> {
```
elharo: Variable names could be cleaned up a bit. We try to avoid abbreviating. Maybe just classFile and sourceFile if that's what these are?

coollog: Did some renaming to get rid of all occurrences of `Dir` abbreviation.

PR #158, opened 2018-03-13
Adds GradleSourceFilesConfiguration.

### LINKED ISSUE #157: Add GradleSourceFilesConfiguration.
Like `MavenSourceFilesConfiguration`, but uses Gradle Project-specific config.

coollog:
Fixes #157


coollog:
@loosebazooka Yea, it was mostly just to keep things consistent between the two plugins. I'm trying to use static constructors for heavy constructions so I changed both to static construction.

coollog:
Looks like NullAway has a bug that is causing the build failure. Filed an issue with them: https://github.com/uber/NullAway/issues/133

PR #160, opened 2018-03-13
Adds main class inference.

### LINKED ISSUE #159: Pull main class from "jar" task.
In Gradle, the `jar` task defines properties for the output jar file, and is configured with a closure in `build.gradle` such as:

```groovy
jar {
    manifest {
        attributes 'Main-Class': 'com.test.HelloWorld'
    }
}
```

We want to default to using the main class defined here if the user doesn't specify a main class.

coollog:
Fixes #159


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  /** Extracts main class from 'jar' task, if available. */
+  @Nullable
+  String getMainClassFromJarTask() {
+    ArrayList<Task> jarTasks = new ArrayList<>(project.getTasksByName("jar", false));
+    if (jarTasks.size() != 1) {
+      return null;
+    }
+    return (String) ((Jar) jarTasks.get(0)).getManifest().getAttributes().get("Main-Class");
```
loosebazooka: Does this just return null if Main-Class is not specified?

coollog: Yep

PR #168, opened 2018-03-14
Changes copyright company to Google LLC.

coollog:


examples/helloworld/src/main/java/example/HelloWorld.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google Inc.
+ * Copyright 2018 Google LLC.
```
TadCordle: Not sure, but I think you should also add an "All Rights Reserved." afterwards. https://g3doc.corp.google.com/company/teams/opensource/copyright.md?cl=head

TadCordle: Not sure, but I think you should also add an "All Rights Reserved." afterwards. https://g3doc.corp.google.com/company/teams/opensource/copyright.md?cl=head#java

coollog: Ah okay adding that.

PR #180, opened 2018-03-20
Adds BuildImageTask configuration.

### LINKED ISSUE #179: Add configuration for jib-gradle-plugin.
`jib-gradle-plugin` will create a task called `jib` that is configurable with an extension of the same name like such:

```groovy
...
jib {
  from {
    image = ‘gcr.io/my-gcp-project/my-base-image’
    credHelper = ‘gcr’
  }
  to {
    image = ‘gcr.io/gcp-project/my-app:built-with-jib’
    credHelper = ‘ecr-login’
  }
  jvmFlags = [‘-Xms512m’, ‘-Xdebug’]
  mainClass = ‘com.mycompany.myproject.Main’
  reproducible = true
  format = OCI
}
tasks.jib.dependsOn build
...
```

A minimal configuration could look like:

```groovy
jib.to.image = ‘gcr.io/my-gcp-project/my-app:built-with-jib’
```

coollog:
Fixes #179


loosebazooka:
ugh, I should have mentioned these earlier, but they slipped my mind: 

- Do you know how `@option` could be used with the nested closures https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/options/Option.html

- Does the groovy closure config translate well if the user is using kotlin DSL to write their build.gradle.kts or do we have do something extra? https://github.com/gradle/kotlin-dsl

coollog:
- For `@Option` it seems that all properties in a closure are optional. I was trying to find how to make a certain property required but did not seem to find anything. Maybe the way I am evaluating the closures (with `ConfigureUtil`) is too low-level and there is some abstraction that includes required property checks.

- I tried playing around with the Kotlin DSL to get the jib task configuration to work, but it doesn't seem to recognize that `image` and `credHelper` are valid properties to set on the closure passed into the `to`/`from` methods. I tried following https://github.com/gradle/kotlin-dsl/blob/master/doc/getting-started/Closures.md but I couldn't get it to actually use the closure to configure until I realized that `to` is a built-in Kotlin method, so I tried changing the method name to `toOMG` to get it to work, but it is still not working (the closure seems to be null?).

This is my `build.gradle.kts`:

```kotlin
...
apply {
    plugin("com.google.cloud.tools.jib")
}

tasks {
    "jib"(com.google.cloud.tools.jib.gradle.BuildImageTask::class) {
        toOMG(closureOf<com.google.cloud.tools.jib.gradle.ImageConfiguration> {
            image = "gcr.io/qingyangc-sandbox/jibgradleimage"
            credHelper = "gcr"
        })
    }
}
```

The error that is currently happening:

```
Parameter specified as non-null is null: method Build_gradle$8$1$1.invoke, parameter $receiver
```

Perhaps we should just stick to flat configuration if Kotlin DSL can't support this type of nested configuration.

coollog:
Finally got it working. I was supposed to use `delegateClosureOf` instead of just `closureOf`. The configuration looks like:

```kotlin
tasks {
    "jib"(com.google.cloud.tools.jib.gradle.BuildImageTask::class) {
        to(delegateClosureOf<com.google.cloud.tools.jib.gradle.ImageConfiguration> {
            image = "gcr.io/qingyangc-sandbox/jibgradleimage"
            credHelper = "gcr"
        })
    }
}
```

However, even though I annotate the `image` and `credHelper` properties as `@Input`, the configuration doesn't fail if they aren't specified.

loosebazooka:
Hrmm, yeah I think we can ignore the `@Option` stuff, since people can just use their own system property to pass it in. We probably want the rest to feel normal with kotlin though.

Could `to`/`from` take `Action<ImageConfiguration>` instead of closure as the input?

loosebazooka:
I think I may have actually misled us. Extension is probably our best bet : https://docs.gradle.org/current/userguide/custom_plugins.html#sec:mapping_extension_properties_to_task_properties

coollog:
@loosebazooka Changed to use extension instead.

coollog:
The corresponding build.gradle.kts would look like:

```kotlin
apply {
    plugin("com.google.cloud.tools.jib")
}

configure<com.google.cloud.tools.jib.gradle.JibExtension> {
    to {
        image = "gcr.io/qingyangc-sandbox/jibgradleimage"
        credHelper = "gcr"
    }
}
```

With the plugin applied in a `plugins` block, the type-safe accessor for the extension should be available as just `jib`.

https://github.com/gradle/kotlin-dsl/blob/master/doc/getting-started/Configuring-Plugins.md

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -0,0 +1,90 @@
...
+import org.gradle.api.DefaultTask;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Optional;
+import org.gradle.api.tasks.TaskAction;
+
+/** Builds a container image. */
+public class BuildImageTask extends DefaultTask {
+
+  /** Linked extension that configures this task. Must be set before the task is executed. */
+  @Nullable private JibExtension extension;
```
loosebazooka: Oh this null check really makes our code weird huh. We can't guarantee set Extension is called with a non-null value before anything else is called?

It feels like the null checker is overlapping with gradle's own checking mechanisms.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -0,0 +1,90 @@
...
+  public Class<? extends BuildableManifestTemplate> getFormat() {
+    return Preconditions.checkNotNull(extension).getFormat();
+  }
+
+  @TaskAction
+  public void buildImage() {
+    // TODO: Implement.
+  }
+
+  void setExtension(JibExtension jibExtension) {
```
loosebazooka: So I think the gradle convention is to only set the members that you expect to use. If we only have one task, and one config though, you can use `@Nested` and delegate the configuring of what's an input or whatever directly on the extension.
```
@Nested
getExtension() {
}
```
Either that or configure the task directly with each param?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
coollog: Yeah, it is overlapping. I changed `setExtension` to `applyExtension` so it is applying the params directly.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
coollog: Since we probably want to share some configuration with a future `dockercontext` task, we probably don't want to have this be specific to just the `jib` task. Changed to configure each param directly.

coollog: Although we might want the `ImageConfiguration` to be a nested configuration rather than be separate parameters on the task (so that we have `to.image` and `to.credHelper` rather than `toImage` and `toCredHelper`, for eg.)

PR #186, opened 2018-03-22
Adds GradleBuildLogger.

coollog:
Adds a `lifecycle` method to the `BuildLogger` interface to be a higher-level logging than `info`.


loosebazooka:
I guess I never noticed this before, but shouldn't we try using a standard logging facade?

coollog:
@loosebazooka  The original reason for using a custom interface is that Maven and Gradle have different logging methods. Specifically, Maven uses its own Log class and Gradle extends the slf4j.Logger with extra methods (lifecycle and quiet).

PR #190, opened 2018-03-23
Applies config to build the image.

coollog:
Applies the `jib-gradle-plugin` configuration added in #180 to build the image with the `jib` task. Adds integration tests to match `jib-maven-plugin`.


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -0,0 +1,238 @@
...
+    return useOnlyProjectCache;
+  }
+
+  @TaskAction
+  public void buildImage() throws InvalidImageReferenceException, IOException {
+    // Asserts required inputs are not null.
+    Preconditions.checkNotNull(from);
+    Preconditions.checkNotNull(to);
+    Preconditions.checkNotNull(jvmFlags);
+    Preconditions.checkNotNull(format);
```
TadCordle: Why flag these as `@Nullable` if they can't be null?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  @Input
+  public boolean getUseOnlyProjectCache() {
+    return useOnlyProjectCache;
+  }
+
+  @TaskAction
+  public void buildImage() throws InvalidImageReferenceException, IOException {
+    // Asserts required inputs are not null.
```
TadCordle: This comment is a bit confusing ("required inputs" sounds like it means "required config params", even though the params it's checking aren't required). Maybe change this to be more along the lines of "Asserts defaults have been set on configuration parameters".

coollog: Oh wait, this was actually referring to the `@Input`-annotated closure-configurable parameters. Here, it asserts that the required `@Input`s (the ones that don't have `@Optional` and are non-primitives) are not `null`. That's why `mainClass` is not in this list.

coollog: Also, this PR is after #180 so I'll make a change there.

coollog: Scratch that, this change is actually introduced in this PR haha.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -71,11 +95,93 @@ public boolean getReproducible() {
...
+  @Input
+  public boolean getUseOnlyProjectCache() {
+    return useOnlyProjectCache;
+  }
+
   @TaskAction
-  public void buildImage() {
-    // TODO: Implement.
+  public void buildImage() throws InvalidImageReferenceException, IOException {
+    // Asserts required @Input parameters are not null.
```
loosebazooka: I don't think @Input parameters can be null unless they're optional.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -71,11 +95,93 @@ public boolean getReproducible() {
...
+    Preconditions.checkNotNull(jvmFlags);
+    Preconditions.checkNotNull(format);
+
+    ImageReference baseImageReference = ImageReference.parse(from.getImage());
+    ImageReference targetImageReference = ImageReference.parse(to.getImage());
+
+    ProjectProperties projectProperties = new ProjectProperties(getProject(), getLogger());
+
+    String mainClass = this.mainClass;
+    if (mainClass == null) {
```
loosebazooka: were we going to do that ${main-class} check thing here too?

jib-maven-plugin/src/test/resources/projects/empty/pom.xml
```diff
@@ -30,6 +30,7 @@
         <configuration>
           <registry>gcr.io</registry>
           <repository>jib-integration-testing/emptyimage</repository>
+          <tag>maven</tag>
```
loosebazooka: gradle?

kokoro/presubmit.sh
```diff
@@ -13,4 +13,4 @@ cd github/jib
 
 (cd jib-core; ./gradlew clean build integrationTest --info)
 (cd jib-maven-plugin; ./mvnw clean install -B -U -X)
-(cd jib-gradle-plugin; ./gradlew clean build integrationTest --info)
+(cd jib-gradle-plugin; ./gradlew clean build --info)
```
loosebazooka: We're intentionally not running integrations tests here, right?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
coollog: Yea, this is just so that NullAway does not barf (compile-time null-checking cannot infer guarantee of `@Input` parameters not being null).

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
coollog: The one regarding the unresolved ${start-class} in spring? That has not been implemented in jib-maven-plugin either yet.

kokoro/presubmit.sh
coollog: Yea, not for presubmit, since presubmit does not have credentials to the GCR account.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
loosebazooka: ah okay, lets track that.

Issue #196, opened 2018-03-26
Different projects on same registry using invalid credentials from Docker config


TadCordle:
Original comment: "Another thing I saw was if for example, the user does not define a credHelper, but Jib tried to use the Docker config for credentials, but the credentials in there were for a different GCP project than the project you want to push to, the build fails with no suggestions. We might want to explore providing better feedback in these cases where we have the same registry but the credentials retrieved are not correct."

coollog:
Will be taken care of with credential diagnosing options.

Issue #192, opened 2018-03-23
Build succeeds with invalid JVM flags


TadCordle:
When the user configures a bad JVM flag, e.g.
```
<configuration>
    ...
    <jvmFlags>
        <jvmFlag>-Xaaaaaaaaaaaaaaaaaaaaaaa</jvmFlag>
    </jvmFlags>
</configuration>
```
`jib:build` succeeds, and the problem doesn't show itself until the user tries to run the built image. It would probably be best for problems like this to be caught as early as possible.

briandealwis:
How could we validate the arguments?  Users may choose to use a different JVM like OpenJ9 from [AdoptOpenJDK's docker images](https://hub.docker.com/r/adoptopenjdk/openjdk8-openj9/) which [supports other arguments](https://www.eclipse.org/openj9/docs/cmdline_specifying/).

coollog:
This issue might not be fixable. What I could think of is to just give a warning for non-standard JVM flags, but then that is cluttering the output with warnings if the user intended to use those flags.

briandealwis:
I wonder if this would be better handled with behavioural testing of the running container with something like [goss](https://github.com/aelsabbahy/goss)/[serverspec](https://serverspec.org).

coollog:
@pilhuhn Yea, standard as in what shows up for `java -help` vs. non-standard as in what shows up for `java -X` and others like prefixed with`-XX:` or `-D`. For "standard" options, there is only a select few like `-javaagent` and `-verbose:...`, so Jib could give a warning if using unrecognized flags not prefixed with `-X` or `-D`.

@briandealwis Yea, this might just be something the developer would handle themselves by writing their own tests that run their jib-built containers.

patflynn:
I'd definitely hesitate to provide more strongly typed and validated behavior around jvm flags.  

I tend to agree with @coollog that jib shouldn't try to solve this problem.

TadCordle:
Yeah, this seems like a pretty old issue that doesn't actually need to be fixed. I think we should just close this.

coollog:
Closed until someone has a strong opinion to have Jib support this.

Issue #189, opened 2018-03-22
Infer main class by searching for class with main method.


coollog:
If main class is not configured, jib should search for a class with the `public static void main` method defined and use that class as the main class. If multiple classes are found, an error should be thrown.

This should be as a last-resort, OR if the provided main class is not a valid main class.

This fixes #170 

The goal of this is to provide an even more minimal configuration for a developer, where not even a JAR configuration is needed to find the main class.

coollog:
@TadCordle want to take this?

PR #197, opened 2018-03-26
Add build failure suggestions

### LINKED ISSUE #27: Add more suggestions for various build failures.
When running `mvn jib:build`, there are various failures caught in the `BuildImageMojo#buildImage` method. These catches provide extra feedback to the user to help them understand the issue and provide suggestions for how to fix the issue.

The actionables are to:

1. Figure out what other build failures there are, if any (compile a list and maybe add that below)
2. Add similar handling for these build failures.
3. (Maybe improve how these failures are handled currently)

Build failures to handle:
- [ ] ~~Could not find or load main class~~ Just noticed #189 is already a thing
- [x] "Unknown host" when registry is misconfigured
- [x] "Obtaining project build output files failed" when project hasn't been compiled yet (e.g. running `mvn clean jib:build`)
- [x] "Unable to parse configuration of mojo: Cannot convert `x` to Enum" if imageFormat isn't `Docker` or `OCI`
- [ ] Unrecognized JVM flag (See #192)

TadCordle:
Fixes #27

Adds suggestions for build failures involving:
* "Unknown host" exception when `registry` is misconfigured
* Exception thrown when build output files are missing
* Misconfigured `imageFormat` parameter

Problems left to address:
* Different projects on same registry using invalid credentials from Docker config (#196)
* Build succeeds when main class is invalid (#189)
* Build succeeds with invalid JVM flags (#192)

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
@@ -38,6 +38,7 @@ import com.google.cloud.tools.jib.registry.credentials.RegistryCredentials;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import java.io.IOException;
+import java.net.UnknownHostException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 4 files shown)


TadCordle:
@coollog ping

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -286,6 +296,19 @@ private void validateParameters() throws MojoFailureException {
         throw new MojoFailureException("Invalid configuration parameters");
       }
     }
+    // Validates 'imageFormat'
+    try {
+      imageFormatToEnum = ImageFormat.valueOf(imageFormat);
```
coollog: We probably want to have this be more local to the place it is used and not as a field (ie. perform the check for if `imageFormat` is part of the enum (therefore not depending on catching an `IllegalArgumentException`), but call `ImageFormat.valueOf` above where the configuration is set on the `BuildConfiguration`)

coollog:
Also, a good practice to follow is to include the change in `CHANGELOG.md` in the same PR so we don't forget to add it to the CHANGELOG later. Something simple like "Better feedback for build failures" is okay since the details of the feedback is not too important to users reading the CHANGELOG (compared to if we were to make a breaking change to the configuration).

PR #198, opened 2018-03-26
Validate JVM flags at build time

### LINKED ISSUE #192: Build succeeds with invalid JVM flags

TadCordle:
Fixes #192.

A couple things I can foresee being problems:
* `JAVA_HOME` may not be set in path
* Java in base image may be different from user's java, and supported jvm flags may differ, which means this check may fail?

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
@@ -38,9 +38,11 @@ import com.google.cloud.tools.jib.registry.credentials.RegistryCredentials;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import java.io.IOException;
+import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


TadCordle:
The method itself doesn't matter so much, I'm just trying to find a way to warn about invalid parameters as early as possible (currently build succeeds and the user doesn't see that they misconfigured jvm flags/main class until they try to run their image). If there is a better way I'm all for it.

loosebazooka:
Ah okay, yeah starting a process seems a little heavy weight, perhaps there's something from the Java community that can do this for us?

patflynn:
There's a bunch of stuff to think about here.  When the user doesn't specify their own base we should be in complete control of the jdk location and version. It would be weird for the user to specify java home for that case.  We could also conceivably have more typesafe jvm configuration parameters if the distroless java version is stable with regard to the plugin version.

TadCordle:
I see what you mean, I guess it doesn't make sense to check against the user's java installation instead of the java in the base image. As for "more typesafe" parameters, do you mean something like keeping an up-to-date list of jvm flags to check against instead of checking against the java binary itself? That'd be much more lightweight, although it could be a pain to maintain and may not be compatible if the user specifies their own base image.

patflynn:
I meant conceivably you could expose the most common params in the config:

<MaxHeapSize>..</MaxHeapSize>
<ThreadStackSize>....

I haven't convinced myself this is a good idea yet, but it's just stuff to
think about.  It would definitely be nice to have something like:
<debug>
     <port>5050</port>
     <server>y</server>
     <suspend>n</server>
</debug>

Come to think of it I'm wondering if I'd ever want to bake any of this into
the image build instead of some means of running the image..

TadCordle:
Yeah, definitely more to think about. I'm gonna close this and we'll revisit it later.

patflynn:
Tad could you schedule a meeting with the team to discuss image
parameterization for when Q is back in the office?

PR #209, opened 2018-04-04
Re-release jib-gradle-plugin v0.1.0.

coollog:
The version in `gradle.properties` needs to be changed rather than the version in the `pluginsBundle`.

Tag `v0.1.0-gradle` has been updated.


coollog:
The plugin works now!

PR #211, opened 2018-04-04
Adds prepare_release script.

coollog:
After #209 

This is pretty-much the same as https://github.com/GoogleCloudPlatform/app-gradle-plugin/blob/master/scripts/prepare_release.sh except that it also pushes the tag.


jib-gradle-plugin/build.gradle
```diff
@@ -1,11 +1,24 @@
 plugins {
+    // Gradle API
     id 'java-gradle-plugin'
+
+    // Checkstyle
```
loosebazooka: Are all these comment necessary?

coollog: Not really, I'll remove the obvious ones.

Issue #208, opened 2018-04-04
Update docs for jib-gradle-plugin.


coollog:
- [x] Add `README.md` to `jib-gradle-plugin`. (#210)
- [x] Add `README.md` to `jib-maven-plugin`. (#219)
- [x] Change root `README.md` to be project description and link to each specific plugin's `README`. (#220)
- [x] Move FAQ to Wiki

PR #220, opened 2018-04-13
 Generifies root README and links to plugin-specific READMEs.

coollog:
Third part of #208 

Requires #219 .


README.md
```diff
@@ -305,21 +87,33 @@ Jib packages your Java application into the following paths on the image:
 
 Running commands like `apt-get` slows down the container build process. We **do not recommend or support** running commands as part of the build. 
 
-However, if you need to run commands, you can build a custom base image. You can then use this custom base image in the `jib-maven-plugin` by adding the following configuration:
+However, if you need to run commands, you can build a custom base image.
```
TadCordle: Maybe this question can be split up into the respective maven and gradle readmes.

TadCordle: Oh wait, there already are `from` examples. Are the custom base image examples here really necessary then?

coollog: Not completely necessary, but just gives a direct pointer for readers looking for the specific solution.

TadCordle: It might be better to keep the FAQ concise though so people can easily scan for what they want.

You could be more specific without examples too, like changing the last sentence to "However, if you need to run commands, you can build a custom base image **and use it with the `from` configuration parameter**."

TadCordle:
For the [last question on L160](https://github.com/google/jib/blob/50b3133a5a6634ad34afb127dee5c440eed17c73/README.md#how-can-i-tag-my-image-with-a-timestamp), could you give a gradle example too?

README.md
coollog: I put it into a collapsible section so that it doesn't take up any extra space.

coollog: The FAQ will probably be moved to a Wiki page or something rather than remain on the README.

Issue #48, opened 2018-02-05
Support build to local Docker daemon directly.


coollog:
Please upvote if you would like this feature to be added.

TadCordle:
Implemented in #266.

PR #223, opened 2018-04-20
Proposal: Build to Docker Daemon

coollog:
Requires #222 .
For issue #48 .

### DIFF
--- a/proposals/docker_build.md
+++ b/proposals/docker_build.md
@@ -1,0 +1,45 @@
+# Proposal: Build to Docker Daemon
+
+## Motivation
+
+Currently, Jib builds and pushes container images to a Docker registry without the need for a Docker daemon. However, for development use cases where the developer does not have a registry set up but do have a Docker daemon available, they may wish to build to the Docker daemon directly.
+
+## Goals
+
+* Build to local Docker daemon
+* Build to [minikube](https://github.com/kubernetes/minikube) (remote) Docker daemon
+* Minimize extra configuration
+
+## Constraints
+
+* Building to local Docker daemon should be an extra feature, not the default
+* Should not depend on a Docker CLI
+
+## Intended Workflow
+
+Building to a Docker daemon should be as simple as calling the build task/goal (`jib:build` for Maven and `jib` for Gradle) with an additional parameter called `buildToDockerDaemon` enabled.
+
+This can also be enabled with a command-line property - `-Ddocker` for Maven and `-Pdocker` for Gradle.
+
+## Implementation
+
+There are three main parts to implementing this proposal:
+
+1. Find the Docker daemon.
+  - This should use the same Docker daemon that the `docker` CLI would.
+  - Remote Docker daemon can be configured with environment variables, such as those output by `minikube docker-env`. These include:
+    - `DOCKER_TLS_VERIFY`
+    - `DOCKER_HOST`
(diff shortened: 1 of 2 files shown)


proposals/docker_build.md
```diff
@@ -0,0 +1,45 @@
...
+* Minimize extra configuration
+
+## Constraints
+
+* Building to local Docker daemon should be an extra feature, not the default
+* Should not depend on a Docker CLI
+
+## Intended Workflow
+
+Building to a Docker daemon should be as simple as calling the build task/goal (`jib:build` for Maven and `jib` for Gradle) with an additional parameter called `buildToDockerDaemon` enabled.
```
loosebazooka: I wonder if we should just add a new task?

jib:buildToDockerDaemon?

loosebazooka: and if we do this, perhaps we can match the bazel dockerless builds and also produce an intermediate tar, which may or may not help us with simple skaffold integration?

coollog: Yeah, a new task does sound good so that we wouldn't need any extra configuration to clutter the default build+push workflow. And yeah, exporting the intermediate tar sounds good. Maybe we can instead do:

`jib:tar` to build the tar
`jib:load` to load the tar into the Docker daemon

This way, we could implement `jib:tar` first to enable integration with Skaffold.

coollog: Decided on `jib:buildDocker`

loosebazooka:
The non-process option would be kind of nice, but it might be worth it to be careful pulling in a complex dependency like docker-java, it itself has a large number of dependencies and we might not want to complicate a user's build dependencies for a single docker operation.

coollog:
Updated to use the `docker` CLI.

proposals/docker_build.md
```diff
@@ -0,0 +1,48 @@
...
+## Goals
+
+* Build to local Docker daemon
+* Build to [minikube](https://github.com/kubernetes/minikube) (remote) Docker daemon
+* Minimize extra configuration
+
+## Constraints
+
+* Building to local Docker daemon should be an extra feature, not the default
+* Should not depend on a Docker CLI
```
loosebazooka: I guess were doing this now?

proposals/docker_build.md
```diff
@@ -0,0 +1,48 @@
...
+2. Send the image tarball stream to the Docker daemon using [`docker load`](https://docs.docker.com/engine/reference/commandline/load/) CLI command.
+
+Note that currently, build, cache, and push are all in `BuildImageSteps`. In order to implement parts 2 and 3, these stages would need be modularized so that push can be replaced with build image tarball stream and send to Docker daemon.
+
+If in the future, we would like to not use the `docker` CLI, we would need to:
+
+1. Find the Docker daemon.
+  - This should use the same Docker daemon that the `docker` CLI would.
+    - The default host is `unix:///var/run/docker.sock` or `tcp://127.0.0.1:2375`.
+  - Remote Docker daemon can be configured with environment variables, such as those output by `minikube docker-env`. These include:
```
loosebazooka: I guess we have to allow the user to configure this on our process?

coollog: I was thinking it would just use the environment variables Jib was run in?

proposals/docker_build.md
coollog: Removed.

PR #226, opened 2018-04-20
Adds auth configuration.

### LINKED ISSUE #215: Add auth to jib-gradle-plugin.
Add configuration to use raw credentials directly in the plugin configuration as such:

```jib
jib {
  from {
    image = 'gcr.io/my-gcp-project/my-base-image'
    auth {
      username = SOME_USERNAME
      password = SOME_PASSWORD
    }
  }
}
```

The auth should be specific to the base image. `SOME_USERNAME` and `SOME_PASSWORD` can be a raw string, but should be defined in `gradle.properties` and not committed to VC.

coollog:
Fixes #215


jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/DirectoryWalker.java
```diff
@@ -66,7 +66,7 @@ public DirectoryWalker filter(@Nullable Predicate<? super Path> pathFilter) {
       if (pathFilter != null) {
         filteredFileStream = fileStream.filter(pathFilter);
       }
-      return filteredFileStream.collect(Collectors.toList());
+      return filteredFileStream.sorted().collect(Collectors.toList());
```
loosebazooka: this seems unnecessary to the feature? Did we need to do this for something?

coollog: Yea, this is actually to fix a bug. I should have probably filed a separate PR for this but since it was just one line of code I just included it here. This file stream needs to be sorted because for reproducibility, even though we are sorting the list of files that are input into each layer, we weren't sorting the subfiles of any directories, and the file walker does not always walk in an alphabetical order.

PR #231, opened 2018-04-23
 Updates README for Gradle v0.1.1.

coollog:


jib-maven-plugin/README.md
```diff
@@ -1,5 +1,6 @@
 [![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/com.google.cloud.tools/jib-maven-plugin/badge.svg)](https://maven-badges.herokuapp.com/maven-central/com.google.cloud.tools/jib-maven-plugin)
+[![Gradle Plugin Portal](https://img.shields.io/badge/gradle%20plugin-v0.1.1-blue.svg)](https://plugins.gradle.org/plugin/com.google.cloud.tools.jib)
```
TadCordle: Should this be in all 3 READMEs?

coollog: Yeah, the reasoning is just to provide as many links to the specific plugins as possible in the root project README. This way, readers get a quick way to see that plugins for both Maven and Gradle are available.

TadCordle: So shouldn't this go in the root README, not the jib-maven-plugin README?

TadCordle: Ok, didn't notice you'd added it to both here and the root in this PR

coollog: You're right, I put it in the wrong place haha.

Issue #204, opened 2018-04-03
Add Docker context export feature.


coollog:
`jib-maven-plugin` has goal `jib:dockercontext` that should be supported in `jib-gradle-plugin` as a new task.

PR #233, opened 2018-04-24
Moves Docker context generation to jib-core.

coollog:
Refactors for #204 

Moves much of what was in `DockerContextMojo` to `jib-core/docker/DockerContextGenerator`. The functionality remains the same. This preps for adding a task to `jib-gradle-plugin` to generate the context as well.


jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/FileOperations.java
```diff
@@ -0,0 +1,48 @@
...
+
+package com.google.cloud.tools.jib.filesystem;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+/** Static methods for operating on the filesystem. */
+public class FileOperations {
```
loosebazooka: If we don't plan on instantiating this, adding a private, no arg constructor is useful.

PR #235, opened 2018-04-24
Updates NullAway and fixes package for Maven.

coollog:
The annotated package configuration for `jib-maven-plugin` was mistakenly `com.uber`, so it wasn't actually doing anything.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -276,6 +293,10 @@ private Authorization getRegistryCredentialsFromSettings(String registry) {
 
   /** Checks validity of plugin parameters. */
   private void validateParameters() throws MojoFailureException {
+    // These @Nullable parameters should never be null.
+    Preconditions.checkNotNull(repository);
```
loosebazooka: Doesn't `required` on `@parameter` handle this? or are we doing some processing here?
https://maven.apache.org/plugin-tools/maven-plugin-annotations/apidocs/org/apache/maven/plugins/annotations/Parameter.html#required()

coollog: NullAway isn't able to process the `@Parameter` annotation to know that it is not nullable.

Issue #212, opened 2018-04-06
Consolidate the configuration naming between Maven and Gradle.


coollog:
Currently, these differ, but should be made as similar as possible.

PR #236, opened 2018-04-25
Proposal: Align jib-maven-plugin configuration with jib-gradle-plugin.

coollog:
Addresses #212 

Preview at https://github.com/GoogleContainerTools/jib/blob/proposal-maven-config/proposals/maven_configuration_v2.md


proposals/maven_configuration_v2.md
```diff
@@ -0,0 +1,93 @@
...
+}
+```
+
+## Desired Configuration Schema
+
+Therefore, the desired `jib-maven-plugin` configuration should look like:
+
+```xml
+<configuration>
+  <from>
```
loosebazooka: I remember this was actually kind of a pain to do? Have you looked at the implementation of it?

coollog: I think this can just be done with a class as the parameter type.

For example, in `docker-maven-plugin`: https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/AbstractDockerMojo.java#L166\

The nested configuration class: https://github.com/fabric8io/docker-maven-plugin/blob/master/src/main/java/io/fabric8/maven/docker/config/RegistryAuthConfiguration.java

loosebazooka: oh weird, so `@Parameter` will implictly trigger nested setting of parameters? and we can set defaults on the inner members?

coollog: Yeah, it might need to implement `Serializable` too.

PR #238, opened 2018-04-26
Refactors out common plugin configuration.

coollog:
First part of #237


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -39,88 +38,48 @@
...
-  }
-
-  @Input
-  public boolean getReproducible() {
-    return reproducible;
-  }
-
-  @Input
+  /** Converts an {@link ImageConfiguration} to an {@link Authorization}. */
+  @Internal
```
loosebazooka: does gradle complain that this is a getter? I'm not sure if my memory is correct, but the `@Internal` doesn't seem required here?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Nested;
+import org.gradle.api.tasks.Optional;
+import org.gradle.api.tasks.TaskAction;
+
+/**
+ * Defines the configuration parameters injected from {@link JibExtension}. {@link Task}s should
+ * extend this class.
+ */
+abstract class TaskConfiguration extends DefaultTask {
```
loosebazooka: Perhaps this can be a `@Nested` input type to the tasks that are extending it. `ImageConfig`? and you pass that to the task directly. Subjectively, I think avoiding this style of extending makes for better reading.

coollog: Okay I'll just change it for Gradle then

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
coollog: Oh actually nvm, it is actually still barfing that it is trying to find the "imageAuthorization" property.

PR #239, opened 2018-04-26
Refactors and moves exception building to jib-core.

coollog:
Second part of #237


loosebazooka:
Can you point to some code where this will be used?

coollog:
@loosebazooka This was mostly intended as part of moving out common logic between maven/gradle to core in order to reduce code duplication and make the plugin "wrappers" thinner. The goal is to have the plugin logic just be adapters from core to maven/gradle.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulExceptionBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+  public E withSuggestion(Throwable cause, String suggestion) {
+    StringBuilder message = getMessageHeader();
+    message.append(", perhaps you should ");
+    message.append(suggestion);
+    return makeException(message.toString(), cause);
+  }
+
+  /** @return an {@link E} with just a cause */
+  public E withNoHelp(Throwable cause) {
+    return makeException(getMessageHeader().toString(), cause);
```
loosebazooka: could this just be (messageHeader, cause), seems a little odd that we'd take the stringbuilder version and immediately string it.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulExceptionBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+  private final String messageHeader;
+
+  /** @param messageHeader the initial message text for the exception message */
+  public HelpfulExceptionBuilder(String messageHeader) {
+    this.messageHeader = messageHeader;
+  }
+
+  /** @return an {@link E} with a cause and a suggestion */
+  public E withSuggestion(Throwable cause, String suggestion) {
+    StringBuilder message = getMessageHeader();
```
loosebazooka: could just be new StringBuilder(messageHeader) ?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulExceptionBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /** @return an {@link E} with just a cause */
+  public E withNoHelp(Throwable cause) {
+    return makeException(getMessageHeader().toString(), cause);
+  }
+
+  /** Implement with constructing a specific wrapper exception class. */
+  protected abstract E makeException(String message, Throwable cause);
+
+  private StringBuilder getMessageHeader() {
```
loosebazooka: It doesn't appear like we really need this method.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -52,6 +52,9 @@
   /** {@code User-Agent} header suffix to send to the registry. */
   private static final String USER_AGENT_SUFFIX = "jib-gradle-plugin";
 
+  private static final HelpfulGradleExceptionBuilder GRADLE_EXCEPTION_BUILDER =
```
loosebazooka: If it's not immutable, I think you need to use cameCase on the member name.

coollog:
Changed to be just a message "builder" class.

PR #240, opened 2018-04-27
Proposal: Improve Jib Log Output

coollog:
Preview at https://github.com/GoogleContainerTools/jib/blob/proposal-log/proposals/log_output_v2.md


proposals/log_output_v2.md
```diff
@@ -0,0 +1,337 @@
...
+The information to include, in the order in which they should appear are:
+
+1. (info) Where inputs are inferred from, including
+    1. The main class
+1. (info) Files that went into each layer
+    1. Classes, resources, dependencies
+1. (lifecycle/info) Important build steps
+    1. Show what’s skipped?
+    1. Asynchronous steps (the order may vary between runs)
+        - [lifecycle] Retrieving credentials for <base image>...
```
TadCordle: FYI I think you need to escape these triangle brackets.

`<base image>` is showing up blank.

proposals/log_output_v2.md
```diff
@@ -0,0 +1,337 @@
...
+### Information to include
+
+The information to include, in the order in which they should appear are:
+
+1. (info) Where inputs are inferred from, including
+    1. The main class
+1. (info) Files that went into each layer
+    1. Classes, resources, dependencies
+1. (lifecycle/info) Important build steps
+    1. Show what’s skipped?
```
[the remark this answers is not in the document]
coollog: Skipped steps. The main problem here is that we have these choices:

1. Output before a step like "Running step..."
- Could output "Skipped step..." after evaluating a step
- Currently what is in the proposal

2. Output after a step like "Ran step..."
- Output "Skipped step..." if the step was skipped
- But, this way, if a step takes long, user does not know which step is actually running

3. Output before and after
- Might be too clutterred

proposals/log_output_v2.md
```diff
@@ -0,0 +1,337 @@
...
+```
+[lifecycle] Containerizing application to gcr.io/my-gcp-project/my-app...
+[lifecycle]
+[lifecycle] Retrieving credentials for gcr.io/distroless/java…
+[lifecycle] Retrieving credentials for gcr.io/my-gcp-project/my-app...
+[lifecycle] Getting base image gcr.io/distroless/java…
+[lifecycle] Building dependencies layer...
+[lifecycle] Building resources layer...
+[lifecycle] Building classes layer...
+[lifecycle] Finalizing…
```
loosebazooka: Finalizing seems like it doesn't do much here, we are already saying `build completed` on the next line. I think we should either make it more descriptive or remove it.

Perhaps `pushing layers?`

coollog: This part is pushing the container configuration and image manifest. I was thinking that these would be more technical detail than needed. Maybe like "Committing image..." or "Configuring metadata..." might work?

TadCordle: I like "configuring metadata", but I also like the info that the build process is finishing up. Maybe something like "Pushing final image configuration..." or just "Finalizing configuration..." could work.

coollog: How about `Finalizing container image...`? The reason I added this line is so that the user isn't misled to think that the classes layer takes the longest to build.

coollog: Or maybe just "Pushing..."? Since after "Building", "Pushing" would be what is expected.

coollog: @loosebazooka

PR #245, opened 2018-05-02
Implements Proposal: Address Base Image Reproducibility

### LINKED ISSUE #217: Use specific 'gcr.io/distroless/java' image digest.
To ensure reproducibility at the plugin version level, each plugin version should be tied to a specific release of the default `gcr.io/distroless/java` image.

For example, the current latest digest is [`@sha256:6af3b801e2beef75307d1eb683e2da792357cacb675a60c10ede16481d2cea741`](https://pantheon.corp.google.com/gcr/images/distroless/GLOBAL/java@sha256:6af3b801e2beef75307d1eb683e2da792357cacb675a60c10ede16481d2cea74/details/info).

### LINKED ISSUE #241: Implement Proposal: Address Base Image Reproducibility
Addresses proposal #224

Fixes #217

coollog:
https://github.com/GoogleContainerTools/jib/blob/master/proposals/reproducible_base_image.md

Fixes #241 
Fixes #217


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildAndCacheApplicationLayersStep.java
```diff
@@ -89,9 +89,7 @@
               return cachedLayer;
             }
 
-            LayerBuilder layerBuilder =
-                new LayerBuilder(
-                    sourceFiles, extractionPath, buildConfiguration.getEnableReproducibleBuilds());
+            LayerBuilder layerBuilder = new LayerBuilder(sourceFiles, extractionPath, true);
```
TadCordle: Is there any reason to keep this last parameter/reproducibility logic in LayerBuilder?

coollog: I kept this here since the LayerBuilder is low-level logic, so the trade-off of removing the explicit "true" here would be a comment added to LayerBuilder saying that it only builds reproducibly.

TadCordle: Another option could be to rename `LayerBuilder` to `ReproducibleLayerBuilder`, if you want to be explicit.

coollog: `ReproducibleLayerBuilder` sounds like a good idea

loosebazooka:
should you remove the proposal as part of this PR? Or move it to "archives" or something.

coollog:
Yeah "archives" sounds like a good idea.

PR #248, opened 2018-05-02
Separates BuildConfiguration image credentials between base and target.

### LINKED ISSUE #205: Change BuildConfiguration to take separate credential helpers for base and target images.
Rather than a list of credential helpers.

coollog:
Fixes #205 

This makes the changes in `jib-core` and has a temporary hack to make the tests pass in `jib-maven-plugin`. That temporary hack will be fixed in a follow-up (for #212/#242).


loosebazooka:
Does RegistryCredentials still need to hold multiple credentials after this change?

coollog:
Looks not so. I will remove that as well.

Issue #243, opened 2018-05-01
Implement Proposal: Build to Docker Daemon


coollog:
Addresses proposal #223

Fixes #48

TadCordle:
So here's my plan for this, not necessarily in this order. I'm hoping each of these checkboxes can be a separate PR.

- [x] Maven/gradle - Add empty tasks/goals
- [ ] Maven/gradle - Add task setup (reading/validating config params, e.g. stuff that happens before `BuildImageTask#doBuildImage`)
- [ ] Core - Refactor/modularize build and cache steps so they can be started separately from push functionality
- [ ] Core - Build tarball stream
- [ ] Core - Send tarball to Docker daemon
- [ ] Maven/gradle - Tie everything together into task/goal

@coollog Let me know if you think this is good approach, or if there's any important info I could be missing (still pretty unfamiliar with a lot of this).

coollog:
@TadCordle looks good!

loosebazooka:
I think it might be good practice to avoid exposing as a task/goal until it is ready. You can either build from core up to maven/gradle. Or start with skeletons in maven/gradle but don't expose them (don't do a `project.tasks.create` or `@Mojo`). In some cases, probably not ours since we're targeting our next release, exposing the incomplete task/goal makes it harder to do releases while this feature is being developed.

TadCordle:
Oh true, I already pushed the empty tasks, should I hide them in the next PR?

PR #250, opened 2018-05-03
Add empty task/goal for building to docker daemon

TadCordle:
Starting #243


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -0,0 +1,35 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.gradle;
+
+import javax.annotation.Nullable;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.tasks.TaskAction;
+
```
coollog: Should probably have a short class-level javadoc

PR #252, opened 2018-05-03
Break apart container configuration build and push step

TadCordle:
Separating build and push steps so we can build and export to Docker daemon.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildContainerConfigurationStep.java
```diff
@@ -112,17 +102,8 @@ private BlobDescriptor afterBaseImageLayerFuturesFuture()
...
-      timer.lap(
-          "Pushing container configuration " + containerConfigurationBlobDescriptor.getDigest());
-
-      // TODO: Use PushBlobStep.
-      // Pushes the container configuration.
-      registryClient.pushBlob(
-          containerConfigurationBlobDescriptor.getDigest(), containerConfigurationBlob);
 
-      return containerConfigurationBlobDescriptor;
+      return new BlobAndDigest(containerConfigurationBlob, digestOutputStream.toBlobDescriptor());
```
coollog: This could probably avoid creating a new tuple class `BlobAndDigest` by just returning the `Blob` here and generating the `Digest` in the push step.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildImageSteps.java
```diff
@@ -159,6 +160,19 @@ public void run()
...
+          timer2.lap("Setting up container configuration push");
+          // Pushes the container configuration.
+          ListenableFuture<ListenableFuture<BlobDescriptor>>
+              pushContainerConfigurationFutureFuture =
+                  Futures.whenAllSucceed(buildContainerConfigurationFutureFuture)
+                      .call(
+                          new PushContainerConfigurationStep(
+                              buildConfiguration,
+                              authenticatePushFuture,
+                              buildContainerConfigurationFutureFuture.get(),
```
coollog: This call to get would be executed before anything is run, which would cause race conditions and blocking. Any `get` for a `Future` should be called `call` method of the step that uses it and should be wrapped in a `NonBlockingFutures.get` for busy-wait safety (it makes sure that the `get` does not block).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildContainerConfigurationStep.java
TadCordle: Sounds good, I wasn't sure if the Digest generation should be kept in the build step.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushContainerConfigurationStep.java
```diff
@@ -0,0 +1,89 @@
...
+  private final ListeningExecutorService listeningExecutorService;
+
+  PushContainerConfigurationStep(
+      BuildConfiguration buildConfiguration,
+      ListenableFuture<Authorization> pushAuthorizationFuture,
+      ListenableFuture<ListenableFuture<Blob>> blobFuturesFuture,
+      ListeningExecutorService listeningExecutorService) {
+    this.buildConfiguration = buildConfiguration;
+    this.pushAuthorizationFuture = pushAuthorizationFuture;
+    this.blobFuturesFuture = blobFuturesFuture;
```
coollog: nit: *FutureFuture since it is just a future wrapping a since future

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushContainerConfigurationStep.java
```diff
@@ -0,0 +1,89 @@
...
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+
+class PushContainerConfigurationStep implements Callable<ListenableFuture<BlobDescriptor>> {
```
coollog: Short javadoc here

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushContainerConfigurationStep.java
coollog: Might also be clearer to name this like buildContainerConfigurationF... to make it explicit where this should be coming from.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildContainerConfigurationStep.java
```diff
@@ -59,11 +60,11 @@
   @Override
   public ListenableFuture<Blob> call() throws ExecutionException, InterruptedException {
     // TODO: This might need to belong in BuildImageSteps.
-    List<ListenableFuture<?>> afterBaseImageLayerFuturesFutureDependencies = new ArrayList<>();
-    afterBaseImageLayerFuturesFutureDependencies.addAll(
+    List<ListenableFuture<?>> afterBaseImageLayerFutureFutureDependencies = new ArrayList<>();
```
coollog: Oh this one is actually a future of futures (it refers to being after `pullBaseImageLayerFuturesFuture`)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushContainerConfigurationStep.java
```diff
@@ -78,7 +80,8 @@ private BlobDescriptor afterBlobFuturesFuture()
 
       CountingDigestOutputStream digestOutputStream =
           new CountingDigestOutputStream(ByteStreams.nullOutputStream());
-      Blob blob = NonBlockingFutures.get(NonBlockingFutures.get(blobFuturesFuture));
+      Blob blob =
```
coollog: Maybe `containerConfigurationBlob` for specificity?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushContainerConfigurationStep.java
```diff
@@ -0,0 +1,92 @@
...
+    this.buildConfiguration = buildConfiguration;
+    this.pushAuthorizationFuture = pushAuthorizationFuture;
+    this.buildContainerConfigurationFutureFuture = buildContainerConfigurationFutureFuture;
+    this.listeningExecutorService = listeningExecutorService;
+  }
+
+  /** Depends on {@code blobFuturesFuture} and {@code pushAuthorizationFuture}. */
+  @Override
+  public ListenableFuture<BlobDescriptor> call() throws ExecutionException, InterruptedException {
+    List<ListenableFuture<?>> blobFuturesFutureDependencies = new ArrayList<>();
```
coollog: This is a `after...FutureFuture` too. `afterBuildContainerConfigurationFutureFutureDependencies` would be the exact name (meaning the dependencies for the `afterBuildContainerConfigurationFutureFuture` method), but it is quite long.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushContainerConfigurationStep.java
```diff
@@ -0,0 +1,92 @@
...
+    blobFuturesFutureDependencies.add(pushAuthorizationFuture);
+    blobFuturesFutureDependencies.add(
+        NonBlockingFutures.get(buildContainerConfigurationFutureFuture));
+    return Futures.whenAllSucceed(blobFuturesFutureDependencies)
+        .call(this::afterBlobFuturesFuture, listeningExecutorService);
+  }
+
+  /** Depends on {@code blobFuturesFuture.get()} and {@code pushAuthorizationFuture}. */
+  private BlobDescriptor afterBlobFuturesFuture()
```
coollog: Same here.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildContainerConfigurationStep.java
```diff
@@ -18,81 +18,63 @@
...
     // TODO: This might need to belong in BuildImageSteps.
-    List<ListenableFuture<?>> afterBaseImageLayerFuturesFutureDependencies = new ArrayList<>();
-    afterBaseImageLayerFuturesFutureDependencies.add(pushAuthorizationFuture);
-    afterBaseImageLayerFuturesFutureDependencies.addAll(
+    List<ListenableFuture<?>> afterBaseImageLayerFutureFutureDependencies = new ArrayList<>();
+    afterBaseImageLayerFutureFutureDependencies.addAll(
         NonBlockingFutures.get(pullBaseImageLayerFuturesFuture));
-    afterBaseImageLayerFuturesFutureDependencies.addAll(buildApplicationLayerFutures);
-    return Futures.whenAllSucceed(afterBaseImageLayerFuturesFutureDependencies)
+    afterBaseImageLayerFutureFutureDependencies.addAll(buildApplicationLayerFutures);
```
coollog: This is actually a `FuturesFuture` since it is wrapping a list of futures.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/PushContainerConfigurationStep.java
TadCordle: How about `afterBuildConfigFutureFuture` so it isn't so long?

TadCordle: Although I guess "BuildConfig" can get confused with the BuildConfiguration that's passed in. Maybe `afterConfigBuildFutureFuture`.

coollog: Hmm yeah we try to avoid abbreviations, so not sure what the best practice here should be. @loosebazooka

loosebazooka: Genereally we want to avoid abbreviations.

PR #260, opened 2018-05-08
Add core functionality for build to docker daemon

TadCordle:


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Depends on {@code pushAuthorizationFuture}, {@code pushBaseImageLayerFuturesFuture.get()},
+   * {@code pushApplicationLayerFutures}, and (@code
+   * containerConfigurationBlobDescriptorFutureFuture.get()}.
+   */
+  private Void afterPushBaseImageLayerFuturesFuture()
+      throws ExecutionException, InterruptedException, IOException {
+    // Add layers to image tarball
+    TarStreamBuilder builder = new TarStreamBuilder();
```
coollog: Prob good to name this fully `tarStreamBuilder` since by like line 120, it might be hard to remember this was a `TarStreamBuilder`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,125 @@
...
+      builder.addEntry(new TarArchiveEntry(layerFile.toFile(), layerFile.getFileName().toString()));
+    }
+
+    // Add config to tarball
+    Blob containerConfigurationBlob =
+        NonBlockingFutures.get(NonBlockingFutures.get(buildConfigurationFutureFuture));
+    Path tempConfig = Files.createTempFile(null, null);
+    tempConfig.toFile().deleteOnExit();
+    containerConfigurationBlob.writeTo(new BufferedOutputStream(Files.newOutputStream(tempConfig)));
+    builder.addEntry(new TarArchiveEntry(tempConfig.toFile(), "config.json"));
```
coollog: We might want to add a functionality to `TarStreamBuilder` to add entries that are `Blob`s and not files, so that we don't have to create temporary files here. A TODO for now is good.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,125 @@
...
+
+    // Add manifest to tarball
+    ImageToJsonTranslator imageToJsonTranslator = new ImageToJsonTranslator(new Image());
+    Blob manifestBlob =
+        imageToJsonTranslator.getDockerLoadManifestBlob(
+            buildConfiguration.getTargetImageRepository(),
+            buildConfiguration.getTargetImageTag(),
```
coollog: `BuildConfiguration` actually stores the target `ImageReference`, so could add a getter for that like `getTargetImage`. Then `ImageReference#toString` would be sufficient to get the string format

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,125 @@
...
+      layerFiles.add(layerFile.getFileName().toString());
+      builder.addEntry(new TarArchiveEntry(layerFile.toFile(), layerFile.getFileName().toString()));
+    }
+
+    // Add config to tarball
+    Blob containerConfigurationBlob =
+        NonBlockingFutures.get(NonBlockingFutures.get(buildConfigurationFutureFuture));
+    Path tempConfig = Files.createTempFile(null, null);
+    tempConfig.toFile().deleteOnExit();
+    containerConfigurationBlob.writeTo(new BufferedOutputStream(Files.newOutputStream(tempConfig)));
```
coollog: These stream should be closed (try-with-resource style).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,125 @@
...
+            buildConfiguration.getTargetImageRepository(),
+            buildConfiguration.getTargetImageTag(),
+            layerFiles);
+    Path tempManifest = Files.createTempFile(null, null);
+    tempManifest.toFile().deleteOnExit();
+    manifestBlob.writeTo(new BufferedOutputStream(Files.newOutputStream(tempManifest)));
+    builder.addEntry(new TarArchiveEntry(tempManifest.toFile(), "manifest.json"));
+
+    // Load the image to docker daemon
+    File tarFile = File.createTempFile(buildConfiguration.getTargetImageRepository(), null);
```
coollog: This should use the `Files.createTempFile` method too (along with `deleteOnExit`)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
TadCordle: Looks like `toString` leaves out tag if it's "latest", but we need to specify the tag in the manifest.

TadCordle:
Thanks for the detailed feedback!

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,131 @@
...
+        DockerLoadManifestBlob.get(buildConfiguration.getTargetImageReference(), layerFiles);
+    Path tempManifest = Files.createTempFile(null, null);
+    tempManifest.toFile().deleteOnExit();
+    try (OutputStream bufferedOutputStream = Files.newOutputStream(tempManifest)) {
+      manifestBlob.writeTo(bufferedOutputStream);
+    }
+    tarStreamBuilder.addEntry(new TarArchiveEntry(tempManifest.toFile(), "manifest.json"));
+
+    // Load the image to docker daemon
+    try (ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream()) {
```
coollog: A `ByteArrayOutputStream` doesn't actually need to be wrapped in try-with-resource since closing it has no effect.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,131 @@
...
+      tarStreamBuilder.toBlob().writeTo(byteArrayOutputStream);
```
coollog: There is `Blobs#writeToByteArray` that can give the byte array directly.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,131 @@
...
+      new Command("docker", "load").run(byteArrayOutputStream.toByteArray());
```
coollog: Although there's already the comment above for refactoring into testable components, this should prob have an explicit comment to not use `Command` since that is an untested and not-very-robust test utility.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlob.java
```diff
@@ -0,0 +1,41 @@
...
+package com.google.cloud.tools.jib.docker;
+
+import com.google.cloud.tools.jib.blob.Blob;
+import com.google.cloud.tools.jib.image.ImageReference;
+import com.google.cloud.tools.jib.json.JsonTemplateMapper;
+import java.util.Collections;
+import java.util.List;
+
+/** Utility for creating image manifest for build to Docker daemon. */
+public class DockerLoadManifestBlob {
```
coollog: We generally put a private constructor for classes that are just static methods.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlob.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Builds a {@link DockerLoadManifestTemplate} from image parameters and returns the result as a
+   * blob.
+   */
+  public static Blob get(ImageReference imageReference, List<String> layerFiles) {
```
coollog: I feel like this could just be a method in `BuildTarballAndLoadDockerStep` and be refactored to whereever the image tar construction component will live.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerLoadManifestTemplate.java
```diff
@@ -0,0 +1,55 @@
+/*
```
coollog: The convention currently for JSON templates in this project is to put them under a `json` package.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerLoadManifestTemplate.java
```diff
@@ -0,0 +1,55 @@
...
+ *     source</a>
+ */
+class DockerLoadManifestTemplate implements JsonTemplate {
+
+  private final String config = "config.json";
+  private final List<String> repoTags = new ArrayList<>(1);
+  private final List<String> layers = new ArrayList<>();
+
+  public void setRepoTags(String repoTags) {
+    this.repoTags.add(repoTags);
```
coollog: If this is supposed to set the repo tag, this should probably be just setting the first element. A test for this would also help confirm the intended behavior.

jib-core/src/main/java/com/google/cloud/tools/jib/json/JsonTemplateMapper.java
```diff
@@ -69,10 +70,16 @@
...
-        });
+    return mapToBlob(template);
+  }
+
+  /** Convert a list of {@link JsonTemplate}s to a {@link Blob} of the JSON string. */
+  public static Blob toBlob(List<JsonTemplate> template) {
+    return mapToBlob(template);
+  }
+
+  private static Blob mapToBlob(Object template) {
```
coollog: This can actually be called `toBlob` as well since I believe Java resolves to the strictest type match. (The calls to it would need to be typecasted to Object though)

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlobTest.java
```diff
@@ -0,0 +1,68 @@
...
+  @Test
+  public void testGetDockerLoadManifest() throws URISyntaxException, IOException {
+    // Loads the expected JSON string.
+    Path jsonFile = Paths.get(Resources.getResource("json/loadmanifest.json").toURI());
+    String expectedJson = new String(Files.readAllBytes(jsonFile), StandardCharsets.UTF_8);
+
+    List<String> layers = Arrays.asList("layer1.tar.gz", "layer2.tar.gz", "layer3.tar.gz");
+    ImageReference testImage = ImageReference.of(null, "testrepo", "testtag");
+    Blob manifestBlob = DockerLoadManifestBlob.get(testImage, layers);
+    try (ByteArrayOutputStream jsonStream = new ByteArrayOutputStream()) {
```
coollog: Same here for the `ByteArrayOutputStream`

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlobTest.java
```diff
@@ -0,0 +1,68 @@
...
+      manifestBlob.writeTo(jsonStream);
```
coollog: `Blobs#writeToString` would give the string directly.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildDockerStepsIntegrationTest.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.cloud.tools.jib.image.ImageReference;
+import com.google.cloud.tools.jib.registry.LocalRegistry;
+import java.nio.file.Path;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/** Integration tests for {@link BuildImageSteps}. */
```
coollog: s/BuildImageSteps/BuildDockerSteps

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildDockerSteps.java
```diff
@@ -0,0 +1,160 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.common.util.concurrent.MoreExecutors;
+import java.io.IOException;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+
+/**
+ * All the steps to build to Docker daemon. TODO: Refactor for less duplicate code w/
```
coollog: Looks like the TODO got inlined

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,131 @@
...
+   */
+  @Override
+  public Void call() throws ExecutionException, InterruptedException {
+    List<ListenableFuture<?>> dependencies = new ArrayList<>();
+    dependencies.addAll(NonBlockingFutures.get(pullBaseImageLayerFuturesFuture));
+    dependencies.addAll(buildApplicationLayerFutures);
+    dependencies.add(NonBlockingFutures.get(buildConfigurationFutureFuture));
+    return Futures.whenAllComplete(dependencies)
+        .call(this::afterPushBaseImageLayerFuturesFuture, listeningExecutorService)
+        .get();
```
coollog: Let's try removing this `get` and see if it still works .

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Depends on {@code pushAuthorizationFuture}, {@code pushBaseImageLayerFuturesFuture.get()},
```
coollog: Should just be the `pushBaseImageLayerFuturesFuture.get` and `buildConfigurationFutureFuture.get`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,131 @@
...
+          new TarArchiveEntry(layerFile.toFile(), layerFile.getFileName().toString()));
+    }
+
+    // Add config to tarball
+    // TODO: Add ability to add blobs as entries to TarStreamBuilder without using temp files
+    Blob containerConfigurationBlob =
+        NonBlockingFutures.get(NonBlockingFutures.get(buildConfigurationFutureFuture));
+    Path tempConfig = Files.createTempFile(null, null);
+    tempConfig.toFile().deleteOnExit();
+    try (OutputStream bufferedOutputStream = Files.newOutputStream(tempConfig)) {
```
coollog: Looks like this is missing a `new BufferedOutputStream`. Although without it is fine, buffering the outputstream should make it be faster for writing to IO by reducing the total amount of expensive IO operations.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,131 @@
...
+      containerConfigurationBlob.writeTo(bufferedOutputStream);
+    }
+    tarStreamBuilder.addEntry(new TarArchiveEntry(tempConfig.toFile(), "config.json"));
+
+    // Add manifest to tarball
+    Blob manifestBlob =
+        DockerLoadManifestBlob.get(buildConfiguration.getTargetImageReference(), layerFiles);
+    Path tempManifest = Files.createTempFile(null, null);
+    tempManifest.toFile().deleteOnExit();
+    try (OutputStream bufferedOutputStream = Files.newOutputStream(tempManifest)) {
```
coollog: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlob.java
```diff
@@ -0,0 +1,41 @@
...
+    // Set up the JSON template.
+    DockerLoadManifestTemplate template = new DockerLoadManifestTemplate();
+    template.setRepoTags(imageReference.getRepository() + ":" + imageReference.getTag());
```
coollog: Oh, this should have registry as well, so if a person uses `gcr.io/my-project/my-app:latest`, it doesn't get rid of the `gcr.io` portion.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerLoadManifestTemplate.java
```diff
@@ -0,0 +1,55 @@
...
+ *     "eb05f3dbdb543cc610527248690575bacbbcebabe6ecf665b189cf18b541e3ca.tar.gz",
+ *     "ba7c544469e514f1a9a4dec59ab640540d50992b288adbb34a1a63c45bf19a24.tar.gz",
+ *     "15705ab016593987662839b40f5a22fd1032996c90808d4a1371eb46974017d5.tar.gz"
+ *   ]
+ * }]</pre>
+ *
+ * @see <a href="https://github.com/moby/moby/blob/master/image/tarexport/load.go">Docker load
+ *     source</a>
+ */
+class DockerLoadManifestTemplate implements JsonTemplate {
```
coollog: I was thinking and maybe a better way to do this would be to have another interface like `ListOfJsonTemplate` that has a method called like `getList` that would be implemented here by returning itself in a singleton list. The benefit of this is that there is no way to forget to wrap this template in a list as it is intended. The logic for the wrapping is also localized to this class rather than living in places like `DockerLoadManifestBlob`

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlobTest.java
TadCordle: It's kind of confusing to have a `writeTo` method in the Blob interface, but then have a bunch of other `writeTo` methods in the Blobs class, would it make sense to combine these somehow?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
TadCordle: Yeah my b, I looked this up and misinterpreted the results I found, I'll change it back.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlobTest.java
coollog: So the design for that is similar to classes like Files vs. Path, where the plural is a bunch of static methods that operate on the singular. `Blob` is an interface that is implemented with different ways it is written out to an `OutputStream`, whereas `Blobs` is a bunch of final static methods for working with `Blob`s. The methods are prob better named `toString` and `toByteArray`.

TadCordle: Files and Path are also pretty confusing.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,149 @@
...
+
+    return null;
+  }
+
+  /**
+   * Builds a {@link DockerLoadManifestTemplate} from image parameters and returns the result as a
+   * blob.
+   */
+  @VisibleForTesting
+  static Blob getManifestBlob(ImageReference imageReference, List<String> layerFiles) {
```
coollog: For this project, we're following the convention of static methods at the top of a class

The ordering is:

1. Constants
1. static fields
1. static methods
1. member fields
1. constructors
1. instance methods

jib-core/src/main/java/com/google/cloud/tools/jib/docker/json/DockerLoadManifestTemplate.java
```diff
@@ -39,14 +39,18 @@
  * @see <a href="https://github.com/moby/moby/blob/master/image/tarexport/load.go">Docker load
  *     source</a>
  */
-class DockerLoadManifestTemplate implements JsonTemplate {
+public class DockerLoadManifestTemplate implements JsonTemplate {
 
   private final String config = "config.json";
-  private final List<String> repoTags = new ArrayList<>(1);
+  private final List<String> repoTags = new ArrayList<>();
```
coollog: Initial capacity can still be 1 since otherwise it allocates 10 for initial capacity be default.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/json/DockerLoadManifestTemplate.java
```diff
@@ -39,14 +39,18 @@
...
   private final List<String> layers = new ArrayList<>();
 
   public void setRepoTags(String repoTags) {
-    this.repoTags.add(repoTags);
+    if (this.repoTags.isEmpty()) {
```
coollog: This is okay, though an alternative could be to just set it to an `ImmutableList.of(repoTags)` or have `repoTags` be initialized as a singleton list of `null` to avoid this logic here.

jib-core/src/main/java/com/google/cloud/tools/jib/json/ListOfJsonTemplate.java
```diff
@@ -0,0 +1,23 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.json;
+
+import java.util.List;
+
+public interface ListOfJsonTemplate extends JsonTemplate {
```
coollog: a short javadoc for what this is / how to implement

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -0,0 +1,148 @@
...
+    // Set up the JSON template.
+    DockerLoadManifestTemplate template = new DockerLoadManifestTemplate();
+    template.setRepoTags(
+        imageReference.toString() + (imageReference.usesDefaultTag() ? ":latest" : ""));
```
coollog: `toString` can be removed. Also, this logic could probably be moved to `ImageReference(#toStringWithTag?)` or have some comment for it since it is not immediately clear why we need to do this.

coollog: A comment for how repotags needs to have a tag in the template would help too.

jib-core/src/main/java/com/google/cloud/tools/jib/json/ListOfJsonTemplate.java
TadCordle: Eh, it's probably fine how it is. It's probably better not to worry about hypotheticals in the API until we actually encounter those cases.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerLoadManifestBlobTest.java
TadCordle: I meant more in the sense that you have functionality for one system split into two different classes, using two different styles. What if `Blob` was abstract and contained the methods currently in `Blobs`? I know abstract classes seem to be avoided in favor of interfaces but I'm not a fan of splitting things out like this.

coollog: Putting the methods in `Blobs` into an abstract `Blob` would not actually work since it introduces circular dependencies where a parent class (`Blob`) depends on implementation classes (like `InputStreamBlob`).

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -209,4 +209,9 @@ public String toString() {
 
     return referenceString.toString();
   }
+
+  /** @return the image reference in Docker-readable format, without hiding the tag. */
+  public String toStringWithTag() {
+    return this + (usesDefaultTag() ? ":latest" : "");
```
coollog: "latest" can be `DEFAULT_TAG`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
coollog: Above the private member fields too.

jib-core/src/main/java/com/google/cloud/tools/jib/json/ListOfJsonTemplate.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+
+/**
+ * All JSON templates to be used with {@link JsonTemplateMapper} that need to be wrapped in a list
+ * must extend this class.
+ *
+ * <p>Json fields should be private fields and fields that are {@code null} will not be serialized.
+ */
+public interface ListOfJsonTemplate extends JsonTemplate {
+  /** Returns the JsonTemplate wrapped as a list. e.g.: [{"property":"value"}] */
```
coollog: nit: newline at start of class block

coollog:
Oh, should link this to https://github.com/GoogleContainerTools/jib/issues/243 for reference.

Issue #262, opened 2018-05-09
Refactor to reduce duplicate code


TadCordle:
Refactor to reduce copy/pasted code between the following files:
- [x] `BuildDockerSteps` and `BuildImageSteps` (#309)
- [x] `BuildContainerConfigurationStep` and `BuildTarballLoadDockerStep`/`PushImageStep` (#269)
- [x] `BuildDockerStepsRunner` and `BuildImageStepsRunner` (#289)
- [x] `BuildDockerTask` and `BuildImageTask` (#296)
- [x] `BuildDockerMojo` and `BuildImageMojo` (#299)
- [x] `gradle.ProjectProperties` and `maven.ProjectProperties` (#310)

coollog:
Oh actually, hold off on that for a bit.

coollog:
Gonna do this first: #305

TadCordle:
Cool, if you want to take those you can, I'm working on the two ProjectProperties right now.

PR #265, opened 2018-05-10
Attach build to docker daemon functionality to gradle task

TadCordle:
Part of #243.

There's a lot of duplicate code here, so this introduces a few refactoring TODOs, which I grouped into issue #262.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -36,10 +36,18 @@ import org.gradle.api.tasks.TaskAction;
 /** Builds a container image and exports to the default Docker daemon. */
 public class BuildDockerTask extends DefaultTask {
 
-  /** Directory name for the cache. The directory will be relative to the build output directory. */
+  /**
+   * Directory name for the cache. The directory will be relative to the build output directory.
+   *
+   * <p>TODO: Move to ProjectProperties.
+   */
   private static final String CACHE_DIRECTORY_NAME = "jib-cache";
 
-  /** Converts an {@link ImageConfiguration} to an {@link Authorization}. */
+  /**
+   * Converts an {@link ImageConfiguration} to an {@link Authorization}.
+   *
+   * <p>TODO: Move to ImageConfiguration.
+   */
   @Nullable
   private static Authorization getImageAuthorization(ImageConfiguration imageConfiguration) {
     if (imageConfiguration.getAuth().getUsername() == null
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.jib.gradle;
 
-import com.google.api.client.http.HttpTransport;
 import com.google.cloud.tools.jib.builder.BuildConfiguration;
 import com.google.cloud.tools.jib.frontend.BuildDockerStepsRunner;
 import com.google.cloud.tools.jib.frontend.BuildImageStepsExecutionException;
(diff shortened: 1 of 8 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -16,18 +16,130 @@
...
 import org.gradle.api.tasks.TaskAction;
 
 /** Builds a container image and exports to the default Docker daemon. */
 public class BuildDockerTask extends DefaultTask {
 
+  /** Directory name for the cache. The directory will be relative to the build output directory. */
+  private static final String CACHE_DIRECTORY_NAME = "jib-cache";
+
+  /** {@code User-Agent} header suffix to send to the registry. */
+  private static final String USER_AGENT_SUFFIX = "jib-gradle-plugin";
```
coollog: This isn't necessary anymore since it's not using a registry.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -16,18 +16,130 @@
...
 import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
+import org.gradle.api.tasks.Nested;
 import org.gradle.api.tasks.TaskAction;
 
 /** Builds a container image and exports to the default Docker daemon. */
 public class BuildDockerTask extends DefaultTask {
 
+  /** Directory name for the cache. The directory will be relative to the build output directory. */
+  private static final String CACHE_DIRECTORY_NAME = "jib-cache";
```
coollog: We'd probably want this to be shared and the same with `BuildImageTask`. Maybe `ProjectProperties` would be a good place for it, and the tasks would call like `ProjectProperties#getCacheDirectory`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -16,18 +16,130 @@
...
+
+  /** {@code User-Agent} header suffix to send to the registry. */
+  private static final String USER_AGENT_SUFFIX = "jib-gradle-plugin";
+
+  /** Converts an {@link ImageConfiguration} to an {@link Authorization}. */
+  @Nullable
+  private static Authorization getImageAuthorization(ImageConfiguration imageConfiguration) {
```
coollog: We could prob just put this in `ImageConfiguration`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -16,18 +16,130 @@
...
+        BuildConfiguration.builder(new GradleBuildLogger(getLogger()))
+            .setBaseImage(baseImageReference)
+            .setBaseImageCredentialHelperName(jibExtension.getFrom().getCredHelper())
+            .setKnownBaseRegistryCredentials(knownBaseRegistryCredentials)
+            .setTargetImage(targetImageReference)
+            .setMainClass(mainClass)
+            .setJvmFlags(jibExtension.getJvmFlags())
+            .build();
+
+    // Disables annoying Apache HTTP client logging.
```
coollog: These HTTP stuff can be removed.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
TadCordle: Do you want me to do this now or hold off on this until last bullet of #262?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
TadCordle: Same question about #262

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
coollog: Can hold off but maybe add a TODO here or at the top of the class

PR #269, opened 2018-05-11
Refactor configuration build/push/tar build steps to reduce duplicate code

### LINKED ISSUE #263: Add ability to add Blobs as entries to TarStreamBuilder without using temp files

TadCordle:
Part of #262 (refactors `BuildContainerConfigurationStep`, `BuildTarballLoadDockerStep`, and `PushImageStep`).

Fixes #263 (adds ability to add blobs to `TarStreamBuilder` without temp files).


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildContainerConfigurationStep.java
```diff
@@ -32,7 +30,7 @@
 import java.util.concurrent.Future;
 
 /** Builds the container configuration. */
-class BuildContainerConfigurationStep implements Callable<ListenableFuture<Blob>> {
+class BuildContainerConfigurationStep implements Callable<ListenableFuture<Image>> {
```
coollog: `BuildImageStep` now?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildImageSteps.java
```diff
@@ -183,17 +182,18 @@ public void run()
           // Pushes the new image manifest.
           ListenableFuture<Void> pushImageFuture =
               Futures.whenAllSucceed(
-                      pushBaseImageLayerFuturesFuture, pushContainerConfigurationFutureFuture)
+                      pushBaseImageLayerFuturesFuture,
+                      pushContainerConfigurationFutureFuture,
+                      buildContainerConfigurationFutureFuture)
```
coollog: The dependency on `buildContainerConfigurationFutureFuture` shouldn't be necessary since `pushContainerConfigurationFutureFuture` already depends on it.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -61,6 +68,17 @@ public void addEntry(TarArchiveEntry entry) {
     entries.add(entry);
   }
 
+  /** Adds a blob to the archive. */
+  public void addEntry(Blob blob, String name) throws IOException {
+    // TODO: Efficiency? We're writing blob contents twice (once here to get entry size, once when
```
coollog: Yeah, I think you could do it by creating a class that encompasses either a file or a blob and a name, and it has a method to write it to a TarArchiveOutputStream with the right TarArchiveEntry and contents.

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarStreamBuilderTest.java
```diff
@@ -66,37 +108,57 @@ public void setUp() throws IOException, URISyntaxException {
...
+    testTarStreamBuilder.addEntry(fileAContents, "some/path/to/resourceFileA");
+    testTarStreamBuilder.addEntry(new TarArchiveEntry(fileB.toFile(), "crepecake"));
+    testTarStreamBuilder.addEntry(new TarArchiveEntry(directoryA.toFile(), "some/path/to"));
+    testTarStreamBuilder.addEntry(
+        fileAContents,
+        "some/really/long/path/that/exceeds/100/characters/abcdefghijklmnopqrstuvwxyz0123456789012345678901234567890");
+  }
+
+  /** Creates a compressed blob from the TarStreamBuilder and verifies it. */
+  private void verifyBlobWithCompression() throws IOException {
```
coollog: This is fine since it's just a test helper, but since theres only two line difference between these functions, it could just be one function with a parameter "compress". Then, like line 140 could be `blob.writeTo(compress?new GZIPOutputStream(tarByteOutputStream):tarByteOutputStream)`.

PR #272, opened 2018-05-14
Fixes Maven integration tests.

coollog:
With workaround from https://github.com/spring-projects/spring-boot/issues/6254#issuecomment-307151464


jib-maven-plugin/pom.xml
```diff
@@ -311,6 +335,7 @@
         <configuration>
           <excludes>
             <exclude>**/*IntegrationTest.java</exclude>
+            <exclude>**/*IT.java</exclude>
```
elharo: This pattern violates Google coding conventions (camel case and no abbreviations) and shouldn't be occurring in our repos.

PR #275, opened 2018-05-15
Remove Command class usage outside of tests

### LINKED ISSUE #261: Refactor BuildTarballAndLoadDockerStep into testable components
- [x] ImageToTarballTranslator (#338)
- [x] DockerLoadClient (#317)

TadCordle:
Command class is intended as a test utility, so I removed it from `BuildTarballAndLoadDockerStep`.

This code will probably be moved around a bit more in the future to address #261.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildTarballAndLoadDockerStep.java
```diff
@@ -105,8 +108,18 @@ private Void afterPushBaseImageLayerFuturesFuture()
...
+    ProcessBuilder processBuilder = new ProcessBuilder("docker", "load");
+    Process process = processBuilder.start();
+    try (OutputStream stdin = process.getOutputStream()) {
+      tarStreamBuilder.toBlob().writeTo(stdin);
+    }
+    try (InputStreamReader stdout =
+        new InputStreamReader(process.getInputStream(), StandardCharsets.UTF_8)) {
+      String output = CharStreams.toString(stdout);
+      if (process.waitFor() != 0) {
+        throw new RuntimeException("'docker load' command failed with output: " + output);
```
coollog: Let's use a checked exception here.

TadCordle: Which one?

PR #278, opened 2018-05-16
Infer main class if it isn't configured anywhere

### LINKED ISSUE #189: Infer main class by searching for class with main method.

TadCordle:
Searches the build output for .class files with a `main` method. If more than one is found, or if one isn't found, it throws an error.

Fixes #189.


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+          pathStream
+              .filter(Files::isRegularFile)
+              .filter(path -> path.toString().endsWith(".class"))
+              .collect(Collectors.toList());
+
+      for (Path classFile : classFiles) {
+        // Convert filename (rootDir/path/to/ClassName.class) to class name (path.to.ClassName)
+        String name = classFile.toAbsolutePath().toString();
+        name = name.substring(rootDirectory.length() + 1);
+        name = name.replace('/', '.').replace('\\', '.');
```
coollog: `\` is actually a valid character for a filename in Unix. A safer way to do this conversion would be to just get the parts of the `Path` by iterating over it.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+        name = name.substring(0, name.length() - ".class".length());
+
+        Class fileClass = new ClassFileLoader(classFile).findClass(name);
```
coollog: Generics should be used with a template - `Class<?>`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+        if (fileClass == null) {
+          // Got an invalid class file, keep searching
+          continue;
+        }
+
+        try {
+          // Check if class contains a public static void main(String[] args)
```
coollog: Wrap in `{@code }`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+          if (main != null
+              && main.getReturnType() == void.class
+              && Modifier.isStatic(main.getModifiers())
+              && Modifier.isPublic(main.getModifiers())) {
+            if (className == null) {
+              // Main method found; save it and continue searching for duplicates
+              className = name;
+            } else {
+              // Found more than one main method
+              throw new MultipleClassesFoundException(className, name);
```
coollog: An alternative could be to change this method to return a list of main classes found so that

1. No need to have this exception
2. Doesn't need to be `@Nullable`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MultipleClassesFoundException.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.frontend;
+
+public class MultipleClassesFoundException extends Exception {
```
coollog: short javadoc

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+            }
+          }
+        } catch (NoSuchMethodException ignored) {
+          // main method not found
+        }
+      }
+    }
+
+    return className;
+  }
```
coollog: Should have a private constructor

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+
+  /**
+   * Searches for a class containing a main method given an absolute root directory.
+   *
+   * @return the name of the class if one is found, null if no class is found.
+   * @throws IOException if searching/reading files fails.
+   * @throws MultipleClassesFoundException if more than one valid main class is found.
+   */
+  @Nullable
+  public static String findMainClass(String rootDirectory)
```
coollog: Should prob pass `rootDirectory` as `Path`

jib-gradle-plugin/src/integration-test/resources/projects/empty/build.gradle
```diff
@@ -15,7 +15,6 @@ jib {
         image = 'gcr.io/jib-integration-testing/emptyimage:gradle'
         credHelper = 'gcr'
     }
-    mainClass = 'com.test.Empty'
```
coollog: :smiley:

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -46,21 +48,38 @@ String getMainClass(@Nullable String mainClass) {
     if (mainClass == null) {
       mainClass = getMainClassFromJarTask();
       if (mainClass == null) {
-        throw new GradleException(
-            HelpfulSuggestionsProvider.get("Could not find main class specified in a 'jar' task")
-                .suggest("add a `mainClass` configuration to jib"));
+        gradleBuildLogger.info(
```
coollog: This could prob just be at the debug level.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -46,21 +48,38 @@ String getMainClass(@Nullable String mainClass) {
...
+            "Could not find main class specified in a 'jar' task; attempting to "
+                + "infer main class.");
+        try {
+          mainClass =
+              MainClassFinder.findMainClass(
+                  project
```
coollog: This should use `GradleSourceConfiguration#getClassesFiles`, which will be the files that go into the container.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -46,21 +48,38 @@ String getMainClass(@Nullable String mainClass) {
...
+                      .toPath()
+                      .resolve("classes")
+                      .resolve("java")
+                      .resolve("main")
+                      .toAbsolutePath()
+                      .toString());
+        } catch (MultipleClassesFoundException | IOException ex) {
+          throw new GradleException(
+              HelpfulSuggestionsProvider.get("Failed to get main class: " + ex.getMessage())
+                  .suggest("add a `mainClass` configuration to jib"));
```
coollog: The `ex` should go as a second argument to `GradleException` rather than a string addition so there is stack hierarchy.

coollog: This suggestion can be a constant together with below.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+      return null;
+    }
+
+    String className = null;
+    try (Stream<Path> pathStream = Files.walk(Paths.get(rootDirectory))) {
+      // Get all .class files
+      List<Path> classFiles =
+          pathStream
+              .filter(Files::isRegularFile)
+              .filter(path -> path.toString().endsWith(".class"))
```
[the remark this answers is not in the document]
TadCordle: http://saltnlight5.blogspot.com/2014/04/be-careful-with-java-pathendswithstring.html

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -0,0 +1,115 @@
...
+  @Nullable
+  public static String findMainClass(String rootDirectory)
+      throws MultipleClassesFoundException, IOException {
+    // Make sure rootDirectory is valid
+    if (!Files.exists(Paths.get(rootDirectory)) || !Files.isDirectory(Paths.get(rootDirectory))) {
+      return null;
+    }
+
+    String className = null;
+    try (Stream<Path> pathStream = Files.walk(Paths.get(rootDirectory))) {
```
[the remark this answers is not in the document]
TadCordle: Is this much of an improvement? Looks like implementing a DirectoryWalker would be more complicated.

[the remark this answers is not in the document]
coollog: It's mostly as a way to abstract away the strategy for walking into a testable class. The use would like something like:

```
new DirectoryWalker(rootDirectory).filter(Files::isRegularFile).filter(path -> path.toString().endsWith(".class")).walk(classFile -> {...});
```

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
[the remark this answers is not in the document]
coollog: Ah right that makes sense. Looks like we used `Filesystems#getPathMatcher` in `MavenSourceFilesConfiguration`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
TadCordle: This returns a list of Paths, how would I use it to get the root directory?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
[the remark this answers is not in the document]
TadCordle: Oh I see, the one you implemented. I was thinking you meant something like this https://commons.apache.org/proper/commons-io/javadocs/api-1.4/org/apache/commons/io/DirectoryWalker.html

coollog:
It would need to do the findMainClass for all the paths since there could
be multiple sourcesets.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -53,63 +53,54 @@ public Class findClass(String name) {
...
+              Class<?> fileClass = new ClassFileLoader(classFile).findClass(name);
+              if (fileClass != null) {
+                try {
+                  // Check if class contains {@code public static void main(String[] args)}
+                  Method main = fileClass.getMethod("main", String[].class);
+                  if (main != null
+                      && main.getReturnType() == void.class
+                      && Modifier.isStatic(main.getModifiers())
+                      && Modifier.isPublic(main.getModifiers())) {
+                    results.add(name);
```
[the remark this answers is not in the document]
TadCordle: Doesn't look like it, I think I need to pass the full name into the class loader.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -48,29 +51,36 @@ String getMainClass(@Nullable String mainClass) {
...
-                      .toAbsolutePath()
-                      .toString());
-        } catch (MultipleClassesFoundException | IOException ex) {
-          throw new GradleException(
-              HelpfulSuggestionsProvider.get("Failed to get main class: " + ex.getMessage())
-                  .suggest("add a `mainClass` configuration to jib"));
-        }
-        if (mainClass == null) {
+          // Adds each file in each classes output directory to the classes files list.
+          JavaPluginConvention javaPluginConvention =
```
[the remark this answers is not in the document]
TadCordle: It was either this or something like `getSourceFilesConfiguration().getClassesFiles().get(0).getParent()` since I need the root directory, and `getClassesFiles` seems to return a list of the files in the root directory. Plus using `getSourceFilesConfiguration()` was weird in testing, since I would either need to refactor a bit, or mock a bunch of the `GradleSourceFilesConfiguration` internals to avoid null pointer exceptions.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -72,34 +71,25 @@ public Class findClass(String name) {
...
-                } catch (NoSuchMethodException ignored) {
-                  // main method not found
+              try {
+                // Check if class contains {@code public static void main(String[] args)}
+                Method main = fileClass.getMethod("main", String[].class);
+                if (main != null
+                    && main.getReturnType() == void.class
+                    && Modifier.isStatic(main.getModifiers())
+                    && Modifier.isPublic(main.getModifiers())) {
+                  classNames.add(fileClass.getCanonicalName());
```
coollog: I believe you would want `getName` since the canonical name would replace a subclass like `package.Hello$Main` with `package.Hello.Main` 

https://stackoverflow.com/questions/15202997/what-is-the-difference-between-canonical-name-simple-name-and-class-name-in-jav

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -19,33 +19,57 @@
...
+  private static final String PLUGIN_NAME = "jib";
+
   private final Project project;
   private final GradleBuildLogger gradleBuildLogger;
+  private final SourceFilesConfiguration sourceFilesConfiguration;
 
   ProjectProperties(Project project, GradleBuildLogger gradleBuildLogger) {
     this.project = project;
     this.gradleBuildLogger = gradleBuildLogger;
+    try {
```
coollog: Looks good!

This should prob go in a static method like `ProjectProperties#getForProject` since constructors generally should be lightweight.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -54,32 +78,29 @@ String getMainClass(@Nullable String mainClass) {
         gradleBuildLogger.debug(
```
coollog: I was talking to @patflynn and Patrick mentioned that it might be better to log something like `Searching for main class... Add a 'mainClass' configuration to 'jib' to improve build speed.` at the `info` level to help users in case they have a ton of classes to search.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProjectProperties.java
```diff
@@ -64,26 +77,25 @@ String getMainClass(@Nullable String mainClass) throws MojoExecutionException {
             "Could not find main class specified in maven-jar-plugin; attempting to infer main "
                 + "class.");
 
-        final String mainClassSuggestion = "add a `mainClass` configuration to jib-maven-plugin";
         try {
           List<String> mainClasses =
-              MainClassFinder.findMainClass(Paths.get(project.getBuild().getOutputDirectory()));
+              MainClassFinder.findMainClasses(Paths.get(project.getBuild().getOutputDirectory()));
```
coollog: This should use `MavenSourceFilesConfiguration` like you did in Gradle.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
TadCordle: K, wasn't sure which one we wanted.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ProjectProperties.java
```diff
@@ -30,45 +33,91 @@
...
+          "Searching for main class... Add a 'mainClass' configuration to '"
+              + PLUGIN_NAME
+              + "' to improve build speed.");
       mainClass = getMainClassFromJarTask();
       if (mainClass == null) {
-        throw new GradleException(
-            HelpfulSuggestionsProvider.get("Could not find main class specified in a 'jar' task")
-                .suggest("add a `mainClass` configuration to jib"));
+        gradleBuildLogger.debug(
+            "Could not find main class specified in a 'jar' task; attempting to "
```
coollog: Let's add a todo to consolidate this code with the Maven one.

TadCordle: Already added it to #262, but I can add a TODO in the code as well if you want.

coollog: Okay that's fine then.

PR #282, opened 2018-05-17
Throw error before build to docker daemon if docker isn't installed

### LINKED ISSUE #268: Make sure Docker is installed before user tries to build to Docker daemon

TadCordle:
Fixes #268


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/BuildDockerStepsRunner.java
```diff
@@ -190,4 +193,20 @@ private void handleRegistryUnauthorizedException(
...
     }
   }
+
+  /** @return true if Docker is installed on the user's system. */
+  public static boolean isDockerInstalled() {
+    try {
+      ProcessBuilder processBuilder = new ProcessBuilder("docker");
+      processBuilder.redirectErrorStream(true);
+      Process process = processBuilder.start();
+      try (InputStreamReader stdout =
```
coollog: Is this stdout reading necessary? Could it just be start the process, if exception then return false, otherwise true?

TadCordle: Looks like that works.

PR #288, opened 2018-05-18
Makes sure all classes passed between async steps are immutable.

### LINKED ISSUE #276: Make sure all classes passed between async steps are immutable.
Looks like there was a `ConcurrentModificationException`: https://fusion.corp.google.com/projectanalysis/current/KOKORO/prod%3Ajib%2Fgcp_ubuntu%2Fcontinuous

coollog:
Fixes #276


jib-core/src/main/java/com/google/cloud/tools/jib/builder/PullAndCacheBaseImageLayerStep.java
```diff
@@ -46,7 +47,7 @@
       BuildConfiguration buildConfiguration,
       Cache cache,
       DescriptorDigest layerDigest,
-      Future<Authorization> pullAuthorizationFuture) {
```
TadCordle: Should probably change the private final field as well from `Future` to `ListenableFuture`

coollog: Good catch - found a bunch of other places too

Issue #291, opened 2018-05-18
Refactor async steps to have dependent steps be explicit dependencies.


coollog:
So, instead of having the steps take `ListenableFuture`s as constructor parameters, they will directly take other step classes.

This relates to #276 and helps with the refactoring that will be done as part of #262.

- [x] Create AsyncStep interface #292
- [x] RetrieveRegistryCredentialsStep #292 
- [x] AuthenticatePushStep #293
- [x] AuthenticatePullStep #294
- [x] PullBaseImageStep #295 
- [x] BuildAndCacheApplicationLayersStep #297
- [x] BuildImageStep #297
- [x] BuildTarballAndLoadDockerStep #297 
- [x] PullAndCacheBaseImageLayerStep #297 
- [x] PullAndCacheBaseImageLayersStep #297  
- [x] PushBlobStep #297 
- [x] PushImageStep #297 
- [x] PushLayersStep #297 
- [x] PushContainerConfigurationStep #297 

- [ ] Add orchestrator that attaches the steps to each other correctly rather than the steps doing it themselves?

PR #292, opened 2018-05-18
Creates AsyncStep interface.

coollog:
Gets started with #291


coollog:
Waiting on #288 to merge master.

loosebazooka:
I don't think I understand this change. It's not clear to me that we should transfer responsibility of submitting to the step.

PR #309, opened 2018-05-22
Consolidates BuildImageSteps and BuildDockerSteps into BuildSteps.

coollog:
Finishes part 1 of #262


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
```diff
@@ -16,23 +16,181 @@
...
-      throws InterruptedException, ExecutionException, CacheMetadataCorruptedException, IOException,
-          CacheDirectoryNotOwnedException;
+  /** Accepts {@link StepsRunner} by running the appropriate steps. */
+  @FunctionalInterface
+  private interface StepsRunnerConsumer {
+
+    void accept(StepsRunner stepsRunner) throws ExecutionException, InterruptedException;
+  }
+
+  private static final String DESCRIPTION_FOR_DOCKER_REGISTRY = "Building and pushing image";
```
TadCordle: Should this naming be less specific (can we push to non-docker registries)?

coollog: Currently, we only can push to Docker registries. I'm not actually sure if there are any other types of container registries out there.

PR #310, opened 2018-05-23
Reduce duplicate code in maven/gradle ProjectProperties

### LINKED ISSUE #262: Refactor to reduce duplicate code

TadCordle:
~~Part of~~ Fixes #262.


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -51,12 +55,78 @@ public Class findClass(@Nullable String name) {
...
+   * If {@code mainClass} is {@code null}, tries to infer main class in this order:
+   *
+   * <ul>
+   *   <li>1. Looks in a {@code jar} plugin provided by {@code projectProperties}.
+   *   <li>2. Searches for a class defined with a main method.
+   * </ul>
+   *
+   * <p>Warns if main class is not valid, or throws an error if no valid main class is not found.
+   */
+  public static String resolveMainClass(
```
coollog: I feel like it might make sense to separate these into two classes - one for searching for a main method class, and one for the actual logic of inferring a main class - these seem like two distinct things on different levels of abstraction.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ProjectProperties.java
```diff
@@ -0,0 +1,41 @@
...
+
+  String getPluginName();
+
+  SourceFilesConfiguration getSourceFilesConfiguration();
+
+  /** @return the name of the main class configured in a jar plugin, or null if none is found. */
+  @Nullable
+  String getMainClassFromJar();
+
+  /** @return a {@link HelpfulSuggestions} instance for main class inference failure. */
```
coollog: Is this just for main class inference failure? If so, we can name this more specifically.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerContextTask.java
```diff
@@ -64,18 +65,18 @@ public void generateDockerContext() {
...
     String targetDir = getTargetDir();
 
     try {
-      new DockerContextGenerator(projectProperties.getSourceFilesConfiguration())
+      new DockerContextGenerator(gradleProjectProperties.getSourceFilesConfiguration())
           .setBaseImage(jibExtension.getBaseImage())
           .setJvmFlags(jibExtension.getJvmFlags())
-          .setMainClass(mainClass)
+          .setMainClass(
+              MainClassFinder.resolveMainClass(
```
coollog: I think we should be explicit about the ordering of log/error messages that show up. We should probably keep all the method calls that could error/log things together near the top before any logic that works on the return values of those methods. This way, we can easily see what order things could be checked in.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -0,0 +1,95 @@
...
+  private final SourceFilesConfiguration sourceFilesConfiguration;
+
+  /** @return a GradleProjectProperties from the given project and logger. */
+  static GradleProjectProperties getForProject(
+      Project project, GradleBuildLogger gradleBuildLogger) {
+    try {
+      return new GradleProjectProperties(
+          project, gradleBuildLogger, GradleSourceFilesConfiguration.getForProject(project));
+    } catch (IOException ex) {
+      throw new GradleException("Obtaining project build output files failed", ex);
```
coollog: We might want to give the same suggestion about compiling as in Maven.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -0,0 +1,95 @@
...
+class GradleProjectProperties implements ProjectProperties {
+
+  private static final String PLUGIN_NAME = "jib";
+
+  private final Project project;
+  private final GradleBuildLogger gradleBuildLogger;
+  private final SourceFilesConfiguration sourceFilesConfiguration;
+
+  /** @return a GradleProjectProperties from the given project and logger. */
+  static GradleProjectProperties getForProject(
```
coollog: static method above instance fields

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -0,0 +1,95 @@
...
+  @Override
+  public String getMainClassFromJar() {
+    List<Task> jarTasks = new ArrayList<>(project.getTasksByName("jar", false));
+    if (jarTasks.size() != 1) {
+      return null;
+    }
+    return (String) ((Jar) jarTasks.get(0)).getManifest().getAttributes().get("Main-Class");
+  }
+
+  public Path getCacheDirectory() {
```
coollog: Looks like this can also be in the interface?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -0,0 +1,95 @@
...
+    return project.getBuildDir().toPath().resolve(CACHE_DIRECTORY_NAME);
+  }
+
+  private GradleProjectProperties(
```
coollog: constructor above methods

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
TadCordle: I'm not sure. `findMainClasses` is only ever called in `resolveMainClass` (and also in the tests, of course), so I see it more as an implementation detail rather than something that belongs on a separate level of abstraction. The functionality is different enough that yeah, maybe it makes sense to break it out by itself, but I'd rather avoid spreading it out when it's only used in that one place. We can split it out later if we need to (maybe when jib-core becomes a standalone library).

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -51,13 +55,80 @@ public Class findClass(@Nullable String name) {
...
+    BuildLogger logger = projectProperties.getLogger();
+    if (mainClass == null) {
+      logger.info(
+          "Searching for main class... Add a 'mainClass' configuration to '"
+              + projectProperties.getPluginName()
+              + "' to improve build speed.");
+      mainClass = projectProperties.getMainClassFromJar();
+      if (mainClass == null || !BuildConfiguration.isValidJavaClass(mainClass)) {
+        logger.debug(
+            "Could not find a valid main class specified in a jar plugin; attempting to infer "
```
coollog: This message should be specific for `maven-jar-plugin` or `jar` task too.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -51,13 +55,80 @@ public Class findClass(@Nullable String name) {
...
+   *
+   * <p>Warns if main class is not valid, or throws an error if no valid main class is not found.
+   */
+  public static String resolveMainClass(
+      @Nullable String mainClass, ProjectProperties projectProperties) {
+    BuildLogger logger = projectProperties.getLogger();
+    if (mainClass == null) {
+      logger.info(
+          "Searching for main class... Add a 'mainClass' configuration to '"
+              + projectProperties.getPluginName()
```
coollog: Oh, `HelpfulSuggestions` was actually intended to be used for things like this. The `HelpfulSuggestionsProvider` would construct with different values for maven vs gradle, and then a method would construct the message to be used in the core code like here. Same goes for the message below about the jar plugin vs task.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -51,13 +55,80 @@ public Class findClass(@Nullable String name) {
...
+          for (Path classPath : projectProperties.getSourceFilesConfiguration().getClassesFiles()) {
+            mainClasses.addAll(findMainClasses(classPath));
+          }
+
+          if (mainClasses.size() == 1) {
+            // Valid class found; use inferred main class
+            mainClass = mainClasses.get(0);
+          } else if (mainClasses.size() == 0 && mainClass == null) {
+            // No main class found anywhere
+            throw new IllegalStateException(
```
coollog: Oh I just noticed this should prob be wrapped in the `GradleException`/`MojoExecutionException` (wrap-unwrap-rewrap a thrown checked exception)

TadCordle:
This PR ended up getting bigger than I was hoping it would.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -48,7 +48,7 @@ public Class findClass(@Nullable String name) {
       try {
         byte[] bytes = Files.readAllBytes(classFile);
         return defineClass(name, bytes, 0, bytes.length);
-      } catch (IOException | ClassFormatError ignored) {
+      } catch (IOException | ClassFormatError | NoClassDefFoundError ignored) {
```
coollog: Oh we should move this fix to another PR.

PR #314, opened 2018-05-24
Fix NoClassDefFoundError during main class inference

### LINKED ISSUE #312: Main class inference fails in petclinic app
```
Caused by: java.lang.NoClassDefFoundError: org/springframework/samples/petclinic/model/BaseEntity (wrong name: org/springframework/samples/petclinic/model/NamedEntity)
    at java.lang.ClassLoader.defineClass1 (Native Method)
    at java.lang.ClassLoader.defineClass (ClassLoader.java:763)
    at java.lang.ClassLoader.defineClass (ClassLoader.java:642)
    at com.google.cloud.tools.jib.frontend.MainClassFinder$ClassFileLoader.findClass (MainClassFinder.java:50)
    at java.lang.ClassLoader.loadClass (ClassLoader.java:424)
    at java.lang.ClassLoader.loadClass (ClassLoader.java:357)
    at java.lang.ClassLoader.defineClass1 (Native Method)
    at java.lang.ClassLoader.defineClass (ClassLoader.java:763)
    at java.lang.ClassLoader.defineClass (ClassLoader.java:642)
    at com.google.cloud.tools.jib.frontend.MainClassFinder$ClassFileLoader.findClass (MainClassFinder.java:50)
    at com.google.cloud.tools.jib.frontend.MainClassFinder.lambda$findMainClasses$2 (MainClassFinder.java:145)
    at com.google.cloud.tools.jib.filesystem.DirectoryWalker.walk (DirectoryWalker.java:62)
    at com.google.cloud.tools.jib.frontend.MainClassFinder.findMainClasses (MainClassFinder.java:143)
    at com.google.cloud.tools.jib.frontend.MainClassFinder.resolveMainClass (MainClassFinder.java:87)
    at com.google.cloud.tools.jib.maven.BuildImageMojo.execute (BuildImageMojo.java:93)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:134)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:51)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:309)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:194)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:107)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:993)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:345)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:191)
    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:498)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:289)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:229)
    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:415)
    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:356)
```

TadCordle:
Fixes #312


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -36,20 +37,49 @@
...
     @Nullable
     @Override
     public Class findClass(@Nullable String name) {
       try {
+        // Name is only ever null when we call findClass manually, and not null otherwise. If null,
+        // we should resolve the correct filename.
+        if (name != null) {
+          classFile = rootDirectory;
+          List<String> folders = Splitter.on(".").splitToList(name);
+          for (int index = 0; index < folders.size() - 1; index++) {
```
coollog: Should use iterator style rather than indexing. (like first remove the last element from folders)

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -36,20 +37,49 @@
...
+            classFile = classFile.resolve(folders.get(index));
+          }
+
+          String className = folders.get(folders.size() - 1);
```
coollog: this can come from removing the last element

coollog: Generally, should prefer using StringBuilder when constructing strings.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -36,20 +37,49 @@
...
     }
 
     @Nullable
     @Override
     public Class findClass(@Nullable String name) {
       try {
+        // Name is only ever null when we call findClass manually, and not null otherwise. If null,
+        // we should resolve the correct filename.
+        if (name != null) {
+          classFile = rootDirectory;
```
coollog: Wouldn't this break the later calls to `findClass(null)`?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -36,20 +37,49 @@
...
+          List<String> folders = Splitter.on(".").splitToList(name);
+          for (int index = 0; index < folders.size() - 1; index++) {
+            classFile = classFile.resolve(folders.get(index));
+          }
+
+          String className = folders.get(folders.size() - 1);
+          if (className.contains("$")) {
```
coollog: Wouldn't inner classes be their own `.class` files (with $ in the name)?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -157,8 +184,9 @@ public static String resolveMainClass(
                     && Modifier.isPublic(main.getModifiers())) {
                   classNames.add(fileClass.getName());
                 }
-              } catch (NoSuchMethodException ignored) {
+              } catch (NoSuchMethodException | NoClassDefFoundError ignored) {
```
coollog: Shouldn't this not be happening anymore if it's taken care of in `findClass`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -115,7 +115,10 @@ public String getJarPluginName() {
    */
   String getMainClass(JibExtension jibExtension) {
     try {
-      return MainClassFinder.resolveMainClass(jibExtension.getMainClass(), this);
+      return MainClassFinder.resolveMainClass(
+          jibExtension.getMainClass(),
+          this,
+          getSourceFilesConfiguration().getClassesFiles().get(0).getParent());
```
coollog: should be for all the `getClassesFiles`, ditto for Maven

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
TadCordle: Not at the moment, since we only ever call it ourselves after constructing a new `ClassFileLoader`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
TadCordle: I mostly just added it for safety in case there are edge cases we haven't considered.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
TadCordle: `splitToList` returns an unmodifiable list.

TadCordle: Nevermind, works if I wrap it in an ArrayList

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
coollog: Yeah, the members should be able to be `private final`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -53,16 +53,11 @@ public Class findClass(@Nullable String name) {
         // we should resolve the correct filename.
         if (name != null) {
           classFile = rootDirectory;
-          List<String> folders = Splitter.on(".").splitToList(name);
-          for (int index = 0; index < folders.size() - 1; index++) {
-            classFile = classFile.resolve(folders.get(index));
+          ArrayList<String> folders = new ArrayList<>(Splitter.on('.').splitToList(name));
```
coollog: Using something like `LinkedList` actually provides a `removeLast` method.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
coollog: Should add a test with inner classes to confirm this actually works.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -37,8 +39,8 @@
   /** Helper for loading a .class file. */
   private static class ClassFileLoader extends ClassLoader {
 
-    private Path rootDirectory;
-    private Path classFile;
+    private final Path rootDirectory;
+    private final Path classFile;
```
coollog: maybe name like `defaultClassFile`?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -51,21 +53,22 @@ public Class findClass(@Nullable String name) {
       try {
         // Name is only ever null when we call findClass manually, and not null otherwise. If null,
         // we should resolve the correct filename.
+        Path file = classFile;
         if (name != null) {
```
coollog: I feel like this the null case should be in a separate method to avoid this null-check logic.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -109,7 +112,16 @@ public static String resolveMainClass(
 
         try {
           // Adds each file in the classes output directory to the classes files list.
-          List<String> mainClasses = findMainClasses(rootDirectory);
+          HashSet<Path> visitedRoots = new HashSet<>();
```
coollog: Generally, in this codebase, we try to use the highest level type possible to limit exposed APIs as much as possible.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -41,6 +43,8 @@
   /** Helper for loading a .class file. */
   private static class ClassFileLoader extends ClassLoader {
 
+    private static final Map<Path, Class> cache = new HashMap<>();
```
coollog: `Class<?>` instead of `Class` (same for `findClass` return type)

coollog: We should be careful with global states like this since having the cache be a static variable means that the `ClassFileLoader` is now not thread-safe. This would mean that we would have to make sure we don't use this `MainClassFinder` in multiple threads concurrently, or we change this cache to be an instance field, or use a thread-safe map.

TadCordle: Can't change the `findClass` return type since it's an override, but I'll change this one.

TadCordle: Should I use a `ConcurrentHashMap`?

coollog: Isn't the return type of `ClassLoader#findClass` `Class<?>`?

https://docs.oracle.com/javase/8/docs/api/java/lang/ClassLoader.html#findClass-java.lang.String-

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -36,23 +43,61 @@
   /** Helper for loading a .class file. */
   private static class ClassFileLoader extends ClassLoader {
 
-    private final Path classFile;
+    private static final Map<Path, Class<?>> cache = new ConcurrentHashMap<>();
```
coollog: I think we should avoid having to use `ConcurrentHashMap` over `HashMap` unless it's absolutely necessary. Since for most cases, and in our case here, we won't be using this class in concurrent situations, using `ConcurrentHashMap` would slow down operations in the usual cases. I was thinking to just have the cache be an instance field. Do you think we should have a global cache?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -36,23 +43,61 @@
...
+        }
+
+        byte[] bytes = Files.readAllBytes(file);
+        Class<?> definedClass = defineClass(name, bytes, 0, bytes.length);
+        cache.put(file, definedClass);
+        return definedClass;
+
+      } catch (IOException
+          | ClassFormatError
+          | IndexOutOfBoundsException
```
coollog: Should `IndexOutOfBoundsException` be ignored?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -36,23 +43,61 @@
...
         // Not a valid class file
+        // TODO: Log search class failure when NoClassDefFoundError/SecurityException is caught?
         return null;
       }
     }
+
+    /**
+     * Converts a class name (pack.age.ClassName) to a Path (rootFolder/pack/age/ClassName.class).
+     */
+    private Path getPathFromClassName(String className) {
```
coollog: Should prob have some tests for this method. For example, what would happen if `className` was empty?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
TadCordle: If it's an instance field we would end up using a new empty cache every time we create a new class loader, which we do a bunch (see L198). I could change it back to a `HashMap` and just leave a comment about it not being threadsafe, considering threading here is still a hypothetical. Or we could make it an instance field of `MainClassFinder`, but then it's still effectively global to `ClassFileLoader`. Or we could change the logic so we only ever instantiate one `ClassFileLoader`, but that's an arbitrary extra constraint, and would require changing some path-finding logic.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
TadCordle: According to the javadoc:

```
* @throws  IndexOutOfBoundsException
*          If either <tt>off</tt> or <tt>len</tt> is negative, or if
*          <tt>off+len</tt> is greater than <tt>b.length</tt>.
```

Sounds to me that it's just another indication that it's not a valid class file.

TadCordle: Actually, it sounds like it depends on the parameters that we pass in. So this should never actually be thrown.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
coollog: Hmm, it looks like the line 198 instantiation should be moved out of the loop so there's only one `ClassFileLoader` instantiated.

PR #317, opened 2018-05-24
Adds DockerClient to run docker load.

coollog:
For #261


jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** Tests for {@link DockerClient}. */
+@RunWith(MockitoJUnitRunner.class)
+public class DockerClientTest {
```
TadCordle: Should we add a test for `isDockerInstalled()`? (Is it even possible to test it?)

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -0,0 +1,100 @@
...
+      if (dockerProcess.waitFor() != 0) {
+        throw new IOException("'docker load' command failed with output: " + output);
+      }
+
+      return output;
+    }
+  }
+
+  /** Runs a {@code docker} command. */
+  private Process docker(String... subCommand) throws IOException {
```
TadCordle: This makes the call `docker("load")` look nice, but I feel like taking one line and pulling it into a method that's called once just makes the code a bit harder to follow. Maybe just use a comment if the line is confusing to read.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
coollog: Not too sure how to test this. Any good ideas?

TadCordle: I guess the main thing we'd want to test is making sure running a command throws an IO exception if the command is missing, so maybe we could use mocks to start a `ProcessBuilder` with a local 
"fake docker" script in one test, and ensure it fails when we try to start it with a non-existent script?

TadCordle: Not sure if `ProcessBuilder` can run local scripts or how this would pan out as far as cross-platform test support goes, etc.

coollog:
Made `isDockerInstalled` testable.

TadCordle:
Hmm, looks like windows doesn't like it.

coollog:
Ah right, Windows is not a POSIX system. Will just ignore the exception.

coollog:
Running an executable between Windows and the unix-based systems seemed to be too much of a hassle, so I removed that.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -56,8 +56,13 @@ public void setUp() throws IOException {
   public void testIsDockerInstalled_pass() throws URISyntaxException, IOException {
     Path executableFile = temporaryFolder.newFolder().toPath().resolve("executable");
     Files.copy(Paths.get(Resources.getResource("executable").toURI()), executableFile);
-    // 755
-    Files.setPosixFilePermissions(executableFile, PosixFilePermissions.fromString("rwxr-xr-x"));
+    try {
+      // 755
+      Files.setPosixFilePermissions(executableFile, PosixFilePermissions.fromString("rwxr-xr-x"));
+
+    } catch (UnsupportedOperationException ex) {
```
loosebazooka: We probably want to avoid catching a runtime exception.

Can you try to get the AttributeView and check for nullness instead?

coollog: I opted to remove this executable-running and just use a mock instead since having a separate executable file and logic for Windows seemed like overkill for a single test.

PR #319, opened 2018-05-29
Docker client fix windows

coollog:


TadCordle:
Does kokoro windows even have bash installed?

coollog:
I believe it does (from the Kokoro configuration page)

PR #320, opened 2018-05-29
Fixes MavenSourceFilesConfiguration adds resource directories into the classes layer.

### LINKED ISSUE #318: MavenSourceFilesConfiguration adds resource directories into the classes layer.

coollog:
Fixes #318


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSourceFilesConfiguration.java
```diff
@@ -121,4 +100,39 @@ public String getResourcesPathOnImage() {
   public String getClassesPathOnImage() {
     return CLASSES_PATH_ON_IMAGE;
   }
+
+  /**
+   * Adds {@code file} to {@link #classesFiles} if it is a {@code .class} file and there is a
+   * corresponding {@code .java} file in the {@code classesSourceDirectory}; otherwise, adds {@code
+   * file} to {@link #resourcesFiles}.
+   */
+  private void addFileToResourcesOrClasses(
```
loosebazooka: I think we have to deal with inner/nested classes here? There might be other quirks we're not covering.
```
MyClass.java
    |
    |
    V
MyClass.class
MyClass$MyInner.class
```

PR #355, opened 2018-06-02
Fixes main class finding with external dependencies.

### LINKED ISSUE #331: Main class inference fails on files with external dependencies
Tried running jib on https://github.com/spring-guides/gs-spring-boot/blob/master/complete with no explicit main class parameter. Fails with "Main class was not found, perhaps you should add a `mainClass` configuration to jib-maven-plugin".

This is caused by a `NoClassDefFoundError` when the class finder is processing `Application.class`. It looks like when it hits the `commandLineRunner` method, it calls `findClass` on `org.springframework.context.ApplicationContext`, but fails to find the corresponding .class file since it's an external dependency.

coollog:
Fixes #331 

Mainly just uses javassist to create an empty placeholder class for external dependencies.


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -44,45 +46,70 @@
...
-    public Class<?> findClass(String name) {
-      return findClass(name, getPathFromClassName(name));
+    public Class<?> findClass(String className) {
+      Path classFile = getPathFromClassName(className);
+
+      if (!Files.exists(classFile)) {
+        // TODO: Log search class failure?
+        // Cannot find corresponding class file. The class is probably in a dependency JAR.
+        // Makes an empty class to prevent NoClassDefFoundError.
+        try {
```
patflynn: This seems very hacky. 

Just to learn a little bit about the problem, the issue here is that the main class isn't in the compiled .class output files and in a dep?

When I was seeing this issue with my dropwizard project I don't think that should have been the case.

loosebazooka: The old issue was that the classloader couldn't load a class if the things it imports weren't resolvable. For instance some jar dep wasn't loaded when we were searching through the classes. Which is what was hitting you probably with dropwizard.

But we should probably look into this a little more.

coollog: In this custom classloader, there's two paths for finding a class. The first is resolving a class that is provided in a file, and second is when the JDK is resolving classes that were used in a class loaded from a file. In the second case, classes that don't exist would cause a `NoClassDefFoundError`. However, since all we care is finding the main class in the provided class files, this error doesn't matter, and this change is a way to tell the JDK that it doesn't matter if classes don't exist since all we are looking for is psvm.

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/MainClassFinderTest.java
```diff
@@ -99,9 +99,18 @@ public void testFindMainClass_extension() throws URISyntaxException, IOException
   }
 
   @Test
-  public void testFindMainClass_externalmethod() throws URISyntaxException, IOException {
+  public void testFindMainClass_importedmethods() throws URISyntaxException, IOException {
```
TadCordle: `importedmethods` -> `importedMethods`

jib-maven-plugin/pom.xml
```diff
@@ -116,6 +122,12 @@
       <version>2.12.0</version>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
```
TadCordle: What's this for?

coollog: Oh oops this was for testing.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -177,47 +115,60 @@ public static String resolveMainClass(
...
+
+      new DirectoryWalker(rootDirectory)
+          .filter(Files::isRegularFile)
+          .filter(path -> path.toString().endsWith(".class"))
+          .walk(
+              classFile -> {
+                try (InputStream classFileInputStream = Files.newInputStream(classFile)) {
+                  CtClass fileClass = classPool.makeClass(classFileInputStream);
+
+                  try {
```
TadCordle: Would it be cleaner to have just one try block with multiple catch clauses instead of having them nested?

Issue #323, opened 2018-05-29
Using image tag with slashes very slow when parsing.


coollog:
For example, this image reference: `gcr.io/qingyangc-sandbox/jibtestimage:askfjl/waegawvw/aweg`

Took 3 min to parse.

coollog:
https://github.com/GoogleContainerTools/jib/tree/fix-imagereference2

coollog:
@chanseokoh I have this PR #357 that's been sitting around for a while - it doesn't solve the issue but it might help write a better regex - feel free to use/not use it/close it in favor of a new solution

PR #357, opened 2018-06-02
Adds RegexBuilder.

coollog:
Makes regular expressions more readable.

Helps with debugging #323.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
@@ -71,14 +71,13 @@ public class ImageReference {
         sequence(any(group(REPOSITORY_PART, literal('/'))), REPOSITORY_PART);
 
     /** Matches a tag of max length 128. */
-    private static final String TAG_REGEX =
-        sequence(wordChars(), range(wordChars('.', '-'), 0, 127));
+    private static final String TAG = sequence(wordChars(), range(wordChars('.', '-'), 0, 127));
 
     /**
      * Matches a full image reference, which is the registry, repository, and tag/digest separated
      * by backslashes. The repository is required, but the registry and tag/digest are optional.
      */
-    private static final String REFERENCE_REGEX =
+    private static final String REFERENCE =
         sequence(
             begin(),
             optional(group(match(REGISTRY), literal('/'))),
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -31,53 +32,73 @@
...
+     * in the middle.
+     */
+    private static final String REGISTRY_PART =
+        group(alphanum(), group(any(alphanum('-')), alphanum()));
+
+    /**
+     * Matches sequences of {@link #REGISTRY_PART} separated by a dot, with an optional {@code
+     * :port} at the end.
+     */
+    private static final String REGISTRY =
```
patflynn: I'm a little skeptical of this change.  It seems to replace something that I'm very familiar with (regex strings),  with a sort of custom DSL that I'd have to do a bunch of work to understand.

jib-core/src/main/java/com/google/cloud/tools/jib/regex/RegexBuilder.java
```diff
@@ -0,0 +1,112 @@
...
+
+package com.google.cloud.tools.jib.regex;
+
+import java.util.regex.Matcher;
+
+/**
+ * Static methods for constructing a regular expression in a more readable functional form. Extend
+ * this class to use the static methods.
+ */
+public abstract class RegexBuilder {
```
patflynn: There's some really interesting ideas in here, but this abstract class with static methods seems strange. Why wouldn't you just use the utility class pattern instead?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
coollog: Yeah, it does provide a custom DSL that requires some extra work to understand. The intention was to try to follow what Docker does for its reference regex: https://github.com/docker/distribution/blob/master/reference/regexp.go

jib-core/src/main/java/com/google/cloud/tools/jib/regex/RegexBuilder.java
coollog: The main reason was to be able to encapsulate regex constants in a child class and also be able to have that class construct the regex using the static methods without statically importing the methods for visibility to others.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -31,53 +32,73 @@
...
+            any(group(literal('.'), REGISTRY_PART)),
+            optional(group(literal(':'), repeated(digit()))));
+
+    /**
+     * Matches all sequences of lowercase alphanumeric characters separated by a separator.
+     *
+     * <p>A separator is either an underscore, a dot, two underscores, or any number of dashes.
+     */
+    private static final String REPOSITORY_PART =
+        sequence(
```
TadCordle: I found this one to be pretty confusing. Could this expression be simplified? Or, I wonder if some of these could be more clear by using a Regex-DSL hybrid rather than just one or the other.

loosebazooka:
Lets hold off on this one till after the beta.

coollog:
Closing this as the issue is fixed in #680.

Issue #346, opened 2018-06-01
Allow user to set default arguments to main.


coollog:
drop wizard requires it

@patflynn 

- [x] Add `Cmd` to `ContainerConfigurationTemplate`
- [x] Add `javaArguments` to `BuildConfiguration`
- [x] Set `Cmd` in `ImageToJsonTranslator`
- [ ] Add parameter `args` to `jib-maven-plugin`
- [x] Add parameter `args` to `jib-gradle-plugin`

coollog:
Should we have the parameter be called `args`, `mainArgs`, or `cmd`?

patflynn:
+1 to `args`

```
~ $ java
Usage: java [-options] class [args...]
           (to execute a class)

```

TadCordle:
+1 `args`

coollog:
Looks like `args` it is

PR #358, opened 2018-06-04
Add core changes for default main arguments

TadCordle:
Part of #346


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
```diff
@@ -76,7 +76,9 @@ public static BuildSteps forBuildToDockerRegistry(
                 .runPullAndCacheBaseImageLayersStep()
                 .runPushBaseImageLayersStep()
                 .runBuildAndCacheApplicationLayerSteps()
-                .runBuildImageStep(getEntrypoint(buildConfiguration, sourceFilesConfiguration))
+                .runBuildImageStep(
+                    getEntrypoint(buildConfiguration, sourceFilesConfiguration),
+                    buildConfiguration.getJavaArguments())
```
coollog: `buildConfiguration` is already passed into the `BuildImageStep`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -55,14 +56,22 @@
       return this;
     }
 
+    public Builder<T> setCmd(List<String> cmd) {
```
coollog: short javadoc

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -79,11 +88,18 @@
...
   private Image(
-      ImageLayers<T> layers, ImmutableList<String> environment, ImmutableList<String> entrypoint) {
+      ImageLayers<T> layers,
+      ImmutableList<String> environment,
+      ImmutableList<String> entrypoint,
+      ImmutableList<String> cmd) {
     this.layers = layers;
     this.environmentBuilder = environment;
     this.entrypoint = entrypoint;
+    this.cmd = cmd;
```
coollog: We should use non-abbreviations for the `Image` class since it is a model and does not need to have the fields be as the same as in the JSON.

TadCordle: "command" or "javaArguments"?

PR #360, opened 2018-06-04
Add 'args' parameter to Gradle plugin, add main args to Docker context generation

TadCordle:
Part of #346.


jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -119,35 +126,36 @@ String makeDockerfile() throws IOException {
...
    * @see <a
    *     href="https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example">https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example</a>
    */
   @VisibleForTesting
-  String getEntrypoint() {
-    List<String> entrypoint =
-        EntrypointBuilder.makeEntrypoint(sourceFilesConfiguration, jvmFlags, mainClass);
-
-    StringBuilder entrypointString = new StringBuilder("[");
+  String makeDockerList(List<String> items) {
```
coollog: Looks like this can be a static method now. Can make this more descriptively named like `joinAsJsonArray`, which is the form that Dockerfile takes.

PR #361, opened 2018-06-04
Add 'args' parameter to Maven plugin

### LINKED ISSUE #346: Allow user to set default arguments to main.

TadCordle:
After #360 because of Docker context stuff.

Fixes #346.


jib-maven-plugin/src/test/resources/projects/simple/pom.xml
```diff
@@ -41,6 +41,7 @@
           <to>
             <image>gcr.io/jib-integration-testing/simpleimage:maven</image>
           </to>
+          <args>An argument.</args>
```
coollog: Didn't know you could actually put it like this.

TadCordle: I didn't either, it was an accident but it worked haha.

PR #364, opened 2018-06-05
Bumps versions to 0.9.0-SNAPSHOT.

coollog:


coollog:
Yea, `jib-core`'s version is never actually bumped because we don't release it.

PR #367, opened 2018-06-05
Ignores InsecureRecursiveDeleteException.

coollog:
This is causing our macos CI builds to fail.

### DIFF
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
@@ -157,9 +157,18 @@ public class JibPluginIntegrationTest {
             .resolve("main")
             .resolve("resources")
             .resolve("newfile"));
-    BuildTask reexecutedJibDockerContextTask =
-        simpleTestProject.build("jibDockerContext").task(":jibDockerContext");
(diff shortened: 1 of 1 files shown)


jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
```diff
@@ -157,9 +157,19 @@ public void testDockerContext() throws IOException, InterruptedException {
...
-        simpleTestProject.build("jibDockerContext").task(":jibDockerContext");
-    Assert.assertNotNull(reexecutedJibDockerContextTask);
-    Assert.assertEquals(TaskOutcome.SUCCESS, reexecutedJibDockerContextTask.getOutcome());
+    try {
+      BuildTask reexecutedJibDockerContextTask =
+          simpleTestProject.build("jibDockerContext").task(":jibDockerContext");
+      Assert.assertNotNull(reexecutedJibDockerContextTask);
+      Assert.assertEquals(TaskOutcome.SUCCESS, reexecutedJibDockerContextTask.getOutcome());
+
+    } catch (UnexpectedBuildFailure ex) {
```
loosebazooka: can you just catch the insecureRescursiveDeleteException directly?

coollog: Since this is running in the `GradleRunner`, it doesn't actually expose the stacktrace from the task failure.

PR #375, opened 2018-06-05
Add default target image name parameter for build to Docker daemon

### LINKED ISSUE #352: Provide default target image reference for build to Docker daemon.
It’d be great to create a default ‘to.image’ if it’s not specified `project-name:version`.

TadCordle:
When `to.image` is not configured, building to Docker daemon will now succeed and give the image the default tag of `projectName:projectVersion` using properties defined in build.gradle/pom.xml.

When running the original image build task with a missing target image configuration, it throws an error.

Fixes #352


TadCordle:
Note - tested this manually; going to write automated tests tomorrow.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -66,6 +67,14 @@ public void buildImage() throws InvalidImageReferenceException {
     // Asserts required @Input parameters are not null.
     Preconditions.checkNotNull(jibExtension);
 
+    String targetImage = jibExtension.getTargetImage();
+    if (Strings.isNullOrEmpty(targetImage)) {
+      throw new GradleException(
+          "Missing target image parameter. Add a 'to.image' configuration parameter to your "
```
coollog: Maybe use `jib.to.image`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -66,6 +67,14 @@ public void buildImage() throws InvalidImageReferenceException {
...
+              + "build.gradle or set the parameter via commandline (e.g. 'gradle jib --image "
+              + "your.image/name').");
```
coollog: We should probably just have this be like `<your image name>` without the symbols, which might be confusing.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -48,7 +49,14 @@ public void execute() throws MojoExecutionException {
 
     // Parses 'from' and 'to' into image reference.
     ImageReference baseImage = parseBaseImageReference(getBaseImage());
-    ImageReference targetImage = parseTargetImageReference(getTargetImage());
+
+    // Parses and validates 'to'.
+    String targetImageString = getTargetImage();
+    if (Strings.isNullOrEmpty(targetImageString)) {
+      // Use default value "project-name:project-version"
+      targetImageString = getProject().getName() + ":" + getProject().getVersion();
```
coollog: I think for this we could just directly build the `ImageReference` rather than passing it through the parser. However, we do need to be careful if the name or version is not in the valid format for a repository or tag. This can just be a TODO for now, but we should take care of these cases in the future.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -74,9 +75,18 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-    // Parses 'from' and 'to' into image reference.
+    // Parses 'from' into image reference.
     ImageReference baseImage = parseBaseImageReference(getBaseImage());
-    ImageReference targetImage = parseTargetImageReference(getTargetImage());
+
+    // Parses and validates 'to'.
+    String targetImageString = getTargetImage();
+    if (Strings.isNullOrEmpty(targetImageString)) {
+      throw new MojoFailureException(
+          "Missing target image parameter. Add a <to.image> configuration parameter to your "
```
coollog: A TODO for consolidating this with the Gradle message.

coollog: `<to.image>` might be misleading. How about `<to><image>`?

patflynn:
:+1: 

I guess it's tricky to make this work for registry based jib builds huh?  There'd need to be a way to specify registry and project but without the final image name and tag component.  Sure would be nice though.. :)

TadCordle:
Yeah, we talked a bit about it yesterday. Figured it would be a bit too involved to fit into the beta release, so maybe after!

patflynn:
:+1:

PR #377, opened 2018-06-05
Fixes Jib task dependency on classes.

### LINKED ISSUE #374: 'classes' dependency should be handled after project evaluation
Currently if the java plugin is applied after the jib plugin, builds will fail due to error:
```
A problem occurred evaluating root project 'hello-gradle'.
> Failed to apply plugin [id 'com.google.cloud.tools.jib']
   > Task with path 'classes' not found in root project 'hello-gradle'.
```

coollog:
Fixes #374

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
@@ -48,8 +48,8 @@ public class JibPlugin implements Plugin<Project> {
 
     // Has all tasks depend on the 'classes' task.
     project.afterEvaluate(
-        ignored -> {
-          Task classesTask = project.getTasks().getByPath("classes");
+        projectAfterEvaluation -> {
+          Task classesTask = projectAfterEvaluation.getTasks().getByPath("classes");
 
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
```diff
@@ -32,23 +32,29 @@ public void apply(Project project) {
...
+            .setJibExtension(jibExtension);
+    Task buildDockerTask =
+        project
+            .getTasks()
+            .create("jibBuildDocker", BuildDockerTask.class)
+            .setJibExtension(jibExtension);
+
+    // Has all tasks depend on the 'classes' task.
+    project.afterEvaluate(
+        ignored -> {
```
loosebazooka: I wonder if the project object in the afterEvaluate block has different properties than the one calling afterEvaluate?

loosebazooka: the one we've `ignored`

coollog: Hmm, okay I'll change to use the one that was ignored.

PR #391, opened 2018-06-06
Consolidate missing target image config error

TadCordle:


jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
```diff
@@ -111,9 +111,9 @@ public void testBuild_defaultTarget() {
       Assert.assertThat(
           ex.getMessage(),
           CoreMatchers.containsString(
-              "Missing target image parameter. Add a 'jib.to.image' configuration parameter to "
-                  + "your build.gradle or set the parameter via commandline (e.g. 'gradle jib "
-                  + "--image <your image name>')."));
+              "Missing target image parameter, perhaps you should add a 'jib.to.image' "
+                  + "configuration parameter to your build.gradle or set the parameter via "
+                  + "commandline (e.g. 'gradle jib --image <your image name>')."));
```
coollog: via *the* commandline?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulSuggestions.java
```diff
@@ -125,4 +125,15 @@ private String forNoCredentialHelpersDefined(
             + "' or "
             + authConfiguration);
   }
+
+  public String forToNotConfigured(String parameter, String buildConfigFilename, String command) {
```
coollog: Should go above private method.

coollog: Maybe also a short javadoc since the name and parameters might be a bit confusing.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulSuggestions.java
```diff
@@ -108,6 +108,24 @@ public String forDockerNotInstalled() {
     return suggest("make sure Docker is installed and you have correct privileges to run it");
   }
 
+  /**
+   * @param parameter the parameter name (should be 'to.image' for gradle, {@literal <to><image>}
```
coollog: We prob shouldn't include specificities for the caller here, since if we change the parameter name, this comment would need to be changed. Having an example is good though.

PR #392, opened 2018-06-06
Fix javadoc warnings when building gradle plugin

### LINKED ISSUE #280: Fix :javadoc warnings when building gradle plugin
There's a lot of them and they're ugly

TadCordle:
Fixes #280 

Although we may want to look into this more, and either fix all the warnings by fixing our javadocs, or limit the type of warnings that's suppressed (more options here: https://gist.github.com/hekar/62772af5b0fe0b6731bb)

### DIFF
--- a/jib-core/build.gradle
+++ b/jib-core/build.gradle
@@ -81,7 +81,7 @@ tasks.withType(JavaCompile) {
 
 // Fail build on javadoc warnings
 tasks.withType(Javadoc) {
-    options.addStringOption('Xwerror', '-quiet')
+    options.addBooleanOption('Xwerror', true)
 }
 
 tasks.withType(Test) {
--- a/jib-core/build.gradle
+++ b/jib-core/build.gradle
@@ -80,10 +80,8 @@ tasks.withType(JavaCompile) {
 }
 
 // Fail build on javadoc warnings
-if (JavaVersion.current().isJava8Compatible()) {
-    tasks.withType(Javadoc) {
-        options.addStringOption('Xwerror', '-quiet')
-    }
+tasks.withType(Javadoc) {
+    options.addStringOption('Xwerror', '-quiet')
 }
 
 tasks.withType(Test) {
--- a/jib-core/build.gradle
+++ b/jib-core/build.gradle
@@ -79,6 +79,13 @@ tasks.withType(JavaCompile) {
     }
 }
 
+// Fail build on javadoc warnings
+if (JavaVersion.current().isJava8Compatible()) {
+    tasks.withType(Javadoc) {
+        options.addStringOption('Xwerror', '-quiet')
+    }
+}
+
 tasks.withType(Test) {
     reports.html.setDestination file("${reporting.baseDir}/${name}")
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildLogger.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildLogger.java
@@ -20,6 +20,11 @@ public interface BuildLogger {
 
   void error(CharSequence message);
 
+  /**
+   * Logs messages as part of normal execution (default log level).
+   *
+   * @param message the message to log
+   */
   void lifecycle(CharSequence message);
 
   void warn(CharSequence message);
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildLogger.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildLogger.java
@@ -20,7 +20,6 @@ public interface BuildLogger {
 
   void error(CharSequence message);
 
-  /** Logs messages as part of normal execution (default log level). */
   void lifecycle(CharSequence message);
 
   void warn(CharSequence message);
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
@@ -60,31 +60,54 @@ public class DockerContextGenerator {
   /**
    * Sets the base image for the {@code FROM} directive. This must be called before {@link
    * #generate}.
+   *
+   * @param baseImage the base image.
+   * @return The {@link DockerContextGenerator} with the base image set.
    */
   public DockerContextGenerator setBaseImage(String baseImage) {
     this.baseImage = baseImage;
     return this;
   }
 
-  /** Sets the JVM flags used in the {@code ENTRYPOINT}. */
+  /**
+   * Sets the JVM flags used in the {@code ENTRYPOINT}.
+   *
+   * @param jvmFlags the jvm flags.
+   * @return The {@link DockerContextGenerator} with the jvm flags set.
+   */
   public DockerContextGenerator setJvmFlags(List<String> jvmFlags) {
     this.jvmFlags = jvmFlags;
     return this;
   }
 
-  /** Sets the main class used in the {@code ENTRYPOINT}. */
+  /**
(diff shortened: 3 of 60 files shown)


coollog:
Yeah, we probably don't want to disable doclint completely since there's a lot of useful checks in there. If we want to fix that issue, we should probably just fix all the warnings.

TadCordle:
I think I'll work on that then. Fixing javadocs has the added benefit of getting me more familiar with the whole codebase as well.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -60,31 +60,54 @@ public DockerContextGenerator(SourceFilesConfiguration sourceFilesConfiguration)
   /**
    * Sets the base image for the {@code FROM} directive. This must be called before {@link
    * #generate}.
+   *
+   * @param baseImage the base image.
+   * @return The {@link DockerContextGenerator} with the base image set.
```
coollog: nit: lowercase for start of `@...` docs and no period

coollog: can also just simplify the `@return` to like `@return this`

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/DirectoryWalker.java
```diff
@@ -31,21 +31,35 @@
...
       throw new NotDirectoryException(rootDir + " is not a directory");
     }
     this.rootDir = rootDir;
   }
 
-  /** Adds a filter to the walked paths. */
+  /**
+   * Adds a filter to the walked paths.
+   *
+   * @param pathFilter the filter.
```
coollog: `filter returns {@code true} if the path should be accepted and {@code false} otherwise`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulSuggestions.java
```diff
@@ -112,7 +112,10 @@ public String none() {
     return messagePrefix;
   }
 
-  /** @return the message containing a suggestion */
+  /**
+   * @param suggestion a suggested fix for the problem described by {@code messagePrefix}
```
coollog: I think for here you can actually use `{@link #messagePrefix}`.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
TadCordle: Oh, should we not be using periods? It seems to be 50/50 everywhere so I've been adding them.

coollog: Yeah, I believe the style is to only use a period for these block tags if you have a second sentence afterwards. I can't quite find where this convention was stated at though...

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -32,36 +32,71 @@
...
     public Builder<T> setEnvironmentVariable(String name, String value) {
       environmentBuilder.add(name + "=" + value);
       return this;
     }
 
-    /** Adds an environment variable definition in the format {@code NAME=VALUE}. */
+    /**
+     * Adds an environment variable definition in the format {@code NAME=VALUE}.
+     *
+     * @param environmentVariableDefinition the string to add
```
coollog: the *definition* to add?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
```diff
@@ -101,12 +114,19 @@ public boolean isEmpty() {
...
+   * @return the layer at the specified index
+   */
   public T get(int index) {
     return layers.get(index);
   }
 
-  /** @return the layer by digest, or {@code null} if not found */
+  /**
+   * @param digest the digest used to retrieve the layer
+   * @return the layer by digest, or {@code null} if not found
```
coollog: the layer *found*?

jib-gradle-plugin/build.gradle
```diff
@@ -98,6 +98,13 @@ tasks.withType(JavaCompile) {
     }
 }
 
+// Fail build on javadoc warnings
+if (JavaVersion.current().isJava8Compatible()) {
```
coollog: We can just configure this directly on `googleJavaFormat` right?

coollog: We should also add this for core and maven.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildLogger.java
```diff
@@ -20,7 +20,6 @@
 
   void error(CharSequence message);
 
-  /** Logs messages as part of normal execution (default log level). */
```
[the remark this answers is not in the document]
TadCordle: It seemed out of place as the only javadoc'd method, and I thought it was self-explanatory enough to remove. But on second thought, the "default log level" part seems important, so I'll put it back.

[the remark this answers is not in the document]
coollog: Yea, this is to make sure callers user lifecycle for normal default log messages (compared to the other ones, which will only show if some verbosity is enabled).

jib-gradle-plugin/build.gradle
TadCordle: How (for both of those)?

coollog: It might not be possible but it looks like the gradle plugin can define options as part of the extension: https://github.com/sherter/google-java-format-gradle-plugin#extended-usage

coollog: Oh okay that's probably my misunderstanding

jib-core/build.gradle
```diff
@@ -79,6 +79,13 @@ tasks.withType(JavaCompile) {
     }
 }
 
+// Fail build on javadoc warnings
+if (JavaVersion.current().isJava8Compatible()) {
```
coollog: Is this compatibility check needed? Our codebase is completely in java 8.

jib-core/build.gradle
```diff
@@ -79,6 +79,13 @@ tasks.withType(JavaCompile) {
...
+    tasks.withType(Javadoc) {
+        options.addStringOption('Xwerror', '-quiet')
```
coollog: What does the `-quiet` do?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -57,7 +57,10 @@ public void set(String image) {
     }
   }
 
-  /** @return the {@link ImageReference} parsed from {@code from}. */
+  /**
+   * @param from the image reference string to parse.
+   * @return the {@link ImageReference} parsed from {@code from}.
+   */
   static ImageReference parseBaseImageReference(String from) {
```
coollog: Just noticed these two methods are essentially identical. We could prob just have one method and a second parameter that's the 'parameter name'.

jib-core/build.gradle
TadCordle: I think it's just a hack to fill the second argument of `addStringOption`, since gradle automatically adds the '-quiet' option anyway.

coollog: It looks like you could possibly try `.addBooleanOption('Xwerror', true)`

https://github.com/gradle/gradle/issues/1393#issuecomment-280287314

Issue #388, opened 2018-06-06
Add support for registries with just HTTP.


coollog:

gintautassulskus:
### Context
I am running docker-registry on a k8s cluster. The access to the registry is via:
`LoadBalancer --- Ingress Controller -- Docker-Registry Service`

Unless I set the Load Balancer to https:443 and configure the Ingress Controller with TLS termination, `jib` fails with: `Connection refused `.  My assumption is that jib supports only https over 443.

Docker client, on the other hand, works just fine with http:80 and  https:443.

Probably it would suffice to document the constraints of jib, e.g.: `works only with secure connections https:443`

Log output:
```
[INFO] Pushing image as docker-registry.cluster.local/demoapp:built-with-jib
[INFO] 
[INFO] RUNNING  Building and pushing image
[INFO] RUNNING  Retrieving registry credentials for docker-registry.cluster.local
[INFO] RUNNING  Retrieving registry credentials for docker-registry.cluster.local
[INFO] No credentials could be retrieved for registry docker-registry.cluster.local
[INFO] Checking credentials from docker-credential-gcr
[INFO] Retrieving registry credentials for docker-registry.cluster.local : 12.627 ms
[INFO] RUNNING  Authenticating with push to docker-registry.cluster.local
[WARNING] The system does not have docker-credential-gcr CLI
[INFO] No credentials could be retrieved for registry gcr.io
[INFO] Retrieving registry credentials for docker-registry.cluster.local : 24.5 ms
[INFO] RUNNING  Building application layers
[INFO] RUNNING  Authenticating pull from gcr.io
[INFO] RUNNING  Building dependencies layer
[INFO] RUNNING  Building resources layer
[INFO] Building application layers : 2.203 ms
[INFO] RUNNING  Building classes layer
[INFO] RUNNING  Setting up to push layers
[INFO] Building dependencies layer : 5.326 ms
[INFO] Setting up to push layers : 3.88 ms
[INFO] Building classes layer : 11.36 ms
[INFO] Building resources layer : 41.223 ms
[INFO] Authenticating with push to docker-registry.cluster.local : 547.814 ms
[INFO] RUNNING  Pushing BLOB sha256:df0e86cb92bfa8ae506bed53e83a81238797203b4e4c85f9d623f673613fa2ef
[INFO] RUNNING  Pushing BLOB sha256:e284665cc3282200f29c695c15c6424a7f659e64b21073cb68d4de492a599b09
[...]

PR #394, opened 2018-06-07
Moves registry endpoint calling to separate class that manages request state.

coollog:
This should help with fixing #388, as it requires handling `HttpHostConnectException` with different requests.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
@@ -1,0 +1,162 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.registry;
+
+import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpResponseException;
+import com.google.api.client.http.HttpStatusCodes;
+import com.google.cloud.tools.jib.http.Authorization;
+import com.google.cloud.tools.jib.http.Connection;
+import com.google.cloud.tools.jib.http.Request;
+import com.google.cloud.tools.jib.http.Response;
+import com.google.cloud.tools.jib.json.JsonTemplateMapper;
+import com.google.cloud.tools.jib.registry.json.ErrorEntryTemplate;
+import com.google.cloud.tools.jib.registry.json.ErrorResponseTemplate;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import javax.annotation.Nullable;
+import javax.net.ssl.SSLPeerUnverifiedException;
+import org.apache.http.NoHttpResponseException;
+
+/** Makes requests to a registry endpoint. */
+class RegistryEndpointCaller<T> {
+
+  private static final String PROTOCOL = "https";
+
+  /** Maintains the state of a request. This is used to retry requests with different parameters. */
+  private static class RequestState {
+
+    @Nullable private final Authorization authorization;
+    private final URL url;
+
+    /**
+     * @param authorization authentication credentials
+     * @param url the endpoint URL to call, or {@code null} to use default from {@code
+     *     registryEndpointProvider}
+     */
+    private RequestState(@Nullable Authorization authorization, URL url) {
+      this.authorization = authorization;
+      this.url = url;
+    }
+  }
+
+  private final RequestState initialRequestState;
+  private final String userAgent;
+  private final RegistryEndpointProvider<T> registryEndpointProvider;
+  private final RegistryEndpointProperties registryEndpointProperties;
+
+  /**
+   * Constructs with parameters for making the request.
+   *
+   * @param userAgent {@code User-Agent} header to send with the request
+   * @param apiRouteBase the endpoint's API root, without the protocol
+   * @param registryEndpointProvider the {@link RegistryEndpointProvider} to the endpoint
+   * @param authorization optional authentication credentials to use
+   * @param registryEndpointProperties properties of the registry endpoint request
+   */
+  RegistryEndpointCaller(
+      String userAgent,
+      String apiRouteBase,
+      RegistryEndpointProvider<T> registryEndpointProvider,
+      @Nullable Authorization authorization,
+      RegistryEndpointProperties registryEndpointProperties)
+      throws MalformedURLException {
+    this.initialRequestState =
+        new RequestState(
+            authorization, registryEndpointProvider.getApiRoute(PROTOCOL + "://" + apiRouteBase));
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Constructs with parameters for making the request.
+   *
+   * @param userAgent {@code User-Agent} header to send with the request
+   * @param apiRouteBase the endpoint's API root, without the protocol
+   * @param registryEndpointProvider the {@link RegistryEndpointProvider} to the endpoint
+   * @param authorization optional authentication credentials to use
+   * @param registryEndpointProperties properties of the registry endpoint request
+   */
```
TadCordle: Need `@throws` to avoid more javadoc warnings

loosebazooka:
Can you points out the fix here? The refactor + fix makes it hard to see the change

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointProperties.java
```diff
@@ -16,7 +16,8 @@
 
 package com.google.cloud.tools.jib.registry;
 
-/** Properties of registry endpoints. */
+/** Properties of registry endpoint requests. */
+// TODO: Probably rename to RegistryEndpointRequestProperties.
```
TadCordle: Why not do this now?

coollog:
@loosebazooka This actually doesn't have the fix. This still has the same functionality as before and is purely a refactor so that we can implement the fix by using the new `RequestState` to manage retries of the registry request.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -0,0 +1,162 @@
...
+      throws MalformedURLException {
+    this.initialRequestState =
+        new RequestState(
+            authorization, registryEndpointProvider.getApiRoute(DEFAULT_PROTOCOL + "://" + apiRouteBase));
+    this.userAgent = userAgent;
+    this.registryEndpointProvider = registryEndpointProvider;
+    this.registryEndpointProperties = registryEndpointProperties;
+  }
+
+  /** Makes the request to the endpoint. */
```
TadCordle: Another javadoc thing - should have `@param <T>`, `@returns`, and `@throws` to avoid warnings.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointProperties.java
coollog: It requires lots of changes in other files (like tests) that I thought would be better to have in a separate PR to make this one simpler to review.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
coollog: I'll merge your javadoc PR so that the build fails if I don't fix these.

coollog:
@TadCordle hmm I don't seem to be getting any javadoc errors or warnings

TadCordle:
Hmm, I tried removing a couple javadoc parameters from Blob.java, and it threw the errors. It looks like it's ignoring your files for some reason?

loosebazooka:
Does the gradle javadoc generation handle the extra code coming from an external place (jib-core?)

TadCordle:
@loosebazooka The gradle plugin does, but the maven one doesn't at the moment.

coollog:
I removed some javadoc from `Blob.java` and ran `./gradlew goJF build` in `jib-core`, and I don't seem to be getting any warnings/errors either.

TadCordle:
@coollog The gradle plugin doesn't error on completely missing javadocs, just javadocs with missing parameters. I think the maven plugin will error on missing javadocs though. Maybe we should fix that.

It looks like the errors aren't being thrown on `RegistryEndpointCaller` because the class and constructor are package private, if you change them to public you should get the errors.

coollog:
Yeah, I removed some some block tags and it still doesn't give me any errors/warnings.

coollog:
Is there a way to enable the checks for package private classes?

TadCordle:
At the bottom of `build.gradle` there's something like this:

```
def fileExtensionsBefore = '<property name="fileExtensions" value="java, properties, xml"/>'
def fileExtensionsAfter = '<property name="fileExtensions" value="properties, xml"/>'
def googleChecksNoJava = googleChecks.replace(fileExtensionsBefore, fileExtensionsAfter)
```

We could maybe do an additional replace like:
```
def scopeBefore = '<property name="scope" value="public"/>'
def scopeAfter = '<property name="scope" value="package"/>'
```

coollog:
Hmm, that seems to be what's suppressing it. Though how were you able to get the warnings/errors?

TadCordle:
try `./gradlew clean goJF build`

coollog:
Yeah, `clean` didn't seem to give errors/warnings.

TadCordle:
And I don't think checkstyle is what's suppressing it, it looks like it's supposed to allow missing tags: https://github.com/checkstyle/checkstyle/blob/0579f2dd058a5cdca0bdc89f2b864fbd60ae5c02/src/main/resources/google_checks.xml#L230

coollog:
Perhaps what we need to do is just use the same `google-checks-no-indent.xml` as in `jib-maven-plugin`.

coollog:
Hmm if checkstyle is allowing it, then what is giving the warnings for your build?

TadCordle:
It's coming from a `:javadoc` task, which I think comes from the `java` plugin.

coollog:
Ah so it looks like when I run `build`, `javadoc` isn't actually part of the tasks:

```shell
$ ./gradlew build --console=plain
:compileJava UP-TO-DATE
:processResources UP-TO-DATE
:classes UP-TO-DATE
:jar UP-TO-DATE
:assemble UP-TO-DATE
:compileTestJava UP-TO-DATE
:processTestResources UP-TO-DATE
:testClasses UP-TO-DATE
:compileIntegrationTestJava UP-TO-DATE
:processIntegrationTestResources UP-TO-DATE
:integrationTestClasses UP-TO-DATE
:checkstyleIntegrationTest UP-TO-DATE
:checkstyleMain UP-TO-DATE
:checkstyleTest UP-TO-DATE
:test UP-TO-DATE
:verifyGoogleJavaFormat
:check
:build

BUILD SUCCESSFUL in 0s
12 actionable tasks: 1 executed, 11 up-to-date
```

But if I just run `./gradlew javadoc` I get the errors.

loosebazooka:
Java doc is only configured to run on release, it might be expensive to run on every build, but we can add it to the check task (maybe there's a check java doc?)

loosebazooka:
is this ready for review?

coollog:
@loosebazooka Yes, although we haven't figured out the javadoc check, but that can just be handled separately.

loosebazooka:
We should also revisit  our custom checkstyle configs, I believe google-java-format and checkstyle at their latest versions are mostly in-line with eachother.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
loosebazooka: maybe fix these for now? before we submit?

coollog:
Sounds good, I'll file an issue for that.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
coollog: The `@param` is taken care of at the class-level.

PR #398, opened 2018-06-11
Adds archive for v0.9.0 docs.

coollog:
This copies the current docs to `docs/v0.9.0`. Any changes for unreleased versions should be made in a `docs/head` folder.


coollog:
@elharo The docs directory generates the website at https://googlecontainertools.github.io/jib/. We are trying to provide documentation that is easily accessible for any version of Jib users may be using. This is to sort of be like what some other docs do like https://docs.docker.com/engine/api/version-history/.

coollog:
@elharo Yes, we should only have one copy of old doc versions in the corresponding version folder in `docs/`. Any other user-facing places should be kept up-to-date with the latest version.

coollog:
I'm not sure, but I don't think there's an easy way to see an old version of a Github Pages website by having it render at some commit. Looking though git history for a README for a certain version may also be too hasslesome for users since we don't label the commits for updating READMEs for certain versions consistently with reference to that version. We can also move the docs to a separate repo like a lot of other projects do so that we can keep documentation commits in a separate history.

TadCordle:
We could just link to the source for previous versions of docs. e.g. something like this link: https://github.com/GoogleContainerTools/jib/tree/03d42652e82c543d0a6c6b69bd8827ec8effd82d/docs

It's maybe not the best user experience since only the latest docs would be rendered in the Github Pages style, but I think it's a happy medium between cluttering the repo (or maintaining a new repo) and making them manually search the git history.

PR #404, opened 2018-06-14
Updates CHANGELOG with issue #401.

coollog:
Will release `v0.9.1` after this.


coollog:
Going to put unreleased changes under `unreleased` from now on and move to specific version after releasing.

PR #407, opened 2018-06-15
 Retries registry call with HTTP if HTTPS fails.

### LINKED ISSUE #388: Add support for registries with just HTTP.

coollog:
Should first merge #430.

Fixes #388

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
@@ -61,7 +61,12 @@ class RegistryEndpointCaller<T> {
     }
   }
 
-  /** Converts the {@link URL}'s protocol to HTTP. */
+  /**
+   * Converts the {@link URL}'s protocol to HTTP.
+   *
+   * @param url the URL to conver to HTTP
+   * @return the URL with protocol set to HTTP
+   */
   private static URL urlWithHttp(URL url) {
     GenericUrl httpUrl = new GenericUrl(url);
     httpUrl.setScheme("http");
@@ -133,7 +138,15 @@ class RegistryEndpointCaller<T> {
     return call(initialRequestState);
   }
 
-  /** Calls the registry endpoint with a certain {@link RequestState}. */
+  /**
+   * Calls the registry endpoint with a certain {@link RequestState}.
+   *
+   * @param requestState the state of the request - determines how to make the request and how to
+   *     process the response
+   * @return an object representing the response, or {@code null}
+   * @throws IOException for most I/O exceptions when making the request
+   * @throws RegistryException for known exceptions when interacting with the registry
+   */
   @Nullable
   private T call(RequestState requestState) throws IOException, RegistryException {
     try (Connection connection = connectionFactory.apply(requestState.url)) {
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -161,14 +205,16 @@ private T call(RequestState requestState) throws IOException, RegistryException
         }
       }
 
+    } catch (HttpHostConnectException | SSLPeerUnverifiedException ex) {
+      // Tries to call with HTTP protocol if HTTPS failed to connect.
+      if (DEFAULT_PROTOCOL.equals(requestState.url.getProtocol())) {
```
TadCordle: I feel like we should explicitly compare against "https" instead of `DEFAULT_PROTOCOL` for readability (and also to avoid problems on the off chance we change the default), or rename the constant. Since we're assuming `DEFAULT_PROTOCOL` is https here.

coollog: Good point, will change to that.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -102,10 +138,18 @@ T call() throws IOException, RegistryException {
...
   }
 
-  /** Calls the registry endpoint with a certain {@link RequestState}. */
+  /**
+   * Calls the registry endpoint with a certain {@link RequestState}.
+   *
+   * @param requestState the state of the request - determines how to make the request and how to
+   *     process the response
+   * @return an object representing the response, or {@code null}
+   * @throws IOException for most I/O exceptions when making the request
```
patflynn: should these be @link?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -161,14 +205,16 @@ private T call(RequestState requestState) throws IOException, RegistryException
         }
       }
 
+    } catch (HttpHostConnectException | SSLPeerUnverifiedException ex) {
```
patflynn: I'm not sure about this, but is it possible users might not want this insecure fallback mechanism?  What if they always want to use https or fail?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
coollog: do you mean like `@throws IOException for most {@link IOException}s when making the request`?

patflynn: nope. Please ignore. I shouldn't make review comments so early in the morning.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
coollog: Hmm, yea it seems that we might want to have the user enable the fallback to HTTP, since Docker only seems to allow insecure registries with a flag.

https://docs.docker.com/registry/insecure/

elharo:
I agree with @patflynn I'm not sure of the background here, but this one is setting off alarm bells in my head. Are you really sure you want to allow non-https at all? Have you consulted the internal security team?

coollog:
Closing in favor of adding a specific configuration parameter to allow insecure `allowInsecureRegistries`. If not set, should not redirect to HTTP.

PR #408, opened 2018-06-15
 Handles 301 and 308 redirects for registry responses.

### LINKED ISSUE #264: Investigate and fix failure to handle 308 Moved permanently for registry responses

coollog:
Fixes #264


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -40,6 +43,12 @@
  */
 class RegistryEndpointCaller<T> {
 
+  /**
+   * @see <a
+   *     href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308">https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308</a>
+   */
+  @VisibleForTesting static final int STATUS_CODE_PERMANENT_REDIRECT = 308;
```
loosebazooka: many of these codes are already defined in org.apache.http.HttpStatus

loosebazooka: oh I see, 308 is missing from these libraries.

PR #411, opened 2018-06-18
Share PATH_ON_IMAGE constants in gradle/maven source file configurations

TadCordle:
Cleaning up another TODO


coollog:
I think we should still keep the getters as interface methods so that when we do jib-core, these paths could be configured to custom paths in other implementations. One way to keep the constants in one place for both `Maven...` and `Gradle...` is to have a separate class that provides the default paths, and this class could prob even be a protected class under `SourceFilesConfiguration`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/SourceFilesConfiguration.java
```diff
@@ -25,6 +25,10 @@
  */
 public interface SourceFilesConfiguration {
 
+  String DEPENDENCIES_PATH_ON_IMAGE = "/app/libs/";
```
coollog: Prefix with `DEFAULT_`?

PR #412, opened 2018-06-18
Add creation date to container config to avoid Docker saying image is 292 years old

### LINKED ISSUE #341: It might be nice if our final image didn't say CREATED in 292 years ago.
We can make each layer reproducible but the final tar can still have a different creation date with the same image-id?

TadCordle:
Now it says 48 years old, but it's better than 292 years (and now it matches distroless).

Fixes #341.


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -54,6 +55,15 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class ContainerConfigurationTemplate implements JsonTemplate {
 
+  /**
+   * A combined date and time at which the image was created. Constant to maintain reproducibility
+   * and avoid Docker's weird "292 years old" bug.
+   *
+   * @see <a
+   *     href="https://github.com/GoogleContainerTools/jib/issues/341">https://github.com/GoogleContainerTools/jib/issues/341</a>
```
loosebazooka: I think it's worth exploring some of the options mentioned in the bazel build bugs also : https://groups.google.com/forum/#!topic/istio-dev/nuv6UaBFGDk (like allowing a user configurable time)? or an env variable. Perhaps file a followup bug for that.

TadCordle: Filed #413

PR #414, opened 2018-06-18
Does not authenticate pull unless the image requires authentication.

### LINKED ISSUE #376: Don't try to authenticate an image unless it requires authentication.
Such as authenticating `gcr.io/distroless/java`.

Currently, authenticating pull happens in the order:

1. RetrieveRegistryCredentialsStep (forBaseImage)
2. AuthenticatePullStep
3. PullBaseImageStep

Rather, it should be in this order:

1. PullBaseImageStep (with no credentials)
If success, skip to step 4.
If not success:
1. RetrieveRegistryCredentialsStep (forBaseImage)
2. PullBaseImageStep (with retrieved credentials)
3. Pass along the credentials to `PullAndCacheBaseImageLayersStep`

Therefore, we should perform `RetrieveRegistryCredentialsStep (forBaseImage)` within `PullBaseImageStep`, since there is no explicit benefit of separating into different asynchronous steps.

Note that our custom `WWW-Authenticate` authenticator (like `AuthenticatePullStep`) is not actually necessary.

coollog:
Fixes #376 

This improves build speed by ~500ms for default distroless base image.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -34,89 +36,127 @@
...
+        // If failed, then, retrieve base registry credentials and try with retrieved credentials.
+        // TODO: Refactor the logic in RetrieveRegistryCredentialsStep out to
+        // registry.credentials.RegistryCredentialsRetriever to avoid this direct executor hack.
+        ListeningExecutorService directExecutorService = MoreExecutors.newDirectExecutorService();
+        RetrieveRegistryCredentialsStep retrieveBaseRegistryCredentialsStep =
+            RetrieveRegistryCredentialsStep.forBaseImage(directExecutorService, buildConfiguration);
+
+        Authorization registryCredentials =
+            NonBlockingSteps.get(retrieveBaseRegistryCredentialsStep);
+        return new Result(pullBaseImage(registryCredentials), registryCredentials);
```
TadCordle: Maybe `pullBaseImage` could just return a `Result` instead of having to wrap the returned image in a new one.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -21,6 +21,8 @@
 import com.google.cloud.tools.jib.async.NonBlockingSteps;
 import com.google.cloud.tools.jib.blob.Blobs;
 import com.google.cloud.tools.jib.builder.BuildConfiguration;
+import com.google.cloud.tools.jib.builder.steps.PullBaseImageStep.Result;
```
TadCordle: Is this import necessary?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
coollog: Usually I think that'd be good, though here it would mean just calling the `new Result` in the returns in `pullBaseImage`, but `pullBaseImage` is intended to be a method that just pulls the base image using the credentials, whereas `Result` is the result of the step as a whole. I'll add a javadoc for this method.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
coollog: Yea, it is actually necessary for the class to use `Result` as the template.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -34,89 +36,138 @@
...
+import javax.annotation.Nullable;
 
 /** Pulls the base image manifest. */
-class PullBaseImageStep implements AsyncStep<Image<Layer>>, Callable<Image<Layer>> {
+class PullBaseImageStep implements AsyncStep<Result>, Callable<Result> {
 
   private static final String DESCRIPTION = "Pulling base image manifest";
 
+  /** Structure for the result returned by this step. */
+  static class Result {
```
loosebazooka: maybe a more descriptive name?

PR #421, opened 2018-06-19
Add 'container' configuration object to wrap container-specific parameters

### LINKED ISSUE #384: Consider grouping container configuration under a `container` configuration object.
If #383 is added.

TadCordle:
`jvmFlags`, `mainClass`, `args`, and `format` configuration parameters are now nested inside a `container` object, much like `image` and `credHelper` being nested inside `to` and `from`. The original parameters are also now deprecated.

Fixes #384.


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -69,7 +69,7 @@ public Blob getContainerConfigurationBlob() {
     // Sets the entrypoint.
     template.setContainerEntrypoint(image.getEntrypoint());
 
-    // Sets the entrypoint.
+    // Sets the main class arguments.
```
coollog: main method?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.List;
+import javax.annotation.Nullable;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Optional;
+
+/**
+ * A bean that configures properties of the container run from the image. This is configurable with
+ * Groovy closures and can be validated when used as a task input.
+ */
+public class ContainerConfiguration {
```
coollog: Maybe `ContainerParameters` so that it's not confused with the container configuration that is part of the image format?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -105,9 +106,9 @@ public void buildImage() throws InvalidImageReferenceException {
             .setTargetImageCredentialHelperName(jibExtension.getTo().getCredHelper())
             .setKnownTargetRegistryCredentials(knownTargetRegistryCredentials)
             .setMainClass(mainClass)
-            .setJavaArguments(jibExtension.getContainer().getArgs())
-            .setJvmFlags(jibExtension.getContainer().getJvmFlags())
-            .setTargetFormat(jibExtension.getContainer().getFormat())
+            .setJavaArguments(jibExtension.getArgs())
```
coollog: This could be left as `jibExtension.getContainer().getArgs()` since `jibExtension.getArgs()` (and the like) will be removed.

TadCordle: I changed `jibExtension.getArgs()` to return `container.getArgs()`, and my plan was to tag `jibExtension.getArgs()` with `@Internal` instead of removing it in the future. Sort of like how `getTargetImage()` and `getBaseImage()` are handled.

coollog: Ah okay that sounds good.

Issue #383, opened 2018-06-06
Consider adding configuration for exposing ports.


coollog:

TadCordle:
I'm thinking the end-goal would be this for gradle:
```
jib {
  container {
    exposedPorts = ['1000', '2000-2010']
  }
}
```

For maven:
```
<configuration>
  <container>
    <exposedPorts>
      <exposedPort>1000</exposedPort>
      <exposedPort>2000-2003</exposedPort>
    </exposedPorts>
  </container>
</configuration>
```

Would be equivalent to adding this to a Dockerfile:
```
EXPOSE 1000
EXPOSE 2000-2003
```

We could name the parameters something differe ("expose" instead of "exposedPorts?")

loosebazooka:
What's the interaction between this being configured on the container and someone exposing ports in their deployment configuration (kube yaml)?

coollog:
I believe exposing the ports on the container would mean that it would not need to be exposed in the pod container definition in the kube yaml.

coollog:
@TadCordle I think calling it `ports` would be fine since I don't think it will have any other connotation than meaning the ports to expose.

PR #432, opened 2018-06-20
Add core changes for configuring exposed ports

TadCordle:
Part of #383 

Still missing Docker context export stuff, but I'll get to that after this


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -193,6 +216,56 @@ public static boolean isValidJavaClass(String className) {
...
+   * @throws NumberFormatException if any of the ports are in an invalid format or out of range
+   */
+  @VisibleForTesting
+  static ImmutableSortedMap<String, EmptyStruct> portListToPortMap(List<String> ports)
+      throws NumberFormatException {
+    SortedMap<String, EmptyStruct> result = new TreeMap<>();
+
+    for (String port : ports) {
+      // Remove all spaces
+      port = port.trim().replace(" ", "");
```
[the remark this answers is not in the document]
TadCordle: What about gradle?

TadCordle:
@loosebazooka I left a comment on the issue for what it might look like: https://github.com/GoogleContainerTools/jib/issues/383

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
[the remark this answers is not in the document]
coollog: In gradle, you would set something to a specific string, so it should be up to the user to make sure they type like `port = '2345'` rather than `port = ' 2345'`, and to error on the latter so they fix their stray space.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -19,19 +19,12 @@
...
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSortedMap;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.*;
```
coollog: Should use explicit imports and no wildcards

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
    * @throws NumberFormatException if any of the ports are in an invalid format or out of range
    */
   @VisibleForTesting
-  static ImmutableSortedMap<String, EmptyStruct> portListToPortMap(List<String> ports)
-      throws NumberFormatException {
-    SortedMap<String, EmptyStruct> result = new TreeMap<>();
+  static ImmutableList<String> expandPortRanges(List<String> ports, BuildLogger logger) {
+    List<String> result = new ArrayList<>();
```
coollog: `ImmutableList.Builder` can avoid doing the `copyOf`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
...
       // Remove all spaces
-      port = port.trim().replace(" ", "");
+      port = port.replace(" ", "");
 
       if (port.matches("\\d+")) {
         // Port is a single number
         int portNum = Integer.parseInt(port);
         if (portNum < 1 || portNum > 65535) {
-          throw new NumberFormatException("Invalid port number " + port);
+          logger.warn("Port number '" + port + "' is out of range (1-65535)");
```
coollog: This is fine for now, but we should add a TODO to give more details on where this port number is from and what the configuration parameter is, and have the configurable message be in `HelpfulSuggestions`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
...
-  static ImmutableSortedMap<String, EmptyStruct> portListToPortMap(List<String> ports)
-      throws NumberFormatException {
-    SortedMap<String, EmptyStruct> result = new TreeMap<>();
+  static ImmutableList<String> expandPortRanges(List<String> ports, BuildLogger logger) {
+    List<String> result = new ArrayList<>();
 
     for (String port : ports) {
       // Remove all spaces
-      port = port.trim().replace(" ", "");
+      port = port.replace(" ", "");
```
coollog: I don't think we should be allowing the user to type like 1_2____4_3 and have it be a valid port.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
    * @throws NumberFormatException if any of the ports are in an invalid format or out of range
    */
   @VisibleForTesting
-  static ImmutableSortedMap<String, EmptyStruct> portListToPortMap(List<String> ports)
-      throws NumberFormatException {
-    SortedMap<String, EmptyStruct> result = new TreeMap<>();
+  static ImmutableList<String> expandPortRanges(List<String> ports, BuildLogger logger) {
```
coollog: It looks like this is only used in the `Builder` class, so it should prob live in there. Making it non-static would remove the need to have `BuildLogger` as the second param.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
...
         // Make sure range is valid (min-max, within port range)
-        if (min > max || min < 1 || max > 65535) {
-          throw new NumberFormatException("Invalid port range " + port);
+        if (min < 1 || max > 65535) {
+          logger.warn("Port range '" + port + "' exceeds normal port range (1-65535)");
         }
 
         for (int portNum = min; portNum <= max; portNum++) {
-          result.put(portNum + "/tcp", EmptyStruct.get());
+          result.add("" + portNum);
```
coollog: If we are just supporting integer ports, we should be building an `ImmutableList<Integer>`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
...
+        if (min > max) {
+          int swapTemp = max;
+          max = min;
+          min = swapTemp;
+        }
+
         // Make sure range is valid (min-max, within port range)
-        if (min > max || min < 1 || max > 65535) {
-          throw new NumberFormatException("Invalid port range " + port);
+        if (min < 1 || max > 65535) {
```
coollog: This check should be combined with the one above for the single port case.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
...
+        result.add(port);
 
       } else if (port.matches("\\d+-\\d+")) {
         // Port is a range
         List<String> range = Splitter.on('-').splitToList(port);
         int min = Integer.parseInt(range.get(0));
         int max = Integer.parseInt(range.get(1));
 
+        // Reverse range if configured as 'max-min' instead of 'min-max'
+        if (min > max) {
```
coollog: I feel like we should prob error in this case instead of fixing it secretly for them, since it might be more likely they typed the wrong port numbers than wrote them going from high to low.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -226,44 +210,50 @@ public static boolean isValidJavaClass(String className) {
...
 
         for (int portNum = min; portNum <= max; portNum++) {
-          result.put(portNum + "/tcp", EmptyStruct.get());
+          result.add("" + portNum);
         }
 
       } else {
         // Port is neither a single number nor a range
-        throw new NumberFormatException("Invalid port number " + port);
+        logger.warn("Port '" + port + "' is not a port number");
```
coollog: This should probably throw an error instead since the user wrote an invalid configuration.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -100,8 +98,8 @@
      *     appear in the configuration json (e.g. "portNum/tcp")
      * @return this
      */
-    public Builder<T> setExposedPorts(Map<String, EmptyStruct> exposedPorts) {
-      this.exposedPorts = ImmutableSortedMap.copyOf(exposedPorts);
+    public Builder<T> setExposedPorts(List<String> exposedPorts) {
```
coollog: I think we can actually change these to `ImmutableList` to avoid doing a `copyOf` on another `ImmutableList`, or just leave a TODO for now.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -123,7 +121,7 @@
           environmentBuilder.build(),
           ImmutableList.copyOf(entrypoint),
           ImmutableList.copyOf(javaArguments),
-          ImmutableSortedMap.copyOf(exposedPorts));
+          ImmutableList.copyOf(exposedPorts));
```
coollog: These `copyOf`s are not necessary anymore.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -124,8 +123,12 @@ public void setContainerCmd(List<String> cmd) {
     config.Cmd = cmd;
   }
 
-  public void setContainerExposedPorts(SortedMap<String, EmptyStruct> exposedPorts) {
-    config.ExposedPorts = exposedPorts;
+  public void setContainerExposedPorts(List<String> exposedPorts) {
+    SortedMap<String, Map<?, ?>> result = new TreeMap<>();
```
coollog: `ImmutableSortedMap.Builder`

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -152,8 +155,16 @@ public void addLayerDiffId(DescriptorDigest diffId) {
...
-  SortedMap<String, EmptyStruct> getContainerExposedPorts() {
-    return config.ExposedPorts;
+  List<String> getContainerExposedPorts() {
+    if (config.ExposedPorts == null) {
+      return null;
+    }
+    List<String> ports = new ArrayList<>();
+    for (Map.Entry<String, Map<?, ?>> entry : config.ExposedPorts.entrySet()) {
+      // Remove the "/tcp"
+      ports.add(Splitter.on('/').splitToList(entry.getKey()).get(0));
```
coollog: This might break if the entry was not in the form `\d+/tcp`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -124,8 +123,12 @@ public void setContainerCmd(List<String> cmd) {
     config.Cmd = cmd;
   }
 
-  public void setContainerExposedPorts(SortedMap<String, EmptyStruct> exposedPorts) {
-    config.ExposedPorts = exposedPorts;
+  public void setContainerExposedPorts(List<String> exposedPorts) {
+    SortedMap<String, Map<?, ?>> result = new TreeMap<>();
+    for (String port : exposedPorts) {
+      result.put(port + "/tcp", Collections.emptyMap());
```
coollog: I believe we can just do port without `/tcp` since by default it is tcp.

https://github.com/opencontainers/image-spec/blob/master/config.md#properties

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
TadCordle: Are we only supporting integer ports [or not](https://github.com/GoogleContainerTools/jib/pull/432#discussion_r196984657)? If we're restricting to numbers I'll change it to be type specific, if not then I forgot to add a `result.add(port)` after the last warning in the method.

TadCordle: Nevermind

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -20,8 +20,8 @@
 import com.google.cloud.tools.jib.image.DescriptorDigest;
 import com.google.cloud.tools.jib.json.JsonTemplate;
 import com.google.common.annotations.VisibleForTesting;
-import java.util.ArrayList;
-import java.util.List;
+import com.google.common.collect.ImmutableSortedMap;
+import java.util.*;
```
TadCordle: I don't know why IntelliJ keeps doing this

coollog: I think there's a configuration you have to disable.

coollog: https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -178,6 +188,69 @@ public BuildConfiguration build() {
...
+                  + "with a '-', with or without protocol specified (e.g. '<portNum>/tcp' or "
+                  + "'<portNum>/udp').");
+        }
+
+        // Parse protocol
+        List<String> splitProtocol = Splitter.on("/").splitToList(port);
+        String protocol = splitProtocol.size() > 1 ? "/" + splitProtocol.get(1) : "";
+
+        // Parse range (or treat as range of min-min if only single port configuration)
+        List<String> splitRange = Splitter.on("-").splitToList(splitProtocol.get(0));
```
coollog: Since you are using pattern match above already, it might make sense to just use the regex matcher to get the parts, so the regex would be like:

```
(\d+)(-\d+|)(/tcp|/udp|)
```

So that it would match 4 groups, where 1 is min, 2 is max or nothing, and 3 is protocol or nothing.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -178,6 +188,69 @@ public BuildConfiguration build() {
...
+
+        // Warn for possibly invalid port numbers
+        if (min < 1 || max > 65535) {
+          // TODO: Add details/use HelpfulSuggestions for these warnings
+          buildLogger.warn("Port number '" + port + "' is out of usual range (1-65535).");
+        }
+
+        // Add all numbers in range to list
+        for (int portNum = min; portNum <= max; portNum++) {
+          result.add(portNum + protocol);
```
coollog: Since we are passing this through the code, we might want to consider putting this in a canonical form rather than the docker image format for of `port/protocol`. Like a class that has a port and a protocol. Can also just be in a TODO for now.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -201,6 +203,7 @@ public BuildConfiguration build() {
     @VisibleForTesting
     ImmutableList<String> expandPortRanges(List<String> ports) throws NumberFormatException {
       ImmutableList.Builder<String> result = new ImmutableList.Builder<>();
+      Pattern portPattern = Pattern.compile("(\\d+)(-\\d+|)(\\/tcp|\\/udp|)");
```
coollog: This can be a constant so that it doesnt have to be compiled each time.

coollog: I just tested it more and realized you can just do 

```
(\d+)(?:-(\d+))?(/tcp|/udp)?
```

And it'll match those groups, with them being null if it didn't match.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -220,16 +225,13 @@ public BuildConfiguration build() {
...
-        String protocol = splitProtocol.size() > 1 ? "/" + splitProtocol.get(1) : "";
-
-        // Parse range (or treat as range of min-min if only single port configuration)
-        List<String> splitRange = Splitter.on("-").splitToList(splitProtocol.get(0));
-        int min = Integer.parseInt(splitRange.get(0));
+        int min = Integer.parseInt(matcher.group(1));
         int max = min;
-        if (splitRange.size() > 1) {
-          max = Integer.parseInt(splitRange.get(1));
+        if (!matcher.group(2).equals("")) {
```
coollog: `Strings.isNullOrEmpty`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
TadCordle: I don't see any advantage to doing this, just seems harder to understand and getting null instead of an empty string just forces us to add more checks.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -178,6 +190,70 @@ public BuildConfiguration build() {
...
+     * Converts/validates a list of ports with ranges to an expanded form without ranges.
+     *
+     * <p>Example: ["1000/tcp", "2000-2002/tcp"] -> ["1000/tcp", "2000/tcp", "2001/tcp", "2002/tcp"]
+     *
+     * @param ports the list of port numbers/ranges
+     * @return the ports as a list of integers
+     * @throws NumberFormatException if any of the ports are in an invalid format or out of range
+     */
+    @VisibleForTesting
+    ImmutableList<String> expandPortRanges(List<String> ports) throws NumberFormatException {
```
coollog: Let's add a TODO for moving this to a class in frontend, so that BuildConfiguration can be kept free of user input parsing.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -118,6 +127,15 @@ public void setContainerCmd(List<String> cmd) {
     config.Cmd = cmd;
   }
 
+  public void setContainerExposedPorts(List<String> exposedPorts) {
```
coollog: The template itself should probably be free of reproducibility logic. This can just be a TODO for now too.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -88,6 +94,9 @@
 
     /** Arguments to pass into main. */
     @Nullable private List<String> Cmd;
+
+    /** Network ports the container exposes. */
+    @Nullable private SortedMap<String, Map<?, ?>> ExposedPorts;
```
coollog: It might make sense to have this be a custom object that stores a list. That way, there doesn't have to be any logic in `getContainerExposedPorts`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -38,6 +39,9 @@
 
   public static class Builder {
 
+    /** Pattern used for parsing information out of exposed port configurations. */
+    private static Pattern portPattern = Pattern.compile("(\\d+)(?:-(\\d+))?(/tcp|/udp)?");
```
coollog: final

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -38,6 +39,9 @@
 
   public static class Builder {
 
+    /** Pattern used for parsing information out of exposed port configurations. */
```
coollog: The comment for examples on line 209-215 should go here

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -227,9 +230,9 @@ public BuildConfiguration build() {
         // Parse protocol
         int min = Integer.parseInt(matcher.group(1));
         int max = min;
-        if (!matcher.group(2).equals("")) {
+        if (!Strings.isNullOrEmpty(matcher.group(2))) {
           // Skip over the hyphen
```
coollog: this comment is not applicable anymore

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -178,6 +198,63 @@ public BuildConfiguration build() {
           throw new IllegalStateException(errorMessage.toString());
       }
     }
+
+    /**
+     * TODO: Move this to a class in frontend
```
coollog: Let's file a bug to take care of these TODOs

TadCordle: https://github.com/GoogleContainerTools/jib/issues/442

PR #437, opened 2018-06-21
Allows only HTTPS unless 'allowHttp' is true.

coollog:
Redo of #407 and regards fixing #388 .

Next PR is adding the plugin configuration `allowInsecureRegistries`.


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -100,15 +100,23 @@ static String getUserAgent() {
...
    * @param imageName the image/repository name (also known as, namespace)
+   * @param allowHttp if {@code true}, allows redirects and fallbacks to HTTP; otherwise, only
+   *     allows HTTPS
    */
-  public RegistryClient(@Nullable Authorization authorization, String serverUrl, String imageName) {
+  public RegistryClient(
+      @Nullable Authorization authorization,
+      String serverUrl,
+      String imageName,
+      boolean allowHttp) {
```
loosebazooka: I think there's something to readability when using builders or static factories here. Having to look at what a "true/false" parameter means when reading the code is not the most readable experience.

coollog: Added static factories.

coollog:
Changed to be a `Factory` class - this will help us to refactor the UserAgent configuration into this Factory class rather than setting it staticly on RegistryClient

PR #438, opened 2018-06-22
Add core 'ports' functionality to docker context exporter, fix cross-platform Dockerfile write inconsistency

TadCordle:
After merging #432. Part of #383.


jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -193,4 +206,19 @@ static String joinAsJsonArray(List<String> items) {
 
     return resultString.toString();
   }
+
+  /**
+   * Builds a list of Dockerfile "EXPOSE" items.
```
coollog: In Dockerfile terminology, they are called "instructions".

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -193,4 +206,19 @@ static String joinAsJsonArray(List<String> items) {
...
+   *
+   * @param exposedPorts the list of ports numbers/ranges to expose
+   * @return a string containing an EXPOSE command for each of the entries
+   */
+  @VisibleForTesting
+  static String makeExposeItems(List<String> exposedPorts) {
+    StringBuilder resultString = new StringBuilder();
+    for (String port : exposedPorts) {
+      resultString.append("EXPOSE ").append(port).append("\n");
```
coollog: nit: can be '\n'

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -193,4 +206,19 @@ static String joinAsJsonArray(List<String> items) {
...
   }
+
+  /**
+   * Builds a list of Dockerfile "EXPOSE" items.
+   *
+   * @param exposedPorts the list of ports numbers/ranges to expose
+   * @return a string containing an EXPOSE command for each of the entries
+   */
+  @VisibleForTesting
+  static String makeExposeItems(List<String> exposedPorts) {
```
coollog: Looks like I messed up the ordering for `joinAsJsonArray` as well, but these static methods should be moved up top.

jib-core/src/main/resources/DockerfileTemplate
```diff
@@ -4,5 +4,5 @@ COPY libs @@DEPENDENCIES_PATH_ON_IMAGE@@
 COPY resources @@RESOURCES_PATH_ON_IMAGE@@
 COPY classes @@CLASSES_PATH_ON_IMAGE@@
 
-ENTRYPOINT @@ENTRYPOINT@@
+@@EXPOSED_PORTS@@ENTRYPOINT @@ENTRYPOINT@@
```
coollog: Should leave some new lines in between

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
coollog: Also, if you don't want the extra new line at the end of the list, you can do a join instead.

jib-core/src/test/resources/sampleDockerfile
```diff
@@ -4,5 +4,7 @@ COPY libs /app/libs/
 COPY resources /app/resources/
 COPY classes /app/classes/
 
+EXPOSE 1000
+EXPOSE 2000-2010
```
coollog: Is this a valid instruction? If not, we might want to reconsider supporting this type of syntax in our configuration, and try to keep it as similar to Dockerfile EXPOSE as possible.

TadCordle: [Yes](https://stackoverflow.com/a/28721643)

coollog: Cools

coollog: What do you think about supporting UDP?

TadCordle: That sounds like a useful thing and probably wouldn't be too hard to add.

jib-core/src/main/resources/DockerfileTemplate
TadCordle: I did that so there are no stray newlines when it gets replaced.

coollog: Yea, the stray newlines should be an implementation fix rather than on the template itself or it decreases the readability of the template.

coollog: Also, a nit, but perhaps EXPOSE_INSTRUCTIONS is a more applicable naming of the template section?

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -83,8 +83,13 @@ static String joinAsJsonArray(List<String> items) {
   @VisibleForTesting
   static String makeExposeItems(List<String> exposedPorts) {
     StringBuilder resultString = new StringBuilder();
+    boolean first = true;
```
coollog: Doing a join would avoid this `first` logic:

```java
String.join("\n", exposedPorts.stream().map(port -> "EXPOSE " + port).collect(Collectors.toList()));
```

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -46,12 +46,61 @@
...
+   *
+   * @see <a
+   *     href="https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example">https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example</a>
+   * @param items the list of items to join into an array.
+   * @return a string in the format: ["item1","item2",...]
+   */
+  @VisibleForTesting
+  static String joinAsJsonArray(List<String> items) {
+    StringBuilder resultString = new StringBuilder("[");
+    boolean firstComponent = true;
```
coollog: Doing a join would avoid this `first` logic:

```java
String.join("\n", exposedPorts.stream().map(port -> "EXPOSE " + port).collect(Collectors.toList()));
```

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -84,7 +84,8 @@ static String joinAsJsonArray(List<String> items) {
   @VisibleForTesting
   static String makeExposeItems(List<String> exposedPorts) {
     return String.join(
-        "\n", exposedPorts.stream().map(port -> "EXPOSE " + port).collect(Collectors.toList()));
+        System.getProperty("line.separator"),
```
coollog: That's interesting. I'm wondering why this would need this platform-specific separator when the template's lines are separated by just LF?

Also, was the error with Docker not accepting non CRLF on windows?

TadCordle: I didn't look that deeply into it, the error just said that the sample and the generated docker file lengths differed by 1 on windows, so I assumed it was using "\r\n" as separators and was missing a "\r".

coollog: Hmm that might mean the test should be fixed. I think the docker context generated should be platform independent so that a user generating on linux or windows would produce the same context for the same project.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -46,12 +47,54 @@
...
+  }
+
+  /**
+   * Builds a list of Dockerfile "EXPOSE" instructions.
+   *
+   * @param exposedPorts the list of ports numbers/ranges to expose
+   * @return a string containing an EXPOSE instruction for each of the entries
+   */
+  @VisibleForTesting
+  static String makeExposeItems(List<String> exposedPorts) {
```
coollog: `makeExposeInstructions`?

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -47,6 +46,18 @@
  */
 public class DockerContextGenerator {
 
+  /** The template to generate the Dockerfile from. */
+  private static final String DOCKERFILE_TEMPLATE =
```
coollog: If we're changing to just building the string in the code, we can just use stringbuilder and append the parts rather than doing the template replacement.

TadCordle:
Yeah I'm just trying things right now.

TadCordle:
@coollog @loosebazooka I think this is review-able, finally.

I got it working with the DockerfileTemplate in [this commit](https://github.com/GoogleContainerTools/jib/pull/438/commits/213d32ce2fd6ed64c930b480cfc4cc0d821520cc), but ended up removing it anyway for efficiency.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -46,12 +46,45 @@
...
+   * Formats a list for the Dockerfile's ENTRYPOINT or CMD.
+   *
+   * @see <a
+   *     href="https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example">https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example</a>
+   * @param items the list of items to join into an array.
+   * @return a string in the format: ["item1","item2",...]
+   */
+  @VisibleForTesting
+  static String joinAsJsonArray(List<String> items) {
+    return "["
```
coollog: Oh for this we could actually use `ObjectMapper` from the JSON library we use (I forgot about it when implementing this). It will automatically generate the JSON string from a list of strings.

You could do like

`Blobs.writeToString(JsonTemplateMapper.toBlob(thelist))`

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -141,56 +185,26 @@ public void generate(Path targetDirectory) throws IOException {
...
-        .replace(
-            "@@ENTRYPOINT@@",
-            joinAsJsonArray(
-                EntrypointBuilder.makeEntrypoint(sourceFilesConfiguration, jvmFlags, mainClass)))
-        .replace("@@CMD@@", joinAsJsonArray(javaArguments));
-  }
-
-  /**
-   * Formats a list for the Dockerfile's ENTRYPOINT or CMD.
+   * Makes the contents of a {@code Dockerfile} using configuration data.
```
coollog: Since we're putting this in code, would be good to have in the javadoc an example of what it would look like.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
TadCordle: I think it needs a `ListOfJsonTemplate`, not a list of strings?

coollog: Oh right, maybe just use `new ObjectMapper().writeValueAsString` directly.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -141,56 +166,40 @@ public void generate(Path targetDirectory) throws IOException {
...
+    ObjectMapper objectMapper = new ObjectMapper();
+    return "FROM "
+        + Preconditions.checkNotNull(baseImage)
+        + "\n\nCOPY libs "
+        + sourceFilesConfiguration.getDependenciesPathOnImage()
+        + "\nCOPY resources "
+        + sourceFilesConfiguration.getResourcesPathOnImage()
+        + "\nCOPY classes "
+        + sourceFilesConfiguration.getClassesPathOnImage()
+        + "\n\n"
```
coollog: Oh wait, wouldn't this add 3 newlines if there weren't any exposed ports?

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -46,12 +47,25 @@
...
 public class DockerContextGenerator {
 
+  /**
+   * Builds a list of Dockerfile "EXPOSE" instructions.
+   *
+   * @param exposedPorts the list of ports numbers/ranges to expose
+   * @return a string containing an EXPOSE instruction for each of the entries
+   */
+  @VisibleForTesting
+  static String makeExposeInstructions(List<String> exposedPorts) {
```
coollog: This could also be moved to `makeDockerfile` and change that to append to a `StringBuilder` as an alternative.

Issue #443, opened 2018-06-22
Add option to disable caching.


coollog:
Similar to how Docker's `--no-cache`, as suggested by @saturnism.

briandealwis:
Docker seems to provide `--no-cache` because of statements like
```
RUN apt-get -y update
```

which may need to be forced to run.  Jib doesn't execute code dynamically, and our caches are based on actual local files.  So I don't think this is necessary.  Please re-open if necessary.

PR #445, opened 2018-06-25
Exposes configuration for configuring cache locations.

coollog:
This helps with fixing #443, but mainly helps with making jib-core a usable library. This gives the ability to configure where the caches are in `BuildConfiguration`.


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/BuildStepsRunner.java
```diff
@@ -44,62 +44,57 @@
...
+  private static Initializer getCacheInitializer(BuildConfiguration buildConfiguration)
       throws CacheDirectoryCreationException {
-    if (!Files.exists(cacheDirectory)) {
-      try {
-        Files.createDirectory(cacheDirectory);
-
-      } catch (IOException ex) {
-        throw new CacheDirectoryCreationException(cacheDirectory, ex);
-      }
+    CacheConfiguration applicationLayersCacheConfiguration;
```
TadCordle: Maybe use a ternary operator here so you can declare and initialize on the same line.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/BuildStepsRunner.java
```diff
@@ -44,62 +44,57 @@
...
+      applicationLayersCacheConfiguration = CacheConfiguration.makeTemporary();
+    } else {
+      applicationLayersCacheConfiguration =
+          buildConfiguration.getApplicationLayersCacheConfiguration();
     }
-    Caches.Initializer cachesInitializer = Caches.newInitializer(cacheDirectory);
-    if (useOnlyProjectCache) {
-      cachesInitializer.setBaseCacheDirectory(cacheDirectory);
+
+    CacheConfiguration baseImageLayersCacheConfiguration;
```
TadCordle: same here

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
```diff
@@ -64,43 +54,59 @@ static void ensureOwnership(Path cacheDirectory)
...
     }
 
-    private final Path applicationCacheDirectory;
-    private Path baseCacheDirectory = DEFAULT_BASE_CACHE_DIRECTORY;
-
-    private Initializer(Path applicationCacheDirectory) {
-      this.applicationCacheDirectory = applicationCacheDirectory;
-    }
+    private final Path baseImageLayersCacheDirectory;
+    private final boolean shouldEnsureOwnershipOfBaseImageLayersCacheDirectory;
```
TadCordle: Are these booleans only used for testing purposes, or do they have other practical uses?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheDirectoryCreationException.java
```diff
@@ -14,15 +14,23 @@
...
 
-package com.google.cloud.tools.jib.frontend;
+package com.google.cloud.tools.jib.cache;
 
-import com.google.cloud.tools.jib.cache.Cache;
 import java.nio.file.Path;
 
-/** Thrown when a directory could not be created to be used as the {@link Cache}. */
+/**
+ * Thrown when a directory could not be created to be used as the {@link
```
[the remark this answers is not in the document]
coollog: I think the meaning is more in terms of failure to create a directory for use as a cache directory rather than failure to create a cache directory (which may imply a special type of directory).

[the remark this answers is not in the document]
TadCordle: Hm I see. Either way the sentence is a bit hard to read

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Caches.java
coollog: These are used in the `init` method below, which would only check for ownership by Jib if set to true.

TadCordle: I was wondering if there was any reason to not check for Jib ownership, but I guess it's probably useful for library purposes

coollog: When using a temporary directory, there's no need to check. Also, it would fail since the temporary directory has already been created.

patflynn:
I'd love to hear more about this change in our design review.  I'd
recommend being conservative when it comes to adding features and
configuration to the build plugins so as not to overwhelm users with
configuration that has low to zero value for them.  It reduces overall
usability.

coollog:
This is actually just exposing the configuration to users of `jib-core`, so in `jib-gradle-plugin`, `jib-maven-plugin` or future jib-library users.

Issue #442, opened 2018-06-22
Move expose ports logic to more appropriate locations


TadCordle:
- [x] Move `BuildConfiguration#expandPortRanges` to a frontend class
- [x] Create class for holding port + protocol
- [x] Use proprietary class for port/protocol storage rather than strings
- [x] Move list <-> map conversion out of `ContainerConfigurationTemplate`

PR #449, opened 2018-06-26
Add ExposedPort class with port/protocol fields, move exposed port config-related parsing to frontend class

TadCordle:
Part of #442. This will be passed around instead of the string representation for easier access.


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ExposedPort.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+package com.google.cloud.tools.jib.configuration;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Splitter;
+import java.util.List;
+
+/** Holds port number and protocol for an exposed port. */
+public class ExposedPort {
```
coollog: Since this is just a structure, `expose` may be out of context. Probably like `PortWithProtocol` is more applicable.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ExposedPort.java
```diff
@@ -0,0 +1,69 @@
...
+
+  private int port;
+  private Protocol protocol;
+
+  /**
+   * @param configurationString the string to parse, in the format "port[/protocol]". Defaults to
+   *     TCP if protocol is not specified.
+   * @return an {@link ExposedPort} with values parsed from the string
+   */
+  public static ExposedPort fromString(String configurationString) {
```
coollog: I think this should belong in a frontend Parser class since this is parsing a user input

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ExposedPort.java
```diff
@@ -0,0 +1,69 @@
...
+  enum Protocol {
+    TCP,
+    UDP
+  }
+
+  private int port;
+  private Protocol protocol;
+
+  /**
+   * @param configurationString the string to parse, in the format "port[/protocol]". Defaults to
```
coollog: would be good to include acceptable examples

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ExposedPort.java
```diff
@@ -0,0 +1,69 @@
...
+   *     TCP if protocol is not specified.
+   * @return an {@link ExposedPort} with values parsed from the string
+   */
+  public static ExposedPort fromString(String configurationString) {
+    List<String> parts = Splitter.on('/').splitToList(configurationString);
```
coollog: What if the input was like: "123/tcp/something" or "notanumber"

TadCordle: This assumes the validation is already done somewhere else (i.e. `BuildConfiguration`), but that stuff is getting refactored out eventually. So I'm not sure yet.

coollog: Yea validation and parsing should definitely be together

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPorts.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    return result.build();
+  }
+
+  /**
+   * @param exposedPorts a list of exposed ports
+   * @return a map of the ports in container config json form ({@code "port/protocol":{}})
+   */
+  public static ImmutableSortedMap<String, Map<?, ?>> listToMap(List<String> exposedPorts) {
```
TadCordle: Just realized these two functions don't belong in frontend, I'm gonna revert these and move them in a different PR.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -200,7 +190,7 @@ public BuildConfiguration build() {
               ImmutableList.copyOf(javaArguments),
               ImmutableList.copyOf(jvmFlags),
               ImmutableMap.copyOf(environmentMap),
-              expandPortRanges(exposedPorts),
+              ExposedPorts.expandPortRanges(exposedPorts, buildLogger),
```
coollog: We should try to keep `BuildConfiguration` free of user input parsing. `BuildConfiguration` should really be passed in a list of `PortWithProtocol` objects.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPorts.java
```diff
@@ -0,0 +1,95 @@
...
+import com.google.cloud.tools.jib.builder.BuildLogger;
+import com.google.cloud.tools.jib.configuration.PortWithProtocol;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ExposedPorts {
```
coollog: If this is just a static parser, should probably name like `ExposedPortsParser` and have a private constructor.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
coollog: Also, in general, we should not have other parts of jib-core depend on frontend classes.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPorts.java
coollog: short javadoc

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortWithProtocol.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  public Protocol getProtocol() {
+    return protocol;
+  }
+
+  /** @return the {@link PortWithProtocol} in string form "port/protocol" */
+  @Override
+  public String toString() {
+    return port + "/" + protocol.toString().toLowerCase();
```
coollog: Might be better to define the enum with the string form so that the enum name is decoupled from the string generation. But, it seems this string conversion should be handled in other places like when this is generated into the Dockerfile/container config and such.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortWithProtocol.java
```diff
@@ -0,0 +1,49 @@
...
+/** Holds port number and protocol for an exposed port. */
+public class PortWithProtocol {
+
+  /** Represents the protocol portion of the port. */
+  public enum Protocol {
+    tcp,
+    udp
+  }
+
+  private int port;
```
coollog: final

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortWithProtocol.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+package com.google.cloud.tools.jib.configuration;
+
+/** Holds port number and protocol for an exposed port. */
+public class PortWithProtocol {
+
+  /** Represents the protocol portion of the port. */
+  public enum Protocol {
+    tcp,
```
coollog: Oh enums should definitely be uppercase since they are constants. I meant that they should be instantiated with the string form, so like `TCP("tcp")`.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortWithProtocol.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public int getPort() {
+    return port;
+  }
+
+  public Protocol getProtocol() {
+    return protocol;
+  }
+
+  /** @return the {@link PortWithProtocol} in string form "port/protocol" */
```
coollog: maybe a TODO to move this to output generation since this is a Docker-specific string format

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPortsParser.java
```diff
@@ -0,0 +1,96 @@
...
+import com.google.cloud.tools.jib.builder.BuildLogger;
+import com.google.cloud.tools.jib.configuration.PortWithProtocol;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class ExposedPortsParser {
```
coollog: private constructor

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortWithProtocol.java
TadCordle: Eh yeah, I'll just remove it and handle it later.

PR #454, opened 2018-06-27
Log important messages at lifecycle level

### LINKED ISSUE #434: Logging at debug level should probably be at lifecycle level
When running exportDockerContext:

```
[DEBUG] Searching for main class... Add a 'mainClass' configuration to 'jib-maven-plugin' to improve build speed.
[DEBUG] Created Docker context at /google/src/cloud/qingyangc/sandbox/google3/experimental/users/qingyangc/sandbox/helloproject/target/jib-docker-context
```

These log messages should probably be at lifecycle level or the goal doesn't actually show anything

TadCordle:
Fixes #434

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
@@ -61,7 +61,7 @@ public class MainClassFinder {
       throws MainClassInferenceException {
     BuildLogger logger = projectProperties.getLogger();
     if (mainClass == null) {
-      logger.lifecycle(
+      logger.info(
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -61,7 +61,7 @@ public static String resolveMainClass(
       throws MainClassInferenceException {
     BuildLogger logger = projectProperties.getLogger();
     if (mainClass == null) {
-      logger.info(
+      logger.lifecycle(
```
coollog: Let's actually hold off on this line for now, since this may be confusing for many first-time users.

Issue #457, opened 2018-06-28
Micronaut


Solido:
**Description of the issue**:

Micronaut service doesn't start on Google Cloud when deployed as Container in Compute Engine.
Container stall in Restarting mode.

http://micronaut.io is a microservice framework by OCI the team behing Grails / Groovy.

**Expected behavior**:

Service should start instantly

**Steps to reproduce**: 

Create a new Micronaut app using the Cli
Add the JIB gradle plugin
Deploy to Container Repo and start a new instance using the Jib deployed container

**Environment**: 

GCE

**Additional Information**: I guess that the generated ENTRYPOINT can't access required 
classes/libs 

Thank you

patflynn:
could this be related to #396 ?

patflynn:
@Solido do you have a sample we could use for testing? Are you using Java, Kotlin, or Groovy?

Solido:
Not sure, Micronaut is using java as the main starter but the rest of the app is impl w/ groovy.

The shortest path to recreate is 

- installing micronaut using SDKMan : https://sdkman.io/ (easier than build from github)
- create a app using the command `mn create-app hello-world --features groovy`
- add JIB to Gradle

The build and the upload of the image is OK

Let me know if it's too complicated I may try to send you a blank project, just not sure about the size of the archive it will generate.

patflynn:
FYI I just tried this without groovy and it worked perfectly.  This may be a generic Groovy related issues similar to #396

Solido:
Good news that java is working, 

#396 can not even compile and he gets log about it.
Looks like all is fine and jib-docker-context contains all the compiled classes and required libs. 
Here's my log from gradle jib with groovy

Retrieving registry credentials for gcr.io...
Retrieving registry credentials for gcr.io...
Building dependencies layer...
Building resources layer...
Building classes layer...
Getting base image gcr.io/distroless/java...
Finalizing...

Container entrypoint set to [java, -Xmx50m, -cp, /app/libs/*:/app/resources/:/app/classes/, awesome.server.Application]

Any idea where to look ?

patflynn:
I just tried running jib with the groovy hello  world and got:
`
Execution failed for task ':jib'.
> Obtaining project build output files failed; make sure you have compiled your project before trying to build the image. (Did you accidentally run "gradle clean jib" instead of "gradle clean compileJava jib"?)
`

Solido:
The app is working well when using a shadow jar and using a basic 
[...]

PR #462, opened 2018-06-28
 Changes GradleSourceFilesConfiguration to warn instead of error when classes not found.

coollog:
This regards #457. It might be enough to fix the issue, or at least help with fixing the issue.

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -14,11 +14,14 @@ All notable changes to this project will be documented in this file.
 - Container creation date set to timestamp 0 ([#341](https://github.com/GoogleContainerTools/jib/issues/341))
 - Does not authenticate base image pull unless necessary - reduces build time by about 500ms ([#414](https://github.com/GoogleContainerTools/jib/pull/414))
 - `jvmFlags`, `mainClass`, `args`, and `format` are now grouped under `container` configuration object ([#384](https://github.com/GoogleContainerTools/jib/issues/384))
+- Warns instead of errors when classes not found ([#462](https://github.com/GoogleContainerTools/jib/pull/462))
 
 ### Fixed 
 
 - Using Azure Container Registry now works - define credentials in `jib.to.auth`/`jib.from.auth` ([#415](https://github.com/GoogleContainerTools/jib/issues/415))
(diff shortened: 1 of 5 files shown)


jib-gradle-plugin/CHANGELOG.md
```diff
@@ -14,11 +14,14 @@ All notable changes to this project will be documented in this file.
 - Container creation date set to timestamp 0 ([#341](https://github.com/GoogleContainerTools/jib/issues/341))
 - Does not authenticate base image pull unless necessary - reduces build time by about 500ms ([#414](https://github.com/GoogleContainerTools/jib/pull/414))
 - `jvmFlags`, `mainClass`, `args`, and `format` are now grouped under `container` configuration object ([#384](https://github.com/GoogleContainerTools/jib/issues/384))
+- Warns instead of errors when classes not found ([#462](https://github.com/GoogleContainerTools/jib/pull/462))
```
TadCordle: Do we need to do this in the maven plugin as well?

coollog: Nah, since jib-maven-plugin does not automatically depend on `compile`. Also, it just looks at build output directories, so it's not tied to some `java` plugin like Gradle.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -44,13 +44,12 @@ static GradleProjectProperties getForProject(
...
+          gradleBuildLogger,
+          GradleSourceFilesConfiguration.getForProject(project, gradleBuildLogger));
+
     } catch (IOException ex) {
-      throw new GradleException(
-          "Obtaining project build output files failed; make sure you have compiled your project "
-              + "before trying to build the image. (Did you accidentally run \"gradle clean "
-              + "jib\" instead of \"gradle clean compileJava jib\"?)",
-          ex);
+      throw new GradleException("Obtaining project build output files failed", ex);
```
TadCordle: It may still be helpful to have a possible reason for this happening in the error message, if it's still likely that this could even happen.

coollog: That will actually just be an `IOException` that happens when trying to list directories, so it's not a likely error anymore.

Issue #415, opened 2018-06-18
Fix using docker-credential-acr-*


coollog:
Thumbs up if you are using Azure.

coollog:
Passing credentials via Maven settings for Azure Container Registry works after #423 , but docker-credential-acr-* still does not work.

coollog:
Filed issue against docker-credential-acr-*: https://github.com/Azure/acr-docker-credential-helper/issues/31

coollog:
Closing unless `docker-credential-acr-*` is needed

PR #466, opened 2018-06-28
Removes Azure container registry not supported limitation.

coollog:
Fixed in `v0.9.1` (#415)

To be merged after documentation is updated to `v0.9.1`


TadCordle:
Why not just push this to the readme branch itself?

jib-gradle-plugin/README.md
```diff
@@ -274,12 +274,6 @@ jib {
 
 See the [Jib project README](/../../#how-jib-works).
 
-## Known Limitations
-
-These limitations will be fixed in later releases.
-
-* Pushing to Azure Container Registry is not currently supported.
```
TadCordle: Should we list docker-credential-acr-* not working?

coollog:
I think that’s a minor-enough issue that we can just track it as an issue
rather than stating it on the documentation.

PR #470, opened 2018-06-29
Make sure items aren't null in BuildConfiguration.Builder collections

TadCordle:


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -98,28 +100,32 @@ public Builder setMainClass(@Nullable String mainClass) {
...
+        Preconditions.checkArgument(Iterables.all(javaArguments, Objects::nonNull));
+        this.javaArguments = ImmutableList.copyOf(javaArguments);
       }
       return this;
     }
 
     public Builder setJvmFlags(@Nullable List<String> jvmFlags) {
       if (jvmFlags != null) {
-        this.jvmFlags = jvmFlags;
+        Preconditions.checkArgument(Iterables.all(jvmFlags, Objects::nonNull));
```
coollog: What about `jvmFlags.contains(null)`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -98,28 +100,32 @@ public Builder setMainClass(@Nullable String mainClass) {
...
+        this.jvmFlags = ImmutableList.copyOf(jvmFlags);
       }
       return this;
     }
 
     public Builder setEnvironment(@Nullable Map<String, String> environmentMap) {
       if (environmentMap != null) {
-        this.environmentMap = environmentMap;
+        Preconditions.checkArgument(Iterables.all(environmentMap.values(), Objects::nonNull));
```
coollog: What about a key `null`?

Issue #473, opened 2018-06-29
Add ability to configure extra layer with arbitrary files.


coollog:
This will help with implementing https://github.com/GoogleContainerTools/jib/blob/more-layers-proposal/proposals/java_agents_and_more_layers.md once that is approved. Probably, for now, we'll just have it as `BuildConfiguration#setExtraFilesLayerConfiguration(LayerConfiguration)`.

Tasks:
- [x] `ReproducibleLayerBuilder` needs to be configured with `extractionPath` for each set of files added. #475
- [x] Add `LayerConfiguration` - built with a set of `sourceFile`, `destinationOnImage`, which represents a layer with those are the `TarArchiveEntry`s. #476
- [x] Add `BuildConfiguration#setExtraFilesLayerConfiguration` #480
- [x] Cache needs to be changed to have each layer be able to take multiple source files and extraction path pairs #508 
- [x] `BuildAndCacheApplicationLayers` needs to build the extra-files layer, and the image should have that layer. #489 
~~- [ ] Add Maven/Gradle configuration.~~
- [x] Add incubating feature to use files in `src/main/jib` as extra layer #562 

Cleanup:
- [x] Consolidate `LayerConfiguration` with `ReproducibleLayerBuilder#LayerEntry` as a `LayerEntry` in package `image` #503

PR #476, opened 2018-06-29
Adds LayerConfiguration.

coollog:
Part of #473
This will be fed into `BuildConfiguration` to configure the extra-files layer.


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @param destinationOnImage the destination on the container image
+     * @return this
+     */
+    public Builder setDestinationOnImage(String destinationOnImage) {
+      this.destinationOnImage = destinationOnImage;
+      return this;
+    }
+
+    public LayerConfiguration build() {
```
loosebazooka: required fields are usually passed into the builder directly.

```
LayerConfiguration.builder(sourceFiles, destinationOnImage).build()
```

loosebazooka: so we don't have to check what are required at what aren't

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * <p>The source files are specified as a list instead of a set to define the order in which they
+   * are added.
+   *
+   * @param sourceFiles the source files to build from. Source files that are directories will have
+   *     all subfiles in the directory added (but not the directory itself)
+   * @param destinationOnImage Unix-style path to add the source files to in the container image
+   *     filesystem
+   */
+  public LayerConfiguration(List<Path> sourceFiles, String destinationOnImage) {
```
TadCordle: Should we verify that `destinationOnImage` is a valid Unix-style path?

coollog: That might be more than we want in the constructor. In terms of using this class in a potential future library release, I think having the javadoc state the constraint is good enough rather than adding a hidden check ourselves. An alternative is to just accept a Path and convert it to its Unix string form.

PR #480, opened 2018-06-29
 Adds BuildConfiguration#setExtraFilesLayerConfiguration.

coollog:
Part of #473


coollog:
Will actually merge #503 first, since that makes `LayerConfiguration` able to take multiple entries.

loosebazooka:
Should just generically accept layer configurations? Especially if that works towards our goal of library-fying jib.

coollog:
@loosebazooka okay I can file a separate PR for that since that includes a lot more changes to the codebase.

coollog:
Filed #516 , but we might want to have this first before that so that we can get the extra files feature out sooner.

PR #489, opened 2018-07-02
 Builds the extra files layer.

coollog:
Part of #473


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
```diff
@@ -41,36 +41,54 @@
...
+              .add(
+                  new BuildAndCacheApplicationLayerStep(
+                      "classes",
+                      listeningExecutorService,
+                      buildConfiguration,
+                      sourceFilesConfiguration.getClassesFiles(),
+                      sourceFilesConfiguration.getClassesPathOnImage(),
+                      cache));
+
+      // Adds the extra layer ot be built if configured.
```
TadCordle: ot -> to

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+package com.google.cloud.tools.jib.configuration;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import java.nio.file.Path;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** Configures how to build a layer in the container image. */
+// TODO: Consolidate with ReproducibleLayerBuilder#LayerEntry.
```
TadCordle: We should make an issue for this

coollog: Added as a cleanup task in #473

coollog:
Still need to add a test for this new extra files layer.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -80,6 +80,7 @@ public void addEntry(TarArchiveEntry entry) {
               ByteStreams.copy(contentStream, tarArchiveOutputStream);
             }
           }
+          // Note that this will skip files that don't exist.
```
briandealwis: This comment seems better suited above?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
```diff
@@ -33,19 +32,18 @@
   /** Entry into the layer metadata. */
   static class LayerMetadataEntry {
 
-    private List<String> sourceFilesStrings;
+    private ImmutableList<String> sourceFilesStrings;
```
briandealwis: Why do these return types need to change?

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheReaderTest.java
```diff
@@ -134,28 +133,29 @@ public void testGetUpToDateLayerBySourceFiles()
...
 
     try (Cache cache = Cache.init(testCacheFolder)) {
       CacheReader cacheReader = new CacheReader(cache);
 
+      ImmutableList<LayerEntry> layerEntryFilter =
+          ImmutableList.of(
+              new LayerEntry(ImmutableList.of(testSourceFiles), "/some/extraction/path"));
+
+      CachedLayerWithMetadata upToDateLayer =
+          cacheReader.getUpToDateLayerByLayerEntries(layerEntryFilter);
```
briandealwis: Aside: this `layerEntryFilter` doesnt seem like a filter (esp given that there is a LayerFilter class), and `getUpToDateLayerByLayerEntries()` seems more `WithLayerEntries()`?  The filter made me think that this could return a cached layer that contained *more* than these layer entries.

briandealwis:
I think the only part that would change for a layers-as-directories would be this section: https://github.com/GoogleContainerTools/jib/pull/489/files#diff-40e82e27bc4a6e9096b9caf1226407d9R89

coollog:
We wouldn't actually need to change anything. The caller can just create the right `LayerConfiguration` to set as the `extraFilesLayerConfiguration` - so like it could be `sourceFiles=src/main/jib` and `extractionPath=/` as the single layer entry in the `LayerConfiguration`.

loosebazooka:
Maybe I'm misunderstanding, but the directories as layers approach allows for multiple layers (<-- is this something we do not want?)

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
coollog: I'm trying to keep everything that is not user-facing (user as in a potential future user of jib-core library) immutable.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheReaderTest.java
coollog: Ah yes, that's a bad naming. Will change.

coollog:
I misunderstood - yea if we do each directory as a separate layer, this will need to change.

coollog:
Yep, in #516, all layers will be passed as a list in `BuildConfiguration#setLayerConfigurations`

PR #490, opened 2018-07-02
Add log messages for using credentials from maven settings

TadCordle:
Another TODO


coollog:
My bad, I just looked at `RetrieveRegistryCredentialsStep` L210 and it already logs where the credentials are from, which is "Maven settings" from `MavenSettingsServerCredentials`. It's just at the info level, which may be what we want anyways.

TadCordle:
Alright, we can just remove the comment then

PR #491, opened 2018-07-02
Moves docs-unreleased/ to docs/ and adds Google Analytics to docs site.

coollog:
We should probably think of a better system for making staging/production changes to the docs site.


docs/_config.yml
```diff
@@ -1 +1,5 @@
 theme: jekyll-theme-slate
+
+# Slate theme-specific configuration.
+# See https://github.com/pages-themes/slate
+google_analytics: UA-121724379-1
```
chanseokoh: Hmm... is this the only way? People will clone our Jib repo, and this may propagate to everywhere.

coollog: Hmm, this wouldn't be the case if we had a separate repo for Jib documentation. Is it a problem to have others clone our tracking ID everywhere? The tracking ID only works for https://googlecontainertools.github.io/jib

PR #492, opened 2018-07-02
Makes jibExportDockerContext.targetDir configurable from build.gradle.

coollog:
This makes `jibExportDockerContext` configurable like:

`build.gradle`:
```groovy
jibExportDockerContext {
    targetDir = 'my-dir'
}
```

Whereas before it could only be configured from a command line property `--jib.dockerDir`

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


patflynn:
@loosebazooka is it gradle custom to allow the user to place build outputs outside of the build directory?

coollog:
Filed #493 to not allow exporting outside of build output directory.

patflynn:
@coollog FWIW I'm not strongly opinionated about whether these files should even be considered build outputs. Just asking the question.

coollog:
Yea we'll have `exportDockerContext` match whatever we decide for the tarball export feature.

PR #502, opened 2018-07-03
Proposal: Build Tarball to Filesystem

### LINKED ISSUE #486: Write proposal for building an image tarball.
Currently, we have `jib:dockerBuild`/`jibDockerBuild` for building the image tarball and `docker load`ing it onto a Docker daemon. We would like to add the functionality to build and output the image tarball itself.

TadCordle:
Fixes #486.

Preview here: https://github.com/GoogleContainerTools/jib/blob/a59c749474b6de8e98c211682fc622ed0d81859f/proposals/build_tarball.md

@dgageot


proposals/build_tarball.md
```diff
@@ -0,0 +1,44 @@
...
+command. To avoid duplicate code, a boolean parameter can be used in the steps runners to determine
+whether to build to a Docker daemon or build to a tarball.
+
+The following changes will be made to the code:
+1. Add a private output path configuration parameter to `BuildConfiguration`
+2. Rename:
+   - `BuildTarballAndLoadDockerStep` to `BuildTarballStep`
+   - `runBuildTarballAndLoadDockerStep` to `runBuildTarballStep`
+   - `waitOnBuildTarballAndLoadDockerStep` to `waitOnBuildTarballStep`
+3. Add a boolean parameter `doDockerLoad` to `runBuildTarballStep` and `BuildTarballStep`'s
```
coollog: An alternative is to pass in a what-to-do-with-the-tarball-that-i-just-built parameter that defines the action to perform on the built tarball blob and have static factories on `BuildTarballStep` to use the right action.

proposals/build_tarball.md
```diff
@@ -36,6 +36,7 @@ The following changes will be made to the code:
    - `waitOnBuildTarballAndLoadDockerStep` to `waitOnBuildTarballStep`
 3. Add a boolean parameter `doDockerLoad` to `runBuildTarballStep` and `BuildTarballStep`'s
 constructor
+   - Alternatively, use in an action to more explicitly define the behavior
```
coollog: parameterize with an action?

proposals/build_tarball.md
```diff
@@ -0,0 +1,45 @@
...
+   - `waitOnBuildTarballAndLoadDockerStep` to `waitOnBuildTarballStep`
+3. Add a boolean parameter `doDockerLoad` to `runBuildTarballStep` and `BuildTarballStep`'s
+constructor
+   - Alternatively, parameterize an action to more explicitly define the behavior
+4. Check the value of `doDockerLoad` in `BuildTarballStep#afterPushBaseImageLayerFuturesFuture()`
+   - If true, run the existing `new DockerClient().load(...)`
+   - If false, write the tarball blob to a file at the output location
+5. Add `BuildSteps#forBuildToTarball()`, which would pass in the required messages/boolean parameter
+6. Add a new task and mojo that would call `BuildSteps.forBuildToTarball()`
+7. (Optional/future) Expose configuration parameter to users
```
loosebazooka: Lets skip `7` completely for now, until we have an obvious use-case.

proposals/build_tarball.md
```diff
@@ -0,0 +1,43 @@
...
+command. To avoid duplicate code, a boolean parameter can be used in the steps runners to determine
+whether to build to a Docker daemon or build to a tarball.
+
+The following changes will be made to the code:
+1. Add a private output path configuration parameter to `BuildConfiguration`
+2. Rename:
+   - `BuildTarballAndLoadDockerStep` to `BuildTarballStep`
+   - `runBuildTarballAndLoadDockerStep` to `runBuildTarballStep`
+   - `waitOnBuildTarballAndLoadDockerStep` to `waitOnBuildTarballStep`
+3. Add a boolean parameter `doDockerLoad` to `runBuildTarballStep` and `BuildTarballStep`'s
```
loosebazooka: build steps probably should not have inputs that change their behavior. For now, to me, it seems like it would make sense to do this as a new static factory on BuildSteps that is just `forBuildToDockerDaemon` without the final step of `docker load`

I guess this is a good argument for choosing a final step and building the steps tree backwards from there? So we can potentially plug in to any random step anywhere as a build target?

loosebazooka: That second paragraph doesn't really have anything to do with this proposal, it's about the underlying mechanism. So you can ignore it.

TadCordle: The way it is right now, the build to docker daemon doesn't actually generate the tar file on disk, it just pipes the tarball data directly to a `docker load` process. The tarball generation + piping to `docker load` happen in the same step at the moment, so it wouldn't really be a matter of just removing the last step, it would mean splitting up the last step into two parts and adding an extra step for generating the file.

proposals/build_tarball.md
```diff
@@ -29,9 +29,9 @@ whether to build to a Docker daemon or build to a tarball.
 
 The following changes will be made to the code:
 1. Add a private output path configuration parameter to `BuildConfiguration`
-2. Break `BuildTarballAndLoadDockerStep` into two step - `BuildTarballStep` and `LoadDockerStep`
+2. Rename `BuildTarballAndLoadDockerStep` to `LoadDockerStep`
 3. Add a new step `WriteTarFileStep` for writing a Blob to disk
 4. Add `BuildSteps#forBuildToTarball()`, which would contain the same steps as
-`forBuildToDockerDaemon()` up until after the `BuildTarballStep`, where it would run a
+`forBuildToDockerDaemon()` up until after the `FinalizingStep`, where it would run a
```
coollog: Oh, not related to this proposal, but perhaps we should log a message like `Loading to Docker daemon...` for the `LoadDockerStep`

TadCordle: I'll file an issue for it.

coollog:
LGTM again from me.

PR #504, opened 2018-07-03
Updates formatter version.

coollog:
This may fix the breaking macOS builds. The sherter/google-java-format plugin was updated to support the latest google-java-format version.


coollog:
This time it looks like its failing to GET the google-java-format v1.6. Going to retry the build because it definitely exists.

coollog:
Passed

PR #508, opened 2018-07-06
Changes cache metadata to take many layer entries instead of just one.

coollog:
Part of #473 

**Rationale:**

The cache metadata works by storing the set of source files that constructed each layer. This way, a layer can be found by the source files it was constructed from. This was fine before when each layer only extracts to a single path. However, for #473, we need to be able to have multiple source files-extraction path pairs (layer entries) for a single layer. This requires the cache metadata to be changed to accommodate this.


jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
```diff
@@ -28,35 +30,72 @@
...
+
+    List<String> getSourceFilesStrings() {
+      return sourceFilesStrings;
+    }
+
+    String getExtractionPath() {
+      return extractionPath;
+    }
+
+    //    void setSourceFilesStrings(List<String> sourceFilesStrings) {
```
TadCordle: Leftover commented code

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
```diff
@@ -28,35 +30,72 @@
...
     this.lastModifiedTime = lastModifiedTime;
   }
 
-  List<String> getSourceFiles() {
-    return sourceFiles;
+  ImmutableList<LayerMetadataEntry> getEntries() {
+    return entries;
   }
 
   public FileTime getLastModifiedTime() {
```
TadCordle: public methods should go first?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -125,13 +125,7 @@ public UnwrittenLayer build() throws IOException {
...
 
-  public List<Path> getSourceFiles() {
-    List<Path> allSourceFiles = new ArrayList<>();
-
-    for (LayerEntry layerEntry : layerEntries) {
-      allSourceFiles.addAll(layerEntry.getSourceFiles());
-    }
-
-    return allSourceFiles;
+  public ImmutableList<LayerEntry> getLayerEntries() {
```
loosebazooka: This isn't really related to this PR, but I'm a little behind.

I wonder if `ReprocibleLayerBuilder` should not have `build()` and `getLayerEntries()` called on it, perhaps `CacheWriter#writeLayer` should take a `ReproducibleLayer` type (from `ReproducibleLayer.builder()`) and have `unwrittenLayer` and `layerEntries` methods on it. Only because we can't programmatically guarantee that there are no changes to ReproducibleLayerBuilder between calls to `build` and `getLayerEntries`.

coollog: Agreed. Another way is to create an interface that's like `LayerBuilder` and `CacheWriter#writeLayer` will take a `LayerBuilder` (with `ReproducibleLayerBuilder` being an interface). All implementations of `LayerBuilder` should be immutable them being built with some builder nested class.

PR #509, opened 2018-07-06
Adds ga-beacon to README.

coollog:
Adds https://github.com/igrigorik/ga-beacon so we can get more in-depth traffic analytics.

Preview at: https://github.com/GoogleContainerTools/jib/tree/coollog-patch-2


chanseokoh:
The instructions say

> The ga-beacon.appspot.com instance is a *demo* instance, good for prototyping and proof of concepts. If you intend to use this in production for your application, you should deploy *your own instance* of this service,

coollog:
I think we can add this for now to try it out (as a proof of concept/prototype), and then deploy our own instance(s) if it works well for us.

coollog:
Will be deploying our own App Engine instance.

coollog:
Made the beacon a pixel since the badge doesn't provide much value to viewers.

PR #511, opened 2018-07-06
Use PortsWithProtocol instead of strings

TadCordle:
Part of #442


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
```diff
@@ -79,6 +79,30 @@ public Protocol getProtocol() {
     return protocol;
   }
 
+  @Override
+  public boolean equals(Object other) {
+    if (other == this) {
+      return true;
+    }
+    if (other == null || other.getClass() != PortsWithProtocol.class) {
```
chanseokoh: Curious. Is there a reason to prefer this over `!other instanceof PortsWithProtocol`? Because of inheritance?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
```diff
@@ -79,6 +79,30 @@ public Protocol getProtocol() {
...
+    }
+    PortsWithProtocol otherPort = (PortsWithProtocol) other;
+    return minPort == otherPort.minPort
+        && maxPort == otherPort.maxPort
+        && protocol == otherPort.protocol;
+  }
+
+  @Override
+  public int hashCode() {
+    return (protocol == Protocol.TCP ? 1 : -1) * minPort + (maxPort << 16);
```
coollog: This may work for the assumption that maxPort is within 16 bits, but this class does not guarantee that maxPort is within that size.

chanseokoh: "0-1024/udp" and "0-1024/tcp" ends up with the same code. How about `Objects.hash(minPort, maxPort, protocol)`? Not that I know now `Objects.hash()` works.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -127,12 +130,14 @@ public void setContainerCmd(List<String> cmd) {
...
 
-  public void setContainerExposedPorts(List<String> exposedPorts) {
+  public void setContainerExposedPorts(List<PortsWithProtocol> exposedPorts) {
     // TODO: Do this conversion somewhere else
     ImmutableSortedMap.Builder<String, Map<?, ?>> result =
         new ImmutableSortedMap.Builder<>(String::compareTo);
-    for (String port : exposedPorts) {
-      result.put(port, Collections.emptyMap());
+    for (PortsWithProtocol port : exposedPorts) {
+      for (int i = port.getMinPort(); i <= port.getMaxPort(); i++) {
```
coollog: We want to avoid single letter variables. Something like `port` would be better.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -127,12 +130,14 @@ public void setContainerCmd(List<String> cmd) {
     config.Cmd = cmd;
   }
 
-  public void setContainerExposedPorts(List<String> exposedPorts) {
+  public void setContainerExposedPorts(List<PortsWithProtocol> exposedPorts) {
```
coollog: This function should probably still take a `String`, and the list of strings should be built somewhere else, and same for the getter.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -161,14 +166,18 @@ public void addLayerDiffId(DescriptorDigest diffId) {
...
+  ImmutableList<PortsWithProtocol> getContainerExposedPorts() {
     // TODO: Do this conversion somewhere else
     if (config.ExposedPorts == null) {
       return null;
     }
-    ImmutableList.Builder<String> ports = new ImmutableList.Builder<>();
+    ImmutableList.Builder<PortsWithProtocol> ports = new ImmutableList.Builder<>();
     for (Map.Entry<String, Map<?, ?>> entry : config.ExposedPorts.entrySet()) {
-      ports.add(entry.getKey());
+      List<String> split = Splitter.on('/').splitToList(entry.getKey());
```
coollog: This can use the same parsing mechanism as in `ExposedPortsParser`

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
```diff
@@ -79,6 +79,30 @@ public Protocol getProtocol() {
     return protocol;
   }
 
+  @Override
+  public boolean equals(Object other) {
```
coollog: btw where is this used (and hashCode)?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
chanseokoh: In that sense, the class does not guarantee non-negative numbers either.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
TadCordle: It's used in tests for assertions, and hashCode was only added to avoid gradle warnings.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
TadCordle: Honestly I used it because I forgot `instanceof` exists. So I'm not sure.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
TadCordle: I'll take care of that when I get rid of the TODO

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
TadCordle: If you mean using pattern matching, I found this to be much simpler, since we'll never encounter ranges here.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
TadCordle: Hmm yeah this seems like a better idea

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
coollog: Ah okay, if it's only for tests, it should just be a check in test files. having equals on a class is adding a lot of complexity and possibilities for error, so we should only use it if absolutely necessary.

TadCordle: Is this going to be a user-facing class?

TadCordle: If it's a user-facing class would it be nice to have?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
coollog: What if the config ports is not in the assumed format? Like `split.get(0)` could be not a parseable integer.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
chanseokoh: Just to be clear. I wrongly assumed for a moment that `hashCode()` should be unique for each combination of the fields. So, technically, your original implementation would probably work OK, even with the possible overflow and such.

loosebazooka: +1

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
coollog: Hmm, okay, I guess that may be useful for users.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/PortsWithProtocol.java
loosebazooka: We should probably use `instanceof` here.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
TadCordle: I guess at that point the question becomes how do we handle a malformed container config in general, though? The Go Container type explicitly states that the exposed ports will be in this format.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPortsParser.java
```diff
@@ -84,11 +81,9 @@
         buildLogger.warn("Port number '" + port + "' is out of usual range (1-65535).");
```
loosebazooka: sounds like we should just error here.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
loosebazooka: It looks like we need to figure out what the spec allows, and handle that (whether it is ranges or not). Once we know that we can fail hard if the image format is invalid.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -121,7 +122,7 @@ public Builder setEnvironment(@Nullable Map<String, String> environmentMap) {
       return this;
     }
 
-    public Builder setExposedPorts(@Nullable List<String> exposedPorts) {
+    public Builder setExposedPorts(@Nullable List<PortsWithProtocol> exposedPorts) {
```
coollog: I took another look at this and I was thinking it might make more sense to just have this either be a `List<PortWithProtocol>` or a `PortsWithProtocol` (no list), where the `PortsWithProtocol` itself can be constructed with a builder that adds single ports and port ranges. This way, we keep the core configuration simple.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
TadCordle: I gave it a try and created a container with a range of exposed ports in the container config:

```
$ docker inspect gcr.io/tcordle-test/hellogradle
...
    "Config": {
        ...
        "ExposedPorts": {
            "1000-1003/udp": {}
        },
        ...
    }
...
```

and then I tried running the container with the -P flag to automatically create port mappings, and I got this error:

```
$ docker run -P gcr.io/tcordle-test/hellogradle
docker: Error response from daemon: driver failed programming external connectivity 
on endpoint affectionate_agnesi8 
(34e06094483a6871f8257c2a26d77935defbb91362fe041c33035602b4f2645b):  
(iptables failed: iptables --wait -t nat -A DOCKER -p udp -d 0/0 --dport 32787 -j DNAT --to-destination 192.168.9.2:0 ! -i docker0: iptables v1.6.1: Port `0' not valid
```

Doesn't look like ranges are supported within the container config.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
TadCordle: Are you suggesting just not keeping track of the ranges again and expanding to individual ports?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
coollog: Yea, I think even if they were we should still follow the spec stated in https://github.com/opencontainers/image-spec/blob/master/config.md#properties, which only allows single ports.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
```diff
@@ -83,7 +83,7 @@ public void testSteps_forBuildToDockerRegistry()
         new Command("docker", "inspect", imageReference).run(),
         CoreMatchers.containsString(
             "            \"ExposedPorts\": {\n"
-                + "                \"1000\": {},\n"
+                + "                \"1000/tcp\": {},\n"
```
loosebazooka: should we keep one in that doesn't have a protocol associated with it?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPortsParser.java
```diff
@@ -70,7 +69,7 @@
       if (!Strings.isNullOrEmpty(matcher.group(2))) {
         max = Integer.parseInt(matcher.group(2));
       }
-      String protocol = matcher.group(3);
+      Protocol protocol = "udp".equals(matcher.group(3)) ? Protocol.UDP : Protocol.TCP;
```
loosebazooka: can we compare directly with the enum for `Protocol.UDP` instead of `"udp"` here?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPortsParser.java
```diff
@@ -80,14 +79,12 @@
 
       // Warn for possibly invalid port numbers
       if (min < 1 || max > 65535) {
-        // TODO: Add details/use HelpfulSuggestions for these warnings
-        buildLogger.warn("Port number '" + port + "' is out of usual range (1-65535).");
+        throw new NumberFormatException(
```
loosebazooka: I wonder if NumberFormatException is not a great exception to throw here. If it's a user configuration error, maybe we should expose it to whoever needs to handle it?

loosebazooka:
This is little subjective, and I'm not sure if we discussed it already and I missed it, but can we just call `PortWithProtocol` -> `Port`?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
TadCordle: We still allow empty protocol configuration in jib (which defaults to TCP), but I felt it was best to keep it explicit in the backend so we don't have to depend on other peoples' defaults.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/ExposedPortsParser.java
TadCordle: So like `Protocol.UDP.toString().equals(...)`?

loosebazooka: I think so? You might need to play around with it a little.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
loosebazooka: Oh, I see, we're just writing in tcp if the user input is just a port. :+1:

PR #516, opened 2018-07-06
Removes SourceFilesConfiguration in favor of list of LayerConfiguration in BuildConfiguration.

### LINKED ISSUE #563: Export extra files layer in Docker context.
Regards https://github.com/GoogleContainerTools/jib/pull/562#pullrequestreview-136397002

Part of #473

### LINKED ISSUE #622: Do not add empty layers to image.
Layers like:
https://github.com/GoogleContainerTools/jib/pull/489
https://github.com/GoogleContainerTools/jib/pull/584

And if dependencies, resources, or classes are empty too.

coollog:
Regards https://github.com/GoogleContainerTools/jib/pull/480#issuecomment-403122259

This is also in preparation for Jib as a library for Java.

This also fixes #622, fixes #563, and moves entrypoint to be provided to `BuildConfiguration` (removing `mainClass` and `jvmFlags`).


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
```diff
@@ -116,24 +108,40 @@ public static BuildSteps forBuildToDockerDaemon(
...
                 .runFinalizingBuildStep()
                 .runBuildTarballAndLoadDockerStep()
                 .waitOnBuildTarballAndLoadDockerStep());
   }
 
   /** Creates the container entrypoint for a given configuration. */
-  private static ImmutableList<String> getEntrypoint(
-      BuildConfiguration buildConfiguration, SourceFilesConfiguration sourceFilesConfiguration) {
+  private static ImmutableList<String> getEntrypoint(BuildConfiguration buildConfiguration) {
+    // TODO: Don't use indexes.
```
loosebazooka: Oh, yeah this is a weird problem.

loosebazooka: Would it be weird to split it into java and non-java layer groups?

coollog:
Fixed a bunch of merge conflicts - still need to make the changes in jib-gradle-plugin and jib-maven-plugin

coollog:
This should be good for review now - this ended up being quite large - I'll do a pass myself too

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
```diff
@@ -42,25 +49,36 @@
...
+  }
+
   @ClassRule public static LocalRegistry localRegistry = new LocalRegistry(5000);
 
   private static final TestBuildLogger logger = new TestBuildLogger();
 
-  @Rule public TemporaryFolder temporaryCacheDirectory = new TemporaryFolder();
-
-  @Rule public TemporaryFolder temporaryTarOutput = new TemporaryFolder();
+  @Rule public TemporaryFolder temporaryFolder = new TemporaryFolder();
```
chanseokoh: I guess it's harmless to have `final`. Is the repo generally lenient about `final`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -200,6 +185,20 @@ public Builder setCreationTime(Instant creationTime) {
       return this;
     }
 
+    /**
+     * Sets the container entrypoint.
+     *
+     * @param entrypoint the tokenized command to run when the container starts
+     * @return this
+     */
+    public Builder setEntrypoint(@Nullable List<String> entrypoint) {
```
chanseokoh: What is the point of allowing `null`? Also, I think we should document the behavior: "passing null does nothing; it does not clear the previous value."

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -65,16 +77,23 @@ public static Builder builder() {
   }
 
   private final ImmutableList<LayerEntry> layerEntries;
+  private String label;
```
chanseokoh: Probably `final`.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
```diff
@@ -42,25 +49,36 @@
 /** Integration tests for {@link BuildSteps}. */
 public class BuildStepsIntegrationTest {
 
+  /** Lists the files in the {@code resourcePath} resources directory. */
+  private static ImmutableList<Path> getFilesList(String resourcePath)
```
TadCordle: If this is specifically for resources maybe it should be renamed to `getResourceFilesList`.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerContextGeneratorTest.java
```diff
@@ -58,23 +56,17 @@ private static void assertSameFiles(Path directory1, Path directory2) throws IOE
...
-    Mockito.when(mockSourceFilesConfiguration.getDependenciesPathOnImage())
-        .thenReturn(expectedDependenciesPath);
-    Mockito.when(mockSourceFilesConfiguration.getResourcesPathOnImage())
-        .thenReturn(expectedResourcesPath);
-    Mockito.when(mockSourceFilesConfiguration.getClassesPathOnImage())
-        .thenReturn(expectedClassesPath);
-  }
+  private String expectedDependenciesPath = "/app/libs/";
+  private String expectedResourcesPath = "/app/resources/";
+  private String expectedClassesPath = "/app/classes/";
```
chanseokoh: Are these suitable for `static final`?

jib-core/src/test/resources/sampleDockerfile
```diff
@@ -4,6 +4,7 @@ COPY libs /app/libs/
 COPY snapshot-libs /app/libs/
 COPY resources /app/resources/
 COPY classes /app/classes/
+COPY root /
```
chanseokoh: Just asking: does this open up the possibility that copying the files at `root/` might overwrite existing files, or is it that `root/` will never have a sub-directory called `app/`?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
```diff
@@ -72,12 +90,16 @@ public void testSteps_forBuildToDockerRegistry()
...
                 .setJavaArguments(Collections.singletonList("An argument."))
                 .setExposedPorts(
                     ExposedPortsParser.parse(Arrays.asList("1000", "2000-2002/tcp", "3000/udp")))
                 .setAllowHttp(true)
+                .setLayerConfigurations(fakeLayerConfigurations)
+                .setEntrypoint(
+                    JavaEntrypointBuilder.makeDefaultEntrypoint(
+                        Collections.emptyList(), "HelloWorld"))
                 .build());
+    buildImageSteps.run();
```
TadCordle: Why are we running this 3 times now?

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ All notable changes to this project will be documented in this file.
...
 
+## 0.9.8
+
+### Added
+
+- Docker context generation now includes snapshot dependencies and extra files ([#516](https://github.com/GoogleContainerTools/jib/pull/516/files))
+
+### Changed
+
+- Only builds non-empty layers ([#516](https://github.com/GoogleContainerTools/jib/pull/516/files))
```
chanseokoh: Is it true? I saw above

```java
   // TODO: Don't generate empty layers.
   /**
    * Creates the Docker context in {@code #targetDirectory}.
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -210,13 +209,10 @@ public BuildConfiguration build() {
...
         errorMessages.add("target image is required but not set");
       }
-      if (mainClass == null) {
-        errorMessages.add("main class is required but not set");
-      }
 
       switch (errorMessages.size()) {
         case 0: // No errors
-          if (baseImageReference == null || targetImageReference == null || mainClass == null) {
+          if (baseImageReference == null || targetImageReference == null) {
```
TadCordle: Should we error if `entrypoint` is empty like we used to error when `mainClass` was null?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
```diff
@@ -49,16 +49,15 @@
...
-    private ImmutableList<String> jvmFlags = ImmutableList.of();
     private ImmutableMap<String, String> environmentMap = ImmutableMap.of();
     private ImmutableList<Port> exposedPorts = ImmutableList.of();
     private Class<? extends BuildableManifestTemplate> targetFormat = V22ManifestTemplate.class;
     @Nullable private CacheConfiguration applicationLayersCacheConfiguration;
     @Nullable private CacheConfiguration baseImageLayersCacheConfiguration;
     private boolean allowHttp = false;
-    @Nullable private LayerConfiguration extraFilesLayerConfiguration;
+    private ImmutableList<LayerConfiguration> layerConfigurations = ImmutableList.of();
+    private ImmutableList<String> entrypoint = ImmutableList.of();
```
TadCordle: This is very much a nit, so feel free to ignore, but for organization it'd be nice to put `entrypoint` next to `javaArguments` instead of at the end.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -36,26 +37,91 @@
...
 
   @Nullable private String baseImage;
   private List<String> jvmFlags = Collections.emptyList();
   private String mainClass = "";
   private List<String> javaArguments = Collections.emptyList();
   private List<String> exposedPorts = Collections.emptyList();
 
-  public DockerContextGenerator(SourceFilesConfiguration sourceFilesConfiguration) {
-    this.sourceFilesConfiguration = sourceFilesConfiguration;
+  // TODO: Just take the LayerConfigurations.
```
TadCordle: Would it be that difficult to just do this now? I think we should avoid adding TODOs while trying to fix other issues.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
```diff
@@ -114,6 +180,7 @@ public DockerContextGenerator setExposedPorts(List<String> exposedPorts) {
     return this;
   }
 
+  // TODO: Don't generate empty layers.
```
TadCordle: Same here.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/BuildStepsRunner.java
```diff
@@ -166,6 +156,13 @@ private static void handleRegistryUnauthorizedException(
     }
   }
 
+  private static String capitalizeFirstLetter(String string) {
```
TadCordle: Could use `StringUtils#capitalize()` instead?

jib-core/src/test/java/com/google/cloud/tools/jib/builder/BuildConfigurationTest.java
```diff
@@ -198,7 +179,7 @@ public void testBuilder_missingValues() {
 
     } catch (IllegalStateException ex) {
       Assert.assertEquals(
-          "base image is required but not set, target image is required but not set, and main class is required but not set",
```
TadCordle: Hmm I think we should keep a test for the `default` case of having 3 or more error messages. Or we can just delete that `default` case logic since we only have two possible error messages at the moment (missing target and missing base).

jib-gradle-plugin/CHANGELOG.md
TadCordle: That's just for generating the Docker context; I don't think empty layers are added to the image itself in `jib:build` etc.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
coollog: We should definitely have `final` where-ever applicable and have things be non-final unless absolutely necessary.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
coollog: The point of allowing `null` is to allow for building an image without an entrypoint (since it's an optional field on the image format). This can also be useful for when we propagate base image container configuration so fields like entrypoint could be propagated. But yes, we should defintely document the behavior.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/BuildStepsRunner.java
```diff
@@ -190,23 +187,16 @@ public void build(HelpfulSuggestions helpfulSuggestions) throws BuildStepsExecut
...
-          .forEach(resourceFile -> buildLogger.info("\t\t" + resourceFile));
-
-      buildLogger.info("\tDependencies:");
-      buildSteps
-          .getSourceFilesConfiguration()
-          .getDependenciesFiles()
-          .forEach(dependencyFile -> buildLogger.info("\t\t" + dependencyFile));
+      for (LayerConfiguration layerConfiguration :
+          buildSteps.getBuildConfiguration().getLayerConfigurations()) {
+        buildLogger.info("\t" + capitalizeFirstLetter(layerConfiguration.getLabel()) + ":");
```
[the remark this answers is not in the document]
coollog: I mostly had this to keep the log message consistent with what it was before. We should definitely consider changing this logic to be something more configurable, but this log message will be part of the "`plugins-common`" rather than `jib-core`. There should probably be an enum from a layer type (DEPENDENCIES, CLASSES, etc.) to the intended label string and log string.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointBuilder.java
```diff
@@ -14,43 +14,51 @@
...
-
-    String classPathsString = String.join(":", classPaths);
-
-    ImmutableList.Builder<String> entrypointBuilder =
-        ImmutableList.builderWithExpectedSize(4 + jvmFlags.size());
+  public static List<String> makeEntrypoint(
+      List<String> classpathElements, List<String> jvmFlags, String mainClass) {
+    String classpathString = String.join(":", classpathElements);
+
+    List<String> entrypointBuilder = new ArrayList<>(4 + jvmFlags.size());
```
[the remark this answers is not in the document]
coollog: I'll change this to `EntrypointConstructor` (like in the jib library design).

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerContextGeneratorTest.java
coollog: Yep!

jib-core/src/test/resources/sampleDockerfile
coollog: Yes, it opens up that possibility.

jib-gradle-plugin/CHANGELOG.md
coollog: Oops forgot to remove that - it actually skips the empty layers now.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
coollog: Oops, removing.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
coollog: I'm intending it to be possibly be empty to allow for propagation of entrypoint from base images in the future (when this is used as a library).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
coollog: Sounds good

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerContextGenerator.java
coollog: I'll look into that again, but the main issue is that since this is in `jib-core`, it doesn't know what layer label corresponds to which copy directive.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
chanseokoh: Alright, just keep in mind you'd normally consider what locale to use if the text can be something else than English, e.g., https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#toUpperCase(java.util.Locale)

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/BuildStepsRunner.java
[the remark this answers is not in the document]
chanseokoh: Alright, just keep in mind you'd normally consider what locale to use if the text can be something else than English, e.g., https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#toUpperCase(java.util.Locale)

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/BuildStepsRunner.java
coollog: I'm trying to avoid having to add the Apache Commons library just for this function, and preferring Guava over Apache Commons.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
chanseokoh: If so, why not just clear the entrypoint if `null` is given? That's more intuitive and what I'd expect normally.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
TadCordle: (This same ordering happens in a bunch of other places too, but I didn't want to overload the comments.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
coollog: That makes a lot of sense.

coollog: Or, actually - I think we should probably have the semantics be like:

`null` - if base image has entrypoint, propagate; otherwise, don't set an entrypoint
non-`null` - set entrypoint to what is given (empty would be empty entrypoint, etc.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
coollog: Yea, maybe we can leave that to another PR that can set up a nice ordering pattern for all of these parameters.

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+## 0.9.8
```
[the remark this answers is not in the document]
coollog: Yes, oops.

coollog:
Trying to fix an integration test on this, will merge after that is fixed.

coollog:
Ah looks like it was just timing out when pulling the image because I am on a slow internet.

PR #517, opened 2018-07-08
Updates versions to 0.9.2.

coollog:


coollog:
Yea, we probably could, but it is quite simple to just change all texts
matching a prior version to a new version except in CHANGELOG files

PR #518, opened 2018-07-09
Fix exception-related errors in Eclipse

briandealwis:
Minor tweaks to fix exception errors flagged in Eclipse.


briandealwis:
Hmm these kokoro builds still haven't completed?

loosebazooka:
It might be an external contributor thing, let me see if I can trigger it.

briandealwis:
> At least 1 approving review is required by reviewers with write access.

Apparently @elharo doesn't have write access here.

PR #537, opened 2018-07-10
Move map<->list translation out of ContainerConfigurationTemplate

### LINKED ISSUE #442: Move expose ports logic to more appropriate locations

TadCordle:
After #511 
Fixes #442 

Moves some translation logic out of the container config JSON template.


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BadConfigurationFormatException.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.image.json;
+
+/** Exception thrown when trying to parse a bad image configuration format. */
+public class BadConfigurationFormatException extends Exception {
```
briandealwis: This exception as is seems fairly generic.  Would it be worth providing the Path or source object to the configuration file?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -115,5 +134,33 @@
...
+  @VisibleForTesting
+  static ImmutableList<Port> portMapToList(@Nullable Map<String, Map<?, ?>> portMap)
+      throws BadConfigurationFormatException {
+    if (portMap == null) {
+      return ImmutableList.of();
+    }
+    ImmutableList.Builder<Port> ports = new ImmutableList.Builder<>();
+    for (Map.Entry<String, Map<?, ?>> entry : portMap.entrySet()) {
+      String port = entry.getKey();
+      Matcher matcher = portPattern.matcher(port);
```
briandealwis: Would this be better on `Port` (though I suppose this could be considered a file-specific format)? Or breaking out to a separate method?

TadCordle: I think we wanted to avoid translation to/from file-specific formats in `Port` since it's going to be a user-facing class in the jib library, and we don't want to break users if the format needed to change.

Maybe we could break this out to a separate method, but this is the only place it's used, so I think it's something we could do later if necessary (similar code is in `ExposedPortParser`, but again I think we wanted to differentiate between parsing plugin configuraiton and parsing container configuration, even if the formats are similar right now).

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BadConfigurationFormatException.java
TadCordle: Hmm, that's a good point, although I'm not sure what would be most useful to provide through this exception. Maybe I can just add a TODO here for now and we can change it if a good use-case pops up?

briandealwis: Sure.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
briandealwis: That makes sense, I suppose.  By plugin-configuration, you mean for Maven/Gradle plugin configuration, right?

jib-core/src/test/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslatorTest.java
```diff
@@ -63,21 +66,52 @@ public void testToImage_v21()
...
+            new Port(1000, Protocol.TCP),
+            new Port(2000, Protocol.TCP),
+            new Port(3000, Protocol.UDP));
+    Assert.assertEquals(expected, JsonToImageTranslator.portMapToList(input));
+
+    ImmutableList<Map<String, Map<?, ?>>> badInputs =
+        ImmutableList.of(
+            ImmutableMap.of("abc", ImmutableMap.of()),
+            ImmutableMap.of("1000-2000", ImmutableMap.of()),
+            ImmutableMap.of("/udp", ImmutableMap.of()));
```
briandealwis: Worth adding an unsupported protocol? 1000/xxx

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
TadCordle: Yeah, Maven/Gradle in our case, and however users choose to parse their input when jib becomes a library.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/Port.java
```diff
@@ -36,6 +37,20 @@
...
+     *
+     * @param protocolString the case insensitive string (e.g. "tcp", "udp")
+     * @return the {@link Protocol} with the corresponding name, or {@code Protocol#TCP} by default
+     *     if invalid
+     */
+    public static Protocol getFromString(@Nullable String protocolString) {
+      if (protocolString == null) {
+        return TCP;
+      }
+      return UDP.toString().toLowerCase().equals(protocolString.toLowerCase()) ? UDP : TCP;
```
coollog: There's a `equalsIgnoreCase` function: https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#equalsIgnoreCase(java.lang.String)

Issue #539, opened 2018-07-11
Failed to connect to docker registry using Basic Authentication


sunnychanwork:
**Description of the issue**:
Failed to connect to docker registry using Basic Authentication

**Expected behavior**:
jib is able to connect to the registry

**Steps to reproduce**: 
Setup a docker registry using https with basic authentication
Pull an image using <from></from>

**Environment**: 
Windows 10
Maven 3.5.3

**Log output**: 
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.2:build (default-cli) on project testing: Build image failed: Tried to retrieve authentication method for myregistry:1234 but failed because: Failed get authentication method from 'WWW-Authenticate' header | If this is a bug, please file an issue at https://github.com/GoogleContainerTools/jib/issues/new: Failed to authenticate with the registry because: 'Bearer' was not found in the 'WWW-Authenticate' header, tried to parse: BASIC realm="Sonatype Nexus Repository Manager" -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.2:build (default-cli) on project testing: Build image failed
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:213)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:154)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:146)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:117)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:81)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:56)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:128)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:305)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:192)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:105)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:956)
[...]

PR #546, opened 2018-07-11
RegistryAuthenticator handle basic, case insensitive (#539)

conblem:
#539


coollog:
Thanks for your contribution!

Issue #514, opened 2018-07-06
Allow user to build image as tar file that can be 'docker load'ed


TadCordle:
https://github.com/GoogleContainerTools/jib/blob/master/proposals/build_tarball.md

- [x] Add private output path configuration parameter
- [x] Rename `BuildTarballAndLoadDockerStep` to `LoadDockerStep`
- [x] Add `WriteTarFileStep`
- [x] Add `BuildSteps#forBuildToTarball()`
- [x] Add new task to Gradle plugin
- [x] Add new mojo to Maven plugin

briandealwis:
Would it make sense to attach the tarball as an artifact to the project?

Is there any value in supporting a new packaging that would:
  - use `package` to generate the tar ball
  - use `install` to build to docker
  - use `deploy` to push to registry

TadCordle:
I'm not sure we want to do that, since currently when we generate the tarball blob, we never actually write it to disk. Adding the tarball as an actual artifact would add some unnecessary I/O to the other steps.

briandealwis:
Most of the repository managers are adding support for docker images anyways.

PR #557, opened 2018-07-11
Add tarOutputPath parameter to BuildConfiguration

TadCordle:
Part of #514, in preparation for implementing https://github.com/GoogleContainerTools/jib/blob/master/proposals/build_tarball.md

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildConfiguration.java
@@ -27,6 +27,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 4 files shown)


TadCordle:
Hm yeah. It looks like we'll probably have to do the validation outside of `BuildConfiguration` construction actually, since different operations require different parameters, and we can't tell which operation will be happening from within `BuildConfiguration`.

coollog:
I think the either/or of `targetImageReference`/`tarOutputPath` might not be the right way of doing it, since for normal `jib:build`, having `tarOutputPath` instead of `targetImageReference` is an incorrect `BuildConfiguration`.

I think we should rather do this like the `DockerContextGenerator`, where the `tarOutputPath` is not configured on `BuildConfiguration`, but rather is a parameter passed to the static factory for the `BuildStepsRunner` for building the tarball. This is a small enough edge-feature that we probably don't need to put into our main `BuildConfiguration`.

TadCordle:
That's a much better idea, I'll close this and handle that when I add the method in `BuildStepsRunner`

PR #559, opened 2018-07-11
Rename BuildTarballAndLoadDockerStep to LoadDockerStep

TadCordle:
Another small part of #514

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -63,7 +63,7 @@ public class StepsRunner {
   @Nullable private BuildImageStep buildImageStep;
   @Nullable private PushContainerConfigurationStep pushContainerConfigurationStep;
   @Nullable private PushImageStep pushImageStep;
-  @Nullable private BuildTarballAndLoadDockerStep buildTarballAndLoadDockerStep;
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -208,7 +208,7 @@ public void waitOnPushImageStep() throws ExecutionException, InterruptedExceptio
 
   public void waitOnBuildTarballAndLoadDockerStep()
```
briandealwis: And same again.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -192,8 +192,8 @@ public StepsRunner runPushImageStep() {
   }
 
   public StepsRunner runBuildTarballAndLoadDockerStep() {
```
briandealwis: Should this method be renamed?  Hmm, I guess not as it will need to build the tarball when that is separated.

TadCordle: Yeah good catch, this should probably be renamed. We ended up agreeing on just creating a separate step instead of splitting this one up - https://github.com/GoogleContainerTools/jib/pull/502#discussion_r200652779

Issue #213, opened 2018-04-06
Allow copying of directory to new layer in image.


coollog:
The current way of adding arbitrary files would be to include them in the `src/main/resources` directory. However, that adds the files to the classpath, which may not be desirable.

The use case of being able to add arbitrary files include being able to add agent libraries to the image, such as [JRebel](http://manuals.zeroturnaround.com/jrebel/remoteserver/docker.html).

The proposed configuration for Maven is:

```xml
<configuration>
  <copy>
    <fromDirectory></fromDirectory>
    <pathOnImage></pathOnImage>
  </copy>
</configuration>
```

The `fromDirectory`'s contents would be placed in the directory on the image at `pathOnImage`.

For Gradle:

```groovy
jib {
  copy {
    fromDirectory = 
    pathOnImage = 
  }
}
```

coollog:
@loosebazooka @TadCordle @patflynn How does this sound? The alternative is to allow to user to define multiple directories to copy.

TadCordle:
I feel like the naming could be improved; I'm not sure why, but I feel like calling it "adding files to the image" gets the point across better than "copying files". Maybe instead of `copy` the parameter could be called `add`?

Also, should there be more symmetry between the two inner parameters? When I see something start with "from" I usually expect the destination to start with "to".

coollog:
For the naming, the main reasoning behind using `copy` is to keep it consistent with what people are used to for Dockerfiles. Dockerfiles has both `COPY` and `ADD` directives, but `ADD` [does some more magic behind the scenes](https://stackoverflow.com/questions/24958140/what-is-the-difference-between-the-copy-and-add-commands-in-a-dockerfile), such as unpacking an archive. We could change this to add, but if we do, it might be best to have it do the same as what `ADD` would do to avoid confusion?

For the inner parameters, perhaps we can call it `fromDirectory` and `toPathOnImage`? Or shorten it to `src` and `dest` to have it like Dockerfile? Or, we could just change `copy` (or `add`) to be a method with two parameters such that you would write something like:

```groovy
jib {
  copy 'sourcefolder' '/path/on/image'
}
```

coollog:
This will be taken care of by https://github.com/GoogleContainerTools/jib/issues/473

PR #562, opened 2018-07-11
Builds extra layer with contents of src/main/jib.

coollog:
This might be what we go with for #473 over what was proposed at #455.

This PR just has the changes for the Gradle plugin. Docker context exporting is also not supported yet.


TadCordle:
Also would this close this old issue? https://github.com/GoogleContainerTools/jib/issues/213

coollog:
@TadCordle yea, though I'll close it manually after the Maven one is added too

PR #565, opened 2018-07-11
 Builds extra layer with contents of src/main/jib.

coollog:
See #562 for more details. This does the same for Maven side.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -85,6 +91,18 @@ public void execute() throws MojoExecutionException {
             .setEnvironment(getEnvironment())
             .setExposedPorts(ExposedPortsParser.parse(getExposedPorts()))
             .setAllowHttp(getAllowInsecureRegistries());
+    if (Files.exists(getExtraDirectory())) {
```
loosebazooka: if they explicitly specify a directory that doesn't exist, should we say so?

coollog: Yes, I'll add a TODO for now that should be resolved when this feature is more solidified.

PR #566, opened 2018-07-11
 Makes extraDirectory configurable with Groovy file().

coollog:


coollog:
Will release v0.9.5 with this incubating feature after this PR is merged.

PR #567, opened 2018-07-11
 Add build steps for building tarball

TadCordle:
Part of #514


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @Override
+  public ListenableFuture<Void> getFuture() {
+    return listenableFuture;
+  }
+
+  @Override
+  public Void call() throws ExecutionException, InterruptedException {
+    ImmutableList.Builder<ListenableFuture<?>> dependenciesBuilder = ImmutableList.builder();
```
coollog: Aside: we should probably figure out a good way to organize all the duplications of the code like this in all the steps.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
```diff
@@ -0,0 +1,106 @@
...
+        .call(this::afterPushBaseImageLayerFuturesFuture, listeningExecutorService)
+        .get();
+  }
+
+  private Void afterPushBaseImageLayerFuturesFuture() throws ExecutionException, IOException {
+    Image<CachedLayer> image = NonBlockingSteps.get(NonBlockingSteps.get(buildImageStep));
+
+    // Build the image to a tarball
+    buildConfiguration.getBuildLogger().lifecycle("Building image to tar file...");
+    if (Files.exists(outputPath.getParent())) {
```
coollog: What is this doing? If the file exists, wouldn't `createDirectories` not do anything?

TadCordle: I added it just in case because I wasn't sure - I should really get into the habit of reading documentation on methods I'm not 100% sure of.

TadCordle: Ah wait now I know why I did it - `createDirectory` throws an exception if the directory exists, `createDirectories` doesn't. So I guess I got the two mixed up.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
```diff
@@ -0,0 +1,104 @@
...
+        .call(this::afterPushBaseImageLayerFuturesFuture, listeningExecutorService)
+        .get();
+  }
+
+  private Void afterPushBaseImageLayerFuturesFuture() throws ExecutionException, IOException {
+    Image<CachedLayer> image = NonBlockingSteps.get(NonBlockingSteps.get(buildImageStep));
+
+    // Build the image to a tarball
+    buildConfiguration.getBuildLogger().lifecycle("Building image to tar file...");
+    Files.createDirectories(outputPath.getParent());
```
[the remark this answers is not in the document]
TadCordle: In our use case, yes, the directory should already exist since we're putting it in the build output. But I imagine if this were eventually configurable, then it'd be a better experience to generate all the directories the user specifies rather than throwing an exception if the directory structure isn't already setup.

Issue #403, opened 2018-06-14
Split dependencies into two layers(silent and mutable)


linux-china:
Now Jib separates your application into multiple layers: dependencies, resources and classes, but still a question here. dependencies layer is still very big. We can split dependencies layer. Why create a new layer for dependencies?  I did some research about dependencies in Java application.  Some jars are very big, for example groovy jar, hibernate jar etc.  These jars are very big but stable, and developers don't change their versions frequently.  The developers change the company internal artifacts regularly, or changed by SNAPSHOT automatically.   For example, a Spring Boot Application is almost 30+M, if we create a layer with following artifacts included, the original dependencies layer will be 5M almost, an new 25M layer(silent dependencies layer) could be stable for a long time.  if the developer changes internal artifact version, only push the 5M layer(mutable dependencies layer), not original 30M layer again.  

In Java, another consideration is SNAPSHOT.  During development and testing phases, most guys use SNAPSHOT version. These jars will be changed regularly from continue integration system.  For some case, for example, security jars always were delivered  as SNAPSHOT version.  100K SNAPSHOT jar will make you rebuild the dependencies layer.

We can introduce a silent dependencies layer and group stable & big jars according to groupId and artifact wild match, or mutable dependencies layer.
**`jib-maven-plugin` Configuration:**
```xml
<silentDependencies>org.springframework:*, org.hibernate:*, *:commons-*, org.webjars:*</silentDependencies>
<mutableDependencies>com.yourcompany:*,*:*:*-SNAPSHOT</mutableDependencies>
```

the Dockerfile will copy slient jars into a new layer

```
COPY silent-libs /app/libs/
COPY mutable-libs /app/libs
```
all artifacts should be in libs directory to make dependencies check easy. 

The followings are some very popular artifacts size information for references, and of course the developers can add other jars.  for a spring boot application, almost silent dependencies layer is almost 80+% in size, and mutable dependencies layer almost  20-% in size. 

### Spring Framework related

* spring all:  spring & spring boot & spring cloud = 12M
* reactor all: 3M

### Apache

* apache commons all: 1.8M

### JVM languages

* groovy: 4.7M
* jruby: 10M
* kotlin all: 3.5M
* scala: 5.7M

### web server

* tomcat all: 3.6M
* jetty all: 2.4
* undertow all: 3M

### Java EE

* javax.*: stable, size is small
* javaee: 2M

### Driver & client

* Oracle: ojdbc8.jar 4M
* MySQL: 2M
[...]

PR #584, opened 2018-07-12
Add support for SNAPSHOT dependencies layer

loosebazooka:
Partial solution to #403 

The behavior of this change is to separate out snapshot and non-snapshot dependencies as their own layers but still writes them both to `/app/libs`.

So `SourceFilesConfiguration` only add a getter on location of the snapshot dependency files, but adds no extra configuration paramter for the location to write those files (it's just shared with regular dependencies)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/EntrypointBuilder.java
```diff
@@ -37,6 +37,7 @@
     ImmutableList<String> classPaths =
         ImmutableList.of(
             sourceFilesConfiguration.getDependenciesPathOnImage() + "*",
+            sourceFilesConfiguration.getSnapshotDependenciesPathOnImage() + "*",
```
coollog: Is there any harm to putting the snapshot dependencies in the same `/app/libs/` directory?

coollog: We also might not want to have the snapshot layer if there are no snapshot dependencies, so having that directory added to the classpath may cause directory not found exceptions?

loosebazooka: Yeah, I don't have any reason reason for picking a separate directory. Let me change it.

patflynn:
+1

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
```diff
@@ -63,6 +63,17 @@
                               sourceFilesConfiguration.getDependenciesPathOnImage())
                           .build(),
                       cache))
+              .add(
```
coollog: We should probably not add this if there are no snapshot dependencies.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/TestSourceFilesConfiguration.java
```diff
@@ -70,12 +81,51 @@ public String getClassesPathOnImage() {
     return EXTRACTION_PATH + "classes/";
   }
 
-  /** Lists the files in the {@code resourcePath} resources directory. */
-  private ImmutableList<Path> getFilesList(String resourcePath)
-      throws URISyntaxException, IOException {
-    try (Stream<Path> fileStream =
-        Files.list(Paths.get(Resources.getResource(resourcePath).toURI()))) {
-      return fileStream.collect(ImmutableList.toImmutableList());
+  public static Builder builder() {
```
coollog: ordering nit: for convention in the codebase, we're trying to follow the class element ordering of:
- public->private
- static->non-static
- members->constructor->methods

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleSourceFilesConfiguration.java
```diff
@@ -92,11 +94,16 @@ private GradleSourceFilesConfiguration(Project project, GradleBuildLogger gradle
       if (resourcesOutputDirectory.equals(dependencyFile.toPath())) {
         continue;
       }
-      dependenciesFiles.add(dependencyFile.toPath());
+      if (dependencyFile.getName().contains("SNAPSHOT")) {
```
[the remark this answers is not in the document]
briandealwis: It does look like Gradle supports [explicitly indicating that a dependency may be changing](
https://stackoverflow.com/a/42058780/600339) and [provides an API](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/ExternalModuleDependency.html#isChanging--) ([docs](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sub:declaring_dependency_with_changing_version)).  Opened #694.

PR #587, opened 2018-07-12
Updates CONTRIBUTING with notice about integration testing.

coollog:
This should help people with running integration tests better.


coollog:
Filed #588

Issue #586, opened 2018-07-12
Credentials for Docker Hub may be stored in different registry entry.


coollog:
Some people are experiencing an authorization error when pushing to Docker Hub even if pushing with `docker` works. This may be because Docker Hub credentials could be stored under a different registry entry instead of `registry.hub.docker.com`. For example, a Docker config (located usually at `$HOME/.docker/config.json`) could have:

```
{
	"auths": {
		"https://index.docker.io/v1/": {},
...
```

We should support this alias as well when using Docker config to fetch credentials.

coollog:
May be what's happening for #564

coollog:
We should also support any auths with suffixes after the registry domain (ie. `index.docker.io/v1/` vs. `index.docker.io`) in [this method](https://github.com/GoogleContainerTools/jib/blob/0b22e7acd7b3e7c2ed56e2e6d9449f302d23d94f/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java#L90).

chanseokoh:
How should it work exactly? For example, my intuition is that, if the given registry is `index.docker.io`, it doesn't match the entry `index.docker.io/v1/` in the Docker config because it is more specific. Conversely, if the given repository is `index.docker.io/v1/`, it can match entries of both `index.docker.io` and `index.docker.io/v1`. Is that right? And is the suffix only possible in the Docker config or it is generally applicable in other stores like `settings.xml` and the build configurations (`pom.xml` and `build.gradle`)?

coollog:
So `v1` or `v2` refers to the Docker Registry API version that route implements. For example, GCR uses Docker Registry API V2, so all requests are to `gcr.io/v2/...`. In the image reference, this `v1` or `v2` is omitted. I think the way it should work is that the domain name should match any entry with the suffix.

ie. `index.docker.io` should match `index.docker.io` and `index.docker.io/v1`

I think in `settings.xml`, we should do the opposite: `index.docker.io/v1` should match `index.docker.io/v1/` and not `index.docker.io`. This is so that specifying credentials for `gcr.io/my-gcp-project` should not have those credentials be used for `gcr.io/my-other-gcp-project`, for example.

chanseokoh:
I think we should do the same to https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java#L105, right?

[...]

PR #605, opened 2018-07-13
Check other potential alises if Docker Hub

chanseokoh:
Part of #586.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -178,10 +195,30 @@ public Authorization call() throws IOException, NonexistentDockerCredentialHelpe
...
     }
   }
 
+  @VisibleForTesting
+  static boolean isDockerHubRegistry(String registry) {
+    return DOCKER_HUB_REGISTRIES.stream().anyMatch(registry::equals);
+  }
+
+  @VisibleForTesting
+  Authorization findOtherDockerHubCredential(String dockerHubRegistry)
```
coollog: Generally, we want to try to keep the build steps (`AsyncStep`) as minimal as possible and delegate the real logic to other classes. So, here for example, we should have probably put the credential helper inference logic inside a `DockerCredentialHelperInferer` rather than in this `AsyncStep` so that we could unit test against it well.

chanseokoh: Just to verify, what I understand from your comments is to minimizing code lines in AsyncStep classes by moving code out to different classes, and not about minimizing operations and time the steps take, right?

chanseokoh:
Looking at the code of `RetrieveRegistryCredentialsStep` again, looks like it's enough to look inside `~/.docker/config` for finding other potential Docker Hub aliases (i.e., no need to change `RetrieveRegistryCredentialsStep` but only change `DockerConfigCredentialRetriever`).

I initially thought it would be more complete to do this at the `RetrieveRegistryCredentialsStep` level so that it would also try other credential supplier, but looks like `knownRegistryCredentials` is independent of the given registry name and the credential helpers are based only on the given suffix and the commonly known suffixes. @coollog am I getting this right?

chanseokoh:
Ready for review, but honestly,

I've talked to @coollog, and for now as the first step and a quick fix, this PR just looks into `docker.config`. However, eventually we should also look for credentials with registry aliases in other sources than `docker.config` (for example, `settings.xml`).

chanseokoh:
> Ready for review, but honestly

Oops. Was going to say honestly, I am not sure if there should be some special handling for the extra path `/v1/` as in `index.docker.io/v1/`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
```diff
@@ -0,0 +1,49 @@
...
+          ImmutableList.of("registry.hub.docker.com", "index.docker.io"));
+
+  /**
+   * Returns the list of registry aliases for the given {@code registry}, including {@code registry}
+   * as the first element.
+   *
+   * @param registry the registry for which the alias group is requested
+   * @return non-empty list of registries where {@code registry} is the first element
+   */
+  public static ImmutableList<String> getAliasesGroup(String registry) {
```
briandealwis: This method should return `List` to minimize exposing Guava classes in APIs. Right?

chanseokoh:
@coollog said the way it manipulated the list to place the given registry at the front (i.e., removing the element and adding it at the front in sequence) did not really feel natural. I agree, so I updated to do it in the functional way.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class RegistryAliasGroup {
+
+  private static final ImmutableList<ImmutableList<String>> REGISTRY_ALIAS_GROUPS =
+      ImmutableList.of(
+          // Docker Hub alias group
+          ImmutableList.of("registry.hub.docker.com", "index.docker.io"));
```
coollog: This can be a set?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
```diff
@@ -0,0 +1,51 @@
...
+   * @param registry the registry for which the alias group is requested
+   * @return non-empty list of registries where {@code registry} is the first element
+   */
+  public static List<String> getAliasesGroup(String registry) {
+    for (ImmutableList<String> aliasGroup : REGISTRY_ALIAS_GROUPS) {
+      if (aliasGroup.contains(registry)) {
+        // Found a group. Move the requested "registry" to the front before returning it.
+        Stream<String> self = Stream.of(registry);
+        Stream<String> withoutSelf = aliasGroup.stream().filter(alias -> !registry.equals(alias));
+        return Stream.concat(self, withoutSelf).collect(Collectors.toList());
```
coollog: Perhaps this could even return the stream itself.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryAliasGroupTest.java
```diff
@@ -0,0 +1,45 @@
...
+  @Test
+  public void testGetAliasesGroup_noKnownAliases() {
+    List<String> singleton = RegistryAliasGroup.getAliasesGroup("something.gcr.io");
+    Assert.assertEquals(1, singleton.size());
+    Assert.assertEquals("something.gcr.io", singleton.get(0));
+  }
+
+  @Test
+  public void testGetAliasesGroup_registryHubDockerCom() {
+    Assert.assertArrayEquals(
```
coollog: nit: I think in this codebase, we prefer to use `List` over raw arrays, such that this can be like `Assert.assertEquals(theList, Arrays.asList(a, b, c)`.

jib-maven-plugin/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 - `jib:buildTar` goal to build an image tarball at `target/jib-image.tar`, which can be loaded into docker using `docker load` ([#514](https://github.com/GoogleContainerTools/jib/issues/514))
 
+- For Docker Hub, also tries registry aliases when getting a credential from `docker.config`
```
coollog: nit: probably just call it "the Docker config" - `docker.config` may imply something else
Also, for formatting, it's good to include the pull request/issue number/link and I think this line should probably be under `Fixed`

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.google.cloud.tools.jib.registry;
+
+import com.google.common.collect.ImmutableList;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class RegistryAliasGroup {
```
coollog: For consistent style, we include a javadoc for each class

jib-maven-plugin/CHANGELOG.md
chanseokoh: +1, also adding this to `jib-gradle-plugin`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
chanseokoh: Yeah, no need to be a list.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
chanseokoh: I just tried returning a stream, but it does not seem concise at the caller side. We can switch to Stream later if it seems fit.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryAliasGroupTest.java
chanseokoh: That is certainly better.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
coollog: Sounds good. For later, the caller could use something like:

```
Authorization authorization = getAliasesGroup()
    .map(registryAlias -> retrieve(dockerConfigTemplate, registryAlias))
    .filter(Objects::nonNull)
    .findFirst();
if (authorization.isPresent()) {
  return authorization.get();
}
```

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ All notable changes to this project will be documented in this file.
 
 - `jibBuildTar` task to build an image tarball at `build/jib-image.tar`, which can be loaded into docker using `docker load` ([#514](https://github.com/GoogleContainerTools/jib/issues/514))
 
+- For Docker Hub, also tries registry aliases when getting a credential from the Docker config
```
coollog: small nit: for consistency, we should keep the entries with no empty lines inbetween (line 8 should not actually be there either)

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAliasGroup.java
chanseokoh: Yeah, probably even do `.findFirst().orElse(null)`;

PR #609, opened 2018-07-13
Apply Maven SettingsDecrypter for encrypted server details

### LINKED ISSUE #592: Maven Encryption does not work
**Description of the issue**:
When credentials in settings.xml are encrypted, jib throws a 401 error, but other tools (including https://github.com/jelmerk/maven-settings-decoder) show the proper password.

If I leave password in plaintext in settings.xml it works.

**Expected behavior**:

Credentials should be decrypted, I expect they are being passed as is.

**Steps to reproduce**: 

```
mvn --encrypt-master-password <some-string>
# save output to settings-security.xml per docs
mvn --encrypt-password <docker registry password>
# save output to settings.xml server section per docs
mvn compile jib:build
#

```

**Environment**: 
```
Apache Maven 3.5.2 (138edd61fd100ec658bfa2d307c43b76940a5d7d; 2017-10-18T03:58:13-04:00)
Maven home: /Users/eddie/.m2/wrapper/dists/apache-maven-3.5.2-bin/28qa8v9e2mq69covern8vmdkj0/apache-maven-3.5.2
Java version: 1.8.0_172, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/jre
```

```xml

			<plugin>
				<version>0.9.4</version>
				<configuration>
					<to>
						<image>registry.hub.docker.com/eddiewebb/blueskygreenbuilds-demo</image>
					</to>
					<container>
						<jvmFlags>
							<jvmFlag>-Dcircle_build_num=99</jvmFlag>
							<jvmFlag>-Dcircle_commit=1234abcdef</jvmFlag>
							<jvmFlag>-Dcircle_user=eddiewebb</jvmFlag>
							<jvmFlag>-Dcircle_repo=demo-repo</jvmFlag>
							<jvmFlag>-Dcircle_workflow_guid=1234</jvmFlag>
							<jvmFlag>-Dvcap.application.name=blueskygreenbuilds-test</jvmFlag>
						</jvmFlags>
						<ports>
							<port>8080</port>
						</ports>
					</container>
				</configuration>
			</plugin>
```

~/.m2/settings.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>

<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
    <servers>
        <server>
            <id>registry.hub.docker.com</id>
            <username>eddiewebb</username>
            <password>{output from --encrypt-password above}</password>
        </server>
    </servers>
</settings>
```

~/.m2/settings-security.xml
```xml
<settingsSecurity>
  <master>{output from --encrypt-master-password above}</master>
</settingsSecurity>
```

**Log output**: 

>[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.4:build (default-cli) on project blueskygreenbuilds: Build image failed, perhaps you should make sure your credentials for 'registry.hub.docker.com' are set up correctly: Unauthorized for registry.hub.docker.com/user/imagename: 401 Unauthorized
[ERROR] {"details":"incorrect username or password"}

**Additional Information**: 

To debug I first ran maven with -X, and it confirms the existence of `env.maven_security_master={output from --encrypt-master-password}`

I then used https://github.com/jelmerk/maven-settings-decoder to decrypt maven credentials.

```
./settings-decoder/bin/settings-decoder -f ~/.m2/settings.xml -s ~/.m2/settings-security.xml 
Master password is : yep-thats-what-I-entered-above
-------------------------------------------------------------------------
Credentials for server registry.hub.docker.com are :
Username : eddiewebb
Password : yep-that-also-is-my-correct-password-for-docker-that-works-in-plaintext

```

There are no special characters in my password (letters and numbers) and no special characters in the generated values (I also tried several iterations)

I have seen the mvn release plugin require certain version, so perhaps some dependency of jib is not where it needs to be?

briandealwis:
My assumption here is that a decryption error in `MavenSettingsServerCredentials` should fail the build rather than either return the still-encrypted password or return `null` (i.e., no configured credentials).

Unsure about `MavenSettingsServerCredentials` throwing a `MojoExecutionException` directly, but it seemed unhelpful to create a new exception class that would be immediately turned into a `MojoExecutionException`.

```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.7-SNAPSHOT:build (default-deploy) on project first: Unable to decrypt settings for registry.hub.docker.com: [ERROR] Failed to decrypt password for server registry.hub.docker.com: org.sonatype.plexus.components.sec.dispatcher.SecDispatcherException: java.io.FileNotFoundException: /Users/bsd/.m2/settings-security.xml (No such file or directory) @ server: registry.hub.docker.com -> [Help 1]
```

Fixes #592


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -32,31 +38,50 @@
...
 
+    if (settingsDecrypter != null) {
+      SettingsDecryptionRequest request = new DefaultSettingsDecryptionRequest(registryServer);
+      SettingsDecryptionResult result = settingsDecrypter.decrypt(request);
+      // un-encrypted passwords are passed through, so a problem indicates a real issue
+      for (SettingsProblem problem : result.getProblems()) {
+        if (problem.getSeverity() == SettingsProblem.Severity.ERROR
+            || problem.getSeverity() == SettingsProblem.Severity.FATAL) {
+          throw new MojoExecutionException(
+              "Unable to decrypt settings for " + registry + ": " + problem);
```
chanseokoh: I think "... decrypt password for ..." is more informative to actual users and has the potential to make them actions on their own.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -44,7 +44,6 @@
     name = BuildDockerMojo.GOAL_NAME,
     requiresDependencyResolution = ResolutionScope.RUNTIME_PLUS_SYSTEM)
 public class BuildDockerMojo extends JibPluginConfiguration {
-
```
coollog: nit: a line removed (we're following a convention of empty line after class definition for this codebase)

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -164,6 +166,8 @@ void handleDeprecatedParameters(BuildLogger logger) {
   @Parameter(defaultValue = "${project.basedir}/src/main/jib", required = true)
   private String extraDirectory;
 
+  @Nullable @Component protected SettingsDecrypter settingsDecrypter;
```
coollog: Would this ever be `null`?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -32,31 +38,50 @@
...
+      for (SettingsProblem problem : result.getProblems()) {
+        if (problem.getSeverity() == SettingsProblem.Severity.ERROR
+            || problem.getSeverity() == SettingsProblem.Severity.FATAL) {
+          throw new MojoExecutionException(
+              "Unable to decrypt password for " + registry + ": " + problem);
+        }
+      }
+      if (result.getServer() != null) {
+        registryServer = result.getServer();
+      }
```
coollog: If it is `null`, should we error?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
briandealwis: I'm not sure.   nullaway complains without it.  I did some searching in the Maven docs but didn't find anything authoritative.

The [Github Maven Plugins](https://github.com/github/maven-plugins/blob/master/github-core/src/main/java/com/github/maven/plugins/core/GitHubProjectMojo.java#L168) look up the SettingsDecrypter through the plexus container and guard against a `ComponentLookupException`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
briandealwis: `SettingsDecrypter` and `SettingsDecryptionResult` are under-documented and does not discuss the possible meanings of the return results.  `SettingsDecryptionResult#getServers()` does note that:
```java
    /**
     * Gets the decrypted servers.
     *
     * @return The decrypted server, can be empty but never {@code null}.
     */
    List<Server> getServers();
```
and the implementation of `getServer()` returns the first element from `getServers()` and `null` if there are no servers.

By this point, we have no errors reported from the decryption request, but we have no servers.  So one interpretation is: no decryption was required, so use the original server.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
chanseokoh: Just a native question, is it possible fix encrypted passwords and plain passwords (for different servers) in `settings.xml`? If that's true, I just want to make sure `settingsDecrypter` handles plain passwords nicely. (I do believe it will.)

briandealwis: I think @chanseokoh is asking about `~/.m2/settings.xml` having both encrypted and unencrypted passwords.  The default `SettingsDecryptor` passes-through unencrypted passwords.

The `SettingsDecrypter` could theoretically be replaced.  In the worse case, where the `SettingsDecrypter` is `null` but we have encrypted passwords, we will attempt to authenticate  with the encrypted password bytes.  But since it was encrypted we won't have divulged anything in the clear.  There's no solid way to identity an encrypted vs unencrypted password though.

coollog: Mmhmm, I was mostly trying to get at the user experience side, like:

1. User sets encrypted password in Maven settings
1. We try to decrypt, but `SettingsDecrypter` is `null`, so the password wasn't decrypted
1. Users sees `401 Unauthorized` and believes their password is wrong

Whereas, I think the better UX would be:

1. User sets encrypted password in Maven settings
1. We try to decrypt, but `SettingsDecrypter` is `null`, so the password wasn't decrypted, and we give an error
1. Users sees `failed to decrypt password` (or similar)

briandealwis: The default encryption approach seems to surround the passwords with braces.

We could pass in the `mavenBuildLogger` here and if `settingsDecrypter == null` and the password starts and ends with braces then issue a warning.

coollog: In `MavenSettingsServerCredentials#retrieve` right? I might be misunderstanding how `SettingsDecrypter` works, but is there any harm in giving the warning just if `settingsDecrypter == null`?

eddiewebb:
Just a note @briandealwis, according to Maven docs it is valid to put notes inside the password element, but outside the curly braces. So it is not a safe assumption that encrypted values will start with {

`<password>Eddie reset this on 2018-07-13 {hsg36sheyeue6duduesy}</password>`. 

So you'd have to anticipate that in any 'is it encrypted' logic

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -82,11 +96,32 @@ RegistryCredentials retrieve(@Nullable String registry) throws MojoExecutionExce
...
 
     return new RegistryCredentials(
         CREDENTIAL_SOURCE,
         Authorizations.withBasicCredentials(
             registryServer.getUsername(), registryServer.getPassword()));
   }
+
+  // pattern cribbed directly from
+  // https://github.com/sonatype/plexus-cipher/blob/master/src/main/java/org/sonatype/plexus/components/cipher/DefaultPlexusCipher.java
+  private static final Pattern ENCRYPTED_STRING_PATTERN =
```
coollog: In this codebase, we're following a convention of organizing classes in the order:

1. static->non-static
1. members->inner class->constructor->methods
1. public->private
1. final->nonfinal

So a `ENCRYPTED_STRING_PATTERN` would be under `CREDENTIAL_SOURCE`, `isEncrypted` would be under `ENCRYPTED_STRING_PATTERN`.

coollog:
Oh I forgot to mention to include the CHANGELOG entry too.

PR #610, opened 2018-07-13
Links to Jib as a library for Java issue in README.

coollog:
So that users can find and watch the relevant issue easily for updates.


README.md
```diff
@@ -15,7 +15,7 @@ Jib builds Docker and OCI images for your Java applications and is available as
 [Maven](https://maven.apache.org/): See documentation for [jib-maven-plugin](jib-maven-plugin).\
 [Gradle](https://gradle.org/): See documentation for [jib-gradle-plugin](jib-gradle-plugin).
 
-*Jib as a container-building library for Java is work-in-progress. Watch for updates.*
+*Jib as a container-building library for Java is work-in-progress. [Watch for updates.](/../../issues/337)*
```
TadCordle: Might be better to use absolute links so they don't die if we move files around for the docs website update?

Issue #599, opened 2018-07-13
Allow authorization also against insecure registries (at least for testing)


Hi-Fi:
**Description of the issue**:
Currently if registry offers only insecure access, there's no way to authenticate against it (https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java#L165). It would be good for testing to be able to authenticate over http (partly because of #543).

**Expected behavior**:
Parameter where user acknowledges that it's very bad idea to send credentials over http, but still allowing that.

**Steps to reproduce**: 
Create local registry with authentication and without TLS.

**Environment**: 
Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T22:39:06+03:00)
Java version: 1.8.0_131, vendor: Oracle Corporation
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 7", version: "6.1", arch: "amd64", family: "windows"
jib-plugin version: 0.9.6

**`jib-maven-plugin` Configuration**: 
```xml
<plugin>
	<groupId>com.google.cloud.tools</groupId>
	<artifactId>jib-maven-plugin</artifactId>
	<version>${jib.version}</version>
	<configuration>
		<from>
			<image>${jib.fromImage}</image>
		</from>
		<to>
			<image>${jib.toImage}</image>
		</to>
		<container>
			<ports>
				<port>${jib.applicationPort}</port>
			</ports>
		</container>
		<allowInsecureRegistries>${jib.allowInsecureRegistries}</allowInsecureRegistries>
	</configuration>
	<executions>
		<execution>
			<phase>compile</phase>
			<goals>
				<goal>build</goal>
			</goals>
		</execution>
	</executions>
</plugin>
```

**Log output**: 
Without allowInsecureRegitstries:
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.6:build (default) on project jib-demo-project: Build image failed, perhaps you should use a registry that supports HTTPS or set the configuration parameter 'allowInsecureRegistries': Only secure connections are allowed, but tried to reach URL http://artifactory/project/baseImage/manifests/latest -> [Help 1]
```
With allowInsecureRegitstries:
```
[...]

PR #613, opened 2018-07-13
Add error message for credentials not being sent over HTTP

TadCordle:
Part of #599

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulSuggestions.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulSuggestions.java
@@ -96,6 +96,10 @@ public class HelpfulSuggestions {
     return suggest("make sure your credentials for '" + registry + "' are set up correctly");
   }
 
+  public String forCredentialsNotSent() {
+    return suggest("use a registry that supports HTTPS so credentials can be sent safely");
+  }
+
   public String forDockerContextInsecureRecursiveDelete(String directory) {
     return suggest("clear " + directory + " manually before creating the Docker context");
   }
@@ -106,6 +110,11 @@ public class HelpfulSuggestions {
 
   public String forDockerNotInstalled() {
     return suggest("make sure Docker is installed and you have correct privileges to run it");
+  }
+
+  public String forInsecureRegistry() {
+    return suggest(
+        "use a registry that supports HTTPS or set the configuration parameter 'allowInsecureRegistries'");
   }
 
   /**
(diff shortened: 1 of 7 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulSuggestions.java
```diff
@@ -108,6 +112,11 @@ public String forDockerNotInstalled() {
     return suggest("make sure Docker is installed and you have correct privileges to run it");
   }
 
+  public String forInsecureRegistry() {
```
coollog: This seems like the more useful message, but I don't see a reference to this?

TadCordle: This one was already here, just moved it because it was all the way at the bottom of the file. Is this actually more useful? I thought the problem happened even if you set `allowInsecureRegistries`.

briandealwis: It's called slightly further down in [BuildStepRunner](https://github.com/GoogleContainerTools/jib/pull/613/files#diff-644d63cbab71bf85ab42dba1ce499fe1R254)

TadCordle: Yeah, I think that's the one that happens if `allowInsecureRegistries` isn't set and it tries to do an HTTP fallback. But the error we're throwing here is when HTTP fallback proceeds, but credentials are required.

coollog: Ah right, I should read more before commenting haha.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/HelpfulSuggestions.java
```diff
@@ -96,6 +96,10 @@ public String forCredentialsNotCorrect(String registry) {
     return suggest("make sure your credentials for '" + registry + "' are set up correctly");
   }
 
+  public String forCredentialsNotSent() {
+    return suggest("use a registry that supports HTTPS so credentials can be sent safely");
```
coollog: I think prefixing with an explanation before the suggestion would be good. Something along the lines of:

`The registry requires credentials but credentials were not sent because the connection was over HTTP.`

TadCordle: Hmm should that prefix go in the new exception instead of `HelpfulSuggestions`, or is it too specific for that?

Issue #543, opened 2018-07-11
Support for self signed certificates


Hi-Fi:
**Description of the issue**:
When (internal) Docker registry is using self signed certificates (or certificates signed with corporation's own root certificate, that's not trusted by Java), jib tries immediately http-connection (according https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java#L219), that's not available in registry. 

Current implementation throws error:

> [ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.2:build (default-cli) on project jib-test-project: Build image failed, perhaps you should use a registry that supports HTTPS or set the configuration parameter 'allowInsecureRegistries': Only secure connections are allowed, but tried to reach URL http://openshift-redhat-registry/v2/testProject/test-container/manifests/latest -> [Help 1]

**Expected behavior**:
`allowInsecureRegistries` -parameter should allow also https connections with self signed certificates, and not directly drop to http.

**Steps to reproduce**: 
mvn compile jib:build with configuration set to point to registry using self signed certificates.

**Environment**: 
Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T22:39:06+03:00)
Java version: 1.8.0_131, vendor: Oracle Corporation
Default locale: en_US, platform encoding: Cp1252
OS name: "windows 7", version: "6.1", arch: "amd64", family: "windows"

**`jib-maven-plugin` Configuration**: 
```xml
<configuration>
	<from>
		<image>openshift-redhat-registry/testProject/test-container</image>
	</from>
	<to>
		<image>openshift-redhat-registry/testProject/${project.artifactId}:${project.version}</image>
	</to>
	<container>
		<jvmFlags>
			<jvmFlag>-Xms512m</jvmFlag>
		</jvmFlags>
	</container>
</configuration>
```

**Log output**: 
> [ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.2:build (default-cli) on project jib-test-project: Build image failed, perhaps you should use a registry that supports HTTPS or set the configuration parameter 'allowInsecureRegistries': Only secure connections are allowed, but tried to reach URL http://openshift-redhat-registry/v2/testProject/test-container/manifests/latest -> [Help 1]

**Additional Information**: 
The actual error causing change to http is:
```
[...]

PR #618, opened 2018-07-14
Add instructions for self signed certicate

velo:
Create doc with the workaround discuessed on https://github.com/GoogleContainerTools/jib/issues/543#issuecomment-404347465

### DIFF
--- a/docs/self_sign_cert.md
+++ b/docs/self_sign_cert.md
@@ -1,0 +1,53 @@
+# Accessing private docker registry with self-signed certificate
+
+Currently, `jib` does not support docker registries with self-signed `https` certificate.
+
+Jib uses the JVM's list of approved CA Certificates to validate SSL certificates. The following instructions describe how to add a registry's self-signed certificate to the JVM's approved CAs.
+
+The certificate will be trusted at the JRE level, affecting all Java applications running on it. You will also need to re-import the certificate when you use a different JRE or upgrade it.
+
+## Using KeyStore Explorer
+
+The easiest way to import the self-signed certificate into JVM is using the [KeyStore Explorer](http://keystore-explorer.org/).
+
+KeyStore Explorer is an open source GUI replacement for the Java command-line utilities keytool and jarsigner. KeyStore Explorer presents their functionality, and more, via an intuitive graphical user interface.
+
+### Installation
+
+Download and install KeyStore Explorer from [official website](http://keystore-explorer.org/downloads.html)
+
+### Indentify java runtime being used by build tool
+
+#### Maven
+
+Run `mvn --version` and take note on java runtime:
+
+```shell
+$ mvn --version
+Apache Maven 3.5.4 (1edded0938998edf8bf061f1ceb3cfdeccf443fe; 2018-06-18T06:33:14+12:00)
+Maven home: /usr/local/Cellar/maven/3.5.4/libexec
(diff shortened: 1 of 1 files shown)


velo:
> just adding the default password

What password? Not sure what you mean.

Hi-Fi:
Default password for Java keystore, that is "changeit". I think it should be in document.

velo:
Mine was not password protected, which step was that?

kaikramer:
To help with the confusion about the password protection of cacerts: Since version 5.2 KeyStore Explorer automatically tries to open cacerts with the default password ("changeit"). It only prompts for a password if this fails.

docs/self_sign_cert.md
```diff
@@ -0,0 +1,51 @@
+# Accessing private docker registry with self signed certificate
+
+Currently, `jib` do not support docker registries with self signed `https` certificate.
```
briandealwis: do not -> does not
self signed certificate -> self-signed certificates

docs/self_sign_cert.md
```diff
@@ -0,0 +1,51 @@
...
+
+The only way to get `jib` working is to import the self signed certificate into jvm `CA Certificates Keystore`.
+
+## Using KeyStore Explorer
+
+The easiest way to import the self signed certificate into jvm is using the [KeyStore Explorer](http://keystore-explorer.org/).
```
briandealwis: the self signed -> a self-signed
jvm -> JVM

docs/self_sign_cert.md
```diff
@@ -0,0 +1,51 @@
+# Accessing private docker registry with self signed certificate
+
+Currently, `jib` do not support docker registries with self signed `https` certificate.
+
+The only way to get `jib` working is to import the self signed certificate into jvm `CA Certificates Keystore`.
```
briandealwis: How about:
> Jib uses the JRE's list of approved _CA Certificates_ to validate SSL certificates.  The following instructions describe how to add a registry's self-signed certificate to the JRE's approved CAs.

docs/self_sign_cert.md
```diff
@@ -0,0 +1,51 @@
...
+Java version: 1.8.0_172, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/jre
+Default locale: en_NZ, platform encoding: UTF-8
+OS name: "mac os x", version: "10.13.6", arch: "x86_64", family: "mac"
+```
+
+On this example the `java runtime` is `/Library/Java/JavaVirtualMachines/jdk1.8.0_172.jdk/Contents/Home/jre`.
+
+### Import certificate
+
+* Launch `KeyStore Explorer`
```
briandealwis: My 2¢: it would be easier to have the user use _Open an existing KeyStore_ and navigate to the `cacerts` file (with default password `changeit`).

Then there are two approaches:
  - if the user has the self-signed certificate, they can import it.
  - if they don't have the self-signed certificate, they can import it from the running service with the _Examine SSL_ option. It allows importing the certificate directly into the keystone.

docs/self_sign_cert.md
chanseokoh: How about also adding

"The certificate will be trusted at the JRE level, affecting all Java applications running on it. You will also need to re-import the certificate when you use a different JRE or upgrade it."

docs/self_sign_cert.md
velo: I guess this import process has multiple possible ways to be done.... I describe the one I used

velo:
@chanseokoh @briandealwis changes applied

PR #627, opened 2018-07-16
Fix writing multi-byte characters to tar archive

### LINKED ISSUE #626: Cannot finish task jibDockerBuild use 2byte character in container args
Use Gradle plugin.  `jib.container.args = ['日本語']`

```
$ ./gradlew clean jibDockerBuild
```

Cannot finish task.

```groovy
plugins {
    id 'org.jetbrains.kotlin.jvm' version '1.2.50'
    id "application"
    id 'com.google.cloud.tools.jib' version '0.9.6'
}

group 'com.github.sakebook'
version '0.0.1-SNAPSHOT'
mainClassName = 'com.github.sakebook.word2vec.Word2VecKt'

jar {
    manifest {
        attributes 'Main-Class': 'com.github.sakebook.word2vec.Word2VecKt'
    }
    from { configurations.compile.collect { it.isDirectory() ? it : zipTree(it) } }
}

jib {
    to {
        image = 'sakebook/word2vec-deeplearning4j-kotlin'
        auth {
            username = USERNAME // Defined in 'gradle.properties'.
            password = PASSWORD
        }
    }
    container {
        args = ['日本語']
    }
}


repositories {
    mavenCentral()
}

dependencies {
    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
    compile "org.deeplearning4j:deeplearning4j-core:0.9.1"
    compile "org.deeplearning4j:deeplearning4j-nlp:0.9.1"
    compile "org.nd4j:nd4j-native-platform:0.9.1"
    compile "org.slf4j:slf4j-simple:1.7.25"
    compile "org.slf4j:slf4j-api:1.7.25"
}

compileKotlin {
    kotlinOptions.jvmTarget = "1.8"
}
compileTestKotlin {
    kotlinOptions.jvmTarget = "1.8"
}
```

**Log output**: 
```
word2vec$ ./gradlew clean jibDockerBuild
warning: Could not find build output directory 'xxxxxx/word2vec/build/classes/java/main'

Containerizing application to Docker daemon as sakebook/word2vec-deeplearning4j-kotlin...

Building resources layer...
Building classes layer...
Finalizing...
Loading to Docker daemon...
<==========---> 83% EXECUTING [1h 19m 22s]
```

**Additional Information**: 

My application wrote by Kotlin.

TadCordle:
Fixes #626. Looks like we were using string length instead of byte array length.


loosebazooka:
can we use codePointCount?

TadCordle:
So it looks like the main problem has been fixed, but using non-ASCII tar archive entry header names on windows might still be sketchy (unicode archive contents should work now, though). I'll keep looking into it, but feel free to review this now so we can get the fix out.

coollog:
@briandealwis `Blob` is essentially our own minimal interface that is essentially a `ByteSource`, but you make a good point - can you file a separate issue for that since it's not directly related to this fix?

coollog:
@TadCordle oh we should add a CHANGELOG entry for this

chanseokoh:
Cool, just in case, have you tested `dockerBuild` and `buildTar`?

TadCordle:
Yeah, tested both with the same args used in the original issue.

PR #650, opened 2018-07-18
Look up cred helper additionally with possible suffixes

chanseokoh:
Part of #586.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java
@@ -19,12 +19,13 @@ package com.google.cloud.tools.jib.registry.credentials.json;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.google.cloud.tools.jib.json.JsonTemplate;
 import com.google.common.annotations.VisibleForTesting;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.function.Predicate;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 
 /**
@@ -97,20 +98,13 @@ public class DockerConfigTemplate implements JsonTemplate {
    */
   @Nullable
   public String getAuthFor(String registry) {
-    Predicate<String> exactMatch = registry::equals;
-    Predicate<String> withHttps = ("https://" + registry)::equals;
-    Predicate<String> startsWith = name -> name.startsWith(registry);
-    Predicate<String> startsWithAndWithHttps = name -> name.startsWith("https://" + registry);
-    AuthTemplate authTemplate =
-        getAuthTemplate(Arrays.asList(exactMatch, withHttps, startsWith, startsWithAndWithHttps));
-
+    AuthTemplate authTemplate = getAuthTemplate(getRegistryMatchersFor(registry));
     return authTemplate != null ? authTemplate.auth : null;
   }
 
   /** Returns the first {@link AuthTemplate} matching the given predicates (short-circuiting). */
-  private AuthTemplate getAuthTemplate(List<Predicate<String>> registryMatches) {
+  private AuthTemplate getAuthTemplate(Stream<Predicate<String>> registryMatches) {
     return registryMatches
-        .stream()
         .map(this::getAuthTemplate)
         .filter(Objects::nonNull)
         .findFirst()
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java
```diff
@@ -97,48 +96,59 @@
...
+  private Stream<Predicate<String>> getRegistryMatchersFor(String registry) {
+    Predicate<String> exactMatch = registry::equals;
+    Predicate<String> withHttps = ("https://" + registry)::equals;
+    Predicate<String> startsWith = name -> name.startsWith(registry);
+    Predicate<String> startsWithAndWithHttps = name -> name.startsWith("https://" + registry);
+    return Stream.of(exactMatch, withHttps, startsWith, startsWithAndWithHttps);
+  }
+
+  /** Returns the first value matching the given key predicates (short-circuiting). */
+  private static <K, T> T findFirstInMapByKey(Map<K, T> map, Stream<Predicate<K>> keyMatches) {
```
coollog: For consistency in the code base, we're using an ordering for the elements of a class of:

```
static->non-static
fields->inner class->constructor->methods
public->private
final->nonfinal
```

So the private static methods would go up top below the inner class

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java
```diff
@@ -97,48 +96,59 @@
...
       }
     }
-    if (credHelpers.containsKey(registry)) {
-      return credHelpers.get(registry);
-    }
-    return null;
+    return findFirstInMapByKey(credHelpers, getRegistryMatchersFor(registry));
+  }
+
+  private Stream<Predicate<String>> getRegistryMatchersFor(String registry) {
```
coollog: Doesn't have to be part of this PR, but given how much logic there is now in this class, it might make sense to break the logic out into a separate `DockerConfig` class that performs the logic on `DockerConfigTemplate`, since `...Template` are mostly to be a Java-mapped JSON template.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java
```diff
@@ -97,48 +96,59 @@
...
    * @param registry the registry to get the credential helpers for
    * @return {@code credsStore} if {@code registry} is present in {@code auths}; otherwise, searches
    *     {@code credHelpers}; otherwise, {@code null} if not found
    */
   @Nullable
   public String getCredentialHelperFor(String registry) {
     if (credsStore != null) {
-      // The registry could be prefixed with the HTTPS protocol.
-      if (auths.containsKey(registry) || auths.containsKey("https://" + registry)) {
+      // The registry could be prefixed with the HTTPS protocol and/or have suffixes.
```
coollog: Should we also handle aliases here?

chanseokoh: I don't think so. Aliases would be handled at a higher-level, from callers.

If you are talking about suffixes, `startsWith()` will cover both the exact and suffix matches. However, I just found a pitfall (https://github.com/GoogleContainerTools/jib/issues/586#issuecomment-405960261), so I think I'll just reuse `getRegistryMatchersFor()`, which actually simplifies the code.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java
chanseokoh: +1 makes sense. Filed https://github.com/GoogleContainerTools/jib/issues/651

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java
```diff
@@ -118,37 +137,19 @@ public String getAuthFor(String registry) {
...
+    List<Predicate<String>> registryMatchers = getRegistryMatchersFor(registry);
+    if (credsStore != null && findFirstInMapByKey(auths, registryMatchers) != null) {
+      return credsStore;
     }
-    return findFirstInMapByKey(credHelpers, getRegistryMatchersFor(registry));
+    return findFirstInMapByKey(credHelpers, registryMatchers);
   }
 
-  private Stream<Predicate<String>> getRegistryMatchersFor(String registry) {
+  private List<Predicate<String>> getRegistryMatchersFor(String registry) {
```
coollog: Hmm what was the reason for changing this to a List?

chanseokoh: Can't reuse the stream once it is closed with `Steam.firstFirst()`. That is,

```java
Stream<...> matchers = getRegistryMatchersFor(...);
findFirstInMapByKey(..., matchers);  // first call OK
findFirstInMapByKey(..., matchers);  // <-- throws an exception
```

coollog: Ah right

PR #652, opened 2018-07-18
Document alternative trust-store option for self-signed certificates

briandealwis:
And improve the documentation.

Document can be viewed [here](
https://github.com/GoogleContainerTools/jib/blob/ca877775f35817215a6ee69fc5f7f19e269e7296/docs/self_sign_cert.md)


docs/self_sign_cert.md
```diff
@@ -31,23 +22,88 @@ Default locale: en_NZ, platform encoding: UTF-8
...
+
+#### Option 1: Create a New Trusted Keystore
+
+This option creates a _new_ list of CA Certificates and configures your build tool to use ths new list as the JRE's list of approved CA certificates, called the _trust store_.
+
+Within _KeyStore Explorer_, select _File > Save As..._ and save the new keystore file as a _JKS_ file alongside your build location. You will be prompted for a password; we use `password` in the examples below.
+
+##### Maven
+
+The following snippet shows how to configure Maven to use this new keystore file. You can configure the registry's username and password with [the `~/.m2/settings.xml` mechanism](https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/README.md#using-maven-settings).
```
chanseokoh: > You can configure the registry's username and password with [the `~/.m2/settings.xml` mechanism]

This could be helpful, but I also think out of scope. I'm mostly concerned the user might be confused this a-bit-independent registry auth topic with this SSL cert setup. Like, if these username and password have something to do with the `-Djavax.net.ssl.trustStorePassword=password`.

docs/self_sign_cert.md
```diff
@@ -31,23 +22,88 @@ Default locale: en_NZ, platform encoding: UTF-8
...
+
+```shell
+$ ./mvnw -Djavax.net.ssl.trustStore=path/to/keystore.jks \
+  -Djavax.net.ssl.trustStorePassword=password \
+  -Dimage=<host>:<port>/<image> jib:build
+```
+
+##### Gradle
+
+You may need to explicitly configure the auth [`username` and `password`](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#using-specific-credentials).
```
chanseokoh: Ditto. This could be helpful, but I also think out of scope. I'm mostly concerned the user might be confused this a-bit-independent registry auth topic with this SSL cert setup.

docs/self_sign_cert.md
```diff
@@ -31,23 +22,88 @@ Default locale: en_NZ, platform encoding: UTF-8
...
+
+```shell
+$ ./gradlew jib \
+  -Djavax.net.ssl.trustStore=path/to/keystore.jks \
+  -Djavax.net.ssl.trustStorePassword=password
+```
+
+#### Option 2: Modify the JRE `cacerts`
+
+The other approach modifies the JRE's list of CA Certificates to include the registry's self-signed certificate.  The certificate will be trusted at the JRE level, affecting all Java applications running on it. You must re-import the certificate when you upgrade or update to a new JRE.
```
chanseokoh: "upgrade or update to a new JRE" seems redundant. Did you mean "use or update to a new JRE"?

docs/self_sign_cert.md
```diff
@@ -31,23 +22,88 @@ Default locale: en_NZ, platform encoding: UTF-8
...
+
+## 4. Save the CA Certificates
+
+Now we save the updated keystore. We can either save to a new keystore and configure our build's JVM to use this new keystore as a _trusted keystore_, or modify the JRE's list of CA Certificates.
+
+#### Option 1: Create a New Trusted Keystore
+
+This option creates a _new_ list of CA Certificates and configures your build tool to use ths new list as the JRE's list of approved CA certificates, called the _trust store_.
+
+Within _KeyStore Explorer_, select _File > Save As..._ and save the new keystore file as a _JKS_ file alongside your build location. You will be prompted for a password; we use `password` in the examples below.
```
chanseokoh: Not sure what is "your build location", like `target/` or `build/`? That would be cleaned a lot.

docs/self_sign_cert.md
briandealwis: It turns out this step is pretty much essential as they'll otherwise need to configure the registry with docker's insecure registries dance.

briandealwis: I think my language is confusing here: _"You can configure your username and password for the registry with the …"_

chanseokoh: Okay, let's do whatever that will make this less confusing. For example, moving the registry part after the shell command example, starting with "Don't forget get that you will also need to ...", which is too verbose, but you know my point.

Issue #582, opened 2018-07-12
Pushing to registry is timing out


alexanderzimmer:
**Description of the issue**:
Our docker-registry is sometimes a little bit slow, so the building-process inside our CI-pipeline fails with the message `Build image failed: Read timed out`. The timeout i selected inside my settings.xml for this server is not respected.

**Expected behavior**:
Respecting the Timeout inside the servers-tag in settings.xml

**Steps to reproduce**:
Try to push an image to an slow docker-registry. Then set the timeout for this registry inside your settings.xml and test ist again. You will see that the timeout is not respected.

Tested with version 0.9.6

coollog:
@alexanderzimmer Thanks for reporting this issue! We'll look into using the timeout from the `settings.xml`. Is your timeout configuration in the old or new wagon format? (https://maven.apache.org/guides/mini/guide-http-settings.html#Connection_Timeouts)

alexanderzimmer:
I've tried both and ended up using:
```
...
            <configuration>
                <httpConfiguration>
                    <all>
                        <connectionTimeout>60000</connectionTimeout>
                        <readTimeout>60000</readTimeout>
                    </all>
                </httpConfiguration>
            </configuration>
...
```

But it still does not work.

coollog:
@alexanderzimmer Yea it will not work in the current state because we do not read timeout configuration from Maven settings. We might support the format in your comment. 

@GoogleContainerTools/java-tools Any ideas on other ways of solving this issue?

patflynn:
I'm not convinced that the maven http setting configuration should be
applied to the container registry interactions.  Perhaps we should provide
our own configuration for this?

chanseokoh:
I think it makes sense, because we read the `<servers>` info from `settings.xml` and get the container registry info from there (including their username and password), and it looks like the HTTP (tmieout) configuration is under `<servers>`.

coollog:
We do have to think about what to do on the Gradle side too to keep the experience consistent. Another option is just to have a much higher timeout globally without configuration.

patflynn:
@coollog what's it set to now? 

So the container registry urls are specified in <servers> in settings.xml?  Are we using an http wagon implementation for our http client?

coollog:
Since we use the Google HTTP client, the default timeout is 20 seconds.

chanseokoh:
> So the container registry urls are specified in in settings.xml?

[...]

PR #656, opened 2018-07-18
Enable jib.httpTimeout system property for HTTP connection and read timeout

chanseokoh:
Part of #582.

Decided to read the system property in `RegistryAuthenticator` and `RegistryEndpointCaller`.


chanseokoh:
Hold on. Chatted with @coollog, and I'll rework this.

chanseokoh:
I think this is good for review.

The next step after this PR is to check the system property value at the frontend to print out warnings, if necessary.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -69,6 +69,11 @@
     }
   }
 
+  private static int getHttpTimeout() {
+    int httpTimeout = Integer.getInteger("jib.httpTimeout", 20000);
```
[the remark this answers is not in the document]
chanseokoh: The Google HTTP client doesn't define the public static constant, so I'll pursue the latter approach.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -103,6 +103,15 @@ public static Initializer initializer(String serverUrl, String repository) {
     return new Initializer(serverUrl, repository);
   }
 
+  @Nullable
+  private static Integer getHttpTimeout() {
```
coollog: Is using `null` a common way to do this? If not, I was thinking maybe use negative to indicate absence?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -103,6 +103,15 @@ public static Initializer initializer(String serverUrl, String repository) {
...
+    Integer httpTimeout = Integer.getInteger("jib.httpTimeout");
+    if (httpTimeout != null && httpTimeout >= 0) {
```
coollog: I think we could do just `return Integer.getInteger("jib.httpTimeout", null)` since in the negative case, we probably want to error (thrown by `HttpRequest#setConnectTimeout`)

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
chanseokoh: I think it's not uncommon to use `null` to indicate a field is unset or uninitialized, or something is absent. Using a negative value works too. However, it won't be distinguishable from the user wrongly giving a negative value. That will matter if we error out on a negative timeout value as you said.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
chanseokoh: I was thinking showing a warning with a negative value, instead of erroring out. Do you think it's not good?

Anyways, if we decide to error out, I'll probably detect it and fail at the frontends with better error messages, rather than just displaying up an unfriendly stack trace from `Preconditions.checkArgument()` with no message.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
coollog: Okay sounds good :+1: let's add a javadoc explicitly mentioning that `null` means the caller should use some default.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
coollog: I think either way is good. I lean towards the errorring since the case I could think of where negative could show up is if the user may have meant to type another timeout and accidentally put a minus sign in front - in which case, a warning may be lost in the logs to a different error when the connection fails.

chanseokoh: I see. I think then the same argument applies to a non-parsable string values. The user could accidentally put any non-digit character, so it also makes sense to error out on non-parsable strings. Going for errorring on both cases?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
chanseokoh: Ah, you weren't talking about a field here. I see what you meant. I'll think about the API-aspect of this method.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
coollog: Yea, I think in the non-parsable version, this already errors, since `Integer#getInteger` would throw a `NumberFormatException`.

chanseokoh: Alright, errorring on both cases. (BTW, `Integer.getInteger()` doesn't throw `NumberFormatException`. `parseInt` does, OTOH.)

chanseokoh: Will update the code tomorrow, but I'm thinking of doing

```java
      Integer httpTimeout = null; // implies using the default timeout
      String timeoutValue = System.getProperty("jib.httpTimeout");
      if (timeoutValue != null) {
        httpTimeout = Integer.parseInt(timeoutValue);
        if (httpTimeout < 0) {
          throw new RegistryAuthenticationFailedException("Negative value of jib.httpTimeout");
        }
      }
      ...
      try (Connection connection = ...) {
        Request.Builder requestBuilder = Request.builder().setHttpTimeout(httpTimeout);
      ...
    } catch (NumberFormatException ex) {
      throw new RegistryAuthenticationFailedException("Non-integer value of jib.httpTimeout");
    }
```

Since all versions of `getInteger()` don't throw `NumberFormatException`, I guess I need to call `parseInt()` after `System.getProperty()`.

chanseokoh: @coollog and I agreed that we make the code simple and not care about the edge cases too much here. If something goes awry, so be it. Instead, we will guard the errors at the plugin frontends.

PR #661, opened 2018-07-19
Wrong gradle option in docs to export docker context

Dviejopomata:
In #499, the option from gradle to export the docker context was changed from **--jib.dockerDir** to **--targetDir**, but it wasn't reflected in the docs.


Dviejopomata:
I signed it

PR #667, opened 2018-07-19
Error on invalid jib.httpTimeout values at plugin front-ends

### LINKED ISSUE #582: Pushing to registry is timing out

chanseokoh:
Fixes #582.


chanseokoh:
You can start reviewing. Once #656 is merged, I'll change the target base for merging.

chanseokoh:
Hmm... I just saw the `.jib.frontend` package in jib-core. I guess the code should go in there.

chanseokoh:
Also, looks like I should throw `MojoExecutionException` for Maven and `GradleException` for Gradle.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SystemPropertyValidator.java
```diff
@@ -0,0 +1,43 @@
...
+   * @param exceptionFactory factory to create an exception with the given description
+   * @param <T> the exception type to throw if invalid values
+   * @throws T if invalid values
+   */
+  public static <T extends Throwable> void checkHttpTimeoutProperty(
+      Function<String, T> exceptionFactory) throws T {
+    try {
+      String value = System.getProperty("jib.httpTimeout");
+      if (value != null && Integer.parseInt(value) < 0) {
+        throw exceptionFactory.apply("negative value of jib.httpTimeout");
```
coollog: How about a wording of "jib.httpTimeout cannot be negative: -1234" (with the negative value)?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SystemPropertyValidator.java
```diff
@@ -0,0 +1,43 @@
...
+      }
+    } catch (NumberFormatException ex) {
+      throw exceptionFactory.apply("non-integer value of jib.httpTimeout");
```
coollog: How about a wording of "jib.httpTimeout must be an integer: notaninteger" (with the attempted value)?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SystemPropertyValidator.java
```diff
@@ -0,0 +1,43 @@
...
+    }
```
coollog: Let's include a private constructor for this static utility class.

coollog:
Cools, we'll be able to do a v0.9.7 release after this is merged.

TadCordle:
@coollog should we get the last bit of https://github.com/GoogleContainerTools/jib/issues/586 in as well?

coollog:
@TadCordle We could probably do the `equals(registry) || startsWith(registry + "/")` since its quick and easy.

chanseokoh:
> @TadCordle We could probably do the `equals(registry) || startsWith(registry + "/")` since its quick and easy.

To clarify, [this is being handled correctly](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/json/DockerConfigTemplate.java#L147) after #650.

PR #671, opened 2018-07-19
Uses matched registry alias for credential helper.

### LINKED ISSUE #651: Refactor DockerConfigTemplate code
https://github.com/GoogleContainerTools/jib/pull/650#discussion_r203412015

coollog:
Credential helpers store credentials for ServerURLs and those credentials are fetched via the exact ServerURL. For example, a Docker config with:

```
	"auths": {
		"https://index.docker.io/v1/": {},
	},
        "credsStore": "gcr"
```

means that even if the original registry was `registry.hub.docker.com`, we need to fetch the credentials using `https://index.docker.io/v1/`.

Also fixes #651 because it needed to.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
@@ -77,6 +77,8 @@ public class DockerConfigCredentialRetriever {
   }
 
   /**
+   * Retrieves credentials for a registry. Tries all possible known aliases.
+   *
    * @return {@link Authorization} found for {@code registry}, or {@code null} if not found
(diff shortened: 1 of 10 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
```diff
@@ -97,20 +99,20 @@ public Authorization retrieve() throws IOException {
   }
 
   @Nullable
-  private Authorization retrieve(DockerConfigTemplate dockerConfigTemplate, String registry) {
+  private Authorization retrieve(DockerConfig dockerConfig, String registryAlias) {
```
TadCordle: Might be worth having a javadoc here even though it's a private method.

Issue #678, opened 2018-07-20
Extreme long parsing time for long repository values in the <to> field


hubermi:
**Description of the issue**:

When defining long docker repository names the parsing process takes extremely long, causing the mvn build job (gradle build job should be affected, too as the root cause is in jib-core) to run with 100% CPU.

**Expected behavior**:
String length should not have a big impact on parsing time.

**Additional Information**:
The root cause is the class com.google.cloud.tools.jib.image.ImageReference;
If calling the parse(String reference) method with a repository value that has 27 characters, it takes about 10 seconds to complete. With 28 characters (like in the example above) it takes about 20 seconds, with 29 characters about 40 seconds ...

coollog:
Hi @hubermi , thanks for reporting the issue! This is related to #323. Labeling this will high priority.

coollog:
The fix will be available in the next release (version `0.9.8`).

coollog:
@hubermi we've released version `0.9.8` and this issue should be fixed in that version.

PR #680, opened 2018-07-20
Fix regex bug

### LINKED ISSUE #323: Using image tag with slashes very slow when parsing.

chanseokoh:
Fixes #323 #678

The Javadoc says
```
<p>A separator is either an underscore, a dot, two underscores, or any number of dashes.
```
So, it did not make sense to allow a zero-length separator, essentially allowing not needing a separator at all.

```java
    ImageReference.parse("gcr.io/qingyangc-sandbox/jibtestimage:askfjl/waegawvw/aweg");
```
The above took 206.050 secs. Now 0.014s.


coollog:
Oh forgot to mention to add a changelog entry too, like `Fixed slow image reference parsing`

Issue #681, opened 2018-07-20
Provide an option to serialize DAG build-step execution


briandealwis:
Users hitting registry authentication/authorization issues find Jib's output text imply that the operations are completed.

```
[INFO] Retrieving registry credentials for registry.hub.docker.com...
[INFO] Getting base image gcr.io/distroless/java...
[INFO] Building dependencies layer…
```

For a good time: `mvn jib:build -Dimage=foo/bar/baz' and try to make sense of the result!

TadCordle:
@briandealwis Is there more to do on this issue or is it fixed by #682?

briandealwis:
I see that image push seems to push up the layers/blobs in parallel.
```
Jul 24, 2018 2:11:40 PM com.google.api.client.http.HttpRequest execute
CONFIG: -------------- REQUEST  --------------
HEAD https://quay.io/v2/briandealwis/jibtest/blobs/sha256:8f125ded1b483229a567e48eece98cc8867c2bee55fb73e9cd65ceecc448b7c7
Accept: 
Accept-Encoding: gzip
Authorization: <Not Logged>
User-Agent: jib 0.9.8-SNAPSHOT jib-maven-plugin Google-HTTP-Java-Client/1.23.0 (gzip)

Jul 24, 2018 2:11:40 PM com.google.api.client.http.HttpRequest execute
CONFIG: -------------- REQUEST  --------------
HEAD https://quay.io/v2/briandealwis/jibtest/blobs/sha256:a988b67f65865e95c004400d61fee34b39651fb3437fedf2e44c9d5f01ff1290
Accept: 
Accept-Encoding: gzip
Authorization: <Not Logged>
User-Agent: jib 0.9.8-SNAPSHOT jib-maven-plugin Google-HTTP-Java-Client/1.23.0 (gzip)

Jul 24, 2018 2:11:40 PM com.google.api.client.http.HttpRequest execute
CONFIG: -------------- REQUEST  --------------
HEAD https://quay.io/v2/briandealwis/jibtest/blobs/sha256:1ad24c91d38c2f3a908e16e428f6fb3d96c4743f0f65559e163b115ddaa521c9
Accept: 
Accept-Encoding: gzip
Authorization: <Not Logged>
User-Agent: jib 0.9.8-SNAPSHOT jib-maven-plugin Google-HTTP-Java-Client/1.23.0 (gzip)

Jul 24, 2018 2:11:40 PM com.google.api.client.http.HttpRequest execute
[...]

PR #682, opened 2018-07-20
System property 'jibSerialize' causes build steps to be executed serially

briandealwis:
For #681.  If system property `jibSerialize` is set then we run all build steps on the main thread.

Our lifecycle messages don't change however — I still see lifecycle messages before any HTTP traffic begins, and we still have some operations happening in parallel (you can see requests for `distroless/java` and to `registry.hub.docker`), suggesting that we still have some parallelism happening without the StepsRunner.

```
[INFO] --- jib-maven-plugin:0.9.8-SNAPSHOT:build (default-cli) @ first ---
[WARNING] Base image 'gcr.io/distroless/java' does not use a specific image digest - build may not be reproducible
[INFO] 
[INFO] Containerizing application to foo/bar/baz...
[INFO] 
[INFO] Retrieving registry credentials for registry.hub.docker.com...
[INFO] Getting base image gcr.io/distroless/java...
[INFO] Building dependencies layer...
Jul 20, 2018 2:25:29 PM com.google.api.client.http.HttpRequest execute
CONFIG: -------------- REQUEST  --------------
GET https://gcr.io/v2/distroless/java/manifests/latest
Accept: application/vnd.oci.image.manifest.v1+json,application/vnd.docker.distribution.manifest.v2+json,application/vnd.docker.distribution.manifest.v1+json
Accept-Encoding: gzip
User-Agent: jib 0.9.8-SNAPSHOT jib-maven-plugin Google-HTTP-Java-Client/1.23.0 (gzip)

Jul 20, 2018 2:25:29 PM com.google.api.client.http.HttpRequest execute
CONFIG: curl -v --compressed -H 'Accept: application/vnd.oci.image.manifest.v1+json,application/vnd.docker.distribution.manifest.v2+json,application/vnd.docker.distribution.manifest.v1+json' -H 'Accept-Encoding: gzip' -H 'User-Agent: jib 0.9.8-SNAPSHOT jib-maven-plugin Google-HTTP-Java-Client/1.23.0 (gzip)' -- 'https://gcr.io/v2/distroless/java/manifests/latest'
Jul 20, 2018 2:25:29 PM com.google.api.client.http.HttpRequest execute
CONFIG: -------------- REQUEST  --------------
GET https://registry.hub.docker.com/v2/
Accept: 
Accept-Encoding: gzip
User-Agent: jib 0.9.8-SNAPSHOT jib-maven-plugin Google-HTTP-Java-Client/1.23.0 (gzip)
```

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -33,6 +33,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import javax.annotation.Nullable;
 
@@ -44,8 +45,7 @@ import javax.annotation.Nullable;
  */
 public class StepsRunner {
 
-  private final ListeningExecutorService listeningExecutorService =
-      MoreExecutors.listeningDecorator(Executors.newCachedThreadPool());
+  private final ListeningExecutorService listeningExecutorService;
   private final BuildConfiguration buildConfiguration;
   private final SourceFilesConfiguration sourceFilesConfiguration;
   private final Cache baseLayersCache;
@@ -76,6 +76,12 @@ public class StepsRunner {
     this.sourceFilesConfiguration = sourceFilesConfiguration;
     this.baseLayersCache = baseLayersCache;
     this.applicationLayersCache = applicationLayersCache;
+
+    ExecutorService executorService =
+        System.getProperty("jibSerialized") == null
+            ? Executors.newCachedThreadPool()
+            : MoreExecutors.newDirectExecutorService();
+    listeningExecutorService = MoreExecutors.listeningDecorator(executorService);
   }
 
   public StepsRunner runRetrieveTargetRegistryCredentialsStep() {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -78,9 +78,9 @@ public class StepsRunner {
     this.applicationLayersCache = applicationLayersCache;
 
     ExecutorService executorService =
-        System.getProperty("jibSerialized") == null
-            ? Executors.newCachedThreadPool()
-            : MoreExecutors.newDirectExecutorService();
+        Boolean.getBoolean("jibSerialized")
+            ? MoreExecutors.newDirectExecutorService()
+            : Executors.newCachedThreadPool();
     listeningExecutorService = MoreExecutors.listeningDecorator(executorService);
   }
 


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -76,6 +76,12 @@ public StepsRunner(
     this.sourceFilesConfiguration = sourceFilesConfiguration;
     this.baseLayersCache = baseLayersCache;
     this.applicationLayersCache = applicationLayersCache;
+
+    ExecutorService executorService =
+        System.getProperty("jibSerialized") == null
```
coollog: Should we check for a specific value of `jibSerialized` for the serialized case (to avoid someone accidentally using serialized when not intentional)?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -76,6 +76,12 @@ public StepsRunner(
...
+            ? Executors.newCachedThreadPool()
+            : MoreExecutors.newDirectExecutorService();
```
coollog: Nice and simple :)

briandealwis:
Oh ok, the ordering of
```
[INFO] Getting base image ghostbuster91/test-jib2...
[INFO] Retrieving registry credentials for registry.hub.docker.com…
```

happens because [`PullBaseImageStep#call()`](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L104) first [tries to pull the image with no credentials](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L107) and, if it fails, then [tries again](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L123) after [obtaining credentials](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L120) (which prints the _Retrieving registry credentials for …_).  So we should just do a bit more logging here.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
briandealwis: Turned it into a boolean property that must be `true`.

briandealwis:
#685 addresses some of the log interleaving. So this is ready.

briandealwis:
/me grumbles about having to continually merge head

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -76,6 +76,12 @@ public StepsRunner(
     this.sourceFilesConfiguration = sourceFilesConfiguration;
     this.baseLayersCache = baseLayersCache;
     this.applicationLayersCache = applicationLayersCache;
+
+    ExecutorService executorService =
+        Boolean.getBoolean("jibSerialized")
```
chanseokoh: @briandealwis is this supposed to be `jibSerialize` or `jibSerialized`? If the latter, please update the title. Can you also update the `CHANGELOG` files?

briandealwis:
Hmm, it should be `jibSerialize`, I suppose.

Issue #620, opened 2018-07-15
Improve error message when `from` image cannot be found


ghostbuster91:
**Description of the issue**:
When jib cannot find image specified in `from` build fails with this not very explanatory message. I also think that it is a little bit missleading because  `Getting base image ghostbuster91/test-jib2..` looks fine.
```
❯ ./gradlew jib
warning: Base image 'ghostbuster91/test-jib2' does not use a specific image digest - build may not be reproducible

Containerizing application to ghostbuster91/test-jib...

Retrieving registry credentials for registry.hub.docker.com...
Getting base image ghostbuster91/test-jib2... <-- HERE IT LOOKS OK
Building dependencies layer...
Building classes layer...
Building resources layer...
Retrieving registry credentials for registry.hub.docker.com...

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':jib'.
> Build image failed

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 11s
2 actionable tasks: 1 executed, 1 up-to-date
```

**Expected behavior**:
IMO the best way would be to fail when executing `Getting base image ghostbuster91/test-jib2..`
At least some better error message should be provided.

**Steps to reproduce**: 
Just make a type when declaring `from` image.
**Environment**: 
Ubuntu 18.04

**`jib-gradle-plugin` Configuration**: 
```groovy
plugins {
    id 'com.google.cloud.tools.jib' version '0.9.6'
}
apply plugin: 'java'

group 'test-jib'
version '1.0-SNAPSHOT'

sourceCompatibility = 1.8

repositories {
    mavenCentral()
}

dependencies {
    testCompile group: 'junit', name: 'junit', version: '4.12'
}

jib {
    from {
        image = 'ghostbuster91/test-jib2'
    }
    to {
        image = 'ghostbuster91/test-jib'
    }
    container {
        mainClass = 'Main'
    }
}
```

**Log output**: 

**Additional Information**:

coollog:
Hi @ghostbuster91 , thanks for filing this issue! This is definitely a place where we could improve the feedback to the user when the build fails and thanks for spotting this.

briandealwis:
It's pretty confusing with Maven too:
```
Getting base image non-existant/image...
[INFO] Building dependencies layer...
[INFO] Building classes layer...
[...]

PR #685, opened 2018-07-20
Improve logging for pulling base image

briandealwis:
This is sort of related to #620.  This PR adds a little bit of logging around the `PullBaseImageStep` to notify the user that jib will retry when the initial unauthenticated pull fails.
```
[INFO] Building dependencies layer...
[INFO] Getting base image nonexistant/image with no auth...
[INFO] Trying again for nonexistant/image with auth...
[INFO] Retrieving registry credentials for registry.hub.docker.com…
```

Perhaps the "auth" is a bit too encoded.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -101,10 +101,7 @@ class PullBaseImageStep
           RegistryAuthenticationFailedException {
     buildConfiguration
         .getBuildLogger()
-        .lifecycle(
-            "Getting base image "
-                + buildConfiguration.getBaseImageReference()
-                + " with no auth...");
+        .lifecycle("Getting base image " + buildConfiguration.getBaseImageReference() + "...");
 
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -101,14 +101,22 @@ public BaseImageWithAuthorization call()
           RegistryAuthenticationFailedException {
     buildConfiguration
         .getBuildLogger()
-        .lifecycle("Getting base image " + buildConfiguration.getBaseImageReference() + "...");
+        .lifecycle(
+            "Getting base image "
+                + buildConfiguration.getBaseImageReference()
+                + " with no auth...");
```
coollog: Since the default case is `gcr.io/distroless/java`, I think we can leave off the `with no auth` here, and then below it can be what @chanseokoh recommended.

briandealwis:
I'd rather include the image ref so that it's clear that the message is tied to the previous pull.

chanseokoh:
> Trying again for \<ref\> with auth..."

Gosh, I keep forget that I need to escape brackets, haha.

PR #687, opened 2018-07-20
Prevent sending multiple requests in a connection

### LINKED ISSUE #686: Connection should explicitly prevent reusing (should not allow sending multiple Requests)
The Javadoc says "Make sure to wrap with a try-with-resource to ensure that the connection is closed after usage." This is to call `disconnect()` on `httpResponse`.

```java
  public void close() throws IOException {
    ...
    httpResponse.disconnect();
  }
```
However, the field `httpResponse` is replaced every time a request is sent. The references to old `httpResponse`s will be lost with multiple sends.

```java
  public Response send(String httpMethod, Request request) throws IOException {
    ...
    httpResponse = httpRequest.execute();
    return new Response(httpResponse);
  }
```

Or, alternatively, the class could be extended to allow reusing, but I think this may become YAGNI, so I prefer having the code simple.

chanseokoh:
Fixes #686.

Had to remove the `@Mock` annotation from the `httpResponse` field, since otherwise it is injected into `Connection` by `@InjectMocks`.

Also added another test class for `Connection` that interacts with a local web server.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
@@ -36,12 +36,12 @@ import org.apache.http.NoHttpResponseException;
  *
  * <p>Example usage:
  *
- * <pre>
+ * <pre>{@code
  * try (Connection connection = new Connection(url)) {
  *   Response response = connection.get(request);
  *   // ... process the response
  * }
- * </pre>
+ * }</pre>
  */
 public class Connection implements Closeable {
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
@@ -23,6 +23,7 @@ import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpResponse;
 import com.google.api.client.http.apache.ApacheHttpTransport;
 import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.common.base.Preconditions;
 import java.io.Closeable;
 import java.io.IOException;
 import java.net.URL;
@@ -30,16 +31,17 @@ import javax.annotation.Nullable;
 import org.apache.http.NoHttpResponseException;
 
 /**
- * Sends an HTTP {@link Request} and stores the {@link Response}.
+ * Sends an HTTP {@link Request} and stores the {@link Response}. Clients should not send more than
+ * one request.
  *
  * <p>Example usage:
  *
- * <pre>{@code
+ * <pre>
  * try (Connection connection = new Connection(url)) {
  *   Response response = connection.get(request);
  *   // ... process the response
  * }
- * }</pre>
+ * </pre>
  */
 public class Connection implements Closeable {
 
@@ -118,6 +120,8 @@ public class Connection implements Closeable {
    * @throws IOException if building the HTTP request fails.
    */
   public Response send(String httpMethod, Request request) throws IOException {
+    Preconditions.checkState(httpResponse == null, "Connection can send only one request");
+
     HttpRequest httpRequest =
         requestFactory
             .buildRequest(httpMethod, url, request.getHttpContent())
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
@@ -106,7 +106,7 @@ class DockerConfig {
   DockerCredentialHelper getCredentialHelperFor(
       DockerCredentialHelperFactory dockerCredentialHelperFactory, String registry) {
     List<Predicate<String>> registryMatchers = getRegistryMatchersFor(registry);
-    Map.Entry<String, AuthTemplate> firstMatchInAuths =
+    Map.Entry<String, ?> firstMatchInAuths =
         findFirstInMapByKey(dockerConfigTemplate.getAuths(), registryMatchers);
     if (dockerConfigTemplate.getCredsStore() != null && firstMatchInAuths != null) {
       return dockerCredentialHelperFactory.newDockerCredentialHelper(
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
@@ -22,11 +22,14 @@ import java.io.OutputStream;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 
 /** Simple local web server for testing. */
 public class TestWebServer implements Closeable {
 
   private final ServerSocket serverSocket;
+  private final List<Socket> sockets = new ArrayList<>();
 
   public TestWebServer() throws IOException, InterruptedException {
     serverSocket = new ServerSocket(0);
@@ -42,6 +45,12 @@ public class TestWebServer implements Closeable {
   @Override
   public void close() throws IOException {
     serverSocket.close();
+    for (Socket socket : sockets) {
+      socket.close();
+      try (Socket toClose = socket) {
+      } catch (IOException ex) {
+      }
+    }
   }
 
   private void waitUntilReady() throws IOException, InterruptedException {
@@ -59,10 +68,10 @@ public class TestWebServer implements Closeable {
   private void serve200() {
     try {
       while (true) {
-        try (Socket socket = serverSocket.accept();
-            OutputStream out = socket.getOutputStream()) {
-          out.write("HTTP/1.1 200 OK\n\nHello World!".getBytes(StandardCharsets.UTF_8));
-        }
+        Socket socket = serverSocket.accept();
+        sockets.add(socket);
+        OutputStream out = socket.getOutputStream();
+        out.write("HTTP/1.1 200 OK\n\nHello World!\n\n".getBytes(StandardCharsets.UTF_8));
       }
     } catch (IOException ex) {
     }
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
@@ -18,27 +18,21 @@ package com.google.cloud.tools.jib.http;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.nio.charset.StandardCharsets;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Semaphore;
 
 /** Simple local web server for testing. */
 class TestWebServer implements Closeable {
 
   private final ServerSocket serverSocket;
-  private final List<Socket> sockets = new ArrayList<>();
-  private final ExecutorService executorService = Executors.newSingleThreadExecutor();
   private final Semaphore threadStarted = new Semaphore(0);
+  private volatile boolean stop = false;
 
   TestWebServer() throws IOException, InterruptedException {
     serverSocket = new ServerSocket(0);
-    executorService.submit(this::serve200);
+    new Thread(this::serve200).start();
     threadStarted.acquire();
   }
 
@@ -49,22 +43,16 @@ class TestWebServer implements Closeable {
 
   @Override
   public void close() throws IOException {
+    stop = true;
     serverSocket.close();
-    for (Socket socket : sockets) {
-      try (Socket toClose = socket) {
-      } catch (IOException ex) {
-      }
-    }
-    executorService.shutdown();
   }
 
-  private Void serve200() throws IOException, InterruptedException {
+  private void serve200() {
     threadStarted.release();
-    while (true) {
-      Socket socket = serverSocket.accept();
-      sockets.add(socket);
-      try (OutputStream out = socket.getOutputStream()) {
-        out.write("HTTP/1.1 200 OK\n\nHello World!".getBytes(StandardCharsets.UTF_8));
+    while (!stop) {
+      try (Socket socket = serverSocket.accept()) {
+        String response = "HTTP/1.1 200 OK\nContent-Length:12\n\nHello World!\n\n";
+        socket.getOutputStream().write(response.getBytes(StandardCharsets.UTF_8));
       } catch (IOException e) {
       }
     }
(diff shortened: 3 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -23,23 +23,25 @@
...
 
 /**
- * Sends an HTTP {@link Request} and stores the {@link Response}.
+ * Sends an HTTP {@link Request} and stores the {@link Response}. Clients should not send more than
+ * one request.
  *
  * <p>Example usage:
  *
- * <pre>{@code
+ * <pre>
```
coollog: I believe the `{@code` is the recommended way of wrapping multi-line code blocks. (https://stackoverflow.com/a/542142/10047775)

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
```diff
@@ -106,7 +106,7 @@ DockerCredentialHelper getCredentialHelperFor(String registry) {
   DockerCredentialHelper getCredentialHelperFor(
       DockerCredentialHelperFactory dockerCredentialHelperFactory, String registry) {
     List<Predicate<String>> registryMatchers = getRegistryMatchersFor(registry);
-    Map.Entry<String, AuthTemplate> firstMatchInAuths =
+    Map.Entry<String, ?> firstMatchInAuths =
```
coollog: Was this intentional?

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerConnectionTest.java
```diff
@@ -0,0 +1,89 @@
...
+/** Tests for {@link Connection} using an actual local server. */
+public class WithServerConnectionTest {
+
+  private static class WebServer implements Closeable {
+
+    private ServerSocket serverSocket;
+    private boolean stop;
+
+    private WebServer() throws IOException {
+      serverSocket = new ServerSocket(8087);
```
coollog: Should we use a 5-digit port here for less chance of collisions?

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerConnectionTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+/** Tests for {@link Connection} using an actual local server. */
+public class WithServerConnectionTest {
+
+  private static class WebServer implements Closeable {
+
+    private ServerSocket serverSocket;
+    private boolean stop;
```
coollog: This will probably need to be `volatile`

chanseokoh: Ah, you're right.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
chanseokoh: TIL

chanseokoh:
Just noticed I am not waiting until the server is properly up.

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerConnectionTest.java
chanseokoh: I'll just let it pick up any available port.

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerConnectionTest.java
chanseokoh: Turns out the field isn't needed at all, since `accept()` will throw an exception once `serverSocket` is closed, terminating the loop.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
chanseokoh: Did not really intend, but I guess we can have it here as a trivial cleanup. Signifies that the value is not being used, and the added benefit of less cramped code.

coollog: kk sounds good

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+      }
+      Thread.sleep(50);
+    }
+    Thread.currentThread().interrupt();
+  }
+
+  private void serve200() {
+    try {
+      while (true) {
```
coollog: I think you could do a `wait/notify` for this rather than the 50ms thread sleep and remove the `waitUntilReady` method.

chanseokoh:
Windows is kind of weird. Could be the CRLF incompatibility, or perhaps I need keep the connection alive.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.nio.charset.StandardCharsets;
+
+/** Simple local web server for testing. */
+public class TestWebServer implements Closeable {
```
coollog: nit: could be package-private (and methods)

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -59,10 +68,10 @@ private void waitUntilReady() throws IOException, InterruptedException {
   private void serve200() {
     try {
       while (true) {
-        try (Socket socket = serverSocket.accept();
-            OutputStream out = socket.getOutputStream()) {
-          out.write("HTTP/1.1 200 OK\n\nHello World!".getBytes(StandardCharsets.UTF_8));
-        }
+        Socket socket = serverSocket.accept();
+        sockets.add(socket);
+        OutputStream out = socket.getOutputStream();
```
coollog: I think the issue is that the `OutputStream` needs to be closed first.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
chanseokoh: I wanted to make sure `serverSocket.accept()` is reached, instead of notifying just before calling `accept()`.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
chanseokoh: I don't really think so... closing socket should just close anything underneath it gracefully. Anyway, the test is now passing, so I think my theory about keep-alive seems right.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
coollog: Hmm, is there an issue with not reaching `serverSocket.accept()`?

chanseokoh: However, it wouldn't hurt to simply the code at the expense of a little bit of imprefection, given the code is only for testing. After all, theoretically I am guessing the server will not respond while it is serving a client, so I think it is still not perfect anyway.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
chanseokoh: But you could be also right. It'd be interesting to test closing `OutputStream` first on Windows.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,79 @@
...
+
+/** Simple local web server for testing. */
+public class TestWebServer implements Closeable {
+
+  private final ServerSocket serverSocket;
+  private final List<Socket> sockets = new ArrayList<>();
+
+  public TestWebServer() throws IOException, InterruptedException {
+    serverSocket = new ServerSocket(0);
+    new Thread(this::serve200).start();
```
coollog: An alternative here would be to do `Executors.newSingleThreadExecutor().submit(this::serve200);` and change `serve200` to `private Void serve200() throws IOException` to remove the empty `catch`.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
coollog: Actually, nevermind, the try-with-resource closes resources with the reverse order of creation.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private Void serve200() throws IOException {
+    threadStarted.release();
+    while (true) {
+      Socket socket = serverSocket.accept();
+      socketsToClose.add(socket);
+      try {
+        String response = "HTTP/1.1 200 OK\nContent-Length:12\n\nHello World!\n\n";
+        socket.getOutputStream().write(response.getBytes(StandardCharsets.UTF_8));
+      } catch (IOException ex) {
```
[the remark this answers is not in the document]
chanseokoh: Good point about flush.

The catch block is needed not to terminate the entire server when an I/O error occurs when responding to some client.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,72 @@
...
+    String host = serverSocket.getInetAddress().getHostAddress();
+    return "http://" + host + ":" + serverSocket.getLocalPort();
+  }
+
+  @Override
+  public void close() throws IOException {
+    serverSocket.close();
+    for (Socket socket : socketsToClose) {
+      try (Socket toClose = socket) {
+      } catch (IOException ex) {
```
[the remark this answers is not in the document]
chanseokoh: Yeah, not a big deal, but this is to continue closing other sockets when closing a certain socket fails. (The exception is thrown when calling `close()` on a socket.)

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+    executorService.shutdown();
+  }
+
+  private Void serve200() throws IOException {
+    threadStarted.release();
+    while (true) {
+      Socket socket = serverSocket.accept();
+      socketsToClose.add(socket);
```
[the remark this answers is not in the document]
chanseokoh: It's safe to call `close()` multiple times on a socket or call it on a closed socket. Probably it is safe on almost all `Closeable` classes.

(And I don't think I get the out-of-bound exception. The enhanced `for` loop never throws it.)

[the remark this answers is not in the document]
coollog: Oh sorry I was referring to `TestWebServer#close`, where it loops over the `socketsToClose` non-concurrent list, but this list is being updated (via `add`) in another thread. Perhaps the `executorService.shutdown()` should be called at the start of `TestWebServer#close`.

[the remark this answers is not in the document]
chanseokoh: Ah, haven't thought about that. Using `ConcurrentLinkedQueue` would make it safe. If `close()` is called concurrently, it may still miss a socket from a running thread, but that's OK.

BTW, I'd like to give a few more final tests on Windows.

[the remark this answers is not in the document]
chanseokoh: That's a very nice idea.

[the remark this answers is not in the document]
chanseokoh: Another idea that can close sockets early would be to wait until the client has closed the channel (i.e., checking `InputStream` is still open). However, the current implementation seems to work and enough for testing.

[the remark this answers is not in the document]
chanseokoh: > wait until the client has closed the channel

Yup, it works. This is definitely the right way. No more weird lazy close, and no need to flush either.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,68 @@
...
+  private Void serve200() throws IOException {
+    threadStarted.release();
+    while (true) {
+      try (Socket socket = serverSocket.accept()) {
+        try {
+          String response = "HTTP/1.1 200 OK\nContent-Length:12\n\nHello World!";
+          socket.getOutputStream().write(response.getBytes(StandardCharsets.UTF_8));
+
+          InputStream in = socket.getInputStream();
+          for (int ch = in.read(); ch != -1; ch = in.read()) ;
```
coollog: Could also do `ByteStreams.copy(socket.getInputStream(), ByteStreams.nullOutputStream());`

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,68 @@
...
+
+  String getEndpoint() {
+    String host = serverSocket.getInetAddress().getHostAddress();
+    return "http://" + host + ":" + serverSocket.getLocalPort();
+  }
+
+  @Override
+  public void close() throws IOException {
+    serverSocket.close();
+    executorService.shutdown();
```
coollog: Should this go before `serverSocket.close()`? I'm thinking if the `serverSocket.close()` is called while line 56 is running, it would throw an exception.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  private Void serve200() throws IOException {
+    threadStarted.release();
+    while (true) {
```
coollog: I'm thinking maybe this web server can just serve one request, since the tests only require that.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
chanseokoh: That's the point of `serverSocket.close()`, because `accept()` is a blocking call; that is, to terminate the thread.

Also, calling `executorService.shutdown()` doesn't help. It just prevents future scheduling. Generally, it's not possible to kill a thread. `shutdownNow()` can't do it either.

chanseokoh: FYI, the main reason for `shutdown()` is related to terminating JVM: https://stackoverflow.com/questions/29999109/executorservice-jvm-doesnt-terminate

PR #689, opened 2018-07-21
Simplify TarStreamBuilder

### LINKED ISSUE #630: Simplify TarStreamBuilder#addEntry() to deal with bytes
From https://github.com/GoogleContainerTools/jib/pull/627#pullrequestreview-137528865

`TarStreamBuilder#addEntry(String content, String fileLocation)` seems to be called by `ImageToTarballTranslator#toTarballBlob()` to add container configuration and manifest files. Oddly the content of these files is passed as a String such that the `TarStreamBuilder` is responsible for performing an text to byte encoding. Rather than provide two variants of `TarStreamBuilder#addEntry()`, why don't we just pass the file contents as `byte[]` and let `toTarballBlob()` deal with the encoding?

TadCordle:
Fix #630

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
@@ -33,6 +33,7 @@ public class Blobs {
     return new FileBlob(file);
   }
 
+  /** Creates a {@link StringBlob} with UTF-8 encoding. */
   public static Blob from(String content) {
     return new StringBlob(content);
   }
@@ -42,7 +43,7 @@ public class Blobs {
   }
 
   /**
-   * Writes the BLOB to a string.
+   * Writes the BLOB to a string with UTF-8 encoding.
    *
    * @param blob the BLOB to write
    * @return the BLOB contents as a string
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
@@ -33,7 +33,12 @@ public class Blobs {
     return new FileBlob(file);
   }
 
-  /** Creates a {@link StringBlob} with UTF-8 encoding. */
+  /**
+   * Creates a {@link StringBlob} with UTF-8 encoding.
+   *
+   * @param content the string to create the blob from
+   * @return the {@link StringBlob}
+   */
   public static Blob from(String content) {
     return new StringBlob(content);
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
@@ -48,7 +48,7 @@ public class Blobs {
   }
 
   /**
-   * Writes the BLOB to a string with UTF-8 encoding.
+   * Writes the BLOB to a string with UTF-8 decoding.
    *
    * @param blob the BLOB to write
    * @return the BLOB contents as a string
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageToTarballTranslator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageToTarballTranslator.java
@@ -65,12 +65,13 @@ public class ImageToTarballTranslator {
     Blob containerConfigurationBlob =
         new ImageToJsonTranslator(image).getContainerConfigurationBlob();
     tarStreamBuilder.addEntry(
-        Blobs.writeToString(containerConfigurationBlob), CONTAINER_CONFIGURATION_JSON_FILE_NAME);
+        Blobs.writeToByteArray(containerConfigurationBlob), CONTAINER_CONFIGURATION_JSON_FILE_NAME);
 
     // Adds the manifest to tarball.
     manifestTemplate.setRepoTags(imageReference.toStringWithTag());
     tarStreamBuilder.addEntry(
-        Blobs.writeToString(JsonTemplateMapper.toBlob(manifestTemplate)), MANIFEST_JSON_FILE_NAME);
+        Blobs.writeToByteArray(JsonTemplateMapper.toBlob(manifestTemplate)),
+        MANIFEST_JSON_FILE_NAME);
 
     return tarStreamBuilder.toBlob();
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageToTarballTranslator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageToTarballTranslator.java
@@ -65,13 +65,12 @@ public class ImageToTarballTranslator {
     Blob containerConfigurationBlob =
         new ImageToJsonTranslator(image).getContainerConfigurationBlob();
     tarStreamBuilder.addEntry(
-        Blobs.writeToByteArray(containerConfigurationBlob), CONTAINER_CONFIGURATION_JSON_FILE_NAME);
+        Blobs.writeToString(containerConfigurationBlob), CONTAINER_CONFIGURATION_JSON_FILE_NAME);
 
     // Adds the manifest to tarball.
     manifestTemplate.setRepoTags(imageReference.toStringWithTag());
     tarStreamBuilder.addEntry(
-        Blobs.writeToByteArray(JsonTemplateMapper.toBlob(manifestTemplate)),
-        MANIFEST_JSON_FILE_NAME);
+        Blobs.writeToString(JsonTemplateMapper.toBlob(manifestTemplate)), MANIFEST_JSON_FILE_NAME);
 
     return tarStreamBuilder.toBlob();
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageToTarballTranslator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageToTarballTranslator.java
@@ -57,19 +57,20 @@ public class ImageToTarballTranslator {
       Path layerContentFile = layer.getContentFile();
       String layerName = layerContentFile.getFileName().toString();
 
-      tarStreamBuilder.addEntry(new TarArchiveEntry(layerContentFile.toFile(), layerName));
+      tarStreamBuilder.addTarArchiveEntry(
+          new TarArchiveEntry(layerContentFile.toFile(), layerName));
       manifestTemplate.addLayerFile(layerName);
     }
 
     // Adds the container configuration to the tarball.
     Blob containerConfigurationBlob =
         new ImageToJsonTranslator(image).getContainerConfigurationBlob();
-    tarStreamBuilder.addEntry(
+    tarStreamBuilder.addTextEntryInUtf8(
         Blobs.writeToString(containerConfigurationBlob), CONTAINER_CONFIGURATION_JSON_FILE_NAME);
 
     // Adds the manifest to tarball.
     manifestTemplate.setRepoTags(imageReference.toStringWithTag());
-    tarStreamBuilder.addEntry(
+    tarStreamBuilder.addTextEntryInUtf8(
         Blobs.writeToString(JsonTemplateMapper.toBlob(manifestTemplate)), MANIFEST_JSON_FILE_NAME);
 
     return tarStreamBuilder.toBlob();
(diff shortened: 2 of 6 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -33,17 +31,11 @@
...
 
-  @FunctionalInterface
-  private interface TarArchiveOutputStreamConsumer {
-    void accept(TarArchiveOutputStream tarArchiveOutputStream) throws IOException;
-  }
-
   /**
-   * Maps from {@link TarArchiveEntry} to function that outputs the entry onto a {@link
-   * TarArchiveOutputStream}. The order of the entries is the order they belong in the tarball.
+   * Maps from {@link TarArchiveEntry} to a {@link ByteSource}. The order of the entries is the
```
coollog: Since we already have `Blob`, we can just use that in place of `ByteSource` - it's basically the same thing except with a much smaller API (just `writeTo`). The alternative is to replace `Blob` with using `ByteSource`.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -56,45 +48,35 @@ private void writeEntriesAsTarArchive(OutputStream tarByteStream) throws IOExcep
...
-        tarArchiveOutputStream -> {
-          // Note that this will skip files that don't exist.
-          if (entry.isFile()) {
-            try (InputStream contentStream =
-                new BufferedInputStream(Files.newInputStream(entry.getFile().toPath()))) {
-              ByteStreams.copy(contentStream, tarArchiveOutputStream);
-            }
-          }
-        });
+        entry.isFile() ? MoreFiles.asByteSource(entry.getFile().toPath()) : ByteSource.empty());
```
coollog: Using `Blob`, it would be `entry.isFile() ? Blobs.from(entry.getFile().toPath()) : Blobs.from(ignored -> {})`

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -56,45 +48,35 @@ private void writeEntriesAsTarArchive(OutputStream tarByteStream) throws IOExcep
...
   /**
    * Adds a blob to the archive.
    *
-   * @param contents the blob contents to add to the tarball.
-   * @param name the name of the entry (i.e. filename).
+   * @param contents the blob contents to add to the tarball
+   * @param name the name of the entry (i.e. filename)
    */
-  public void addEntry(String contents, String name) {
+  public void addEntry(byte[] contents, String name) {
```
coollog: I think the point of using `String` as a parameter type is to prevent callers from accidentally using this signature for files, where they read the file into a `byte[]` and then call this method.

TadCordle: How does using `String` instead of `byte[]` prevent that?

coollog: There's a `Files.readAllBytes` method that is used to read files to a `byte[]` and not a method for reading to string - and reading files to a string in general is not a good idea unless that file is guaranteed to be a text file.

TadCordle: I'd argue that making it a `String` doesn't help much with that, considering you can just wrap the `Files.readAllBytes` in a `new String()` (which we actually used to do in the [test](https://github.com/GoogleContainerTools/jib/pull/689/files/52c6e480812342d79bc4e29bd8c6436c3dc262c5#diff-843239071ea940328c65b2c1f0d2ba94L58)). Might just be more useful to update the javadoc to warn against using this signature with files.

coollog: That sounds fine (updating the javadoc), although I'd argue that generally, we should keep APIs as type-restricted as possible (ie. favor `SpecificType` over `Object` and `DataType` over `byte[]`).

coollog: Perhaps `JsonTemplate` might be the more applicable type here instead of `String` - although that introduces a dependency from the `tar` package to the `json` package.

chanseokoh: > `String` is encoded by `UTF-8`

I think I wasn't really correct with this statement. https://docs.oracle.com/javase/6/docs/api/java/lang/Character.html#unicode says

> The Java 2 platform uses the UTF-16 representation in `char` arrays and in the `String` and `StringBuffer` classes.

Anyways, it doesn't matter regarding converting a `String` to bytes with any encoding; `getBytes(Charset)` will do the righr job. So, I guess the question here is, does the TAR stream always expect that text strings should be written in the UTF-8 encoding? That is, if someone reads the written bytes later, will it always decode the bytes with UTF-8? If the answer is yes, I *think* it is not so wrong to accept `String` (which should be plain texts, by the way).

coollog: Perhaps the right solution here is to change the signature to `public void addEntry(String jsonString, String name)`, since JSON is defined to be encoded in UTF-8 [[RFC8259](https://tools.ietf.org/html/rfc8259#section-8.1)],

TadCordle: @elharo Would be interested in hearing what you think.

TadCordle: @briandealwis as well

chanseokoh: I don't know if the TAR string entry specification mandates writing strings in UTF-8, but if it is in fact true, I'd be OK with something like `addTextEntryInUtf8(String text, String name)`.

briandealwis: As tar files are just bags of bytes, I'd be happy with using `Blob`.  It has the advantage of minimizing Guava types in our method signatures.  Let the caller determine how to encode whatever they're passing in.

coollog: I agree - `Blob` would be good too as we have a purpose-built API for our use cases for that.

TadCordle: I guess the direction we're headed with this class is less a general tarball builder and more something to fit our specific purposes, and the only things we're ever adding are files and raw strings. I'll stick with the string parameter and just rename the method as @chanseokoh suggested.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -56,45 +46,36 @@ private void writeEntriesAsTarArchive(OutputStream tarByteStream) throws IOExcep
...
+   * @param name the name of the entry (i.e. filename)
    */
-  public void addEntry(String contents, String name) {
+  public void addTextEntryInUtf8(String jsonString, String name) {
     TarArchiveEntry entry = new TarArchiveEntry(name);
-    byte[] contentsBytes = contents.getBytes(StandardCharsets.UTF_8);
-    entry.setSize(contentsBytes.length);
-    archiveMap.put(entry, tarArchiveOutputStream -> tarArchiveOutputStream.write(contentsBytes));
+    entry.setSize(jsonString.getBytes(StandardCharsets.UTF_8).length);
+    archiveMap.put(entry, Blobs.from(jsonString));
```
coollog: I think we need to expose the encoding in `Blobs.from(String)` since that creates a `StringBlob` with UTF8 encoding too (or at least document it explicitly). (same with `Blobs#writeToString`)

jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
```diff
@@ -42,7 +48,7 @@ public static Blob from(BlobWriter writer) {
   }
 
   /**
-   * Writes the BLOB to a string.
+   * Writes the BLOB to a string with UTF-8 encoding.
```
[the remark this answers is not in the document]
chanseokoh: This one only. You assume the raw bytes are in UTF-8, so you do UTF-8 decoding to covert them to a `String` object.

For the `from(String) one, you convert a `String` to a sequence of bytes in UTF-8 format (encoding the string in UTF-8).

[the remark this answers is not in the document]
TadCordle: Ah makes sense, thanks

jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
```diff
@@ -42,7 +48,7 @@ public static Blob from(BlobWriter writer) {
   }
 
   /**
-   * Writes the BLOB to a string.
+   * Writes the BLOB to a string with UTF-8 decoding.
```
briandealwis: This is where the carry-files-as-strings breaks down: although any String can be encoded as a set of bytes in UTF-8, not all bytes are valid UTF-8 strings.

If you want to go down this route, I'd much prefer `Blob encodeStringBlob()` and `decodeStringBlob()`

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageToTarballTranslator.java
```diff
@@ -57,19 +57,20 @@ public Blob toTarballBlob(ImageReference imageReference) throws IOException {
...
+          new TarArchiveEntry(layerContentFile.toFile(), layerName));
       manifestTemplate.addLayerFile(layerName);
     }
 
     // Adds the container configuration to the tarball.
     Blob containerConfigurationBlob =
         new ImageToJsonTranslator(image).getContainerConfigurationBlob();
-    tarStreamBuilder.addEntry(
+    tarStreamBuilder.addTextEntryInUtf8(
         Blobs.writeToString(containerConfigurationBlob), CONTAINER_CONFIGURATION_JSON_FILE_NAME);
```
briandealwis: So we're converting a blob to a string only for the tarStreamBuilder to write it back out as bytes? Really?

TadCordle: I have to agree with you here. I can see the reason for wanting to make it a String parameter considering the use case, but the operation itself is so much more straightforward if we do everything with bytes.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -56,45 +46,35 @@ private void writeEntriesAsTarArchive(OutputStream tarByteStream) throws IOExcep
...
 
   /**
-   * Adds a blob to the archive.
+   * Adds a blob to the archive. Note that this should be used with raw bytes and not file contents;
+   * for adding files to the archive, use {@code TarStreamBuilder#addEntry(TarArchiveEntry entry)}.
    *
-   * @param contents the blob contents to add to the tarball.
-   * @param name the name of the entry (i.e. filename).
+   * @param contents the bytes to add to the tarball
+   * @param name the UTF-8 encoded name of the entry (i.e. filename)
```
chanseokoh: The "UTF-8 encoded" doesn't sound right. I see `TarArchiveEntry` just accepts a `String`, and we are just good. Basically, a `String` in memory in Java can be thought of as a conceptual thing irrelevant of encoding. The encoding matters only when you must write it down in bytes.

TadCordle: Oh ok, makes sense.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -56,45 +46,35 @@ private void writeEntriesAsTarArchive(OutputStream tarByteStream) throws IOExcep
...
-            }
-          }
-        });
+        entry, entry.isFile() ? Blobs.from(entry.getFile().toPath()) : Blobs.from(ignored -> {}));
   }
 
   /**
-   * Adds a blob to the archive.
+   * Adds a blob to the archive. Note that this should be used with raw bytes and not file contents;
+   * for adding files to the archive, use {@code TarStreamBuilder#addEntry(TarArchiveEntry entry)}.
```
chanseokoh: Oh, "addEntry" should be renamed too.

coollog: I'm a bit late here, but this could be just `{@link #addTarArchiveEntry}`, which will actually add compile-time checks I believe.

PR #696, opened 2018-07-23
Simplify RegistryEndpointCaller code

chanseokoh:
I notice that `RequestState.authorization` is always the value given to the `RegistryEndpointCaller` constructor, so there is not much reason to carry this around in `RequestState`.

After that, `RequestState` becomes a meaningless class holding only a single `URL` field, so it seems obvious to get rid of it.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
@@ -86,7 +86,7 @@ public class RegistryAuthenticator {
         throw new RegistryAuthenticationFailedException(ex);
 
       } catch (InsecureRegistryException ex) {
-        // HTTP is not allowed, so just return null.
+        // Cannot skip certificate validation or use HTTP, so just return null.
         return null;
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -86,7 +86,7 @@ public RegistryAuthenticator initialize()
         throw new RegistryAuthenticationFailedException(ex);
 
       } catch (InsecureRegistryException ex) {
-        // HTTP is not allowed, so just return null.
+        // Cannot skip certificate validation or use HTTP, so just return null.
```
briandealwis: This doesn't seem related.

TadCordle: I think it's small enough that it's fine to keep it in the PR

chanseokoh: No worries. Will have this in a later PR.

Issue #643, opened 2018-07-17
Make allowInsecureRegistries follow Docker semantics for use-insecure-registries


chanseokoh:
Just ignore the certification error and proceed with HTTPS.

https://docs.docker.com/registry/insecure/#deploy-a-plain-http-registry

> With insecure registries enabled, Docker goes through the following steps:
> 
> - First, try using HTTPS.
>    - If HTTPS is available but the certificate is invalid, ignore the error about the certificate.
>    - If HTTPS is not available, fall back to HTTP.

When implemented, cases like #543 will be supported naturally and we will be able to remove the workaround doc: https://github.com/GoogleContainerTools/jib/blob/master/docs/configure-gcp-credentials.md

Should update docs once fixed.

chanseokoh:
Just documenting here what the security implication of `allowInsecureRepositories` that I am aware of, once this is fixed.

Jib will proceed with HTTPS when the server certificate (including a self-signed one) cannot be verified. The user may mistype the server URL and accidentally connect to a malicious server whose URL is set up to resemble the original server URL. TLS can't verify the server, but since `allowInsecureRepositories` is set, Jib proceeds and sends the password to the malicious server.

chanseokoh:
Fortunately, `ApacheHttpTransport` already provides a helper method to skip the validation.

```java
    Builder transportBuilder = new ApacheHttpTransport.Builder();
    if (!turnOffCertificateValidation) {
      transportBuilder.doNotValidateCertificate();
    }
    requestFactory = transportBuilder.build().createRequestFactory();
```

chanseokoh:
Should also update the docs, once this is fixed.

PR #708, opened 2018-07-24
Support insecure TLS connection

chanseokoh:
Towards #643.

Note that this just adds the ability to create an insecure `Connection` instance; nothing is wired up yet.

Also sets up a local SSL server for testing.


jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -45,15 +47,43 @@
...
-  /**
-   * Do not use {@link NetHttpTransport}. It does not process response errors properly. A new {@link
-   * ApacheHttpTransport} needs to be created for each connection because otherwise HTTP connection
-   * persistence causes the connection to throw {@link NoHttpResponseException}.
-   *
-   * @see <a
-   *     href="https://github.com/google/google-http-java-client/issues/39">https://github.com/google/google-http-java-client/issues/39</a>
-   */
-  private HttpRequestFactory requestFactory = new ApacheHttpTransport().createRequestFactory();
+  public static class Builder {
```
coollog: short javadoc here

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -66,7 +96,12 @@
    * @param url the url to send the request to
    */
   public Connection(URL url) {
```
coollog: If we are using a `Builder` for the class, we should make all the constructors private.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -45,15 +47,43 @@
...
+     */
+    public Builder doNotValidateCertificate() throws GeneralSecurityException {
+      ApacheHttpTransport.Builder transportBuilder = new ApacheHttpTransport.Builder();
+      transportBuilder.doNotValidateCertificate();
+      httpTransport = transportBuilder.build();
+      return this;
+    }
+
+    public Connection build() {
+      return new Connection(url, httpTransport);
```
coollog: We could probably create the `HttpTransport` here to avoid creating the extra `ApacheHttpTransport` that gets overwritten by a call to `doNotValidateCertificate`. Maybe have a flag like `shouldValidateCertificate` default to `true` and set to `false` in `doNotValidateCertificate`?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -45,15 +47,43 @@
...
+    }
+
+    /**
+     * Turns off the normal TLS peer verification.
+     *
+     * @throws GeneralSecurityException if unable to turn off
+     * @return this
+     */
+    public Builder doNotValidateCertificate() throws GeneralSecurityException {
+      ApacheHttpTransport.Builder transportBuilder = new ApacheHttpTransport.Builder();
```
coollog: I think it would be fine to have this as one line `httpTransport = new ApacheHttpTransport.Builder().doNotValidateCertificate().build();`

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerConnectionTest.java
```diff
@@ -52,4 +56,22 @@ public void testErrorOnSecondSend() throws IOException, InterruptedException {
...
+            new Connection.Builder(new URL(server.getEndpoint()))
+                .doNotValidateCertificate()
+                .build()) {
+      Response response = connection.send("GET", new Request.Builder().build());
+
+      Assert.assertEquals(200, response.getStatusCode());
+
+      ByteArrayOutputStream out = new ByteArrayOutputStream();
+      response.getBody().writeTo(out);
+      Assert.assertEquals("Hello World!", out.toString("UTF-8"));
```
coollog: Should prob use `StandardCharsets.UTF_8` rather than the string form.

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerConnectionTest.java
```diff
@@ -52,4 +56,22 @@ public void testErrorOnSecondSend() throws IOException, InterruptedException {
       }
     }
   }
+
+  @Test
+  public void testInsecureConnection()
+      throws IOException, InterruptedException, GeneralSecurityException, URISyntaxException {
+    try (TestWebServer server = new TestWebServer(true);
+        Connection connection =
+            new Connection.Builder(new URL(server.getEndpoint()))
```
coollog: Generally, in this codebase, we've been doing the builder instantiation with a static method `builder()` so that this can be like `Connection.builder(new URL(server.getEndpoint())).doNot...`

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -43,6 +43,12 @@
  */
 class RegistryEndpointCaller<T> {
 
+  @FunctionalInterface
+  @VisibleForTesting
+  static interface ConnectionFactory {
```
coollog: short javadoc here too

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -43,6 +43,12 @@
...
+    Connection create(URL url) throws GeneralSecurityException;
```
coollog: nit: newline after class header

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
chanseokoh: This is to avoid forcing callers that don't call `doNotValidateCertificate()` to unnecessary add an empty try-catch block to ignore the `GeneralSecurityException` that will never be thrown.

coollog: That's a good point - perhaps instead of a builder pattern, we should just use static factories that instantiate `Connection` with the right `HttpRequestFactory `.

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerConnectionTest.java
chanseokoh: `toString()` doesn't have the version to accept `Charset` directly, but I can do `StandardCharsets.UTF_8.name()`.

coollog: In other parts of the codebase, we've been using like `new String(out.toByteArray(), StandardCharsets.UTF_8)`

jib-core/src/main/java/com/google/cloud/tools/jib/registry/InsecureRegistryException.java
```diff
@@ -24,6 +24,6 @@
 public class InsecureRegistryException extends RegistryException {
 
   InsecureRegistryException(URL insecureUrl) {
-    super("Only secure connections are allowed, but tried to reach URL " + insecureUrl);
+    super("Only secure connections are allowed, but failed to verify the server at " + insecureUrl);
```
TadCordle: The phrasing here is a bit awkward, maybe something like "Failed to verify the server at `<url>` because only secure connections are allowed"

TadCordle:
I think this deserves a changelog entry?

chanseokoh:
> I think this deserves a changelog entry?

There is no user-visible change yet. Will add a cangelog entry in the PR that closes #643.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -46,14 +48,37 @@
...
 
   /**
-   * Do not use {@link NetHttpTransport}. It does not process response errors properly. A new {@link
-   * ApacheHttpTransport} needs to be created for each connection because otherwise HTTP connection
-   * persistence causes the connection to throw {@link NoHttpResponseException}.
+   * Returns a factory for {@link Connection}.
    *
-   * @see <a
-   *     href="https://github.com/google/google-http-java-client/issues/39">https://github.com/google/google-http-java-client/issues/39</a>
+   * @return {@link Connection} factory
```
coollog: nit: maybe include a description of what the factory is (like "a function that generates a Connection to a URL")

Issue #595, opened 2018-07-12
Propagate container configuration from base image?


coollog:
Thumbs up if this is useful for you.

chanseokoh:
Hmm... if the images built by docker in the normal way don't inherit the container configuration, I think we don't need to do this. OTOH, if they do, this is something we can consider.

coollog:
Yes, I believe Dockerfile `FROM` uses the container configuration from the base image. Any instructions afterwards are extensions on that base container configuration.

essh:
This is useful for me to pick up on the Env from a base image (specifically `PATH` and `JAVA_HOME`). As already mentioned these were previously coming along when using a DockerFile `FROM` and would save having to specify them when launching a container built with jib.

coollog:
I think for this feature, we can start with propagating environment variables since that seems to be the most useful case.

TadCordle:
Adjusting the milestone since we were only planning on propagating environment variables for this next release.

chanseokoh:
For those who are following this issue, environment variables from the base image are not being propagated into the target image.

chanseokoh:
> environment variables from the base image are not being propagated

Sorry, typo. Environment variables from the base image are **now** being propagated

chanseokoh:
Turns out we also need to propagate `workingDir` to make use of distroless-jetty as a base image for #431. Will work on this.

chanseokoh:
Also `ExposedPorts` in the container configuration.

chanseokoh:
#924: `Entrypoint` and `CMD`

GuustavoPaiva:
@chanseokoh I like using the base image ports if none is specified. 

Also, I think volumes must be inherited as well. After the PR for supporting volume configuration (#1215) is accepted we can do it.

chanseokoh:
@GuustavoPaiva yeah, we actually do need to inherit the ports and the volumes, which docker does with `FROM` in `Dockerfile`.

coollog:
We have two left: volumes and healthcheck

Updated original post with a TODO list.

chanseokoh:
I think we haven't handled ports either. Updated the TODO list.

chanseokoh:
Just in case, I am documenting here that there is one subtle to be careful about when enabling inheriting ports, volumes, etc. In `BuildImageSteps`, the code currently looks like

```java
      // inherit properties
      imageBuilder
          .addEnvironment(baseImage.getEnvironment())
[...]

Issue #15, opened 2018-01-29
Add abstraction for container configuration environment.


coollog:
From https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/issues/91

Container configuration environment was added as a list of `NAME=VALUE` strings in https://github.com/GoogleCloudPlatform/minikube-build-tools-for-java/pull/72. It should be changed to be a separate class to handle the environment rather than just a list of strings.

coollog:
Might still be useful for https://github.com/GoogleContainerTools/jib/issues/595

PR #716, opened 2018-07-25
Propagate base image's environment variable config to built image

TadCordle:
Part of #595. Also did a small bit of refactoring/error checking for translating between environment variable list/map


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -96,8 +100,15 @@
           buildAndCacheApplicationLayerSteps) {
         imageBuilder.addLayer(NonBlockingSteps.get(buildAndCacheApplicationLayerStep));
       }
+
+      // Use environment from base image if not configured
+      if (buildConfiguration.getEnvironment() == null) {
```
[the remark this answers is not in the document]
TadCordle: Possibly, I'm not sure what the plan is for that.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -67,6 +68,24 @@
...
+   */
+  @VisibleForTesting
+  @Nullable
+  static ImmutableList<String> environmentMapToList(@Nullable Map<String, String> environment) {
+    if (environment == null) {
+      return null;
+    }
+    ImmutableList.Builder<String> builder = ImmutableList.builder();
+    for (Map.Entry<String, String> environmentVariable : environment.entrySet()) {
+      builder.add(environmentVariable.getKey() + "=" + environmentVariable.getValue());
```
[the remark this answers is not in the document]
TadCordle: Yeah, I definitely think this can be more bullet proof. Actually I think this breaks if there's an `=` in the value, not just the name.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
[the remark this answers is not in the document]
coollog: Yea, I think environment variables should be overwritten only at the specific variable level, and not at the entire environment level.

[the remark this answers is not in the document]
coollog: So it could be like initialize `imageBuilder` with the base image, then apply overwrites from `buildConfiguration`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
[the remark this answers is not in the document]
coollog: Hmm I think we might want to guard this at the environment-setting level - so maybe we might need to have a separate `Environment` class to be this abstraction (like mentioned in #15)

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -49,6 +49,16 @@
...
 
+  /**
+   * Pattern used for parsing environment variables in the format {@code NAME=VALUE}. {@code NAME}
+   * should start with a letter and contain only letters (uppercase and lower) and numbers. {@code
+   * VALUE} can contain anything.
+   *
+   * <p>Example matches: NAME=VALUE, A12345=$$$$$
+   */
+  @VisibleForTesting
+  static final Pattern environmentPattern = Pattern.compile("([A-Za-z][A-Za-z0-9_]+)=(.*)");
```
coollog: I believe environment variable names can actually contain any character (http://pubs.opengroup.org/onlinepubs/000095399/basedefs/xbd_chap08.html) - `These strings have the form name=value; names shall not contain the character '='.`, so we should probably have this pattern be like `(.*?)=(.*)`.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/BuildImageStepTest.java
```diff
@@ -46,6 +47,8 @@
 
   @Mock private BuildConfiguration mockBuildConfiguration;
   @Mock private BuildLogger mockBuildLogger;
+  @Mock private PullBaseImageStep mockPullBaseImageStep;
+  @Mock private Image<Layer> mockBaseImage;
```
coollog: We could prob just have a real `Image` since there's nothing in Image that needs to be mocked.

jib-core/src/test/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslatorTest.java
```diff
@@ -110,6 +111,32 @@ public void testPortMapToList() throws BadContainerConfigurationFormatException
     }
   }
 
+  @Test
+  public void testJsonToImageTranslatorRegex() {
+    assertGoodPattern("NAME=VALUE", "NAME", "VALUE");
```
coollog: Might want to call this like `assertGoodEnvironmentPattern`

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 
 - Only builds non-empty layers ([#516](https://github.com/GoogleContainerTools/jib/pull/516/files))
 - Fixed slow image reference parsing ([#680](https://github.com/GoogleContainerTools/jib/pull/680))
+- Environment variable configuration is propagated from the base image ([#716](https://github.com/GoogleContainerTools/jib/pull/716))
```
coollog: nit: `Propagates environment variables from base image...`

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
TadCordle: [Here](https://github.com/opencontainers/runtime-spec/blob/master/config.md) it says:

> **env**: (array of strings, OPTIONAL) with the same semantics as IEEE Std 1003.1-2008's environ.

And in the link you shared:

"Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set and do not begin with a digit"

coollog: Yep, for environment variables used in shell utilities (like `bash`), the names shouldn't consist of anything besides the alphanumerics, but for process environment variables, the definition is more lenient to allow any character besides `=`.

TadCordle: reading is hard

TadCordle: Shouldn't we make sure the name has at least one character though? I think what you gave will match something like "=ABC". Is that valid?

coollog: That’s a good point - let’s change the star to a plus.

TadCordle: That matches "====", I think `([^=].*?)=(.*)` should work

coollog: Yea, I think having = in the value is okay

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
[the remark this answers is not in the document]
TadCordle: We can handle this when we handle #15 then?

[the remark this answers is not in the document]
coollog: Maybe we can just throw an `IllegalArgumentException` for now?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -49,6 +49,15 @@
...
   private static final Pattern portPattern = Pattern.compile("(\\d+)(?:/(tcp|udp))?");
 
+  /**
+   * Pattern used for parsing environment variables in the format {@code NAME=VALUE}. {@code NAME}
+   * should start with a letter and contain only letters (uppercase and lower) and numbers. {@code
+   * VALUE} can contain anything.
+   *
+   * <p>Example matches: NAME=VALUE, A12345=$$$$$
+   */
+  @VisibleForTesting static final Pattern environmentPattern = Pattern.compile("([^=].*?)=(.*)");
```
chanseokoh: This seems weird. `([^=].*?)` matches any string that doesn't start with `=`. I think it's conceptually correct to say `([^=]+)`.

chanseokoh: So as a concrete example,

```java
    Pattern pattern = Pattern.compile("([^=].*?)");
    Matcher matcher = pattern.matcher("ab==c==d");
    matcher.matches();
    System.out.println(matcher.group(1));
```

It prints out `ab==c==d`.

TadCordle: Yeah I guess within the context of the rest of the expression, a name containing an `=` would only be detected if it started with an `=`. But yours makes more sense and is easier to read.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -139,7 +147,14 @@
...
     if (containerConfigurationTemplate.getContainerEnvironment() != null) {
       for (String environmentVariable : containerConfigurationTemplate.getContainerEnvironment()) {
-        imageBuilder.addEnvironmentVariableDefinition(environmentVariable);
+        Matcher matcher = environmentPattern.matcher(environmentVariable);
+        if (!matcher.matches()) {
+          throw new BadContainerConfigurationFormatException(
+              "Invalid environment variable definition: " + environmentVariable);
+        }
+        String name = matcher.group(1);
+        String value = matcher.group(2);
```
briandealwis: When patterns are far away from their use, it can help to use a named group (e.g., `matcher.group("name")` and `matcher.group("value")`).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -96,8 +101,16 @@
...
         imageBuilder.addLayer(NonBlockingSteps.get(buildAndCacheApplicationLayerStep));
       }
+
+      // Use environment from base image
+      imageBuilder.setEnvironment(
+          NonBlockingSteps.get(pullBaseImageStep).getBaseImage().getEnvironment());
+      if (buildConfiguration.getEnvironment() != null) {
+        for (Map.Entry<String, String> entry : buildConfiguration.getEnvironment().entrySet()) {
+          imageBuilder.setEnvironmentVariable(entry.getKey(), entry.getValue());
+        }
```
briandealwis: I wonder if this could be simplified by instead having just a `imageBuilder.addEnvironment()`:
```diff
+      // Initialize to environment from base image
+      imageBuilder.addEnvironment(
+          NonBlockingSteps.get(pullBaseImageStep).getBaseImage().getEnvironment());
+      // override with configured environment
+      if (buildConfiguration.getEnvironment() != null) {
+         imageBuilder.addEnvironment(buildConfiguration.getEnvironment());
+      }

File: jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java [hunk 10]
```diff
@@ -67,6 +68,27 @@
...
+  @Nullable
+  static ImmutableList<String> environmentMapToList(@Nullable Map<String, String> environment) {
+    if (environment == null) {
+      return null;
+    }
+    ImmutableList.Builder<String> builder = ImmutableList.builder();
+    for (Map.Entry<String, String> environmentVariable : environment.entrySet()) {
+      if (environmentVariable.getKey().contains("=")) {
+        throw new IllegalArgumentException("Environment variable name cannot contain '='");
+      }
```
briandealwis: This is exactly what `Preconditions.checkArgument()` does?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -67,6 +68,27 @@
...
+      builder.add(environmentVariable.getKey() + "=" + environmentVariable.getValue());
+    }
+    return builder.build();
```
briandealwis: If you like streams, you could simplify this a bit:
```
    Preconditions.checkArgument(
        Iterables.any(environment.keySet(), s -> s.contains("=")),
        "Environment variable name cannot contain '='");
    return environment
        .entrySet()
        .stream()
        .map(entry -> entry.getKey() + "=" + entry.getValue())
        .collect(ImmutableList.toImmutableList());
```

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -67,6 +68,27 @@
...
+  @VisibleForTesting
+  @Nullable
+  static ImmutableList<String> environmentMapToList(@Nullable Map<String, String> environment) {
+    if (environment == null) {
+      return null;
+    }
+    ImmutableList.Builder<String> builder = ImmutableList.builder();
+    for (Map.Entry<String, String> environmentVariable : environment.entrySet()) {
+      if (environmentVariable.getKey().contains("=")) {
+        throw new IllegalArgumentException("Environment variable name cannot contain '='");
```
coollog: It might be good to include the illegal name in the message.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -49,6 +49,14 @@
    */
   private static final Pattern portPattern = Pattern.compile("(\\d+)(?:/(tcp|udp))?");
 
+  /**
+   * Pattern used for parsing environment variables in the format {@code NAME=VALUE}. {@code NAME}
+   * should not contain an '='.
+   *
+   * <p>Example matches: NAME=VALUE, A12345=$$$$$
+   */
+  @VisibleForTesting static final Pattern environmentPattern = Pattern.compile("([^=]+)=(.*)");
```
coollog: nit: constants should be in all caps

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
coollog: +1 We should probably do this for some of the other regex patterns too.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -49,18 +50,14 @@
...
-    public Builder<T> setEnvironment(@Nullable Map<String, String> environment) {
-      if (environment == null) {
-        this.environmentBuilder = ImmutableList.builder();
-      } else {
-        for (Map.Entry<String, String> environmentVariable : environment.entrySet()) {
-          setEnvironmentVariable(environmentVariable.getKey(), environmentVariable.getValue());
-        }
+    public Builder<T> addEnvironment(@Nullable Map<String, String> environment) {
+      if (environment != null) {
+        this.environmentBuilder.putAll(ImmutableMap.copyOf(environment));
```
chanseokoh: I'm thinking this is no different from and just could be `environmentBuilder.putAll(environment)`, am I right? You are just putting references to objects (`String`s in this case) into the builder, not like you are doing a deep copy of `String` instances. (And `String`s are immutable, so there is no point of copying, actually.) Am I right?

TadCordle: Sounds right to me.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -67,6 +69,27 @@
...
+   * @return the list
+   */
+  @VisibleForTesting
+  @Nullable
+  static ImmutableList<String> environmentMapToList(@Nullable Map<String, String> environment) {
+    if (environment == null) {
+      return null;
+    }
+    Preconditions.checkArgument(
+        environment.keySet().stream().noneMatch(s -> s.contains("=")),
```
chanseokoh: `key -> key.contains("=")` seems more readable.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -47,7 +47,17 @@
...
+      Pattern.compile("(?<portNum>\\d+)(?:/(?<protocol>tcp|udp))?");
+
+  /**
+   * Pattern used for parsing environment variables in the format {@code NAME=VALUE}. {@code NAME}
+   * should not contain an '='.
+   *
+   * <p>Example matches: NAME=VALUE, A12345=$$$$$
+   */
+  @VisibleForTesting
+  static final Pattern environmentPattern = Pattern.compile("(?<name>[^=]+)=(?<value>.*)");
```
chanseokoh: `ENVIRONMENT_PATTERN` too.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -96,8 +100,14 @@
           buildAndCacheApplicationLayerSteps) {
         imageBuilder.addLayer(NonBlockingSteps.get(buildAndCacheApplicationLayerStep));
       }
+
+      // Use environment from base image
+      imageBuilder.addEnvironment(
+          NonBlockingSteps.get(pullBaseImageStep).getBaseImage().getEnvironment());
+      if (buildConfiguration.getEnvironment() != null) {
+        imageBuilder.addEnvironment(buildConfiguration.getEnvironment());
```
chanseokoh: Since the parameter of `addEnvironment()` is `@Nullable`, we could probably remove the null check.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -67,6 +69,27 @@
...
+   * @return the list
+   */
+  @VisibleForTesting
+  @Nullable
+  static ImmutableList<String> environmentMapToList(@Nullable Map<String, String> environment) {
+    if (environment == null) {
+      return null;
+    }
+    Preconditions.checkArgument(
+        environment.keySet().stream().noneMatch(key -> key.contains("=")),
```
coollog: Or, even more clear may be s/key/variableName/

TadCordle: I like `key` more since it's descriptive about which part of the map itself it's dealing with. And we can still tell that the key corresponds to the name with the error message right below.

PR #717, opened 2018-07-25
Fix gradle warnings about @Input annotations on java.io.File

### LINKED ISSUE #705: Fix warnings when building gradle plugin
`$ ./gradlew clean goJF build` causes two types of warnings at the moment.

- [x] Unchecked future result
```
> Task :compileTestJava 
(...)/jib/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java:40: warning: [FutureReturnValueIgnored] Return value of methods returning Future must be checked. Ignoring returned Futures suppresses exceptions thrown from the code that completes the Future.
    executorService.submit(this::serve200);
                          ^
    (see http://errorprone.info/bugpattern/FutureReturnValueIgnored)
  Did you mean to remove this line?
1 warning
```

- [x] `@Input` annotation used on `java.io.File`
```
> Task :validateTaskProperties 
Task property validation finished with warnings:
  - Warning: Task type 'com.google.cloud.tools.jib.gradle.BuildDockerTask': property 'jib.extraDirectory' has @Input annotation used on property of type java.io.File.
  - Warning: Task type 'com.google.cloud.tools.jib.gradle.BuildImageTask': property 'jib.extraDirectory' has @Input annotation used on property of type java.io.File.
  - Warning: Task type 'com.google.cloud.tools.jib.gradle.BuildTarTask': property 'jib.extraDirectory' has @Input annotation used on property of type java.io.File.
  - Warning: Task type 'com.google.cloud.tools.jib.gradle.DockerContextTask': property 'jib.extraDirectory' has @Input annotation used on property of type java.io.File.
```

TadCordle:
Fixes #705


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -185,7 +184,7 @@ public void setAllowInsecureRegistries(boolean allowInsecureRegistries) {
     this.allowInsecureRegistries.set(allowInsecureRegistries);
   }
 
-  public void setExtraDirectory(File extraDirectory) {
+  public void setExtraDirectory(Path extraDirectory) {
```
coollog: Does this work for `build.gradle`? I remember we used `File` originally because `Path` doesn't work with the Gradle project `file` method.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -272,7 +271,14 @@ boolean getAllowInsecureRegistries() {
   }
 
   @Input
-  File getExtraDirectory() {
+  String getExtraDirectory() {
```
coollog: What about using `Path` directly here?

TadCordle: Gets the same warnings as File.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
TadCordle: Oh I didn't check, I thought an integration test would have covered this already. One sec.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -184,8 +187,8 @@ public void setAllowInsecureRegistries(boolean allowInsecureRegistries) {
     this.allowInsecureRegistries.set(allowInsecureRegistries);
   }
 
-  public void setExtraDirectory(Path extraDirectory) {
-    this.extraDirectory.set(extraDirectory);
+  public void setExtraDirectory(String extraDirectory) {
+    this.extraDirectory.set(projectDir.resolve(extraDirectory));
```
TadCordle: So this I'm not too sure of. The way I changed it here means that people can only configure paths relative to the project directory, which makes things easier in general, but also doesn't let people add absolute paths (maybe that's ok?)

coollog: I think we can just have this take a `File`?

coollog: That way users can use the `file` method in Gradle.

PR #722, opened 2018-07-26
Add 'auth' configuration to maven plugin

### LINKED ISSUE #693: Setting auth credentials from command line or environment variable
In the case jib-maven-plugin is used in a CI environment, it can be useful to pass the authentication credentials directly via the command line or via environment variable.

For example with Gitlab CI, it is the case that CI jobs have environment variables with temporary tokens to push images to the Gitlab Registry.

Currently, it seems to me that the only way to take advantage of these is either via `docker login` (which means I need docker installed now) or via some hack by generating a custom `settings.xml`.

It would be great either to have jib-maven-plugin accept credentials via command line or to document what is the preferred way to achieve the requirement.

TadCordle:
So now I think the order of auth checking is something like:
1. Cred helper in build config
2. Username/password system properties
3. Username/password in build config
4. Maven settings
5. Common cred helpers
6. Docker config

Fix #693

### DIFF
--- a/jib-maven-plugin/CHANGELOG.md
+++ b/jib-maven-plugin/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- <to><auth> and <from><auth> parameters with <username> and <password> fields for simple authentication, similar to the Gradle plugin ([#693](https://github.com/GoogleContainerTools/jib/issues/693))
+- Can set credentials via commandline using `jib.to.auth.username`, `jib.to.auth.password`, `jib.from.auth.username`, and `jib.from.auth.password` system properties ([#693](https://github.com/GoogleContainerTools/jib/issues/693))
+
 ### Changed
 
 - Propagates environment variables from the base image ([#716](https://github.com/GoogleContainerTools/jib/pull/716))
--- a/jib-maven-plugin/CHANGELOG.md
+++ b/jib-maven-plugin/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- <to><auth> and <from><auth> parameters with <username> and <password> fields for simple authentication, similar to the Gradle plugin ([#693](https://github.com/GoogleContainerTools/jib/issues/693))
+- `<to><auth>` and `<from><auth>` parameters with `<username>` and `<password>` fields for simple authentication, similar to the Gradle plugin ([#693](https://github.com/GoogleContainerTools/jib/issues/693))
 - Can set credentials via commandline using `jib.to.auth.username`, `jib.to.auth.password`, `jib.from.auth.username`, and `jib.from.auth.password` system properties ([#693](https://github.com/GoogleContainerTools/jib/issues/693))
 
 ### Changed
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -26,6 +26,7 @@ import com.google.cloud.tools.jib.frontend.ExposedPortsParser;
 import com.google.cloud.tools.jib.frontend.HelpfulSuggestions;
 import com.google.cloud.tools.jib.frontend.JavaEntrypointConstructor;
 import com.google.cloud.tools.jib.frontend.SystemPropertyValidator;
+import com.google.cloud.tools.jib.http.Authorization;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.registry.RegistryClient;
 import com.google.cloud.tools.jib.registry.credentials.RegistryCredentials;
@@ -76,8 +77,11 @@ public class BuildDockerMojo extends JibPluginConfiguration {
     MavenSettingsServerCredentials mavenSettingsServerCredentials =
         new MavenSettingsServerCredentials(
             Preconditions.checkNotNull(session).getSettings(), settingsDecrypter, mavenBuildLogger);
+    Authorization fromAuthorization = getBaseImageAuth();
     RegistryCredentials knownBaseRegistryCredentials =
-        mavenSettingsServerCredentials.retrieve(baseImage.getRegistry());
+        fromAuthorization != null
+            ? new RegistryCredentials("<jib><from><auth>", fromAuthorization)
+            : mavenSettingsServerCredentials.retrieve(baseImage.getRegistry());
 
     String mainClass = mavenProjectProperties.getMainClass(this);
 
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -80,7 +80,8 @@ public class BuildDockerMojo extends JibPluginConfiguration {
     Authorization fromAuthorization = getBaseImageAuth();
     RegistryCredentials knownBaseRegistryCredentials =
         fromAuthorization != null
-            ? new RegistryCredentials("<jib><from><auth>", fromAuthorization)
+            ? new RegistryCredentials(
+                "jib-maven-plugin <from><auth> configuration", fromAuthorization)
             : mavenSettingsServerCredentials.retrieve(baseImage.getRegistry());
 
     String mainClass = mavenProjectProperties.getMainClass(this);
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
@@ -97,12 +97,13 @@ public class BuildImageMojo extends JibPluginConfiguration {
     Authorization fromAuthorization = getBaseImageAuth();
     RegistryCredentials knownBaseRegistryCredentials =
         fromAuthorization != null
-            ? new RegistryCredentials("<jib><from><auth>", fromAuthorization)
+            ? new RegistryCredentials(
+                "jib-maven-plugin <from><auth> configuration", fromAuthorization)
             : mavenSettingsServerCredentials.retrieve(baseImage.getRegistry());
     Authorization toAuthorization = getTargetImageAuth();
     RegistryCredentials knownTargetRegistryCredentials =
         toAuthorization != null
-            ? new RegistryCredentials("<jib><to><auth>", toAuthorization)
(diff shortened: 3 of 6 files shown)


jib-maven-plugin/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- <to><auth> and <from><auth> parameters with <username> and <password> fields for simple authentication, similar to the Gradle plugin ([#693](https://github.com/GoogleContainerTools/jib/issues/693))
```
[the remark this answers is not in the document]
TadCordle: > Also I'd put "Maven" somewhere to say this applies to Maven users.

It's in the Maven changelog, so I think it's ok that it doesn't mention Maven. Also looks like I forgot to wrap the parameters in \`\`, oops.

jib-maven-plugin/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
...
+- Can set credentials via commandline using `jib.to.auth.username`, `jib.to.auth.password`, `jib.from.auth.username`, and `jib.from.auth.password` system properties ([#693](https://github.com/GoogleContainerTools/jib/issues/693))
```
chanseokoh: Q: are these system probably already work in Gradle. (But I believe Gradle users can already do this on the command line, so the system properties aren't need at all, perhaps?)

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -50,6 +73,8 @@
     private String image = "gcr.io/distroless/java";
 
     @Nullable @Parameter private String credHelper;
+
+    @Parameter private AuthConfiguration auth = new AuthConfiguration();
```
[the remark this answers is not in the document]
briandealwis: They can't as they're injected by Plexus after object creation.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -39,6 +41,27 @@
 /** Defines the configuration parameters for Jib. Jib {@link Mojo}s should extend this class. */
 abstract class JibPluginConfiguration extends AbstractMojo {
 
+  /** Used to configure {@code from.auth} and {@code to.auth} parameters. */
+  public static class AuthConfiguration {
```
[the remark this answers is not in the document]
TadCordle: I tried making all the inner classes non-public and Maven failed:

```
Unable to parse configuration of mojo com.google.cloud.tools:jib-maven-plugin:0.9.8-SNAPSHOT:dockerBuild for parameter container: Cannot create instance of class com.google.cloud.tools.jib.maven.JibPluginConfiguration$ContainerParameters
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -89,13 +90,20 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-        mavenSettingsServerCredentials.retrieve(baseImage.getRegistry());
+        fromAuthorization != null
+            ? new RegistryCredentials("<jib><from><auth>", fromAuthorization)
+            : mavenSettingsServerCredentials.retrieve(baseImage.getRegistry());
+    Authorization toAuthorization = getTargetImageAuth();
     RegistryCredentials knownTargetRegistryCredentials =
-        mavenSettingsServerCredentials.retrieve(targetImage.getRegistry());
+        toAuthorization != null
+            ? new RegistryCredentials("<jib><to><auth>", toAuthorization)
+            : mavenSettingsServerCredentials.retrieve(targetImage.getRegistry());
```
briandealwis: I wonder if this code could be pushed up to `JibPluginConfiguration` and added by providing the `BuildConfiguration.Builder` instance. For example
```
  setBaseImageCredentialProvisioning(baseConfigurationBuilder);
  setTargetImageCredentialProvisioning(baseConfigurationBuilder);
```

It would reduce code duplicated across the mojos.

jib-maven-plugin/CHANGELOG.md
chanseokoh: Also, I remember there are other properties like `-Dimage=...`, which I'm gussing corresponds to this code.
```
  @Parameter(property = "image")
  private ToConfiguration to = new ToConfiguration();
```
Are they different from system properties? Can we make `jib.to.auth.username` work in this way? Although not sure which one is for which situation, honestly.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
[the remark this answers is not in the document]
chanseokoh: Makes sense. May need to revert the changes, sorry.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
TadCordle: I think in general we're trying to keep `JibPluginConfiguration` as a class that just contains all the Maven configuration properties without much extra logic. If anything I should probably be moving the system property/build config choosing stuff out of there, but I left it since I'm essentially using the system properties as a replacement for commandline properties (because I couldn't figure out how to tag the correct auth parameters as such, and system properties are easier and identical as far as setting them is concerned). We can work on reducing duplicate code when we work on the `jib-plugins-common` module.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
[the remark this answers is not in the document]
TadCordle: No problem, I skipped over this one because I wasn't sure either.

jib-maven-plugin/CHANGELOG.md
TadCordle: I was trying to figure out how to do this; Q told me that the `property` field works by calling a `set()` method within `ToConfiguration`. The problem I hit is that `auth` is also within `ToConfiguration`, so I wouldn't be able to make another `set` method. So to avoid the headache I just went with system properties instead since they're set in the same way.

chanseokoh: I see. I am curious if @coollog knows something about it.

TadCordle: I'll try messing around with it some more.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -76,8 +77,11 @@ public void execute() throws MojoExecutionException {
     MavenSettingsServerCredentials mavenSettingsServerCredentials =
         new MavenSettingsServerCredentials(
             Preconditions.checkNotNull(session).getSettings(), settingsDecrypter, mavenBuildLogger);
+    Authorization fromAuthorization = getBaseImageAuth();
     RegistryCredentials knownBaseRegistryCredentials =
-        mavenSettingsServerCredentials.retrieve(baseImage.getRegistry());
+        fromAuthorization != null
+            ? new RegistryCredentials("<jib><from><auth>", fromAuthorization)
```
coollog: The credential source might be clearer as `jib-maven-plugin <from><auth> configuration` or something.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -59,6 +84,8 @@
 
     @Nullable @Parameter private String credHelper;
 
+    @Parameter private AuthConfiguration auth = new AuthConfiguration();
```
coollog: If the below are final, this could be final too?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -141,12 +168,12 @@ void handleDeprecatedParameters(BuildLogger logger) {
   @Parameter(defaultValue = "${session}", readonly = true)
   MavenSession session;
 
-  @Parameter private FromConfiguration from = new FromConfiguration();
+  @Parameter private final FromConfiguration from = new FromConfiguration();
```
coollog: nit: final variables up top - also, it looks like the `handleDeprecatedParameters` method is out of place, and the package-private `session` should be above private `project`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -183,6 +210,24 @@ String getBaseImageCredentialHelperName() {
...
+   * Gets an {@code Authorization} from system properties, or from the {@code <from><auth>}
+   * configuration if the system properties are not set, or {@code null} if neither are set.
+   *
+   * @return a new {@code Authorization} from the configured username and password, or {@code null}
+   *     if one isn't configured.
+   */
+  @Nullable
+  Authorization getBaseImageAuth() {
+    // System property takes priority over build configuration
+    String commandlineUsername = System.getProperty("jib.from.auth.username");
```
coollog: Can consolidate this logic with the `getTargetImageAuth` under a private static method.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
TadCordle: Oh, I think I meant to change those back.

TadCordle:
> Username/password system properties
> Username/password in build config
> Cred helper in build config
> Maven settings

Considering maven settings credentials and username/password are passed in using the same "known credentials" variable, splitting them up should probably wait until another PR.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -333,4 +307,26 @@ void setTargetImage(@Nullable String targetImage) {
...
+   * tries build configuration, otherwise returns null.
+   *
+   * @param usernameProperty the name of the username system property
+   * @param passwordProperty the name of the password system property
+   * @param auth the configured credentials
+   * @return a new {@link Authorization} from the system properties or build configuration, or
+   *     {@code null} if neither is configured.
+   */
+  @Nullable
+  private static Authorization getImageAuth(
```
coollog: nit: private static method up top

PR #725, opened 2018-07-27
Update readme for 0.9.8

TadCordle:
To be merged after 0.9.8 release.


jib-maven-plugin/README.md
```diff
@@ -311,6 +320,43 @@ Configure credential helpers to use by specifying them as a `credHelper` for the
 </configuration>
 ```
 
+#### Using Specific Credentials
+
+You can specify credentials directly in the <auth> parameter for the `from` and/or `to` images. In the example below, `to` credentials are retrieved from the `REGISTRY_USERNAME` and `REGISTRY_PASSWORD` environment variables.
```
chanseokoh: Place `<auth>` in the black quotes. (Otherwise not visible at all.)

jib-maven-plugin/README.md
```diff
@@ -311,6 +320,43 @@ Configure credential helpers to use by specifying them as a `credHelper` for the
...
+    <auth>
+      <username>my_username</username>
+      <password>my_password</password>
+    </auth>
+  </from>
+  <to>
+    <image>gcr.io/my-gcp-project/my-app</image>
+    <auth>
+      <username>${env.REGISTRY_USERNAME}</username>
+      <password>${env.REGISTRY_PASSWORD}</password>
```
chanseokoh: Aha, this is the way to get env variables. Very good to know.

jib-maven-plugin/README.md
chanseokoh: Perhaps worth mentioning to use this as a last resort, as the plain password is entered into the file? (The command line apporach is not so different, as people on the same system can see them too.)

jib-maven-plugin/README.md
```diff
@@ -311,6 +320,44 @@ Configure credential helpers to use by specifying them as a `credHelper` for the
...
+Alternatively, you can specify credentials via commandline using the following system properties.
+
+Property | Description
+--- | ---
+`-Djib.from.auth.username` | Username for base image registry.
+`-Djib.from.auth.password` | Password for base image registry.
+`-Djib.to.auth.username` | Username for target image registry.
+`-Djib.to.auth.password` | Password for target image registry.
+
+e.g. `mvn package jib:build -Djib.to.auth.username=user -Djib.to.auth.password=pass`
```
coollog: Let's prefer to use `compile` instead of `package` for example code.

jib-maven-plugin/README.md
```diff
@@ -311,6 +320,44 @@ Configure credential helpers to use by specifying them as a `credHelper` for the
...
+
+**Note** This method of authentication should be used only as a last resort, as it is insecure to make your password visible in plain text.
```
coollog: nit: colon after Note

TadCordle:
Should be good to merge once the Maven release is done.

@chanseokoh if you want, you can also add what you'd mentioned here https://github.com/GoogleContainerTools/jib/issues/643#issuecomment-409010147

chanseokoh:
> @chanseokoh if you want, you can also add what you'd mentioned here #643 (comment)

Thanks for the reminder. I'll do it after fixing #767.

PR #732, opened 2018-07-27
Split BuildConfiguration into multiple classes

### LINKED ISSUE #684: Group configuration into separate classes.
In `BuildConfiguration`, the configuration should be grouped into separate classes in preparation for the library:

- [x] `javaArguments` (should probably be renamed), `environmentMap`, `exposedPort`, `entrypoint`, should be moved to a separate immutable `ContainerConfiguration`
- [x] `baseImageReference`, `baseImageCredentialHelper`, `knownBaseRegistryCredentials` should be moved to a separate `ImageConfiguration`
- [x] same for the target image configuration

### LINKED ISSUE #709: Improve BuildConfiguration parameter layout.
From https://github.com/GoogleContainerTools/jib/pull/516#pullrequestreview-140073132

- [x] Reorder parameters to be more intuitive.
- [x] Remove java-specific naming like `javaArguments`.

TadCordle:
Fixes #684, fixes #709

Breaks `BuildConfiguration` parameters into new `ImageConfiguration` (`from`/`to`) and `ContainerConfiguration` classes.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
@@ -49,16 +49,34 @@ public class BuildConfiguration {
       this.buildLogger = buildLogger;
     }
 
+    /**
+     * Sets the base image configuration.
+     *
+     * @param imageConfiguration the {@link ImageConfiguration} describing the base image
+     * @return this
+     */
     public Builder setBaseImageConfiguration(ImageConfiguration imageConfiguration) {
       this.baseImageConfiguration = imageConfiguration;
       return this;
     }
 
+    /**
+     * Sets the target image configuration.
+     *
+     * @param imageConfiguration the {@link ImageConfiguration} describing the target image
+     * @return this
+     */
     public Builder setTargetImageConfiguration(ImageConfiguration imageConfiguration) {
       this.targetImageConfiguration = imageConfiguration;
       return this;
     }
 
+    /**
+     * Sets configuration parameters for the container.
+     *
+     * @param containerConfiguration the {@link ContainerConfiguration}
+     * @return this
+     */
     public Builder setContainerConfiguration(ContainerConfiguration containerConfiguration) {
       this.containerConfiguration = containerConfiguration;
       return this;
@@ -121,7 +139,11 @@ public class BuildConfiguration {
       return this;
     }
 
-    /** @return the corresponding build configuration */
+    /**
+     * Builds a new {@link BuildConfiguration} using the parameters passed into the builder.
+     *
+     * @return the corresponding build configuration
+     */
     public BuildConfiguration build() {
       // Validates the parameters.
       List<String> errorMessages = new ArrayList<>();
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
@@ -192,7 +192,7 @@ public class BuildConfiguration {
   @Nullable private final String targetImageCredentialHelperName;
   @Nullable private final RegistryCredentials knownTargetRegistryCredentials;
   private final Instant creationTime;
-  @Nullable private final ImmutableList<String> mainArguments;
+  @Nullable private final ImmutableList<String> programArguments;
   @Nullable private final ImmutableMap<String, String> environmentMap;
   @Nullable private final ImmutableList<Port> exposedPorts;
   private final Class<? extends BuildableManifestTemplate> targetFormat;
@@ -213,7 +213,7 @@ public class BuildConfiguration {
       @Nullable RegistryCredentials knownTargetRegistryCredentials,
       Instant creationTime,
       @Nullable ImmutableList<String> entrypoint,
-      @Nullable ImmutableList<String> mainArguments,
+      @Nullable ImmutableList<String> programArguments,
       @Nullable ImmutableMap<String, String> environmentMap,
       @Nullable ImmutableList<Port> exposedPorts,
       Class<? extends BuildableManifestTemplate> targetFormat,
@@ -229,7 +229,7 @@ public class BuildConfiguration {
     this.targetImageCredentialHelperName = targetImageCredentialHelperName;
     this.knownTargetRegistryCredentials = knownTargetRegistryCredentials;
     this.creationTime = creationTime;
-    this.mainArguments = mainArguments;
+    this.programArguments = programArguments;
     this.environmentMap = environmentMap;
     this.exposedPorts = exposedPorts;
     this.targetFormat = targetFormat;
@@ -316,8 +316,8 @@ public class BuildConfiguration {
    * @return the list of arguments, or {@code null} if not set
    */
   @Nullable
-  public ImmutableList<String> getMainArguments() {
-    return mainArguments;
+  public ImmutableList<String> getProgramArguments() {
+    return programArguments;
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
@@ -19,18 +19,13 @@ package com.google.cloud.tools.jib.configuration;
 import com.google.cloud.tools.jib.builder.BuildLogger;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.json.BuildableManifestTemplate;
-import com.google.cloud.tools.jib.image.json.V22ManifestTemplate;
 import com.google.cloud.tools.jib.registry.credentials.RegistryCredentials;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Iterables;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
-import java.util.Objects;
 import javax.annotation.Nullable;
 import javax.lang.model.SourceVersion;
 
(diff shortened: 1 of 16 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -40,94 +35,33 @@
...
-    }
-
-    public Builder setBaseImageCredentialHelperName(@Nullable String credentialHelperName) {
-      baseImageCredentialHelperName = credentialHelperName;
-      return this;
-    }
-
-    public Builder setTargetImageCredentialHelperName(@Nullable String credentialHelperName) {
-      targetImageCredentialHelperName = credentialHelperName;
+    public Builder setBaseImageConfiguration(ImageConfiguration imageReference) {
```
coollog: `imageConfiguration`

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -40,94 +35,33 @@
...
-    @Nullable private RegistryCredentials knownBaseRegistryCredentials;
-    @Nullable private ImageReference targetImageReference;
-    @Nullable private String targetImageCredentialHelperName;
-    @Nullable private RegistryCredentials knownTargetRegistryCredentials;
-    // TODO: Should rename to not be java-specific.
-    @Nullable private ImmutableList<String> javaArguments;
-    @Nullable private ImmutableMap<String, String> environmentMap;
-    @Nullable private ImmutableList<Port> exposedPorts;
-    private Class<? extends BuildableManifestTemplate> targetFormat = V22ManifestTemplate.class;
+    private ImageConfiguration baseImageConfiguration = new ImageConfiguration();
```
coollog: We should probably have these be the Builder classes or be `null` to start, like how `CacheConfiguration` is `null` to start.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -228,22 +137,22 @@ public BuildConfiguration build() {
           }
           return new BuildConfiguration(
               buildLogger,
-              creationTime,
               baseImageReference,
-              baseImageCredentialHelperName,
-              knownBaseRegistryCredentials,
+              baseImageConfiguration.getCredentialHelper(),
```
coollog: These should just use the configuration objects directly, since they are immutable.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -391,14 +296,28 @@ public RegistryCredentials getKnownTargetRegistryCredentials() {
     return knownTargetRegistryCredentials;
   }
 
+  public Instant getCreationTime() {
```
coollog: Hence, down here, these could all be just a `getContainerConfiguration`.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  private final Instant creationTime;
+  @Nullable private final ImmutableList<String> entrypoint;
+  @Nullable private final ImmutableList<String> programArguments;
+  @Nullable private final ImmutableMap<String, String> environmentMap;
+  @Nullable private final ImmutableList<Port> exposedPorts;
+  private final Class<? extends BuildableManifestTemplate> targetFormat;
+
+  ContainerConfiguration() {
```
coollog: With the above change to use the Builder or be nullable, this can be private.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/** Immutable configuration options for the container. */
+public class ContainerConfiguration {
+
+  /** Builder for instantiating a {@link ContainerConfiguration}. */
+  public static class Builder {
+
+    private Instant creationTime = Instant.EPOCH;
```
coollog: I have a feeling we might want to put this default somewhere besides the configuration classes, but that could be a separate discussion/PR.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
TadCordle: Any reason why? Seems like making them `null` to start just adds an extra check later (`baseImageConfiguration.getImage()` -> `baseImageConfiguration == null ? null : baseImageConfiguration.getImage()`). Seems safer to have them default to instantiated but empty.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
TadCordle: Are we expecting the user to use these getters?

TadCordle: Wait nevermind

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
coollog: I think it's mostly to keep it consistent with what we have for `CacheConfiguration`, for instance. The reason that one is nullable is because initializing one (for example, the temporary directory one) may be a costly operation and we don't want to have our configuration code be coupled with possibly-costly initialization implementations. Also, in terms of semantics, having a empty `ImageConfiguration`, for example, to start, doesn't actually mean use defaults, since defaults right now are actually provided at the plugin side. We eventually want to probably add validation for different configuration types, so for `ImageConfiguration`, it should not be instantiated without an actual `ImageReference`, and having an empty constructor would mean either there would be this code path where that validation is not performed, or that the validation would need to be duplicated in the `Builder.build` and in the constructor(s) as well.

coollog: `null` also provides the semantics at the higher-level configuration building (`BuildConfiguration`), where it can simply validate on whether required fields were `null` instead of having to understand the details of the lower-level configuration to know which fields should be validated, or call out to a lower-level configuration's validation, which probably should be performed when it was built in the first place.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
TadCordle: This will require changing code in a bunch of other files, so I'll leave this for another PR.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -115,40 +114,36 @@ public Builder setLayerConfigurations(List<LayerConfiguration> layerConfiguratio
...
-          if (baseImageReference.usesDefaultTag()) {
+          if (baseImageConfiguration.getImage().usesDefaultTag()) {
             buildLogger.warn(
                 "Base image '"
-                    + baseImageReference
+                    + baseImageConfiguration.getImage()
                     + "' does not use a specific image digest - build may not be reproducible");
           }
+
+          // If a container configuration hasn't been specified, use an empty one by default
```
coollog: Can this be nullable too?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -0,0 +1,196 @@
...
+      return this;
+    }
+
+    /**
+     * Sets the commandline arguments for main.
+     *
+     * @param programArguments the list of arguments
+     * @return this
+     */
+    public Builder setProgramArguments(@Nullable List<String> programArguments) {
```
chanseokoh: For the others like `setEnvironment` or `setExposedPorts`, we set the field to `null` if the given argument is `null`. It's not the case here. Explanation?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -0,0 +1,196 @@
...
+     * @param environmentMap the map
+     * @return this
+     */
+    public Builder setEnvironment(@Nullable Map<String, String> environmentMap) {
+      if (environmentMap == null) {
+        this.environmentMap = null;
+      } else {
+        Preconditions.checkArgument(
+            !Iterables.any(environmentMap.keySet(), Objects::isNull)
+                && !Iterables.any(environmentMap.values(), Objects::isNull));
```
chanseokoh: Huh, now I see I missed this in the original PR a while ago. This should actually be broken into two `checkArgument()`s, so that we know which one was at fault if these are violated.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
TadCordle: If we leave it null, we'd need to add extra checks to a bunch of the getters; or worse, if we consolidate the getters into a single `getContainerConfiguration` later, then that's an extra check we'd need to do everywhere we try to get something from the container configuration. And if we instead error on a `ContainerConfiguration` not being passed into the builder like we do for the base/target image configuration, that's just a pain for the user to have to pass in a default `ContainerConfiguration` themselves. It seems better to me to instantiate the container configuration with null/default values if the user hasn't configured it themselves. Plus the worry about constructing an empty `ContainerConfiguration` being potentially expensive is purely hypothetical at this point.

chanseokoh: Hmm... so this is like a default value? Isn't it customary to set default values when declaring fields?

```java
  private ContainerConfiguration containerConfiguration = ContainerConfiguration.builder().build();
```

PR #733, opened 2018-07-27
Make allowInsecureRegistries follow Docker semantics

### LINKED ISSUE #643: Make allowInsecureRegistries follow Docker semantics for use-insecure-registries

chanseokoh:
Fixes #643.

Before:
`call(URL)` embedded the logic for handling `allowInsecureRegistries`.

Now:
The logic is broken out of `call(URL)` into `callWithInsecureRegistryHandling()`. `call(URL)` always contacts the endpoint and is irrelevant of `allowInsecureRegistries`. `callWithInsecureRegistryHandling()` is the one that handles `allowInsecureRegistries`.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
@@ -30,6 +30,7 @@ import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.security.GeneralSecurityException;
 import java.util.function.Function;
 import javax.annotation.Nullable;
 import javax.net.ssl.SSLPeerUnverifiedException;
@@ -50,6 +51,10 @@ class RegistryEndpointCaller<T> {
   @VisibleForTesting static final int STATUS_CODE_PERMANENT_REDIRECT = 308;
 
   private static final String DEFAULT_PROTOCOL = "https";
+
+  private static boolean isHttpsProtocol(URL url) {
+    return "https".equals(url.getProtocol());
+  }
 
   private final URL initialRequestUrl;
   private final String userAgent;
@@ -125,7 +130,39 @@ class RegistryEndpointCaller<T> {
    */
   @Nullable
   T call() throws IOException, RegistryException {
-    return call(initialRequestUrl);
+    return callWithInsecureRegistryHandling(initialRequestUrl);
+  }
+
+  @VisibleForTesting
+  @Nullable
+  T callWithInsecureRegistryHandling(URL url) throws IOException, RegistryException {
+    try {
+      if (!isHttpsProtocol(url) && !allowInsecureRegistries) {
+        throw new InsecureRegistryException(url);
+      }
+      return call(url, connectionFactory);
+
+    } catch (SSLPeerUnverifiedException exception) {
+      // Cannot verify the server (e.g., self-signed certificate or plain HTTP).
+      if (!allowInsecureRegistries) {
+        throw new InsecureRegistryException(url);
+      }
+
+      try {
+        if (insecureConnectionFactory == null) {
+          insecureConnectionFactory = Connection.getInsecureConnectionFactory();
+        }
+        return call(url, insecureConnectionFactory);
+
+      } catch (GeneralSecurityException ex) {
+        throw new RegistryException("cannot turn off TLS peer verification", ex);
+      } catch (SSLPeerUnverifiedException | HttpHostConnectException ex) {
+        // Try HTTP as a last resort.
+        GenericUrl httpUrl = new GenericUrl(url);
+        httpUrl.setScheme("http");
+        return call(httpUrl.toURL(), connectionFactory);
+      }
+    }
   }
 
   /**
@@ -136,15 +173,11 @@ class RegistryEndpointCaller<T> {
    * @throws IOException for most I/O exceptions when making the request
    * @throws RegistryException for known exceptions when interacting with the registry
    */
-  @VisibleForTesting
   @Nullable
-  T call(URL url) throws IOException, RegistryException {
-    boolean isHttpProtocol = "http".equals(url.getProtocol());
-    if (!allowInsecureRegistries && isHttpProtocol) {
-      throw new InsecureRegistryException(url);
-    }
+  private T call(URL url, Function<URL, Connection> connectionFactory)
+      throws IOException, RegistryException {
     // Only sends authorization if using HTTPS or explicitly forcing over HTTP.
-    boolean sendCredentials = !isHttpProtocol || Boolean.getBoolean("sendCredentialsOverHttp");
+    boolean sendCredentials = isHttpsProtocol(url) || Boolean.getBoolean("sendCredentialsOverHttp");
 
     try (Connection connection = connectionFactory.apply(url)) {
       Request.Builder requestBuilder =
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -125,7 +130,37 @@
    */
   @Nullable
   T call() throws IOException, RegistryException {
-    return call(initialRequestUrl);
+    return callWithInsecureRegistryHandling(initialRequestUrl);
```
coollog: Hmm, I feel like the checking should be done here for whether or not to handle insecure registries based on the flag so that it launches two codepaths - one for calling without allowing insecure registries, and one for calling with allowing insecure registries - rather than have that logic be encapsulated in `callWithInsecureRegistryHandling`. From the function call here, it seems like insecure registry handling is the default.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -125,7 +130,37 @@
...
+        if (!allowInsecureRegistries) {
+          throw new InsecureRegistryException(url);
+        }
+        if (insecureConnectionFactory == null) {
+          insecureConnectionFactory = Connection.getInsecureConnectionFactory();
+        }
+        return call(url, insecureConnectionFactory);
+
+      } catch (GeneralSecurityException ex) {
+        throw new RegistryException("cannot turn off TLS peer verification", ex);
```
coollog: nit: newline after

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
chanseokoh: > it launches two codepaths

It's not exactly like there will be two codepaths based on `allowInsecureRegistries`. They cross-over, for example, you can first contact an HTTP server (failover-to-HTTP occurred) which redirects the client to a secure HTTPS server. With `callWithInsecureRegistryHandling()`, you don't have to think about those things anymore. You can pass a URL to `callWithInsecureRegistryHandling()` recursively and it will handle things correctly.

> From the function call here, it seems like insecure registry handling is the default.

You are saying the naming `callWithInsecureRegistryHandling` is giving you the impression that it will attempt insecure communication, right? I'll think about a better name, but if you have some suggestions, let me know.

coollog: Ah yes, I'm thinking about this more in the lines of having `allowInsecureRegistries` be checked in one location to launch separate codepaths, one for secure and one for insecure. The one for insecure will be essentially the one for secure, except with additional functionality for handling HTTPS connect/TLS verification errors (codepaths with cross-over, but a different entrypoint). So, rather than checking `allowInsecureRegistries` in multiple places in `callWithInsecureRegistryHandling`, a single check would determine if the insecure handling code entrypoint would be launched vs the secure handling code entrypoint. I think I need to think more about this too.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -125,7 +130,38 @@
...
+
+      } catch (SSLPeerUnverifiedException | HttpHostConnectException ex) {
+        // Try HTTP as a last resort.
```
coollog: For example, here, I think this would allow HTTP fallback even if `allowInsecureRegistries` was false.

chanseokoh: That can never happen. If `allowInsecureRegistries` is false and the server is not a verifiable HTTPS, it will throw an exception.

```java
        // Cannot verify the server (e.g., self-signed certificate or plain HTTP).
        if (!allowInsecureRegistries) {
          throw new InsecureRegistryException(url);
        }
```

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
chanseokoh: > So, rather than checking allowInsecureRegistries in multiple places in callWithInsecureRegistryHandling, a single check would determine if the insecure handling code entrypoint would be launched vs the secure handling code entrypoint.

I think I got what you mean. That makes sense too. I'll try your idea and push up changes, and then we can evaluate which one will look better.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
coollog: Ah right, I need to read this more carefully.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
chanseokoh: Here's the attempt with your idea, if I understood it correctly: https://github.com/GoogleContainerTools/jib/pull/733/commits/6df4b42d37cc9359a1364fe6eedebc913648808b

I'm fine with either ways.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -125,7 +131,48 @@
...
+  @Nullable
+  private T possbilyInsecureCall(URL url) throws IOException, RegistryException {
+    Preconditions.checkState(allowInsecureRegistries);
+    try {
+      return call(url, connectionFactory);
+
+    } catch (SSLPeerUnverifiedException exception) {
+      try {
+        if (insecureConnectionFactory == null) {
+          insecureConnectionFactory = Connection.getInsecureConnectionFactory();
```
coollog: Hmm, it seems that `insecureConnectionFactory` doesn't need to be an instance variable.

chanseokoh: In fact, `connectionFactory` doesn't have to be so either, but it was there from the beginning, and I believe it is for mocking `Connection`. `insecureConnectionFactory` is there for the same reason.

But I guess you will wonder why the code has this null check.

```java
        Connection.getConnectionFactory(),
        null /* might never be used, so create lazily to delay throwing potential GeneralSecurityException */);
```

So this is to avoid the potential `GeneralSecurityException` when creating the `insecureConnectionFactory`. For example, if the user never sets `allowInsecureRegistries` or the server was secure even if it was set, we can proceed without an error in the situation where `GeneralSecurityException` will otherwise be thrown if blindly created at the beginning.

chanseokoh:
> Here's the attempt with your idea, if I understood it correctly: 6df4b42
> I'm fine with either ways.

> I think this looks better, but going to request for more opinions @GoogleContainerTools/java-tools

Personally, I slightly prefer the previous approach for the following reasons:
1. After 6df4b42, you add one more method (`possiblyInsecureCall`) with similar names:
   - `call()`
   - `callWithAllowInsecureRegistryHandling()`
   - `possiblyInsecureCall()`
   - `call(URL)`
2. Although `callWithAllowInsecureRegistryHandling()` is now basically for determining the two different codepaths based on `allowInsecureRegistries`, the behavior of `possiblyInsecureCall()` is identical to that of `callWithAllowInsecureRegistryHandling()` if the endpoint is secure HTTPS, whether `allowInsecureRegistries` is true or not.

However, I usually and casually follow the feedback from reviewers, and the new code seems fine to me, so I'll go for the new code. It has its own merits too.

coollog:
I think drawing out a state diagram might help with this - I'll sketch up something when I get back to somewhere with paper.

Issue #478, opened 2018-06-29
Move plugins specific code to separate module.


coollog:
Since `jib-core` will be converted to a standalone Java library for building container images, the plugins-specific code in there (mostly in the `frontend` package) should be moved to a separate module, perhaps called `jib-plugins-common`. This module would have most of the shared code for `jib-maven-plugin` and `jib-gradle-plugin`.

patflynn:
Just a thought: Most of the use-cases that make sense for jib are for
building Java images from some kind of UI (whether build system or IDE),
and conceivably a lot of that package (or as much as I could gather form
looking at the class names :) ) could apply to those clients.

coollog:
Yes, we should definitely think more about which classes we could provide in `jib-core` vs in `jib-plugins-common`. For example, the messages in the suggestions messages class (`HelpfulSuggestions`) mostly only apply to Maven or Gradle and has Maven/Gradle-specific messages and that should only go into `jib-plugins-common`, where `MainClassFinder` should definitely be in `jib-core` since it is core to Jib's simple configuration capability.

Issue #431, opened 2018-06-20
Support WARs.


coollog:
Thumbs up if you would like Jib to be able to containerize applications normally packaged as WARs.

xiangfeid:
Is there a timeline for this?

coollog:
@xiangfeid We are actively working on this issue and welcome input for solutions from anyone in the community. We expect a solution to be provided by the `0.10.0` version milestone.

@chanseokoh will be able to guide the conversation and design

YJFinBJ:
I have issue in "comand and tomcat war support #646"

chanseokoh:
For documentation after some preliminary research:

`pom.xml` of an example Maven WAR project:
```xml
  <packaging>war</packaging>
  <version>0.1.0-SNAPSHOT</version>

  <groupId>com.example</groupId>
  <artifactId>war-example</artifactId>
```

1. `$ mvn package`
1. ```
   $ cat > target/Dockerfile <<EOF
   FROM gcr.io/distroless/java/jetty:debug

   # By convention, the exploded WAR directory is <artifact-id>-<version>.
   COPY war-example-0.1.0-SNAPSHOT /jetty/webapps/ROOT

   ENTRYPOINT ["java","-jar","/jetty/start.jar"]
   CMD []
   EOF
   ```
1. `$ docker build -t my-app-on-jetty target/`
1. `$ docker run  -p 8080:8080 -t my-app-on-jetty`

The app will be available at `localhost:8080`. I expect the Tomcat base image would not be too different from this flow.

There are other images like the official Jetty Docker image and the [Google-maintained Jetty runtime image](https://github.com/GoogleCloudPlatform/jetty-runtime),  but I think we'll stick to the distroless Jetty. I hope it enabled all the typical web app support like JSP and JSTL.

briandealwis:
So we could put `WEB-INF/classes` in one layer, `WEB-INF/lib` in a second layer, and everything else in a third layer.  Does that make sense?

coollog:
Yep! Should be fairly simple to write an adapter for this war format once #516 is merged.

loosebazooka:
@tofflos, are you wondering about provided deps? We do not include provided deps so if MicroProfile is providing them for you that's fine.

In the case of the app server, you would still pull `payara/micro`. The one issue is determining `$DEPLOY_DIR`, we will have a provision for custom application locations, our default case will probably support distroless/jetty though.

joelhandwell:
[...]

PR #735, opened 2018-07-28
Skips execution if packaging is 'pom'.

### LINKED ISSUE #734: Skip Mojo execution if packaging is POM.
This will help in multi-module Maven projects to skip containerizing a parent POM and only containerize modules that are actually applications to containerize.

coollog:
Fixes #734 

The code duplication will be taken care of by when we add a plugins-common framework (#478).

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -53,6 +53,11 @@ public class BuildDockerMojo extends JibPluginConfiguration {
 
   @Override
   public void execute() throws MojoExecutionException {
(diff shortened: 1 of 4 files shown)


loosebazooka:
I mean this is an easy way to fix the problem, but maven is configurable to avoid this.

loosebazooka:
Should we only be executing on "jar" builds, and ignoring all other builds?

coollog:
@loosebazooka We could, though we would want to enable "war" too when we take care of #431. However, we might want to allow for custom packaging types, though I'm not if people actually do that.

loosebazooka:
Yeah, I'm not sure if I know what the right way is. Perhaps just leave it as is for now, maybe using the "jib" packaging type option can clear this up for most people.

coollog:
Oops, forgot to add CHANGELOG. Will do in followup PR.

Issue #540, opened 2018-07-11
Improve examples.


coollog:
- Improve the documentation for [helloworld](https://github.com/GoogleContainerTools/jib/tree/master/examples/helloworld)
- Add examples for other registries besides GCR (like **Docker Hub**, ECR, ACR, Nexus)
- Provide examples for multi-module projects (DONE - #736)
  - Multi-module example with one module as common library (low priority)
- Move the [Micronaut example](https://github.com/coollog/micronaut-jib) here (DONE - #893)

Please comment with any examples you would like to us to provide.

giskarda:
Fairly interested in the multi-module example!

wilder:
@coollog, do you think adding an example using Gradle and [Docker Hub](https://hub.docker.com/) as registry would be a good idea? I'm interested in setting up a project using this configuration.

coollog:
@wilder thanks for the suggestion - added to the list of registries!

coollog:
Hi @giskarda , we have added a simple multi-module example at https://github.com/GoogleContainerTools/jib/tree/master/examples/multi-module :)

coollog:
Will close this and file separate issues for example requests as needed.

PR #736, opened 2018-07-28
 Adds multi-module example.

coollog:
For #540 

Preview at https://github.com/GoogleContainerTools/jib/tree/multimodule-example/examples

A lot of the code is just the wrappers, so when reviewing just skip over those.


examples/multi-module/hello-service/src/main/java/hello/Application.java
```diff
@@ -0,0 +1,12 @@
+package hello;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+
+@SpringBootApplication
+public class Application {
+
+  public static void main(String[] args) {
```
chanseokoh: Indentation is off and not consistent. (Other files do 2-space indentation.)

examples/multi-module/README.md
```diff
@@ -0,0 +1,72 @@
...
+# Where are the containers
+
+The output of the build should have the container image references highlighted in <span style="color: cyan">cyan</span>. You can expect them to be at:
+
+- `name-service`: `gcr.io/PROJECT_ID/name-service:0.1.0`
+- `hello-service`: `gcr.io/PROJECT_ID/hello-service:0.1.0`
+
+# How to run on Kubernetes
+
+[`kubernetes.yaml`](kubernetes.yaml) defines the manifests for running the two microservices on Kubernetes. Make sure to change `PROJECT_ID` to your own Google Cloud Platform project.
```
chanseokoh: I thought `PROJECT_ID` here referred to the env variable mentioned just above, until I looked into `kubernetes.yaml`. I think "Make sure to open the file and change ..." will clear any possible confusion.

examples/multi-module/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+
+    <build>
+        <!-- Defines plugins that are used in the modules. -->
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>com.google.cloud.tools</groupId>
+                    <artifactId>jib-maven-plugin</artifactId>
+                    <!-- TODO: Change to released version. -->
+                    <version>0.9.8-SNAPSHOT</version>
```
chanseokoh: Maybe just better to use a non-SNAPSHOT version, as otherwise users need to install it into their local repo. Also, I see you used `0.9.7` (non-SNAPSHOT) in `build.gradle`.

examples/multi-module/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+        <module>name-service</module>
+        <module>hello-service</module>
+    </modules>
+
+    <!-- Uses Java 8. -->
+    <properties>
+        <java.version>1.8</java.version>
+
+        <!-- IMPORTANT: Set the environment variable PROJECT_ID to your own Google Cloud Platform project. -->
+        <image>gcr.io/${env.PROJECT_ID}/${artifactId}:${version}</image>
```
chanseokoh:
```
[WARNING] The expression ${artifactId} is deprecated. Please use ${project.artifactId} instead.
[WARNING] The expression ${version} is deprecated. Please use ${project.version} instead.
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
```

loosebazooka:
It seems to me that some people are also having issues with this if they include a maven library from the same project that they don't want containerized?

for example
```
.
├── pom.xml (parent/root)
├── container1
│   └── pom.xml
├── container2
│   └── pom.xml
└── non-container-shared-lib
    └── pom.xml
```
We should cover this case as well either in this PR or a followup.

coollog:
@loosebazooka Would that non-containerized library be just a `pom` packaging?

examples/multi-module/pom.xml
coollog: Yep, this should be changed after releasing 0.9.8 since it depends on #735 to work.

loosebazooka:
@coollog, it would be a jar of some kind, a common library pulled in to both container1 and container2. The more I look at maven multi module stuff though, the less I like how it is done, since it appears you still need to specify dependency (lib) in all other modules that use it, just as you would any other dependency. It kinda feels like it's just doing an `mvn install` for you but making every else WAY harder. Maybe, just file the library addition to the example low priority and we can deal with it later.

coollog:
Added the common library example to #540 as low-priority. This should be good to finalize/merge now.

examples/multi-module/.gitignore
```diff
@@ -0,0 +1,8 @@
+.DS_Store
```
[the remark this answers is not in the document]
coollog: Yeah, this is for other users, but I think we can just remove this since it assumes people would use git.

loosebazooka:
Furthering the non-pom non-jib module point:

```
.
├── pom.xml (parent/root)
├── container1
│   └── pom.xml
├── container2
│   └── pom.xml
└── non-container-shared-lib
    └── pom.xml
```

In this case, we don't want to deploy non-container-shared-jib. Now I'm not sure what exactly happens here when we run `mvn compile jib:build` and `jib-maven-plugin` isn't applied on non-container-shared-lib, but I think the user has to do something like

```
$ mvn compile && mvn jib:build -pl container1,container2
```

to avoid the maven reactor plugin from throwing an error about jib not being found on non-container-shared-lib.

Although if the user is truly sharing this library in container1 or container2, then non-container-shared-lib needs to be installed, so it might even look like

```
$ mvn compile && mvn install -pl non-container-shared-lib && mvn jib:build -pl container1,container2
```

but there's also all this -am (also make), I'm not sure what lifecycle goals that runs.

I think I'm rambling a little here, lets just go with the simple example you have and we can update it as we get questions.

PR #737, opened 2018-07-30
 Factors MainClassFinder out from MainClassResolver.

coollog:
This works towards #478 by decoupling `MainClassFinder` from `ProjectProperties`.


jib-core/src/test/java/com/google/cloud/tools/jib/frontend/MainClassFinderTest.java
```diff
@@ -1,211 +1,122 @@
-/*
```
TadCordle: Accidental deletion?

coollog: Oh yea oops.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -38,101 +36,159 @@
...
-    return mainClass;
+      if (mainClasses.size() == 1) {
+        // Valid class found; use inferred main class
+        return Result.success(mainClasses.get(0));
+      }
+      if (mainClasses.size() == 0) {
+        // No main class found anywhere
+        return Result.mainClassNotFound();
+      }
+      // More than one main class found with no jar plugin to fall back on; error
```
chanseokoh: I think the comment "with no jar plugin to fall back on" does not apply here.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * If {@code mainClass} is {@code null}, tries to infer main class in this order:
+   *
+   * <ul>
+   *   <li>1. Looks in a {@code jar} plugin provided by {@code projectProperties} ({@code
+   *       maven-jar-plugin} for maven or {@code jar} task for gradle).
+   *   <li>2. Searches for a class defined with a main method.
+   * </ul>
+   *
+   * <p>Warns if main class is not valid, or throws an error if no valid main class is not found.
```
chanseokoh: "if no valid main class is not found" --> "if no valid main class is found"

That said, the difference between "main class is not valid" and "no valid main class is found" isn't clear to me just reading the Javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -38,101 +36,159 @@
...
-   *   <li>1. Looks in a {@code jar} plugin provided by {@code projectProperties} ({@code
-   *       maven-jar-plugin} for maven or {@code jar} task for gradle).
-   *   <li>2. Searches for a class defined with a main method.
-   * </ul>
-   *
-   * <p>Warns if main class is not valid, or throws an error if no valid main class is not found.
+   * @param classesFiles the classes files to check
+   * @param buildLogger used for displaying status messages.
+   */
+  public MainClassFinder(ImmutableList<Path> classesFiles, BuildLogger buildLogger) {
```
chanseokoh: `BuildLogger` doesn't seem necessary.

chanseokoh: Oh, never mind. Didn't see the other part of the class.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
```diff
@@ -0,0 +1,108 @@
...
+        MainClassFinder.Result mainClassFinderResult =
+            new MainClassFinder(
+                    projectProperties.getClassesLayerEntry().getSourceFiles(),
+                    projectProperties.getLogger())
+                .find();
+
+        if (mainClassFinderResult.isSuccess()) {
+          mainClass = mainClassFinderResult.getFoundMainClass();
+
+        } else if (mainClass == null) {
```
chanseokoh: These multiple branches at various levels are a bit hard to understand... we call `MainClassFinder.find()` even if `mainClass` is not null, if `!BuildConfiguration.isValidJavaClass(mainClass)` (Line 54). Why do we check the error of `MainClassFinder.find()` only when `mainClass` is null?

coollog: Agreed, I'll reorganize this a bit.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
coollog: Reworded to include "provided by projectProperties"

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
```diff
@@ -0,0 +1,136 @@
...
+    // If mainClass is null, try to find via projectProperties.
+    if (mainClass == null) {
+      mainClass = getMainClassFromJar(projectProperties);
+    }
+
+    // If mainClass is still null, try to search in class files.
+    if (mainClass == null) {
+      mainClass = findMainClassInClassFiles(projectProperties, true);
+
+    } else if (!BuildConfiguration.isValidJavaClass(mainClass)) {
```
chanseokoh: Looks like the static `isValidJavaClass` method is called only from this class and no one else. Move the method here?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
```diff
@@ -0,0 +1,136 @@
...
+        .info(
+            "Searching for main class... Add a 'mainClass' configuration to '"
+                + projectProperties.getPluginName()
+                + "' to improve build speed.");
+    return projectProperties.getMainClassFromJar();
+  }
+
+  @Nullable
+  private static String findMainClassInClassFiles(
+      ProjectProperties projectProperties, boolean shouldError) throws MainClassInferenceException {
```
chanseokoh: Making the method throw or suppress exceptions based on an argument seems weird to me. I'm thinking it's better to remove `shouldError` and have caller to catch the exception if they really want to ignore it.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
coollog: Oh it's used on line 66 too.

chanseokoh: Yeah, so only in this class. Why not moving it from `BuildConfiguration` to `MainClassResolver` (this class)?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
```diff
@@ -0,0 +1,137 @@
...
+    if (mainClass == null) {
+      mainClass = findMainClassInClassFiles(projectProperties);
+
+    } else if (!BuildConfiguration.isValidJavaClass(mainClass)) {
+      // If mainClass found in projectProperties is not valid, try to search in class files, but
+      // don't error if not found in class files.
+      try {
+        String mainClassInClassFiles = findMainClassInClassFiles(projectProperties);
+        if (mainClassInClassFiles != null) {
+          mainClass = mainClassInClassFiles;
```
chanseokoh: I think you can just do `mainClass = findMainClassInClassFiles(...)`, because it will be assigned only when it succeeds (when no exception), right?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
```diff
@@ -0,0 +1,137 @@
...
+    projectProperties
+        .getLogger()
+        .info(
+            "Searching for main class... Add a 'mainClass' configuration to '"
+                + projectProperties.getPluginName()
+                + "' to improve build speed.");
+    return projectProperties.getMainClassFromJar();
+  }
+
+  @Nullable
```
chanseokoh: Probably you can remove this, since `getFoundMainClass()` is not `@Nullable`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassResolver.java
coollog: Oh oops, I misread your comment.

PR #739, opened 2018-07-30
Makes sure there are no duplicate cache entries.

### LINKED ISSUE #721: jib cache metadata-v2.json seems to grow without bound
```
$ cd ~/Library/Application\ Support/google-cloud-tools-java/jib

$ ls -1 *.tar.gz | wc -l
      15

$ jq '.layers[].reference.digest' metadata-v2.json | wc -l
313

$ jq '.layers[].reference.digest' metadata-v2.json | sort -u | wc -l
13
```

coollog:
Fixes #721 

The solution here is to, when adding a new layer, remove any prior layers with the same digest (essentially bumping layers to the top if re-added). This should work, for example, since applying layers A and B in order ABA should produce an identical filesystem as just BA.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
@@ -32,9 +32,9 @@ public class ImageLayers<T extends Layer> implements Iterable<T> {
         ImmutableSet.builder();
 
     /**
-     * Adds a layer. Removes any prior occurrences of the same layer. Note that not all {@link
-     * Layer} types implement {@code equals/hashCode} - only those that are intended to be not
-     * duplicated will.
+     * Adds a layer. Removes any prior occurrences of the same layer.
+     *
+     * Note that only subclasses of {@link Layer} that implement {@code equals/hashCode} will be guaranteed to not be duplicated.
      *
      * @param layer the layer to add
      * @return this
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
@@ -34,7 +34,8 @@ public class ImageLayers<T extends Layer> implements Iterable<T> {
     /**
      * Adds a layer. Removes any prior occurrences of the same layer.
      *
-     * Note that only subclasses of {@link Layer} that implement {@code equals/hashCode} will be guaranteed to not be duplicated.
+     * <p>Note that only subclasses of {@link Layer} that implement {@code equals/hashCode} will be
+     * guaranteed to not be duplicated.
      *
      * @param layer the layer to add
      * @return this
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
@@ -19,6 +19,8 @@ package com.google.cloud.tools.jib.image;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 /** Holds the layers for an image. Makes sure that each layer is only added once. */
@@ -26,7 +28,7 @@ public class ImageLayers<T extends Layer> implements Iterable<T> {
 
   public static class Builder<T extends Layer> {
 
-    private final ImmutableList.Builder<T> layersBuilder = ImmutableList.builder();
+    private final Set<T> layersBuilder = new LinkedHashSet<>();
     private final ImmutableSet.Builder<DescriptorDigest> layerDigestsBuilder =
         ImmutableSet.builder();
 
@@ -34,19 +36,19 @@ public class ImageLayers<T extends Layer> implements Iterable<T> {
     @Nullable private T lastLayer;
 
     /**
-     * Adds a layer.
+     * Adds a layer. Removes any prior occurrences of the same layer. Note that not all {@link
+     * Layer} types implement {@code equals/hashCode} - only those that are intended to be not
+     * duplicated will.
      *
      * @param layer the layer to add
      * @return this
      * @throws LayerPropertyNotFoundException if adding the layer fails
      */
     public Builder<T> add(T layer) throws LayerPropertyNotFoundException {
-      // Doesn't add the layer if the last layer is the same.
-      if (!isSameAsLastLayer(layer)) {
-        layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
-        layersBuilder.add(layer);
-        lastLayer = layer;
-      }
+      layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
+      layersBuilder.remove(layer);
+      layersBuilder.add(layer);
+      lastLayer = layer;
 
       return this;
     }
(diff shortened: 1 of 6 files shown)


loosebazooka:
@briandealwis, as far as I can tell in how we layer in jib, `ABA == BA` and I can't really think of a case where `ABA != BA`?

coollog:
Each layer (ie. A, B) would have a set of changes for files (ie. x, y, z). Each of these changes can be either a create, update, or delete, which translates to a change operation to set the contents of a file. So each layer can be represented by a set of pairs `(file, newContents)` (eg. `{(x, content1), (y, content2), (z, content3)}`). These changes, applied in sequence, have independent histories for each file. Therefore, for each file in its layer history chain, each change essentially replaces any prior change (only the last change takes effect). Therefore, applying layers ABA should have the same effect as applying layers BA, as all changes in the second A replace the changes in the first A (since they have the same change sets).

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
```diff
@@ -19,34 +19,36 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Set;
 import javax.annotation.Nullable;
 
 /** Holds the layers for an image. Makes sure that each layer is only added once. */
```
chanseokoh: We should remove "Makes sure that each layer is only added once" or update the Javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
```diff
@@ -19,34 +19,36 @@
...
-      // Doesn't add the layer if the last layer is the same.
-      if (!isSameAsLastLayer(layer)) {
-        layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
-        layersBuilder.add(layer);
-        lastLayer = layer;
-      }
+      layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
+      layersBuilder.remove(layer);
+      layersBuilder.add(layer);
+      lastLayer = layer;
```
chanseokoh: `lastLayer` seems unused.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
```diff
@@ -19,34 +19,36 @@
...
+import java.util.Set;
 import javax.annotation.Nullable;
 
 /** Holds the layers for an image. Makes sure that each layer is only added once. */
 public class ImageLayers<T extends Layer> implements Iterable<T> {
 
   public static class Builder<T extends Layer> {
 
-    private final ImmutableList.Builder<T> layersBuilder = ImmutableList.builder();
+    private final Set<T> layersBuilder = new LinkedHashSet<>();
```
chanseokoh: - Isn't the order of layers important? (`Set` does not provide any ordering guarantees.)
- Are you comfortable calling this a builder?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
```diff
@@ -19,34 +19,36 @@
...
     public Builder<T> add(T layer) throws LayerPropertyNotFoundException {
-      // Doesn't add the layer if the last layer is the same.
-      if (!isSameAsLastLayer(layer)) {
-        layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
-        layersBuilder.add(layer);
-        lastLayer = layer;
-      }
+      layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
+      layersBuilder.remove(layer);
+      layersBuilder.add(layer);
```
chanseokoh: I don't think `Set` allows duplicates, so probably there is no need to remove before adding.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
coollog: I'm using `LinkedHashSet` to ensure the order is the order of insertion - I'll change the signature to `LinkedHashSet<T> layers` to make this clearer.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
```diff
@@ -19,34 +19,31 @@
...
 import javax.annotation.Nullable;
 
-/** Holds the layers for an image. Makes sure that each layer is only added once. */
+/** Holds the layers for an image. Makes sure that there are no duplicate layers. */
 public class ImageLayers<T extends Layer> implements Iterable<T> {
 
   public static class Builder<T extends Layer> {
 
-    private final ImmutableList.Builder<T> layersBuilder = ImmutableList.builder();
+    private final LinkedHashSet<T> layers = new LinkedHashSet<>();
```
briandealwis: I think explicitly marking this a `LinkedHashSet` is a good idea, but it warrants a bit of explanation that the set aspect is for the fast comparison.  TBH given the need to remove-and-then-add, it would be better to just make this a List. Docker is limited to 40 layers (I believe), so the O(n) vs O(ln n) doesn't matter.

coollog: I think the speed here might actually matter a bit, since this class is used to represent the layers in the cache as well (as this is mostly fixing #721). Having the remove-then-add would be `O(n)` performed `O(n)` times (for `n` layers). Since the cache could potentially have many layers after extensive use (not limited to the Docker image layer limit), the `O(n2)` vs `O(n ln n)` could potentially be a noticeable difference when manipulating the cache metadata.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
```diff
@@ -19,34 +19,32 @@
...
      */
     public Builder<T> add(T layer) throws LayerPropertyNotFoundException {
-      // Doesn't add the layer if the last layer is the same.
-      if (!isSameAsLastLayer(layer)) {
-        layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
-        layersBuilder.add(layer);
-        lastLayer = layer;
-      }
+      layerDigestsBuilder.add(layer.getBlobDescriptor().getDigest());
+      layers.remove(layer);
```
TadCordle: Don't `LinkedHashSet`s already handle duplicates?

chanseokoh:
> `ABA == BA` and I can't really think of a case where `ABA != BA`?

I wonder if this is possible: for example, both A and B create a read-only file (no permission to write or modify it) in the same location, which once placed no one can overwrite it. That is, whoever writes the file first wins. If that's possible, `ABA != BA`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageLayers.java
chanseokoh: Turns out this is to re-order the element (https://github.com/GoogleContainerTools/jib/pull/739#discussion_r206358351).

That said, it seems worth to put a comment like "remove necessary to rearrange the insertion-order of `LinkedHashSet`".

coollog:
@chanseokoh The file permissions for the files will not affect the overwrite. In the layered container filesystem at runtime, the order of the search for a file goes from the last layer backwards, so if prior layers with a file will not be reached if later layers have the same file.

coollog:
@briandealwis Yep, the intention for this fix is to also have the same optimization applied to generated images as well.

chanseokoh:
Okay, this PR looks good to me.

coollog:
Looks like this helped to fix the failing integration test for the simple project, but the empty project still fails (because the second build time is longer).

PR #747, opened 2018-07-31
Minor fixes

chanseokoh:
- `pathFilter` is not `@Nullable`.
- We are trying all possible registries, including the original one.


jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/DirectoryWalker.java
```diff
@@ -90,10 +90,7 @@ public DirectoryWalker filterRoot() {
    */
   public ImmutableList<Path> walk() throws IOException {
     try (Stream<Path> fileStream = Files.walk(rootDir)) {
-      Stream<Path> filteredFileStream = fileStream;
-      if (pathFilter != null) {
-        filteredFileStream = fileStream.filter(pathFilter);
-      }
+      Stream<Path> filteredFileStream = fileStream.filter(pathFilter);
```
coollog: Could probably even combine this and the next line.

Issue #714, opened 2018-07-25
Add event system for propagating asynchronous events to user of jib-core.


coollog:
- [x] Proposal for eventing system #909 
- [x] Add eventing system #921, #931 
- [x] Define events for each log point
  - [x] Set up and `RetrieveRegistryCredentialsStep` #971
  - [x] Replace all uses of `JibLogger` with emitting `LogEvent`/`TimerEvent` and remove `JibLogger`. #1034, #1043, #1048
- [ ] Rename `EventEmitter` to `EventDispatcher` #1053

Followup:
- Create new events that contain structured information for users of Jib Core for various parts of the build execution

coollog:
@GoogleContainerTools/java-tools It would be good to have someone else take a stab at the rest of the tasks so they can try out using the eventing system API.

TadCordle:
I can try working on this.

PR #748, opened 2018-07-31
Inform user of fallback to HTTP

### LINKED ISSUE #573: Inform user of fallback to HTTP.
Regards https://github.com/GoogleContainerTools/jib/issues/545#issuecomment-404378398

TadCordle:
Fixes #573


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/AuthenticationMethodRetrieverIntegrationTest.java
```diff
@@ -24,17 +25,20 @@
 /** Integration tests for {@link AuthenticationMethodRetriever}. */
 public class AuthenticationMethodRetrieverIntegrationTest {
 
+  private static final EmptyBuildLogger BUILD_LOGGER = new EmptyBuildLogger();
```
briandealwis: Can't this just be a mock?

TadCordle: Wasn't sure if it was a good idea to use mocks in integration tests.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/BlobCheckerIntegrationTest.java
```diff
@@ -28,11 +29,12 @@
 public class BlobCheckerIntegrationTest {
 
   @ClassRule public static LocalRegistry localRegistry = new LocalRegistry(5000);
+  private final EmptyBuildLogger buildLogger = new EmptyBuildLogger();
```
chanseokoh: Let's be consistent. `static final ... BUILD_LOGGER`?

coollog:
Oh for the `BuildLogger`, we should probably not have the `http` package classes depend on `builder` package classes (since the `http` package is lower-level and should be decoupled from the Jib execution - `builder` should be the highest level package). We could either have the `http` package have its own logger, or have some way of reporting back log messages that doesn't require injecting a `builder.BuildLogger`, or move the `BuildLogger` to a lower-level package.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/BlobCheckerIntegrationTest.java
briandealwis: Or even just a singleton: `EmptyBuilderLogger.INSTANCE`.

TadCordle:
@coollog I think `BuildLogger` fits in the root package with `Timer`, WDYT?

coollog:
@TadCordle We can probably have it there for now, though the `Timer` is mostly a debugging thing that should probably be replaced with a more robust solution. With #714, we'll probably be able to rip out the timer logic since eventing will allow clients to handle their own timing.

briandealwis:
Should we rename `BuildLogger` to something like `JibLogger` or `JibReporting`?

TadCordle:
> Should we rename BuildLogger to something like JibLogger or JibReporting?

I think I'll submit a followup for this, it changes a lot of files.

PR #750, opened 2018-07-31
Enable Labels in Container Config

loosebazooka:
- only exposes in core the opportunity to add labels
- this PR does NOT
  - pass through labels from a base image (should be in followup)
  - enable frontend labels access (should be in followup)

part of #751


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -98,6 +99,23 @@ public Builder setExposedPorts(@Nullable List<Port> exposedPorts) {
       return this;
     }
 
+    /**
+     * Sets the container's exposed ports.
```
TadCordle: Leftover copy-pasted comment?

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -110,15 +111,26 @@
...
         this.exposedPorts = null;
       } else {
-        this.exposedPorts = exposedPorts;
+        this.exposedPorts = ImmutableList.copyOf(exposedPorts);
       }
       return this;
     }
 
+    /**
+     * Sets the items in the "ExposedPorts" field in the container configuration.
```
TadCordle: Same here.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
loosebazooka: copy-pasta strikes again, I'll fix these.

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildConfigurationTest.java
```diff
@@ -250,6 +253,17 @@ public void testBuilder_nullValues() {
       Assert.assertNull(ex.getMessage());
     }
 
+    // Labels element should not be null.
+    try {
+      Map<String, String> badLabels = new HashMap<>();
+      badLabels.put("label-key", null);
+      BuildConfiguration.builder(Mockito.mock(BuildLogger.class))
+          .setContainerConfiguration(ContainerConfiguration.builder().setLabels(badLabels).build());
```
chanseokoh: The try block should only be focused on the method being tested. I actually think `BuildConfiguration.builder()` doesn't play a role. Maybe

```java
Map<String, String> badLabels = ...
badLabels.put(...)
Builder builder = ContainerConfiguration.builder();
try {
  builder.setLabels(badLabels);
  Assert.fail(...);
} catch (...) {
```

loosebazooka: Yeah, makes sense.

loosebazooka: hrmm... looks like we might need to cleanup this whole test class :\

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -110,15 +111,26 @@
      * @param exposedPorts the list of exposed ports to add
      * @return this
      */
-    public Builder<T> setExposedPorts(@Nullable ImmutableList<Port> exposedPorts) {
+    public Builder<T> setExposedPorts(@Nullable List<Port> exposedPorts) {
       if (exposedPorts == null) {
```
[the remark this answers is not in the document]
loosebazooka: Yeah, I was going to, but figured it should be in a followup. This ImmutableList -> List change probably shouldn't be part of this PR either. I might go through and do some cleanup after this.

PR #753, opened 2018-07-31
Has integration test check actual build time.

coollog:
https://github.com/GoogleContainerTools/jib/pull/749 did not seem to fix the issue completely (simple project still fails, although running the macOS integration tests locally does not fail). This PR runs clean before the first integration test run so that the first run always starts with a clean cache.

This PR also changes the time comparison to be the actual execution time of the `jib:build` goal rather than the time the `Verifier` took so that the comparison is more accurate - this is the main fix.


coollog:
Okay it works with checking time from the actual build logs for `jib:build`.

coollog:
It finally works! @GoogleContainerTools/java-tools

coollog:
@TadCordle invoking a few times

coollog:
Looks like it failed again but with a much smaller difference (~20ms). Will merge this for now since this is probably a correct way to do it, but we'll still need to determine how to alleviate this difference in a future PR.

PR #762, opened 2018-08-01
Maven release v0.9.8

chanseokoh:


loosebazooka:
This should probably go in before the labels stuff goes in.

PR #766, opened 2018-08-01
Experimental badge

loosebazooka:
Use a more aesthetically pleasing experimental badge


jib-gradle-plugin/README.md
```diff
@@ -1,4 +1,4 @@
-[![experimental](http://badges.github.io/stability-badges/dist/experimental.svg)](http://github.com/badges/stability-badges)
+![experimental](https://img.shields.io/badge/stability-experimental-darkorange.svg)
 [![Analytics](https://cloud-tools-for-java-metrics.appspot.com/UA-121724379-2/jib-gradle-plugin)](https://github.com/igrigorik/ga-beacon)
```
TadCordle: Maybe we should move these analytics badges to the bottom of the page too (to match the main readme, and also because the analytics badge looks slightly different and it looks weird).

loosebazooka: :+1:

patflynn:
@loosebazooka is there a beta badge?  Seems like we're further along the road than experimental since we're tracking a 1.0 release

loosebazooka:
@patflynn it's whatever we want to, I can change it to beta

patflynn:
up to you all. Just seems a bit less scary.

Issue #632, opened 2018-07-17
NullPointerException  when jib-maven-plugin is configured with environment variables


kemitix:
**Description of the issue**:

`NullPointerException` when `jib-maven-plugin` is configured with environment variables.

**Expected behavior**:

Normal image to be built.

**Steps to reproduce**:

`mvn jib:dockerBuild`

**Environment**:

Apache Maven 3.5.4 (1edded0938998edf8bf061f1ceb3cfdeccf443fe; 2018-06-17T19:33:14+01:00)
Maven home: C:\Apps\apache-maven-3.5.4
Java version: 1.8.0_172, vendor: Oracle Corporation, runtime: C:\Program Files\Java\jdk1.8.0_172\jre
Default locale: en_GB, platform encoding: Cp1252
OS name: "windows 10", version: "10.0", arch: "amd64", family: "windows"

**`jib-maven-plugin` Configuration**:

```xml
<plugin>
  <groupId>com.google.cloud.tools</groupId>
  <artifactId>jib-maven-plugin</artifactId>
  <version>0.9.7-SNAPSHOT</version>
  <configuration>
    <environment>
      <SPRING_CONFIG_LOCATION>file:/path/application.properties</SPRING_CONFIG_LOCATION>
    </environment>
  </configuration>
  ...
</plugin>
```

**Log output**:

```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.7-SNAPSHOT:dockerBuild (default-cli) on project mig-management-portal: Execution default-cli of goal com.google.cloud.tools:jib-maven-plugin:0.9.7-SNAPSHOT:dockerBuild failed. NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.7-SNAPSHOT:dockerBuild (default-cli) on project mig-management-portal: Execution default-cli of goal com.google.cloud.tools:jib-maven-plugin:0.9.7-SNAPSHOT:dockerBuild failed.
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:213)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
[...]

PR #768, opened 2018-08-01
Cleanup null validation when configuration image

loosebazooka:
This is just a followup to #750 to cleanup and organize some code


jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildConfigurationTest.java
```diff
@@ -215,90 +212,4 @@ public void testBuilder_missingValues() {
           ex.getMessage());
     }
   }
-
-  @Test
-  @SuppressWarnings("JdkObsolete")
-  public void testBuilder_nullValues() {
```
loosebazooka: These tests were moved to ContainerConfigurationTest

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildConfigurationTest.java
```diff
@@ -215,90 +212,4 @@ public void testBuilder_missingValues() {
...
-      Assert.assertNull(ex.getMessage());
-    }
-
-    // Can accept empty environment.
-    BuildConfiguration.builder(Mockito.mock(JibLogger.class))
-        .setContainerConfiguration(
-            ContainerConfiguration.builder().setEnvironment(ImmutableMap.of()).build());
-
-    // Environment map can accept TreeMap and Hashtable.
-    BuildConfiguration.builder(Mockito.mock(JibLogger.class))
```
loosebazooka: These were removed, doesn't make sense to ensure `TreeMap`/`HashMap` are valid `Map` types.

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/ContainerConfigurationTest.java
```diff
@@ -0,0 +1,76 @@
...
+      ContainerConfiguration.builder().setEntrypoint(Arrays.asList("first", null));
+      Assert.fail("The IllegalArgumentException should be thrown.");
+    } catch (IllegalArgumentException ex) {
+      Assert.assertNull(ex.getMessage());
+    }
+
+    // Exposed ports element should not be null.
+    try {
+      ContainerConfiguration.builder()
+          .setExposedPorts(Arrays.asList(new Port(1000, Protocol.TCP), null));
```
chanseokoh: This is still not narrow enough. For example, suppose Port was modified to throw IllegalArgumentException when given 1000. The test will still pass if we introduce a bug into .setExposedPorts() that won't throw the exception.

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/ContainerConfigurationTest.java
```diff
@@ -0,0 +1,76 @@
+package com.google.cloud.tools.jib.configuration;
+
+import static org.junit.Assert.*;
```
coollog: looks like there's a static wildcard import here.

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/ContainerConfigurationTest.java
```diff
@@ -0,0 +1,76 @@
+package com.google.cloud.tools.jib.configuration;
```
coollog: copyright

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/ContainerConfigurationTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.cloud.tools.jib.configuration.Port.Protocol;
+import com.google.common.collect.ImmutableMap;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ContainerConfigurationTest {
```
coollog: short javadoc here

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/ContainerConfigurationTest.java
loosebazooka: Ah right. Will update

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/ContainerConfigurationTest.java
loosebazooka: done

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildConfigurationTest.java
chanseokoh: Now I remember: https://github.com/GoogleContainerTools/jib/issues/632#issue-341871451

The point of the test was to ensure that we don't call things like `containsKey(null)`, because while `ImmutableMap` is OK with that, some `Map` implementations like `TreeMap` throws an exception right away. Maybe adding the issue link as a comment make this clearer.

loosebazooka: that's a good point, I'll put them back in.

chanseokoh:
I pushed the "Delete Branch" button. You can always restore it later.

PR #770, opened 2018-08-01
Add FAQ entries for diagnosing connectivity issues

### LINKED ISSUE #615: Add docs for using Jib behind a proxy.
Solution by @chanseokoh : https://github.com/GoogleContainerTools/jib/issues/525#issuecomment-404176907

briandealwis:
Fixes #615


docs/faq.md
```diff
@@ -200,9 +202,46 @@ kubectl patch serviceaccount default \
...
+   - **The registry does not support the [Docker Registry V2 Schema 2](https://github.com/GoogleContainerTools/jib/issues/601)** (sometimes referred to as _v2-2_).  This problem is usually shown by failures wth `INVALID_MANIFEST` errors. Some registries can be configured to support V2-2 such as [Artifactory](https://www.jfrog.com/confluence/display/RTF/Docker+Registry#DockerRegistry-LocalDockerRepositories). Other registries, such as Quay.io/Quay Enterprise and OpenShift, are in the process of adding support.
+
+### How can I examine network traffic?
+
+It can be useful to examine network traffic to diagnose connectivity issues. Jib uses the Google HTTP client library to interact with registries which logs HTTP requests using the JVM-provided `java.util.logging` facilities.  It is very helpful to serialize Jib's actions using the `jibSerialize` property.
+
+To see the HTTP traffic, create a `logging.properties` file with the following:
```
loosebazooka: It shouldn't matter, but should we tell people where to create this file (project root) for it to work as we describe?

briandealwis: I'll make it more clear in the example that they specify the path to the file

docs/faq.md
```diff
@@ -200,9 +202,46 @@ kubectl patch serviceaccount default \
...
+And then launch your build tool as follows:
+```sh
+mvn -Djava.util.logging.config.file=path/to/log.properties -DjibSerialize=true ...
+```
+or
+```sh
+gradle -Djava.util.logging.config.file=path/to/log.properties -DjibSerialize=true ...
+```
+
 ### How do I enable debugging?
```
coollog: Oh this was actually meant to be answered with how to debug the running container via attaching, for example, an IDE debugger. The answer here could be along the lines of "How do I view debug logs for Jib?", though I think most people would know how to do this for Maven/Gradle.

docs/faq.md
```diff
@@ -200,9 +202,46 @@ kubectl patch serviceaccount default \
 
 See more at [Using Google Container Registry (GCR) with Minikube](https://ryaneschinger.com/blog/using-google-container-registry-gcr-with-minikube/).
 
+### How do I configure a proxy?
+
+Jib currently requires configuring your build tool to use the appropriate [Java networking properties](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) (`https.proxyHost`, `https.proxyPort`, `https.proxyUser`, `https.proxyPassword`).
+
+### How can I diagnose problems pulling or pushing from remote registries?
```
coollog: Looks like this is missing an entry in the links up top.

docs/faq.md
```diff
@@ -200,9 +202,46 @@ kubectl patch serviceaccount default \
...
+
+There are a few reasons why Jib may be unable to connect to a remote registry, including:
+
+   - **Access requires a proxy.** See [_How do I configure a proxy?_](#how-do-i-configure-a-proxy) for details.
```
coollog: nit: the tabulation before `-` can be removed since the markdown will render it regardless of whitespacing

docs/faq.md
```diff
@@ -200,9 +202,46 @@ kubectl patch serviceaccount default \
...
+   - **The registry does not support HTTPS or its certificates are not configured properly.**  We have a separate document on [handling registries that use self-signed certificates](self_sign_cert.md).  We support an  `allowInsecureRegistries` flag to ignore SSL certificate validation, but we do recommend its use (_Jib 0.9.8_). Finally we do not pass authentication details on non-HTTPS connections, but this can be overridden with the `sendCredentialsOverHttp` system property (_Jib 0.9.8_).
```
coollog: nit: `s/_Jib 0.9.8_/_version 0.9.8_/`?

docs/faq.md
```diff
@@ -200,9 +202,46 @@ kubectl patch serviceaccount default \
...
+   - **The registry does not support the [Docker Registry V2 Schema 2](https://github.com/GoogleContainerTools/jib/issues/601)** (sometimes referred to as _v2-2_).  This problem is usually shown by failures wth `INVALID_MANIFEST` errors. Some registries can be configured to support V2-2 such as [Artifactory](https://www.jfrog.com/confluence/display/RTF/Docker+Registry#DockerRegistry-LocalDockerRepositories). Other registries, such as Quay.io/Quay Enterprise and OpenShift, are in the process of adding support.
```
coollog: This should be `Docker Image Format V2 Schema 1`

docs/faq.md
```diff
@@ -14,6 +14,8 @@ If a question you have is not answered below, please [submit an issue](/../../is
 [Can I ADD a custom directory to the image?](#can-i-add-a-custom-directory-to-the-image)\
 [Can I build to a local Docker daemon?](#can-i-build-to-a-local-docker-daemon)\
 [I am seeing `ImagePullBackoff` on my pods.](#i-am-seeing-imagepullbackoff-on-my-pods-in-minikube)\
+[How do I configure a proxy?](#how-do-i-configure-a-proxy)\
+[How can I examine network traffic?](#how-can-i-examine-network-traffic)\
```
chanseokoh: Why no entry for "How can I diagnose problems pulling or pushing from remote registries?"

docs/faq.md
chanseokoh: > The registry does not support HTTPS or its certificates are not configured properly.

I'm wondering naybe we should break this down, because what "The registry does not support HTTPS" sounds to me is that it includes the case of an HTTP server, and in this case, "handling registries that use self-signed certificates" does not apply. Also, about "its certificates are not configured properly", it doesn't mean the certificates are always self-signed. Self-signed certificates are one such case, so if the server indeed has a wrong or expired certificate, installing such certificates to a local trusted CA store does not seem like a good idea. Moreover, although not sure, but especially if the certificate is expired, it may never succeed even after installing it locally?

> we do recommend its use (_Jib 0.9.8_)

It's not clear what "its use" means: installing the self-signed certificates or using `allowInsecureRegistries`? If recommending the latter, I'd move it to the top.

docs/faq.md
chanseokoh: Oh, is it confirmed that `OpenShift` does not support the Schema 2?

coollog: My bad, I misread this formatting of this answer. This is correct.

coollog: Oh, but replacing `s/Registry/Image Format/`

docs/faq.md
briandealwis: I'll separate this into SSL-related and HTTP-related.

_Its_ refers to the `allowInsecureRegistries`.

docs/faq.md
```diff
@@ -14,7 +14,10 @@ If a question you have is not answered below, please [submit an issue](/../../is
 [Can I ADD a custom directory to the image?](#can-i-add-a-custom-directory-to-the-image)\
 [Can I build to a local Docker daemon?](#can-i-build-to-a-local-docker-daemon)\
 [I am seeing `ImagePullBackoff` on my pods.](#i-am-seeing-imagepullbackoff-on-my-pods-in-minikube)\
-[How do I enable debugging?](#how-do-i-enable-debugging)\
+[How do I configure a proxy?](#how-do-i-configure-a-proxy)\
+[How can I diagnose problems pulling or pushing from remote registries?](#how-can-i-diagnose-problems-pulling-or-pushing-from=-remote=registries)\
```
coollog: Looks like there's a few `=` in the link

docs/faq.md
```diff
@@ -200,9 +203,49 @@ kubectl patch serviceaccount default \
...
+gradle -Djava.util.logging.config.file=path/to/log.properties -DjibSerialize=true ...
+```
+
+### How do I view debug logs for Jib?
+
+Maven: use `mvn -X -DjibSerialize=true` to enable more detailed logging and serialize Jib's actions.
+
+Gradle: use `grade --debug -DjibSerialize=true` to enable more detailed logging and serialize Jib's actions.
+
+See the [`CONTRIBUTING.md`](../CONTRIBUTING.md), found in the root of the Jib repository, for hints on debugging Jib with an IDE.
```
coollog: nit: this link should probably be absolute since this markdown generates the page at https://googlecontainertools.github.io/jib/

docs/faq.md
```diff
@@ -200,9 +203,49 @@ kubectl patch serviceaccount default \
 
 See more at [Using Google Container Registry (GCR) with Minikube](https://ryaneschinger.com/blog/using-google-container-registry-gcr-with-minikube/).
 
-### How do I enable debugging?
```
coollog: We should probably still keep this entry as a placeholder to remind us to add something about enabling debugging (via the debugger agent and port forwarding). Can be in an HTML comment block, though.

docs/faq.md
```diff
@@ -200,9 +203,49 @@ kubectl patch serviceaccount default \
...
 
-*TODO: Provide solution.*
+Jib currently requires configuring your build tool to use the appropriate [Java networking properties](https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html) (`https.proxyHost`, `https.proxyPort`, `https.proxyUser`, `https.proxyPassword`).
+
+### How can I diagnose problems pulling or pushing from remote registries?
+
+There are a few reasons why Jib may be unable to connect to a remote registry, including:
+
+- **Access requires a proxy.** See [_How do I configure a proxy?_](#how-do-i-configure-a-proxy) for details.
+- **The registry does not support HTTPS.** We do not pass authentication details on non-HTTPS connections, though this can be overridden with the `sendCredentialsOverHttp` system property (_version 0.9.8_).
```
chanseokoh: I think it's worth mentioning here too that in order to connect to plain HTTP registries  `allowInsecureRegistries` should also be enabled but not recommended.

briandealwis:
PTAL

docs/faq.md
```diff
@@ -200,9 +204,51 @@ kubectl patch serviceaccount default \
...
+### How do I view debug logs for Jib?
+
+Maven: use `mvn -X -DjibSerialize=true` to enable more detailed logging and serialize Jib's actions.
+
+Gradle: use `grade --debug -DjibSerialize=true` to enable more detailed logging and serialize Jib's actions.
+
 ### How do I enable debugging?
 
-*TODO: Provide solution.*
+See the [`CONTRIBUTING.md`](https://github.com/GoogleContainerTools/jib/blob/master/CONTRIBUTING.md), found in the root of the Jib repository, for hints on debugging Jib with an IDE.
```
coollog: Oh sorry I meant that this question was actually asking in the lines of "How do I enable debugging when running the container built with Jib so that I can attach a debugger to it and debug the container remotely?" rather than debugging Jib itself.

briandealwis: Oh! So many things to debug.

docs/faq.md
```diff
@@ -248,7 +248,13 @@ Gradle: use `grade --debug -DjibSerialize=true` to enable more detailed logging
 
 ### How do I enable debugging?
 
-See the [`CONTRIBUTING.md`](https://github.com/GoogleContainerTools/jib/blob/master/CONTRIBUTING.md), found in the root of the Jib repository, for hints on debugging Jib with an IDE.
+If using the `distroless/java` base image, then use the [`JAVA_TOOL_OPTIONS`](#how-do-i-set-parameters-for-my-image-at-runtime) to pass along debugging configuration arguments.  For example, to have the remote VM accept debug connections on port 8000, but not suspend:
```
coollog: port 5005*

docs/faq.md
```diff
@@ -248,7 +248,13 @@ Gradle: use `grade --debug -DjibSerialize=true` to enable more detailed logging
...
+```
+-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+```
+
+Then connect your debugger to port 5005 on the given host.  Exposing the debug port will likely require additional configuration of the container runtime environment.
```
coollog: Might be good to add something about port-forwarding for Kubernetes and Docker.

`kubectl port-forward <pod name> 5005:5005`
`docker run -p 5005:5005 <image>`

docs/faq.md
briandealwis: 🤦‍♂️

docs/faq.md
briandealwis: Thx — didn't know the kubectl magic.

docs/faq.md
```diff
@@ -200,9 +204,58 @@ kubectl patch serviceaccount default \
...
+
 ### How do I enable debugging?
 
-*TODO: Provide solution.*
+If using the `distroless/java` base image, then use the [`JAVA_TOOL_OPTIONS`](#how-do-i-set-parameters-for-my-image-at-runtime) to pass along debugging configuration arguments.  For example, to have the remote VM accept debug connections on port 5005, but not suspend:
+```
+-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005
+```
+
+Then connect your debugger to port 5005 on the given host.  You can port-forward the container port to a localhost port for easy access.  Using Docker: `docker run -p 5005:5005 <image>`
```
coollog: nit: Using Docker... on a new line like Using Kubernetes...?

PR #771, opened 2018-08-01
Fall back to HTTP on right condition when HttpHostConnectException

### LINKED ISSUE #767: Try HTTP right away if secure HTTPS fails with HttpHostConnectException if allowInsecureRegistries
Suppose the registry part of an image reference doesn't have a port specified and the registry is HTTP (at port 80).

0.9.7 firsts tries port 443, and it will fail with `HttpHostConnectException` because the port 443 is not listening. 0.9.7 catches the exception and fall backs to HTTP (trying port 80).

In #733, I did not fully understand when exactly `HttpHostConnectException` can happen, so I caught the exception only when attempting an insecure HTTPS, thinking that this preserves the 0.9.7 behavior.

However, in this scenario, we should catch `HttpHostConnectException` at the first attempt of secure HTTPS, and fall back right into HTTP, bypassing insecure HTTPS.

Note that 0.9.8 still works in this scenario if the image reference contains the port (e.g., `localhost:80/repository/image`. Hopefully people usually don't run an HTTP registry on port 80.

chanseokoh:
Fixes #767.

I think it makes sense to me to fall back to HTTP only when certain conditions are met: https://github.com/GoogleContainerTools/jib/issues/767#issuecomment-409723110

Does it make sense?


chanseokoh:
> Looks good - although I'm wondering if there will be many repeated warning messages for someone trying to use an HTTP registry.

That's a valid point. We could add a counter somewhere to print only a limited number of repeated warnings. But I don't think I'd like to add that feature now. I'd think it's a small price to pay for using insecure servers. However, I'll file an issue so that we can discuss.

Issue #727, opened 2018-07-27
Cannot run image running with container generated by jib as non-root in k8s


JamesWojewoda:
**Description of the issue**: I cannot run the image generated via jib-maven-plugin as nonroot. When attempting to set the securitycontext runasuser as a custom one, it errors as the app cannot find the main class. Only root seems to work, but this is a security issue and I'd rather run the container as a unpriviledged user.

**Expected behavior**: set the security context in the k8s deployment as uid 10000 (just an example), and the app thats generated has /app folders owned or at least executable by this uid.

**Steps to reproduce**: create a k8s deployment, set securitycontext: runasUser: to custom uid. use a custom jdk image and use that USER uid in the from block of the maven-jib-plugin section in the pom.xml

**Environment**:

jdk image utilized by pom to generate the new image for the java app:
```
FROM openjdk:8-jdk-alpine

RUN set -x && \
	sed -i -e "s/\#networkaddress.cache.ttl=-1/networkaddress.cache.ttl=300/g" -e "s/securerandom.source=file\:\/dev\/random/securerandom.source=file\:\/dev\/\.\/urandom/g" /usr/lib/jvm/java-1.8-openjdk/jre/lib/security/java.security

ENV JAVA_OPTS="-XX:+UnlockExperimentalVMOptions -XX:+UseCGroupMemoryLimitForHeap"

RUN addgroup -g 10000 runner && \
    adduser -G runner -D -H -u 10000 runner

USER 10000
```
**`jib-maven-plugin` Configuration**: 
```xml
<plugin>
	        <groupId>com.google.cloud.tools</groupId>
	        <artifactId>jib-maven-plugin</artifactId>
	        <version>0.9.7</version>
	        <configuration>
	          <from>
	          	<image>XXXXXXX.dkr.ecr.us-east-1.amazonaws.com/foo:custom-jdk</image>
	          </from>
	          <to>
	            <image>XXXXXXX.dkr.ecr.us-east-1.amazonaws.com/foo:java-app</image>
	            <credHelper>ecr-login</credHelper>
	          </to>
	          <container>
	          	<ports>
	          		<port>8080</port>
	          	</ports>
	          </container>
	          <imageFormat>Docker</imageFormat>
	        </configuration>
	      </plugin>
```

**Log output**: Error: Could not find or load main class com.test.hello.HelloApplication

**Additional Information**: the folders that are generated in /app are owned by root. resources is available to all users, but the other 2 are root write and executable only.

[...]

PR #772, opened 2018-08-01
Adds entries for all directories in layer tarball.

### LINKED ISSUE #523: /app not readable when running on Openshift
**/app folder permissions is set to 0700** on docker  version 1.13.1 on Redhat Enterprice 7.5:
```
/ # ls -al / | grep app
drw-------    1 root     root          4096 Jul 10 07:49 app
```

**Expected behavior**:
/app folder permissions should be 0755, as it is with docker version 18.03.1-ce

```
/ # ls -al / | grep app
drwxr-xr-x    1 root     root          4096 Jul 10 07:49 app
```

**Steps to reproduce**:
Build using configuration pasted below

**Environment**:
Server:
 Version:         1.13.1
 API version:     1.26 (minimum version 1.12)
 Package version: docker-1.13.1-63.git94f4240.el7.x86_64
 Go version:      go1.9.2
 Git commit:      94f4240/1.13.1
 Built:           Mon Apr 30 15:45:42 2018
 OS/Arch:         linux/amd64
 Experimental:    false
   Static hostname: node
         Icon name: computer-server
           Chassis: server
        Machine ID: <machine_id>
           Boot ID: <boot_it>
  Operating System: OpenShift Enterprise
       CPE OS Name: cpe:/o:redhat:enterprise_linux:7.5:GA:server
            Kernel: Linux 3.10.0-862.3.3.el7.x86_64
      Architecture: x86-64

```xml
        <plugin>
          <version>0.9.2</version>
          <configuration>
            <from>
              <image>my.local/registry/openjdk:8-alpine</image>
            </from>
            <to>
              <image>my.local/registry/${project.artifactId}:${project.version}</image>
            </to>
            <container>
              <ports>
                <port>8080</port>
                <port>8081</port>
              </ports>
              <jvmFlags>-Djava.security.egd=file:/dev/./urandom</jvmFlags>
            </container>
          </configuration>
        </plugin>
```

**Log output**: 
Error: Could not find or load main class my.package.Application

coollog:
Fixes #523 
Works towards #727

This makes sure that permissions for directories in the container filesystem will all be the default 755.


jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -44,6 +45,12 @@
       throws IOException {
     List<TarArchiveEntry> tarArchiveEntries = new ArrayList<>();
 
+    // Adds the extraction path itself.
+    TarArchiveEntry extractionPathDirectoryEntry =
+        new TarArchiveEntry(Paths.get(".").toFile(), layerEntry.getExtractionPath());
```
coollog: Not sure how to do this better than just using the current directory `.`. Otherwise, we would have to either have a directory to use, or duplicate a lot of the logic in `TarArchiveEntry`'s constructor.

TadCordle: Maybe add a more explicit comment about this? (e.g. `// Adds the extraction path itself; we need to add some existing directory to the {@link TarArchiveEntry}, so to do this, we just add the current directory named as the extraction path`)

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -45,7 +45,8 @@
       throws IOException {
     List<TarArchiveEntry> tarArchiveEntries = new ArrayList<>();
 
-    // Adds the extraction path itself.
+    // Adds the extraction path itself; to do this, are adding the current directory to act as the
```
TadCordle: to do this, **we** are adding

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -44,6 +45,13 @@
       throws IOException {
     List<TarArchiveEntry> tarArchiveEntries = new ArrayList<>();
 
+    // Adds the extraction path itself; to do this, we are adding the current directory to act as
+    // the file input (since all directories are treated the same in TarArchiveEntry).
+    TarArchiveEntry extractionPathDirectoryEntry =
+        new TarArchiveEntry(Paths.get(".").toFile(), layerEntry.getExtractionPath());
```
chanseokoh: Given https://github.com/GoogleContainerTools/jib/issues/727#issuecomment-409951959, I guess the extraction path is like `/app/classes`, this directory gets the readable permission, but `/app` does not.

coollog: Hmm, yea that might be it. Will try to think of a good way to add parent directories of the extraction too.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -102,9 +123,28 @@ public ReproducibleLayerBuilder addFiles(List<Path> sourceFiles, String extracti
...
+    // Adds the extraction paths (along with any parent directories) to explicitly set permissions
+    // for those directories.
+    LinkedHashSet<String> directoriesToAdd = new LinkedHashSet<>();
+    for (LayerEntry layerEntry : layerEntries) {
+      String extractionPath = layerEntry.getExtractionPath();
+      for (Path parentDirectory : getParentDirectories(Paths.get(extractionPath))) {
+        directoriesToAdd.add(parentDirectory.toString());
+      }
+    }
+    // We are using the current directory to act as the file input (since all directories are
```
loosebazooka: I'm not sure I understand this comment

chanseokoh:
Unfortunately, this still doesn't fix https://github.com/GoogleContainerTools/jib/issues/727#issuecomment-410043365 :(

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
coollog: The `TarArchiveEntry` constructor can use a directory and sets up defaults like normalizing the path, normalizing the end separator (adds `/` if not present), and sets default permissions. It just needs to check `file.isDirectory`, so any `File` that is a directory will do.

coollog: I think we could think about trying to find a better `TarArchiveEntry` or implement our own...

coollog:
@chanseokoh Yea, perhaps we need to just set default permissions for all files.

coollog:
This can be merged and I'll file a followup PR for adding all directories to have default directory mode.

chanseokoh:
I thought this was to set the 755 permissions to the directories (which turned out to work only for /app, /app/classes, app/resources, app/lib, ... and not for directories nested in them). So, if you are going to have a PR that will make all directories to have the default directory mode, I think this PR is not necessary? Or, is this PR still necessary and fix something?

coollog:
@chanseokoh That is true, I'll just modify this PR then.

loosebazooka:
is this PR ready?

coollog:
This PR is in terms of what it does, but it still doesn't seem to fix https://github.com/GoogleContainerTools/jib/issues/727#issuecomment-411517143

coollog:
This fixes #523 (https://github.com/GoogleContainerTools/jib/issues/523#issuecomment-416114189) and can be merged.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -101,15 +154,35 @@ public ReproducibleLayerBuilder addFiles(List<Path> sourceFiles, String extracti
...
-      filesystemEntries.addAll(buildAsTarArchiveEntries(layerEntry));
+    List<LayerEntry> layerEntries = this.layerEntries.build();
+    for (LayerEntry layerEntry : layerEntries) {
+      // Converts layerEntry to list of TarArchiveEntrys.
+      List<TarArchiveEntry> tarArchiveEntries = buildAsTarArchiveEntries(layerEntry);
+
+      for (TarArchiveEntry tarArchiveEntry : tarArchiveEntries) {
+        // Adds all directories along extraction paths to explicitly set permissions for those
+        // directories.
+        for (Path parentDirectory : getParentDirectories(Paths.get(tarArchiveEntry.getName()))) {
```
briandealwis: I wonder if this wouldn't be better in TarStreamBuilder.  This approach also generates a lot of garbage since for only the first file per parent actually needs to walk the parents.

An alternative would be to first sort the file list, then walk the list and add directory elements all at once to the TarStreamBuilder, and then add the files.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -34,6 +38,36 @@
  */
 public class ReproducibleLayerBuilder {
 
+  // Uses the current directory to act as the file input to TarArchiveEntry (since all directories
+  // are treated the same in TarArchiveEntry).
+  private static final File DIRECTORY_FILE = Paths.get(".").toFile();
+
+  /** Decorates an object with a separate object for {@link #hashCode}. */
```
briandealwis: I'd appreciate detail as to why this class is required.  Is `Path#hashCode()` possibly inconsistent?  (`UnixPath#hashCode()` looks fine)

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -34,6 +38,44 @@
  */
 public class ReproducibleLayerBuilder {
 
+  // Uses the current directory to act as the file input to TarArchiveEntry (since all directories
+  // are treated the same in TarArchiveEntry).
+  private static final File DIRECTORY_FILE = Paths.get(".").toFile();
+
+  /** Holds a list of {@link TarArchiveEntry}s with unique extraction paths. */
```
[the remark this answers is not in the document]
coollog: Added a note, though the additions do not necessary need to be regular files.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -34,6 +38,48 @@
  */
 public class ReproducibleLayerBuilder {
 
+  // Uses the current directory to act as the file input to TarArchiveEntry (since all directories
+  // are treated the same in TarArchiveEntry).
+  private static final File DIRECTORY_FILE = Paths.get(".").toFile();
+
+  /**
+   * Holds a list of {@link TarArchiveEntry}s with unique extraction paths. The list also consists
```
briandealwis: s/consists of/includes/

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -34,6 +38,48 @@
  */
 public class ReproducibleLayerBuilder {
 
+  // Uses the current directory to act as the file input to TarArchiveEntry (since all directories
+  // are treated the same in TarArchiveEntry).
+  private static final File DIRECTORY_FILE = Paths.get(".").toFile();
```
briandealwis: Move into `UniqueTarArchiveEntries`?  Make as a java doc and include a link to TarArchiveEntry constructor as a `@seealso`?

briandealwis:
👍

PR #773, opened 2018-08-01
Add warning when auth configuration is incomplete

### LINKED ISSUE #759: Show warning when ignoring auth section due to null username or password
From https://github.com/GoogleContainerTools/jib/issues/712#issuecomment-409394470

If either auth.username or auth.password is null, we just silently ignores the auth section:
```java
  /** Converts the {@link ImageParameters} to an {@link Authorization}. */
  @Internal
  @Nullable
  Authorization getImageAuthorization() {
    if (auth.getUsername() == null || auth.getPassword() == null) {
      return null;
    }

    return Authorizations.withBasicCredentials(auth.getUsername(), auth.getPassword());
  }
```
(This is the Gradle plugin code, but probably we are doing the same thing for Maven.)

However, if the user put the auth section but either value is null, the user is mostly likely doing something wrong. In this case, instead of silently ignoring the auth section, we should print out a warning. (No need to error out, of course.) That should have resolved #712 at the beginning.

TadCordle:
Fixes #759


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -145,14 +144,52 @@ static ImageReference parseImageReference(String image, String type) {
...
+      String usernameProperty,
+      String passwordProperty,
+      AuthConfiguration auth) {
     // System property takes priority over build configuration
     String commandlineUsername = System.getProperty(usernameProperty);
     String commandlinePassword = System.getProperty(passwordProperty);
     if (commandlineUsername != null && commandlinePassword != null) {
       return Authorizations.withBasicCredentials(commandlineUsername, commandlinePassword);
+    } else {
+      if (commandlinePassword != null) {
```
TadCordle: I'm not sure how I feel about putting this logic in the configuration classes. I'm expecting we'll be refactoring the plugins pretty soon in preparation for the jib-plugins-common module though, so I decided to put it here until we have that structure figured out. However, I can move this somewhere else now if you want.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ImageParameters.java
```diff
@@ -70,22 +71,42 @@ public void setCredHelper(String credHelper) {
...
+   *
+   * @param logger the {@link JibLogger} used to print warnings
+   * @param imageProperty the image configuration's name (i.e. "from" or "to")
+   * @return the {@link Authorization}, or null if the username and password aren't both configured
+   */
   @Internal
   @Nullable
-  Authorization getImageAuthorization() {
+  Authorization getImageAuthorization(JibLogger logger, String imageProperty) {
     if (auth.getUsername() == null || auth.getPassword() == null) {
```
chanseokoh: I'm tempted to also warn when these are `.trim().isEmpty()` (although in such cases, we won't return null). WDYT?

TadCordle: Hmm, I'm not sure about usernames, but I could imagine someone making their password all whitespace.

chanseokoh: Alright, maybe just `.isEmpty()`? I think it'll be fine because we don't error but just warn.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/AuthParameters.java
```diff
@@ -24,7 +24,7 @@
  * A bean that configures authorization credentials to be used for a registry. This is configurable
  * with Groovy closures and can be validated when used as a task input.
  */
-public class AuthConfiguration {
+public class AuthParameters {
```
coollog: just a question: what is the reasoning for this rename?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ImageParameters.java
```diff
@@ -70,22 +71,42 @@ public void setCredHelper(String credHelper) {
...
+      if (auth.getPassword() != null) {
```
coollog: Wouldn't this warn if `jib.auth` was null too? (aka should not warn if user did not define `jib.auth`)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/AuthParameters.java
TadCordle: Consistency (`ContainerParameters`, `ImageParameters`, `AuthConfiguration` -> `AuthParameters`)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ImageParameters.java
TadCordle: If the user doesn't configure auth, both username and password should both be null.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
coollog: Maybe where this authorization is used (ie. Build...Mojo where it makes the knownRegistryCredentials)? (So it would try to fetch and validate on it)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ImageParameters.java
coollog: Ah right - maybe it would make more sense to split this logic into 3 parts (vs entwining the 3 cases in if-else-fallthrough):

both are null - return null
only password is null - warn about password
only username is null - warn about username

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
TadCordle: Yeah, but there are 3 files in each plugin where that happens. I know we're going to be trying to reduce copy-paste soon, but it seemed like a little much.

coollog: I think if we keep the actual snippet of code that does the knownRegistryCredential setup the same (with this added), it can actually help in the refactoring process so that we have one chunk of the same code to move and refactor to a common place (compared to trying to find this code hidden in the configuration class to factor away too).

TadCordle: Ok, but we should refactor soon because this is going to be really bad.

coollog: Yes, definitely, should be a top priority since the library depends on the frontend package being refactored away too

chanseokoh: Not following this, so are we good now, or will you make further changes in this PR?

TadCordle: I think it should be good for review.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -174,4 +179,42 @@ BuildImageTask setJibExtension(JibExtension jibExtension) {
     this.jibExtension = jibExtension;
     return this;
   }
+
+  /**
+   * Converts the {@link ImageParameters} to an {@link Authorization}.
```
chanseokoh: I don't see `ImageParameters` in the method signature. `AuthParameters`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -164,4 +168,41 @@ BuildDockerTask setJibExtension(JibExtension jibExtension) {
     this.jibExtension = jibExtension;
     return this;
   }
+
+  /**
+   * Converts the {@link ImageParameters} to an {@link Authorization}.
```
chanseokoh: I don't see `ImageParameters` in the method signature. `AuthParameters`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -193,4 +197,41 @@ BuildTarTask setJibExtension(JibExtension jibExtension) {
     this.jibExtension = jibExtension;
     return this;
   }
+
+  /**
+   * Converts the {@link ImageParameters} to an {@link Authorization}.
```
chanseokoh: I don't see `ImageParameters` in the method signature. `AuthParameters`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
TadCordle: Oops yeah, outdated comments.

TadCordle:
Should this have a changelog entry as well, or is this a small enough change that it's not worth mentioning?

coollog:
Yeah, this can be left out since it's not something users need to be informed of since it's just a log message.

PR #786, opened 2018-08-02
Consolidates Maven- and GradleLayerConfigurations.

### LINKED ISSUE #710: Have DockerContextGenerator take LayerConfigurations instead of LayerEntrys.
From https://github.com/GoogleContainerTools/jib/pull/516#pullrequestreview-140073132

This makes the class more maintainable as we add more layer types, and also makes it more usable as part of the eventual library.

### LINKED ISSUE #711: Consolidate Gradle- and MavenLayerConfigurations.
Related to #478 
From https://github.com/GoogleContainerTools/jib/pull/516/files#diff-33ad1cd8b95da047edb9ad28b1f3c6b5R160

coollog:
Fixes #711 
Fixes #710

- Renames `DockerContextGenerator` to `JavaDockerContextGenerator` because it is Java-specific
- Adds `JavaLayerConfigurations` to build Java-specific `LayerConfiguration`s
- Various other refactoring to make these changes work


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfigurations.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/** Defines a list of {@link LayerConfiguration}s. */
+public class LayerConfigurations {
+
+  private final ImmutableList<LayerConfiguration> layerConfigurations;
+  private final ImmutableMap<String, LayerConfiguration> layerConfigurationMap;
```
loosebazooka: Seems strange to store both these?
Why not just `ImmutableMap`? it stores it in insertion order (like `LinkedHashMap`)?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -0,0 +1,134 @@
...
+import com.google.cloud.tools.jib.configuration.LayerConfigurations;
+import com.google.cloud.tools.jib.image.LayerEntry;
+import com.google.common.base.Preconditions;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/** Builds {@link LayerConfiguration}s for a Java application. */
+public class JavaLayerConfigurations extends LayerConfigurations {
```
loosebazooka: I'm not convinced that it's necessary that `JavaLayerConfigurations` extends `LayerConfigurations`. Personally, stylistically, I would avoid inheritance, and just have `JavaLayerConfigurations` implement some interface that defined a `LayerConfigurationCollection` type. And have `LayerConfigurations` be our default implementation for `jib-core` users. The lines like `return Preconditions.checkNotNull(getByLabel(RESOURCES_LAYER_LABEL)).getLayerEntries().get(0);` look kinda weird to me.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurations.java
```diff
@@ -124,79 +114,14 @@ static GradleLayerConfigurations getForProject(
...
-        dependenciesLayerConfiguration,
-        snapshotDependenciesLayerConfiguration,
-        resourcesLayerConfiguration,
-        classesLayerConfiguration,
-        extraFilesLayerConfiguration);
-  }
-
-  LayerEntry getDependenciesLayerEntry() {
-    return dependenciesLayerConfiguration.getLayerEntries().get(0);
+    return JavaLayerConfigurations.builder()
```
loosebazooka: this code shrinkage is really nice :+1:

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
loosebazooka: Separating them out might also simplify the implementation of `LayerConfigurations` which seems (to me) to only have labeling to enable the `JavaLayerConfigurations` class?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -0,0 +1,129 @@
...
+    public Builder setExtraFiles(List<Path> extraFiles) {
+      layerFilesMap.put(LayerType.EXTRA_FILES, extraFiles);
+      return this;
+    }
+
+    public JavaLayerConfigurations build() {
+      ImmutableMap.Builder<LayerType, LayerConfiguration> layerConfigurationsMap =
+          ImmutableMap.builderWithExpectedSize(LayerType.values().length);
+      for (LayerType layerType : LayerType.values()) {
+        List<Path> layerFiles = Preconditions.checkNotNull(layerFilesMap.get(layerType));
```
TadCordle: Question - why use a builder rather than just a constructor if all of the arguments are required?

chanseokoh: I haven't looked into the code, but yeah, I think we don't have to be too obsessed with the Builder pattern. And even if some parameters are optional, I personally think sometimes it's just more succinct and convenient to have multiple constructors if there are a small parameters to set. I also find nothing really wrong with having both (a few) public constructors and a builder, if manageable.

TadCordle: I guess it makes sense if you want to break up the construction of the builder (i.e. create builder, do some logic, set first parameter, do some logic, set second parameter, etc. until everything is in and we can `build()`). But in our case it looks like we only ever add everything to the builder all at once.

coollog: Yep, we could have separate constructors for this class, but the main purpose of having the builder is to avoid an explosion of constructors - ie. each time we add a new field to set, a new constructor would have to be added with a larger set of parameters, and we would end up having a large set of non-idempotent constructors that may be easy to misuse (by ordering the parameters incorrectly).

chanseokoh: Yeah, that's one of the problems that the Builder pattern can solve: the explosion of constructors. But we are not having a explosive number of constructors. IMO, if you are having just 3-4 constructors and a very small set of optional parameters, I think the pattern is an over-engineering that clutters code unnecessarily. Currently, `LayerConfiguration` has only two parameters, so I'm with @TadCordle about the matter. If the number of constructors and optional parameters grows later, then that's the time when Builder pattern should be considered. As an analogy, `IOException` has four constructors (0-arg, with a message, with a cause, and with a message+cause), and I'd gripe if I always have to use its Builder that blows up the code into 4~5 lines every time I need to create an exception.

coollog: Oh, this is referring to `LayerConfiguration`. For that, the main purpose for the builder was to allow for adding `LayerEntry`s without needing too much boilerplate code. Without the builder, it would require calling like:

```java
new LayerConfiguration(
    Arrays.asList(
        new LayerEntry(ImmutableList.copyOf(sourceFiles1), destinationOnImage1)
        new LayerEntry(ImmutableList.copyOf(sourceFiles2), destinationOnImage2)),
    label);
```

rather than

```java
LayerConfiguration.builder()
    .addEntry(sourceFiles1, destinationOnImage1)
    .addEntry(sourceFiles2, destinationOnImage2)
    .setLabel(label)
    .build();
```

chanseokoh: Oh, missed that part of the code. That makes sense. That said, now it seems like I was the only one talking about `LayerConfiguration`, haha.

TadCordle: I was talking specifically about `JavaLayerConfiguration`, not `LayerConfiguration`; is 5 parameters too many for a constructor? It just seems weird to assign a bunch of non-optional values to a builder using separate method calls then have a method call that specifically makes sure none of those values are missing and copies them somewhere else to actually make the thing you want, rather than just constructing the thing you want directly with missing parameters being verified by the compiler.

> each time we add a new field to set, a new constructor would have to be added with a larger set of parameters

If the new field were required, we'd just be changing the same constructor, not adding more. If not, then yeah maybe we'd have to add another constructor, or we can just use a setter. Either way, we don't really know at the moment how many parameters we might add, if any, do we? I generally don't think it's a good idea to plan so far in advance for hypotheticals. If the hypothetical happens, then that's cool that we don't have to refactor as much as we would have needed to, but if it doesn't happen, we're left with code that's over-engineered because we were trying to solve a non-issue.

> that may be easy to misuse (by ordering the parameters incorrectly)

This is the only thing we'd actually need to worry about at the moment; all the parameters being the same type of `List<Path>` makes it a lot more error prone, especially if we ever changed the order of the parameters in the constructor ourselves. So I am more inclined to agree with you. That said, it is still a hypothetical situation...

coollog: > It just seems weird to assign a bunch of non-optional values to a builder using separate method calls then have a method call that specifically makes sure none of those values are missing and copies them somewhere else to actually make the thing you want, rather than just constructing the thing you want directly with missing parameters being verified by the compiler.

I'm intending the setters on `JavaLayerConfigurations#Builder` to be optional so that users can set only the layers that they need to.

> If the new field were required, we'd just be changing the same constructor, not adding more. If not, then yeah maybe we'd have to add another constructor, or we can just use a setter.

Once constructors are set, it's quite difficult to change them. It would have to require a deprecation cycle before a certain constructor can be safely removed, and it might not be good practice to deprecate constructors in the first place since those will be the most-used methods on a class (deprecating the constructor would break all uses of that class). Also, changing a constructor via rearranging parameters would probably not be safe to do at all, since it could break uses in invisible ways.

> Either way, we don't really know at the moment how many parameters we might add, if any, do we?

Yea, we will probably be adding more parameters in the future depending on how we want to structure more layer types.

chanseokoh: > I'm intending the setters on `JavaLayerConfigurations#Builder` to be optional so that users can set only the layers that they need to.

Looking at the code, I don't think the setters on `JavaLayerConfigurations#Builder` are optional now. With the `Precondtions` check below, currently the client of the `Builder` should call all the setters of the `Builder`. They must provide all the five layers.

```java
for (LayerType layerType : LayerType.values()) {
        List<Path> layerFiles = Preconditions.checkNotNull(layerFilesMap.get(layerType));
```

If you intent the setters to be optional, I think you should remove the precondition check?

chanseokoh: Ah, my bad. Seems like an empty values were put at the start of the `builder()`. Please disregard my previous comment.

TadCordle: > I'm intending the setters on JavaLayerConfigurations#Builder to be optional so that users can set only the layers that they need to.

Ok, I looked back and realized I missed the part in the builder's constructor where the map is preloaded with an entry for each layer type. I assumed it just started out empty and the `Preconditions.checkNotNull` would just fail if the user didn't call every setter. My fault.

> (deprecating the constructor would break all uses of that class)

which is why I was confused about this whole point, because it made no difference here whether we used a constructor or a builder without optional parameters. Makes sense now.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -0,0 +1,129 @@
...
+      }
+      return new JavaLayerConfigurations(layerConfigurationsMap.build());
+    }
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  private final ImmutableMap<LayerType, LayerConfiguration> layerConfigurationMap;
```
loosebazooka: Any reason we can't use the layers directly instead of using a map? I would think our JavaLayerConfiguration is going to be pretty static.

loosebazooka: If this is to fit in with some future interface, I think we should include the interface in this PR? And maybe even then, could `JavaLayerConfigurations` hide the "map"-ness of the interface and provide stronger accessors?

Issue #594, opened 2018-07-12
Reduce duplicate code in image building tasks/mojos


TadCordle:
- [x] Maven/gradle specific code
- [x] Auth configuration
- [x] `ProjectProperties`

coollog:
We could probably organize this in the framework style like how the registry requests work.

coollog:
Related to https://github.com/GoogleContainerTools/jib/issues/478 as well

TadCordle:
I think this is done for the most part. Feel free to reopen if you think there's more that should be done.

PR #793, opened 2018-08-03
Reduce duplicate code in gradle plugin

TadCordle:
Part of #594. Next step is maven plugin, and after that seeing how much between the maven/gradle common code we can move to the plugins package in jib-core.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTaskConfigurator.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTaskConfigurator.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.gradle;
+
+import com.google.cloud.tools.jib.JibLogger;
+import com.google.cloud.tools.jib.configuration.BuildConfiguration;
+import com.google.cloud.tools.jib.configuration.CacheConfiguration;
+import com.google.cloud.tools.jib.configuration.ContainerConfiguration;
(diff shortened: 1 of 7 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTaskConfigurator.java
```diff
@@ -0,0 +1,172 @@
...
+import org.gradle.api.GradleException;
+
+/** Configures and provides builders for the image building tasks. */
+class BuildTaskConfigurator {
+
+  /** {@code User-Agent} header suffix to send to the registry. */
+  private static final String USER_AGENT_SUFFIX = "jib-gradle-plugin";
+
+  /**
+   * Sets up configuration that is common among the image building tasks. This includes setting up
```
coollog: "Sets up the `{@link BuildConfiguration}`..."

I think it's good to be clear in the distinction between the plugin configuration and the Jib execution configuration (`BuildConfiguration`). This class is essentially meant to convert the plugin-level configuration into a canonical Jib execution configuration without parts that vary amongst the tasks..

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTaskConfigurator.java
```diff
@@ -0,0 +1,172 @@
...
+import com.google.cloud.tools.jib.plugins.common.SystemPropertyValidator;
+import com.google.cloud.tools.jib.registry.RegistryClient;
+import com.google.cloud.tools.jib.registry.credentials.RegistryCredentials;
+import com.google.common.base.Strings;
+import java.time.Instant;
+import javax.annotation.Nullable;
+import org.gradle.api.GradleException;
+
+/** Configures and provides builders for the image building tasks. */
+class BuildTaskConfigurator {
```
coollog: Not sure about the naming here haha. We could probably name this more along the lines of a `PluginConfigurationProcessor`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
```diff
@@ -36,23 +36,23 @@
...
    *
    * @param logger the logger used to display messages.
    * @param jibExtension the {@link JibExtension} providing the configuration data
    * @param projectProperties used for providing additional information
-   * @return a new {@link BuildTaskConfigurator} containing pre-configured builders
+   * @return a new {@link PluginConfigurationProcessor} containing pre-configured builders
    * @throws InvalidImageReferenceException if parsing the base image configuration fails
    */
-  static BuildTaskConfigurator newCommonConfiguration(
+  static PluginConfigurationProcessor newCommonConfiguration(
```
coollog: perhaps this should be name along the lines of `processCommonConfiguration`? `newCommonConfiguration` may give the meaning of constructing an object of type `CommonConfiguration`.

Issue #720, opened 2018-07-26
Questions around secure/insecure registry connection


chanseokoh:
### First Question

To create a `RegistryAuthenticator` instance, one should use `RegistryAuthenticator.Initializer.initialize()`. It may return `null` if the registry uses insecure HTTPS or plain HTTP.

```java
    public RegistryAuthenticator initialize() throws ..., RegistryException {
      ...
      } catch (InsecureRegistryException ex) {
        // Cannot skip certificate validation or use HTTP, so just return null.
        return null;
      }
```

There are two places where a `RegistryAuthenticator` instance is created: `AuthenticatePushStep.call()` and `PullBaseImageStep.call()`. And they handle `null` (insecure HTTPS or plain HTTP) differently.

`PullBaseImageStep`:
```java
          RegistryAuthenticator registryAuthenticator =
              RegistryAuthenticator.initializer(...) ... .initialize();
          if (registryAuthenticator == null) {
            ... .error(
                    "Failed to retrieve authentication challenge for registry that required token authentication");
            throw registryUnauthorizedException;
          }
          registryCredentials =
              registryAuthenticator.setAuthorization(registryCredentials).authenticatePull();
```

`AuthenticatePushStep`:
```java
      RegistryAuthenticator registryAuthenticator =
          RegistryAuthenticator.initializer(...) ... .initialize();
      if (registryAuthenticator == null) {
        return registryCredentials;
      }
      return registryAuthenticator.setAuthorization(registryCredentials).authenticatePush();
```
Why? I'm trying to understand the difference of the steps for the logic.

### Second Question

The core functionality of `RegistryAuthenticator` is the eventual method `authenticate()` which contacts the URL `<realm>?service=<service>&scope=repository:<repository>:<scope>`. I'm thinking `<realm>` would probably contain a protocol (e.g., `https://`), but I'm not so sure. Anyways, currently `authenticate()` is not being affected by `allowInsecureRegistries`, as it doesn't use `RegistryEndpointCaller` to make a call which handles `allowInsecureRegistries`. Instead, it establishes a `Connection` on its own.

```java
      try (Connection connection = Connection.getConnectionFactory().apply(authenticationUrl)) {
        Request.Builder requestBuilder =
            Request.builder().setHttpTimeout(Integer.getInteger("jib.httpTimeout"));
        if (authorization != null) {
          requestBuilder.setAuthorization(authorization);
        }
```

[...]

PR #797, opened 2018-08-03
Apply allowInsecureRegistries to RegistryAuthenticator too

chanseokoh:
Tests are completely broken. This is for early discussion. Please review the general approach, not details.

This is about the second part in #720, the issue that `allowInsecureRegistries` has no effect on `RegistryAuthenticator`. My approach here is

- Refactor the `allowInsecureRegistries` logic into a lower-level utility-like class, `EndpoinCaller`.
- Both `RegistryEndpointCaller` and `RegistryAuthenticator` uses `EndpointCaller` for handling `allowInsecureRegistries`.
- Rename `RegistryException` to `EndpointException`, because it's common that the auth server is different from a registry server.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/AuthenticationMethodRetrieverIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/AuthenticationMethodRetrieverIntegrationTest.java
@@ -29,7 +29,7 @@ public class AuthenticationMethodRetrieverIntegrationTest {
 
   @Test
   public void testGetRegistryAuthenticator()
-      throws RegistryAuthenticationFailedException, IOException, RegistryException {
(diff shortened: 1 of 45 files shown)


chanseokoh:
But at the same time, I think it would be pretty rare that an auth server is insecure. https://github.com/GoogleContainerTools/jib/issues/720#issuecomment-410271972 may be a very special case where SSL validation fails unpredictably possibly due to some issue in the customer's local environment.

So, now I think we don't really need to try to apply `allowInsecureRegistries` to `RegistryAuthenticator` at least for now. If we find actual use-cases, we can consider it then.

chanseokoh:
Per https://github.com/GoogleContainerTools/jib/issues/720#issuecomment-410376321, I'm closing this. If we later find a case where the auth server is required and not secure, we can revisit the branch. (You can always restore a closed PR.) However, as I said in the comment, I think it would be extremely rare that an insecure registry requires talking to an auth server (even if the server is self).

PR #801, opened 2018-08-06
Consolidate auth validation in maven/gradle plugins

TadCordle:
Part of #594.

A couple things to note:

1. This change makes the system properties `jib.to.auth.username` and `jib.to.auth.password` available for use in the gradle plugin as well as the maven plugin.
2. Apparently incorrect credentials were being used in the validation function in `BuildImageTask` (see [this line](https://github.com/GoogleContainerTools/jib/compare/consolidate-auth-config?expand=1#diff-17f1a336c0588462ae5457e9870fb7d7L86)), it should be fixed now. I think this was from the friday refactoring, so I don't think it was in any releases.


chanseokoh:
> Apparently incorrect credentials were being used in the validation function in BuildImageTask (see this line), it should be fixed now.

Since it really happened and the tests could not catch it, it makes sense to add a test.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/AuthParameters.java
```diff
@@ -24,14 +25,15 @@
...
-public class AuthParameters {
+public class AuthParameters implements AuthProperty {
 
   @Nullable private String username;
   @Nullable private String password;
 
   @Input
-  @Nullable
   @Optional
+  @Override
```
coollog: nit: Not specific to this PR, but I was thinking maybe we could follow a convention of alphabetical ordering for annotations just for consistency in the future - not important though.

jib-core/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -0,0 +1,115 @@
...
+
+import com.google.cloud.tools.jib.JibLogger;
+import com.google.cloud.tools.jib.http.Authorization;
+import com.google.cloud.tools.jib.http.Authorizations;
+import com.google.common.base.Strings;
+import java.util.function.Function;
+import javax.annotation.Nullable;
+
+/** Validator for system properties. */
+public class ConfigurationPropertyValidator {
```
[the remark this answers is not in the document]
TadCordle: That's what it used to be named, but with the addition of `getImageAuth`, it's validating more than just system properties, so I renamed it.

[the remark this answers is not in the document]
coollog: Ah okay, maybe change the javadoc to be lke `Validator for plugin configuration parameters and system properties`?

jib-core/src/main/java/com/google/cloud/tools/jib/plugins/common/AuthProperty.java
```diff
@@ -0,0 +1,33 @@
...
+/** Holds a username and password property. */
+public interface AuthProperty {
+
+  @Nullable
+  String getUsername();
+
+  @Nullable
+  String getPassword();
+
+  String getUsernamePropertyDescriptor();
```
coollog: Should have some short javadoc about how to implement these methods.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/AuthParameters.java
```diff
@@ -16,22 +16,48 @@
...
+  @Nullable private String usernameDescriptor;
+  @Nullable private String passwordDescriptor;
+
+  @Internal
+  @Override
+  public String getUsernamePropertyDescriptor() {
+    return Preconditions.checkNotNull(usernameDescriptor);
+  }
+
+  void setUsernamePropertyDescriptor(String usernameDescriptor) {
```
coollog: These could be set via a constructor. (`ObjectFactory#newInstance` takes constructor parameters)

TadCordle:
Filed an issue for https://github.com/GoogleContainerTools/jib/pull/801#issuecomment-410793210

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ImageParameters.java
```diff
@@ -39,8 +39,12 @@
   @Nullable private String credHelper;
 
   @Inject
-  public ImageParameters(ObjectFactory objectFactory) {
-    auth = objectFactory.newInstance(AuthParameters.class);
+  public ImageParameters(ObjectFactory objectFactory, String imageDescriptor) {
+    auth =
+        objectFactory.newInstance(
+            AuthParameters.class,
+            imageDescriptor + ".auth.username",
```
coollog: Oh, right this can actually be just `descriptor + ".auth"`, and then `AuthParameters` adds the `username` and `password` suffixes. Essentially, the idea is to pass along the descriptor name for each of the parameters at the level at which it is applicable (and the same for Maven).

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -77,7 +96,9 @@ String getPassword() {
 
     @Nullable @Parameter private String credHelper;
 
-    @Parameter private AuthConfiguration auth = new AuthConfiguration();
+    @Parameter
+    private AuthConfiguration auth =
+        new AuthConfiguration("<from><auth><username>", "<from><auth><password>");
```
coollog: This can be done in the same manner:

`new AuthConfiguration(descriptor + "<auth>");`
```
private AuthConfiguration(String descriptor) {
      this.usernameDescriptor = descriptor + "<username>";
      this.passwordDescriptor = descriptor + "<password>";
```

And below for example:

`new FromConfiguration("<from>")`

This way, the same hierarchical model can be applied to the other parameters and fetched via a "getDescriptor" method in other logging code.

TadCordle: I'll change it for the auth related stuff, but I don't want to get too out of scope with this PR by changing up the other parameters.

coollog: Sounds good.

Issue #751, opened 2018-07-31
Label Support


loosebazooka:
Add provision for labels
- [ ] in core
- [ ] in plugins
- [ ] with passthrough
- [ ] set jib builder labels

loosebazooka:
@GoogleContainerTools/java-tools does anyone want to pick this up, I'm a little short on time.

We would be exposing a way for users to add labels to their builds.

TadCordle:
I could do this. Would this just be another configuration option in the `container` block?

PR #804, opened 2018-08-06
Passthrough labels from base image

loosebazooka:
part of #751


coollog:
Oh let's add CHANGELOG entries.

PR #807, opened 2018-08-07
Do not put JIB_INTEGRATION_TESTING_KEY into environment variable

chanseokoh:
I don't see a reason to put this in the environment variable. There is no code using this variable except `continuous.sh`.

With this, the key will no longer exposed in our continuous build log.

### DIFF
--- a/kokoro/continuous.sh
+++ b/kokoro/continuous.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 
 set -e
-
-echo ${JIB_INTEGRATION_TESTING_KEY} > ./keyfile.json
-
 set -x
 
 gcloud components install docker-credential-gcr
@@ -12,7 +9,7 @@ gcloud components install docker-credential-gcr
 export PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/
 
 # docker-credential-gcr uses GOOGLE_APPLICATION_CREDENTIALS as the credentials key file
-export GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/keyfile.json
+export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_KEYSTORE_DIR}/72743_jib_integration_testing_key
 docker-credential-gcr configure-docker
 
 # Stops any left-over containers.


kokoro/continuous.sh
```diff
@@ -12,7 +9,7 @@ gcloud components install docker-credential-gcr
 export PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/
 
 # docker-credential-gcr uses GOOGLE_APPLICATION_CREDENTIALS as the credentials key file
-export GOOGLE_APPLICATION_CREDENTIALS=$(pwd)/keyfile.json
+export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_KEYSTORE_DIR}/72743_jib_integration_testing_key
```
coollog: Wouldn't this mean that the key is still in the environment variable `GOOGLE_APPLICATION_CREDENTIALS`?

chanseokoh: The difference is that the contents of the key were in `JIB_INTEGRATION_TESTING_KEY`, and those contents were being logged in plain text. `GOOGLE_APPLICATION_CREDENTIALS` just holds a path.

coollog: Kk, we'll just have to make sure we don't ever log the contents of that file in our build logs.

loosebazooka: I think my concern is that some malicious PR will expose this and someone accidentally triggers a build for it.

chanseokoh: Yeah, anyone could do `echo $JIB_INTEGRATION_TESTING_KEY` to get the contents of the service account key, so this PR doesn't make things worse. That's why we should never auto-trigger Kokoro builds for an external contributor, which is the current setting.

What I did for CT4E is to use the service key of the GCE machine where Kokoro runs, but unfortunately this required using a dedicated machine pool that we maintain ourselves.

etanshaul: @chanseokoh 

> That's why we should never auto-trigger Kokoro builds for an external contributor, which is the current setting.

How do you configure this?

chanseokoh: Chatted with with @etanshaul.

Issue #803, opened 2018-08-06
Add more thorough authentication tests


TadCordle:
Found a bug recently where we were passing the `from` auth plugin configuration into the `to` image configuration for the jib builder (see https://github.com/GoogleContainerTools/jib/pull/801), and none of our tests seemed to pick it up. We should add more tests to make sure these sorts of things don't happen again.

We should probably add several different kinds of auth configurations in the integration tests to make sure everything works from start to finish.

PR #814, opened 2018-08-08
Add a comprehensive maven integration test

TadCordle:
Adds a test that uses most of jib's parameters, including `<auth>` for both `from` and `to`.

Note that this doesn't cover every auth case. Things we can consider adding:
- [ ] testing with an https registry
- [ ] verifying error messages for refusal to communicate with insecure registry/send credentials over http
- [ ] different registries/credentials for pull/push

Part of #803


jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -192,4 +196,79 @@ public void testExecute_defaultTarget() {
                   + "compile jib:build -Dimage=<your image name>')."));
     }
   }
+
+  @Test
+  public void testExecute_complex()
+      throws IOException, InterruptedException, VerificationException {
+    // Runs the Docker registry.
```
coollog: Let's add a TODO to refactor this into `LocalRegistry`.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -192,4 +196,79 @@ public void testExecute_defaultTarget() {
...
+            "docker",
+            "run",
+            "-d",
+            "-p",
+            "5000:5000",
+            "--restart=always",
+            "--name",
+            containerName,
+            "-v",
+            complexTestProject.getProjectRoot().resolve("auth") + ":/auth",
```
coollog: A comment here would help clarify what this volume mount is for.

jib-maven-plugin/src/test/resources/projects/complex/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+          <to>
+            <image>localhost:5000/complex-image</image>
+            <auth>
+              <username>testuser</username>
+              <password>testpassword</password>
+            </auth>
+          </to>
+          <container>
+            <useCurrentTimestamp>true</useCurrentTimestamp>
+            <args>An argument.</args>
```
coollog: Should we use the usual format for parameter lists here?

jib-maven-plugin/src/test/resources/projects/complex/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+            <mainClass>com.test.HelloWorld</mainClass>
+            <jvmFlags>
+              <jvmFlag>-Xms512m</jvmFlag>
+              <jvmFlag>-Xdebug</jvmFlag>
+            </jvmFlags>
+            <ports>
+              <port>1000/tcp</port>
```
coollog: The test should probably test for these ports actually being exposed.

jib-maven-plugin/src/test/resources/projects/complex/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+              <port>2000-2003/udp</port>
+            </ports>
+            <!-- <format>OCI</format> -->
```
coollog: Could set this format to `Docker`

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -192,4 +196,79 @@ public void testExecute_defaultTarget() {
...
+        Instant.parse(
+            new Command("docker", "inspect", "-f", "{{.Created}}", "localhost:5000/complex-image")
+                .run()
+                .trim());
+    Assert.assertTrue(buildTime.isAfter(before) || buildTime.equals(before));
+    new Command("docker", "logout", "localhost:5000").run();
+
+    // Stops the local registry.
+    try {
+      new Command("docker", "stop", containerName).run();
```
coollog: This should probably go in a `finally` block so it always executes.

jib-maven-plugin/src/test/resources/projects/complex/src/main/java/com/test/HelloWorld.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.IOException;
+import java.lang.management.ManagementFactory;
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/** Example class that uses a dependency and a resource file. */
+public class HelloWorld {
```
coollog: If the code for this `complex` project is mostly the same as the `simple` project, it could just use the same `simple` project, but have a different POM (like `pom-complex.xml`) that can be passed into the verifier with the `--file` CLI option.

jib-maven-plugin/src/test/resources/projects/complex/pom.xml
TadCordle: Doesn't that already happen here? https://github.com/GoogleContainerTools/jib/pull/814/files#diff-dd10efd511098bc1a27585924f70f8f1R247

chanseokoh:
Nice!

Issue #813, opened 2018-08-08
Clean up integration tests


TadCordle:
In addition to adding more comprehensive tests (#76, #803) we should think about cleaning up/speeding up the tests we have.

Cleanup:
- [x] Add optional credentials to `LocalRegistry`
- [x] Remove, or modify `buildAndRun` in maven/gradle to allow more flexibility in which parameters get verified

Speedup:
- [ ] Run tests in parallel?

PR #818, opened 2018-08-09
Allow LocalRegistry to require credentials

TadCordle:
Part of #813

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
@@ -91,8 +91,7 @@ public class LocalRegistry extends ExternalResource {
               "REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd"));
     }
     dockerTokens.add("registry:2");
-    String[] tokenArray = new String[dockerTokens.size()];
-    new Command(dockerTokens.toArray(tokenArray)).run();
+    new Command(dockerTokens).run();
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -17,77 +17,145 @@
...
 
+  private final String containerName = "registry-" + UUID.randomUUID();
   private final int port;
+  private final boolean pullBusyBox;
+  @Nullable private final String username;
+  @Nullable private final String password;
 
-  /** The name for the container running the registry. */
-  private final String containerName = "registry-" + UUID.randomUUID();
+  public LocalRegistry(int port, boolean pullBusyBox) {
```
coollog: Maybe `pullBusybox` can be removed and have it be a method instead? (so `Blob...IntegrationTest` will call that method)

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -17,77 +17,145 @@
...
-            "--restart=always",
-            "--name",
-            containerName,
-            "registry:2")
-        .run();
-
-    // Pulls 'busybox'.
-    new Command("docker", "pull", "busybox").run();
+    if (username == null || password == null) {
+      new Command(
```
coollog: Could prob consolidate this with L82 since it has mostly the same tokens.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -96,14 +88,11 @@ protected void before() throws IOException, InterruptedException {
...
-          .run();
-    }
-
-    if (pullBusyBox) {
-      pullAndPushToLocal("busybox", "busybox");
+              "REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd"));
     }
+    dockerTokens.add("registry:2");
+    String[] tokenArray = new String[dockerTokens.size()];
+    new Command(dockerTokens.toArray(tokenArray)).run();
```
coollog: I think we should just add a constructor that takes a List rather than doing this the round-about-array-way

Issue #774, opened 2018-08-01
Have registry credentials be provided via a list of CredentialRetrievers.


coollog:
Currently, `RetrieveRegistryCredentialsStep` handles registry credentials retrieval with logic internal to itself. Rather, the user (of the library) should be able to provide the actual strategies and order in which credentials are fetched.

Currently, in `ImageConfiguration`, credential-fetching methods are provided either via `setCredentialHelper`, which sets a potential credential helper to use, or via `setKnownRegistryCredentials`, which pass in credentials already fetched from places like Maven settings or the `auth` configuration. This should be coalesced into a single `setCredentialRetrievers` that takes a list of `CredentialRetriever`, where each could be implemented with, for example, fetching from credential helper, fetching from Docker config, provided from Maven settings, etc. The order would also determine the priority at which they are checked, and the evaluation of each provision strategy happens lazily as needed.

This is derived from the original comment at https://github.com/GoogleContainerTools/jib/issues/724#issuecomment-408450283

Tasks:

- [x] Add `CredentialRetriever` interface. #822 
- [x] Add setter for list of `CredentialRetriever`s to `BuildConfiguration` #823
- [x] Add factories for common provided `CredentialRetriever`s, including #824
  - [x] Retrieve from defined Docker credential helper #824
  - [x] Retrieve from inferred Docker credential helper #825
  - [x] Return predetermined known credentials #839
  - [x] Retrieve from Docker config #828 
- [x] Create list of `CredentialRetriever`s that are the same as what `RetrieveRegistryCredentialStep` does now and pass into `BuildConfiguration` #839
- [x] Use `CredentialRetriever`s in `RetrieveRegistryCredentialStep` #847
- [x] Remove `setCredentialHelper` and `setKnownRegistryCredentials` #851
- [x] Remove code duplication by creating a `DefaultCredentialRetrievers` in `plugins.common` #860
- [ ] Consider having the classes that are used in `CredentialRetrieverFactory` directly implement `CredentialRetriever`

PR #822, opened 2018-08-10
Adds interface for CredentialRetriever.

coollog:
Part of #774

- Adds `CredentialRetriever` that can be implemented to return `Credentials`.


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/credentials/Credentials.java
```diff
@@ -0,0 +1,56 @@
...
+  public String getPassword() {
+    return password;
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (other == null || getClass() != other.getClass()) {
```
TadCordle: I think we can use `instanceof` here and remove the null check?

PR #825, opened 2018-08-10
 Adds CredentialRetrieverFactory#inferCredentialHelper.

coollog:
Part of #774


jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildConfigurationTest.java
```diff
@@ -51,6 +53,8 @@ public void testBuilder() {
     String expectedTargetImageCredentialHelperName = "anotherCredentialHelper";
     RegistryCredentials expectedKnownTargetRegistryCredentials =
         Mockito.mock(RegistryCredentials.class);
+    List<CredentialRetriever> expectedCredentialRetrievers =
```
chanseokoh: You don't use this variable anywhere than `setCredentialRetrievers`, so the naming seems confusing. Probably just `credentialRetrievers`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -196,10 +199,13 @@ Authorization retrieveFromCredentialHelper(String credentialHelperSuffix)
     buildLogger.info("Checking credentials from docker-credential-" + credentialHelperSuffix);
 
     try {
-      Authorization authorization =
+      Credentials credentials =
           dockerCredentialHelperFactory
-              .newDockerCredentialHelper(registry, credentialHelperSuffix)
+              .newDockerCredentialHelper(
```
chanseokoh: You still seems to have the suffix version of `newDockerCredentialHelper`. Why using the `Path` version?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
```diff
@@ -0,0 +1,185 @@
...
+        throw new IllegalStateException("No COMMON_CREDENTIAL_HELPERS should be null");
+      }
+      inferredCredentialHelperSuffixes.add(inferredCredentialHelperSuffix);
+    }
+
+    return () -> {
+      for (String inferredCredentialHelperSuffix : inferredCredentialHelperSuffixes) {
+        try {
+          return retrieveFromDockerCredentialHelper(
+              registry,
```
chanseokoh: I think we can skip passing `registry`, unless you want to make `retrieveFromDockerCredentialHelper` `static`. It can get `registry` from `imageReference`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
```diff
@@ -0,0 +1,185 @@
...
+              Paths.get(
+                  DockerCredentialHelperFactory.CREDENTIAL_HELPER_PREFIX
```
chanseokoh: If you are always using `CREDENTIAL_HELPER_PREFIX`, maybe it's simpler to just pass a suffix and use the suffix version of `newDockerCredentialHelper`?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
```diff
@@ -0,0 +1,185 @@
...
+                      + inferredCredentialHelperSuffix),
+              dockerCredentialHelperFactory);
+
+        } catch (NonexistentDockerCredentialHelperException ex) {
```
chanseokoh: I think you should also catch `NonexistentServerUrlDockerCredentialHelperException`, so that you try the next helper in the inferred list if you can't get the credentials from the first one in the list.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
coollog: The `dockerCredentialHelper(Path credentialHelper)` would need to pass the full Path to `retrieveFromDockerCredentialHelper`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
coollog: I think we can leave that to another PR since it changes the behavior from the original `RetrieveRegistryCredentialsStep`. Filed #827

PR #835, opened 2018-08-13
Fix auth integration test on Mac

TadCordle:
~~Fixes a problem we were hitting with a failing volume mount on Mac.~~

PR #836, opened 2018-08-13
Fixes macOS auth integration test.

coollog:
From #835

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -73,7 +74,9 @@ protected void before() throws IOException, InterruptedException {
                   username,
                   password)
               .run();
-      Path tempFolder = Files.createTempDirectory("auth");
+      // Creates the temporary directory in /tmp since that is one of the default directories mounted into Docker.
```
loosebazooka: Were our tests flaky or broken?

coollog: Docker for Mac only mounts certain directories like `/tmp` into the VM and the default temporary directory was generated in `/var`, which isn't mounted into the VM and therefore the `htpasswd` did not exist for the integration test container to use (or rather, the folder could not be mounted to `/auth`).

PR #839, opened 2018-08-14
Makes list of credential retrievers.

coollog:
Part of #774 

There is a lot of duplicate code that should be refactored later (after `setKnownRegistryCredentials` and `setCredHelper` are removed).

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
@@ -109,7 +109,7 @@ public class BuildImageTask extends DefaultTask {
     }
     // Makes credential retriever list.
     List<CredentialRetriever> credentialRetrievers = new ArrayList<>();
-    String credentialHelperSuffix = jibExtension.getFrom().getCredHelper();
+    String credentialHelperSuffix = jibExtension.getTo().getCredHelper();
     if (credentialHelperSuffix != null) {
       credentialRetrievers.add(
           credentialRetrieverFactory.dockerCredentialHelper(credentialHelperSuffix));
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
@@ -19,7 +19,10 @@ package com.google.cloud.tools.jib.gradle;
 import com.google.cloud.tools.jib.cache.CacheDirectoryCreationException;
 import com.google.cloud.tools.jib.configuration.BuildConfiguration;
 import com.google.cloud.tools.jib.configuration.ImageConfiguration;
-import com.google.cloud.tools.jib.http.Authorization;
+import com.google.cloud.tools.jib.configuration.credentials.Credential;
+import com.google.cloud.tools.jib.configuration.credentials.CredentialRetriever;
+import com.google.cloud.tools.jib.frontend.CredentialRetrieverFactory;
+import com.google.cloud.tools.jib.http.Authorizations;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.InvalidImageReferenceException;
 import com.google.cloud.tools.jib.plugins.common.BuildStepsExecutionException;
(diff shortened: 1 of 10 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -81,20 +86,45 @@ public void buildImage() throws InvalidImageReferenceException {
...
+      knownTargetRegistryCredentials =
+          new RegistryCredentials(
+              "jib.to.auth",
+              Authorizations.withBasicCredentials(
+                  toCredential.getUsername(), toCredential.getPassword()));
+      knownCredentialRetriever = credentialRetrieverFactory.known(toCredential, "jib.to.auth");
     }
+    // Makes credential retriever list.
+    List<CredentialRetriever> credentialRetrievers = new ArrayList<>();
+    String credentialHelperSuffix = jibExtension.getFrom().getCredHelper();
```
TadCordle: This should be `jibExtension.getTo().getCredHelper()`, right? Maybe we should add another test that uses different credential helpers for `from` and `to`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -81,20 +86,45 @@ public void buildImage() throws InvalidImageReferenceException {
...
+    }
+    credentialRetrievers.add(credentialRetrieverFactory.inferCredentialHelper());
+    credentialRetrievers.add(credentialRetrieverFactory.dockerConfig());
+
     ImageConfiguration targetImageConfiguration =
-        ImageConfiguration.builder(ImageReference.parse(jibExtension.getTargetImage()))
+        ImageConfiguration.builder(targetImage)
             .setCredentialHelper(jibExtension.getTo().getCredHelper())
             .setKnownRegistryCredentials(knownTargetRegistryCredentials)
+            .setCredentialRetrievers(credentialRetrievers)
```
TadCordle: Is there any reason anymore to set all 3 of these, or do we just need `setCredentialRetrievers()`? Not necessarily in this PR, just in general.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -83,19 +88,55 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                Authorizations.withBasicCredentials(
+                    toCredential.getUsername(), toCredential.getPassword()));
+        knownCredentialRetriever =
+            credentialRetrieverFactory.known(
+                toCredential, MavenSettingsServerCredentials.CREDENTIAL_SOURCE);
+      }
+    } else {
+      knownTargetRegistryCredentials =
+          new RegistryCredentials(
+              "jib-maven-plugin <from><auth> configuration",
```
TadCordle: "jib-maven-plugin \<to\>\<auth\> configuration"

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -83,19 +88,55 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+              Authorizations.withBasicCredentials(
+                  toCredential.getUsername(), toCredential.getPassword()));
+      knownCredentialRetriever =
+          credentialRetrieverFactory.known(
+              toCredential, "jib-maven-plugin <from><auth> configuration");
```
TadCordle: Same thing here, should be "\<to\>\<auth\>"

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
coollog: Yep, it will eventually be just `setCredentialRetrievers`. I'm just doing these changes piece by piece so we have the behavior does not change at each step.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
coollog: This is what happens when I copy-paste too much code 🤦‍♂️

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
coollog: I'll be filing a followup PR to refactor this code / remove the duplication once the `setKnownRegistryCredentials` and `setCredHelper` are removed. It will also make this list-making more testable.

PR #843, opened 2018-08-15
Change missing build output directory message

### LINKED ISSUE #677: Kotlin build gives warning of no Java classes
**Description of the issue**:
If a gradle project only has Kotlin source files the Jib build will complain that 
```
warning: Could not find build output directory '/build/classes/java/main'
```

**Expected behavior**:
If Kotlin classes are found then there shouldn't be a warning of no java classes found

**Steps to reproduce**: 
1. Create a gradle project with kotlin source code
2. Build the project
3. run ```gradle jib```

**Environment**: Mac OS X, gradle 4.9, java 1.8

TadCordle:
Fixes #677 

Changes it to log at `info` level instead of `warn`, and lists all directories as well as which ones are skipped so it is more clear what is happening.


coollog:
I think if we were to remove the warning, we wouldn't need to add these info logs since we already have logs in `BuildStepsRunner` to log all the files that go into the container image.

coollog:
I think the warning should either be made clearer in that it doesn't matter if it doesn't need `java/main` source set, or to actually detect non-java projects and not give warnings for those. I think we should still keep the warning for pure-java projects though.

TadCordle:
@coollog the reason I took out the warning for specific source sets missing is because we still have the warning for all of them being empty, so I wasn't sure how useful it would be for the original warning to exist in the first place. Is there any reason only some output directories would be added for a pure java project?

coollog:
@TadCordle I probably need to check this further to confirm, but perhaps they have some Groovy+Java code and somehow did not attach the Java sourceset (so the Groovy code cannot actually find the Java code to use). We would probably need to do some more investigation into common pitfalls of projects with multiple languages to provide the best error/warning messages.

TadCordle:
@briandealwis we already do warn in that case - "No classes files were found - did you compile your project?"

To me it seems like an obscure enough problem that using --info to debug further would suffice, rather than showing a warning for every non-java project (or doing the extra work to detect this case and skip the warning).

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurations.java
```diff
@@ -61,13 +61,13 @@ static JavaLayerConfigurations getForProject(
 
     // Adds each file in each classes output directory to the classes files list.
     FileCollection classesOutputDirectories = mainSourceSet.getOutput().getClassesDirs();
+    gradleJibLogger.info("Adding corresponding output directories of source sets to image.");
```
briandealwis: nit: we don't use a trailing period elsewhere

PR #845, opened 2018-08-16
Move plugins.common package to new module

### LINKED ISSUE #478: Move plugins specific code to separate module.

TadCordle:
Fixes #478


TadCordle:
Oh oops, I branched off my other change instead of master.

jib-maven-plugin/pom.xml
```diff
@@ -293,6 +293,60 @@
               </resources>
             </configuration>
           </execution>
+
+          <!-- Imports jib-plugins-common sources -->
```
coollog: I think we could combine this with the jib-core source adding so that it just imports both sources (and the comment then becomes like `Imports jib-core and jib-plugins-common sources`

jib-plugins-common/gradle.properties
```diff
@@ -0,0 +1 @@
+version = 0.1.0-SNAPSHOT
```
coollog: tiny nit: Looks like we don't have ending newlines for the gradle.properties but probably should just add them

jib-plugins-common/settings.gradle
```diff
@@ -0,0 +1,2 @@
+rootProject.name = 'jib-plugins-common'
+
```
coollog: extra newline

TadCordle: Wait so do we want newlines or not?

Issue #832, opened 2018-08-10
Consider renaming NonexistentDockerCredentialHelperException to CannotFindDockerCredentialHelperException.


coollog:
From https://github.com/GoogleContainerTools/jib/pull/824/files#r209374222

chanseokoh:
I vote for `DockerCredentialHelperNotFoundException`, in line with `FileNotFoundException`, but whatever.

PR #847, opened 2018-08-16
 Changes RetrieveRegistryCredentialsStep to use credential retrievers list.

coollog:
Part of #774


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -21,42 +21,28 @@
...
-class RetrieveRegistryCredentialsStep implements AsyncStep<Authorization>, Callable<Authorization> {
+class RetrieveRegistryCredentialsStep implements AsyncStep<Credential>, Callable<Credential> {
 
   private static final String DESCRIPTION = "Retrieving registry credentials for %s";
 
-  /**
-   * Defines common credential helpers to use as defaults. Maps from registry suffix to credential
-   * helper suffix.
-   */
-  private static final ImmutableMap<String, String> COMMON_CREDENTIAL_HELPERS =
```
TadCordle: Maybe I'm just missing it, do we not look for these anymore?

coollog: These (and most of the code originally here) were moved to `CredentialRetrieverFactory`, where the default `CredentialRetriever`s are constructed.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
```diff
@@ -169,6 +170,9 @@ CredentialRetriever inferCredentialHelper(
               logger.info("  Caused by: " + ex.getCause().getMessage());
             }
           }
+
+        } catch (NonexistentServerUrlDockerCredentialHelperException | IOException ex) {
```
briandealwis: Should `NonexistentServerUrlDockerCredentialHelperException` and `NonexistentDockerCredentialHelperException ` be subclasses of `CredentialRetrievalException`?

coollog: That will potentially be done as part of #832

PR #860, opened 2018-08-20
Adds DefaultCredentialRetrievers to consolidate generation of CredentialRetrievers list.

coollog:
Part of #774 

This moves all of the duplicate logic in the plugins into `plugins-common`.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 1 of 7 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Creates a new {@link DefaultCredentialRetrievers} with a given {@link
+   * CredentialRetrieverFactory}.
+   *
+   * @param credentialRetrieverFactory the {@link CredentialRetrieverFactory} to generate the {@link
+   *     CredentialRetriever}s
+   * @return a new {@link DefaultCredentialRetrievers}
+   */
+  public static DefaultCredentialRetrievers init(
```
TadCordle: Why not just use a public constructor?

coollog: I'm intending this class to be potentially used by users of Jib library (this may be moved into `jib-core`). For all public APIs, I'm trying to keep it so that there are no constructors exposed (ie. static factories over constructors).

Issue #637, opened 2018-07-17
Reorganize caching to be an interface to allow for different implementations.


coollog:
This will help:
- Allow for layer up-to-date checking to be different strategies
- Have all layers be independent so no cache data is lost during concurrent executions

The cache implementation will consist of a storage engine and a query system.

Tasks:
- [x] Add design proposal for new cache mechanism. #878
- [x] Create new cache interfaces under a new package `ncache` that will eventually replace the existing `cache` package. #869
- [x] Add default implementations of the new cache interfaces. #874, #896, #879, #989, #901 
  - [x] DefaultCacheWriteEntry #874, #896
  - [x] DefaultCacheReadEntry #874, #896
  - [x] DefaultCacheStorage #879, #898, #901 
- [x] Add a new `ncache.Cache` that provides the higher-level operations used by the builder steps. #975, #1022, #1031, #1037, #1046
- [x] Have the builder steps use the new `ncache.Cache` and compare performance with old `cache.Cache` (performance is basically the same after checking with 100,000+ files)
- [x] Change the builder steps to use the new `ncache.Cache`. #1047
- [x] Remove old `cache` package #1047
- [ ] Move `ncache` to `cache`. #1056
- [ ] Consider adding a cache pruner that deletes all cache entries without a selector.
- [ ] Make sure all issues in #922 are fixed.
- [ ] Explore detections of potential cache corruptions - see #980

coollog:
Closing this in favor of the followup/cleanup tasks: #1056 #922 and #980

PR #869, opened 2018-08-22
Create new cache interfaces under a new package ncache.

coollog:
Part of #637 

The new `ncache` package will eventually replace `cache` (once the components are built out).

This establishes the interface to use for the new storage engine. The parts are separated into 3 components:

- `CacheStorage` - interface to read/write to the underlying storage
- `CacheReadEntry` - result of entries fetched by querying the storage engine
- `CacheWriteEntry` - data to give to the storage engine to write

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheReadEntry.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheReadEntry.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.ncache;
+
+import com.google.cloud.tools.jib.blob.Blob;
+import com.google.cloud.tools.jib.image.DescriptorDigest;
+import java.util.Optional;
+
+/**
+ * Represents a cache entry for a layer stored in the cache. <b>Implementations must be
+ * immutable.</b>
+ */
+public interface CacheReadEntry {
+
+  /**
+   * Gets the digest of the layer.
+   *
+   * @return the layer digest
+   */
+  DescriptorDigest getLayerDigest();
+
+  /**
+   * Gets the diff ID of the layer. The diff ID is the digest of the uncompressed layer contents,
+   * whereas the {@link #getLayerDigest} is the digest of the compressed layer contents.
+   *
+   * @return the layer diff ID
+   */
+  DescriptorDigest getLayerDiffId();
+
+  /**
+   * Gets the size of the layer, in bytes.
+   *
+   * @return the layer size
+   */
+  long getLayerSize();
+
+  /**
+   * Gets the {@link Blob} for the layer. This {@link Blob} should be able to be used multiple
+   * times.
+   *
+   * @return the layer {@link Blob}
+   */
+  Blob getLayerBlob();
+
+  /**
+   * Gets the optional metadata blob for the layer. The metadata is an arbitrary format chosen by
+   * the storage engine implementation.
+   *
+   * @return the metadata {@link Blob}
+   */
+  Optional<Blob> getMetadataBlob();
+}
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheReadEntry.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheReadEntry.java
@@ -57,8 +57,9 @@ public interface CacheReadEntry {
   Blob getLayerBlob();
 
   /**
-   * Gets the optional metadata blob for the layer. The metadata is an arbitrary format chosen by
-   * the storage engine implementation.
+   * Gets the optional metadata blob for the layer. The metadata is in the same format as supplied
+   * when writing to the cache with {@link CacheWriteEntry}. This {@link Blob} should be able to be
+   * used multiple times.
    *
    * @return the metadata {@link Blob}
    */
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheStorage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheStorage.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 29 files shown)


TadCordle:
To clarify, everything in the current `cache` package will be removed and this new `ncache` package will be called `cache` again when the redesign is done? (I ask because `ncache` is a weird name)

coollog:
@TadCordle Yep!

coollog:
@loosebazooka Yea, this is mostly for us to organize our own storage engine interface and reduce the tight coupling of the cache-checking strategies from the actual storage engine implementation.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheReadEntry.java
```diff
@@ -0,0 +1,66 @@
...
+   */
+  DescriptorDigest getLayerDigest();
+
+  /**
+   * Gets the diff ID of the layer. The diff ID is the digest of the uncompressed layer contents,
+   * whereas the {@link #getLayerDigest} is the digest of the compressed layer contents.
+   *
+   * @return the layer diff ID
+   */
+  DescriptorDigest getLayerDiffId();
```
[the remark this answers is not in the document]
coollog: Following Docker's naming convention: https://github.com/opencontainers/image-spec/blob/master/config.md#layer-diffid

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Gets the {@link Blob} to write as the layer contents.
+   *
+   * @return the layer {@link Blob}
+   */
+  Blob getLayerBlob();
+
+  /**
+   * Gets the optional selector digest to also reference this layer data.
```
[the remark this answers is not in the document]
coollog: It is a separate hash to use to reference a layer. We'll have it be a hash of the layer entries in our storage engine implementation.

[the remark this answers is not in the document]
loosebazooka: Should it be optional? Are we implying that CacheStorage will have a default implementation to calculate this? Who would and wouldn't return null here?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
```diff
@@ -0,0 +1,46 @@
...
+   * @return the selector digest
+   */
+  Optional<DescriptorDigest> getSelector();
+
+  /**
+   * Gets the optional {@link Blob} to write as the arbitrary layer metadata.
+   *
+   * @return the metadata {@link Blob}
+   */
+  Optional<Blob> getMetadataBlob();
```
loosebazooka: Same question as "selector" as to who would and wouldn't provide a value for this.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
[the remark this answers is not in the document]
coollog: The selector digest will be calculated at a higher level above the `CacheStorage`. Application layers will have selector (since they have layer entries), whereas base image layers will not.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
coollog: Same as selector - application layers will have metadata, base image layers will not.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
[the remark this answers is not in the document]
loosebazooka: Okay, so this is a secondary identifier that can be generated by a cache implementer? Can we add a little more detail to the javadoc?

```
Get an optional selector digest to also reference this layer data. A selector digest may be a secondary identifier for a layer that is distinct from the default layer digest. Useful as an inexpensive alternative reference to a layer rather than going through the motions of calculating the primary layer digest (sha256 of compressed tar)
```

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
loosebazooka: also here, perhaps some more detail.

```
.... to write as the arbitrary layer metadata (for example: last modified time, file hashes, etc)
```

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+package com.google.cloud.tools.jib.ncache;
+
+import com.google.cloud.tools.jib.blob.Blob;
+import com.google.cloud.tools.jib.image.DescriptorDigest;
+import java.util.Optional;
+
+/** Represents layer data to write to the cache. <b>Implementations must be immutable.</b> */
+public interface CacheWriteEntry {
```
briandealwis: Having a write entry feels weird.  How would I create an instance of this?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheReadEntry.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.google.cloud.tools.jib.blob.Blob;
+import com.google.cloud.tools.jib.image.DescriptorDigest;
+import java.util.Optional;
+
+/**
+ * Represents a cache entry for a layer stored in the cache. <b>Implementations must be
+ * immutable.</b>
+ */
+public interface CacheReadEntry {
```
briandealwis: How about `CachedLayer`?

The design doc talks of caching layers and metadata, and so this type should seem to do both.  But it seems specific to layers.  And what's an entry?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheStorage.java
```diff
@@ -0,0 +1,72 @@
...
+public interface CacheStorage {
+
+  /**
+   * Saves the {@link CacheWriteEntry}.
+   *
+   * @param cacheWriteEntry the {@link CacheWriteEntry}
+   * @return the {@link CacheReadEntry} for the written {@link CacheWriteEntry}
+   * @throws IOException if an I/O exception occurs
+   */
+  CacheReadEntry save(CacheWriteEntry cacheWriteEntry) throws IOException;
```
briandealwis: I don't know how to create a CacheWriteEntry.  Why not have this method instead take the required pieces:
```
CacheReadEntry saveLayer(Blob layerBlob, @Nullable Blob metadataBlob, @Nullable DescriptorDigest selector);
```

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheReadEntry.java
coollog: The original cache defined `CachedLayer`, which implemented the `Layer` interface. However, the coupling with the `Layer` interface meant that there was a lot of extra logic and code to support `CachedLayer` in places where `Layer` would be used. Rather, I think with the new cache interfaces, it'd be better to keep the cache entry decoupled from the `Layer` interface. A `Layer` instance can be constructed from a cache entry, but instead of having to use the subclass `CachedLayer` in many places like in the old cache, we can just use `Layer` everywhere. **I think maybe `CacheEntry` could work as a name?** (removing the `Read`)

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWriteEntry.java
coollog: Via static factories that builds the `CacheWriteEntry` for specific use cases (https://github.com/GoogleContainerTools/jib/pull/874/files#diff-d25f928c70b6196718d3234abbdbf5e3R33)

PR #871, opened 2018-08-22
Update READMEs for v0.9.10

TadCordle:
For merging after 0.9.10 is released. Fixes #868.

PR #873, opened 2018-08-22
Add support for defining an entrypoint

### LINKED ISSUE #579: Provide a way to use a custom entrypoint which is not a java command
**Description of the issue**:

Current jib is too opinionated on how to run a java command, even for spring boot apps it is limiting.
Allowing to run a script would be very beneficial for the whole ecosystem.

**Expected behavior**:

Entrypoint is configurable and can take any command.
It would be neat to synchronize src/main/jib/scripts folder into the image (same layer than the build classes?) to make it smooth and easy to maintain for teams.

### LINKED ISSUE #850: Allow setting an entry point to an explicit pathed binary
**Description of the issue**:
I'm trying to use an image where the entry point is not the java binary, and the java binary is in a different path than the default distroless image.  (Specifically, adoptopenjdk/openjdk10)

**Expected behavior**:
I should be able to specify the image's java execution path

briandealwis:
Fixes #579
Fixes #850

Adds support for providing an entrypoint using `<container><entrypoint>` (Maven) / `container.entrypoint` (Gradle).  `entrypoint` causes `jvmFlags` and `mainClass` to be ignored, and a warning is issued if they are found.
```
[WARNING] <mainClass> and <jvmFlags> are ignored when entrypoint is specified
```

Maven supports both using a list or just a string:
```xml
 <container>
  <entrypoint>/busybox/sh</entrypoint>
</container>
```

And it turns out that Maven doesn't actually check the names of child elements of a list.  So you can use something like the following if you like:
```xml
 <container>
  <entrypoint>
    <executable>/busybox/sh</executable>
    <argument>-x</argument>
    <scriptFile>/app/start.sh</scriptFile>
  </entrypoint>
</container>
```

(_PR: Not Ready_ while I get the ITs working.)


loosebazooka:
Is there a way we could provide a shortcut to the jib specific java classpath through some kind of template or keyword?
```
<entrypoint>
  <java>/weird/path/to/jre/on/my/corp/image/java</java>
  <cp>-cp</cp>
  <classpath>__jib_classpath__</classpath>
  ...
</entrypoint>
```

briandealwis:
I wonder if we can expose the class path as a property — so that it can be referenced but also changed.

coollog:
@briandealwis I believe that can be done by calling [`MavenProject#getProperties`](https://maven.apache.org/ref/3.2.3/apidocs/org/apache/maven/project/MavenProject.html#getProperties()) and then `setProperty`. The main thing though is that the pom.xml would not be able to resolve this property before some Jib goal actually sets it. I believe it may need to be done in the [`initialize` phase](http://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference) ("initialize build state, e.g. set properties or create directories."). Haven't tried this though.

coollog:
@briandealwis 

The late-binding construct is pretty neat!

For Gradle, yea, a variable is sufficient - it could be part of the `jib` extension (`jib.classpath`).

I think we wouldn't need to give a special construct for just setting the classpath. The current workaround is to just place other jars in `src/main/jib/app/libs/`.

briandealwis:
I've opened #894 for exposing the classpath and other possibly other variables.

jib-maven-plugin/README.md
```diff
@@ -285,6 +286,25 @@ In this configuration, the image:
...
+```xml
+<configuration>
+  <from>
+    <image>gcr.io/distroless/java:debug</image>
+  </from>
+  <container>
+    <!-- Maven doesn't check the names of child tags --> 
+    <entrypoint>
+      <executable>/busybox/sh</executable>
+      <script>/bin/start.sh</script>
```
chanseokoh: I couldn't understand what it means by "Maven doesn't check the names of child tags" until I saw examples using `<arg>` below. This also gives me the impression that `<executable>` and `<script>` are fixed names that I must use. TBH, I think this is confusing. It also makes me to believe that the order of `<executable>` and `<script>` shouldn't matter. I'd just go with using a single tag name.

coollog:
Is this PR ready for review? (The Not Ready is present)

briandealwis:
Oops, let me fix up @chanseokoh's findings and I'll remove the label.

briandealwis:
PTAL

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -149,6 +150,18 @@ public JavaDockerContextGenerator setBaseImage(String baseImage) {
     return this;
   }
 
+  /**
+   * Sets the entrypoint to be used as the {@code ENTRYPOINT}. If not empty, then overrides the
+   * {@link #setJvmFlags(List) jvmFlags} and {@link #setMainClass(String) mainclass}.
```
coollog: I think if we have this method, we should just remove the `setJvmFlags` and `setMainClass` methods and delegate the call of `JavaEntrypointConstructor#makeEntrypoint` to the user of `JavaDockerContextGenerator`.

jib-gradle-plugin/README.md
```diff
@@ -197,6 +197,7 @@ Property | Type | Default | Description
 `ports` | `List<String>` | *None* | Ports that the container exposes at runtime (similar to Docker's [EXPOSE](https://docs.docker.com/engine/reference/builder/#expose) instruction).
 `format` | `String` | `Docker` | Use `OCI` to build an [OCI container image](https://www.opencontainers.org/).
 `useCurrentTimestamp` | `boolean` | `false` | By default, Jib wipes all timestamps to guarantee reproducibility. If this parameter is set to `true`, Jib will set the image's creation timestamp to the time of the build, which sacrifices reproducibility for easily being able to tell when your image was created.
+`entrypoint` | list | *None* | The executable to be run (similar to Docker's [ENTRYPOINT](https://docs.docker.com/engine/reference/builder/#entrypoint) instruction). If set then `jvmFlags` and `mainClass` are ignored.
```
coollog: We should leave changes to documentation to another PR (like #871) after the feature is released as a new version of Jib, so users do not see an unreleased feature documented.

TadCordle: In fact, if this is going in 0.9.10, feel free to push to that branch so we can merge all the documentation changes in at once after releasing.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
briandealwis: Makes sense.

jib-gradle-plugin/README.md
briandealwis: will back out the docs.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
briandealwis: There's not much specific to Java in `JavaDockerContextGenerator` now!

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/DockerContextMojo.java
```diff
@@ -70,8 +70,7 @@ public void execute() throws MojoExecutionException {
 
       new JavaDockerContextGenerator(mavenProjectProperties.getJavaLayerConfigurations())
           .setBaseImage(getBaseImage())
-          .setJvmFlags(getJvmFlags())
-          .setMainClass(mainClass)
+          .setEntrypoint(getEntrypoint())
```
coollog: I think this is missing the logic to handle JVM flags and main class? We might want to add some tests to make sure JVM flags and mainClass still work.

jib-gradle-plugin/src/integration-test/resources/projects/default-target/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
 
 jib {
     container {
+        entrypoint = ['foo', 'bar', 'baz']
```
TadCordle: These cause the integration tests to fail, since the integration tests actually try to run the built container and 'foo' isn't an executable. For example, for `testBuildTar_simple`:

```
java.lang.RuntimeException: Command 'docker run simpleimage:gradle3455722031277561' failed: docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"foo\": executable file not found in $PATH": unknown.
time="2018-08-29T14:41:10-04:00" level=error msg="error waiting for container: context canceled"
```

TadCordle: To go off of what @coollog said, maybe it'd be a good idea to put this in a separate test that doesn't actually run the image and just inspects it, but actually runs the images generated using `mainClass` and `jvmFlags`.

briandealwis:
Sorry :-(   Had someone drop by and pushed — ITs don't run well locally.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/EntrypointIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+  @Test
+  public void testExecute() throws VerificationException, IOException {
+    String targetImage = "entrypoint:maven" + System.nanoTime();
+
+    Verifier verifier = new Verifier(entrypointTestProject.getProjectRoot().toString());
+    verifier.setSystemProperty("_TARGET_IMAGE", targetImage);
+    verifier.setAutoclean(false);
+    verifier.executeGoal("package");
+
+    verifier.executeGoal("jib:" + BuildTarMojo.GOAL_NAME);
```
coollog: It looks like this should be part of `BuildTarMojoIntegrationTest`?

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
```diff
@@ -350,4 +362,58 @@ public void testDockerContext() throws IOException, InterruptedException {
...
+    Assert.assertNotNull(classesTask);
+    Assert.assertEquals(TaskOutcome.SUCCESS, classesTask.getOutcome());
+    Assert.assertNotNull(jibTask);
+    Assert.assertEquals(TaskOutcome.SUCCESS, jibTask.getOutcome());
+    Assert.assertThat(
+        buildResult.getOutput(), CoreMatchers.containsString("Created Docker context at "));
+  }
+
+  @Test
+  public void testBuildTar_entrypoint() throws IOException {
```
coollog: If we do choose only one of the tasks to test, I would lean towards `jib` (build to registry) since it is the main functionality of Jib.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/EntrypointIntegrationTest.java
coollog: Okay, let's add a javadoc describing what this class is testing/

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
coollog: Maybe this method should just be called `testCustomEntrypoint` then?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/EntrypointIntegrationTest.java
TadCordle: But we are using `jib:buildTar` to build a tar and then verifying that it has the correct entrypoint here, aren't we? I think I agree that it does make more sense to put this in `BuildTarMojoIntegrationTest`. As far as I understand, the test file should correspond directly to a source file (and we don't have an Entrypoint class in jib-maven-plugin).

briandealwis:
I've reverted the integration tests and instead added tests around `PluginConfigurationProcessor` instead.  Much simpler.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessorTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class PluginConfigurationProcessorTest {
```
coollog: nit: short javadoc `Tests for {@link PluginConfigurationProcessor}.`

coollog: tiny nit: newline after class definition

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessorTest.java
```diff
@@ -0,0 +1,115 @@
...
+  @Mock GradleJibLogger gradleJibLogger;
```
coollog: nit: in this codebase, we prefer to name mocks with a mock- prefix, so like `mockGradleJibLogger`.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessorTest.java
```diff
@@ -0,0 +1,115 @@
...
+    Mockito.doReturn(Collections.emptyList()).when(containerParameters).getEntrypoint();
+
+    Mockito.doReturn(JavaLayerConfigurations.builder().build())
+        .when(projectProperties)
+        .getJavaLayerConfigurations();
+    Mockito.doReturn("java.lang.Object").when(projectProperties).getMainClass(jibExtension);
+  }
+
+  @Test
+  public void testBase() throws InvalidImageReferenceException {
```
coollog: Generally, we try to name the test methods with `test<method/behavior being tested>[_<thing this test is testing for>]`, so for example, this would be named `testProcessCommonConfiguration_default`.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessorTest.java
```diff
@@ -0,0 +1,115 @@
...
+    PluginConfigurationProcessor processor =
+        PluginConfigurationProcessor.processCommonConfiguration(
+            gradleJibLogger, jibExtension, projectProperties);
+    ContainerConfiguration configuration = processor.getContainerConfigurationBuilder().build();
```
coollog: We should prob add a TODO for testing the other processing results too.

.gitignore
```diff
@@ -1,3 +1,4 @@
+jib-*/bin
```
coollog: What is this referring to?

briandealwis: Oh I wasn't supposed to commit that.  I have Eclipse set to generate to bin/, which prevents clashes when running Maven and Gradle from outside.

Issue #534, opened 2018-07-10
'MANIFEST_INVALID' when uploading to Artifactory


Joeskyyy:
**Description of the issue**: When running jib as part of a maven goal, I get the following:

```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.2:build (build-docker-image) on project jengel-service: Build image failed: Tried to push image manifest for $OUR_REGISTRY/sre/jengel-test:jib but failed because: manifest invalid (something went wrong) | If this is a bug, please file an issue at https://github.com/GoogleContainerTools/jib/issues/new: 400 Bad Request
[ERROR] {"errors":[{"code":"MANIFEST_INVALID","message":"manifest invalid","detail":{"description":"null"}}]}
```

When looking at the artifactory logs, it seems like the manifest being uploaded is `null`:

```
2018-07-10 19:55:25,650 [http-nio-8081-exec-55] [INFO ] (o.a.e.UploadServiceImpl:360) - Deploy to 'et-docker-local:sre/jengel-test/_uploads/49928209-6f48-451f-8bd3-57a329be6880' Content-Length: unspecified
2018-07-10 19:55:25,748 [http-nio-8081-exec-66] [INFO ] (o.j.r.d.v.r.h.DockerV2LocalRepoHandler:257) - Deploying docker manifest for repo 'sre/jengel-test' and tag 'jib' into repo 'et-docker-local'
2018-07-10 19:55:25,761 [http-nio-8081-exec-66] [ERROR] (o.j.r.d.v.r.h.DockerV2LocalRepoHandler:298) - Error uploading manifest: 'null'
```

**Expected behavior**: The image should be uploaded to artifactory :D

**Steps to reproduce**:

* Add an artifactory repo as your `<to><image>` block
* Execute `mvn` to publish image
* Receive error ):

**Environment**:

* Current trying on macOS 10.13.5
* mvn 3.5.3
* jib 0.9.2
* Artifactory 6.0.1

**`jib-maven-plugin` Configuration**: 
```xml
                    <plugin>
                        <groupId>com.google.cloud.tools</groupId>
                        <artifactId>jib-maven-plugin</artifactId>
                        <version>0.9.2</version>
                        <configuration>
                            <from>
                                <image>$OUR_REGISTRY/frolvlad/alpine-oraclejdk8:slim</image>
                            </from>
                            <to>
                                <image> $OUR_REGISTRY/sre/jengel-test:jib</image>
[...]

Issue #875, opened 2018-08-23
write history as part of layer meta data.


loosebazooka:
History might be necessary for some tools to work.

- [x] add history fields to manifest object
- [x] passthrough original history
- [x] write jib layer history
- [x] allow custom "created_by" field

expected format: 

```yaml
{
  "author": "Jib",
  "created": "1970-01-01T00:00:00Z", # or useCurrentTimeStamp
  "created_by": "jib" # or jib-gradle-plugin jib-maven-plugin (tool identifier)
}
```

mattmoor:
```
$ crane config gcr.io/distroless/java/jetty@sha256:4bec71e275183f6306ea1bfe490dcf54732fcc7e37c137129ad30ed9dc62b99d | python -mjson.tool
{
    "architecture": "amd64",
    "author": "Bazel",
    "config": {
        "Cmd": [
            "/jetty/start.jar"
        ],
        "Entrypoint": [
            "/usr/bin/java",
            "-jar"
        ],
        "Env": [
            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/busybox",
            "SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt"
        ],
        "ExposedPorts": {
            "8080/tcp": {}
        },
        "WorkingDir": "/jetty"
    },
    "created": "1970-01-01T00:00:00Z",
    "history": [
        {
            "author": "Bazel",
            "created": "1970-01-01T00:00:00Z",
            "created_by": "bazel build ..."
        },
        {
            "author": "Bazel",
            "created": "1970-01-01T00:00:00Z",
            "created_by": "bazel build ..."
        },
        {
            "author": "Bazel",
            "created": "1970-01-01T00:00:00Z",
            "created_by": "bazel build ..."
        },
        {
            "author": "Bazel",
            "created": "1970-01-01T00:00:00Z",
            "created_by": "bazel build ..."
        },
        {
            "author": "Bazel",
            "created": "1970-01-01T00:00:00Z",
            "created_by": "bazel build ..."
        }
    ],
    "os": "linux",
    "rootfs": {
        "diff_ids": [
            "sha256:c09300da45505100cbbad5ae5cd8f80389ed86a162e60f6fa211828ff3f80268",
            "sha256:6f1db24e440882e47a86c6cbad091e98fda9e57cbbcb0ff8e34b46e6a435b20a",
[...]

PR #877, opened 2018-08-24
Write history as part of layer metadata

TadCordle:
Part of #875. Base image history is propagated, and history elements for each of the new layers are added on top. "created_by" is always just "jib" for now; I'll submit a followup later where we can pass in different strings.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -94,22 +96,40 @@
     try (Timer ignored = new Timer(buildConfiguration.getBuildLogger(), DESCRIPTION)) {
       // Constructs the image.
       Image.Builder<CachedLayer> imageBuilder = Image.builder();
-      for (PullAndCacheBaseImageLayerStep pullAndCacheBaseImageLayerStep :
-          NonBlockingSteps.get(pullAndCacheBaseImageLayersStep)) {
+      Image<Layer> baseImage = NonBlockingSteps.get(pullBaseImageStep).getBaseImage();
+      ContainerConfiguration containerConfiguration =
+          buildConfiguration.getContainerConfiguration();
+      String buildTime =
```
coollog: nit: `creationTime` may be less ambiguous - `buildTime` can be construed as time it took to build

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -94,22 +96,40 @@
...
       }
-
-      // Parameters that we passthrough from the base image
-      Image<Layer> baseImage = NonBlockingSteps.get(pullBaseImageStep).getBaseImage();
       imageBuilder.addEnvironment(baseImage.getEnvironment());
       imageBuilder.addLabels(baseImage.getLabels());
 
-      ContainerConfiguration containerConfiguration =
-          buildConfiguration.getContainerConfiguration();
+      // Add history elements if base layers/history sizes don't match
```
coollog: This can be used to add the application layer history objects too - just add the `buildAndCacheApplicationLayerSteps.size()`. This way, there's only one place that calls `new HistoryObjectTemplate(buildTime, "Jib", "jib")`

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -42,6 +42,18 @@
  *     "ExposedPorts": { "6000/tcp":{}, "8000/tcp":{}, "9000/tcp":{} }
  *     "Labels": { "com.example.label": "value" }
  *   },
+ *   "history": [
+ *     {
+ *       "author": "Jib",
+ *       "created": "1970-01-01T00:00:00Z",
+ *       "created_by": "jib build ..."
```
coollog: nit: `"created_by": "jib"`

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/BuildImageStepTest.java
```diff
@@ -139,5 +143,9 @@ public void test_propagateBaseImageConfiguration()
     Assert.assertEquals(
         ImmutableMap.of("base.label", "base.label.value", "my.label", "my.label.value"),
         image.getLabels());
+    Assert.assertTrue(
+        "History doesn't contain expected base image layer",
+        image.getHistory().contains(expectedHistory));
```
coollog: Assert the first history is the `expectedHistory`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryObjectTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+  private String created;
+
+  /** The name of the author specified when committing the image. */
+  @Nullable private String author;
+
+  /** The command used while building the image. */
+  @Nullable private String created_by;
+
+  /** Whether or not the layer is empty */
+  @Nullable private Boolean empty_layer;
```
coollog: `boolean`? I believe we can default to `false`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryObjectTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public HistoryObjectTemplate(
+      String created, @Nullable String author, @Nullable String createdBy) {
+    this.author = author;
+    this.created = created;
+    this.created_by = createdBy;
+  }
+
+  @Override
+  public boolean equals(Object other) {
```
coollog: nit: can short circuit with `if (this == other) return true;`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -94,22 +96,40 @@
...
+      if (baseImage.getHistory().size() < baseImageLayers.size()) {
+        int sizeDifference = baseImageLayers.size() - baseImage.getHistory().size();
+        for (int count = 0; count < sizeDifference; count++) {
```
coollog: History with `empty_layer=true` should not be taken into account when calculating the history objects to add (they don't correspond to a layer). This should also get a test to make sure this is accounted for correctly.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryObjectTemplate.java
TadCordle: I wanted to make this field optional, so I made it a `@Nullable Boolean` instead of `boolean` (if it's `boolean` then the field will always show up in the generated config).

coollog: Ah right, that makes sense :+1:

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryObjectTemplate.java
```diff
@@ -0,0 +1,66 @@
...
+
+package com.google.cloud.tools.jib.image.json;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.google.cloud.tools.jib.json.JsonTemplate;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HistoryObjectTemplate implements JsonTemplate {
```
coollog: javadoc

coollog: Also, was there a reason for having this outside of `ContainerConfigurationTemplate` (besides just a long file)? If we keep it outside, might want to name it like `ContainerConfigurationHistoryObjectTemplate` for something.

TadCordle: Wasn't sure if it was a good idea to use nested classes in this case since `HistoryObjectTemplate` is used in multiple places outside of `ContainerConfigurationTemplate`. But I could nest it if you think it's a better idea.

coollog: It's mostly that even though it is used outside of `ContainerConfigurationTemplate`, it is just a type of object inside the config JSON, so some sort of association there that ties them closely would probably make more sense for the class relationship.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -99,36 +99,43 @@
...
-        int sizeDifference = baseImageLayers.size() - baseImage.getHistory().size();
-        for (int count = 0; count < sizeDifference; count++) {
-          imageBuilder.addHistory(new HistoryObjectTemplate(buildTime, "Jib", "jib"));
-        }
+      // Add history elements for non-empty layers that don't have one yet
+      int sizeDifference =
+          baseImageLayers.size() + buildAndCacheApplicationLayerSteps.size() - nonEmptyLayerCount;
+      for (int count = 0; count < sizeDifference; count++) {
+        imageBuilder.addHistory(
+            new HistoryObjectTemplate(
```
coollog: Oh right, I just realized that this was passing a JSON template object into the `Image` builder. `Image` is intended to be a canonicalized Java object model of an image independent of the format (hence why we have `ImageToJsonTranslator`/`JsonToImageTranslator`). Making this distinction would allow for this image-building code not have to touch JSON templates at all.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -108,6 +110,74 @@
...
+     * Returns whether or not the history object corresponds to an empty layer.
+     *
+     * @return whether or not the history object corresponds to an empty layer.
+     */
+    @JsonIgnore
+    public boolean isEmptyLayer() {
+      return empty_layer == null ? false : empty_layer;
+    }
+
+    @VisibleForTesting
```
coollog: Is this used in testing?

TadCordle: Oh, not anymore.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
TadCordle: Is there a way to address this without making two nearly identical classes?

TadCordle: Hmm actually I guess they might not be that identical.

swilibc:
fwiw, this issue is causing problems for me with a private repo. The owner has confirmed that your fix addresses that issue.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -94,22 +96,47 @@
...
+      }
       imageBuilder.addEnvironment(baseImage.getEnvironment());
       imageBuilder.addLabels(baseImage.getLabels());
 
-      ContainerConfiguration containerConfiguration =
-          buildConfiguration.getContainerConfiguration();
+      // Add history elements for non-empty layers that don't have one yet
+      int sizeDifference =
+          baseImageLayers.size() + buildAndCacheApplicationLayerSteps.size() - nonEmptyLayerCount;
+      for (int count = 0; count < sizeDifference; count++) {
```
briandealwis: My only concern here is that we're claiming (and setting ourselves up for blame for) layers that were not generated by us.  Perhaps we could only add missing-history items here, and then add jib layers in the next `for()` where you're adding the Jib-generated layers (lines 136–139)?

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryItem.java
```diff
@@ -0,0 +1,85 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.google.cloud.tools.jib.json.JsonTemplate;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/** Represents an item in the container configuration's {@code history} list. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HistoryItem implements JsonTemplate {
```
briandealwis: Or `HistoryEntry`?  (Just throwing this out as a suggestion.)

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryItem.java
```diff
@@ -0,0 +1,85 @@
...
+  private String created;
+
+  /** The name of the author specified when committing the image. */
+  @Nullable private String author;
+
+  /** The command used while building the image. */
+  @Nullable private String created_by;
+
+  /** Whether or not the layer is empty ({@code @Nullable Boolean} to make field optional). */
+  @Nullable private Boolean empty_layer;
```
briandealwis: Question: should we follow [Google naming style](https://google.github.io/styleguide/javaguide.html#s5-naming) and use `@JsonProperty` for the external?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
TadCordle: Yeah, I originally had it this way and changed it so we only needed to use `new HistoryItem()` once, but on second thought I agree it's clearer the original way (easier to tell which history entries correspond to which layers).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -119,23 +119,19 @@
...
-                    : containerConfiguration.getCreationTime().toString(),
-                "Jib",
-                "jib",
-                null));
+      String layerCreationTime =
+          containerConfiguration == null
+              ? Instant.EPOCH.toString()
+              : containerConfiguration.getCreationTime().toString();
+      for (int count = 0; count < baseImageLayers.size() - nonEmptyLayerCount; count++) {
+        imageBuilder.addHistory(new HistoryEntry(layerCreationTime, "Jib", "jib", null));
```
briandealwis: But I don't think this should be "Jib"/"jib" — maybe author = _"(unknown)"_ and createdBy= _"(unavailable)"_?

TadCordle: I can change them, but right now createdBy is just a placeholder. After this I think I'm going to add a way to pass in something like "jib-gradle-plugin" or "jib-maven-plugin". Maybe I should do the same for author?

@coollog WDYT

coollog: I think Brian means for the base image layers that don't have history to have `author` and `createdBy` be `(unknown)`, though I think leaving them empty would be fine too.

TadCordle: Oh ok, makes sense.

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
```diff
@@ -0,0 +1,90 @@
...
+import com.google.cloud.tools.jib.json.JsonTemplate;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/** Represents an item in the container configuration's {@code history} list. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HistoryEntry implements JsonTemplate {
+
+  /** The timestamp at which the image was created. */
+  private String created;
```
coollog: Do we want to be making the `created` field non-nullable (in the config spec, it is optional)?

coollog: Can name this in code style and use `JsonProperty` here too. Something like `creationTimestamp` is more descriptive.

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
```diff
@@ -0,0 +1,90 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.cloud.tools.jib.json.JsonTemplate;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/** Represents an item in the container configuration's {@code history} list. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HistoryEntry implements JsonTemplate {
+
+  /** The timestamp at which the image was created. */
```
coollog: A comment about this needing to be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6)

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /** The name of the author specified when committing the image. */
+  @Nullable private String author;
+
+  /** The command used while building the image. */
+  @JsonProperty("created_by")
+  @Nullable
+  private String createdBy;
+
+  /** Whether or not the layer is empty ({@code @Nullable Boolean} to make field optional). */
```
coollog: This field is actually about whether or not this history entry corresponds to an actual layer in the container.

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
```diff
@@ -0,0 +1,90 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HistoryEntry implements JsonTemplate {
+
+  /** The timestamp at which the image was created. */
+  private String created;
+
+  /** The name of the author specified when committing the image. */
+  @Nullable private String author;
+
+  /** The command used while building the image. */
```
coollog: s/while building/to build/?

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
```diff
@@ -0,0 +1,90 @@
...
+package com.google.cloud.tools.jib.image;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.cloud.tools.jib.json.JsonTemplate;
+import java.util.Objects;
+import javax.annotation.Nullable;
+
+/** Represents an item in the container configuration's {@code history} list. */
```
coollog: Let's add a link to https://github.com/opencontainers/image-spec/blob/master/config.md#properties and pointing specifically to the `history` field.

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
```diff
@@ -0,0 +1,90 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HistoryEntry implements JsonTemplate {
```
coollog: I think this might better belong under the `image.json` package, next to `ContainerConfigurationTemplate`.

I'm thinking it might make more sense to call this `ContainerConfigurationHistory` and have it manage the actual JSON templates. Then, `Image` will have a `ContainerConfigurationHistory` and not have to directly manage `HistoryEntry` JSON templates.

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
```diff
@@ -0,0 +1,90 @@
...
+
+  /** The command used while building the image. */
+  @JsonProperty("created_by")
+  @Nullable
+  private String createdBy;
+
+  /** Whether or not the layer is empty ({@code @Nullable Boolean} to make field optional). */
+  @JsonProperty("empty_layer")
+  @Nullable
+  private Boolean emptyLayer;
```
coollog: Could be named like `isLayer`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
coollog: Hmm, interesting - okay, let's match that behavior then since we don't want to break compatibility with Docker.

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
TadCordle: I think that naming is too different from the field it corresponds to and could get confusing.

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
TadCordle: I'll move it to the package, but I'm just going to make an issue for the second point for now.

TadCordle: #889

jib-core/src/main/java/com/google/cloud/tools/jib/image/HistoryEntry.java
briandealwis: But on the flip side we may error out reading an OCI-conformant image that has no `created` field. Right?

briandealwis:
Might be worth including the `comment` field on `HistoryEntry` — we can fill that in for our auto-generated missing layer history entries.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -124,14 +124,15 @@
               ? Instant.EPOCH.toString()
               : containerConfiguration.getCreationTime().toString();
       for (int count = 0; count < baseImageLayers.size() - nonEmptyLayerCount; count++) {
-        imageBuilder.addHistory(new HistoryEntry(layerCreationTime, null, null, null));
+        imageBuilder.addHistory(
+            new HistoryEntry(layerCreationTime, null, null, "auto-generated by Jib", null));
```
coollog: We probably want to go with a builder pattern here since it's unclear which field is which.

coollog: Also, "auto-generated by Jib" might not be the text we want for `created_by`, since `created_by` refers to what created the layer the history entry refers to (and that layer was not "auto-generated by Jib") - I think "unknown" would be more suitable here.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * Represents an item in the container configuration's {@code history} list.
+ *
+ * @see <a href=https://github.com/opencontainers/image-spec/blob/master/config.md#properties>OCI
+ *     image spec ({@code history} field)</a>
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class HistoryEntry implements JsonTemplate {
+
+  /** The RFC 3339 formatted timestamp at which the image was created. */
```
coollog: nit: specifically section 5.6 of RFC 3339

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
```diff
@@ -0,0 +1,108 @@
...
+  @JsonProperty("created")
+  @Nullable
+  private String creationTimestamp;
+
+  /** The name of the author specified when committing the image. */
+  @JsonProperty("author")
+  @Nullable
+  private String author;
+
+  /** The command used to build the image. */
```
coollog: s/image/layer/

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
```diff
@@ -0,0 +1,108 @@
...
+  @Nullable
+  private String createdBy;
+
+  /** A custom message set when creating the layer. */
+  @JsonProperty("comment")
+  @Nullable
+  private String comment;
+
+  /**
+   * Whether or not the entry corresponds to an empty layer ({@code @Nullable Boolean} to make field
```
coollog: This comment should be updated too

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
```diff
@@ -0,0 +1,108 @@
...
+    this.emptyLayer = emptyLayer;
+  }
+
+  /**
+   * Returns whether or not the history object corresponds to a layer in the container.
+   *
+   * @return {@code true} if the history object corresponds to a layer in the container
+   */
+  @JsonIgnore
+  public boolean hasLayer() {
```
coollog: nit: `hasLayer` seems to mean if this `HistoryEntry` has a layer - perhaps `hasLayerInContainer` or `hasCorrespondingLayer` or `isForLayer`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -137,10 +152,18 @@ public void addLayerDiffId(DescriptorDigest diffId) {
     rootfs.diff_ids.add(diffId);
   }
 
+  public void addHistory(HistoryEntry historyEntry) {
```
coollog: nit: since `history` is a plural list of `HistoryEntry`s, perhaps this method name should be `addToHistory` instead

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -70,6 +82,9 @@
   /** Execution parameters that should be used as a base when running the container. */
   private final ConfigurationObjectTemplate config = new ConfigurationObjectTemplate();
 
+  /** Build commands used to create the image. */
```
coollog: "Describes the history of each layer"?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
TadCordle: That's `comment`, not `created_by`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
TadCordle: To what? I already updated this from earlier.

coollog: The same as the `hasLayer` method? "whether or not the history object corresponds to a layer in the container"

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
coollog: Ah right, then my first comment applies haha.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
```diff
@@ -32,7 +32,56 @@
...
-  /** The RFC 3339 formatted timestamp at which the image was created. */
+  public static class Builder {
+
+    @Nullable private String creationTimestamp;
+    @Nullable private String author;
+    @Nullable private String createdBy;
+    @Nullable private String comment;
+    @Nullable private Boolean emptyLayer;
+
+    public Builder setCreationTimestamp(String creationTimestamp) {
```
coollog: Would it make sense to take an `Instant` here instead?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
```diff
@@ -105,4 +154,17 @@ public boolean equals(Object other) {
   public int hashCode() {
     return Objects.hash(author, creationTimestamp, createdBy, comment, emptyLayer);
   }
+
+  @Override
+  public String toString() {
```
coollog: Is this used?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
```diff
@@ -32,7 +32,56 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class HistoryEntry implements JsonTemplate {
 
-  /** The RFC 3339 formatted timestamp at which the image was created. */
+  public static class Builder {
```
coollog: nit: short javadoc

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
TadCordle: Whoops, added that to test something and forgot to get rid of it.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/HistoryEntry.java
TadCordle: I didn't see any other Builders in the code base with a javadoc (and I don't think it's necessary, this is pretty self-explanatory).

coollog:
This should fix #534

PR #878, opened 2018-08-24
Proposal RFC: New cache mechanism

coollog:
Regards #637 

Preview at: https://github.com/GoogleContainerTools/jib/blob/ncache-proposal/proposals/cache_v2.md


coollog:
Updated according to discussion. Will merge #874 and file another PR according to the changed specs.

coollog:
@GoogleContainerTools/java-tools Added a section about locking since I realized that `FileLock` is not sufficient for concurrent executions of Jib within the same JVM.

proposals/cache_v2.md
```diff
@@ -0,0 +1,177 @@
...
+
+The layer directory should be written atomically to avoid reads that may read an incomplete layer directory.
+
+The selector file can be written atomically anytime after the layer directory is fully written.
+
+All atomic writes should be done by first writing to a temporary location and then moving to the desired location.
+
+#### List the entries in the cache
+
+Listing the entries in the cache just requires obtaining a list of all the layer directories under `layers/`. Since the layer directory may write to a temporary directory within the `layers/` directory, the listing would just need to filter out those temporary directories.
```
chanseokoh: Perhaps we may avoid this filtering if we create a temp directory or file in other locations like `/tmp`?

coollog: I believe on Windows moving a file from one directory to another is a costly operation (though I cannot find any sources for that currently). However, we could just go with creating the temp directory not in `layers/` and just absorb the cost on Windows for the cost we save during reads.

proposals/cache_v2.md
```diff
@@ -0,0 +1,177 @@
...
+  36a2b7401dcddc50a35aeaa81085718b9d5fbce9d607c55a1d79beec2469f9ac/
+    326a609681777ee4ca02b1898579c9e07801ef066a629a3c59fa6df6ab42b7aa.layer
+    metadata
+selectors/
+  65de3b72aaf98e4f300ccdf7d64bf9a3b1e23c8c44a1242265f717db1a0877e9
+```
+
+The `layers/` directory consists of directories for each layer, named by the layer digest.
+Inside each of the layer directories:
+- The layer tarball is named by the layer diff ID with extension `.layer`
```
TadCordle: Is this still the plan? Noticed you removed this extension here https://github.com/GoogleContainerTools/jib/pull/901/files#diff-68ffab573db32fda375d7fdebc437e37L55

coollog: Right, I'll update this with that change.

coollog:
Yea, this proposal can be merged.

PR #879, opened 2018-08-24
 Adds DefaultCacheStorage#write.

coollog:
Part of #637 

`DefaultCacheStorage` - storage engine implementation with just `write` implemented so far
`DefaultCacheStorageFiles` - resolves files used in the cache storage directory
`DefaultCacheStorageWriter` - the logic behind `DefaultCacheStorage#write`


jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorage.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param cacheDirectory the directory for this cache
+   * @return a new {@link CacheStorage}
+   */
+  public static CacheStorage withDirectory(Path cacheDirectory) {
+    return new DefaultCacheStorage(cacheDirectory);
+  }
+
+  private final DefaultCacheStorageWriter defaultCacheStorageWriter;
+
+  private DefaultCacheStorage(Path cacheDirectory) {
```
loosebazooka: Would we want to pass in defaultCacheStorageWriter so we can mock this out? Perhaps in `withDirectory`?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorage.java
```diff
@@ -0,0 +1,91 @@
...
+    DefaultCacheStorageFiles defaultCacheStorageFiles =
+        new DefaultCacheStorageFiles(cacheDirectory);
+    this.defaultCacheStorageWriter = new DefaultCacheStorageWriter(defaultCacheStorageFiles);
+  }
+
+  @Override
+  public CacheEntry write(CacheWrite cacheWrite) throws IOException {
```
loosebazooka: I don't know where CacheWrite came from? Perhaps an earlier PR? But it's naming is confusing. I would expect it to be "CacheLayerMetadata" or something

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   * Writes the {@link CacheWrite}.
+   *
+   * @param cacheWrite the {@link CacheWrite} to write out
+   * @return the {@link CacheEntry} representing the written entry
+   * @throws IOException if an I/O exception occurs
+   */
+  CacheEntry write(CacheWrite cacheWrite) throws IOException {
+    Path temporaryLayerFile = Files.createTempFile(null, null);
```
loosebazooka: I would expect this write to happen in the cache directory? I feel like mv is not free in Windows (but renames are?)

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorage.java
coollog: Hmm, it used be called `CacheWriteEntry` and `CacheEntry` used to be `CacheReadEntry`. `CacheWrite` is intended to be the interface for cache writes and `CacheLayerMetadata` might give a sense of being the layer metadata read out from the cache. Perhaps it should be called `CacheWriteData` instead?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorage.java
```diff
@@ -0,0 +1,91 @@
...
+ *   <layer digest>/
+ *     <layer diff ID>.layer
+ *     metadata
+ *   ...
+ * selectors/
+ *   <selector digest>
+ *   ...
+ * }</pre>
+ *
+ * Layers entires are stored in their own directories under the {@code layers/} directory. Each
```
TadCordle: ~~entires~~ entries

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageFiles.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  /**
+   * Resolves a selector file.
+   *
+   * @param selector the selector digest
+   * @return the selector file
+   */
+  Path getSelectorFile(DescriptorDigest selector) {
+    return getSelectorDirectory(selector);
```
TadCordle: Is `getSelectorDirectory` necessary, or can its logic just go here?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -0,0 +1,106 @@
...
+import com.google.cloud.tools.jib.image.DescriptorDigest;
+import java.io.BufferedOutputStream;
+import java.io.IOException;
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.zip.GZIPOutputStream;
+
+/** Writes to the default cache storage engine. */
+class DefaultCacheStorageWriter {
```
TadCordle: I think since this class only contains one write method that's used only in `DefaultCacheStorage`, we could move that logic there instead of adding this extra class. Although I'm not sure what the plan is for reusing this.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -0,0 +1,106 @@
...
+      // The GZIPOutputStream must be closed in order to write out the remaining compressed data.
+      compressorStream.close();
+      BlobDescriptor compressedBlobDescriptor = compressedDigestOutputStream.toBlobDescriptor();
+      DescriptorDigest layerDigest = compressedBlobDescriptor.getDigest();
+
+      // Renames the temporary layer file to the correct filename.
+      Path layerFile = defaultCacheStorageFiles.getLayerFile(layerDigest, layerDiffId);
+      Files.createDirectories(layerFile.getParent());
+      try {
+        Files.move(temporaryLayerFile, layerFile);
```
TadCordle: Could moving this file while we have an output stream open on it cause problems? Can we close the output stream earlier?

coollog:
Incorporated new proposals to account for atomicity of writes from https://github.com/GoogleContainerTools/jib/pull/878#issuecomment-419127182. This changes the write to only produce a layer directory at the intended location in full. This way, cache reads never see an incomplete layer directory.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/FileOperations.java
```diff
@@ -52,5 +57,27 @@ public static void copy(ImmutableList<Path> sourceFiles, Path destDir) throws IO
     }
   }
 
+  /**
+   * Acquires an exclusive lock on the {@code file} and opens an {@link OutputStream} to write to
```
loosebazooka: I think should make our intentions a little more clear here on why we're locking. They are an assertion rather than concurrency management feature?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/TemporaryDirectory.java
```diff
@@ -0,0 +1,70 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * A temporary directory that tries to delete itself upon close. Note that deletion is <b>NOT</b>
+ * guaranteed.
+ */
+public class TemporaryDirectory implements Closeable {
```
loosebazooka: guava `Files.createTempDir` and `MoreFiles.deleteRecursively`?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -0,0 +1,220 @@
...
+    private WrittenLayer(
+        DescriptorDigest layerDigest, DescriptorDigest layerDiffId, long layerSize) {
+      this.layerDigest = layerDigest;
+      this.layerDiffId = layerDiffId;
+      this.layerSize = layerSize;
+    }
+  }
+
+  /**
+   * Attempts to move {@code source} to {@code destination}. If {@code destination} already exists,
```
loosebazooka: we talked offline about this, but just making note that we need to overwrite selector files.

coollog:
Definitely. We'll do a side-by-side comparison with `cache` once `ncache` is working.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -0,0 +1,236 @@
...
+   * @throws IOException if an I/O exception occurs
+   */
+  private WrittenLayer writeLayerBlobToDirectory(Blob layerBlob, Path layerDirectory)
+      throws IOException {
+    Path temporaryLayerFile = Files.createTempFile(layerDirectory, null, null);
+    temporaryLayerFile.toFile().deleteOnExit();
+
+    try (CountingDigestOutputStream compressedDigestOutputStream =
+        new CountingDigestOutputStream(
+            new BufferedOutputStream(FileOperations.newLockingOutputStream(temporaryLayerFile)))) {
```
loosebazooka: Sorry I missed this, but it doesn't seem necessary to use locking outputstream on temporary files. Shouldn't their uniqueness be provided by guava?

coollog: Oh yes, that's true. Will change this in #898

PR #883, opened 2018-08-27
Only warn if inferred credential helper threw NonexistentServerUrlDockerCredentialHelperException

### LINKED ISSUE #827: Don't fail credential retrieval if inferred credential helper threw NonexistentServerUrlDockerCredentialHelperException.
From https://github.com/GoogleContainerTools/jib/pull/825#discussion_r209344161

TadCordle:
Fixes #827

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
@@ -171,17 +171,17 @@ public class CredentialRetrieverFactory {
                       + inferredCredentialHelperSuffix),
               dockerCredentialHelperFactory);
 
-        } catch (DockerCredentialHelperNotFoundException ex) {
+        } catch (DockerCredentialHelperNotFoundException
+            | NonexistentServerUrlDockerCredentialHelperException ex) {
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
```diff
@@ -171,17 +171,17 @@ CredentialRetriever inferCredentialHelper(
                       + inferredCredentialHelperSuffix),
               dockerCredentialHelperFactory);
 
-        } catch (DockerCredentialHelperNotFoundException ex) {
+        } catch (DockerCredentialHelperNotFoundException
+            | NonexistentServerUrlDockerCredentialHelperException ex) {
```
loosebazooka: How does this failure now affect the build? Just it just continue to some other auth mechanism?

TadCordle: Yeah, currently if the first cred helper exists, but doesn't have the correct server url, it'll just fail without checking other cred helpers on the system. So now it will treat cred helpers that are missing the correct url the same way we've been treating missing cred helpers.

PR #887, opened 2018-08-28
Enhance build.sh

briandealwis:
Enhance build.sh to allow running ITs locally…

  - add `it` target to run integration tests
  - make `-e` for gradle use `--info --stacktrace` to see test errors


build.sh
```diff
@@ -84,5 +85,10 @@ for target in "$@"; do
       fi
       ;;
 
+    it)
+      doBuild jib-core  ./gradlew $gradleOptions integrationTest
+      doBuild jib-maven-plugin   ./mvnw $mavenOptions integration-test -U
```
coollog: To run the Maven integration test, we actually have it be in a profile: `./mvnw $mavenOptions verify -Pintegration-tests`

PR #888, opened 2018-08-28
Revert "Adds entries for all directories in layer tarball. (#772)"

coollog:
This is breaking the integration tests. Going to revert it and then re-file a pull request when it's fixed.


coollog:
Unfixes #523

PR #891, opened 2018-08-28
Adds entries for all directories in layer tarball.

### LINKED ISSUE #523: /app not readable when running on Openshift

### LINKED ISSUE #727: Cannot run image running with container generated by jib as non-root in k8s

coollog:
Originally #772 , but reverted in #888 .

This one adds a fix for constructing the `TarArchiveEntry` extraction path correctly, whereas before it generated paths like `/app/classes//com/test...`, which caused errors like https://github.com/GoogleContainerTools/jib/issues/727#issuecomment-414448807. The main difference from #772 is in `buildAsTarArchiveEntries`.

Fixes #523 
Fixes #727


jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -34,31 +40,75 @@
  */
 public class ReproducibleLayerBuilder {
 
+  /**
+   * Holds a list of {@link TarArchiveEntry}s with unique extraction paths. The list also includes
+   * all parent directories for each extraction path.
+   */
+  private static class UniqueTarArchiveEntries {
```
TadCordle: Why not just use a `LinkedHashMap` or something instead of making a custom data structure for it?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -34,31 +40,75 @@
...
-                   */
-                  StringBuilder subExtractionPath =
-                      new StringBuilder(layerEntry.getExtractionPath());
+                  // Builds the same file path as in the source file for extraction. The iteration
+                  // is necessary because the path needs to be in Unix-style.
+                  Path subExtractionPath = Paths.get(layerEntry.getExtractionPath());
                   Path sourceFileRelativePath = sourceFile.getParent().relativize(path);
                   for (Path sourceFileRelativePathComponent : sourceFileRelativePath) {
-                    subExtractionPath.append('/').append(sourceFileRelativePathComponent);
+                    subExtractionPath = subExtractionPath.resolve(sourceFileRelativePathComponent);
```
TadCordle: Wasn't the point of using `append('/')` instead of `resolve` for forcing the paths to be Unix-style? (Comment on line 107)

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
coollog: The `LinkedHashMap` would require creating a custom wrapper that uses the `TarArchiveEntry#getName` as the hash object, which seemed more confusing than just having a custom data structure for storing a list of unique `TarArchiveEntry`s

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
coollog: Actually, that seems to be a relic of the old way - removing.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -34,41 +40,82 @@
...
-                      new TarArchiveEntry(path.toFile(), subExtractionPath.toString()));
+                      new TarArchiveEntry(path.toFile(), extractionPath.toString()));
                 });
 
       } else {
         TarArchiveEntry tarArchiveEntry =
             new TarArchiveEntry(
                 sourceFile.toFile(),
-                layerEntry.getExtractionPath() + "/" + sourceFile.getFileName());
+                Paths.get(layerEntry.getExtractionPath(), sourceFile.getFileName().toString())
```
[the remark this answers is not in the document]
coollog: Sounds good. I'll actually break this out into another method to provide more readability.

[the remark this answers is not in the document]
coollog: Actually, nevermind, it made it less readable.

PR #893, opened 2018-08-28
Adds micronaut example.

coollog:
Part of #540 
Copied over from https://github.com/coollog/micronaut-jib

Preview at https://github.com/GoogleContainerTools/jib/tree/micronaut-example/examples/micronaut


examples/micronaut/src/main/groovy/example/micronaut/HelloController.groovy
```diff
@@ -0,0 +1,14 @@
...
+import groovy.transform.CompileStatic
+import io.micronaut.http.annotation.Controller
+import io.micronaut.http.annotation.Get
+
+@CompileStatic
+@Controller("/hello") // <1>
+class HelloController {
+    @Get("/") // <2>
+    String index() {
+        "Hello Jib" // <3>
```
TadCordle: Everything on the README says the output is "Hello World", maybe this should match that.

coollog: Fixed

examples/micronaut/README.md
```diff
@@ -20,7 +20,7 @@ docker run -d -p 8080:8080 micronaut-jib:0.1
 ```
 ```shell
 curl localhost:8080/hello
-> Hello World
+> Hello Jib
```
TadCordle: Maybe we should change the code instead, the gif still says Hello World.

coollog: Ah, good point, I'll just change back to Hello World

examples/micronaut/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+        exclude group: 'org.codehaus.groovy', module: 'groovy-all'
+    }
+    testCompile 'io.micronaut:inject-java'
+}
+
+compileJava.options.compilerArgs += '-parameters'
+compileTestJava.options.compilerArgs += '-parameters'
+
+/** CHANGE THIS TO THE IMAGE YOU WANT TO PUSH TO */
+// jib.to.image = 'gcr.io/PROJECT_ID/micronaut-jib'
```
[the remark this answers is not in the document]
coollog: Yea this line is just there to give the user an example of how to configure `jib-gradle-plugin` if they want to build directly to a registry. The project version is 0.1 so you would need to use the `micronaut-jib:0.1` reference that Jib generates for `jibDockerBuild`.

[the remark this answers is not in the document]
chanseokoh: Ah, that was my copy-paste error. The only thing to fix is to add the `--image=micronaut-jib` parameter then.

https://github.com/GoogleContainerTools/jib/pull/893/files#diff-0a4703c3cc2535016cc80dff59feaa82R17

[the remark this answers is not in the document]
coollog: Yep, the reference is generated by `jibDockerBuild`

PR #897, opened 2018-08-29
Adds FAQ about using debug base image.

### LINKED ISSUE #354: It’d be nice to have a flag in the config to use the debug base
`I need to inspect this container but I can’t bash into it cause it’s distroless :(`

coollog:
Fixes #354 

Preview at https://github.com/GoogleContainerTools/jib/blob/where-is-bash/docs/faq.md


docs/faq.md
```diff
@@ -41,6 +42,40 @@ kubectl run jib-deployment --image=<image name>
...
+
+If you would like to include a shell for debugging, set the base image to `gcr.io/distroless/java:debug` instead. The shell will be located at `/busybox/sh`.
+
+<details>
+<summary>Base image configuration examples</summary>
+<p>
+
+#### Maven
+
+In [`jib-maven-plugin`](../jib-maven-plugin), you can then use this custom base image by adding the following configuration:
```
TadCordle: you can ~~then~~ use ~~this~~ **the** custom base image

docs/faq.md
```diff
@@ -41,6 +42,40 @@ kubectl run jib-deployment --image=<image name>
...
+<configuration>
+  <from>
+    <image>gcr.io/distroless/java:debug</image>
+  </from>
+</configuration>
+```
+
+#### Gradle
+
+In [`jib-gradle-plugin`](../jib-gradle-plugin), you can then use this custom base image by adding the following configuration:
```
TadCordle: you can ~~then~~ use ~~this~~ **the** custom base image

docs/faq.md
```diff
@@ -41,6 +42,40 @@ kubectl run jib-deployment --image=<image name>
...
+
+```groovy
+jib.from.image = 'gcr.io/distroless/java/debug'
```
TadCordle: Should this be `:debug`, not `/debug`?

docs/faq.md
```diff
@@ -41,6 +42,42 @@ kubectl run jib-deployment --image=<image name>
 
 For more information, see [steps 4-6 of the Kubernetes Engine deployment tutorial](https://cloud.google.com/kubernetes-engine/docs/tutorials/hello-app#step_4_create_a_container_cluster).
 
+#### Where is bash?
+
+By default, Jib uses [`distroless/java`](https://github.com/GoogleContainerTools/distroless/tree/master/java) as the base image. Visit the [distroless project](https://github.com/GoogleContainerTools/distroless#distroless-docker-images) for more information about distroless images.
```
TadCordle: I think it'd still be a good idea to mention that distroless doesn't have a shell here. ("By default, Jib uses `distroless/java` as the base image, which does not have a shell. Visit the...")

PR #898, opened 2018-08-29
 Adds DefaultCacheStorage#listDigests.

coollog:
Part of #637 

`DefaultCacheStorageReader` - the logic behind `DefaultCacheStorage#listDigests/retrieve/select`


loosebazooka:
Assuming this needs to be rebased?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageReader.java
```diff
@@ -0,0 +1,61 @@
...
+    try (Stream<Path> layerDirectories =
+        Files.list(defaultCacheStorageFiles.getLayersDirectory())) {
+      List<Path> layerDirectoriesList = layerDirectories.collect(Collectors.toList());
+      List<DescriptorDigest> layerDigests = new ArrayList<>(layerDirectoriesList.size());
+      for (Path layerDirectory : layerDirectoriesList) {
+        try {
+          layerDigests.add(DescriptorDigest.fromHash(layerDirectory.getFileName().toString()));
+
+        } catch (DigestException ex) {
+          throw new CacheCorruptedException("Found non-digest file in layers directory", ex);
```
loosebazooka: What's the solution when this happens? Delete the layer directory?

loosebazooka: Do we plan on having a cacheFixer?

coollog: Yea, for now we'll just defer to the user of the class to handle corruptions like this.

coollog:
@loosebazooka Reading won't be corrupted by a writer is "guaranteed" by writes always writing atomically (writing to a temporary file and atomically moving to the actual location).

PR #899, opened 2018-08-29
Make createdBy in history configurable

### LINKED ISSUE #875: write history as part of layer meta data.

TadCordle:
Fixes #875 

`created_by` is now `jib-gradle-plugin` for gradle and `jib-maven-plugin` for maven


TadCordle:
@briandealwis Hmm we use it for the User-Agent header suffix and a few other random places for logging. Maybe it'd be good to combine these in the future (I don't think I want to make any changes like that in this PR though).

PR #901, opened 2018-08-29
Adds DefaultCacheStorage#retrieve.

coollog:
Part of #637


jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageReader.java
```diff
@@ -58,4 +60,46 @@
...
+      return Optional.empty();
+    }
+
+    DefaultCacheEntry.Builder cacheEntryBuilder =
+        DefaultCacheEntry.builder().setLayerDigest(layerDigest);
+
+    try (Stream<Path> filesInLayerDirectory = Files.list(layerDirectory)) {
+      for (Path fileInLayerDirectory : filesInLayerDirectory.collect(Collectors.toList())) {
+        if (DefaultCacheStorageFiles.isLayerFile(fileInLayerDirectory)) {
+          if (cacheEntryBuilder.getLayerBlob().isPresent()) {
```
loosebazooka: I'm a little confused, so bear with me: `CacheEntry` seem kind of like a union type, so this code feels weird. Should it either build `metadataCacheEntry` or `layerCacheEntry`? Can a cache entry contain both `metadata` and `layer`? If not, can we have multiple metadata files? can line 98 be hit more than once?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheEntry.java
```diff
@@ -61,6 +61,10 @@ public Builder setMetadataBlob(@Nullable Blob metadataBlob) {
       return this;
     }
 
+    public Optional<Blob> getLayerBlob() {
+      return Optional.ofNullable(layerBlob);
```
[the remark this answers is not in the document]
coollog: This method is actually just used in `DefaultCacheStorageReader#retrieve` for checking if the `layerBlob` has been set yet on the builder.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageFiles.java
```diff
@@ -27,6 +28,44 @@
...
   private static final String SELECTORS_DIRECTORY = "selectors";
 
+  /**
+   * Returns whether or not {@code file} is a layer contents file.
+   *
+   * @param file the file to check
+   * @return {@code true} if {@code file} is a layer contents file; {@code false} otherwise
+   */
+  static boolean isLayerFile(Path file) {
+    return file.getFileName().toString().endsWith(LAYER_FILENAME_SUFFIX);
```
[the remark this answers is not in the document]
coollog: Yes, I'll change it to just use the length since it's more efficient.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageReader.java
coollog: Yea, `CacheEntry` is a union type of either just `layer` data or both `layer` and `metadata` data (the former being base image layers and the latter being application layers). Line 98 won't be hit more than once because the metadata file is named by exact name, though this code should probably not make that assumption. I'll change it to not make that assumption.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageFiles.java
```diff
@@ -18,15 +18,53 @@
...
   private static final String SELECTORS_DIRECTORY = "selectors";
 
+  /**
+   * Returns whether or not {@code file} is a layer contents file.
+   *
+   * @param file the file to check
+   * @return {@code true} if {@code file} is a layer contents file; {@code false} otherwise
+   */
+  static boolean isLayerFile(Path file) {
+    return file.getFileName().toString().length() == DescriptorDigest.HASH_LENGTH;
```
TadCordle: Can/should we do better here? This doesn't seem like a very thorough check. I could see someone setting an arbitrary 64-character filename to get past this check or something.

coollog: I think the assumptions for this cache is that the user will not be messing with its contents, and if they do, we have no guarantees about possible data corruption, so such a check here I think would be okay rather than fully matching against the hash regex. For example, the user could modify the layer file itself and produce completely corrupted layer contents, but we would not be liable for detecting/fixing that.

loosebazooka: Maybe we should at least verify the layer contents still hash to the same value?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageFiles.java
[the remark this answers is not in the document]
loosebazooka: I'm not sure if endswith is particularly expensive? Is length actually a better check?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageReader.java
loosebazooka: Are they functionally different in any other part of the code? Should they be handled as different entry types?

coollog: When being read, they are not functionally different. The only use for the metadata currently is to check the last modified time when trying to see if a layer needs rebuilding (select the layer by the selector, then check if the files have any modified times greater than that stored in the metadata).

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageFiles.java
coollog: Yea, the hashing is an expensive operation. I believe that even if we had that check, since the selector file could be modified as well, a selector generated from a set of layer entries could select a completely different layer than intended and we don't have an easy way of checking the validity of this with anything less than having to decompress the layer and read all the tar archive entries.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageFiles.java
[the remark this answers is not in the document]
coollog: Probably negligibly more expensive. Length is definitely just a constant time value fetch though.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageFiles.java
loosebazooka: Oh it's a hash of the decompressed file? Can we verify integrity with a hash of the compressed file somewhere?

coollog: If the compressed hash doesn't actually match the uncompressed hash, the Docker registry would actually reject it (hence why they have diff ID in addition to the layer digest), though the corruption would be hidden under a manifest invalid error from the registry during push. I'll file a separate issue to explore potential cache corruptions to detect and test for solutions that may be adequate and fast.

loosebazooka: :+1:

PR #902, opened 2018-08-29
Propagate WorkingDir in container configuration

chanseokoh:
Part of #595.


chanseokoh:
Ready for review.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -135,6 +136,11 @@
       return this;
     }
 
+    public Builder<T> setWorkingDirectory(@Nullable String workingDirectory) {
```
coollog: javadoc here

PR #906, opened 2018-08-30
Add unzip util method

chanseokoh:
For use in #431.


jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import org.apache.commons.compress.utils.IOUtils;
+
+/** Utility class for Zip archives. */
+public class ZipUtil {
+
+  // TODO: prevent Zip-Slip vulnerability: https://snyk.io/research/zip-slip-vulnerability#java
+  public static void unzip(Path archive, Path destination) throws IOException {
+    try (ZipInputStream zipIn = new ZipInputStream(Files.newInputStream(archive))) {
```
coollog: Should we buffer these file streams?

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import org.apache.commons.compress.utils.IOUtils;
+
+/** Utility class for Zip archives. */
+public class ZipUtil {
```
coollog: Will this file be used outside of the plugins? If not, we could just put it in `plugins-common`.

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+  // TODO: prevent Zip-Slip vulnerability: https://snyk.io/research/zip-slip-vulnerability#java
+  public static void unzip(Path archive, Path destination) throws IOException {
+    try (ZipInputStream zipIn = new ZipInputStream(Files.newInputStream(archive))) {
+
+      for (ZipEntry entry = zipIn.getNextEntry(); entry != null; entry = zipIn.getNextEntry()) {
+        Path entryPath = destination.resolve(entry.getName());
```
coollog: Could `entry.getName()` possibly refer to paths preceding `destination`?

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+        if (entry.isDirectory()) {
+          Files.createDirectories(entryPath);
+        } else {
+          try (OutputStream out = Files.newOutputStream(entryPath)) {
+            IOUtils.copy(zipIn, out);
```
coollog: nit: `ByteStreams#copy` is what is used in the rest of the codebase.

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
```diff
@@ -0,0 +1,47 @@
...
+import java.nio.file.Path;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import org.apache.commons.compress.utils.IOUtils;
+
+/** Utility class for Zip archives. */
+public class ZipUtil {
+
+  // TODO: prevent Zip-Slip vulnerability: https://snyk.io/research/zip-slip-vulnerability#java
+  public static void unzip(Path archive, Path destination) throws IOException {
```
coollog: Should this allow unzipping to non-empty destination directories? Or would it be safer to require an empty/non-existing destination directory?

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
chanseokoh: That's the Zip-Slip vulnerability I mentioned above. I can address that in a later PR.

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
chanseokoh: Yeah, probably unzipping will happen at the plugin level, so I'll move it to `plugins-common`.

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
chanseokoh: I looked into the code of `ZipInputStream`, and it does look like it uses a buffer. However, it won't hurt buffer the stream anyway.

jib-core/src/main/java/com/google/cloud/tools/jib/zip/ZipUtil.java
chanseokoh: As a general util method, I think it makes sense to allow unzipping to non-empty destinations, which is what other zip/tar CLIs do.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java
```diff
@@ -14,31 +14,41 @@
...
+        ZipInputStream zipIn = new ZipInputStream(new BufferedInputStream(fileIn))) {
 
       for (ZipEntry entry = zipIn.getNextEntry(); entry != null; entry = zipIn.getNextEntry()) {
+        // TODO: check Zip-Slip vulnerability: https://snyk.io/research/zip-slip-vulnerability#java
         Path entryPath = destination.resolve(entry.getName());
 
         if (entry.isDirectory()) {
           Files.createDirectories(entryPath);
         } else {
           try (OutputStream out = Files.newOutputStream(entryPath)) {
```
coollog: And `BufferedOutputStream` here too?

chanseokoh: Looking at the code of `ByteStreams.copy`, it does use a buffer in a very obvious way, so I'm inclined to leave the code simple. You could still argue that's an implementation detail of `ByteStreams` though.

chanseokoh:
Zip-Slip fix merged into this PR.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java
```diff
@@ -0,0 +1,64 @@
...
+
+    try (InputStream fileIn = Files.newInputStream(archive);
+        ZipInputStream zipIn = new ZipInputStream(new BufferedInputStream(fileIn))) {
+
+      for (ZipEntry entry = zipIn.getNextEntry(); entry != null; entry = zipIn.getNextEntry()) {
+        Path entryPath = destination.resolve(entry.getName());
+
+        String canonicalTarget = entryPath.toFile().getCanonicalPath();
+        if (!canonicalTarget.startsWith(canonicalDestination + File.separator)) {
+          throw new IOException("Blocked unzipping files outside destination: " + entry.getName());
```
briandealwis: Could we include the filename too?

chanseokoh: It's printing `entry.getName()` which is the file path in the zip. If this error happens, it would look like `../../../evil.attack`.

briandealwis: I meant the zip name

chanseokoh: Oops!

PR #908, opened 2018-08-30
Eliminate Zip-Slip vulnerability

chanseokoh:
On top of #906.

https://snyk.io/research/zip-slip-vulnerability


chanseokoh:
Seems like Kokoro won't run when not merging into master. I'll just put the "PR: Not Ready" until merging #906.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java
```diff
@@ -37,13 +38,19 @@
...
+
     try (InputStream fileIn = Files.newInputStream(archive);
         ZipInputStream zipIn = new ZipInputStream(new BufferedInputStream(fileIn))) {
 
       for (ZipEntry entry = zipIn.getNextEntry(); entry != null; entry = zipIn.getNextEntry()) {
-        // TODO: check Zip-Slip vulnerability: https://snyk.io/research/zip-slip-vulnerability#java
         Path entryPath = destination.resolve(entry.getName());
 
+        String canonicalTarget = entryPath.toFile().getCanonicalPath();
+        if (!canonicalTarget.startsWith(canonicalDestination + File.separator)) {
```
coollog: Could this be done with NIO? Like maybe `entryPath.toAbsolutePath().startsWith(destination.toAbsolutePath())`?

coollog: Okay, though it feels off having to use the deprecated `java.io.File` API to account for a security vulnerability. Perhaps we could use `Path#normalize`?

chanseokoh: I still don't think `Path.normalize()` would work. A canonical path is a unique path for any given file; there can ever be only one canonical path. Computing a canonical path may involve "resolving symbolic links (on UNIX platforms), and converting drive letters to a standard case (on Microsoft Windows platforms)" according to its Javadoc. So I think it's safe to stick with the canonical path approach.

PR #911, opened 2018-08-30
Add skip parameter to base plugin configuration

### LINKED ISSUE #865: Add configuration to skip in Maven plugin
In multi-module projects usually the parent POM is used to specify plugin and dependency versions to be used in child modules. Then, there are build plugins that are not desired to be enabled for all modules. Therefore it has become a common practice by Maven plugin to support a 'skip' configuration item. If skip = true then the plugin should skip that module. Then we can for example set skip = true as default and enable it (skip = false) for modules that need that plugin.

So it would be awesome if the jib-maven-plugin also could support such a 'skip' configuration.

loosebazooka:
- can use system property -Djib.skip=true to skip if part of
  lifecycle
- still requires "required parameters" on jib to be configured
  even if skipping, since maven doesn't provide a api for actual
  goal skipping

fixes #865

### DIFF
--- a/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
+++ b/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
@@ -17,6 +17,9 @@
 package com.google.cloud.tools.jib.maven;
 
 import com.google.cloud.tools.jib.Command;
+import com.google.common.base.Charsets;
+import com.google.common.io.Files;
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.time.Instant;
@@ -41,6 +44,9 @@ public class BuildDockerMojoIntegrationTest {
   @ClassRule
   public static final TestProject defaultTargetTestProject =
       new TestProject(testPlugin, "default-target");
+
+  @ClassRule
+  public static final TestProject skippedTestProject = new TestProject(testPlugin, "empty");
 
   /**
    * Builds and runs jib:buildDocker on a project at {@code projectRoot} pushing to {@code
@@ -109,5 +115,25 @@ public class BuildDockerMojoIntegrationTest {
         buildToDockerDaemonAndRun(
             defaultTargetTestProject.getProjectRoot(),
             "default-target-name:default-target-version"));
+  }
+
+  @Test
+  public void testExecute_skipJibGoal() throws VerificationException, IOException {
+    String targetImage = "emptyimage:maven" + System.nanoTime();
+
+    Verifier verifier = new Verifier(skippedTestProject.getProjectRoot().toString());
+    verifier.setSystemProperty("_TARGET_IMAGE", targetImage);
+    verifier.setAutoclean(false);
+    verifier.setSystemProperty("jib.skip", "true");
+
+    verifier.executeGoal("jib:" + BuildDockerMojo.GOAL_NAME);
+    File logFile = new File(verifier.getBasedir(), verifier.getLogFileName());
+    Assert.assertTrue(
+        Files.asCharSource(logFile, Charsets.UTF_8)
+            .read()
+            .contains(
+                "[INFO] Skipping containerizations because jib-maven-plugin: skip = true\n"
+                    + "[INFO] ------------------------------------------------------------------------\n"
+                    + "[INFO] BUILD SUCCESS"));
   }
 }
--- a/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
+++ b/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
@@ -18,10 +18,7 @@ package com.google.cloud.tools.jib.maven;
 
 import com.google.cloud.tools.jib.Command;
 import java.io.IOException;
-import java.nio.charset.Charset;
-import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.time.Instant;
 import org.apache.maven.it.VerificationException;
 import org.apache.maven.it.Verifier;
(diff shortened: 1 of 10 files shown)


loosebazooka:
If anyone has any decent testing strategies for this, let me know. It's kind of hard to do since maven doesn't actually define "skipping" a goal like gradle does.

briandealwis:
You could run something like the `BuildImageMojoIntegrationTest` and use the `Verifier#verifyTextInLog()` that your log string appeared.

loosebazooka:
@briandealwis I was hoping to ensure that the return was called, maybe I can ensure the log has no more writes after that?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
```diff
@@ -110,4 +116,24 @@ public void testExecute_defaultTarget()
...
+  public void testExecute_skipJibGoal() throws VerificationException, IOException {
+    String targetImage = "emptyimage:maven" + System.nanoTime();
+
+    Verifier verifier = new Verifier(skippedTestProject.getProjectRoot().toString());
+    verifier.setSystemProperty("_TARGET_IMAGE", targetImage);
+    verifier.setAutoclean(false);
+    verifier.setSystemProperty("jib.skip", "true");
+
+    verifier.executeGoal("jib:" + BuildDockerMojo.GOAL_NAME);
+    File logFile = new File(verifier.getBasedir(), verifier.getLogFileName());
```
coollog: nit: use `Path`? And `new String(Files.readAllBytes(logFile), StandardCharsets.UTF_8)`

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
```diff
@@ -110,4 +116,24 @@ public void testExecute_defaultTarget()
...
+    Assert.assertTrue(
```
coollog: `Assert.assertThat(string, CoreMatchers.containsString())`

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
chanseokoh: Yeah, that way, it will remove the Guava dependency.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,34 @@
...
+    Verifier verifier = new Verifier(testProject.getProjectRoot().toString());
+    verifier.setSystemProperty("_TARGET_IMAGE", targetImage);
+    verifier.setAutoclean(false);
+    verifier.setSystemProperty("jib.skip", "true");
+
+    verifier.executeGoal("jib:" + goal);
+
+    Path logFile = Paths.get(verifier.getBasedir(), verifier.getLogFileName());
+    Assert.assertThat(
+        new String(Files.readAllBytes(logFile), Charset.forName("UTF-8")),
```
coollog: `StandardCharsets.UTF_8`

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,34 @@
...
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.hamcrest.CoreMatchers;
+import org.junit.Assert;
+
+public class SkippedGoalVerifier {
```
coollog: short javadoc

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,34 @@
...
+
+  public static void verifyGoalIsSkipped(TestProject testProject, String goal)
```
coollog: visibility can all be package-private?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,34 @@
...
+
+    Path logFile = Paths.get(verifier.getBasedir(), verifier.getLogFileName());
+    Assert.assertThat(
+        new String(Files.readAllBytes(logFile), Charset.forName("UTF-8")),
+        CoreMatchers.containsString(
+            "[INFO] Skipping containerization because jib-maven-plugin: skip = true\n"
+                + "[INFO] ------------------------------------------------------------------------\n"
+                + "[INFO] BUILD SUCCESS"));
+  }
+}
```
coollog: private constructor

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,34 @@
...
+import org.junit.Assert;
+
+public class SkippedGoalVerifier {
+
+  public static void verifyGoalIsSkipped(TestProject testProject, String goal)
+      throws VerificationException, IOException {
+    String targetImage = "neverbuilt:maven" + System.nanoTime();
+
+    Verifier verifier = new Verifier(testProject.getProjectRoot().toString());
+    verifier.setSystemProperty("_TARGET_IMAGE", targetImage);
```
coollog: I think we can just remove this line?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,34 @@
+package com.google.cloud.tools.jib.maven;
```
coollog: copyright

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
loosebazooka: without target image specified, it will fail?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,52 @@
...
+import java.nio.file.Paths;
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.hamcrest.CoreMatchers;
+import org.junit.Assert;
+
+/** A simple verifier utility to test goal skipping accross all our jib goals. */
+class SkippedGoalVerifier {
+
+  private SkippedGoalVerifier() {}
```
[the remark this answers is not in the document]
loosebazooka: oh right

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/SkippedGoalVerifier.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
```
briandealwis: I meant to ask about this _All rights reserved._ It's [not normally seen on ASL headers](https://www.apache.org/licenses/LICENSE-2.0.html#apply), and @elharo indicated that there was [some internal discussion about whether to include it or not](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/pull/2964#issuecomment-375420457).

loosebazooka: Okay, yeah I just looked at the updating docs, and `All rights reserved` is gone. I'll update that, and then push a larger update across the codebase.

PR #917, opened 2018-08-31
Preparation for Gradle WAR support

chanseokoh:
Towards #431.

For now, create a single layer for the exploded WAR directory for the first iteration. I thought about creating a separate class `WarLayerConfiguration`, but it seems not worth.


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -56,6 +58,16 @@ static GradleProjectProperties getForProject(
     }
   }
 
+  @Nullable
+  static War getWar(Project project) {
```
coollog: `getWarTask`?

chanseokoh: Heads-up: this is going to be called from `GradleLayerConfigurations` (to get the files inside a WAR), so not `private`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -97,6 +98,11 @@ public Builder setExtraFiles(List<Path> extraFiles) {
       return this;
     }
 
+    public Builder setExplodedWar(List<Path> explodedWar) {
```
coollog: Hmm, didn't we want to split the exploded war into the classes, resources, and dependencies layers?

chanseokoh: Yeah, that is true. The single layer is for the first iteration.

patflynn:
just curious if this is going to suffer from the same packaging manifest issues as #jar

chanseokoh:
@patflynn not sure about that, thinking servlets engines might pick up the manifest even if the web app is loaded from an exploded directory. This is something I can test.

briandealwis:
The app servers explode the wars as a normal course of operation.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
```diff
@@ -37,6 +38,15 @@
         mainClass);
   }
 
+  /**
+   * Constructs the container entrypoint for the gcr.io/distroless/jetty base image.
+   *
+   * @return ["java", "-jar", "/jetty/start.jar"]
+   */
+  public static List<String> makeDistrolessJettyEntrypoint() {
```
briandealwis: Should this take jvmFlags?

jib-core/src/test/resources/sampleDockerfile
```diff
@@ -4,6 +4,7 @@ COPY libs /app/libs/
 COPY snapshot-libs /app/libs/
 COPY resources /app/resources/
 COPY classes /app/classes/
+COPY exploded-war /jetty/webapps/ROOT/
```
briandealwis: I don't think normal jar projects should see `exploded-war` references, and the `/app` locations don't seem right.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
briandealwis: Also, the `distroless/java/jetty` image has:
  - `entrypoint: ["/usr/bin/java", "-jar"]`
  - `cmd: ["/jetty/start.jar"]`

So we could just not set anything by default, right?

(As compared to `distroless/java` which sets `entrypoint: ["/usr/bin/java", "-jar"]` and no `cmd`, so we must change the entrypoint.)

jib-core/src/test/resources/sampleDockerfile
chanseokoh: Normal jar projects won't see `exploded-war`. The `DockerContextGenerator` will add only what it sees. The `sampleDockerfile` is just for testing that the generator has the capacity to handle everything.

chanseokoh: Also, `/app` will not be there for a WAR project.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
chanseokoh: Sounds good if we are propagating both `Entrypoint` and `Cmd`. I think we do?

chanseokoh:
> I guess I was expecting that you would just configure the `JavaLayerConfigurations` with the appropriate locations within the WAR.

That makes sense. For now, I'm deferring to later making `JavaLayerConfigurations` work with fine-grained WAR layers with a possibly customizable webapp location in the image to support beyond the distroless/jetty, since for now, it is easy to dump the whole webapp. (The later PR to unzip and set sources would be smaller and easier to review this way.) I think the task for fine-grained WAR layers for arbitrary Servlet engine base image can be done in an isolated manner later, which will aid focusing on only that task.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
briandealwis: Hmm, [it doesn't look like we do](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java#L118).

chanseokoh:
I think I should continue working on this. #923 became stalled. Ready for review.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
```diff
@@ -37,6 +38,15 @@
         mainClass);
   }
 
+  /**
+   * Constructs the container entrypoint for the gcr.io/distroless/jetty base image.
+   *
+   * @return ["java", "-jar", "/jetty/start.jar"]
```
coollog: Might be good to add a link for easy reference:

`@see <a href="https://github.com/GoogleContainerTools/distroless/blob/master/java/jetty/BUILD">https://github.com/GoogleContainerTools/distroless/blob/master/java/jetty/BUILD</a>`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -111,6 +123,12 @@ public String getJarPluginName() {
     return JAR_PLUGIN_NAME;
   }
 
+  @Override
+  public boolean isWarProject() {
+    // TODO: replace with "getWar(project) != null" once ready
```
coollog: Is this ready?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -41,6 +45,9 @@
   @Mock private Jar mockJar;
   @Mock private Jar mockJar2;
   @Mock private Project mockProject;
+  @Mock private Convention mockConvention;
+  @Mock WarPluginConvention mockWarPluginConvection;
```
coollog: private?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -130,6 +130,11 @@ public String getJarPluginName() {
     return JAR_PLUGIN_NAME;
   }
 
+  @Override
+  public boolean isWarProject() {
+    return false; // TODO: to be implemented. For now, assume false.
```
coollog: Is this for a future PR?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
coollog: And also a TODO to just set the CMD?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
chanseokoh: I'll use this as a feature flag during development, and make it work only when all the parts are working.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: Had to add the method, since we should implement the interface.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
chanseokoh: I'm always confused when mixing CMD and ENTRYPOINT so not understanding the entire story, but I'll add a TODO to use CMD instead of ENTRYPOINT.

coollog: Oh I meant like set the CMD for `/jetty/start.jar` and have ENTRYPOINT be inherited (`java -jar`), so that the container entrypoint becomes `java -jar /jetty/start.jar`.

chanseokoh: Oh, in that case, I think we don't need to set CMD either if we inherit both CMD and ENTRYPONIT. CMD is set with `/jetty/start.jar` and ENTRYPOINT `[/usr/bin/java, -jar]` in the base Jetty image.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
coollog: Let's add a TODO to remove this? (and elsewhere where we want to move the exploded war files into the classes/resources/dependencies? like `JavaDockerContextGenerator`)

loosebazooka:
Should we include the changes for configuring the target directory in this PR? Or should we do that first for all of jib? Or even as a followup PR?

PR #921, opened 2018-08-31
Adds JibEvent, JibEventType, and EventHandlers.

coollog:
Regards #909 

A preliminary implementation that shows a bit of how it will work.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventHandlers.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventHandlers.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.event;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+/** Builds a set of event handlers to handle {@link JibEvent}s. */
+public class EventHandlers {
+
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/event/EventHandlers.java
```diff
@@ -0,0 +1,101 @@
...
+   * emitted.
+   *
+   * <p><b>Note: Implementations of {@code eventConsumer} must be thread-safe.</b>
+   *
+   * @param eventType the event type that {@code eventConsumer} should handle
+   * @param eventConsumer the event handler
+   * @param <E> the type of {@code eventClass}
+   * @return this
+   */
+  public <E extends JibEvent> EventHandlers add(
```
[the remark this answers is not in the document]
coollog: Yes, `EventHandlers` is meant to set up the map of handlers. An immutable version will be used in the actual execution.

jib-core/src/test/java/com/google/cloud/tools/jib/event/EventHandlersTest.java
```diff
@@ -0,0 +1,98 @@
...
+  /** Test implementation of {@link JibEvent}. */
+  private static class TestJibEvent1 implements JibEvent {
+
+    private boolean called = false;
+
+    private void assertGetPayloadCalled() {
+      Assert.assertTrue(called);
+    }
+
+    private void getPayload() {
```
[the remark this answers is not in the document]
coollog: Oh yea, `getPayload` is meant to mock a real method that would actually give back some event-related information. I'll actually change this to use mocks to make it more like the real deal.

jib-core/src/test/java/com/google/cloud/tools/jib/event/EventHandlersTest.java
```diff
@@ -53,46 +45,47 @@ private void sayHello(String name) {
...
-  /** Test class to hold mutable int. */
-  private static class Counter {
-
-    private int count = 0;
-  }
-
   @Test
   public void testAdd() {
-    Counter counter = new Counter();
+    int[] counter = new int[1];
```
TadCordle: Assuming you're just using this so you can pass by reference, maybe an `Integer` would look less weird?

chanseokoh: I think `Integer` is immutable, once set. Here we want to increment values, so I think it won't work that way.

coollog: Yea, I think the array method should be fine here since this is just a test and this array method is a common way to get around the effectively final requirement for lambdas.

TadCordle: Ah ok, didn't know Integer was immutable.

PR #930, opened 2018-09-04
Add plugin configuration for environment variables

### LINKED ISSUE #890: How to set environment variables in the container
Using the maven plugin.  How can I set environment variables within the container?

Also, when is 0.9.10 likely to be released as I need the label feature?

### LINKED ISSUE #929: Add CHANGELOG for skip parameter
#911

TadCordle:
Fixes #890 

This adds a new `environment` configuration parameter to the maven and gradle plugins to set environment variables on the container image.

Maven:
```xml
<configuration>
  <container>
    <environment>
      <key1>value1</key1>
      <key2>value2</key2>
    </environment>
  </container>
</configuration>
```

Gradle:
```groovy
jib {
  container {
    environment = [key1:"value1", key2:"value2"]
  }
}
```

This PR also adds the corresponding "ENV" command to the docker context generator.

Also slipped `skip` into the changelog to fix #929

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
@@ -216,10 +216,10 @@ public class BuildStepsIntegrationTest {
                 + "                \"key1\": \"value1\",\n"
                 + "                \"key2\": \"value2\"\n"
                 + "            }"));
-    Assert.assertThat(
-        dockerContainerConfig,
-        CoreMatchers.containsString(
-            "                \"var1=value1\",\n                \"var2=value2\"\n"));
+
+    String dockerConfigEnv =
+        new Command("docker", "inspect", "-f", "{{.Config.Env}}", imageReference).run();
+    Assert.assertThat(dockerConfigEnv, CoreMatchers.containsString("var1=value1 var2=value2"));
     String history = new Command("docker", "history", imageReference).run();
     Assert.assertThat(history, CoreMatchers.containsString("jib-integration-test"));
     Assert.assertThat(history, CoreMatchers.containsString("bazel build ..."));
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
@@ -85,7 +85,7 @@ public class BuildStepsIntegrationTest {
                 + "            }"));
     String dockerConfigEnv =
         new Command("docker", "inspect", "-f", "{{.Config.Env}}", imageReference).run();
-    Assert.assertThat(dockerConfigEnv, CoreMatchers.containsString("var1=value1 var2=value2"));
+    Assert.assertThat(dockerConfigEnv, CoreMatchers.containsString("env1=value1 env2=value2"));
     String history = new Command("docker", "history", imageReference).run();
     Assert.assertThat(history, CoreMatchers.containsString("jib-integration-test"));
     Assert.assertThat(history, CoreMatchers.containsString("bazel build ..."));
@@ -210,7 +210,7 @@ public class BuildStepsIntegrationTest {
                 JavaEntrypointConstructor.makeDefaultEntrypoint(
                     Collections.emptyList(), "HelloWorld"))
             .setProgramArguments(Collections.singletonList("An argument."))
-            .setEnvironment(ImmutableMap.of("var1", "value1", "var2", "value2"))
+            .setEnvironment(ImmutableMap.of("env1", "value1", "env2", "value2"))
             .setExposedPorts(
                 ExposedPortsParser.parse(Arrays.asList("1000", "2000-2002/tcp", "3000/udp")))
             .setLabels(ImmutableMap.of("key1", "value1", "key2", "value2"))
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
@@ -85,7 +85,8 @@ public class BuildStepsIntegrationTest {
                 + "            }"));
     String dockerConfigEnv =
         new Command("docker", "inspect", "-f", "{{.Config.Env}}", imageReference).run();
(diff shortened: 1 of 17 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
```diff
@@ -226,4 +197,31 @@ private BuildConfiguration getBuildConfiguration(
...
+        dockerContainerConfig,
+        CoreMatchers.containsString(
+            "            \"Labels\": {\n"
+                + "                \"key1\": \"value1\",\n"
+                + "                \"key2\": \"value2\"\n"
+                + "            }"));
+    Assert.assertThat(
+        dockerContainerConfig,
+        CoreMatchers.containsString(
+            "                \"var1=value1\",\n                \"var2=value2\"\n"));
```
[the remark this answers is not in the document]
TadCordle: The issue is that there are other environment variables already in there, like PATH and SSL_CERT_FILE. I guess I could use regex to skip those or just match those variables in the string as well (I'm not sure if they change though)?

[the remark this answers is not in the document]
TadCordle: Actually I can just use `docker inspect -f {{.Config.Env}} <image>`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -170,6 +172,17 @@ public JavaDockerContextGenerator setJavaArguments(List<String> javaArguments) {
     return this;
   }
 
+  /**
+   * Sets the environment variables
+   *
+   * @param environment the environment variables
```
coollog: nit: map from environment variable name to value?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -267,14 +283,20 @@ String makeDockerfile() throws JsonProcessingException {
       dockerfile.append("\nEXPOSE ").append(port);
     }
 
-    boolean firstLabel = true;
+    StringJoiner joiner = new StringJoiner(" \\\n    ", "\nENV ", "");
+    for (Entry<String, String> env : environment.entrySet()) {
+      joiner.add(env.getKey() + "=" + objectMapper.writeValueAsString(env.getValue()));
+    }
+    if (environment.entrySet().size() > 0) {
```
coollog: nit: can just be `environment.size()`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -267,14 +283,20 @@ String makeDockerfile() throws JsonProcessingException {
...
+      dockerfile.append(joiner.toString());
+    }
+
+    joiner = new StringJoiner(" \\\n      ", "\nLABEL ", "");
```
coollog: The logic for labels seems to be the same as the logic for environment - maybe condense to a static method?

jib-gradle-plugin/src/integration-test/resources/projects/simple/complex-build.gradle
```diff
@@ -34,6 +34,7 @@ jib {
         args = ['An argument.']
         mainClass = 'com.test.HelloWorld'
         jvmFlags = ['-Xms512m', '-Xdebug']
+        environment = [var1:'value1', var2:'value2']
```
coollog: nit: it might be clearer if the values were like `env1`/`env2`?

jib-gradle-plugin/src/integration-test/resources/projects/simple/complex-build.gradle
```diff
@@ -34,7 +34,7 @@ jib {
         args = ['An argument.']
         mainClass = 'com.test.HelloWorld'
         jvmFlags = ['-Xms512m', '-Xdebug']
-        environment = [var1:'value1', var2:'value2']
+        environment = [env1:'value1', env2:'value2']
```
coollog: Oh I was referring to `value1` and `value2` since it would make the `assertEquals` be something like `\nenv1value\nenv2value\n` rather than `\nvalue1\nvalue2\n` where `value1` and `value2` are not immediately clear where they should have come from.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -99,6 +99,35 @@ private static void addIfNotEmpty(
...
+   * }</pre>
+   *
+   * @param map the map to convert
+   * @param command the dockerfile command to prefix the map values with
+   * @param objectMapper used to json-ify map values
+   * @return the new dockerfile command as a string
+   * @throws JsonProcessingException if getting the json string of a map value fails
+   */
+  private static String mapToDockerfileString(
+      Map<String, String> map, String command, ObjectMapper objectMapper)
```
coollog: `ObjectMapper` can prob just be eliminated and made into a constant.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -99,6 +99,35 @@ private static void addIfNotEmpty(
             layerEntry.getSourceFiles(), directoryInContext, layerEntry.getExtractionPath()));
   }
 
+  /**
+   * Converts a map to a corresponding dockerfile string in the form of:
+   *
+   * <pre>{@code
+   * command key1="value1" \
+   *     key2="value2" \
```
coollog: nit: should we line up `key1` and `key2` since we're adding spaces anyways?

patflynn:
It'd be really nice if these PRs had a bit more content in the description so it's really easy for reviewers to understand what the changes are without looking at code. 

For instance if this PR actually changes plugin config params it should include examples of what it enables.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
TadCordle: The command names are different lengths, and adding a variable number of spaces here would have harmed readability or required a weird extra parameter, so I don't think it's worth it, unless there's a function we can use to generate whitespace that I don't know about (that doesn't require another `import`). I think it's fine as is.

PR #931, opened 2018-09-04
Adds EventEmitter.

coollog:
Continuation of #921

Regards #909

`EventEmitter` will be used in the Jib execution to emit the actual events.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.event;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.Collections;
+import java.util.List;
+
+/** Emits {@link JibEvent}s to event handlers. */
+public class EventEmitter {
+
+  /** Maps from {@link JibEvent} class to handlers for that event type. */
+  private final ImmutableMap<Class<? extends JibEvent>, List<Handler<? extends JibEvent>>> handlers;
+
+  /**
+   * Creates from {@link Handler}s in an {@link EventHandlers}.
+   *
+   * @param eventHandlers the {@link EventHandlers} to get the {@link Handler}s from
+   */
+  public EventEmitter(EventHandlers eventHandlers) {
+    handlers = ImmutableMap.copyOf(eventHandlers.getHandlers());
+    System.out.println(handlers);
+  }
+
+  /**
+   * Emits {@code jibEvent} to all the handlers that can handle it.
+   *
+   * @param jibEvent the {@link JibEvent} to emit
+   */
+  public void emit(JibEvent jibEvent) {
+    for (Handler<? extends JibEvent> handler : getHandlersFor(JibEvent.class)) {
+      handler.handle(jibEvent);
+    }
+    for (Handler<? extends JibEvent> handler : getHandlersFor(jibEvent.getClass())) {
+      handler.handle(jibEvent);
+    }
+  }
+
+  private List<Handler<? extends JibEvent>> getHandlersFor(
+      Class<? extends JibEvent> jibEventClass) {
+    List<Handler<? extends JibEvent>> handlerList = handlers.get(jibEventClass);
+    if (handlerList == null) {
+      return Collections.emptyList();
+    }
+    return handlerList;
+  }
+}
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
@@ -33,7 +33,6 @@ public class EventEmitter {
    */
   public EventEmitter(EventHandlers eventHandlers) {
     handlers = ImmutableMap.copyOf(eventHandlers.getHandlers());
-    System.out.println(handlers);
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
@@ -16,23 +16,21 @@
 
 package com.google.cloud.tools.jib.event;
 
-import com.google.common.collect.ImmutableMap;
-import java.util.Collections;
-import java.util.List;
+import com.google.common.collect.ImmutableMultimap;
 
 /** Emits {@link JibEvent}s to event handlers. */
 public class EventEmitter {
 
   /** Maps from {@link JibEvent} class to handlers for that event type. */
-  private final ImmutableMap<Class<? extends JibEvent>, List<Handler<? extends JibEvent>>> handlers;
+  private final ImmutableMultimap<Class<? extends JibEvent>, Handler<? extends JibEvent>> handlers;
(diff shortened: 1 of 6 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
```diff
@@ -0,0 +1,61 @@
...
+  private final ImmutableMap<Class<? extends JibEvent>, List<Handler<? extends JibEvent>>> handlers;
+
+  /**
+   * Creates from {@link Handler}s in an {@link EventHandlers}.
+   *
+   * @param eventHandlers the {@link EventHandlers} to get the {@link Handler}s from
+   */
+  public EventEmitter(EventHandlers eventHandlers) {
+    handlers = ImmutableMap.copyOf(eventHandlers.getHandlers());
+    System.out.println(handlers);
```
TadCordle: Forgot to remove?

coollog: Oops yes

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+/** Emits {@link JibEvent}s to event handlers. */
+public class EventEmitter {
+
+  /** Maps from {@link JibEvent} class to handlers for that event type. */
+  private final ImmutableMap<Class<? extends JibEvent>, List<Handler<? extends JibEvent>>> handlers;
+
+  /**
+   * Creates from {@link Handler}s in an {@link EventHandlers}.
```
briandealwis: This doesn't parse well.  _Creates an instance…_?

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Emits {@code jibEvent} to all the handlers that can handle it.
+   *
+   * @param jibEvent the {@link JibEvent} to emit
+   */
+  public void emit(JibEvent jibEvent) {
+    for (Handler<? extends JibEvent> handler : getHandlersFor(JibEvent.class)) {
+      handler.handle(jibEvent);
+    }
+    for (Handler<? extends JibEvent> handler : getHandlersFor(jibEvent.getClass())) {
```
briandealwis: If you'll forgive a moment of bikeshedding, could be condensed to:
```
    handlers
        .getOrDefault(JibEvent.class, Collections.emptyList())
        .forEach(handler -> handler.handle(jibEvent));
    handlers
        .getOrDefault(jibEvent.getClass(), Collections.emptyList())
        .forEach(handler -> handler.handle(jibEvent));
```

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventHandlers.java
```diff
@@ -26,34 +26,7 @@
...
-        eventConsumer.accept(eventClass.cast(jibEvent));
-        return true;
-      }
-      return false;
-    }
-  }
-
-  // Maps from JibEvent class to handlers for that event type.
+  /** Maps from {@link JibEvent} class to handlers for that event type. */
   private final Map<Class<? extends JibEvent>, List<Handler<? extends JibEvent>>> handlers =
```
briandealwis: Multimap?

jib-core/src/main/java/com/google/cloud/tools/jib/event/Handler.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Handles a {@link JibEvent}.
+   *
+   * @param jibEvent the event to handle
+   * @return {@code true} if this {@link Handler} handled the event; {@code false} if this {@link
+   *     Handler} does not handle the event
+   */
+  boolean handle(JibEvent jibEvent) {
+    if (eventClass.isInstance(jibEvent)) {
```
briandealwis: This should be a logical error, shouldn't it? ` Preconditions.checkArgument()`

jib-core/src/main/java/com/google/cloud/tools/jib/event/JibEventType.java
```diff
@@ -21,7 +21,7 @@
 /** Holds references to all {@link JibEvent} types. */
 public class JibEventType<E extends JibEvent> {
```
briandealwis: Why is this class necessary?

jib-core/src/main/java/com/google/cloud/tools/jib/event/Handler.java
```diff
@@ -0,0 +1,46 @@
...
+  Handler(Class<E> eventClass, Consumer<E> eventConsumer) {
+    this.eventClass = eventClass;
+    this.eventConsumer = eventConsumer;
+  }
+
+  /**
+   * Handles a {@link JibEvent}.
+   *
+   * @param jibEvent the event to handle
+   * @return {@code true} if this {@link Handler} handled the event; {@code false} if this {@link
```
briandealwis: How do you see this return result being used?

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package com.google.cloud.tools.jib.event;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.Collections;
+import java.util.List;
+
+/** Emits {@link JibEvent}s to event handlers. */
+public class EventEmitter {
```
chanseokoh: Who creates this instance and call `.emit()`? For example, I'm roughly guessing at the beginning of `PushBlobStep.call()`, will the `call()` method directly do

`new EventEmitter(eventHandler).emit(JibEventType.BEGIN_PUSH_BLOB)`

and at the end do

`new EventEmitter(eventHandlers).emit(JibEventType.END_PUSH_BLOB)`

? If so, how could `PushBlobStep` get the `eventHandlers` from whom? Does `PushBlobStep` need to know about `eventHandlers` and pass them to `EventEmitter`? Or, will there be another manager or layer that `PushBlobStep` will use?

coollog:
@chanseokoh Yes, that is correct - the class information is necessary to cast the emitted `JibEvent` into the subtype that the consumer specifically consumes. If there's a way to do it without the extra `Handler`, I would definitely prefer that approach.

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
coollog: The `BuildConfiguration` will have an `EventEmitter` that is created from an `EventHandlers` before execution begins. All steps will then have access to that `EventEmitter` to emit whichever events they wish.

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventHandlers.java
coollog: TIL that is perfect

jib-core/src/main/java/com/google/cloud/tools/jib/event/JibEventType.java
coollog: It's not completely necessary, but it's there to centralize all the event types so that users can find all the available event types to add handlers for in one location.

jib-core/src/main/java/com/google/cloud/tools/jib/event/Handler.java
coollog: That's true, I originally had this in a different usage (where all handlers are "tried" against `jibEvent`), but this is not necessary anymore.

jib-core/src/main/java/com/google/cloud/tools/jib/event/JibEventType.java
briandealwis: How does it do that?  Oh, I see:
```
// TODO: Add entries for all JibEvent types.
```
Can't that be handled by an IDE's _Type Hierarchy_ and Javadoc showing all subclasses of `JibEvent`?

coollog: That's true - should this just be removed then? The benefit I can see from `JibEventType` is that users would not need to have imports for the various `JibEvent` classes when they add handlers.

loosebazooka: Are we expecting this to be enum-esque?

PR #932, opened 2018-09-05
Improve the error message when pushing to a registry that does not support v2-2

### LINKED ISSUE #601: Handle Docker V2.1 image format or provide better error message.
Original comment by @jorgemoralespou https://github.com/GoogleContainerTools/jib/issues/534#issuecomment-404801072:

> I would think that nowadays there are very few (close to none) users with older versions of docker cli, but it would be good to support both schema1 and schema2 in the plugin, defaulting to schema2 but with the ability to override the value, and obviously a nicer error message explaining to the user what to do in case he get's a response from the registry that schema2 is not supported.

briandealwis:
Turns `MANIFEST_INVALID` / `TAG_INVALID` errors into slightly more helpful error messages:

>  _Tried to push image manifest for localhost:5000/foo:bar but failed because: Registry does not support Image Manifest Version 2, Schema 2_

Docker Registry 2.1, which returns a `400` / `TAG_INVALID`:
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.11-SNAPSHOT:build (default-cli) on project first: Tried to push image manifest for localhost:5000/foo:bar but failed because: Registry does not support Image Manifest Version 2, Schema 2 | If this is a bug, please file an issue at https://github.com/GoogleContainerTools/jib/issues/new: 400 Bad Request
[ERROR] {"errors":[{"code":"TAG_INVALID","message":"manifest tag did not match URI"}]}
[ERROR] -> [Help 1]
```

Docker Registry 2.2, which returns a `400` / `MANIFEST_INVALID`:
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.11-SNAPSHOT:build (default-cli) on project first: Tried to push image manifest for localhost:5000/foo:bar but failed because: Registry does not support Image Manifest Version 2, Schema 2 | If this is a bug, please file an issue at https://github.com/GoogleContainerTools/jib/issues/new: 400 Bad Request
[ERROR] {"errors":[{"code":"MANIFEST_INVALID","message":"manifest invalid","detail":{}}]}
[ERROR] -> [Help 1]
```

Pushing to quay.io, which returns a `415` / `MANIFEST_INVALID` (doesn't adhere to spec):
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.11-SNAPSHOT:build (default-cli) on project first: Tried to push image manifest for quay.io/briandealwis/foo:popo but failed because: Registry does not support Image Manifest Version 2, Schema 2 | If this is a bug, please file an issue at https://github.com/GoogleContainerTools/jib/issues/new: 415 UNSUPPORTED MEDIA TYPE
[ERROR] {"errors":[{"code":"MANIFEST_INVALID","detail":{"message":"manifest schema version not supported"},"message":"manifest invalid"}]}
[ERROR] -> [Help 1]
```

Fixes #601

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
@@ -17,14 +17,19 @@
 package com.google.cloud.tools.jib.registry;
 
 import com.google.api.client.http.HttpMethods;
+import com.google.api.client.http.HttpResponseException;
 import com.google.cloud.tools.jib.http.BlobHttpContent;
 import com.google.cloud.tools.jib.http.Response;
 import com.google.cloud.tools.jib.image.json.BuildableManifestTemplate;
 import com.google.cloud.tools.jib.json.JsonTemplateMapper;
+import com.google.cloud.tools.jib.registry.json.ErrorEntryTemplate;
+import com.google.cloud.tools.jib.registry.json.ErrorResponseTemplate;
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Collections;
 import java.util.List;
+import org.apache.http.HttpStatus;
 
 /** Pushes an image's manifest. */
 class ManifestPusher implements RegistryEndpointProvider<Void> {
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -53,6 +58,62 @@ public BlobHttpContent getContent() {
...
+          throw new RegistryErrorExceptionBuilder(getActionDescription(), httpResponseException)
+              .addReason("Registry does not support Image Manifest Version 2, Schema 2")
+              .build();
+        }
+      }
+
+    } catch (IOException ex) {
+      throw new RegistryErrorExceptionBuilder(getActionDescription(), httpResponseException)
+          .addReason("Failed to parse registry error response body")
+          .build();
```
chanseokoh: Maybe we should just throw `httpResponseException`? The rationale is that, if we didn't have this extra parsing at all like before, it would have thrown `httpResponseException`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -53,6 +58,62 @@ public BlobHttpContent getContent() {
...
+      ErrorResponseTemplate errorResponse =
+          JsonTemplateMapper.readJson(errorContent, ErrorResponseTemplate.class);
+      List<ErrorEntryTemplate> errors = errorResponse.getErrors();
+      if (errors.size() == 1) {
+        String errorCodeString = errors.get(0).getCode();
+        if (errorCodeString == null) {
+          // Did not get an error code back.
+          throw httpResponseException;
+        }
+        ErrorCodes errorCode = ErrorCodes.valueOf(errorCodeString);
```
chanseokoh: I noticed `valueOf(String)` will throw `IllegalArgumentException` if there is no matching enum, so I think we should catch that.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -53,6 +58,62 @@ public BlobHttpContent getContent() {
...
+        if (errorCode.equals(ErrorCodes.MANIFEST_INVALID)
+            || errorCode.equals(ErrorCodes.TAG_INVALID)) {
+          throw new RegistryErrorExceptionBuilder(getActionDescription(), httpResponseException)
+              .addReason("Registry does not support Image Manifest Version 2, Schema 2")
```
chanseokoh: I wonder if it's better to be a little bit less assertive with the claim, like "Registry may not be supporting ...". Just thinking `TAG_INVALID` might imply something different in some cases or there were genuinely other reasons (e.g., bugs) that produced an invalid manifest. WDYT?

TadCordle: Yeah, for example something like this would have been misleading for the artifactory errors, where the issue wasn't that they didn't support v2.2 schema, the issue was that they were expecting fields we didn't provide.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
briandealwis: Good catch.  Will handle this in #937 as this is a lightly modified block taken from elsewhere.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
briandealwis: Good catch.  Will handle this in #937 as this is a lightly modified block taken from elsewhere.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
briandealwis: Makes sense.  Note that the original Artifactory error report seemed to be due to [Artifactory not being configured to support v2-2 out of the box](
https://github.com/GoogleContainerTools/jib/issues/534#issuecomment-404679808).

It's too bad the servers don't return any actionable information to identify the registry type.  Docker Registry has the following headers:
```
{content-length=[79], content-type=[application/json; charset=utf-8], date=[Wed, 05 Sep 2018 18:12:25 GMT], docker-distribution-api-version=[registry/2.0], connection=[close]}
```
And `quay.io` has:
```
Headers: {content-length=[131], content-type=[application/json], date=[Wed, 05 Sep 2018 18:13:03 GMT], server=[nginx/1.14.0], connection=[keep-alive]}
```

PR #935, opened 2018-09-05
Consolidate tool name usage

### LINKED ISSUE #33: Refactor User-Agent injection.
`User-Agent` is currently passed in to the `RegistryClient` in `BuildImageMojo` with `RegistryClient.setUserAgentSuffix(USER_AGENT_SUFFIX);`.

### LINKED ISSUE #900: Consolidate usage of tool name under one parameter
For example, `PluginConfigurationProcessor#USER_AGENT_SUFFIX`, `BuildConfiguration#createdBy`, and several other logging places use the name of the plugin. It'd be nice to consider consolidating these values.

TadCordle:
Fixes #900 
Also fixes #33 

Replaces `setCreatedBy` on `BuildConfiguration` with `setToolName`, which can also be used to set the user agent suffix in `RegistryClient`.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -186,7 +186,7 @@ class PullBaseImageStep
           LayerCountMismatchException, BadContainerConfigurationFormatException {
     RegistryClient registryClient =
         buildConfiguration
-            .newRegistryClientFactory(buildConfiguration.getBaseImageConfiguration())
+            .newBaseImageRegistryClientFactory()
             .setAuthorization(registryAuthorization)
             .newRegistryClient();
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -191,6 +191,7 @@ class PullBaseImageStep
                 buildConfiguration.getBaseImageConfiguration().getImageRepository())
             .setAllowInsecureRegistries(buildConfiguration.getAllowInsecureRegistries())
             .setAuthorization(registryAuthorization)
+            .setUserAgentSuffix(buildConfiguration.getToolName())
             .newRegistryClient();
 
     ManifestTemplate manifestTemplate =
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -185,13 +185,9 @@ class PullBaseImageStep
       throws IOException, RegistryException, LayerPropertyNotFoundException,
           LayerCountMismatchException, BadContainerConfigurationFormatException {
     RegistryClient registryClient =
-        RegistryClient.factory(
-                buildConfiguration.getBuildLogger(),
-                buildConfiguration.getBaseImageConfiguration().getImageRegistry(),
-                buildConfiguration.getBaseImageConfiguration().getImageRepository())
-            .setAllowInsecureRegistries(buildConfiguration.getAllowInsecureRegistries())
+        buildConfiguration
+            .newRegistryClientFactory(buildConfiguration.getBaseImageConfiguration())
             .setAuthorization(registryAuthorization)
-            .setUserAgentSuffix(buildConfiguration.getToolName())
             .newRegistryClient();
 
     ManifestTemplate manifestTemplate =
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
@@ -304,7 +304,7 @@ public class BuildConfiguration {
   }
 
   /**
-   * Creates a new {@link RegistryClient.Factory} for the target image with fields from the build
+   * Creates a new {@link RegistryClient.Factory} for the base image with fields from the build
    * configuration.
    *
    * @return a new {@link RegistryClient.Factory}
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.jib.configuration;
 import com.google.cloud.tools.jib.JibLogger;
 import com.google.cloud.tools.jib.image.json.BuildableManifestTemplate;
 import com.google.cloud.tools.jib.image.json.V22ManifestTemplate;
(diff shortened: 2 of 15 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
```diff
@@ -38,9 +37,6 @@
 /** Configures and provides builders for the image building tasks. */
 class PluginConfigurationProcessor {
 
-  /** {@code User-Agent} header suffix to send to the registry. */
-  private static final String USER_AGENT_SUFFIX = "jib-gradle-plugin";
```
coollog: I would favor this constant over having the tool name be deep within the `processCommonConfiguration` method.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -191,6 +191,7 @@ public BaseImageWithAuthorization call()
                 buildConfiguration.getBaseImageConfiguration().getImageRepository())
             .setAllowInsecureRegistries(buildConfiguration.getAllowInsecureRegistries())
             .setAuthorization(registryAuthorization)
+            .setUserAgentSuffix(buildConfiguration.getToolName())
```
coollog: Hmm, it might be better to have a centralized location where this is set so that it doesn't need to be called for every time we create a new `RegistryClient`. I'm thinking perhaps something at the `BuildConfiguration` level (the class itself or some other class that uses it) could have a `newRegistryClientFactory` that generates a new `RegistryClient.Factory` with the user agent suffix (and other fields) pre-populated so that the only methods that need to be called are `setAuthorization` and `newRegistryClient`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/PluginConfigurationProcessor.java
```diff
@@ -38,8 +37,7 @@
 /** Configures and provides builders for the image building goals. */
 class PluginConfigurationProcessor {
 
-  /** {@code User-Agent} header suffix to send to the registry. */
-  private static final String USER_AGENT_SUFFIX = "jib-maven-plugin";
+  private static final String TOOL_NAME = "jib-maven-plugin";
```
coollog: Hmm, actually it seems a bit out-of-place to be in the `PluginConfigurationProcessor` - it's actually already defined in `MavenProjectProperties`.

TadCordle: In gradle, the project properties plugin name is just `jib`. Should we change it to `jib-gradle-plugin`?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -302,4 +302,20 @@ public boolean getAllowInsecureRegistries() {
...
   }
+
+  /**
+   * Creates a new {@link RegistryClient.Factory} and sets fields from the build configuration.
+   *
+   * @param imageConfiguration the {@link ImageConfiguration} to use for the server URL and image
+   *     name
+   * @return a new {@link RegistryClient.Factory}
+   */
+  public RegistryClient.Factory newRegistryClientFactory(ImageConfiguration imageConfiguration) {
```
briandealwis: It seems odd for BuildConfiguration, which is essentially a value object that encodes the intent, now knows how to create RegistryClients, but has to have part of itself passed back in.

TadCordle: How should we select which `ImageConfiguration` we're using to create the `RegistryClient`?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/PluginConfigurationProcessor.java
coollog: Oh right, so that was named `jib` because the configuration is applied to `jib` (the extension). Perhaps we should have a separate constant for the tool name then.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
TadCordle: >But RegistryClient already has a factory(), and I don't think another level of indirection is helpful.

In this case maybe RegistryClient should have a newFactory method that takes a build configuration?

TadCordle: Although maybe we don't want to require `BuildConfiguration` this deep down?

coollog: Yea, the `registry` package is definitely lower-level than the `configuration` package (which is only "seen" by the `builder` package and higher).

TadCordle: I ended up adding a separate method for the target image and base image; I'm not sure if the method fits on `ImageConfiguration` considering it pulls a few other values from `BuildConfiguration`.

coollog: Yea, if the method goes into `ImageConfiguration`, it shouldn't directly take a `BuildConfiguration` since `ImageConfiguration` should not depend on `BuildConfiguration`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -34,6 +34,7 @@
 /** Obtains information about a {@link MavenProject}. */
 class MavenProjectProperties implements ProjectProperties {
 
+  public static final String TOOL_NAME = "jib-maven-plugin";
```
coollog: I think it'd be good to add some javadocs for these constants now that there's 3 so it's immediately clear what each's purpose is.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -39,8 +39,13 @@
 /** Obtains information about a Gradle {@link Project} that uses Jib. */
 class GradleProjectProperties implements ProjectProperties {
 
-  public static final String TOOL_NAME = "jib-gradle-plugin";
+  /** Used to generate the User-Agent header and history metadata. */
+  static final String TOOL_NAME = "jib-gradle-plugin";
+
+  /** Used for logging during main class inference. */
```
coollog: Oh by purpose I meant like javadoc that says what the constant actually refers to. I think it'll be fine here, but in general, javadoc should be within the scope of the class and not refer to the actual actions of the caller/user of that class. For example, this constant refers to the name of the configuration object (extension) for Jib Gradle tasks, and the `JAR_PLUGIN_NAME` below is a description of where to configure the jar. 

If we were to have these be specifically used in main class reference, they should probably be grouped under a separate class for defining main class inference logging parameters.

PR #945, opened 2018-09-06
Lower verbosity of insecure registry failover

### LINKED ISSUE #778: Consider limiting warnings related to allowInsecureRegistries failover if too verbose
https://github.com/GoogleContainerTools/jib/pull/771#issuecomment-409943409

chanseokoh:
Fixes #778.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
@@ -169,7 +169,7 @@ class RegistryEndpointCaller<T> {
     }
 
     try {
-      logger.warn(
+      logger.info(
           "Cannot verify server at " + url + ". Attempting again with no TLS verification.");
       return call(url, getInsecureConnectionFactory());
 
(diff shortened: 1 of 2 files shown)


briandealwis:
Using INFO makes sense.  But I think the real problem with the logging is that it leaves me hanging: it's not clear whether the insecure-attempt worked!
```
[WARNING] Failed to connect to https://localhost:5000/v2/foo/blobs/sha256:077396f4e79fec38d66633aed03c94066e3ea8677fd7925b6c799ea676da2896 over HTTPS. Attempting again with HTTP: http://localhost:5000/v2/foo/blobs/sha256:077396f4e79fec38d66633aed03c94066e3ea8677fd7925b6c799ea676da2896
[WARNING] Cannot verify server at https://localhost:5000/v2/foo/blobs/uploads/. Attempting again with no TLS verification.
[WARNING] Failed to connect to https://localhost:5000/v2/foo/blobs/uploads/ over HTTPS. Attempting again with HTTP: http://localhost:5000/v2/foo/blobs/uploads/
[WARNING] Cannot verify server at https://localhost:5000/v2/foo/manifests/latest. Attempting again with no TLS verification.
[WARNING] Failed to connect to https://localhost:5000/v2/foo/manifests/latest over HTTPS. Attempting again with HTTP: http://localhost:5000/v2/foo/manifests/latest
```

chanseokoh:
> it's not clear whether the insecure-attempt worked!

Make sense, and it's easy to log success. However, I think it is quite verbose and somehow unnecessary to confirm the success. With INFO, most people will not see or pay attention to them, especially if there was no error. If there is an error, it will fail right away.

briandealwis:
I guess what would be better is that we figure this out once and then use that knowledge for all future requests, rather than rediscover the appropriate settings each time.

chanseokoh:
That's a viable optimization. I'll create a new issue.

PR #947, opened 2018-09-07
Adds Jib and JibContainerBuilder.

### LINKED ISSUE #881: Start implementing Jib and JibContainerBuilder.
`Jib` will be the main entrypoint for using Jib Core.

`Jib` API:
- `JibContainerBuilder from(String imageReference)`
- `JibContainerBuilder from(ImageReference imageReference)`

`JibContainerBuilder` configures container to build. The name is not final.

`JibContainerBuilder` API:

- `JibContainerBuilder addFiles(List<Path> files, String pathInContainer)`
- `JibContainerBuilder setLayers(List<LayerConfiguration>/LayerConfiguration...)`
- `JibContainerBuilder addLayer(LayerConfiguration layerConfiguration)`
- `JibContainerBuilder setEntrypoint(List<String>/String...)`
- `JibContainerBuilder setProgramArguments(List<String>/String...)`
- `JibContainerBuilder setEnvironment(Map<String, String> environmentMap)`
- `JibContainerBuilder setEnvironment(String name, String value)`
- `JibContainerBuilder setExposedPorts(List<Port>/Port... ports)`
- `JibContainerBuilder addExposedPort(int port[, Protocol protocol])`
- `JibContainerBuilder setLabels(Map<String, String> labelMap)`
- `JibContainerBuilder setLabel(String labelKey, String labelValue)`
- `JibContainer containerize(Containerizer containerizer)`
- For the future: `JibContainer containerize(Containerizer containerizer)`

coollog:
Fixes #881

This adds `Jib` and `JibContainerBuilder` to a new package `api`. This package is temporary and both classes will be moved to the root package `com.google.cloud.tools.jib` once they are fully ready and the other extraneous classes are removed from that root package.

This also changes massages the structure and javadocs of `LayerConfiguration` and `Port`.


jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+
+  /**
+   * Sets a variable in the container environment.
+   *
+   * @param name the environment variable name
+   * @param value the environment variable value
+   * @return this
+   * @see #setEnvironment for more details
+   */
+  public JibContainerBuilder setEnvironmentVariable(String name, String value) {
```
chanseokoh: `addEnvironmentVariable`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+
+  /**
+   * Sets the ports to expose from the container. Ports exposed will allow ingress traffic. This
+   * replaces any previously-set exposed ports.
+   *
+   * @param ports the list of ports to expose
+   * @return this
+   */
+  public JibContainerBuilder setExposedPorts(List<Integer> ports) {
+    this.ports = ports.stream().map(Port::tcp).collect(Collectors.toCollection(ArrayList::new));
```
chanseokoh: In general, when do you think you will check if given values are sane? At the beginning of `containerize()`? For example, I see a user can pass a negative port number here without an error or provide a non-Unix-style path for `pathInContainer` in `addFiles`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  /**
+   * Sets the ports to expose from the container. This replaces any previously-set exposed ports.
+   *
+   * @param ports the ports to expose
+   * @return this
+   * @see #setExposedPorts(List) for more details
+   */
+  public JibContainerBuilder setExposedPorts(int... ports) {
```
chanseokoh: Just curious if we won't care about UDP.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+
+  private List<LayerConfiguration> layerConfigurations = new ArrayList<>();
+  private Map<String, String> environment = new HashMap<>();
+  private List<Port> ports = new ArrayList<>();
+  private Map<String, String> labels = new HashMap<>();
+  @Nullable private ImmutableList<String> entrypoint;
+  @Nullable private ImmutableList<String> programArguments;
+
+  /** Instantiate with {@link Jib#from}. */
+  JibContainerBuilder(ImageReference baseImageReference) {
```
TadCordle: Should this be private or `@VisibleForTesting` if `Jib#from` is supposed to be used instead?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,259 @@
...
+   *   <li>{@code /path/in/container/directory/}
+   *   <li>{@code /path/in/container/directory/...} (all contents of {@code directory/})
+   * </ul>
+   *
+   * @param files the source files to copy to a new layer in the container
+   * @param pathInContainer the Unix-style path to copy the source files to in the container file
+   *     system
+   * @return this
+   */
+  public JibContainerBuilder addFiles(List<Path> files, String pathInContainer) {
```
TadCordle: Hmm maybe this should just be an overload of `addLayer` instead of being named differently?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
TadCordle: If the environment variable already exists, I think the naming is still accurate.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
TadCordle: Yeah, why not use the `Port` class here?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: For port, I don't think we need any validation since a negative port is still valid for an image's exposed ports. For the non-Unix-style path, I think we can add that to `LayerEntry` and throw a runtime exception if invalid.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: I was thinking of not exposing the `Port` class in this interface and keep everything as primitives. (Ideally, we wouldn't expose the `LayerConfiguration` class either.) But, that might just make the API too bloated for setting exposed ports. Will change to just using `Port`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: This is package-private for `Jib#from` to call it.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: That sounds good, condenses the method names.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: I think we should keep this consistent with `add/setLabel`. Should we change this to `add...` or change `addLabel` to `setLabel`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -38,11 +37,11 @@
...
  *    .addFiles(sourceFiles, extractionPath)
- *    .setEntrypoint(entrypoint)
- *    .setProgramArguments(arguments)
- *    .addEnvironmentVariable(key, value)
- *    .addExposedPort(
- *    .addLabel(key, value)
+ *    .setEntrypoint("myprogram", "--flag", "subcommand")
+ *    .setProgramArguments("hello", "world")
+ *    .addEnvironmentVariable("HOME", "/app")
+ *    .addExposedPort(8080)
```
TadCordle: You may need a different example here since you changed it to use the `Port` class.

coollog: Ah right

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,261 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Builds a container with Jib.
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * Jib.from(baseImage)
+ *    .addFiles(sourceFiles, extractionPath)
```
chanseokoh: Renamed: `addFiles()` --> `addLayer()`

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -34,7 +34,7 @@
     private Builder() {}
 
     /**
-     * Sets a label for this layer.
+     * Sets a label for this layer. This label does not affect the contents of the layer.
```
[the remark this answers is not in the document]
coollog: The label is for identifying which `LayerConfiguration` is which in the logging (like `Building classes layer...`). Once the eventing system is set, this label would not be necessary - the `LayerConfiguration` will be included in the event that is emitted.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,272 @@
...
+
+  /**
+   * Sets the container entrypoint. This is the beginning of the command that is run when the
+   * container starts. {@link #setProgramArguments} sets additional tokens.
+   *
+   * @param entrypointTokens a list of tokens for the entrypoint command
+   * @return this
+   */
+  public JibContainerBuilder setEntrypoint(List<String> entrypointTokens) {
+    entrypoint = ImmutableList.copyOf(entrypointTokens);
```
briandealwis: _token_ seems funny.  Did you choose it for a reason?  _component_?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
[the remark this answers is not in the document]
briandealwis: So the event consumers need to track this mapping?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -90,10 +114,20 @@ private LayerConfiguration(String label, ImmutableList<LayerEntry> layerEntries)
     this.layerEntries = layerEntries;
   }
 
+  /**
+   * Gets the label.
```
briandealwis: Would `name` be better to prevent confusion with container labels?

coollog: Yea that sounds good - or even `id`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
loosebazooka: I feel like `entrypoint` might be fine here?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/Port.java
```diff
@@ -37,34 +36,64 @@
...
-     * @return the {@link Protocol} with the corresponding name, or {@code Protocol#TCP} by default
-     *     if invalid
-     */
-    public static Protocol getFromString(@Nullable String protocolString) {
-      if (protocolString == null) {
-        return TCP;
-      }
-      return UDP.toString().equalsIgnoreCase(protocolString) ? UDP : TCP;
-    }
+  /**
```
loosebazooka: I like these additions :+1:

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: Okay will remove tokens

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/Port.java
```diff
@@ -83,9 +112,15 @@ public boolean equals(Object other) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(port, protocol);
+    return Objects.hash(port, protocol.toString());
```
TadCordle: Why the `toString` here? Just curious.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -38,6 +39,10 @@
   private final String extractionPath;
 
   public LayerEntry(ImmutableList<Path> sourceFiles, String extractionPath) {
```
TadCordle: Javadoc (maybe should mention that `extractionPath` is validated here)

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/Port.java
coollog: Since `protocol` is just an enum to better organize string constants, I thought it'd make more sense to hash the actual string rather than the enum constants.

TadCordle: It shouldn't affect the result either way so it's probably fine, however I feel like adding the `toString` makes it marginally less readable.

coollog: I think we can just make protocol a string.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -0,0 +1,277 @@
...
+   * @return this
+   * @see #setEntrypoint(List) for more details
+   */
+  public JibContainerBuilder setEntrypoint(String... entrypoint) {
+    return setEntrypoint(Arrays.asList(entrypoint));
+  }
+
+  /**
+   * Sets the container entrypoint program arguments. These are additional tokens added to the end
+   * of the entrypoint command.
```
TadCordle: I wonder if we should comment about how this corresponds to "Cmd" in the container config.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/Port.java
```diff
@@ -32,40 +31,65 @@
...
+   */
+  public static Port parseProtocol(int port, String protocolString) {
+    Protocol protocol =
+        Protocol.UDP.toString().equalsIgnoreCase(protocolString) ? Protocol.UDP : Protocol.TCP;
+    return new Port(port, protocol);
   }
 
   private final int port;
-  private final Protocol protocol;
+  private final String protocol;
```
TadCordle: If we're representing this as a string internally now I don't see how the enum is necessary anymore. Now `udp(int port)` and `tcp(int port)` might as well just `return new Port(port, "udp")` and `return new Port(port, "tcp")` respectively. It's especially weird in `parseProtocol` to be taking `protocolString`, converting it to an enum, passing that into the constructor, and immediately converting it back to a string.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Jib.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.api;
+// TODO: Move to com.google.cloud.tools.jib
```
loosebazooka: why not now?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
loosebazooka: Or maybe if this should just be cmd? Since that translates to both docker directly? But I guess kubernetes uses args?

coollog: Yea, `cmd` is a confusing naming that I think Kubernetes tried to move away from.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Jib.java
coollog: There are some other classes in `com.google.cloud.tools.jib` that shouldn't be there after when the library API is moved there. Once those are moved to their respective locations (or deleted), these `api` package classes should be moved there. This will help us make sure we don't forget to clean up the `com.google.cloud.tools.jib` package.

PR #952, opened 2018-09-07
Adds RegistryImage, DockerDaemonImage, and TarImage.

### LINKED ISSUE #948: Add RegistryImage, DockerDaemonImage, and TarImage.
Continuing from #881...

The three `TargetImage` types (`RegistryImage`, `DockerDaemonImage`, and `TarImage`) define the 3 different targets Jib can build to.

`RegistryImage` builds to a Docker registry. API:
- `static RegistryImage named(ImageReference/String imageReference)`
- `RegistryImage setCredential(String username, String password)`
- `RegistryImage addCredentialRetriever(CredentialRetriever credentialRetrievers)`

`DockerDaemonImage` build to the Docker daemon. API:
- `static DockerDaemonImage named(ImageReference/String imageReference)`
- `DockerDaemonImage setDockerExecutable(Path dockerExecutable)`

`TarImage` builds to a tarball file. API:
- `static TarImage.Builder named(ImageReference/String imageReference)`
`TarImage.Builder` constructs the `TarImage`. API:
- `TarImage saveTo(Path outputFile)`

coollog:
Fixes #948

Adds the 3 target image configurations (to a container registry, to a Docker daemon, and to a tar file).
Also changes `CredentialRetriever` to return `Optional` rather than `Nullable`.


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/credentials/Credential.java
```diff
@@ -22,18 +22,39 @@
...
 public class Credential {
 
+  /**
+   * Gets a {@link Credential} configured with a username and password.
+   *
+   * @param username the username
+   * @param password the password
+   * @return a new {@link Credential}
+   */
+  public static Credential basic(String username, String password) {
```
TadCordle: Are we going to be adding non-basic ways of constructing credentials?

coollog: Maybe - we could in the future maybe add token-based credentials, but I think basic is sufficient for now.

TadCordle: I'm just not sure about whether a static method here is better than a simple constructor. Just seems like an extra hurdle for someone unfamiliar with the API trying to figure out how to create a `Credential`. If we were adding more ways of creating credentials, then it maybe makes sense, but as it is right now I don't think it does.

TadCordle: Makes sense

jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistryImage.java
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Sets the credentials to use to push the image. This replaces any previously-added {@link
+   * CredentialRetriever}s.
+   *
+   * @param username the username
+   * @param password the password
+   * @return this
+   */
+  public RegistryImage setCredential(String username, String password) {
+    credentialRetrievers = new ArrayList<>();
```
TadCordle: It feels weird to have a `setCredential` that removes all other credentials in a builder. Would people have any reason to want to remove all credentials they've set, and if they do, would it be better to just explicitly have a `clearCredentials` method instead of overwriting with a new credential? Alternatively, should we have a `setCredential` that takes a generic `CredentialRetriever` instead of only allowing a username and password?

coollog: Hmm, that is a good point that having `setCredential` mean clear credential retrievers and use these credentials instead might be putting too much on one setter. Perhaps we should name this `useCredential`? The alternative would be to add a `CredentialRetriever` that returns the provided credential to the list of credential retrievers, but that might be confusing since there's already an `addCredentialRetriever`. We could have a `setCredentialRetriever` that replaces the list of credential retrievers with a single credential retriever, but I think `setCredentialRetrievers` to set a list would make more sense in that case. However, I steered away from setting full lists of credential retrievers that are built somewhere else since a list of credential retrievers is very specific to the `RegistryImage` they are for. Also, having this `(String username, String password)` signature is mostly meant to be a shorthand for users to easily provide a basic credential without needing to mess with `CredentialRetriever` or `Credential` classes.

So I think the options I'm considering are:
1. Rename to `useCredential`
1. Rename to `withCredential`
1. Rename to `addCredential` and change to add instead of replace

TadCordle: To me, option 3 makes the most sense; I can't think of a reason someone would want to add credentials and replace them later for the same registry.

loosebazooka: Yeah, there's a little bit of a disconnect between a single item and a list here. I would think `set...` mostly matches the underlying type (list). And `add` could augment the list.  I don't predict a lot of use from something like `clear` or `remove`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistryImage.java
```diff
@@ -0,0 +1,125 @@
...
+    credentialRetrievers.add(() -> Optional.of(Credential.basic(username, password)));
```
loosebazooka: Just wondering why Credential.basic would ever return null?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
```diff
@@ -66,15 +66,15 @@ static PluginConfigurationProcessor processCommonConfiguration(
     }
     DefaultCredentialRetrievers defaultCredentialRetrievers =
         DefaultCredentialRetrievers.init(CredentialRetrieverFactory.forImage(baseImage, logger));
-    Credential fromCredential =
+    Optional<Credential> optionFromCredential =
```
chanseokoh: `optionFrom...` --> `optionalFrom...`

jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistryImage.java
loosebazooka: Assuming the tool developer has to know when to use basic credentials, I don't see it being particularly expensive for a tool developer to use the slightly longer form `addCrendentials(Credentials.basic(asdf,asfd))` or whatever.

jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistryImage.java
coollog: `Credential.basic` would not return null, but `CredentialRetriever` can return an optional `Credential`.

PR #955, opened 2018-09-07
Adds Containerizer.

### LINKED ISSUE #949: Add Containerizer.
Continuing from #948 ...

`Containerizer` configures how and where to containerize to. It is passed into `JibContainerBuilder#containerize`. API:
- `static Containerizer to(RegistryImage registryImage)`
- `static Containerizer to(DockerDaemonImage dockerDaemonImage)`
- `static Containerizer to(TarImage tarImage)`
- `Containerizer setExecutorService(ExecutorService executorService)`
- `Containerizer setCacheConfiguration(CacheConfiguration cacheConfiguration)`
- `Containerizer setEventHandlers(EventHandlers eventHandlers)`

coollog:
Fixes #949


jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @param tarImage the {@link TarImage} that defines target output file
+   * @return a new {@link Containerizer}
+   */
+  public static Containerizer to(TarImage tarImage) {
+    return new Containerizer(tarImage);
+  }
+
+  private final ImageTarget imageTarget;
+  @Nullable private ExecutorService executorService;
```
loosebazooka: Could these have defaults and not be nullable (`ExecutorService` and `CacheConfiguration`)? Or am I missing something?

coollog: I think the default choice for these should belong in lower-level classes (like `BuildConfiguration` or some defaults class) rather than in the user-level `Containerizer` API. The nullness here just indicates that the user did not intentionally set anything and the underlying system should make its own choice.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -0,0 +1,116 @@
...
+    return new Containerizer(tarImage);
+  }
+
+  private final TargetImage targetImage;
+  @Nullable private ExecutorService executorService;
+  @Nullable private CacheConfiguration cacheConfiguration;
+  @Nullable private EventHandlers eventHandlers;
+
+  /** Instantiate with {@link #to}. */
+  private Containerizer(TargetImage targetImage) {
```
briandealwis: Isn't it really an image target?  (The image target will be provided the target image.)

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * @param executorService the {@link ExecutorService}
+   * @return this
+   */
+  public Containerizer setExecutorService(@Nullable ExecutorService executorService) {
+    this.executorService = executorService;
+    return this;
+  }
+
+  // TODO: Rethink this method.
```
briandealwis: Unclear as to what needs to be rethought :-)

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
coollog: Yea, I was unsure which was clearer, since for the base image `SourceImage` made more sense since it is the source image rather than the image source (which has the idea of where the image comes from instead). Filed an issue about the naming: #998

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
coollog: We would actually need two cache configurations - one for the base image layers cache and one for the application layers cache. This interface could remain the same and `CacheConfiguration` could be changed to take care of both, or this interface would need to have `setApplicationLayersCacheConfiguration` and `setBaseImageLayersCacheConfiguration`. Filed #999

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -0,0 +1,116 @@
...
+   * @return a new {@link Containerizer}
+   */
+  public static Containerizer to(TarImage tarImage) {
+    return new Containerizer(tarImage);
+  }
+
+  private final TargetImage targetImage;
+  @Nullable private ExecutorService executorService;
+  @Nullable private CacheConfiguration cacheConfiguration;
+  @Nullable private EventHandlers eventHandlers;
```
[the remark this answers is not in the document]
chanseokoh: Oh, just read another comment of the style choice. Please disregard.

Issue #864, opened 2018-08-21
Create a task/goal to display jib source files


loosebazooka:
This task goal should provide a list of files that represent project files. The idea is to identify which files are required to monitor to trigger a rebuild from an external system.
- [x] maven #958 
- [ ] gradle #1021 

Output should be in yaml for easy parsing. I would expect something along the lines of:

```
---
# jib-files
files:
  # a directory will include all files in it
  - /src/main/java 
  # a single file
  - /files/file.txt
---
```

in our case it might looks like this.
```
---
# jib-files
files:
  - /path/to/pom.xml
  - /path/to/build.gradle
  - /path/to/sources1
  - /path/to/sources2
  - /path/to/resources
  - /path/to/dependency1.jar
  - /path/to/dependency2.jar
  - /path/to/dependency3.jar
  - /path/to/dependency4.jar
  - /path/to/dependency5.jar
---
```

Ideally the task goal would display an output even when maven and gradle are run in "quiet" mode. I believe we might still need delimiters on the output incase some rogue plugin is spewing out stuff.

I think this can be run without any lifecycle phases or prerequisite tasks/goals:
```
$ ./gradlew jibFiles
$ ./mvnw jib:files
```

But I'm not sure if we need to force dependency resolution to truly obtain the path on disk for certain dependencies.

Another considerations is if we should include non-project dependencies in this list as they are expected to be stable and only manipulated by the build system itself.

PR #959, opened 2018-09-07
Add in _skaffold-files _skaffold-package-goals

### LINKED ISSUE #958: Add skaffold helpers to jib-maven-plugin
Add helpers so any integration with skaffold can extract the necessary information.

~~1. See if multimodule or not~~ (this can be done directly with `mvn help:evaluate -Dexpression=project.modules`)
2. Get all source files to watch (`jib:_skaffold-files`)
3. Get jib goals on package phase (more useful for multimodule builds - might be ignore for single module) (`jib:_skaffold-package-goals`)

### LINKED ISSUE #962: extra directory should resolve smarter for non-absolute paths
Currently extra directory is resolved to the execution directory, not the project path. This is fine unless you're in a multi module situation, in which case you want it to resolve to the project.baseDir.

For example a user may configure it one of two ways
```xml
<extraDirectory>src/main/custom-jib</extraDirectory>
<!-- OR -->
<extraDirectory>${project.baseDir}/src/main/custom-jib</extraDirectory>
```

solution is to do something like this:
```java
    Path extraDirectoryPath = Paths.get(extraDirectory);
    if (extraDirectoryPath.isAbsolute()) {
      return extraDirectoryPath;
    }
    else {
      return project.getBasedir().toPath().resolve(extraDirectory);
    }
```

This is how maven-resources appear to be resolved.

loosebazooka:
The Maven part of #864 .

Built in skaffold support goals:
1. _skaffold-files: a list of files to watch
2. _skaffold-package-goals: a list of jib goals bound to the
   package phase.

fixes #958 
fixes #962

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
@@ -301,11 +301,6 @@ abstract class JibPluginConfiguration extends AbstractMojo {
     return allowInsecureRegistries;
   }
 
-  /**
-   * Directory where non-java container files should be placed.
-   *
-   * @return path to extras directory, if not absolute then resolve against this project directory.
-   */
   Path getExtraDirectory() {
     // TODO: Should inform user about nonexistent directory if using custom directory.
     return Preconditions.checkNotNull(extraDirectory).toPath();
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
@@ -21,8 +21,8 @@ import com.google.cloud.tools.jib.plugins.common.AuthProperty;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import java.io.File;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -166,9 +166,10 @@ abstract class JibPluginConfiguration extends AbstractMojo {
   @Parameter(defaultValue = "false", required = true)
   private boolean allowInsecureRegistries;
 
+  // this parameter is cloned in FilesMojo
   @Nullable
   @Parameter(defaultValue = "${project.basedir}/src/main/jib", required = true)
-  private String extraDirectory;
+  private File extraDirectory;
 
   @Parameter(defaultValue = "false", property = "jib.skip")
   private boolean skip;
@@ -300,9 +301,14 @@ abstract class JibPluginConfiguration extends AbstractMojo {
     return allowInsecureRegistries;
   }
 
+  /**
+   * Directory where non-java container files should be placed.
+   *
+   * @return path to extras directory, if not absolute then resolve against this project directory.
+   */
   Path getExtraDirectory() {
     // TODO: Should inform user about nonexistent directory if using custom directory.
-    return Paths.get(Preconditions.checkNotNull(extraDirectory));
+    return Preconditions.checkNotNull(extraDirectory).toPath();
   }
 
   boolean isSkipped() {
@@ -339,7 +345,7 @@ abstract class JibPluginConfiguration extends AbstractMojo {
   }
 
   @VisibleForTesting
-  void setExtraDirectory(String extraDirectory) {
+  void setExtraDirectory(File extraDirectory) {
     this.extraDirectory = extraDirectory;
   }
 }
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
@@ -37,7 +37,7 @@ public class MavenProjectProperties implements ProjectProperties {
   /** Used for logging during main class inference and analysis of user configuration. */
   public static final String PLUGIN_NAME = "jib-maven-plugin";
 
-  /** Used to indentify this plugin when interacting with the maven system. */
+  /** Used to identify this plugin when interacting with the maven system. */
   public static final String PLUGIN_KEY = "com.google.cloud.tools:" + PLUGIN_NAME;
 
   /** Used to generate the User-Agent header and history metadata. */
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
@@ -34,8 +34,11 @@ import org.codehaus.plexus.util.xml.Xpp3Dom;
 /** Obtains information about a {@link MavenProject}. */
 public class MavenProjectProperties implements ProjectProperties {
 
-  /** Used for logging during main class inference and analysis of user configuration */
+  /** Used for logging during main class inference and analysis of user configuration. */
   public static final String PLUGIN_NAME = "jib-maven-plugin";
+
+  /** Used to indentify this plugin when interacting with the maven system. */
+  public static final String PLUGIN_KEY = "com.google.cloud.tools:" + PLUGIN_NAME;
 
   /** Used to generate the User-Agent header and history metadata. */
   static final String TOOL_NAME = "jib-maven-plugin";
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
@@ -32,13 +32,13 @@ import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
 
 /** Obtains information about a {@link MavenProject}. */
-class MavenProjectProperties implements ProjectProperties {
+public class MavenProjectProperties implements ProjectProperties {
+
+  /** Used for logging during main class inference and analysis of user configuration */
+  public static final String PLUGIN_NAME = "jib-maven-plugin";
 
   /** Used to generate the User-Agent header and history metadata. */
   static final String TOOL_NAME = "jib-maven-plugin";
-
-  /** Used for logging during main class inference. */
-  private static final String PLUGIN_NAME = "jib-maven-plugin";
 
   /** Used for logging during main class inference. */
   private static final String JAR_PLUGIN_NAME = "'maven-jar-plugin'";
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
@@ -110,7 +110,6 @@ public class FilesMojo extends AbstractMojo {
       System.out.println(extraDirectory.getAbsoluteFile().toPath());
     }
 
-
     // Grab non-project SNAPSHOT dependencies for this project
     // TODO: this whole sections relies on internal maven API, it could break. We need to explore
     // TODO: better ways to resolve dependencies using the public maven API.
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
@@ -93,17 +93,18 @@ public class FilesMojo extends AbstractMojo {
       return;
     }
 
-    // print out sources directory
+    // print out sources directory (resolved by maven to be an absolute path)
     System.out.println(project.getBuild().getSourceDirectory());
 
-    // print out resources directory
+    // print out resources directory (resolved by maven to be an absolute path)
     ImmutableSet.copyOf(project.getBuild().getResources())
         .stream()
         .map(FileSet::getDirectory)
         .forEach(System.out::println);
 
-    // this seems weird, but we will only print out the "jib" directory on projects where
-    // the plugin is explicitly configured (even though _skaffold-files is a jib-maven-plugin goal).
+    // this seems weird, but we will only print out the jib "extraFiles" directory on projects where
+    // the plugin is explicitly configured (even though _skaffold-files is a jib-maven-plugin goal
+    // and is expected to run on all projects irrespective of their configuring of the jib plugin).
     if (project.getPlugin(MavenProjectProperties.PLUGIN_KEY) != null) {
       // print out extra directory
       System.out.println(extraDirectory.getAbsoluteFile().toPath());
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
@@ -135,11 +135,11 @@ public class FilesMojo extends AbstractMojo {
         };
 
     try {
-      DependencyResolutionResult t =
+      DependencyResolutionResult resolutionResult =
           projectDependenciesResolver.resolve(
               new DefaultDependencyResolutionRequest(project, session.getRepositorySession())
                   .setResolutionFilter(ignoreProjectDependenciesFilter));
-      t.getDependencies()
+      resolutionResult.getDependencies()
           .stream()
           .map(org.eclipse.aether.graph.Dependency::getArtifact)
           .filter(org.eclipse.aether.artifact.Artifact::isSnapshot)
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
@@ -139,7 +139,8 @@ public class FilesMojo extends AbstractMojo {
           projectDependenciesResolver.resolve(
               new DefaultDependencyResolutionRequest(project, session.getRepositorySession())
                   .setResolutionFilter(ignoreProjectDependenciesFilter));
-      resolutionResult.getDependencies()
+      resolutionResult
+          .getDependencies()
           .stream()
           .map(org.eclipse.aether.graph.Dependency::getArtifact)
           .filter(org.eclipse.aether.artifact.Artifact::isSnapshot)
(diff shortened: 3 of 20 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/PackageGoalsMojo.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+/**
+ * Print out all jib goals tied to the package phase. Useful in multimodule situations to determine
+ * if the correct jib goals are configured when running skaffold.
+ *
+ * <p>It is intended to be used from the root project and only in multimodule situations:
+ *
+ * <p>./mvnw jib:_skaffold-package-goals -pl module [-Pprofile]
```
coollog: `-q` as well?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Test
+  public void testFilesMojo_singleModule() throws VerificationException, IOException {
+    Path projectRoot = simpleTestProject.getProjectRoot();
+
+    verifyGoals(
+        projectRoot,
+        null,
+        ImmutableList.of(projectRoot.resolve("pom.xml"), projectRoot.resolve("src/main/java")));
```
coollog: Might be good to add a resource file and just make sure `src/main/resources` shows up as well.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+/** Integration tests for {@link FilesMojo}. */
+public class FilesMojoIntegrationTest {
```
coollog: I think this can be just a unit test since it doesn't depend on any external resources?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoIntegrationTest.java
coollog: And also a snapshot dependency?

jib-maven-plugin/src/test/resources/projects/multi/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+    <module>lib</module>
+  </modules>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>com.google.cloud.tools</groupId>
+          <artifactId>jib-maven-plugin</artifactId>
+          <version>@@PluginVersion@@</version>
```
chanseokoh: Question: how is this injected?

jib-maven-plugin/src/test/resources/projects/multi/lib/src/test/java/com/lib/LibTest.java
```diff
@@ -0,0 +1,17 @@
+package com.lib;
```
chanseokoh: nit: copyright

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.apache.maven.project.MavenProject;
+
+/**
+ * Print out changing source dependencies on a module. In multimodule applications it should be run
+ * by activating a single module and it's dependent modules. Dependency collection will ignore
```
chanseokoh: "it's dependent" --> "its dependent"

jib-maven-plugin/src/test/resources/projects/multi/pom.xml
loosebazooka: It's in TestProject.java:72

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoIntegrationTest.java
loosebazooka: Yeah, probably.

loosebazooka:
Still need to add in a snapshot dependency. But code can be reviewed till then. This also fixes #962

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -300,9 +301,14 @@ boolean getAllowInsecureRegistries() {
     return allowInsecureRegistries;
   }
 
+  /**
+   * Directory where non-java container files should be placed.
+   *
+   * @return path to extras directory, if not absolute then resolve against this project directory.
```
coollog: wording change?
`path to the extra files directory. If the directory is not absolute, then it should have been resolved against the current project directory.`

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/TestPlugin.java
```diff
@@ -42,6 +42,8 @@ protected void before() throws IOException, XmlPullParserException, Verification
     Verifier verifier = new Verifier(".", true);
     verifier.setAutoclean(false);
     verifier.addCliOption("-DskipTests");
+    verifier.addCliOption("-Dfmt.skip");
```
coollog: Are these necessary?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
```diff
@@ -0,0 +1,102 @@
...
+    // print out sources directory
+    System.out.println(project.getBuild().getSourceDirectory());
+
+    // print out resources directory
+    ImmutableSet.copyOf(project.getBuild().getResources())
+        .stream()
+        .map(FileSet::getDirectory)
+        .forEach(System.out::println);
+
+    // this seems weird, but we will only print out the "jib" directory on projects where
```
coollog: print out the `src/main/jib` directory?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
```diff
@@ -0,0 +1,102 @@
...
+    // the plugin is explicitly configured (even though _skaffold-files is a jib-maven-plugin goal).
+    if (project.getPlugin(MavenProjectProperties.PLUGIN_KEY) != null) {
+      // print out extra directory
+      System.out.println(extraDirectory.getAbsoluteFile().toPath());
```
coollog: Should this only print if the directory actually exists?

coollog: Perhaps we should have a method that always prints the absolute path and use that method to make sure all the prints print out an absolute path? Like:

```java
private static void printFile(Path file) {
  System.out.println(file.toAbsolutePath());
}
```

loosebazooka: I believe users should be able to add things in there while skaffold is active and it will trigger a rebuild.

loosebazooka: I mean it's a question of ensuring we using printFile vs file.toAbsolutePath, but there's no way to enforce we're actually calling either. I'd rather not someone reading this have to go find out what "printFile" is doing. Absolute file path printing is ensure in the tests at the moment.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/TestPlugin.java
loosebazooka: They let you run the plugin dependent tests from the IDE faster without having to invoke fmt/checkstyle.  fmt/checkstyle will still run in the end. But theres no reason for us to double run this. It make the installs run faster as well.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
loosebazooka: looks like I don't need this comment at all. I change this section of code.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -32,14 +32,17 @@
 import org.codehaus.plexus.util.xml.Xpp3Dom;
 
 /** Obtains information about a {@link MavenProject}. */
-class MavenProjectProperties implements ProjectProperties {
+public class MavenProjectProperties implements ProjectProperties {
+
+  /** Used for logging during main class inference and analysis of user configuration. */
+  public static final String PLUGIN_NAME = "jib-maven-plugin";
+
+  /** Used to indentify this plugin when interacting with the maven system. */
```
coollog: identify

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+/** Integration tests for {@link FilesMojo}. */
```
coollog: Just `Tests`

jib-maven-plugin/src/test/resources/projects/multi/complex-service/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>com.jib.test</groupId>
+      <artifactId>lib</artifactId>
+      <version>1.0.0.TEST-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.tools</groupId>
+      <artifactId>appengine-plugins-core</artifactId>
```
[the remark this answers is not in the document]
loosebazooka: Oh oops. This shouldn't have made it in.

jib-maven-plugin/src/test/resources/projects/multi/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+    <module>lib</module>
+  </modules>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>com.google.cloud.tools</groupId>
+          <artifactId>jib-maven-plugin</artifactId>
+          <version>0.9.11-SNAPSHOT</version>
```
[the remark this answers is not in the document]
loosebazooka: No. I need to remove this. I'm getting messy.

briandealwis: `@@PluginVersion@@`?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
```diff
@@ -0,0 +1,153 @@
...
+                  .setResolutionFilter(ignoreProjectDependenciesFilter));
+      resolutionResult.getDependencies()
+          .stream()
+          .map(org.eclipse.aether.graph.Dependency::getArtifact)
+          .filter(org.eclipse.aether.artifact.Artifact::isSnapshot)
+          .map(org.eclipse.aether.artifact.Artifact::getFile)
+          .forEach(System.out::println);
+
+    } catch (DependencyResolutionException ex) {
+      throw new MojoExecutionException("Failed to resolve dependencies", ex);
```
briandealwis: Is this really the right thing to do?  If I specify the wrong version, this is what I see:

```
$ mvn -q jib:_skaffold-files
/Users/bsd/Manumitting/Projects/GooglePluginEclipse/jibplayground/maven/app/pom.xml
/Users/bsd/Manumitting/Projects/GooglePluginEclipse/jibplayground/maven/app/src/main/java
/Users/bsd/Manumitting/Projects/GooglePluginEclipse/jibplayground/maven/app/src/main/resources
/Users/bsd/Manumitting/Projects/GooglePluginEclipse/jibplayground/maven/app/src/main/jib
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:0.9.11-SNAPSHOT:_skaffold-files (default-cli) on project first: Failed to resolve dependencies: Could not resolve dependencies for project ca.mt.jib.examples:first:jar:0.0.1-SNAPSHOT: Failure to find com.google.cloud.tools:appengine-plugins-core:jar:0.6.0-SNAPSHOT in http://nexus.mt.ca:8081/nexus/content/groups/public was cached in the local repository, resolution will not be reattempted until the update interval of nexus has elapsed or updates are forced -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```
(Note: the Maven error logging goes to stdout too.)

Should Skaffold ignore the results with a non-zero exit code and retry after some timeout?

loosebazooka: I think the error result should actually result in an error. I'm also wondering if I should change this to just calculate the correct filename from the artifact coordinates, instead of explicitly resolving (downloading) everything.

loosebazooka: Oh yeah, we shouldn't be "stopping" skaffold during a dev run. But we have to surface those errors to the users.

That error is also a little weird, I believe us using ProjectDependencyResolver in a weird way is providing a result that doesn't make full sense.

r2d4:
We don't treat the build files from the the build dependencies, and just recompute every cycle

e.g. Dockerfile changes - need to recompute dependencies

e.g. Dockerfile instruction `COPY . .` 
file is added to the project, that changes the glob pattern so the build file needs to be recomputed anyways.

loosebazooka:
@r2d4, just to clarify, if `src/.../myfile.java` changes would you recompute the build?

r2d4:
How the dependencies are parsed for each artifact
Main switch

https://github.com/GoogleContainerTools/skaffold/blob/24aefd21060f3c7e32f8c82ac959e67835c6c997/pkg/skaffold/runner/runner.go#L359-L363

Docker artifact implementation
https://github.com/GoogleContainerTools/skaffold/blob/abe80329cf875d78257c654d177f297d1a9cf15d/pkg/skaffold/docker/parse.go#L222

Bazel artifact implementation
https://github.com/GoogleContainerTools/skaffold/blob/426e56e1a6b70c7942e734d76db42240d7cc7c04/pkg/skaffold/bazel/bazel.go#L39

Jib implemention

```golang
func GetDependencies(workspace string, artifact *v1alpha2.JibArtifact) []string {
   // jib artifact might have a field that take in the a list of globs that are dependencies
   // it could just expand those globs and return the list of files
   return filepath.Match(artifact.JibFiles)
}
```

loosebazooka:
Thanks @r2d4, I think we should be good to go then?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -166,9 +166,10 @@ public void set(String image) {
   @Parameter(defaultValue = "false", required = true)
   private boolean allowInsecureRegistries;
 
+  // this parameter is cloned in FilesMojo
   @Nullable
   @Parameter(defaultValue = "${project.basedir}/src/main/jib", required = true)
-  private String extraDirectory;
+  private File extraDirectory;
```
TadCordle: I feel like we used to do this and I had to change it to a String to get rid of a warning, is that not the case anymore?

loosebazooka: oh, was it some errorprone or nullaway warning? 
Let me make sure this still works even if the directory doesn't exist.

TadCordle: Oh actually it might have been a Gradle warning, not Maven.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
```diff
@@ -0,0 +1,154 @@
...
+            .collect(Collectors.toSet());
+
+    DependencyFilter ignoreProjectDependenciesFilter =
+        (node, parents) -> {
+          if (node == null || node.getDependency() == null) {
+            // if nothing, then ignore
+            return false;
+          }
+          if (projectArtifacts.contains(node.getArtifact().toString())) {
+            // ignore project dependency artifacts, ignore
```
TadCordle: ignore project dependency artifacts ~~, ignore~~

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoTest.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+/** Tests for {@link FilesMojo}. */
+public class FilesMojoTest {
```
TadCordle: Noticed the thread earlier, but I'm not sure I understand the decision to make this a unit test and not an integration test.

coollog: I think the convention we use for deciding between unit vs. integration test is to favor unit tests unless:

1) The test depends on external resources that not every machine that runs the test has (like `docker`, registry credentials, or access to the Internet), and
2) Lesser so, the test just takes way too long.

PR #960, opened 2018-09-08
Changes LayerEntry to be more like TarArchiveEntry.

### LINKED ISSUE #956: Change LayerEntry to be more like TarArchiveEntry.
A `LayerEntry` should be a tuple of (`Path`, `extractionPath`) rather than (`List<Path>`, `extractionPath`). `extractionPath` will be the exact location of the file in the container file system rather than the directory to extract the source file to. This way, a `LayerConfiguration` becomes a list of `LayerEntry`s, each with its own source file and target path in the container file system.

The benefits of these are:
- Simplifies `LayerConfiguration` to not be a list of lists
- Does not hide the directory-walking logic within `ReproducibleLayerBuilder` - each file must be explicitly added
- Allows files to be renamed (to fix #810)
- Allow adding only select files in a directory (to fix #957)

Fixes #957

coollog:
Fixes #956

`LayerEntry` is now a tuple of (`Path`, `extractionPath`) rather than (`List<Path>`, `extractionPath`). This means that `LayerConfiguration` is a list of `LayerEntry`s, each referring to a single source file, whereas before, each `LayerEntry` referred to a list of source files all extracting to the same `extractionPath`.

Also changes `MainClassFinder` to take a list of class files instead of walking directories possibly containing class files.

Also makes changes to `JavaDockerContextGenerator` that leads to `FileOperations` being able to be removed.

TODO: Consider changing `extractionPath` to be `ContainerPath` class that always serializes in Unix form.


jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheMetadata.java
```diff
@@ -78,31 +75,24 @@ static boolean doLayerEntriesMatchMetadataEntries(
...
+        LayerMetadata.LayerMetadataEntry metadataEntry = metadataEntriesIterator.next();
+
+        boolean areSourceFilesEqual =
+            layerEntry.getSourceFileString().equals(metadataEntry.getSourceFilesString());
+        boolean areExtractionPathsEqual =
+            layerEntry.getExtractionPathString().equals(metadataEntry.getExtractionPathString());
+        if (!areSourceFilesEqual || !areExtractionPathsEqual) {
+          return false;
+        }
+      }
```
chanseokoh: nit: I think below will work, but not saying I want this code to be like that. Just commenting.
```java
Streams.zip(
        layerEntries.stream().map(LayerEntry::getSourceFilesString),
        metadataEntries.stream().map(LayerMetadataEntry::getSourceFilesString),
        String::equals)
   .reduce(true, Boolean::logicalAnd);
```

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheMetadataTranslator.java
```diff
@@ -105,17 +101,14 @@ static CacheMetadataTemplate toTemplate(CacheMetadata cacheMetadata) {
...
-                            layerMetadataEntry.getExtractionPath()))
-                .collect(Collectors.toList());
+        ImmutableList<LayerMetadata.LayerMetadataEntry> metadataEntries =
+            cachedLayerWithMetadata.getMetadata().getEntries();
+        List<LayerEntryTemplate> layerEntryTemplates = new ArrayList<>(metadataEntries.size());
+        for (LayerMetadata.LayerMetadataEntry metadataEntry : metadataEntries) {
+          layerEntryTemplates.add(
+              new LayerEntryTemplate(
+                  metadataEntry.getSourceFilesString(), metadataEntry.getExtractionPathString()));
+        }
```
chanseokoh: Preferring non-stream style?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/json/CacheMetadataLayerPropertiesObjectTemplate.java
```diff
@@ -31,27 +31,28 @@
...
     @Nullable
-    public List<String> getSourceFiles() {
-      return sourceFiles;
+    public String getSourceFileString() {
+      return sourceFile;
     }
 
     @Nullable
-    public String getExtractionPath() {
+    public String getExtractionPathString() {
```
chanseokoh: I'm fine without the `String` suffix. The type says it. But up to you.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -45,20 +47,47 @@ public Builder setLabel(String label) {
...
-     * they are added.
+     * <p>For example, {@code addEntry(Paths.get("myfile"), Paths.get("/path/in/container"))} would
+     * add {@code myfile} to the container to be accessed at {@code /path/in/container}.
      *
-     * @param sourceFiles the source files to build from. Source files that are directories will
-     *     have all subfiles in the directory added (but not the directory itself)
-     * @param destinationOnImage Unix-style path to add the source files to in the container image
-     *     filesystem
+     * @param sourceFile the source file to add to the layer
+     * @param pathInContainer the path to add the source files to in the container file system
```
chanseokoh: "the path of the source file (not files) in the..."?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -74,31 +74,32 @@
...
+      String directoryInContext,
+      String extractionPath) {
+    if (layerEntries.isEmpty()) {
       return;
     }
 
     listBuilder.add(
-        new CopyDirective(
-            layerEntry.getSourceFiles(), directoryInContext, layerEntry.getExtractionPath()));
+        new CopyDirective(ImmutableList.copyOf(layerEntries), directoryInContext, extractionPath));
```
chanseokoh: `getDependencylayerEntries` is already returning `ImmutableList`, so we could have `addIfNotEmtpy` accept `ImmutableList<LayerEntry>`, and we won't need to copy?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -259,8 +265,18 @@ public void generate(Path targetDirectory) throws IOException {
       Path directoryInContext = targetDirectory.resolve(copyDirective.directoryInContext);
       Files.createDirectory(directoryInContext);
 
-      // Copies dependencies.
-      FileOperations.copy(copyDirective.sourceFiles, directoryInContext);
+      // Copies the source files.
+      for (LayerEntry layerEntry : copyDirective.layerEntries) {
+        Path destination =
+            directoryInContext.resolve(
+                Paths.get(copyDirective.extractionPath).relativize(layerEntry.getExtractionPath()));
```
chanseokoh: I'm a bit confused. Shouldn't this be `directoryInContext.resolve(copyDirective.extractionPath).resolve(layerEntry.getExtractionPath())`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -16,37 +16,92 @@
...
 public class LayerEntry {
 
+  private final Path sourceFile;
+  private final Path extractionPath;
+
   /**
-   * The source files to build from. Source files that are directories will have all subfiles in the
-   * directory added (but not the directory itself).
+   * Instantiates with a source file and the path to place the source file in the container file
+   * system.
```
chanseokoh: It'd be nice to eliminate any ambiguity with an example. I believe it is not like `extractionPath` is a directory and the file will be placed inside `extractionPath`, put rather `extractionPath` is the final path of the file, right?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -16,37 +16,92 @@
...
    *
-   * <p>The source files are specified as a list instead of a set to define the order in which they
-   * are added.
+   * @param sourceFile the source file to add to the layer
+   * @param extractionPath the path to place the source file in the container file system (relative
```
chanseokoh: Does this mean `extractionPath` shouldn't start with `/`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -16,37 +16,92 @@
...
-    this.sourceFiles = sourceFiles;
+  public LayerEntry(Path sourceFile, Path extractionPath) {
+    this.sourceFile = sourceFile;
     this.extractionPath = extractionPath;
   }
 
-  public ImmutableList<Path> getSourceFiles() {
-    return sourceFiles;
+  /**
+   * Gets the source file. Do <b>not</b> call {@link Path#toString} on this - use {@link
```
chanseokoh: Curious. Will it be wrong to call `toString()`? Can `getSourceFile().toString()` and `getSourceFileString()` be different? The emphasis on "not" gives me the impression that they can be different.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheReaderTest.java
```diff
@@ -60,6 +61,8 @@ public void setUp() throws IOException, URISyntaxException {
 
     // Copies the test resource cache to the temporary test cache folder.
     Path resourceCache = Paths.get(Resources.getResource("cache").toURI());
+    // TODO: Uncomment
+    // FileOperations.copy(ImmutableList.of(resourceCache), testCacheFolder);
```
chanseokoh: Can you update the DOTO why it should be uncommented for what?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java
```diff
@@ -112,7 +113,13 @@ private static String findMainClassInClassFiles(ProjectProperties projectPropert
...
-        projectProperties.getJavaLayerConfigurations().getClassesLayerEntry().getSourceFiles();
+        projectProperties
+            .getJavaLayerConfigurations()
+            .getClassLayerEntries()
+            .stream()
+            .peek(layerEntry -> Verify.verify(layerEntry != null))
+            .map(LayerEntry::getSourceFile)
+            .collect(ImmutableList.toImmutableList());
 
     MainClassFinder.Result mainClassFinderResult =
```
briandealwis: I'm not a fan of this result-style.  It would make more sense to be to turn the `isSuccess()` into a result-type, and rename `getErrorType()` -> `getResultType()`.  Then your switch would just be the different result types.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheMetadata.java
coollog: Yea, I was trying to prefer using for-loops in the computationally-intensive sections (lots of computations itself and/or called many times) since streams come with a bit of overhead. For non-computationally-intensive sections, though, streams can definitely be nice to make the code flow better.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/json/CacheMetadataLayerPropertiesObjectTemplate.java
coollog: I'm mostly trying to keep the method naming for the `LayerEntry` the same as this template object - so `getExtractionPath` would be the path form and `getExtractionPathString` would be the string form.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheMetadata.java
chanseokoh: Too bad, since the inception of Java, Java has always been slow, and we should still care about that.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
coollog: `destination` is the location to copy the `layerEntry.sourceFile` to, so it should be resolved to the directory relative to the `directoryInContext`. That relative path is the path from the base extraction path for that `copyDirective` (`copyDirective.extractionPath`) to the actual extraction path for that source file (`layerEntry.extractionPath`).

For example, a `copyDirective` has:
- `directoryInContext = 'classes'`
- `extractionPath = '/app/classes'`

Then, with a context `targetDirectory = 'jib-docker-context'`, `destination` for a `layerEntry.extractionPath = '/app/classes/com/test/HelloWorld'` would be `jib-docker-context/classes/com/test/HelloWorld`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
coollog: Yes, it is the final path. Will add an example.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
coollog: `extractionPath` can start with `/` or not - it will be normalized when building the actual tarball archive.

coollog:
@briandealwis 

For `Optional`, I think we can start with the changes here, but I'll file an issue to finish the effort for the rest of the codebase.

> One thing that this PR changes is that we won't detect if a new file has been added. I'm ok with that.
I believe that this PR doesn't change the old behavior and will detect if a new file has been added. If a new file is added, the `LayerEntry` will be generated for that file, which will try to match against metadata layer entries via `CacheReader#getUpToDateLayerByLayerEntries`, which will fail and build the new layer with the new file.

> I wonder if it's worth using a custom `FileSystem` for the container's file system so we can ensure consistent cross-filesystem behaviour. (Not required in this PR.)
I'll file a new issue for this as well.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -42,12 +44,12 @@
...
 
     private final String label;
-    private final String extractionPath;
+    private final Path extractionPath;
 
     /** Initializes with a label for the layer and the layer files' default extraction path root. */
     LayerType(String label, String extractionPath) {
       this.label = label;
-      this.extractionPath = extractionPath;
+      this.extractionPath = Paths.get(extractionPath);
```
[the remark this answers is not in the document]
coollog: It's mostly since this is just a constructor for the enum and pushing it to the caller would require putting `Paths.get` multiple times in the enum list above.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
```diff
@@ -32,20 +30,21 @@
   /** Entry into the layer metadata. */
   static class LayerMetadataEntry {
 
-    private ImmutableList<String> sourceFilesStrings;
-    private String extractionPath;
+    private final String sourceFilesString;
```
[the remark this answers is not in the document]
coollog: Hmm, I'm thinking perhaps it might be good to have an `ExtractionPath` class that handles non-OS-specific representation of an `extractionPath`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -17,25 +17,21 @@
...
 import java.io.InputStream;
 import java.lang.reflect.Modifier;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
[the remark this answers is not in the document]
coollog: Ah was coding on a new IDE instance, forgot to set this up.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheMetadataTranslatorTest.java
```diff
@@ -110,11 +108,11 @@ public void testFromTemplate()
...
     Assert.assertEquals(
-        CLASSES_LAYER_SOURCE_FILES,
-        classesLayer.getMetadata().getEntries().get(0).getSourceFilesStrings());
+        CLASSES_LAYER_SOURCE_FILE.toString(),
+        classesLayer.getMetadata().getEntries().get(0).getSourceFilesString());
     Assert.assertEquals(
-        CLASSES_LAYER_EXTRACTION_PATH,
-        classesLayer.getMetadata().getEntries().get(0).getExtractionPath());
+        CLASSES_LAYER_EXTRACTION_PATH.toString(),
+        classesLayer.getMetadata().getEntries().get(0).getExtractionPathString());
```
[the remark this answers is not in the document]
coollog: Yea, we will need to change this serialization if the serialization method in `getExtractionPathString` changes.

jib-core/src/test/resources/json/metadata_windows-v3.json
```diff
@@ -0,0 +1 @@
+{"layers":[{"reference":{"size":631,"digest":"sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef","diffId":"sha256:b56ae66c29370df48e7377c8f9baa744a3958058a766793f821dadcb144a4647"}},{"reference":{"size":223,"digest":"sha256:8c662931926fa990b41da3c9f42663a537ccd498130030f9149173a0493832ad","diffId":"sha256:a3f3e99c29370df48e7377c8f9baa744a3958058a766793f821dadcb144a8372"},"properties":{"layerEntries":[{"sourceFile":"\\some\\source\\path","extractionPath":"some\\extraction\\path"}],"lastModifiedTime":255073580723571}}]}
```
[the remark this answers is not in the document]
coollog: That is true, I'm leaning more towards an `ExtractionPath` class now. Or it could be named like `ContainerPath`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java
```diff
@@ -112,7 +113,13 @@ private static String findMainClassInClassFiles(ProjectProperties projectPropert
                 + "; attempting to infer main class.");
 
     ImmutableList<Path> classesSourceFiles =
-        projectProperties.getJavaLayerConfigurations().getClassesLayerEntry().getSourceFiles();
+        projectProperties
+            .getJavaLayerConfigurations()
+            .getClassLayerEntries()
+            .stream()
+            .peek(layerEntry -> Verify.verify(layerEntry != null))
```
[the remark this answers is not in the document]
coollog: Ah oops this was left over from testing.

jib-core/src/test/resources/cache/metadata-v3.json
```diff
@@ -0,0 +1,34 @@
...
+    {
+      "reference": {
+        "size": 0,
+        "digest": "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        "diffId": "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+      },
+      "properties": {
+        "layerEntries": [
+          {
+            "sourceFile": "/some/source/directory",
```
chanseokoh: Just got this question. By just looking at the medata file, it's not possible to determine is this file should be a directory or a cache. So is it possible that Jib can be confused by this? For example, the cache saved is actually for a file, but someone was querying with the path intended to be a directory.

chanseokoh: > to determine is this path should be a directory or a cache

Sorry, I meant "to determine if this file should be a directory or a file."

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
```diff
@@ -322,6 +323,13 @@ public void testDockerContext() throws IOException, InterruptedException {
                 .toString())
         .run();
 
+    new DirectoryWalker(
+            simpleTestProject.getProjectRoot().resolve("build").resolve("jib-docker-context"))
+        .walk(
+            path -> {
```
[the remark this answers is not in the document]
coollog: Oops, also left over from testing.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -259,8 +264,23 @@ public void generate(Path targetDirectory) throws IOException {
...
 
-      // Copies dependencies.
-      FileOperations.copy(copyDirective.sourceFiles, directoryInContext);
+      // Copies the source files to the directoryInContext.
+      for (LayerEntry layerEntry : copyDirective.layerEntries) {
+        // This resolves the path to copy the source file to in the {@code directory}.
+        // For example, for a 'baseDirectory' of 'target/jib-docker-context', a 'baseExtractionPath'
+        // of '/app/classes', and an 'actualExtractionPath' of
+        // '/app/classes/com/test/HelloWorld.class', the resolved destination would be
+        // 'target/jib-docker-context/com/test/HelloWorld.class'.
```
chanseokoh: Something feels not right, or the example is not a good one. I know in reality the file should be at `target/jib-docker-context/classes/com/test/HelloWorld.class' (not directly into the root of `target/jib-docker-context`.

coollog: Ah right, I messed up the example - should be using `jib-docker-context/classes` to be more like the real deal.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -45,20 +47,47 @@ public Builder setLabel(String label) {
...
-     * they are added.
+     * <p>For example, {@code addEntry(Paths.get("myfile"), Paths.get("/path/in/container"))} would
+     * add {@code myfile} to the container to be accessed at {@code /path/in/container}.
      *
-     * @param sourceFiles the source files to build from. Source files that are directories will
-     *     have all subfiles in the directory added (but not the directory itself)
-     * @param destinationOnImage Unix-style path to add the source files to in the container image
-     *     filesystem
+     * @param sourceFile the source file to add to the layer
+     * @param pathInContainer the path to add the source file to in the container file system
```
briandealwis: Is it the path to correspond to the source file (that is, the actual name) or the path to the directory to add the source file to?

_The path in the container file system corresponding to the {@code sourceFile}_?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -45,20 +47,47 @@ public Builder setLabel(String label) {
...
 
     /**
-     * Adds an entry to the layer.
+     * Adds an entry to the layer. Only adds the single source file to the exact path in the
+     * container file system.
      *
-     * <p>The source files are specified as a list instead of a set to define the order in which
-     * they are added.
+     * <p>For example, {@code addEntry(Paths.get("myfile"), Paths.get("/path/in/container"))} would
+     * add {@code myfile} to the container to be accessed at {@code /path/in/container}.
```
briandealwis: This language is still ambiguous.  I think you mean _to be accessed **as**_ (not **at**).

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -54,7 +54,7 @@ public Builder setLabel(String label) {
      * add {@code myfile} to the container to be accessed at {@code /path/in/container}.
      *
      * @param sourceFile the source file to add to the layer
-     * @param pathInContainer the path to add the source files to in the container file system
+     * @param pathInContainer the path to add the source file to in the container file system
```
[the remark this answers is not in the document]
coollog: This can be either a directory or a file, but either would be added to that exact path. Directory contents would not be added.

`hold` the file might sound like it is a directory containing the file. Maybe like `the exact path for the file in the container file system`?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -45,20 +47,47 @@ public Builder setLabel(String label) {
     }
 
     /**
-     * Adds an entry to the layer.
+     * Adds an entry to the layer. Only adds the single source file to the exact path in the
```
briandealwis: The _file_ in `sourceFile` somewhat implies it shouldn't be a directory. _sourcePath_?  And add a note that the contents of a directory are not added.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -45,20 +47,47 @@ public Builder setLabel(String label) {
...
+      return this;
+    }
+
+    /**
+     * Adds an entry to the layer. If the source file is a directory, the directory and its contents
+     * will be added recursively.
+     *
+     * <p>For example, {@code addEntryRecursive(Paths.get("mydirectory"),
+     * Paths.get("/path/in/container"))} would add {@code mydirectory} to the container to be
+     * accessed at {@code /path/in/container} and the contents of {@code mydirectory} to be accessed
```
briandealwis: s/at/as/

Could I suggest:
> For example, {@code addEntryRecursive(Paths.get("mydirectory"), Paths.get("/path/in/container"))} would add {@code mydirectory} to the container to be accessed as {@code /path/in/container} and  {@code mydirectory/fileA} to be accessed as {@code /path/in/container/fileA}.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -135,104 +109,88 @@ public Throwable getErrorCause() {
     }
   }
 
-  private final ImmutableList<Path> classesFiles;
+  private final ImmutableList<Path> files;
   private final JibLogger buildLogger;
 
   /**
-   * Finds a class with {@code psvm} in {@code classesFiles}.
+   * Finds a class with {@code psvm} in {@code files}.
```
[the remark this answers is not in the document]
briandealwis: (I thought it referenced some VM bytecode.)

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -16,37 +16,102 @@
 
 package com.google.cloud.tools.jib.image;
 
-import com.google.common.collect.ImmutableList;
+import com.google.common.annotations.VisibleForTesting;
 import java.nio.file.Path;
+import java.util.Objects;
 
 /**
  * Represents an entry in the layer. A layer consists of many entries that can be converted into tar
```
briandealwis: I think it's worth mentioning that (1) entries can be files or directories, and (2) a directory is copied as is: its contents are not copied.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -16,37 +16,102 @@
...
  */
 public class LayerEntry {
 
+  private final Path sourceFile;
+  private final Path extractionPath;
+
   /**
-   * The source files to build from. Source files that are directories will have all subfiles in the
-   * directory added (but not the directory itself).
+   * Instantiates with a source file and the path to place the source file in the container file
```
briandealwis: _source file_ here reads like it only supports files.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -16,37 +16,102 @@
...
+   * system. For example, adding a file {@code HelloWorld.class} to be in container file system at
+   * {@code /app/classes/HelloWorld.class} would be {@code new
+   * LayerEntry(Paths.get("HelloWorld.class"), Paths.get("/app/classes/HelloWorld.class"))}.
```
briandealwis: You could add an example here of `com` &rarr; `/app/classes/com` but not its contents.

coollog:
@briandealwis The more harping the better :)

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
coollog: I've been using file to mean directories and files in this codebase. Do you think we should have file only mean regular files for the Jib Core API?

coollog: Did you mean `LayerEntry::new`? Yes that should be explaining the same thing, although `LayerEntry` is intended to be an internal class (no user-facing API accepts a `LayerEntry`) whereas `LayerConfiguration` is a user-facing API. I'll change the javadocs to be more aligned.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -23,6 +23,14 @@
 /**
  * Represents an entry in the layer. A layer consists of many entries that can be converted into tar
  * archive entries.
+ *
+ * <p>Note that:
+ *
+ * <ul>
+ *   <li>Entry source files can be either files or directories.
+ *   <li>A directory is a single tar archive entry - its contents are not added. To add the contents
+ *       of a directory, each subfile should have be its own {@link LayerEntry}.
```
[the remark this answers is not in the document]
coollog: That is a lot clearer.

Issue #802, opened 2018-08-06
Allow pushing multiple tags to registries


Techtony96:
**Description of the feature**:
Currently, in order to push multiple tags (of the same image) you must call jib multiple times. This is common with docker where you have the latest tag and version specific tags. I suggest there be a way to specify multiple tags within jib's configuration.

**Expected behavior**:
```
jib {
    to {
        image = ["repo/name:tag", "repo/name:tag2", "repo2/name2:tag3"]
    }
}
```

Jib would handle this by building the container once and then (preferably in parallel) push the container with the given tags.

chanseokoh:
We should decided if we would allow multiple tags only or pushing to multiple registries. In the latter case, we may need to think about how auth info could be specified in the build files or through the system properties.

loosebazooka:
@Techtony96, while personally I think deployment configs should use a specific version instead of `latest`, you can kind of do this by doing multiple builds and using project properties. If you're not doing that already, it's fairly simple, both gradle and jib are cached enough that this should be pretty fast, the second push especially shouldn't take more than a couple seconds.

```
jib {
  to {
    image = "registry/repo/image:${project.tag}"
  }
}
```
```
$./gradlew jib -Ptag=1.0.0 && ./gradlew jib -Ptag=latest
```

Techtony96:
@loosebazooka I do something very similar to that, shown here: https://github.com/DiscordBolt/BoltBot/blob/cacbd52103893ad629c00ff1b066bce7c24a25db/build.gradle#L95-L99

Hronom:
Same here, need this feature for maven to do something like this:
```
 <plugin>
                <groupId>com.google.cloud.tools</groupId>
                <artifactId>jib-maven-plugin</artifactId>
                <version>0.9.9</version>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>build</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <from>
                        <image>openjdk:10-jre</image>
                    </from>
                    <to>
                        <image>registry.hub.docker.com/hronom/eureka-service:${git.commit.id}</image>
                        <image>registry.hub.docker.com/hronom/eureka-service:${project.version}</image>
                        <image>registry.hub.docker.com/hronom/eureka-service:latest</image>
                    </to>
                    <container>
                        <mainClass>${start-class}</mainClass>
                        <ports>
                            <port>8081</port>
                        </ports>
                    </container>
[...]

PR #963, opened 2018-09-10
Adds additional tags to BuildConfiguration.

coollog:
Prepares for #802.

Next is to add building to these additional tags in `PushImageStep`.


chanseokoh:
> Next is to add building to these additional tags in `PushImageStep`.

Are these simply additional tags? If yes, I'd rename the methods and fields so that they are extra tags. If not, are you planning to reconcile these with image references given to `ImageConfiguration` (i.e., do something about `ImageConfiguration.getImageTag()`, etc.?)

coollog:
@chanseokoh That sounds good, I'll change these to `additionalTags` for now until we decide to merge it with the tag in the `ImageConfiguration`.

PR #970, opened 2018-09-11
Add constants for configuration defaults

### LINKED ISSUE #738: Consolidate defaults into one class
Default values (e.g. `Instant.EPOCH` for creation time, `V22ManifestTemplate.class` for target format, etc.) are spread out and should be more centralized somewhere.

TadCordle:
Fixes #738


jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildConfigurationTest.java
```diff
@@ -172,14 +171,14 @@ public void testBuilder_default() {
             .setTargetImageConfiguration(targetImageConfiguration)
             .build();
 
-    Assert.assertEquals(V22ManifestTemplate.class, buildConfiguration.getTargetFormat());
+    Assert.assertEquals(
+        BuildConfiguration.DEFAULT_TARGET_FORMAT, buildConfiguration.getTargetFormat());
```
coollog: I think we can leave these using the actual value and make the constants private.

PR #971, opened 2018-09-11
Replaces JibLogger with EventEmitter in RetrieveRegistryCredentialsStep.

coollog:
Part of #714

- Adds a `LogEvent` that replaces the logging originally done by the `JibLogger` interface.
- Adds a new `TimerEvent` and accompanying `TimerEventEmitter` to emit timer events to handle durations
- Has `RetrieveRegistryCredentialsStep` emit `LogEvents` rather than call `JibLogger` interface
- Has `RetrieveRegistryCredentialsStep` emit `TimerEvents` rather than log timing debug messages
- Sets the appropriate logback methods on the Maven/Gradle side to handle the `LogEvent`s and `TimerEvent`s
- TODO in followup PRs: Replace all uses of `JibLogger` with emitting `LogEvent`/`TimerEvent` and remove `JibLogger`.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -82,22 +87,29 @@ static RetrieveRegistryCredentialsStep forTargetImage(
   @Override
   @Nullable
   public Credential call() throws CredentialRetrievalException {
-    buildLogger.lifecycle(String.format(DESCRIPTION, registry) + "...");
+    String description = makeDescription(registry);
+    emit(LogEvent.lifecycle(description + "..."));
 
-    try (Timer ignored = new Timer(buildLogger, String.format(DESCRIPTION, registry))) {
+    try (Timer ignored =
+        LoggingTimer.newTimer(logMessage -> emit(LogEvent.debug(logMessage)), description)) {
```
briandealwis: IMHO the base code shouldn't decide that timer events should be logged (or lifecycle events, for that matter).  Rather this should be emitting `TimerEvent` messages. The logger should decide whether it wants to consume and report those messages.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -51,6 +53,7 @@
     private ImmutableList<LayerConfiguration> layerConfigurations = ImmutableList.of();
     private Class<? extends BuildableManifestTemplate> targetFormat = DEFAULT_TARGET_FORMAT;
     private String toolName = DEFAULT_TOOL_NAME;
+    @Nullable private EventEmitter eventEmitter;
```
briandealwis: Could we have a `NonEmitter` that just silently drops all events?  Then we could avoid the `orElse` and null tests elsewhere.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -82,22 +87,29 @@ static RetrieveRegistryCredentialsStep forTargetImage(
...
       }
 
-      /*
-       * If no credentials found, give an info (not warning because in most cases, the base image is
-       * public and does not need extra credentials) and return null.
-       */
-      buildLogger.info("No credentials could be retrieved for registry " + registry);
+      // If no credentials found, give an info (not warning because in most cases, the base image is
+      // public and does not need extra credentials) and return null.
+      emit(LogEvent.info("No credentials could be retrieved for registry " + registry));
```
briandealwis: IMHO: this should be a `NoCredentialsForRegistryEvent`, and it should include the list of credentialRetrievers requested.  This base code should be totally oblivious to knowing how the results are reported to the user.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/LoggingTimer.java
```diff
@@ -0,0 +1,69 @@
...
+        nanoTime ->
+            logMessageConsumer.accept(
+                getTabs(depth)
+                    .append("TIMED\t")
+                    .append(label)
+                    .append(" : ")
+                    .append(nanoTime / 1000 / 1000.0)
+                    .append(" ms")
+                    .toString()));
+  }
```
briandealwis: This class seems odd to me: tracking depth and creating a representation seems like something that should be tracked by the logger.  What if the event consumer wants to emit a different representation of the timing data?  Eclipse, for example, has its own log format.

What I'd like is a `TimerEvent` that has a `Timer` instance (so I can track lap records), an event timestamp (absolute start time) as well as the elapsed time (which I should be able to fetch with a desired `TimeUnit` instance), and a state (started, in-progress, ended).  And `Timer` instances should have a parent timer.  Then my consumer can track the timers and turn them into a hierarchical format

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
coollog: That is a better design - I'll change to that.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
coollog: Yea, will do that.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
coollog: Oh yes, we will definitely be adding more events with actual structured data for the user to use. For now, these pull requests will first just replace the `JibLogger` code with `LogEvent`s to maintain the same functionality.

coollog:
As suggested, reorganized the timer classes by adding `TimerEvent` and moving the log message generation for the `TimerEvent`s to `plugins-common`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/Timer.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Captures the time since last lap or creation, and resets the start time.
+   *
+   * @return the duration of the last lap, or since creation
+   */
+  Duration lap() {
+    Instant now = Instant.now();
+    Duration duration = Duration.between(startTime, now);
+    startTime = now;
```
briandealwis: Hmm, I think we'd want to keep the original start time too.  So we can get both the duration and the elapsed time (since creation)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -33,14 +34,17 @@
...
+    return "Retrieving registry credentials for " + registry;
+  }
 
   /** Retrieves credentials for the base image. */
   static RetrieveRegistryCredentialsStep forBaseImage(
       ListeningExecutorService listeningExecutorService, BuildConfiguration buildConfiguration) {
     return new RetrieveRegistryCredentialsStep(
         listeningExecutorService,
-        buildConfiguration.getBuildLogger(),
+        // TODO: Replace with ExecutionMainframe.emit so that it is never nullable.
```
briandealwis: What's ExecutionMainframe?

jib-core/src/main/java/com/google/cloud/tools/jib/event/JibEventType.java
```diff
@@ -24,6 +26,12 @@
   /** All event types. Handlers for this will always be called first. */
   public static final JibEventType<JibEvent> ALL = new JibEventType<>(JibEvent.class);
 
+  /** Log message event. */
+  public static final JibEventType<LogEvent> LOG = new JibEventType<>(LogEvent.class);
+
+  /** Timer event for timing various part of Jib's execution. */
+  public static final JibEventType<TimerEvent> TIMER = new JibEventType<>(TimerEvent.class);
```
briandealwis: Question — since these are related to functionality, maybe they should be LOGGING and TIMING?

jib-core/src/main/java/com/google/cloud/tools/jib/event/events/TimerEvent.java
```diff
@@ -0,0 +1,114 @@
...
+  /** The state of the timing. */
+  public enum State {
+
+    /** The timer has started timing. Duration will be 0. */
+    START,
+
+    /** The timer is still in progress. Duration captures the time since the last event. */
+    IN_PROGRESS,
+
+    /** The timer has finished timing. Duration captures the time since the last event. */
```
briandealwis: I think it makes more sense to return the total time.  A swimmer's race records the individual lap times and their total time.  Or maybe TimerEvent should carry the duration and elapsed (total)

jib-core/src/test/java/com/google/cloud/tools/jib/builder/TimerTest.java
```diff
@@ -0,0 +1,45 @@
...
+public class TimerTest {
+
+  @Test
+  public void testLap() throws InterruptedException {
+    Timer parentTimer = new Timer();
+    TimeUnit.MILLISECONDS.sleep(5);
+    Duration parentDuration1 = parentTimer.lap();
+    TimeUnit.MILLISECONDS.sleep(10);
+    Duration parentDuration2 = parentTimer.lap();
+    TimeUnit.MILLISECONDS.sleep(1);
```
briandealwis: This isn't safe: Windows' sleep() is notoriously unpredictable.  The `Timer` should be provided an `Clock` instance that is driven by this test.  (And then we don't need to actually sleep.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/Timer.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/** Times code execution intervals. Call {@link #lap} at the end of each interval. */
+// TODO: Replace com.google.cloud.tools.jib.Timer with this.
+class Timer implements TimerEvent.Timer {
+
+  @Nullable private final Timer parentTimer;
+
+  private Instant startTime = Instant.now();
```
briandealwis: For testing we should be able to provide an instance of `Clock` rather than try to drive tests via the wall-clock.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
coollog: Ah will remove this - not needed anymore since we're using default no-op emitter.

jib-core/src/main/java/com/google/cloud/tools/jib/event/events/TimerEvent.java
coollog: Made TimerEvent carry the duration and elapsed

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -82,21 +85,21 @@ static RetrieveRegistryCredentialsStep forTargetImage(
   @Override
   @Nullable
   public Credential call() throws CredentialRetrievalException {
-    buildLogger.lifecycle(String.format(DESCRIPTION, registry) + "...");
+    String description = makeDescription(registry);
+    eventEmitter.emit(LogEvent.lifecycle(description + "..."));
```
TadCordle: Why not make the "..." part of `makeDescription`? Actually, `makeDescription` seems simple enough to just inline here?

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
```diff
@@ -16,30 +16,8 @@
 
 package com.google.cloud.tools.jib.event;
 
-import com.google.common.collect.ImmutableMultimap;
+/** Emits {@link JibEvent}s. */
+public interface EventEmitter {
```
TadCordle: Not sure I understand why this is necessary, what might non-default implementations of this look like?

jib-core/src/main/java/com/google/cloud/tools/jib/event/events/TimerEvent.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * The timer has finished timing. {@link #getDuration} is the time since the last event. {@link
+     * #getElapsed} is the total elapsed time.
+     */
+    FINISHED
+  }
+
+  /** Defines a timer hierarchy. */
+  public interface Timer {
```
TadCordle: Same comment here, are we expecting to have multiple implementations of `Timer`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/Timer.java
```diff
@@ -0,0 +1,69 @@
...
+import java.time.Instant;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/** Times code execution intervals. Call {@link #lap} at the end of each interval. */
+// TODO: Replace com.google.cloud.tools.jib.Timer with this.
+class Timer implements TimerEvent.Timer {
+
+  private final Clock clock;
+  @Nullable private final Timer parentTimer;
```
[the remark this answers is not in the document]
coollog: I believe the style is that `Optional`s should only be used as return types to indicate the potential for emptiness to users of a method. For parameters or fields, nullables are okay.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
coollog: The description without the "..." is passed into the `TimerEventEmitter`. I have `makeDescription` separate mostly to have it sort of like a static constant up top.

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
coollog: For example, a no-op EventEmitter such as on BuildConfiguration line 55.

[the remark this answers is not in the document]
coollog: Hmm, maybe `EventDispatcher`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/TimerEventEmitter.java
```diff
@@ -0,0 +1,81 @@
...
+
+  private static final Clock DEFAULT_CLOCK = Clock.systemUTC();
+
+  private final EventEmitter eventEmitter;
+  private final String description;
+
+  private final Clock clock;
+  private final Timer timer;
+
+  public TimerEventEmitter(EventEmitter eventEmitter, String description) {
```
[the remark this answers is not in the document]
coollog: I was thinking of having the `Timer` object be the reference to differentiate between timers.

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventEmitter.java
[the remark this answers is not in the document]
coollog: The rename seems to be a large change - I'll file a separate PR for that. #1001

jib-core/src/main/java/com/google/cloud/tools/jib/event/JibEventType.java
```diff
@@ -24,6 +26,12 @@
   /** All event types. Handlers for this will always be called first. */
   public static final JibEventType<JibEvent> ALL = new JibEventType<>(JibEvent.class);
 
+  /** Log message event. */
+  public static final JibEventType<LogEvent> LOGGING = new JibEventType<>(LogEvent.class);
```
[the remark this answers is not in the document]
coollog: That's a good point - I'll update the javadoc for that.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/TimerEventHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+    if (timerEvent.getState() == State.START) {
+      return logMessageBuilder.append("TIMING\t").append(timerEvent.getDescription()).toString();
+
+    } else {
+      return logMessageBuilder
+          .append("TIMED\t")
+          .append(timerEvent.getDescription())
+          .append(" : ")
+          .append(timerEvent.getDuration().toNanos() / 1000 / 1000.0)
```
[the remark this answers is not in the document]
coollog: I think the 0 is what we want, since this is at the millisecond granularity with 3 extra decimal digits.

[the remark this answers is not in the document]
chanseokoh: Oh, gotcha.

jib-core/src/main/java/com/google/cloud/tools/jib/event/DefaultEventEmitter.java
```diff
@@ -0,0 +1,46 @@
...
+    handlers = eventHandlers.getHandlers();
+  }
+
+  /**
+   * Emits {@code jibEvent} to all the handlers that can handle it.
+   *
+   * @param jibEvent the {@link JibEvent} to emit
+   */
+  @Override
+  public void emit(JibEvent jibEvent) {
```
loosebazooka: Should this happen on an "event" thread? So slow event handlers don't slow down the build steps?

Could events should be submitted to the dispatcher and then it just watches its queue to dispatch them to handlers (on a single event handler thread?)

chanseokoh: I still don't get the exact workings of our eventing design, but from my experience on some event models, I think it is a sane design that the event handlers run on the same thread where the actual event is taking place and often it is desirable. That is, if an event handler blocks, the main execution path will block. Often, you want to make sure certain things happen in a synchronized manner with the event, and if the handlers run on a separate thread that is independent of the main execution thread, I think it will be almost impossible to sync. Think about UI a text input widget as example: you want to validate if what the user entered in a field is correct, and the validation is triggered through text input events.

coollog: Yea, I think we should keep it on the same thread and delegate to the user to handle with inexpensive handlers. That would avoid the cost of dispatching to another thread for our plugins' case of just logging a log message.

coollog:
Merged master.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/LogEventHandler.java
```diff
@@ -0,0 +1,60 @@
...
+      case DEBUG:
+        log.debug(logEvent.getMessage());
+        break;
+
+      case ERROR:
+        log.error(logEvent.getMessage());
+        break;
+
+      case INFO:
+        // Use lifecycle for progress-indicating messages.
```
briandealwis: Should put this on `LogEvent.INFO` too.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/LogEventHandler.java
```diff
@@ -0,0 +1,60 @@
...
+        log.debug(logEvent.getMessage());
+        break;
+
+      case WARN:
+        log.warn("warning: " + logEvent.getMessage());
```
briandealwis: The `MavenJibLogger` wasn't prefixing messages with `warning:` previously.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/LogEventHandler.java
loosebazooka: Also, lifecycle and info might be two different things in some situations (for example: gradle).

loosebazooka: oh oops, I didn't realize this was in the maven plugin. Brian's point still holds though. :)

PR #972, opened 2018-09-11
Define system property names in one class

### LINKED ISSUE #829: Define system property names in a consistent location
`jib.httpTimeout`, `sendCredentialsOverHttp`, `jibSerialize`, and the `jib.to.auth.username` etc. system properties are all defined and used in different locations.

TadCordle:
Fixes #829


jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
```diff
@@ -0,0 +1,38 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.jib;
+
+/** Names of system properties defined/used by Jib. */
+public class JibSystemProperties {
+
+  /** (Integer) HTTP connection/read timeouts for registry interactions, in milliseconds. */
+  public static String HTTP_TIMEOUT = "jib.httpTimeout";
```
coollog: final for all

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
```diff
@@ -0,0 +1,38 @@
...
+
+  /** (Boolean) Enables more detailed logging and serializes Jib's execution. */
+  public static String SERIALIZE = "jibSerialize";
```
coollog: `Enables ... if set to {@code true}` and similar phrasing for below?

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib;
+
+/** Names of system properties defined/used by Jib. */
+public class JibSystemProperties {
```
coollog: I think if we have a class for this, we might as well have it be the class with the getters for the system properties like:

`int getHttpTimeout`
`bool isSerializedExecutionEnabled`
`bool isSendCredentialsOverHttpEnabled`
`bool isUserAgentEnabled`

TadCordle: Yeah I started out in this direction but wasn't sure how to deal with `ConfigurationPropertyValidator#checkHttpTimeoutProperty(...)`, since it gets `jib.httpTimeout` as a String first. I guess maybe we could move this method to the new class?

coollog: Hmm, yea that method can live in this new class.

coollog: I think you could just expose the property name constants for tests and not have to add more methods just for testing purposes.

TadCordle:
Side note: `jib.to.auth.username` and `jib.to.auth.password` are left out, since they seem to be a bit plugin-specific? Should I add these to core as well?

coollog:
@TadCordle Yea, those should be left out of `jib-core` since they are just aliases for the plugin configuration.

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright 2018 Google LLC. All rights reserved.
```
briandealwis: No _All rights reserved_

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
```diff
@@ -52,14 +53,14 @@ static PluginConfigurationProcessor processCommonConfiguration(
...
-    ConfigurationPropertyValidator.checkHttpTimeoutProperty(GradleException::new);
+    JibSystemProperties.checkHttpTimeoutProperty(GradleException::new);
 
     // TODO: Instead of disabling logging, have authentication credentials be provided
     GradleJibLogger.disableHttpLogging();
 
     ImageReference baseImage = ImageReference.parse(jibExtension.getBaseImage());
 
-    if (Boolean.getBoolean("sendCredentialsOverHttp")) {
+    if (Boolean.getBoolean(JibSystemProperties.SEND_CREDENTIALS_OVER_HTTP)) {
```
briandealwis: This should be `JibSystemProperties.isSendCredentialsOverHttpEnabled()`?

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
TadCordle: Oh, looks like there are other places with "All rights reserved" too. I'll fix this one and fix the rest in a followup.

jib-core/src/test/java/com/google/cloud/tools/jib/JibSystemPropertiesTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.jib;
+
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class JibSystemPropertiesTest {
```
coollog: nit: javadoc

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
```diff
@@ -0,0 +1,96 @@
...
+  public static final String SEND_CREDENTIALS_OVER_HTTP = "sendCredentialsOverHttp";
+
+  private static final String SERIALIZE = "jibSerialize";
+
+  private static final String DISABLE_USER_AGENT = "_JIB_DISABLE_USER_AGENT";
+
+  /**
+   * Gets the HTTP connection/read timeouts for registry interactions in milliseconds.
+   *
+   * @return the value of the {@code jib.httpTimeout} system property
```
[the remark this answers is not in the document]
briandealwis: Or `{@link #XXX}` to the new constants :-)

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
```diff
@@ -0,0 +1,96 @@
...
+   */
+  public static Integer getHttpTimeout() {
```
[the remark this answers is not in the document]
TadCordle: Null pointer exception in a bunch of places if I do this.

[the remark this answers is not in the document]
coollog: Oh, then this should be `@Nullable` or return an `Optional` to reflect nullability, and probably indicate that this can return null if no timeout is set. Another option is to choose a sensible default timeout for indicating no timeout set.

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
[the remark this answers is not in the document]
TadCordle: Maybe shouldn't `@link` since the constants are only exposed for testing?

jib-core/src/main/java/com/google/cloud/tools/jib/JibSystemProperties.java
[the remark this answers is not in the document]
TadCordle: I think we mention in the readme that the default is 20000, so I think I'll just return that if it's not set.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/DockerContextMojo.java
```diff
@@ -61,7 +61,11 @@ public void execute() throws MojoExecutionException {
 
     MavenJibLogger mavenJibLogger = new MavenJibLogger(getLog());
     handleDeprecatedParameters(mavenJibLogger);
-    JibSystemProperties.checkHttpTimeoutProperty(MojoExecutionException::new);
+    try {
+      JibSystemProperties.checkHttpTimeoutProperty();
```
coollog: `Assert.fail(...)`

TadCordle: Not a test

coollog: Oh oops I misread it lol.

PR #975, opened 2018-09-11
 Adds Cache with initial method for fetching a cache entry by layer entries.

coollog:
Part of #637

`Cache` - high-level operations on the cache - to be used by Jib builder steps
`LayerEntriesSelector` - serializes `LayerEntry`s into a selector digest


jib-core/src/main/java/com/google/cloud/tools/jib/json/JsonTemplateMapper.java
```diff
@@ -109,7 +109,7 @@
    * @return the template filled with the values parsed from {@code jsonFile}
```
loosebazooka: comment `{@code jsonFile} -> {@code jsonString}`? Also a little curious what this is doing? Does this just parse a string that we know is a list?

coollog: Yep, this would try to parse a string as a JSON list. It would fail if it is not a list.

loosebazooka: Is there a reason we can just use `readJson`?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.util.Optional;
+
+/** Cache for storing data to be shared between Jib executions. */
+public class Cache {
+
+  private final CacheStorage cacheStorage;
+
+  private Cache(CacheStorage cacheStorage) {
```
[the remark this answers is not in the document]
coollog: No way to instantiate this yet. Will add way to instantiate after `retrieveCacheEntry` and `write` are implemented.

jib-core/src/main/java/com/google/cloud/tools/jib/json/JsonTemplateMapper.java
coollog: Filed #997

chanseokoh: If feasible, can we fix #997 first and add "not ready for review" to this PR? Seems like #997 has a significant cleanup impact on this. Or, will it be too much work?

coollog: Not sure if #997 is valid to do yet. We'll have to explore that more and see if things will work without `ListOfJsonTemplate`, but I don't think it'll affect this PR much since the serialization of `LayerEntryTemplate` can just store it in a java List.

jib-core/src/test/java/com/google/cloud/tools/jib/ncache/LayerEntriesSelectorTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+  @Test
+  public void testLayerEntriesTemplate_getList() {
+    Assert.assertEquals(
+        toLayerEntryTemplates(IN_ORDER_LAYER_ENTRIES),
+        new LayerEntriesTemplate(OUT_OF_ORDER_LAYER_ENTRIES).getList());
+  }
+
+  @Test
+  public void testGenerateSelector_empty() throws IOException {
```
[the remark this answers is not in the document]
coollog: Oops this was meant to be a test for no layer entries.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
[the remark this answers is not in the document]
loosebazooka: add todo?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
```diff
@@ -0,0 +1,65 @@
...
+   * Retrieves the {@link CacheEntry} that was built from the {@code layerEntries}.
+   *
+   * @param layerEntries the layer entries to match against
+   * @return a {@link CacheEntry} that was built from {@code layerEntries}, or {@link
+   *     Optional#empty} if none found
+   * @throws IOException if an I/O exception occurs
+   * @throws CacheCorruptedException if the cache is corrupted
+   */
+  // TODO: Add test once other unimplemented methods are implemented.
+  public Optional<CacheEntry> retrieveCacheEntry(ImmutableList<LayerEntry> layerEntries)
```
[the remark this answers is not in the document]
coollog: Yea, we won't be exposing Guava types in user-facing APIs for Jib Core, but will still use Guava types for internal APIs (like the cache).

PR #976, opened 2018-09-12
 Serializes LayerEntry in unix path form.

### LINKED ISSUE #967: Ensure LayerEntry path usage does not create issues when converting to string
From original comment https://github.com/GoogleContainerTools/jib/pull/960#pullrequestreview-153897158:

> I'm a teeny bit leery of the many uses of `Path.toString()` and `Paths.get()` as they return platform-specific formats.  Windows does funny things with the [Yen symbol](https://docs.microsoft.com/en-us/windows/desktop/intl/character-sets-used-in-file-names).  I think it's hypothetical as I don't think these paths will leak outside of our cache — and the caches are very unlikely to ever be shared between platforms — and the `TarArchiveEntry` already handles names with backslashes, but I worry that we'll have *someone* try to pull in files with valid *nix filenames (e.g., with a colon or multibyte character) that is somehow rejected.  **I wonder if it's worth using a custom `FileSystem` for the container's file system so we can ensure consistent cross-filesystem behaviour.**

### LINKED ISSUE #979: Consider making LayerEntry source file and extraction path be absolute.
From https://github.com/GoogleContainerTools/jib/pull/976/files/2530a8af3f94e0a70602b75b6194096e8fc0cf2a#r217207527

coollog:
Fixes #967 
Fixes #979

This provides a consistent serialization form of the paths in `LayerEntry`. The serialized source file and extraction paths are absolute.


jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -34,6 +35,21 @@
...
 public class LayerEntry {
 
+  /**
+   * Stringifies {@code path} in unix form.
+   *
+   * @param path the path
+   * @return the string form of the path
+   */
+  private static String toUnixPath(Path path) {
+    StringJoiner pathJoiner =
```
loosebazooka: Should we just be always using absolute path?

chanseokoh:
The Windows test failure seems relevant.

coollog:
@chanseokoh Ah yes, I'll fix that.

coollog:
Windows tests are passing now.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -87,12 +104,12 @@ public Path getExtractionPath() {
   }
 
   /**
-   * Gets the source file path in string form.
+   * Gets the source file absolute path in string form.
```
chanseokoh: How about "Get the source file as an absolute path in string form. The path is made to absolute first, if not already absolute."

And do this for all the other `get...FileString()` methods.

Also, I think it's worth adding a test that a relative path is actually converted to an absolute path.

chanseokoh: Maybe renaming `getSourceFileString` to `getAbsoluteSourceFileString` as well?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
coollog: Actually resolving this in this PR now.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -34,6 +37,21 @@
  */
 public class LayerEntry {
 
+  /**
+   * Stringifies {@code path} in Unix form. The path must be absolute.
+   *
+   * @param path the path
+   * @return the string form of the absolute path
+   */
+  private static String toAbsoluteUnixPath(Path path) {
```
chanseokoh: `toUnixPath` sounds better to me. I mistakenly thought `toAbsolute...` would convert `path` to absolute first.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -61,10 +79,11 @@ public LayerEntry(Path sourceFile, Path extractionPath) {
   /**
    * Gets the source file.
    *
-   * <p>Do <b>not</b> call {@link Path#toString} on this - use {@link #getSourceFileString} instead.
-   * This path can be relative or absolute, and {@link #getSourceFileString} can also be relative or
-   * absolute, but callers should rely on {@link #getSourceFileString} for the serialized form since
-   * the serialization could change independently of the path representation.
+   * <p>Do <b>not</b> call {@link Path#toString} on this - use {@link #getAbsoluteSourceFileString}
+   * instead. This path can be relative or absolute, and {@link #getAbsoluteSourceFileString} can
+   * also be relative or absolute, but callers should rely on {@link #getAbsoluteSourceFileString}
```
chanseokoh: "can also be relative or absolute"

It can no longer be relative.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
```diff
@@ -31,23 +31,23 @@
...
-      return extractionPathString;
+    String getAbsoluteExtractionPathString() {
+      return absoluteExtractionPathString;
     }
 
     @VisibleForTesting
-    LayerMetadataEntry(String sourceFileString, String extractionPathString) {
-      this.sourceFileString = sourceFileString;
-      this.extractionPathString = extractionPathString;
+    LayerMetadataEntry(String absoluteSourceFileString, String absoluteExtractionPathString) {
```
chanseokoh: Is there someone verifying these arguments are absolute?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/json/CacheMetadataLayerPropertiesObjectTemplate.java
```diff
@@ -47,9 +47,9 @@ public String getExtractionPathString() {
       return extractionPath;
     }
 
-    public LayerEntryTemplate(String sourceFile, String extractionPath) {
-      this.sourceFile = sourceFile;
-      this.extractionPath = extractionPath;
+    public LayerEntryTemplate(String absoluteSourceFile, String absoluteExtractionPath) {
```
chanseokoh: Is there someone verifying these arguments are absolute?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
```diff
@@ -31,23 +31,23 @@
   static class LayerMetadataEntry {
 
     /** The source file path string, in OS-specific format. */
```
chanseokoh: Just a question: so this path will contain "/" on Unix and "\" on Windows, for example?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -75,10 +94,11 @@ public Path getSourceFile() {
   /**
    * Gets the extraction path.
    *
-   * <p>Do <b>not</b> call {@link Path#toString} on this - use {@link #getExtractionPathString}
-   * instead. This path can be relative or absolute, and {@link #getExtractionPathString} can also
-   * be relative or absolute, but callers should rely on {@link #getExtractionPathString} for the
-   * serialized form since the serialization could change independently of the path representation.
+   * <p>Do <b>not</b> call {@link Path#toString} on this - use {@link
+   * #getAbsoluteExtractionPathString} instead. This path can be relative or absolute, and {@link
+   * #getAbsoluteExtractionPathString} can also be relative or absolute, but callers should rely on
```
chanseokoh: "can also be relative or absolute"

It can no longer be relative.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
coollog: The path representation can still be relative or absolute, but the serialized string form is always absolute.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
coollog: Not currently, but the variable names are more of an indication of what to pass in rather than a requirement for being a valid `LayerMetadataEntry`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
chanseokoh: Sorry, I was referring to

"and {@link #getAbsoluteSourceFileString} can also be relative or absolute"

I believe "getAbsoluteSourceFileString" is always absolute.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerMetadata.java
coollog: Ah nope, need to update this comment.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
coollog: Ah right, I'll fix that.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -34,6 +37,25 @@
  */
 public class LayerEntry {
 
+  private static final Path ROOT = Paths.get("/");
```
TadCordle: Is this necessary? Only used in one place.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -87,22 +110,23 @@ public Path getExtractionPath() {
   }
 
   /**
-   * Gets the source file path in string form.
+   * Get the source file as an absolute path in Unix form. The path is made to absolute first, if
```
TadCordle: "... The path is made **~~to~~** absolute first..."?

PR #978, opened 2018-09-12
Adds to.tags configuration for Gradle.

coollog:


jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
```diff
@@ -190,6 +190,11 @@ public void testBuild_empty() throws IOException, InterruptedException {
         new Command("docker", "inspect", "-f", "{{.Created}}", targetImage).run().trim());
   }
 
+  @Test
+  public void testBuild_multipleTags() throws IOException, InterruptedException {
+    // TODO: Implement
```
chanseokoh: If this is to be implemented in a later PR, I'd remove it from here. It's just a test method, and I don't think we need a placeholder.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -0,0 +1,68 @@
...
+
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+import org.gradle.api.Action;
+import org.gradle.api.model.ObjectFactory;
+
+/** {@link ImageParameters} that configure the base image. */
+public class BaseImageParameters implements ImageParameters {
+
+  private AuthParameters auth;
```
chanseokoh: Seems like this can be `final`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ImageParameters.java
```diff
@@ -31,47 +29,25 @@
...
-  @Nullable private String image;
-  @Nullable private String credHelper;
-
-  @Inject
-  public ImageParameters(ObjectFactory objectFactory, String imageDescriptor) {
-    auth = objectFactory.newInstance(AuthParameters.class, imageDescriptor + ".auth");
-  }
+interface ImageParameters {
 
   @Input
```
chanseokoh: Is this `@Input` in effect in implementation classes? Should this be on implementation classes? (Not sure.)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -0,0 +1,68 @@
...
+    this.credHelper = credHelper;
+  }
+
+  @Override
+  public AuthParameters getAuth() {
+    return auth;
+  }
+
+  @Override
+  public void auth(Action<? super AuthParameters> action) {
```
chanseokoh: Does `Action<AuthParameters>` work?

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
coollog: Oh yea, I will implement this in this PR. I wanted to solicit some comments about the way `BaseImageParameters` and `TargetImageParameters` are laid out first.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
coollog: Yes that would work - is there a preference for using the specific type over allowing super types?

chanseokoh: This is not a public facing API, and I was wondering if there is anyone calling `auth()` with a type `Action<AuthProperty>` for example. Maybe I'm not familiar with Gradle idioms.

coollog: This is for like the `jib-gradle-plugin` `jib.to.auth` configuration so that users can configure the auth for a base or target image. The actual action that Groovy generates for this may be just `Action<AuthParameters>`, but I guess someone could decide to use like an `Action<Object>` that did some arbitrary thing on the auth object.

loosebazooka: I think it might have something to do with gradle decorating things.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibPluginIntegrationTest.java
```diff
@@ -106,6 +127,15 @@ private static String buildToDockerDaemonAndRun(TestProject testProject, String
     return new Command("docker", "run", imageReference).run();
   }
 
+  private static String pullAndRun(String imageReference) throws IOException, InterruptedException {
```
briandealwis: `pullAndRunJibBuiltImage`?  `pullAndRun` doesn't convey that it checks for a jib image layer.

coollog: Added some javadoc as well.

coollog:
Will merge master after #1019 to fix the failing tests.

PR #982, opened 2018-09-13
Logs additional tags.

coollog:

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
@@ -75,6 +75,16 @@ public class ImageReference {
   private static final Pattern REFERENCE_PATTERN = Pattern.compile(REFERENCE_REGEX);
 
   /**
+   * Parses a string {@code reference} into an {@link ImageReference}.
+   *
+   * <p>Image references should generally be in the form: {@code <registry>/<repository>:<tag>} For
+   * example, an image reference could be {@code gcr.io/distroless/java:debug}.
+   *
+   * <p>See <a
+   * href="https://docs.docker.com/engine/reference/commandline/tag/#extended-description">https://docs.docker.com/engine/reference/commandline/tag/#extended-description</a>
+   * for a description of valid image reference format. Note, however, that the image reference is
+   * referred confusingly as {@code tag} on that page.
+   *
    * @param reference the string to parse
    * @return an {@link ImageReference} parsed from the string
    * @throws InvalidImageReferenceException if {@code reference} is formatted incorrectly
@@ -134,9 +144,12 @@ public class ImageReference {
   }
 
   /**
-   * @param registry the image registry
+   * Constructs an {@link ImageReference} from the image reference components, consisting of an
+   * optional registry, a repository, and an optional tag.
+   *
+   * @param registry the image registry, or {@code null} to use the default registry (Docker Hub)
    * @param repository the image repository
-   * @param tag the image tag
+   * @param tag the image tag, or {@code null} to ues the default tag ({@code latest})
    * @return an {@link ImageReference} built from the given registry, repository, and tag
    */
   public static ImageReference of(
@@ -151,6 +164,9 @@ public class ImageReference {
   }
 
   /**
+   * Returns {@code true} if {@code registry} is a valid registry string. For example, a valid
+   * registry could be {@code gcr.io} or {@code localhost:5000}.
+   *
    * @param registry the registry to check
    * @return {@code true} if is a valid registry; {@code false} otherwise
    */
@@ -159,6 +175,9 @@ public class ImageReference {
   }
 
   /**
+   * Returns {@code true} if {@code repository} is a valid repository string. For example, a valid
+   * registry could be {@code distroless} or {@code my/container-image/repository}.
+   *
    * @param repository the repository to check
    * @return {@code true} if is a valid repository; {@code false} otherwise
    */
@@ -167,6 +186,9 @@ public class ImageReference {
   }
 
   /**
+   * Returns {@code true} if {@code tag} is a valid tag string. For example, a valid tag could be
+   * {@code v120.5-release}.
+   *
    * @param tag the tag to check
    * @return {@code true} if is a valid tag; {@code false} otherwise
    */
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -134,9 +144,12 @@ public static ImageReference parse(String reference) throws InvalidImageReferenc
...
 
   /**
-   * @param registry the image registry
+   * Constructs an {@link ImageReference} from the image reference components, consisting of an
+   * optional registry, a repository, and an optional tag.
+   *
+   * @param registry the image registry, or {@code null} to use the default registry (Docker Hub)
    * @param repository the image repository
-   * @param tag the image tag
+   * @param tag the image tag, or {@code null} to ues the default tag ({@code latest})
```
briandealwis: _to ~~ues~~ use_
`{@link #DEFAULT_TAG}`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
```diff
@@ -38,12 +39,45 @@
...
+
+  private static String buildMessageWithTargetImageReferences(
+      BuildConfiguration buildConfiguration, String prefix, String suffix) {
+    String targetRegistry = buildConfiguration.getTargetImageConfiguration().getImageRegistry();
+    String targetRepository = buildConfiguration.getTargetImageConfiguration().getImageRepository();
+
+    StringJoiner successMessageBuilder = new StringJoiner(", ", prefix, suffix);
+    for (String tag : buildConfiguration.getAllTargetImageTags()) {
+      successMessageBuilder.add(
+          colorCyan(ImageReference.of(targetRegistry, targetRepository, tag).toString()));
```
briandealwis: These ANSI sequences don't work for IDEs.  But they can be moved out to the logger once the emitting rolls out.

coollog: This is actually part of the Maven/Gradle plugins, so this will actually be code outside of Jib Core. Will definitely add startup and success events in Jib Core though.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
coollog: I was thinking since `latest` is a image reference spec constant, it should be referred to exactly, whereas `DEFAULT_TAG` is just a Jib constant that should always be equal to the real spec constant.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
briandealwis: Missed that this was in -plugins-common. 👍

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
briandealwis: Makes sense.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -159,6 +175,9 @@ public static boolean isValidRegistry(String registry) {
   }
 
   /**
+   * Returns {@code true} if {@code repository} is a valid repository string. For example, a valid
+   * registry could be {@code distroless} or {@code my/container-image/repository}.
```
chanseokoh: a valid registry --> a valid repository

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
chanseokoh: "to ues" --> "to use"

chanseokoh: > I was thinking since `latest` is a image reference spec constant, it should be referred to exactly, whereas `DEFAULT_TAG` is just a Jib constant that should always be equal to the real spec constant.

Are you saying `DEFAULT_TAG` may not be `latest`?

coollog: `DEFAULT_TAG` will always be `latest`, so having a link will just add an extra jump for readers of the javadoc.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
```diff
@@ -55,7 +89,13 @@ public static BuildStepsRunner forBuildImage(BuildConfiguration buildConfigurati
       throws CacheDirectoryCreationException {
     return new BuildStepsRunner(
         BuildSteps.forBuildToDockerRegistry(
-            buildConfiguration, getCacheInitializer(buildConfiguration)));
+            buildConfiguration, getCacheInitializer(buildConfiguration)),
+        String.format(
```
chanseokoh: I don't see any "%s" in `STARTUP_MESSAGE_PREFIX_FOR_DOCKER_REGISTRY`. Is `String.format()` working as intended?

coollog: Oops, this was supposed to be unwrapped

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -167,38 +186,72 @@ public static boolean isValidRepository(String repository) {
...
   }
 
+  /**
+   * Returns {@code true} if {@code tag} is the default tag (@code latest}; {@code false} if not.
+   *
+   * @param tag the tag to check
+   * @return {@code true} if {@code tag} is the default tag (@code latest}; {@code false} if not
+   */
+  public static boolean isDefaultTag(String tag) {
+    return tag.length() == 0 || DEFAULT_TAG.equals(tag);
```
chanseokoh: nit: `tag.isEmpty()`

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -167,38 +186,72 @@ public static boolean isValidRepository(String repository) {
...
+   *
    * @param tag the tag to check
    * @return {@code true} if is a valid tag; {@code false} otherwise
    */
   public static boolean isValidTag(String tag) {
     return tag.matches(TAG_REGEX);
   }
 
+  /**
+   * Returns {@code true} if {@code tag} is the default tag (@code latest}; {@code false} if not.
```
chanseokoh: "is the default tag latest or empty"? BTW, I'm not sure if it makes sense that this method returns true for an empty input as a utility method named `isDefaultTag`.

coollog: Hmm, I'm thinking that `default` means that it would be the default if the tag was empty.

PR #983, opened 2018-09-13
Builds to Docker daemon with additional tags.

coollog:

Now, build to Docker daemon first loads the tarball, and then does `docker tag` to tag with the additional tags.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
@@ -96,9 +96,9 @@ class LoadDockerStep implements AsyncStep<Void>, Callable<Void> {
 
     // Tags the image with all the tags. This only needs to be done for additional tags, so this can
     // be skipped if there is only one tag - the load above already loads that tag.
-    if (buildConfiguration.getAllTargetImageTags().size() > 0) {
+    if (buildConfiguration.getAllTargetImageTags().size() > 1) {
       for (String tag : buildConfiguration.getAllTargetImageTags()) {
-        dockerClient.tag(targetImageReference, targetImageReference.retag(tag));
+        dockerClient.tag(targetImageReference, targetImageReference.withTag(tag));
       }
     }
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
@@ -94,12 +94,13 @@ class LoadDockerStep implements AsyncStep<Void>, Callable<Void> {
     DockerClient dockerClient = new DockerClient();
     dockerClient.load(new ImageToTarballTranslator(image).toTarballBlob(targetImageReference));
 
-    // Tags the image with all the tags. This only needs to be done for additional tags, so this can
-    // be skipped if there is only one tag - the load above already loads that tag.
-    if (buildConfiguration.getAllTargetImageTags().size() > 1) {
-      for (String tag : buildConfiguration.getAllTargetImageTags()) {
-        dockerClient.tag(targetImageReference, targetImageReference.withTag(tag));
+    // Tags the image with all the additional tags, skipping the one 'docker load' already loaded.
+    for (String tag : buildConfiguration.getAllTargetImageTags()) {
+      if (tag.equals(targetImageReference.getTag())) {
+        continue;
       }
+
+      dockerClient.tag(targetImageReference, targetImageReference.withTag(tag));
     }
 
     return null;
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
```diff
@@ -85,13 +86,21 @@ public Void call() throws ExecutionException, InterruptedException {
...
-    new DockerClient()
-        .load(
-            new ImageToTarballTranslator(image)
-                .toTarballBlob(buildConfiguration.getTargetImageConfiguration().getImage()));
+    DockerClient dockerClient = new DockerClient();
+    dockerClient.load(new ImageToTarballTranslator(image).toTarballBlob(targetImageReference));
+
+    // Tags the image with all the tags. This only needs to be done for additional tags, so this can
+    // be skipped if there is only one tag - the load above already loads that tag.
+    if (buildConfiguration.getAllTargetImageTags().size() > 0) {
```
briandealwis: Shouldn't this be `> 1`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -201,6 +201,16 @@ public boolean usesDefaultTag() {
...
   }
 
+  /**
+   * Gets an {@link ImageReference} with the same registry and repository, but a different tag.
+   *
+   * @param tag the new tag
+   * @return an {@link ImageReference} with the same registry/repository and the new tag
+   */
+  public ImageReference retag(String tag) {
+    return ImageReference.of(registry, repository, tag);
```
TadCordle: This is a bit confusing to look at, since `registry`, `repository`, and `tag` are all fields in `ImageReference`, but in this case only `tag` is local. So maybe we should rename `tag` to `newTag` here.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -201,6 +201,16 @@ public boolean usesDefaultTag() {
     return DEFAULT_TAG.equals(tag);
   }
 
+  /**
+   * Gets an {@link ImageReference} with the same registry and repository, but a different tag.
+   *
+   * @param tag the new tag
+   * @return an {@link ImageReference} with the same registry/repository and the new tag
+   */
+  public ImageReference retag(String tag) {
```
briandealwis: My only hesitation is that this sounds like an action: retag this image as _tag_. `withTag()`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
```diff
@@ -85,13 +86,21 @@ public Void call() throws ExecutionException, InterruptedException {
...
-        .load(
-            new ImageToTarballTranslator(image)
-                .toTarballBlob(buildConfiguration.getTargetImageConfiguration().getImage()));
+    DockerClient dockerClient = new DockerClient();
+    dockerClient.load(new ImageToTarballTranslator(image).toTarballBlob(targetImageReference));
+
+    // Tags the image with all the tags. This only needs to be done for additional tags, so this can
+    // be skipped if there is only one tag - the load above already loads that tag.
+    if (buildConfiguration.getAllTargetImageTags().size() > 1) {
+      for (String tag : buildConfiguration.getAllTargetImageTags()) {
```
TadCordle: Not a big deal I guess, but this means we'll also retag with the tag that's already been docker loaded, correct?

coollog: Yea, that's true. I'll just cut that out.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
```diff
@@ -94,12 +94,13 @@ private Void afterPushBaseImageLayerFuturesFuture()
...
     dockerClient.load(new ImageToTarballTranslator(image).toTarballBlob(targetImageReference));
 
-    // Tags the image with all the tags. This only needs to be done for additional tags, so this can
-    // be skipped if there is only one tag - the load above already loads that tag.
-    if (buildConfiguration.getAllTargetImageTags().size() > 1) {
-      for (String tag : buildConfiguration.getAllTargetImageTags()) {
-        dockerClient.tag(targetImageReference, targetImageReference.withTag(tag));
+    // Tags the image with all the additional tags, skipping the one 'docker load' already loaded.
+    for (String tag : buildConfiguration.getAllTargetImageTags()) {
+      if (tag.equals(targetImageReference.getTag())) {
```
TadCordle: Could invert this logic and remove the continue since only 1 line follows.

coollog: This is just a personal preference thing, but I preferred to do it this way here because it clearly conveys that the original tag should be skipped (a single special case for the loop), whereas for most of the cases of the loop, the usual logic of performing the tagging should happen. Inverting the logic here conveys the sense of "only do the tagging if the tag is not the original", which is the same logic, but gives a sense that we are filtering for a small subset of the tags that is actually not the original tag.

PR #984, opened 2018-09-13
Add container.appRoot config parameter

### LINKED ISSUE #964: Add "container.appRoot" config parameter
The default will be `/app`, so when not explicitly given, the normal non-WAR Jib builds will work as before.

Knowing that users can already configure `ENTRYPOINT`, this new config parameter will help people use an arbitrary Servlet engine as a base image for building a WAR image (#431). In the WAR case, the default value will be `/jetty/webapps/ROOT`. (The default base image will be `gcr.io/distroless/java/jetty`.)

Don't forget to add docs after releasing it.

chanseokoh:
Fixes #964


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -32,6 +32,12 @@
   /** The default creation time of the container (constant to ensure reproducibility by default). */
   public static final Instant DEFAULT_CREATION_TIME = Instant.EPOCH;
 
+  /**
+   * The default app root in the image. For example, if this is set to {@code "/helloworld-app"},
```
loosebazooka: I feel like the example should just use the current default value?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -131,6 +131,11 @@ private static String mapToDockerfileString(Map<String, String> map, String comm
     return joiner.toString();
   }
 
+  @VisibleForTesting
+  static String getUnixPath(Path path) {
```
loosebazooka: I think another PR has a similar utility, might be worth using the same for both?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -36,47 +37,61 @@
...
+    private boolean isAppRootRelative() {
+      return appRootRelative;
+    }
   }
 
   /** Builds with each layer's files. */
   public static class Builder {
 
     private final Map<LayerType, List<Path>> layerFilesMap = new EnumMap<>(LayerType.class);
+    private Path appRoot = Paths.get(ContainerConfiguration.DEFAULT_APP_ROOT);
```
loosebazooka: I'm worried path might be implementation specific (different on windows), and this should be a string? Specifically when running Jib on a windows machine.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -110,6 +125,8 @@ public Builder setExplodedWarFiles(List<Path> explodedWarFiles) {
     }
 
     public JavaLayerConfigurations build() throws IOException {
+      Preconditions.checkState(appRoot.isAbsolute(), "'appRoot' must be an absolute path");
```
loosebazooka: this seems like we might need to use a custom validator that is not OS specific.

loosebazooka:
oops sorry I jumped the gun on the review there a little. Didn't see the `NOT READY` label.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
chanseokoh: Certainly. Since that utility is not yet available, I'll add a TODO here to use the utility.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: I had the same concern intially, but since the existing code is actually using `Path` to pass to `addEntryRecursively()`, I think it seemed better to accept `Path` for now: https://github.com/GoogleContainerTools/jib/pull/984/files#diff-1105d9e04ba077f47449d5c7095d4252L123 Maybe `addEntryRecursively()` shouldn't accept `Path`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: Not sure I get what you mean. (Is `C:\dir` not absolute, for example? Even so, I think it is right to fail, since we assume we build a Linux image.)

Anyways, I think we have been assuming that this `appRoot` should always in Unix-style in the end.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: Actually, your suggestion seems better. I changed it to accept Unix-style path in `String` instead of `Path`. This also removes the need for the `toUnixPath()` helper.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: Whatever you meant, I think this will be taken care of properly after I `appRoot` from `Path` to `String`. The new condition will look like

```java
      // Windows filenames cannot have "/", so this also blocks Windows-style path.
      Preconditions.checkState(appRoot.startsWith("/"),
          "appRoot should be an absolute path in Unix-style");
```

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
loosebazooka: Looks like addEntryRecursively actually boils down to some Unix path conversion eventually. https://github.com/GoogleContainerTools/jib/blob/8f14bd545339e372aae61522f35b783a9e2da855/jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java#L126

chanseokoh:
Ready for review.

You might first think too many files were touched, but the actual code change to make this work isn't really that large, when excluding tests.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: Yeah, so I'm wondering all the methods and classes that need the extraction path info should get the path in `String` rather than `Path` to indicate clearly that the extraction path is and will end up in Unix-style.

chanseokoh:
I agree it'd be good to have a `UnixPath` class.

> That way, we won't be limited to just a `String` or be subject to the OS-dependent effects of `Path`.

One thing to note is that, on Windows, you can construct a `Path` with a drive letter. (The root of such a path is a drive letter.) So, I thought explicitly accepting a `String` instead of `Path` has the advantage of preventing passing a `Path` with a drive letter. Since we don't support creating a Windows image, I thought `String` makes sense. WDYT?

coollog:
I think although `String` instead of `Path` has the advantage that it does not automatically add the drive letter, it doesn't really prevent there being a drive letter, and therefore requires extra logic for checks that it does not contain a drive letter. For `Path`, even though it could contain a drive letter when using `Path#toString`, when iterating over its path components, the drive letter is not included. I think either way requires special logic to handle this and the `UnixPath` class would be good to have. I think we can replace `LayerEntry` to use `UnixPath` as well and consolidate all of this logic. This would probably be good to have in a separate PR.

chanseokoh:
> I think we can replace `LayerEntry` to use `UnixPath` as well and consolidate all of this logic. This would probably be good to have in a separate PR.

Good idea. I'll file a new issue. It may require another big PR. For this PR, I guess we can stick to use `String` then.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -146,35 +146,38 @@ private static String mapToDockerfileString(Map<String, String> map, String comm
    * @param javaLayerConfigurations the {@link JavaLayerConfigurations}
    */
   public JavaDockerContextGenerator(JavaLayerConfigurations javaLayerConfigurations) {
+    String appRoot = javaLayerConfigurations.getAppRoot();
+    appRoot = appRoot.endsWith("/") ? appRoot : appRoot + '/';
```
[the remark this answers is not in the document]
chanseokoh: To be fixed by #992.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
```diff
@@ -16,24 +16,29 @@
...
-  public static final String DEFAULT_CLASSES_PATH_ON_IMAGE = "/app/classes/";
-  public static final String DEFAULT_DEPENDENCIES_PATH_ON_IMAGE = "/app/libs/";
-  public static final String DEFAULT_JETTY_BASE_ON_IMAGE = "/jetty/webapps/ROOT/";
+  public static final String DEFAULT_RESOURCES_PATH_ON_IMAGE = "resources/";
+  public static final String DEFAULT_CLASSES_PATH_ON_IMAGE = "classes/";
+  public static final String DEFAULT_DEPENDENCIES_PATH_ON_IMAGE = "libs/";
+
+  public static List<String> makeDefaultEntrypoint(
+      String appRoot, List<String> jvmFlags, String mainClass) {
+    Preconditions.checkArgument(
```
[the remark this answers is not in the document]
chanseokoh: To be fixed by #992.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -36,47 +36,62 @@
   /** Represents the different types of layers for a Java application. */
   @VisibleForTesting
   enum LayerType {
-    DEPENDENCIES("dependencies", JavaEntrypointConstructor.DEFAULT_DEPENDENCIES_PATH_ON_IMAGE),
+    DEPENDENCIES(
```
chanseokoh: Thinking more about this enum where we need to dynamically decide the final extraction path, I'm now leaning toward getting rid of static extraction path initialization with the `appRootRelative` switch. Instead, the `JavaLayerConfiguration.Builder()` could accept the extraction path when setting layer files. That is, change `setClassFiles(List<Path> classFiles)` to `setClassFiles(String extractionPath, List<Path> classFiles)`. That seems more straightforward for our dynamic path requirement and simpler. What do you think?

coollog: Yea that sounds good. One nit is to have it be `setClassFiles(List<Path> classFiles, String extractionPath)` to be more like `LayerEntry`.

chanseokoh:
Will re-design this per https://github.com/GoogleContainerTools/jib/pull/984#discussion_r218169036. Not ready for review until then.

chanseokoh:
Ready for review

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -32,6 +32,12 @@
   /** The default creation time of the container (constant to ensure reproducibility by default). */
   public static final Instant DEFAULT_CREATION_TIME = Instant.EPOCH;
 
+  /**
+   * The default app root in the image. For example, if this is set to {@code "/app"}, dependency
+   * JARs will be in {@code "/app/libs"}.
+   */
+  public static final String DEFAULT_APP_ROOT = "/app";
```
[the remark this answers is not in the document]
coollog: Was this resolved?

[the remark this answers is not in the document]
chanseokoh: Moved to `JavaLayerConfigurations`, but then I reworked that class, and this is not used in `JavaLayerConfigurations` anymore, so ended up back here. It's being referred to by the plugins (`ContainerParameters` and `JibPluginConfiguration`). If you think there's a better place, let me know and I'll move this.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -134,10 +146,13 @@ public static Builder builder() {
   }
 
   private final ImmutableMap<LayerType, LayerConfiguration> layerConfigurationMap;
+  private final ImmutableMap<LayerType, String> defaultExtractionPathMap;
```
coollog: Just `extractionPathMap`? This doesn't look like it's getting any defaults.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -134,10 +146,13 @@ public static Builder builder() {
...
 
   private JavaLayerConfigurations(
-      ImmutableMap<LayerType, LayerConfiguration> layerConfigurationsMap) {
-    this.layerConfigurationMap = layerConfigurationsMap;
+      ImmutableMap<LayerType, LayerConfiguration> layerConfigurationsMap,
+      ImmutableMap<LayerType, String> defaultExtractionPathMap) {
+    layerConfigurationMap = layerConfigurationsMap;
```
coollog: Perhaps we should name these both `layerConfigurationsMap` so this can be the more familiar `this.x = x` form?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -170,7 +185,35 @@ private JavaLayerConfigurations(
     return getLayerEntries(LayerType.EXPLODED_WAR);
   }
 
+  public String getDependencyDefaultExtractionPath() {
```
coollog: Are these getting defaults or the actual extraction path?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -36,96 +36,108 @@
...
 
     private Builder() {
       for (LayerType layerType : LayerType.values()) {
         layerFilesMap.put(layerType, new ArrayList<>());
+        extractionPathMap.put(layerType, "/");
       }
     }
 
-    public Builder setDependencyFiles(List<Path> dependencyFiles) {
+    public Builder setDependencyFiles(List<Path> dependencyFiles, String extractionPath) {
```
coollog: Should perhaps `JavaLayerConfigurations` take an `appRoot` and build the `extractionPaths` rather than taking a separate `extractionPath` for each and delegating the `extractionPath` building to the caller?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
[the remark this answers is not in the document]
chanseokoh: Alright, I'll move it to `JavaLayerConfigurations`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: I'll remove `Default` from the name.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: I think what you said is basically the original design, where the builder had `setAppRoot()` (https://github.com/GoogleContainerTools/jib/pull/984#discussion_r218169036)? The new design allows more flexibility that can cover the WAR case too. For example, the resources layer goes into "<appRoot>/resources" for non-WAR, but it should go to "<appRoot>" in the WAR case.

coollog: Ah right.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -36,108 +36,129 @@
...
   /** Builds with each layer's files. */
   public static class Builder {
 
     private final Map<LayerType, List<Path>> layerFilesMap = new EnumMap<>(LayerType.class);
+    private final Map<LayerType, String> extractionPathMap = new EnumMap<>(LayerType.class);
 
     private Builder() {
       for (LayerType layerType : LayerType.values()) {
         layerFilesMap.put(layerType, new ArrayList<>());
+        extractionPathMap.put(layerType, "/");
```
coollog: Is there a need to initialize these with the path as "/"? (Perhaps the code that builds the `LayerConfiguration` can skip if the files or extraction path is empty?)

chanseokoh: I think it's good to have "/". Otherwise, we may go into the business of returning `Optional<String>` for `getResourceExtractionPath()` instead of `String`.

chanseokoh:
> Are you planning to merge the `Absolute/RelativeUnixPath` changes in this PR or first merge this PR?

I was about to do this. If you think you can approve this PR soon, let's merge it, and I'll work on `Absolute` path right away. Otherwise, I can work it in this PR.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: That's a good idea. I was thinking the same. I'll file an issue.

coollog:
Okay let's merge this first then so the changeset is smaller. I think we may possibly want to have #1009 fixed before changing this over.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: Filed #1010.

Issue #337, opened 2018-06-01
Consider separating jib into class library and plugins


elharo:
This project could be divided into an independent library for building images and pushing them to repos; and the Maven and gradle specific parts. 

Among other uses, this would allow us to adopt jib for building/pushing Docker images from Eclipse and IntelliJ.

coollog:
A quick survey for everyone who thumbs upped this issue - *please comment below with your intended use case*. This will help us think about how we could better structure the features of the library. Thanks!

schmitch:
For a Sbt Plugin: https://github.com/schmitch/sbt-jib

rmannibucau:
To be able to manipulate layers of multiple images and merge them into a single one with some customizations. Can likely end up as a maven/gradle plugin but goal is to aggregate images before a cloud vendor specific deployment.

coollog:
Hi everyone, we have released an initial alpha version of the Jib Core library for building containers in Java. Please see the draft documentation at https://github.com/GoogleContainerTools/jib/tree/master/jib-core and let us know if there's any issues or questions!

Issue #953, opened 2018-09-07
Consider making JibContainerBuilder interface more fluent.


coollog:
Continuing from #881...

Considering that we are modeling `JibContainerBuilder`'s interface off of the `Stream` API, and that `JibContainerBuilder` begins with `Jib.from`, we should probably make the rest of the API more fluent as well.

`addLayer(List<Path>, String)` -> `layer(List<Path>, String)`
`addLayer(LayerConfiguration)` -> `layer(LayerConfiguration)`
`setLayers(List<LayerConfiguration>)` -> `layer(List<LayerConfiguration>)`
`setEntrypoint(List<String>/String...)` -> `entrypoint(List<String>/String...)`
`setProgramArguments(List<String>/String...)` -> `programArguments(List<String>/String...)`
`setEnvironment(Map<String, String>)` -> `environment(Map<String, String>)`
`setEnvironmentVariable(String, String)` -> `environmentVariable(String, String)`
`setExposedPorts(List<Port>/Port...)` -> `expose(List<Port>/Port...)`
`addExposedPort(Port)` -> `expose(Port)`
`setLabels(Map<String, String>)` -> `label(Map<String, String>)`
`addLabel(String, String)` -> `label(String, String)`

Also, currently, the collection setters (like `setExposedPorts`) replace previously-set ports. If converting to use this more fluent API, these collection setters should append, not replace. We could add methods like `exposeOnly(Map)` that behaves like the original setters.

briandealwis:
We could provide `exposeOnly(Map<>)` variants that take a `Map`?

TadCordle:
Should we close this? Haven't had any discussions about it in a while (personally I like the explicit API over the fluent one).

briandealwis:
@loosebazooka recently turned appengine-plugins-core to the more fluent approach.  I think it looks better, but I don't feel strongly.

loosebazooka:
I just made everything a builder: ¯\_(ツ)_/¯

briandealwis:
You also did `setX()` &rarr; `x()`

coollog:
Closing and sticking with current design since it is clearer in the functions' intentions, especially for methods like `addExposedPort` vs `setExposedPort` (`expose` vs `exposeOnly` is less clear and still involves extra `Only` portion).

PR #987, opened 2018-09-14
Proposal RFC: Jib Core as a library for Java

coollog:
For #337 

Preview at: https://github.com/GoogleContainerTools/jib/blob/jib-core-proposal/proposals/jib_core_library.md

Note that the proposed API is currently in implementation.


proposals/jib_core_library.md
```diff
@@ -0,0 +1,74 @@
...
+- `JibContainerBuilder setLayers(List<LayerConfiguration>/LayerConfiguration...)`
+- `JibContainerBuilder addLayer(LayerConfiguration)`
+- `JibContainerBuilder setEntrypoint(List<String>/String...)`
+- `JibContainerBuilder setProgramArguments(List<String>/String...)`
+- `JibContainerBuilder setEnvironment(Map<String, String> environmentMap)`
+- `JibContainerBuilder addEnvironmentVariable(String name, String value)`
+- `JibContainerBuilder setExposedPorts(List<Port>/Port...)`
+- `JibContainerBuilder addExposedPort(Port port)`
+- `JibContainerBuilder setLabels(Map<String, String> labelMap)`
+- `JibContainerBuilder addLabel(String key, String value)`
```
briandealwis: Weren't we talking about having a more fluent-looking `environment()`, `labels()`?

coollog: Right, that was in #953 - I'll update these to reflect the proposal there.

coollog:
Thought about the `TargetImage` types more and we should probably have `SourceImage` as well with the initial implementing class be `RegistryImage`. `Jib#from` would also need to take a `RegistryImage` in order to specify credentials for the base image.

proposals/jib_core_library.md
briandealwis: But I don't think we ever replace layers, do we.  At least, not for the API. Right?

coollog: `setLayers` currently would actually replace layers (https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java#L109)

proposals/jib_core_library.md
```diff
@@ -0,0 +1,76 @@
...
+# Proposed API
+
+`Jib` - the main entrypoint for using Jib Core
+- `JibContainerBuilder from(String baseImageReference)`
+- `JibContainerBuilder from(ImageReference baseImageReference)`
+- `JibContainerBuilder from(RegistryImage baseImage)`
+
+`JibContainerBuilder` - configures the container to build
+- `JibContainerBuilder layer(List<Path> files, Path pathInContainer)`
+- `JibContainerBuilder layer(LayerConfiguration)`
```
TadCordle: These two were changed to `addLayer`, right?

proposals/jib_core_library.md
```diff
@@ -0,0 +1,76 @@
...
+`DockerDaemonImage` - builds to a Docker daemon
+- `static DockerDaemonImage named(ImageReference/String)`
+- `DockerDaemonImage setDockerExecutable(Path)`
+
+`TarImage` - builds to a tarball archive
+- `Builder`
+  - `TarImage saveTo(Path outputFile)`
+- `static Builder named(ImageReference/String)`
+
+`Containerizer` - configures how and where to containerize to
```
TadCordle: Is `withExectionContext` not a thing anymore?

coollog: Nope - it made sense to just have the things that would have been under `ExecutionContext` on the `Containerizer` itself since that is the execution context.

proposals/jib_core_library.md
coollog: Yep, I'll change this over.

coollog:
Merged, but final API may be subject to change.

Issue #992, opened 2018-09-17
UnixPath class for strictly accepting Unix-style path


chanseokoh:
From https://github.com/GoogleContainerTools/jib/pull/984#issuecomment-422042180

We don't build Windows images, so some image-building components work on the assumption that given paths are in Unix-style. We could introduce `UnixPath` to replace `Path`, and a class like `LayerType` or `LayerEntry` could accept `UnixPath`. If implementing `UnixPath` as a replacement for `Path` may look like an over-engineering, at least we could make `UnixPath` as a convenient util class for conversion to and from Unix paths and checking if a path is a Unix one.

coollog:
Possible `UnixPath` API:

`static UnixPath fromPath(Path path)`
`static UnixPath fromPathString(String path)`
`UnixPath resolve(UnixPath/String/Path)`
`boolean isAbsolute()`
`String toString()`

coollog:
I'll take a stab at this and see how useful it is.

TadCordle:
Looks like this was already finished.

PR #995, opened 2018-09-17
Adds AbsoluteUnixPath to represent Unix paths in absolute form.

coollog:
Part of #992.

Followup PR will change `LayerEntry#extractionPath` to use this new `AbsoluteUnixPath`.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
@@ -62,7 +62,10 @@ public class AbsoluteUnixPath {
     return new AbsoluteUnixPath(pathJoiner.toString());
   }
 
-  /** Unix-style path, in absolute form. Does not end with trailing slash. */
+  /**
+   * Unix-style path, in absolute form. Does not end with trailing slash, except for the file system
+   * root ({@code /}).
+   */
   private final String unixPath;
 
   private AbsoluteUnixPath(String unixPath) {
@@ -76,8 +79,11 @@ public class AbsoluteUnixPath {
    * @return a new {@link AbsoluteUnixPath} representing the resolved path
    */
   public AbsoluteUnixPath resolve(RelativeUnixPath relativeUnixPath) {
-    return AbsoluteUnixPath.fromPath(
-        Paths.get(unixPath).resolve(relativeUnixPath.getRelativePath()));
+    StringJoiner pathJoiner = new StringJoiner("/", unixPath + '/', "");
+    for (String pathComponent : relativeUnixPath.getRelativePathComponents()) {
+      pathJoiner.add(pathComponent);
+    }
+    return AbsoluteUnixPath.get(pathJoiner.toString());
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
@@ -52,7 +52,7 @@ public class AbsoluteUnixPath {
    * @param path the absolute {@link Path} to convert to an {@link AbsoluteUnixPath}.
    * @return a new {@link AbsoluteUnixPath}
    */
-  private static AbsoluteUnixPath fromPath(Path path) {
+  public static AbsoluteUnixPath fromPath(Path path) {
     Preconditions.checkArgument(
         path.getRoot() != null, "Cannot create AbsoluteUnixPath from non-absolute Path: " + path);
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
@@ -32,24 +32,6 @@ import javax.annotation.concurrent.Immutable;
 public class AbsoluteUnixPath {
 
   /**
-   * Gets a new {@link AbsoluteUnixPath} from a {@link Path}. The {@code path} must be absolute
-   * (indicated by a non-null {@link Path#getRoot}).
-   *
-   * @param path the absolute {@link Path} to convert to an {@link AbsoluteUnixPath}.
-   * @return a new {@link AbsoluteUnixPath}
-   */
-  public static AbsoluteUnixPath fromPath(Path path) {
-    Preconditions.checkArgument(
-        path.getRoot() != null, "Cannot create AbsoluteUnixPath from non-absolute Path: " + path);
-
-    StringJoiner pathJoiner = new StringJoiner("/", "/", "");
-    for (Path pathComponent : path) {
-      pathJoiner.add(pathComponent.getFileName().toString());
-    }
-    return new AbsoluteUnixPath(pathJoiner.toString());
-  }
-
-  /**
    * Gets a new {@link AbsoluteUnixPath} from a Unix-style path string. The path must begin with a
    * forward slash ({@code /}).
    *
@@ -62,6 +44,24 @@ public class AbsoluteUnixPath {
     return fromPath(Paths.get(path));
   }
 
+  /**
+   * Gets a new {@link AbsoluteUnixPath} from a {@link Path}. The {@code path} must be absolute
+   * (indicated by a non-null {@link Path#getRoot}).
+   *
+   * @param path the absolute {@link Path} to convert to an {@link AbsoluteUnixPath}.
+   * @return a new {@link AbsoluteUnixPath}
+   */
+  private static AbsoluteUnixPath fromPath(Path path) {
+    Preconditions.checkArgument(
+        path.getRoot() != null, "Cannot create AbsoluteUnixPath from non-absolute Path: " + path);
+
+    StringJoiner pathJoiner = new StringJoiner("/", "/", "");
+    for (Path pathComponent : path) {
+      pathJoiner.add(pathComponent.getFileName().toString());
+    }
+    return new AbsoluteUnixPath(pathJoiner.toString());
+  }
+
   /** Unix-style path, in absolute form. Does not end with trailing slash. */
   private final String unixPath;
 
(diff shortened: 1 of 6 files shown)


chanseokoh:
I think `UnixPath` can be useful. I can definitely some places where I want `UnixPath` instead of `AbsoluteUnixPath`. How about having `isAbsolute()` on `UnixPath` instead?

coollog:
@chanseokoh Hmm, where would you see use of `UnixPath`? I think of just paths in the container, and those will always be absolute.

chanseokoh:
For example, https://github.com/GoogleContainerTools/jib/pull/984/files#diff-d4e5570720a7c83a526956eb06f31cffR27

So we will need to concat two `UnixPath`s.
```java
UnixPath appRoot = ...;
UnixPath DEFAULT_CLASSES_PATH_ON_IMAGE = ...;

UnixPath finalPathInContainer = appRoot.resolve(DEFAULT_CLASSES_PATH_ON_IMAGE);
```

chanseokoh:
Technically that certainly works, but conceptually, `RELATIVE_CLASSES_PATH_ON_IMAGE` should be a Unix path, so if we can declare that it is a Unix path, that will convey the clear message. And since we will be introducing a new class to present a Unix path, it makes sense to apply that to any path expected to be a Unix path.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
```diff
@@ -0,0 +1,110 @@
...
+   *
+   * @param path the absolute {@link Path} to convert to an {@link AbsoluteUnixPath}.
+   * @return a new {@link AbsoluteUnixPath}
+   */
+  public static AbsoluteUnixPath fromPath(Path path) {
+    Preconditions.checkArgument(
+        path.getRoot() != null, "Cannot create AbsoluteUnixPath from non-absolute Path: " + path);
+
+    StringJoiner pathJoiner = new StringJoiner("/", "/", "");
+    for (Path pathComponent : path) {
```
loosebazooka: I think there's a way to use streams here if you want:
```java
Streams.stream(path.iterator()).collect(Collectors.joining("/","/",""));
```

coollog:
Okay that sounds reasonable. For `toString` though, we would probably want to only serialize absolute paths, so `UnixPath` may be better served as an interface with two subclasses - `AbsoluteUnixPath` and `RelativeUnixPath`, where `AbsoluteUnixPath` would be the only one that is serializable to string?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Gets a new {@link AbsoluteUnixPath} from a {@link Path}. The {@code path} must be absolute
+   * (indicated by a non-null {@link Path#getRoot}).
+   *
+   * @param path the absolute {@link Path} to convert to an {@link AbsoluteUnixPath}.
+   * @return a new {@link AbsoluteUnixPath}
+   */
+  public static AbsoluteUnixPath fromPath(Path path) {
+    Preconditions.checkArgument(
```
loosebazooka: Are we expecting to have input somewhere where we can't go directly from string -> unix path?
Should we only be parsing strings?

chanseokoh:
That can work too. (BTW, by "serializing", do you mean printing out a path in Unix-style? Every class will have `toString()` anyway.)

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
coollog: I think since this might be a class that is called many times (for processing many `LayerEntry`s or serializing for cache metadata), we should prefer to use iterative style over streams since streams add a significant amount of overhead.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
coollog: Yea, although it is a weird use case. For `LayerEntry#getSourceFileString`, it will try to first get an `AbsoluteUnixPath` for the source file and then serialize it to get the serialized form for the cache metadata. We could just remove this method in favor of only the `get` method and just convert the `LayerEntry#sourceFile` to a string first.

coollog:
@chanseokoh Yep, serialization as in serializing to a Unix-style path string in absolute form.

coollog:
I was thinking maybe we should limit this to our use case (container file systems) and call this `PathInContainer`?

coollog:
Going to rework `AbsoluteUnixPath` to have it only parse Unix-style paths like `RelativeUnixPath`

coollog:
Done. Ready for review again.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  /**
+   * Gets a new {@link AbsoluteUnixPath} from a {@link Path}. The {@code path} must be absolute
+   * (indicated by a non-null {@link Path#getRoot}).
+   *
+   * @param path the absolute {@link Path} to convert to an {@link AbsoluteUnixPath}.
+   * @return a new {@link AbsoluteUnixPath}
+   */
+  private static AbsoluteUnixPath fromPath(Path path) {
```
chanseokoh: No one seems to use this `private` method.

coollog: Oops meant to make this public. This will be used by LayerEntry to turn the source file path into an AbsoluteUnixPath for serialization.

jib-core/src/test/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPathTest.java
```diff
@@ -0,0 +1,64 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.filesystem;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+/** Test for {@link AbsoluteUnixPath}. */
+public class AbsoluteUnixPathTest {
```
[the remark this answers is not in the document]
coollog: I included those in `UnixPathParserTest`. I think we don't have to take care of `..` and `.` since those are valid unix path components and just have this class be absolute unix paths with no canonicalization guarantees.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
chanseokoh: I actually had a question that I was about to ask.

The source file path of `LayerEntry` is, e.g., the path of an actual file on a local machine, right? That is, it could be like "D:\my-proj\build\classes\..." on Windows. And I see we are converting it to a Unix path for serialization. (Will be `AbsoluteUnixPath` after this PR.) Doesn't it discard the drive letter (D:\)?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (!(other instanceof AbsoluteUnixPath)) {
+      return false;
+    }
+    AbsoluteUnixPath otherAbsoluteUnixPath = (AbsoluteUnixPath) other;
```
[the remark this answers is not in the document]
coollog: Yes, since for `AbsoluteUnixPath`, the string form is the only one that is externally-accessible and should be what identifies the path.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
coollog: Yea, that makes me think we might just need to store the windows form path for the source file in the cache metadata and remove this method.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/RelativeUnixPath.java
```diff
@@ -0,0 +1,57 @@
...
+public class RelativeUnixPath {
+
+  /**
+   * Gets a new {@link RelativeUnixPath} from a Unix-style path in relative form. The {@code path}
+   * must be relative (does not begin with a leading slash {@code /}).
+   *
+   * @param relativePath the relative path
+   * @return a new {@link RelativeUnixPath}
+   */
+  public static RelativeUnixPath get(String relativePath) {
```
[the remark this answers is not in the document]
chanseokoh: File a new issue?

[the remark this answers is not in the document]
coollog: Oh I had made this change already - `get` now splits using `/`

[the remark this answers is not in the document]
coollog: (Using `UnixPathParser`)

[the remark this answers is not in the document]
chanseokoh: Oops sorry, put a comment in a wrong conversation.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
chanseokoh: This is the right conversation. File a new issue for https://github.com/GoogleContainerTools/jib/pull/995#discussion_r218589436?

coollog: Ah yes, filed #1008

PR #1006, opened 2018-09-19
Wait until local registry is ready when setting it up in integration tests

### LINKED ISSUE #965: Flaky integration test with LocalRegistry(5000, "testuser", "testpassword")
I've seen this error a few times. Looking at the code of `LocalRegistry`. One possibility I'm thinking is that the local registry in the image needs some time to be fully operational, while the test code is trying to access the registry too early.

```
com.google.cloud.tools.jib.gradle.JibPluginIntegrationTest > testDockerDaemon_defaultTarget FAILED
    java.lang.RuntimeException: Command 'docker login localhost:5000 -u testuser --password-stdin' failed: Error response from daemon: Get http://localhost:5000/v2/: EOF
        at com.google.cloud.tools.jib.Command.run(Command.java:68)
        at com.google.cloud.tools.jib.registry.LocalRegistry.login(LocalRegistry.java:145)
        at com.google.cloud.tools.jib.registry.LocalRegistry.pullAndPushToLocal(LocalRegistry.java:135)
        at com.google.cloud.tools.jib.gradle.JibPluginIntegrationTest.setup(JibPluginIntegrationTest.java:177)
```

chanseokoh:
Fixes #965.

Waits for up to 20 seconds.

- I don't think `--restart=always` is ever needed. What the flag would do is that the container will automatically start up if the Docker daemon restarts.
- No need to explicitly call `docker rm <container>` when giving `--rm` for `docker run`.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
@@ -18,6 +18,9 @@ package com.google.cloud.tools.jib.registry;
 
 import com.google.cloud.tools.jib.Command;
 import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -55,10 +58,10 @@ public class LocalRegistry extends ExternalResource {
             Arrays.asList(
                 "docker",
                 "run",
+                "--rm",
                 "-d",
                 "-p",
                 port + ":5000",
-                "--restart=always",
                 "--name",
                 containerName));
     if (username != null && password != null) {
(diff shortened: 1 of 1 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -151,4 +148,20 @@ private void logout() throws IOException, InterruptedException {
...
+
+    for (int i = 0; i < 20; i++) {
+      try {
+        HttpURLConnection connection = (HttpURLConnection) queryUrl.openConnection();
+        int code = connection.getResponseCode();
+        if (code == HttpURLConnection.HTTP_OK || code == HttpURLConnection.HTTP_UNAUTHORIZED) {
+          return;
+        }
+      } catch (IOException ex) {
+        Thread.sleep(1000);
```
briandealwis: Might be worth halving or quartering this: it's unlikely to need more than 250ms or 500ms.  Could use a `Stopwatch` and output how much time was waited?

No sleeping on a non-200/401?

PR #1012, opened 2018-09-19
Use AbsoluteUnixPath for container.appRoot propagation

### LINKED ISSUE #1011: Use AbsoluteUnixPath for container.appRoot propagation

chanseokoh:
Fixes #1011.


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -282,9 +285,9 @@ public void generate(Path targetDirectory) throws IOException {
         // 'baseExtractionPath' of '/app/classes', and an 'actualExtractionPath' of
         // '/app/classes/com/test/HelloWorld.class', the resolved destination would be
         // 'target/jib-docker-context/classes/com/test/HelloWorld.class'.
-        Path destination =
-            directoryInContext.resolve(
-                Paths.get(copyDirective.extractionPath).relativize(layerEntry.getExtractionPath()));
+        Path baseExtractionPath = Paths.get(copyDirective.extractionPath.toString());
+        Path relativeEntryPath = baseExtractionPath.relativize(layerEntry.getExtractionPath());
```
coollog: Hmm, maybe we can use `copyDirective.extractionPath` directly and add a `relativize` method to `AbsoluteUnixPath` after we change `LayerEntry#getExtractionPath` to return an `AbsoluteUnixPath`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
```diff
@@ -16,29 +16,29 @@
...
 import java.util.List;
 
 /** Constructs an image entrypoint for the Java application. */
 public class JavaEntrypointConstructor {
 
-  public static final String DEFAULT_RELATIVE_RESOURCES_PATH_ON_IMAGE = "resources/";
-  public static final String DEFAULT_RELATIVE_CLASSES_PATH_ON_IMAGE = "classes/";
-  public static final String DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE = "libs/";
+  public static final RelativeUnixPath DEFAULT_RELATIVE_RESOURCES_PATH_ON_IMAGE =
+      RelativeUnixPath.get("resources/");
```
coollog: Can remove the trailing slash.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaEntrypointConstructor.java
```diff
@@ -16,29 +16,29 @@
...
-
+      AbsoluteUnixPath appRoot, List<String> jvmFlags, String mainClass) {
     return makeEntrypoint(
         Arrays.asList(
-            appRoot + DEFAULT_RELATIVE_RESOURCES_PATH_ON_IMAGE,
-            appRoot + DEFAULT_RELATIVE_CLASSES_PATH_ON_IMAGE,
-            appRoot + DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE + "*"),
+            appRoot.resolve(DEFAULT_RELATIVE_RESOURCES_PATH_ON_IMAGE).toString(),
+            appRoot.resolve(DEFAULT_RELATIVE_CLASSES_PATH_ON_IMAGE).toString(),
+            appRoot.resolve(DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE).toString() + "/*"),
```
coollog: This can be `appRoot.resolve(DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE).resolve(RelativeUnixPath.get("*")).toString()` to avoid the explicit string concatenation.

jib-core/src/test/resources/sampleDockerfile
```diff
@@ -1,10 +1,10 @@
 FROM somebaseimage
 
-COPY libs /app/libs/
-COPY snapshot-libs /app/libs/
-COPY resources /app/resources/
-COPY classes /app/classes/
-COPY exploded-war /exploded/war/
+COPY libs /app/libs
```
coollog: I think we actually need a trailing slash.

From https://docs.docker.com/engine/reference/builder/#copy:
> `If <dest> does not end with a trailing slash, it will be considered a regular file and the contents of <src> will be written at <dest>.`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
chanseokoh: Oh, good catch.

jib-core/src/test/resources/sampleDockerfile
chanseokoh: Oh, good catch.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
chanseokoh: Perhaps we can think about this after #1014 (or in #1014).

PR #1014, opened 2018-09-19
 Changes LayerEntry#extractionPath to use AbsoluteUnixPath.

### LINKED ISSUE #1009: Change LayerEntry#extractionPath to use the new AbsoluteUnixPath.
This is a followup for #995

coollog:
Fixes #1009


chanseokoh:
Meger with master, and I should be able to look at this.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
```diff
@@ -98,6 +98,42 @@ public AbsoluteUnixPath resolve(RelativeUnixPath relativeUnixPath) {
...
+  // TODO: Add test.
+  public AbsoluteUnixPath resolve(Path relativePath) {
+    Preconditions.checkArgument(
+        relativePath.getRoot() == null, "Cannot resolve against absolute Path: " + relativePath);
+
+    ImmutableList.Builder<String> newPathComponents =
+        ImmutableList.builderWithExpectedSize(pathComponents.size() + relativePath.getNameCount());
+    newPathComponents.addAll(pathComponents);
+    for (Path pathComponent : relativePath) {
+      if (pathComponent.toString().isEmpty()) {
```
chanseokoh: You needed this when parsing `String` directly, but I don't think it's needed here. For example, `System.out.println(Paths.get("///a///a///a//").getNameCount())` return 3.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
```diff
@@ -98,6 +98,42 @@ public AbsoluteUnixPath resolve(RelativeUnixPath relativeUnixPath) {
...
+        // Skips empty components.
+        continue;
+      }
+      newPathComponents.add(pathComponent.toString());
+    }
+    return new AbsoluteUnixPath(newPathComponents.build());
```
chanseokoh: If going for simplicity, I guess `AbsoluteUnixPath.fromPath(Paths.get(unixPath).resolve(relativePath))` will work? The code of `fromPath()` is quite similar. (We still need to check that `relativePath` is not absolute.)

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
```diff
@@ -98,6 +98,42 @@ public AbsoluteUnixPath resolve(RelativeUnixPath relativeUnixPath) {
...
   }
 
+  /**
+   * Resolves this path against another relative path (by the name elements of {@code
+   * relativePath}).
+   *
+   * @param relativePath the relative path to resolve against
+   * @return a new {@link AbsoluteUnixPath} representing the resolved path
+   */
+  // TODO: Add test.
```
chanseokoh: I think we should add tests in this PR, unless you will add these method in another PR.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
coollog: I'm using this mostly for `Paths.get("")`, which would return a `nameCount` of 1, but also as a just-in-case since it's not guaranteed that name elements are not empty for different implementations of `Path`.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
coollog: Oops, I've already added tests and forgot to take this out.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/AbsoluteUnixPath.java
chanseokoh: Huh, TIL about `Paths.get("")` having an empty filename.

coollog: However, since resolving against a `""` path doesn't add another name element, not having this in `fromPath` works.

coollog: *having this in the call to `fromPath` in `resolve(Path relativePath)`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -286,7 +286,8 @@ public void generate(Path targetDirectory) throws IOException {
         // '/app/classes/com/test/HelloWorld.class', the resolved destination would be
         // 'target/jib-docker-context/classes/com/test/HelloWorld.class'.
         Path baseExtractionPath = Paths.get(copyDirective.extractionPath.toString());
-        Path relativeEntryPath = baseExtractionPath.relativize(layerEntry.getExtractionPath());
+        Path relativeEntryPath =
+            baseExtractionPath.relativize(Paths.get(layerEntry.getExtractionPath().toString()));
```
chanseokoh: I think we can now use `copyDirective.extractionPath.resolve(relativeEtryPath)`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -124,7 +118,7 @@ public String getAbsoluteSourceFileString() {
    * @return the extraction path
    */
   public String getAbsoluteExtractionPathString() {
```
chanseokoh: The Javadoc says "The path is made absolute first", but `extractionPath` is always absolute. Let's update it.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheMetadataTest.java
```diff
@@ -177,7 +187,8 @@ public void testFilter_byNoEntries() throws CacheMetadataCorruptedException {
 
     LayerEntry fakeLayerEntry =
         new LayerEntry(
-            Paths.get("some/source/file", "some/source/directory"), Paths.get("extractionPath"));
+            Paths.get("some/source/file", "some/source/directory"),
```
chanseokoh: Unrelated, but this example seems weird, which just results in a path of `/some/source/file/some/source/directory`. Maybe the intention in the past was to pass two `Path` as a list.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
coollog: I'm not sure what you mean here - the `copyDirective.extractionPath` is converted into the `baseExtractionPath` in order to relativize against the `layerEntry.getExtractionPath`. I decided not to implement the relativize method on `AbsoluteUnixPath` here since it seemed overkill for this one scenario.

chanseokoh: Ah, thought it was `resolve` than `relativize`.

PR #1015, opened 2018-09-20
Add "format" target

chanseokoh:
Sometimes I want to do a quick formatting.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,8 +9,8 @@ usage()
 {
   eval 1>&2
   echo "Simple builder for Jib for jib-core, jib-plugins-common, jib-maven-plugin, and jib-gradle-plugin"
-  echo "use: $0 [-qe] [clean | core | plugins | maven | gradle | all | it]"
-  echo "  'all' is the same as 'core plugins gradle maven'"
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -32,13 +32,13 @@ while getopts qe c; do
   \?) usage;;
   esac
 done
-shift `expr $OPTIND - 1`
+shift $( expr $OPTIND - 1 )
```
TadCordle: Newby question: what's the difference?

build.sh
```diff
@@ -32,13 +32,13 @@ while getopts qe c; do
...
 
 if [ $# -eq 0 ]; then
   set -- core plugins gradle maven
 fi
 
-set -e  # exit on error
+set -o errexit  # exit on error
```
TadCordle: Second newby question: what's the difference here as well?

chanseokoh: Same, `-e` is just a shortcut.

build.sh
chanseokoh: They are same, but `$( )` has the advance that you can nest. According to http://mywiki.wooledge.org/BashFAQ/082, `$( )` is slightly preferred.

PR #1021, opened 2018-09-21
Add _jibSkaffoldFiles task to gradle plugin

### LINKED ISSUE #864: Create a task/goal to display jib source files

### LINKED ISSUE #977: Add skaffold helpers to jib-gradle-plugin
Add helpers so any integration with skaffold can extract the necessary information.
This should be less involved than the maven plugin (#958) since gradle tasks define their dependencies and gradle multimodule resolution is a lot more robust.
 
~~1. See if multimodule or not~~ (this can be done directly with `gradle projects`)
2. Get all source files to watch (`_jibSkaffoldFiles`)
~~3. Get jib goals on package phase~~ (not necessary, jib gradle goals define all their dependencies)

TadCordle:
Fixes #977
Fixes #864

Adds an internal `_jibSkaffoldFiles` task that prints a list of changing files in a jib project (build scripts, source folder, resource folders, extra layer directory, and SNAPSHOT dependency jars). Recursively handles project dependencies as well to print the source files instead of the jar.


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,140 @@
...
+    this.jibExtension = jibExtension;
+    return this;
+  }
+
+  @TaskAction
+  public void listFiles() {
+    Preconditions.checkNotNull(jibExtension);
+    Project project = getProject();
+
+    // If this is not the root project, print the root project's build.gradle and settings.gradle
```
coollog: `gradle.properties` might be good to list as well

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,140 @@
...
+    if (project != project.getRootProject()) {
+      System.out.println(project.getRootProject().getBuildFile());
+      Path settingsFiles = project.getRootDir().toPath().resolve(Settings.DEFAULT_SETTINGS_FILE);
+      if (Files.exists(settingsFiles)) {
+        System.out.println(settingsFiles);
+      }
+    }
+
+    // Print extra layer
+    if (project.getPlugins().hasPlugin(JibPlugin.class)) {
```
coollog: What is the reasoning for checking if there is `JibPlugin`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,140 @@
...
+      System.out.println(jibExtension.getExtraDirectoryPath());
+    }
+
+    // Print sub-project sources
+    Set<File> skippedJars = new HashSet<>();
+    listFilesForProject(project, skippedJars);
```
coollog: This will print the project's build and settings file again?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,140 @@
...
+
+    // Print out SNAPSHOT non-project dependency jars
+    for (File file : project.getConfigurations().getByName("runtime")) {
+      if (!skippedJars.contains(file) && file.toString().contains("SNAPSHOT")) {
+        System.out.println(file);
+        skippedJars.add(file);
```
coollog: It's a bit hard to follow the logic behind using the `Set<File> skippedJars` here. It might be good to isolate the mutation of `skippedJar` to a more encapsulated location, since `skippedJars` seems to be mutated here, and inside `listFilesForProject`. Also, in `listFilesForProject`, the set is called `projectDependencyJars`, but here it is called `skippedJars`. A comment would help explain the discretion in naming.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/FilesTaskTest.java
```diff
@@ -0,0 +1,137 @@
...
+   * @param expected the expected list of paths
+   * @param actual the actual list of paths
+   * @throws IOException if checking if two files are the same fails
+   */
+  private static void assertPathListsAreEqual(List<Path> expected, List<Path> actual)
+      throws IOException {
+    Assert.assertEquals(expected.size(), actual.size());
+    for (int index = 0; index < expected.size(); index++) {
+      Assert.assertEquals(
+          expected.get(index).toFile().getCanonicalFile(),
```
coollog: I believe you can use `Path#toRealPath`

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/FilesTaskTest.java
```diff
@@ -0,0 +1,137 @@
...
+  @Test
+  public void testFilesTask_singleProject() throws IOException {
+    Path projectRoot = simpleTestProject.getProjectRoot();
+    List<Path> result = verifyTaskSuccess(simpleTestProject, null);
+    List<Path> expected =
+        ImmutableList.of(
+            projectRoot.resolve("src/main/custom-extra-dir"),
+            projectRoot.resolve("build.gradle"),
+            projectRoot.resolve("src/main/java"),
+            projectRoot.resolve("src/main/resources"));
```
coollog: Should this have the dependency jar?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/FilesTaskTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.gradle.testkit.runner.BuildResult;
+import org.gradle.testkit.runner.BuildTask;
+import org.gradle.testkit.runner.TaskOutcome;
+import org.hamcrest.CoreMatchers;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+/** Tests for {@link FilesTask}. */
+public class FilesTaskTest {
```
coollog: Might be good to include a test for a project that would trigger the recursive call to `listFilesForProject`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
TadCordle: The mojo does this, I'm not sure if it's necessary here.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
TadCordle: No, the root build/settings files are only printed initially if the task is called on a sub-project.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/FilesTaskTest.java
TadCordle: Not currently, since it's not a SNAPSHOT. Issue is open here #1027

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/FilesTaskTest.java
TadCordle: `testFilesTask_multiProjectComplexService` does this since lib/ is a project dependency.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
TadCordle: The reason it's called `skippedJars` here and `projectDependencyJars` within the method is because within the method, the set is only used to keep track of project dependency jars, while here the set is also used to avoid printing the same files twice. I'll rethink this or add a comment.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,140 @@
...
+    if (Files.exists(settingsFile)) {
+      System.out.println(settingsFile);
+    }
+
+    // Print sources
+    mainSourceSet.getAllJava().getSourceDirectories().forEach(System.out::println);
+
+    // Print resources
+    mainSourceSet.getResources().getSourceDirectories().forEach(System.out::println);
+
```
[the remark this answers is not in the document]
TadCordle: Not sure what you mean by this? This method only keeps track of dependencies where we'd rather watch the project source instead of the jar. We actually print jars in `listFiles()` after the recursion excludes all the project dependency jars.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Recursive function for printing out a project's artifacts. Calls itself when it encounters a
+   * project dependency.
+   *
+   * @param project the project to list the artifacts for
+   * @param projectDependencyJars the set of jar files associated with each project dependency
+   *     encountered
+   */
+  private static void listFilesForProject(Project project, Set<File> projectDependencyJars) {
```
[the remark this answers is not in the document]
TadCordle: Returning a `Set` rather than passing a mutable `Set` as a parameter seemed more complicated initially, maybe I can look into it again.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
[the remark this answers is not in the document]
coollog: Ah yes that makes sense now - the subprojects' dependency JARs is for skipping, not listing. Although, why are subprojects' dependency JARs skipped?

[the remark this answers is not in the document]
TadCordle: If you mean the project dependencies, it's because we watch their source/resources/etc. instead of the jars.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,140 @@
...
+    mainSourceSet.getAllJava().getSourceDirectories().forEach(System.out::println);
+
+    // Print resources
+    mainSourceSet.getResources().getSourceDirectories().forEach(System.out::println);
+
+    // Find project dependencies
+    for (Configuration configuration :
+        project.getConfigurations().getByName("runtime").getHierarchy()) {
+      for (Dependency dependency : configuration.getDependencies()) {
+        if (dependency instanceof ProjectDependency) {
```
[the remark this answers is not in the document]
TadCordle: I believe these get picked up in `listFiles()` if they're snapshots.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
loosebazooka: That was because maven is weird and needed to run the goal directly on the other projects (we couldn't access enough information from the project under consideration). And we needed to make sure non-jib modules didn't give us jib data.

But with gradle just lets us go inspect all the sub-projects, so it shouldn't be necessary here.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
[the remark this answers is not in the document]
loosebazooka: All dependency jars are resolved in the project under consideration: https://github.com/GoogleContainerTools/jib/pull/1021/files#diff-0ba9cbdfb012fc358d5e50ed9c05c410R133

It does the resolution and transitive resolution.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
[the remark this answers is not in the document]
loosebazooka: Would it be easier to read if we just pulled out all projectConfigurations in one step? Either recursive or some kind of iterative dfs/bfs. Then process each configuration after that?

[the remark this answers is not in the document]
TadCordle: Hmm, do you mean pulling L109-120 out of the recursive function? Wouldn't that still involve maintaining a set to pull out the project dependencies?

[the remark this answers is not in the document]
loosebazooka: Yeah, you would, I mean the set is inevitable since you have to track it. And I think my point is different than Q's. So I probably shouldn't have continued on this stream.

But in the end, this function is doing two separate things (printing and collecting), and it might be a little easier to read if they were separate.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,174 @@
...
+   * @param project the project
+   */
+  private static void printGradleFiles(Project project) {
+    Path projectPath = project.getProjectDir().toPath();
+
+    // Print build.gradle
+    System.out.println(project.getBuildFile());
+
+    // Print settings.gradle
+    Path settingsFiles = projectPath.resolve(Settings.DEFAULT_SETTINGS_FILE);
```
loosebazooka: Naming: settingsFile**s** -> settingsFile
Also, don't expect a lot of customization here, but we can check if the settings file was configured on the command line.

```
getProject().getGradle().getStartParameter().getSettingsFile() != null
```
and use that instead of default.

TadCordle: Is there an equivalent for gradle.properties?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * Recursive function for printing out a project's artifacts. Calls itself when it encounters a
+   * project dependency.
+   *
+   * @param project the project to list the artifacts for
+   * @param projectDependencies the set of project dependencies encountered. When a project
+   *     dependency is encountered, it is added to this set.
+   */
+  private static void findProjectDependencies(
+      Project project, Set<ProjectDependency> projectDependencies) {
```
loosebazooka: maybe projectDependenciesResult or something along those lines?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,174 @@
...
+    printProjectFiles(project);
+
+    // Print extra layer
+    if (Files.exists(jibExtension.getExtraDirectoryPath())) {
+      System.out.println(jibExtension.getExtraDirectoryPath());
+    }
+
+    // Find project dependencies
+    Set<ProjectDependency> projectDependencies = new HashSet<>();
+    findProjectDependencies(project, projectDependencies);
```
coollog: nit: Usually prefer a more functional style: `Set<ProjectDependency> projectDependencies = findProjectDependencies(project);`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -0,0 +1,174 @@
...
+
+    Set<File> projectDependencyJars = new HashSet<>();
+    for (ProjectDependency projectDependency : projectDependencies) {
+      printProjectFiles(projectDependency.getDependencyProject());
+
+      // Keep track of project dependency jars for filtering out later
```
coollog: Should this comment on on line 148?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
loosebazooka: Perhaps in another PR we can flatten it out using a non-recursive BFS or DFS (So you're not constantly creating and merging sets).

TadCordle: Filed #1033

PR #1022, opened 2018-09-21
Finishes implementing new Cache.

coollog:
Part of #637
Continuation from #975

- Finishes implementing `Cache` (instantiation with `withDirectory`, and `retrieve/write` methods)
- Implements `DefaultCacheStorageReader#select` as well.


jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
```diff
@@ -16,22 +16,82 @@
...
+   *
+   * @param cacheDirectory the directory for the cache. Creates the directory if it does not exist.
+   * @return a new {@link Cache}
+   * @throws NotDirectoryException if {@code cacheDirectory} is not a directory
+   * @throws IOException if an I/O exception occurs
+   */
+  public static Cache withDirectory(Path cacheDirectory) throws IOException {
+    if (Files.exists(cacheDirectory)) {
+      if (!Files.isDirectory(cacheDirectory)) {
+        throw new NotDirectoryException("The cache can only write to a directory");
```
loosebazooka: Not right or wrong, but createDirectories throws FileAlreadyExistsException in this case: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createDirectories(java.nio.file.Path,%20java.nio.file.attribute.FileAttribute...)

It also uses some "optional specific exception" hint in the java doc.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWrite.java
```diff
@@ -21,7 +21,7 @@
 import java.util.Optional;
 
 /** Represents layer data to write to the cache. <b>Implementations must be immutable.</b> */
-public interface CacheWrite {
+interface CacheWrite {
```
loosebazooka: It looks like CacheWrite isn't part of the API and may not need to be an interface? We don't seem to really allow customizing it and most creation events of CacheWrite are directly from calling DefaultCacheWrite under the hood.

I would think, given how we've implemented it here, that CacheWrite could just be what DefaultCacheWrite is now doing?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
loosebazooka: My actual comment is, should we just use `FileAlreadyExistsException` here?

coollog: Yea, that sounds good.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheWrite.java
coollog: Yep, I'll remove the interface.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
```diff
@@ -16,22 +16,73 @@
...
+import javax.annotation.concurrent.Immutable;
 
-/** Cache for storing data to be shared between Jib executions. */
+/**
+ * Cache for storing data to be shared between Jib executions.
+ *
+ * <p>Uses the default cache storage engine ({@link DefaultCacheStorage}) and layer entries as the
+ * selector ({@link LayerEntriesSelector}).
+ *
+ * <p>Implementation is immutable and thread-safe.
```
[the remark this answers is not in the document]
loosebazooka: Yeah we might want to work on wording here. The object is immutable, but it does in-fact mutate the underlying storage mechanism.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
```diff
@@ -16,22 +16,73 @@
...
+   * @return the {@link CacheEntry} for the written {@link CacheWrite}
+   * @throws IOException if an I/O exception occurs
+   */
+  public CacheEntry write(Blob layerBlob) throws IOException {
+    return cacheStorage.write(CacheWrite.layerOnly(layerBlob));
+  }
+
+  /**
+   * Saves a cache entry with a layer {@link Blob}, an additional selector digest, and a metadata
+   * {@link Blob}. Use {@link #write(Blob)} to save only a layer {@link Blob}.
```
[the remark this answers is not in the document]
coollog: I'll actually change this to take `LayerEntry`s instead so that selectors are just an implementation detail of the cache. *Reference* can be an alternative name, but that may be confusing since it is overloaded with both the meaning of *image reference* and the *reference digest*.

jib-core/src/test/java/com/google/cloud/tools/jib/ncache/CacheTest.java
```diff
@@ -0,0 +1,251 @@
...
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/** Tests for {@link Cache}. */
+public class CacheTest {
```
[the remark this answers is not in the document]
coollog: I think the methods that would be used in the implementation would be `compress` and `sizeOf` (by `DefaultCacheStorageWriter#writeLayerBlobToDirectory`. I'll try to refactor these in a followup PR - possibly with some Blob-manipulation utilities in the `blob` package. #1030

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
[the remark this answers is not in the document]
briandealwis: Sounds good.

PR #1023, opened 2018-09-22
Remove ListOfJsonTemplate class

### LINKED ISSUE #997: Consider removing ListOfJsonTemplate.
From the thread at https://github.com/GoogleContainerTools/jib/pull/975/files/e810e02988681b9d6f9f06a6fd6565f6e4c778c8#r218225233, it seems that `ListOfJsonTemplate` is a misnomer and can be removed.

chanseokoh:
Fixes #997.


jib-core/src/main/java/com/google/cloud/tools/jib/docker/json/DockerLoadManifestTemplate.java
```diff
@@ -43,7 +43,7 @@
  * @see <a href="https://github.com/moby/moby/blob/master/image/tarexport/load.go">Docker load
  *     source</a>
  */
-public class DockerLoadManifestTemplate implements ListOfJsonTemplate {
+public class DockerLoadManifestTemplate implements ListOfJsonTemplate<JsonTemplate> {
```
coollog: I think `ListOfJsonTemplate<DockerLoadManifestTemplate>` would be more applicable here.

jib-core/src/main/java/com/google/cloud/tools/jib/json/JsonTemplateMapper.java
```diff
@@ -132,7 +132,7 @@ public static Blob toBlob(JsonTemplate template) {
    * @param template the list of JSON templates to convert
    * @return a {@link Blob} of the JSON string
    */
-  public static Blob toBlob(ListOfJsonTemplate template) {
+  public static Blob toBlob(ListOfJsonTemplate<?> template) {
```
coollog: should this be `? extends JsonTemplate`?

jib-core/src/main/java/com/google/cloud/tools/jib/docker/json/DockerLoadManifestTemplate.java
chanseokoh: This sounds weird. The actual `DockerLoadManifestTemplate` class matches and represents a single Docker Manifest element. But `ListofJsonTemplate<itself>` also implies that the class can act as a list of Docker Manifests. This sounds confusing to me. I think I'll fix #997, and things should be significantly simplified.

coollog: Sounds good. Let me know if that works out well - I feel like we had a reason we had to add the `ListOfJsonTemplate` class.

chanseokoh:
Updated the PR to fix #997. `DockerLoadManifestTemplate` now represents a single element, not a list.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/LayerEntriesSelector.java
```diff
@@ -92,30 +90,9 @@ public int hashCode() {
...
-     */
-    @VisibleForTesting
-    LayerEntriesTemplate(ImmutableList<LayerEntry> layerEntries) {
-      for (LayerEntry layerEntry : layerEntries) {
-        layerEntryTemplates.add(new LayerEntryTemplate(layerEntry));
-      }
-      Collections.sort(layerEntryTemplates);
-    }
+  static List<LayerEntryTemplate> toSortedJsonTemplates(List<LayerEntry> layerEntries) {
+    return layerEntries.stream().map(LayerEntryTemplate::new).sorted().collect(Collectors.toList());
```
coollog: I think we might want to keep it non-stream style since we might not want to incur the stream overhead in this class as it one of the lower-level methods.

jib-core/src/test/resources/json/loadmanifest.json
```diff
@@ -1 +1 @@
-[{"config":"config.json","repoTags":["testregistry/testrepo:testtag"],"layers":["layer1.tar.gz","layer2.tar.gz","layer3.tar.gz"]}]
```
coollog: This should also keep the `[]` since without it it isn't a valid loadable manifest.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/json/DockerLoadManifestTemplate.java
```diff
@@ -29,21 +28,21 @@
  * <p>Example manifest JSON:
  *
  * <pre>{@code
- * [{
+ * {
```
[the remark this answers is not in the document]
chanseokoh: I'll rename the class to `DockerLoadManifestEntryTemplate` and update the Javadoc. Then I don't think we'll need `DockerLoadManifestTemplate` anymore, since it'd be just `List<DockerLoadManifestEntryTemplate>`.

PR #1024, opened 2018-09-22
Re-enable mount-from

### LINKED ISSUE #169: Have cross-repo mount automatically put source repository for base image in same registry.
If the base image and the target image are both in the same registry, we can add a `&from=` when pushing a blob to use the base image repository to find the blob to mount.

We might also want to rearrange the build steps to:

1) Pull base image manifest
2) Begin upload of base image layers (get upload location) with mount, if on same registry.
3) Pull base image layers and finish upload if not on same registry, or mount failed.

briandealwis:
Re-enables `mount`/`from` for #169.  The speed-up can be immense.  Building and pushing an image with `openjdk:8` as the base image to a new Docker Hub image takes 3:41 *minutes*, whereas with `mount`/`from` it took 7.976s.

  - updated `AuthenticatePushStep` to obtains `pull` scope for the base repository if the base and target registries are the same
  - update `PushBlobStep` to pass along the base repository as the mount location if the base and target registries are the same
  - add a system property `jib.useMountFrom` to disable using `mount`/`from` in case somebody hits a registry that doesn't support it (plus it was useful for manual testing)
  - lightly reworked `RegistryAuthenticator` to pass around the `RegistryEndpointRequestProperties` object where possible, and enhanced `RegistryEndpointRequestProperties` to include a set of dependent images

I think I over-generalized `RegistryEndpointRequestProperties` and should just maintain a single additional image rather than a set of images.

Fixes #169


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -102,12 +105,20 @@ public RegistryAuthenticator initialize()
...
+   *     requested with {@code pull} scope; may be {@code null} or empty
    * @return the new {@link Initializer}
    */
   public static Initializer initializer(
-      JibLogger buildLogger, String serverUrl, String repository) {
-    return new Initializer(buildLogger, serverUrl, repository);
+      JibLogger buildLogger,
+      String serverUrl,
+      String repository,
+      @Nullable String... dependentRepositories) {
```
coollog: Hmm.. It feels a bit off that this is both a varargs and nullable. Perhaps this should be two different `initializer` methods or it should take a `List<String>` instead (so the empty case means none)?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
```diff
@@ -83,8 +85,13 @@ public BlobDescriptor call() throws IOException, RegistryException, ExecutionExc
         return blobDescriptor;
       }
 
-      // todo: leverage cross-repository mounts
-      registryClient.pushBlob(blobDescriptor.getDigest(), blob, null);
+      // If base and target images are in the same registry, then use mount/from to try mounting the
+      // BLOB from the base image repository to the target image repository and possibly avoid
+      // having to push the BLOB. See the following for details:
+      // https://docs.docker.com/registry/spec/api/#cross-repository-blob-mount
+      // Must also be checked in {@link AuthenticatePushStep}
```
coollog: > Must also be checked in {@link AuthenticatePushStep}

Is this a left-over comment?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
```diff
@@ -83,8 +85,13 @@ public BlobDescriptor call() throws IOException, RegistryException, ExecutionExc
...
+      Optional<String> mountFrom = MountFromSupport.getMountFrom(buildConfiguration);
+      registryClient.pushBlob(blobDescriptor.getDigest(), blob, mountFrom.orElse(null));
```
coollog: Not sure, but I think it might be preferable to use `isPresent` and `get` over defaulting to `null` when retrieving the value from an `Optional`, so this would be more clearly written as:

```java
if (mountFrom.isPresent()) {
  registryClient.pushBlobWithMount(blobDescriptor.getDigest(), blob, mountFrom.get());
} else {
  registryClient.pushBlob(blobDescriptor.getDigest(), blob);
}
```

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -31,6 +31,8 @@
 
   private static final String DISABLE_USER_AGENT = "_JIB_DISABLE_USER_AGENT";
 
+  @VisibleForTesting static final String USE_MOUNT_FROM = "jib.useMountFrom";
```
coollog: tiny nit: package-private above private.

coollog: May consider naming closer to the Docker term `Cross Repository Blob Mount`?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -249,7 +254,18 @@ public Authorization authenticatePush() throws RegistryAuthenticationFailedExcep
 
   @VisibleForTesting
   URL getAuthenticationUrl(String scope) throws MalformedURLException {
-    return new URL(authenticationUrlBase + scope);
+    StringBuilder urlBase = new StringBuilder(authenticationUrlBase);
```
coollog: nit: `urlBuilder`?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -163,7 +163,19 @@ private String makeUserAgent() {
...
+  }
+
+  /**
+   * Creates a new {@link Factory} for building a {@link RegistryClient}.
+   *
+   * @param buildLogger the build logger used for printing messages
+   * @param registryEndpointRequestProperties properties of registry endpoint requests
+   * @return the new {@link Factory}
+   */
+  public static Factory factory(
```
coollog: nit: package-private?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointRequestProperties.java
```diff
@@ -21,14 +21,19 @@
 
   private final String serverUrl;
   private final String imageName;
+  private final String[] dependentImageNames;
```
coollog: I think in this codebase, we prefer to use `List` interface over raw arrays.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointRequestProperties.java
```diff
@@ -21,14 +21,19 @@
...
 
   /**
    * @param serverUrl the server URL for the registry (for example, {@code gcr.io})
    * @param imageName the image/repository name (also known as, namespace)
+   * @param dependentImageNames any additional images required for access; may be {@code null} or
```
coollog: Is `required for access` the right explanation here? Maybe it's more like `any additional images that may provide blobs to mount into the {@code imageName} repository`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
briandealwis: Sadly no, but I'll reword it slightly.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
briandealwis: I didn't like this either.  What I really want to put is a map of _repository &rarr; scope_.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
```diff
@@ -83,8 +85,13 @@ public BlobDescriptor call() throws IOException, RegistryException, ExecutionExc
...
       }
 
-      // todo: leverage cross-repository mounts
-      registryClient.pushBlob(blobDescriptor.getDigest(), blob, null);
+      // If base and target images are in the same registry, then use mount/from to try mounting the
+      // BLOB from the base image repository to the target image repository and possibly avoid
+      // having to push the BLOB. See the following for details:
+      // https://docs.docker.com/registry/spec/api/#cross-repository-blob-mount
+      // Must also be checked in {@link AuthenticatePushStep}
+      Optional<String> mountFrom = MountFromSupport.getMountFrom(buildConfiguration);
```
chanseokoh: Where is this class?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
```diff
@@ -83,8 +85,13 @@ public BlobDescriptor call() throws IOException, RegistryException, ExecutionExc
         return blobDescriptor;
       }
 
-      // todo: leverage cross-repository mounts
-      registryClient.pushBlob(blobDescriptor.getDigest(), blob, null);
+      // If base and target images are in the same registry, then use mount/from to try mounting the
```
chanseokoh: Haven't looked into the code yet, so asking at a high level: what if the base and target images are in the same registry but are in different repositories requiring different auths? Is that case covered?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
briandealwis: I didn't commit it (blush).  Will upload in next commit with review fixes.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
briandealwis: That's the note below, which I'll reword.

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -27,12 +27,13 @@
   @VisibleForTesting
   public static final String SEND_CREDENTIALS_OVER_HTTP = "sendCredentialsOverHttp";
 
+  @VisibleForTesting
+  static final String USE_CROSS_REPOSITORY_BLOB_MOUNTS = "jib.useCrossRepositoryBlobMounts";
```
coollog: Maybe this can be shortened to just `jib.useBlobMounts`? (There's currently no other types of blob mounts)

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -254,18 +255,21 @@ public Authorization authenticatePush() throws RegistryAuthenticationFailedExcep
...
+    urlBuilder
         .append("&scope=repository:")
         .append(registryEndpointRequestProperties.getImageName())
         .append(':')
         .append(scope);
 
-    for (String repository : registryEndpointRequestProperties.getDependentImageNames()) {
-      urlBase.append("&scope=repository:").append(repository).append(":pull");
+    if (registryEndpointRequestProperties.getMountedImageName() != null) {
+      urlBuilder
```
coollog: To actually authenticate for the pull, I believe we would need to add the authentication credentials for the base image.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
chanseokoh: > what if the base and target images are in the same registry but are in different repositories requiring different auths?

I don't think this case works.

Below works, showing that I can pull and push to my `francium25` repo.
```xml
          <from>
            <image>francium25/my-test</image>
            <auth>
              <username>francium25</username><password>...</password>
            </auth>
          </from>
          <to>
            <image>francium25/my-test2</image>
            <auth>
              <username>francium25</username><password>...</password>
            </auth>
          </to>
```

Below also works, showing that I can push to another repo of mine, `francium25paran`:
```xml
          <!--from>
            <image>francium25/my-test</image>
            <auth>
              <username>francium25</username><password>...</password>
            </auth>
          </from-->
          <to>
            <image>francium25paran/my-test2</image>
            <auth>
              <username>francium25paran</username><password>...</password>
            </auth>
          </to>
```

However, if I uncomment the `<from>` section above, it doesn't work. The reason is that `francium25` and `francium25paran` require different auths. Checking if they are on the same registry isn't enough.

```
com.google.cloud.tools.jib.registry.RegistryUnauthorizedException: Unauthorized for registry.hub.docker.com/francium25paran/my-test2
        at com.google.cloud.tools.jib.registry.RegistryEndpointCaller.call(RegistryEndpointCaller.java:268)
        at com.google.cloud.tools.jib.registry.RegistryEndpointCaller.callWithAllowInsecureRegistryHandling(RegistryEndpointCaller.java:151)
        at com.google.cloud.tools.jib.registry.RegistryEndpointCaller.call(RegistryEndpointCaller.java:141)
        at com.google.cloud.tools.jib.registry.RegistryClient.callRegistryEndpoint(RegistryClient.java:367)
        at com.google.cloud.tools.jib.registry.RegistryClient.pushBlob(RegistryClient.java:316)
        at com.google.cloud.tools.jib.builder.steps.PushBlobStep.call(PushBlobStep.java:94)
        at com.google.cloud.tools.jib.builder.steps.PushBlobStep.call(PushBlobStep.java:36)
        at com.google.common.util.concurrent.CombinedFuture$CallableInterruptibleTask.runInterruptibly(CombinedFuture.java:181)
        at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:57)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: com.google.api.client.http.HttpResponseException: 401 Unauthorized
{"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Type":"repository","Class":"","Name":"francium25paran/my-test2","Action":"pull"},{"Type":"repository","Class":"","Name":"francium25paran/my-test2","Action":"push"},{"Type":"repository","Class":"","Name":"francium25/my-test","Action":"pull"}]}]}

        at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:1070)
        at com.google.cloud.tools.jib.http.Connection.send(Connection.java:161)
        at com.google.cloud.tools.jib.registry.RegistryEndpointCaller.call(RegistryEndpointCaller.java:229)
        ... 11 more
```

chanseokoh: Oh, and for testing, you might want to temporarily force skipping the BLOB availability check.

```
      // check if the BLOB is available
      if (false && registryClient.checkBlob(blobDescriptor.getDigest()) != null) { ...
```

briandealwis: I misunderstood.  Major bummer.

chanseokoh: Yeah, I'm afraid that's a major bummer that may require much more non-trivial changes.

briandealwis:
Can't just compare the credential retrievers between the base and target image as the credentials for the target may allow accessing the base.  The only real way to test is to try authenticating for both the target and base and if that fails then retry with just the target.  

We could save the accessible repositories in `AuthenticatePushStep` and then access that detail in `PushBlobStep`.  Or we could track the accessible repositories in `Authorization` itself.

briandealwis:
Closing this in favour of a new implementation

PR #1031, opened 2018-09-25
 Has Cache use last modified time as metadata.

coollog:
For #637

This changes the new `Cache` to not expose metadata in its API. Rather, the metadata is generated as the latest last modified time of the `LayerEntry`s. Retrieve-by-layer-entries is also changed to retrieve only layers that match in last modified time (as stored by the metadata).

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.attribute.FileTime;
+import java.time.Instant;
 import java.util.Optional;
 import javax.annotation.concurrent.Immutable;
 
@@ -58,7 +59,8 @@ public class Cache {
   }
 
   /**
-   * Saves a cache entry with only a layer {@link Blob}. Use {@link #write(Blob, ImmutableList)} to include a selector and metadata.
+   * Saves a cache entry with only a layer {@link Blob}. Use {@link #write(Blob, ImmutableList)} to
+   * include a selector and metadata.
    *
    * @param layerBlob the layer {@link Blob}
    * @return the {@link CacheEntry} for the written layer
@@ -81,11 +83,15 @@ public class Cache {
       throws IOException {
     return cacheStorage.write(
         CacheWrite.withSelectorAndMetadata(
-            layerBlob, LayerEntriesSelector.generateSelector(layerEntries), LastModifiedMetadata.generateMetadata(layerEntries)));
+            layerBlob,
+            LayerEntriesSelector.generateSelector(layerEntries),
+            LastModifiedMetadata.generateMetadata(layerEntries)));
   }
 
   /**
-   * Retrieves the {@link CacheEntry} that was built from the {@code layerEntries}. The last modified time of the {@code layerEntries} must match the last modified time as stored by the metadata of the {@link CacheEntry}.
+   * Retrieves the {@link CacheEntry} that was built from the {@code layerEntries}. The last
+   * modified time of the {@code layerEntries} must match the last modified time as stored by the
+   * metadata of the {@link CacheEntry}.
    *
    * @param layerEntries the layer entries to match against
    * @return a {@link CacheEntry} that was built from {@code layerEntries}, if found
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
```diff
@@ -100,7 +106,26 @@ public CacheEntry write(Blob layerBlob, ImmutableList<LayerEntry> layerEntries,
...
+    }
+
+    CacheEntry cacheEntry = optionalCacheEntry.get();
+    if (!cacheEntry.getMetadataBlob().isPresent()) {
+      return Optional.empty();
+    }
+
+    Blob metadataBlob = cacheEntry.getMetadataBlob().get();
+    FileTime cacheEntryLastModifiedTime =
+        FileTime.from(Instant.parse(Blobs.writeToString(metadataBlob)));
```
chanseokoh: Hmm... so this assumes that this cache metadata `Blob` is an instance of `Instant`. I feel like `Blob` here is somewhat like the `Object` class in Java, obscuring anything inside it. Can't we have `CacheEntry.getLastSourceModifiedTime()`, much like we currently have `CacheEntry.getLayerSize()`, or have `getMetadata()` return a readable structure (i.e., return a Java class)?

coollog: I intended `CacheEntry` to have metadata as a generic metadata whose contents would be determined by the user. `Cache` is basically the default API implementation for Jib builder to use, and defines the metadata as the last modified time of the layers (via its helper `LastModifiedMetadata`). Having `getMetadata` return `Blob` is intended to allow for different implementations of the metadata (like perhaps a fast file checksum or some other information).

chanseokoh: Then, is it possible to use generics somehow?

coollog: Hmm, I think to have generics work, the cache retrieval (`DefaultCacheStorage`) would need to have knowledge of how to serialize/deserialize the metadata, which would either require the user providing a serialization/deserialization helper to `DefaultCacheStorage` or having `DefaultCacheStorage` use a specific serialization/deserialization method.

I think it might make sense to put this logic here in `LastModifiedMetadata`, like `LastModifiedMetadata#getLastModifiedTime(CacheEntry)`.

chanseokoh: Okay, I will leave it up to you, given that you have given enough thoughts on this.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/LastModifiedTimeMetadata.java
```diff
@@ -0,0 +1,83 @@
...
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.attribute.FileTime;
+import java.time.Instant;
+import java.util.Optional;
+
+/**
+ * Metadata that is the latest last modified time of all the source files in a list of {@link
+ * LayerEntry}s.
```
chanseokoh: Since we now have `getLastModifiedTime(CacheEntry)`, this Javadoc (saying "time of all the source files ... of `LayerEntry`s") doesn't fit very well.

coollog: This javadoc is referring to the actual contents of the metadata, which is the last modified time of layer entries. Getting the last modified time from a `CacheEntry` is just deserializing the metadata, which still is the last modified time of the layer entries that made up the layer.

chanseokoh: But all the methods are just static util methods. There is no field, and the class is stateless, not representing anything actually. So I'm not sold that this class is referring to the actual contents of the metadata, since there is no metadata actually associated with this class.

coollog: Hmm, I'll try rewording this. This is intended as a serialization/deserialization helper for metadata that stores the last modified time, in comparison to another possible `...Metadata` class that could serialize/deserialize for different contents.

Issue #1008, opened 2018-09-19
Cache metadata source files needs to store windows path for windows.


coollog:
Cache metadata for serialized source file paths should store the Windows-style path for Windows since the drive letter could be different and storing in Unix-style would not capture the difference.

From https://github.com/GoogleContainerTools/jib/pull/995#discussion_r218589436

coollog:
Irrelevant now that the new cache mechanism is in place.

chanseokoh:
Maybe we should remove `LayerEntry.getAbsoluteSourceFileString()` and have `LayerEntriesSelector.LayerEntryTemplate` just call `layerEntry.getSourcePath()` below? `getAbsoluteSourceFileString()` strips out the drive letter, so I feel like we shouldn't use or need it.

```java
    @VisibleForTesting
    LayerEntryTemplate(LayerEntry layerEntry) {
      sourceFile = layerEntry.getAbsoluteSourceFileString();
      extractionPath = layerEntry.getAbsoluteExtractionPathString();
    }
```

coollog:
@chanseokoh That's true - we should probably change `getAbsoluteSourceFileString` to return `sourceFile.toAbsolutePath().toString()` directly then since we still do want the absolute path rather than a relative path.

PR #1032, opened 2018-09-25
Use AbsoluteUnixPath for LayerEntry.getAbsoluteSourceFileString()

chanseokoh:
Making a quick try without any testing.


chanseokoh:
That's right. That should definitely fixed by #1008. This PR is more of a necessary clean-up before then, not changing any current behavior.

PR #1034, opened 2018-09-25
Start replacing JibLogger calls with Events

TadCordle:
Part of #714.

This PR is progress towards replacing `JibLogger` with the new event system. I haven't finished purging all uses of `JibLogger` yet, but the PR was getting large, so I decided to submit now.

In followups I'll finish the rest, work on replacing Timers with events, and eventually remove the old `Timer` and `JibLogger` classes.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -70,7 +70,7 @@ public void execute() throws MojoExecutionException {
       ImageReference targetImage =
           ConfigurationPropertyValidator.getGeneratedTargetDockerTag(
               getTargetImage(),
-              mavenJibLogger,
+              mavenProjectProperties.getEventEmitter(),
```
chanseokoh: Who's setting an `EventEmitter` and where? Is logging actually working? I believe there should exist code that adds a real event handler for `LogEvent` that actually stream logs to `AbstractMojo.getLog()`. I don't see such code.

chanseokoh: Oh, maybe the log handler is already added somewhere before this PR?

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
```diff
@@ -63,7 +67,12 @@ private static DockerCredentialHelperFactory getTestFactory(
     };
   }
 
-  @Mock private JibLogger mockJibLogger;
+  private StringBuilder logMessages = new StringBuilder();
+  private EventEmitter eventEmitter =
+      new DefaultEventEmitter(
+          new EventHandlers()
+              .add(JibEventType.LOGGING, logEvent -> logMessages.append(logEvent.getMessage())));
```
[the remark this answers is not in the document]
TadCordle: I admittedly did not try it. I was going off of a different test that checked log messages and it did it this way.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidatorTest.java
```diff
@@ -32,10 +37,20 @@
 @RunWith(MockitoJUnitRunner.class)
 public class ConfigurationPropertyValidatorTest {
 
-  @Mock private JibLogger mockLogger;
+  private StringBuilder logMessages;
```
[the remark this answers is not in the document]
TadCordle: I wanted to create a new one for each test.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
TadCordle: Yeah, that happens in `LogEventHandler` I think

chanseokoh: Yeah, I see `makeEventEmitter()` in `GradleProjectProperties` and `MavenProjectProperties` adds the log event handler. We're good.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidatorTest.java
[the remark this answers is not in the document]
chanseokoh: I believe a new instance of `ConfigurationPropertyValidatorTest` is created for each test. The fields won't be reused between tests.

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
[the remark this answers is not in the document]
TadCordle: It looks like it'll work if I add an `equals` to `LogEvent`. Which I think is worth it, since it makes the tests a lot more straightforward. I'll just make the `equals` visible for testing.

TadCordle:
It looks like I'm getting some weird behavior with gradle now:

```
tcordle@tcordle33:~/IdeaProjects/hellogradle$ ./gradlew clean jib
Retrieving registry credentials for gcr.io...
Building classes layer...
Finalizing...

> Task :jib 

Containerizing application to gcr.io/tcordle-test/hello:u5ij32ro...


Container entrypoint set to [java, -cp, /app/resources:/app/classes:/app/libs/*, Hello]

Built and pushed image as gcr.io/tcordle-test/hello:u5ij32ro
```

It looks like a bunch of the log messages are appearing out of order.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/LogEventHandler.java
```diff
@@ -17,12 +17,19 @@
...
 import java.util.function.Consumer;
 import org.gradle.api.logging.Logger;
 
 /** Handles {@link LogEvent}s by passing to the Gradle {@link Logger}. */
+// We don't care about the return values of the logging futures.
+@SuppressWarnings("FutureReturnValueIgnored")
 class LogEventHandler implements Consumer<LogEvent> {
 
+  /** This executor keeps all log messages in order. */
+  private static final ExecutorService executorService = Executors.newSingleThreadExecutor();
```
chanseokoh: Hmm... this will do serialize the log messages in sequence, but I think this will make the logs out of sync with the actual event points, since logs are being handled in a new, separate thread.

I'm still wondering why this is happening all of a sudden with Gradle, because in the end, the code seems to do the same thing as before.

TadCordle: This apparently happened a long time ago, and the same fix needed to be applied to `GradleJibLogger`. The same fix was just missing from the event handler. @coollog can probably explain better.

coollog: Yeah, this was because calling the Gradle logging from different threads may cause the log messages to be out-of-order. I think having the log messages in sync with the actual event points is not too important as long as we have them in the same order in which they are emitted (ensured via the single-thread executor).

chanseokoh: Okay, this is not a regression, so we don't have to be concerned in this PR.

I think we can just file an issue about the out-of-sync log messages in Gradle to keep track of it.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -68,7 +67,6 @@ static MavenProjectProperties getForProject(
       return new MavenProjectProperties(
           project,
           makeEventEmitter(log),
-          new MavenJibLogger(log),
```
chanseokoh: Can we remove `MavenJibLogger` and its interface `JibLogger`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -64,8 +63,6 @@ static GradleProjectProperties getForProject(
       return new GradleProjectProperties(
           project,
           makeEventEmitter(logger),
-          // TODO: Remove
-          new GradleJibLogger(logger),
```
chanseokoh: Can we remove `GradleJibLogger` and its interface `JibLogger`?

jib-core/src/main/java/com/google/cloud/tools/jib/event/events/LogEvent.java
```diff
@@ -89,4 +91,24 @@ public Level getLevel() {
   public String getMessage() {
     return message;
   }
+
+  @VisibleForTesting
+  @Override
+  public boolean equals(Object obj) {
```
coollog: tiny nit: s/obj/other/ to avoid using abbreviations. The `other` below can be `otherLogEvent`.

jib-core/src/main/java/com/google/cloud/tools/jib/event/events/LogEvent.java
```diff
@@ -89,4 +91,24 @@ public Level getLevel() {
...
+    if (obj == this) {
+      return true;
+    }
+    if (!(obj instanceof LogEvent)) {
+      return false;
+    }
+
+    LogEvent other = (LogEvent) obj;
+    return level == other.getLevel() && message.equals(other.getMessage());
```
coollog: nit: can use the private fields directly since checking for equality is checking on the contents rather than the return values of getters.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
TadCordle: I'll get to these eventually. Loggers are still passed around in `BuildConfiguration` for the timers, which I'll be working on updating in later PRs. There are still some loose ends in this PR, but I wanted to submit some changes before it got too big.

PR #1037, opened 2018-09-26
Stores temporary cache files under cache directory.

### LINKED ISSUE #1036: DefaultCacheStorageWriter#write fails for moving between different FileStores
Since we write a layer directory to a temporary location first and then atomically move the directory to its intended location, using `Files.move` can fail if the temporary location is on a different `FileStore` than the intended location.

From javadoc for `Files#move`:

> When invoked to move a directory that is not empty then the directory is moved if it does not require moving the entries in the directory.  For example, renaming a directory on the same {@link FileStore} will usually not require moving the entries in the directory. When moving a directory requires that its entries be moved then this method fails (by throwing an {@code IOException}).

Example error:

```
java.nio.file.DirectoryNotEmptyException: /tmp/2774970389576489081
        at sun.nio.fs.UnixCopyFile.move(UnixCopyFile.java:498)
        at sun.nio.fs.UnixFileSystemProvider.move(UnixFileSystemProvider.java:262)
        at java.nio.file.Files.move(Files.java:1395)
        at com.google.cloud.tools.jib.ncache.DefaultCacheStorageWriter.moveIfDoesNotExist(DefaultCacheStorageWriter.java:75)
```

I believe the solution would be to use a temporary location within the cache directory instead. This will be under a `tmp/` directory.

coollog:
Fixes #1036 
Uses `[CACHE DIRECTORY]/tmp` as the temporary directory.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/TemporaryDirectory.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/TemporaryDirectory.java
@@ -32,7 +32,7 @@ public class TemporaryDirectory implements Closeable {
   private final Path temporaryDirectory;
 
   /**
-   * Creates a new temporary directory under {@code parentDirectory}.
+   * Creates a new temporary directory under an existing {@code parentDirectory}.
    *
    * @param parentDirectory the directory to create the temporary directory within
    * @throws IOException if an I/O exception occurs
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/TemporaryDirectory.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/TemporaryDirectory.java
@@ -32,12 +32,13 @@ public class TemporaryDirectory implements Closeable {
   private final Path temporaryDirectory;
 
   /**
-   * Creates a new temporary directory.
+   * Creates a new temporary directory under {@code parentDirectory}.
    *
+   * @param parentDirectory the directory to create the temporary directory within
    * @throws IOException if an I/O exception occurs
    */
-  public TemporaryDirectory() throws IOException {
-    temporaryDirectory = Files.createTempDirectory(null);
+  public TemporaryDirectory(Path parentDirectory) throws IOException {
+    temporaryDirectory = Files.createTempDirectory(parentDirectory, null);
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
@@ -25,10 +25,13 @@ import com.google.cloud.tools.jib.image.DescriptorDigest;
 import java.io.BufferedOutputStream;
 import java.io.IOException;
 import java.nio.file.AtomicMoveNotSupportedException;
+import java.nio.file.DirectoryNotEmptyException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import java.util.zip.GZIPOutputStream;
 
 /** Writes to the default cache storage engine. */
@@ -59,6 +62,10 @@ class DefaultCacheStorageWriter {
    * @throws IOException if an I/O exception occurs
    */
   private static void moveIfDoesNotExist(Path source, Path destination) throws IOException {
+    if (Files.exists(destination)) {
+      return;
+    }
+
     try {
       Files.move(source, destination, StandardCopyOption.ATOMIC_MOVE);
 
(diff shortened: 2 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -108,8 +108,11 @@ CacheEntry write(CacheWrite cacheWrite) throws IOException {
     // Creates the layers directory if it doesn't exist.
     Files.createDirectories(defaultCacheStorageFiles.getLayersDirectory());
 
-    // Creates the temporary directory.
-    try (TemporaryDirectory temporaryDirectory = new TemporaryDirectory()) {
+    // Creates the temporary directory. The temporary directory must be in the same FileStore as the
```
TadCordle: `{@link FileStore}` and `{@code Files.move}`?

coollog: The brace notations only work in javadoc, not in comments.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/TemporaryDirectory.java
```diff
@@ -32,12 +32,13 @@
...
-   * Creates a new temporary directory.
+   * Creates a new temporary directory under {@code parentDirectory}.
    *
+   * @param parentDirectory the directory to create the temporary directory within
    * @throws IOException if an I/O exception occurs
    */
-  public TemporaryDirectory() throws IOException {
-    temporaryDirectory = Files.createTempDirectory(null);
+  public TemporaryDirectory(Path parentDirectory) throws IOException {
+    temporaryDirectory = Files.createTempDirectory(parentDirectory, null);
```
chanseokoh: `Files.createTempDirectory` will error if `parentDirectory` does not exist. Should we be worried?

coollog:
Realized that on a file system that doesn't support simply renaming a directory with contents, moves still throw the `DirectoryNotEmptyException`. Will need to take care of this case as well.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/TemporaryDirectory.java
coollog: Will delegate the creation of the parent directory to the caller. Touched up the javadoc for that.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -75,6 +82,15 @@ private static void moveIfDoesNotExist(Path source, Path destination) throws IOE
 
       } catch (FileAlreadyExistsException alsoIgnored) {
         // Same reasoning
+
+      } catch (DirectoryNotEmptyException ex) {
+        // The file system cannot rename directories, so we must resort to copying the directory.
```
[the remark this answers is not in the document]
coollog: This happens only on a file system that doesn't support simply renaming a directory with contents (for example, I tried it on a network-mounted virtual file system). The only possible concurrency issue I can see with this is if the layer file is copied first, and the metadata is not copied yet, then another execution will try to rebuild the layer.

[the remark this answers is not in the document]
coollog: Filed #1044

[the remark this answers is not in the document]
chanseokoh: Just making a note: `DirectoryNotEmptyException` is thrown in our test case because Java can't delete the source directory: https://github.com/frohoff/jdk8u-jdk/blob/master/src/solaris/classes/sun/nio/fs/UnixCopyFile.java#L497

The exception name is a bit misleading at the higher level, since what it is saying at the low level is that, it failed to delete the source directory hence the source directory remains non-empty. (Copy to the target is already done, but because Java couldn't delete the source directory, it does the cleanup to remove the copied target directory.)

BTW, in our case, the source and the target file system providers were identical: the Linux file system.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -75,6 +82,15 @@ private static void moveIfDoesNotExist(Path source, Path destination) throws IOE
...
+        Files.createDirectory(destination);
+        try (Stream<Path> sourceFiles = Files.list(source)) {
+          for (Path sourceFile : sourceFiles.collect(Collectors.toList())) {
+            Files.copy(sourceFile, destination.resolve(sourceFile.getFileName()));
+          }
+        }
```
chanseokoh: Try removing the tmp source?

coollog: The tmp source would be removed via the `TemporaryDirectory` mechanism - do you think it'd be good to just do that here as well?

chanseokoh: This method is not aware of `TemporaryDirectory`, so it won't hurt to try to keep the promise of "move". But then, we will probably ignore failures from deleting the source fails, so I'm fine with leaving it there, considering also that this manual copy is somewhat a corner case.

PR #1043, opened 2018-09-26
Replace Timer with timer events

TadCordle:
Part of #714. Intending to have one more followup that removes `JibLogger`.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/TimerEventEmitter.java
```diff
@@ -63,11 +63,24 @@ public TimerEventEmitter subTimer(String description) {
...
   }
 
+  /**
+   * Captures the time since last lap or creation and emits an {@link State#LAP} {@link TimerEvent}.
+   *
+   * @param newDescription the description to use instead of the {@link TimerEventEmitter}'s
+   *     description
+   */
+  public void lap(String newDescription) {
+    eventEmitter.emit(
```
coollog: Should we just change `emitTimerEvent` to take a description as well?

PR #1045, opened 2018-09-26
Remove deprecated parameters

### LINKED ISSUE #461: Remove deprecated parameters
The parameters

- `jvmFlags`
- `mainClass`
- `args`
- `format`

all have been moved under the nested `container` configuration, and the non-nested versions should be removed.

TadCordle:
Fixes #461


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -220,8 +177,7 @@ public ContainerParameters getContainer() {
     return container;
   }
 
-  // TODO: Make @Internal (deprecated)
-  @Input
+  @Internal
```
coollog: Hmm, should we just get rid of these instead?

TadCordle: I suppose.

PR #1046, opened 2018-09-26
 Changes CacheWrite into UncompressedCacheWrite and adds ability to write compressed layer blob.

### LINKED ISSUE #1042: CacheWrite needs to be split into UncompressedCacheWrite and CompressedCacheWrite.
The current implementation assumes `layerBlob` is uncompressed. However, for pulling base image layers, `layerBlob` would already be compressed. This would require a separate implementation of `DefaultCacheStorageWriter#write` as well.

coollog:
Fixes #1042


jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
```diff
@@ -57,31 +57,32 @@ private Cache(CacheStorage cacheStorage) {
...
+   * a metadata {@link Blob}. Use {@link #write(Blob)} to save a compressed layer {@link Blob}.
    *
-   * @param layerBlob the layer {@link Blob}
+   * @param uncompressedLayerBlob the layer {@link Blob}
    * @param layerEntries the layer entries that make up the layer
    * @return the {@link CacheEntry} for the written layer and metadata
    * @throws IOException if an I/O exception occurs
    */
-  public CacheEntry write(Blob layerBlob, ImmutableList<LayerEntry> layerEntries)
+  public CacheEntry write(Blob uncompressedLayerBlob, ImmutableList<LayerEntry> layerEntries)
```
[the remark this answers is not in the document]
coollog: I was thinking since the `compressed` or `uncompressed` qualifier only applies to the `Blob`, having it on the method name sounds a bit off since the `layerEntries` is not compressed or uncompressed, but I do agree distinguishing in the method name would be good. Would there be a better naming than `writeCompressed` vs `writeUncompressed`?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -95,32 +100,91 @@ private static void moveIfDoesNotExist(Path source, Path destination) throws IOE
...
+          .setLayerBlob(Blobs.from(layerFile))
+          .build();
+    }
+  }
+
+  /**
+   * Writes the {@link UncompressedCacheWrite}.
    *
-   * <p>The {@link CacheWrite} is written out to the cache directory in the form:
+   * <p>The {@link UncompressedCacheWrite} is written out to the cache directory in the form:
```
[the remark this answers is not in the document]
coollog: I think I included this line mostly to clarify that it is written out the *cache directory*.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/Cache.java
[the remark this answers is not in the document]
TadCordle: Maybe something that points out what is being written, like `writeCompressedBlob`/`Layer`/`Entry`. `writeCompressed` and `writeUncompressed` might be good enough, though.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheStorage.java
```diff
@@ -36,13 +37,22 @@
...
+  CacheEntry write(UncompressedCacheWrite uncompressedCacheWrite) throws IOException;
+
+  /**
+   * Saves a compressed layer {@link Blob}.
+   *
+   * @param compressedLayerBlob the compressed layer {@link Blob}
+   * @return the {@link CacheEntry} for the written layer
+   * @throws IOException if an I/O exception occurs
+   */
+  CacheEntry write(Blob compressedLayerBlob) throws IOException;
```
chanseokoh: From the class Java "The cache storage engine stores layer data as UncompressedCacheWrites", I assume this method will/should uncompress the given blob, right?

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -154,30 +220,60 @@ CacheEntry write(CacheWrite cacheWrite) throws IOException {
...
    * @param layerDirectory the directory for the layer
    * @return a {@link WrittenLayer} with the written layer information
    * @throws IOException if an I/O exception occurs
    */
-  private WrittenLayer writeLayerBlobToDirectory(Blob layerBlob, Path layerDirectory)
-      throws IOException {
+  private WrittenLayer writeCompressedLayerBlobToDirectory(
+      Blob compressedLayerBlob, Path layerDirectory) throws IOException {
+    // Writes the layer file to the temporary directory.
+    Path temporaryLayerFile = Files.createTempFile(layerDirectory, null, null);
```
chanseokoh: Just saying: we don't really need to create a file, since we'll just overwrite it. We just need a random name, but admittedly, `Files.createTempFile()` is an easy way to have a random name.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
```diff
@@ -154,30 +220,60 @@ CacheEntry write(CacheWrite cacheWrite) throws IOException {
...
+    try (OutputStream fileOutputStream =
+        new BufferedOutputStream(Files.newOutputStream(temporaryLayerFile))) {
+      layerBlobDescriptor = compressedLayerBlob.writeTo(fileOutputStream);
+    }
+
+    // Gets the diff ID.
+    DescriptorDigest layerDiffId = getDiffIdByDecompressingFile(temporaryLayerFile);
+
+    Path layerFile = layerDirectory.resolve(defaultCacheStorageFiles.getLayerFilename(layerDiffId));
+    moveIfDoesNotExist(temporaryLayerFile, layerFile);
```
chanseokoh: It'd be nice if we remove the double moves (first moving a temp file in a temp directory here, and then moving the temp directory by the caller). However, I am fine if the double moves keep the code more readable.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/CacheStorage.java
coollog: Ah right, I need to fix that javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
coollog: Yea, maybe we can just use a static file name like "layer".

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
coollog: The moving the temp file is to name it correctly (with the diff ID, which can only be obtained after the file is written). The moving the temp directory is to atomically move the directory into place rather having the directory in the final location with empty contents before the layer contents are added.

jib-core/src/main/java/com/google/cloud/tools/jib/ncache/DefaultCacheStorageWriter.java
coollog: I'm going to actually create a `TemporaryFile` class to encapsulate this logic better.

coollog: Actually, nevermind, I'll leave that to a separate PR.

PR #1049, opened 2018-09-27
New JavaLayerConfiguration API for more flexibility

### LINKED ISSUE #1007: Change JavaLayerConfiguration.set___Files to use the new API to add source files
Current signature: `setExtraFiles(List<Path> extraFiles)`

Currently, the source files (`extraFiles`) should be top-level files and directories. If a source path is a directory, `JavaLayerConfiguration` calls `LayerConfiguration.Builder.addEntryRecursive()` to add all the files and sub-directories in the directory recursively. That is, it is not possible to separate files in a directory into different layers (much like the problem of #957). This is also problematic in the WAR case. For example,

```
WAR-root/
    index.html
    WEB-INF/
        classes/
            HelloServlet.class
            cool.properties
        web.xml
```

The resources layer should have
```
index.html                       -->  to be placed in <appRoot>
WEB-INF/web.xml                  -->  to be placed in <appRoot>
WEB-INF/classes/cool.properties  -->  to be placed in <appRoot>
```
which should be unpacked at the root of the application in an image, whereas the classes layer should have
```
WEB-INF/classes/HelloServlet.class  -->  to be placed in <appRoot>
```
or
```
HelloSevlet.class  -->  to be placed in <appRoot>/WEB-INF/classes
```

Fortunately, #956 is fixed, so we can now change `setExtraFiles()` to accept an explicit list of individual files to add. That is, instead of adding fiels with `addEntryRecursive()`, just add each file with `addEntry()`. However, it may be inconvenient for callers to always explicitly list all the files, so it'll be good to have `setExtraFilesDirectory()` and `setExtraFilesDirectories()` too.

chanseokoh:
Fixes #1007.

See if this sounds reasonable.

I will see if I can further simplify `GradleLayerConfigurations` and `MavenLayerConfigurations` with the new API.


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -74,16 +72,9 @@
     /** The directory in the context to put the source files for the layer */
     private final String directoryInContext;
 
-    /** The extraction path in the image. */
-    private final AbsoluteUnixPath extractionPath;
-
-    private CopyDirective(
-        ImmutableList<LayerEntry> layerEntries,
-        String directoryInContext,
-        AbsoluteUnixPath extractionPath) {
```
coollog: I think we might want to keep `extractionPath` here, or rename it, since `CopyDirectory` is meant to be a representation of the Docker `COPY` directive. In this current solution, though, `extractionPath` would always be `/`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -65,78 +65,238 @@ String getName() {
   /** Builds with each layer's files. */
   public static class Builder {
 
-    private final Map<LayerType, List<Path>> layerFilesMap = new EnumMap<>(LayerType.class);
-    private final Map<LayerType, AbsoluteUnixPath> extractionPathMap =
-        new EnumMap<>(LayerType.class);
+    private static class LayerFile {
+      private Path sourceFile;
```
coollog: tiny nit: newline after class declaration

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -65,78 +65,238 @@ String getName() {
...
+      private AbsoluteUnixPath pathInContainer;
+      private boolean recursiveAdd;
```
coollog: final for all fields

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -65,78 +65,238 @@ String getName() {
...
+
+      private LayerFile(Path sourceFile, AbsoluteUnixPath pathInContainer, boolean recursiveAdd) {
+        this.sourceFile = sourceFile;
+        this.pathInContainer = pathInContainer;
+        this.recursiveAdd = recursiveAdd;
+      }
+    }
+
+    private final Map<LayerType, List<LayerFile>> layerFilesMap = new EnumMap<>(LayerType.class);
```
coollog: Could this just be a `Map<LayerType, LayerConfiguration>`? The `add...File` methods would just call `LayerConfiguration#add/addRecursive`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -65,78 +65,238 @@ String getName() {
...
+     * Adds a file to the dependency layer. If the source file is a directory, the directory and its
+     * contents will be added recursively. See {@link LayerConfiguration.Builder#addEntryRecursive}
+     * for concrete examples about how the file will be placed in the image.
+     *
+     * @param sourceFile the source file to add to the layer
+     * @param pathInContainer the path in the container file system corresponding to the {@code
+     *     sourceFile}
+     * @return this
+     */
+    public Builder addDependencyFileRecursive(Path sourceFile, AbsoluteUnixPath pathInContainer) {
```
coollog: Is it okay if all files are added recursively? (aka are the non-rescursive `add...File` methods necessary?)

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: It's certainly doable. But I also think this provides a small degree of convenience to the caller that they don't have to use multiple builders of `LayerConfiguration` for the builder of `JavaLayerConfigurations`. I think if you don't feel strongly against of this API, we could try this for a bit and decide later if we should go for `LayerConfiguration` or think about a totally different API?

coollog: Oh I wasn't referring to changing the API - just the internal representation here. The API for `JavaLayerConfigurations` could still be the same.

chanseokoh: Oh, I see. That makes sense. The only reason I am constructing `LayerConfiguration` in `build()` was to maintain the same API where `add...FileRecursive()` methods do not throw an exception. Currently, `LayerConfiguration.Builder.addEntryRecursive()` throws `IOException`, so if we were to change the map to contain `LayerConfiguration`, we will need to expose the `IOException` at the time of calling `add...FileRecursive()`, which may not be bad, since `LayerConfiguration` is already doing so. WDYT?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: I tried to be in line with `LayerConfiguration` having both recursive and non-recursive methods, but now I think it is possible to cover all cases only with recursive adds. Then perhaps I can rename `add...FileRecursive()` to just `add...File`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
coollog: Yea I think having the `IOException` should be fine.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
coollog: Sounds good!

chanseokoh: Now I remember. The non-recursive case is useful when the caller is walking a directory, enumerating all files and directories and adding each path.

PR #1052, opened 2018-09-27
Remove recursion from files task

### LINKED ISSUE #1033: Consider removing recursion from skaffold files task
https://github.com/GoogleContainerTools/jib/pull/1021#discussion_r220351260

TadCordle:
Fixes #1033

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
@@ -21,6 +21,7 @@ import java.io.File;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.HashSet;
 import java.util.Set;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -96,25 +97,35 @@ private static void printProjectFiles(Project project) {
...
-    for (Configuration configuration :
-        project.getConfigurations().getByName("runtime").getHierarchy()) {
-      for (Dependency dependency : configuration.getDependencies()) {
-        if (dependency instanceof ProjectDependency) {
-          projectDependenciesResult.add((ProjectDependency) dependency);
-          findProjectDependencies(
-              ((ProjectDependency) dependency).getDependencyProject(), projectDependenciesResult);
+  private static Set<ProjectDependency> findProjectDependencies(Project project) {
+    Set<ProjectDependency> projectDependencies = new HashSet<>();
+    ArrayDeque<Project> projects = new ArrayDeque<>();
```
coollog: nit: Not entirely sure, but for a queue, we might want the type to be `Queue` (doesn't need to be double-ended here) and use `add` and `remove` (or `offer` and `poll`). `push` and `pop` are more for using `ArrayDeque` as a stack.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -96,25 +97,35 @@ private static void printProjectFiles(Project project) {
   }
 
   /**
-   * Recursive function for printing out a project's artifacts. Calls itself when it encounters a
-   * project dependency.
+   * Function for collecting a project's project dependencies.
```
coollog: `Collects a project's project dependencies, including all transitive project dependencies`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
TadCordle: That was intended, is there any reason we should use a queue instead of a stack here?

coollog: Oh okay, that makes sense. Then maybe use the `Deque` type?

chanseokoh:
Personally, I often prefer recursion, since it's more straightforward.

PR #1056, opened 2018-09-27
Renames ncache to cache.

coollog:
Part of #637


loosebazooka:
Can you add a issue for stress testing on the 0.10.0 milestone by the team. Both size and concurrency.

coollog:
Yep - filed #1059

PR #1061, opened 2018-09-28
Fix to/from closure type in JibExtension

### LINKED ISSUE #1060: Cannot acccess 'image'
After upgrading to 0.9.11 I get this error message for build.gradle.kts:
`Cannot acccess 'image': It is public in <anonymous>`

grafik

loosebazooka:
fixes #1060


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -107,11 +107,11 @@ public JibExtension(Project project) {
     extraDirectory.set(resolveDefaultExtraDirectory(project.getProjectDir().toPath()));
   }
 
-  public void from(Action<? super ImageParameters> action) {
+  public void from(Action<? super BaseImageParameters> action) {
     action.execute(from);
   }
 
-  public void to(Action<? super ImageParameters> action) {
+  public void to(Action<? super TargetImageParameters> action) {
```
loosebazooka: It's possible we don't need `ImageParameters` anymore. It passes through some annotations, but we can probably apply those on the subclasses directly.

TadCordle: Yeah, I think we should get rid of it.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -34,34 +36,31 @@ public BaseImageParameters(ObjectFactory objectFactory, String imageDescriptor)
     auth = objectFactory.newInstance(AuthParameters.class, imageDescriptor + ".auth");
   }
 
+  @Input
   @Nullable
-  @Override
   public String getImage() {
```
TadCordle: I think we still need the `@Optional` on each of these methods, right?

chanseokoh:
I don't know about Gradle, but logically, I feel like referencing an interface that has necessary getters and setters should have worked. Doesn't seem difference really, so not understanding why it didn't work.

coollog:
@chanseokoh Gradle works fine, but the issue was IntelliJ wasn't able to pick up the interface annotations.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
chanseokoh: Good catch. Seems like `getCredHelper` and `auth` should be optional too?

TadCordle: Yeah, in both classes; didn't want to spam the same comment on each method haha

chanseokoh:
Oh, I see. Sounds like the fault of IntelliJ then.

loosebazooka:
The Target Image Params has extra params though, and it shouldn't be able to apply the action on it even if it wanted to?

TadCordle:
@loosebazooka I think this is good after we fix the `@Optional` stuff. Can we get this in soon? I want to work on #1083 and I expect that to touch these files a bunch.

loosebazooka:
Yeah sorry, sometimes I forget I write code from time to time :disappointed:

PR #1062, opened 2018-09-28
Construct correct classes and resources layers for Maven builds

### LINKED ISSUE #1005: Logic to filter resource files from classes/ is not perfect
`MavenLayerConfigurations.getForJarProject()`:

I know we're doing our best, and maybe the logic below is the best we can do with the current API, but making a note here that the logic isn't perfect and resource files may slip into the classes layer, e.g., if a resource file is `classes/com/example/my-cool.properties`. However, now that #956 is fixed, this issue can be resolved somehow by leveraging the new API?

```java
    // Gets the classes files in the 'classes' output directory. It finds the files that are classes
    // files by matching them against the .java source files. All other files are deemed resources.
    try (Stream<Path> classFileStream = Files.list(classesOutputDirectory)) {
      classFileStream.forEach(
          classFile -> {
            /*
             * Adds classFile to classesFiles if it is a .class file or is a directory that also
             * exists in the classes source directory; otherwise, adds file to resourcesFiles.
             */
            if (Files.isDirectory(classFile)
                && Files.exists(
                    classesSourceDirectory.resolve(classesOutputDirectory.relativize(classFile)))) {
              classesFiles.add(classFile);
              return;
            }

            if (FileSystems.getDefault().getPathMatcher("glob:**.class").matches(classFile)) {
              classesFiles.add(classFile);
              return;
            }

            resourcesFiles.add(classFile);
          });
    }
```

chanseokoh:
Fixes #1005.

Explicitly walks through `target/classes` and put only `.class` files into the classes layer. All other files go into the resources layer.

Note that's another behavior change. Because [the new `JavaLayerConfigurations` API always adds files recursively](https://github.com/GoogleContainerTools/jib/pull/1049#discussion_r220984045), I had to skip non-empty directories during the directory walk. (Empty directories are added, however.) Previously, such directories will also be added as `LayerEntry`. I think it's fine, because we had the extra step to add tar archive entries for all possible directories to set correct permissions on directories.

However, it might be better to just revert https://github.com/GoogleContainerTools/jib/pull/1049#discussion_r220984045 and use non-recursive addition for this kind of use cases. Thoughts?


chanseokoh:
Oh, I forgot to push up some tests.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
```diff
@@ -46,85 +43,74 @@
...
+
+  private static void addFilesToLayer(
+      Path sourceRoot,
+      Predicate<Path> pathFilter,
+      AbsoluteUnixPath basePathInContainer,
+      FileToLayerAdder addFileToLayer)
+      throws IOException {
+    Predicate<Path> isEmptyDirectory =
+        path -> Files.isDirectory(path) && path.toFile().list().length == 0;
+    // Always add empty directories. However, ignore non-empty directories because otherwise
```
coollog: I think we can just skip adding any directories since directories are all added in `ReproducibleLayerBuilder`.

chanseokoh: But empty directories won't be added, will they? Such a directory doesn't have any file, so if we don't add it, `ReproducibleLayerBuilder` won't even know it existed. Isn't it?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
```diff
@@ -46,85 +43,74 @@
...
+  }
+
+  private static void addFilesToLayer(
+      Path sourceRoot,
+      Predicate<Path> pathFilter,
+      AbsoluteUnixPath basePathInContainer,
+      FileToLayerAdder addFileToLayer)
+      throws IOException {
+    Predicate<Path> isEmptyDirectory =
+        path -> Files.isDirectory(path) && path.toFile().list().length == 0;
```
coollog: I think we should favor using `Files.list` since it at most would need to list one file from a directory rather than listing all the files every time.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
coollog: Hmm, that's true - we probably want to honor the exact directories in the source files. However, I believe when you compile with Maven, it actually ignores empty directories.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
chanseokoh: I used `File.list()` because `Files.list()` needs closing the stream. I can still go with `Files.list()`, but it doesn't seem to justify the reduced readability due to increased cognitive load. I think `path.toFile().list()` just works out fine here, since all we want to know is if it's empty.

chanseokoh: But it's also true that `toFile().list()` will construct the entire list, so `Files.list()` has an advantage too. Actually, let me use `Files.list()`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
chanseokoh: This is a general method, and we should honor the exact directory structure for other layers like extra files and resources. The program in an image may make assumptions about such empty directories.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
```diff
@@ -46,85 +43,74 @@
...
+    // Always add empty directories. However, ignore non-empty directories because otherwise
+    // JavaLayerConfigurations will add files recursively.
+    Predicate<Path> shouldAdd =
+        path -> isEmptyDirectory.test(path) || (!Files.isDirectory(path) && pathFilter.test(path));
+
+    DirectoryWalker filteredWalker = new DirectoryWalker(sourceRoot).filter(shouldAdd);
```
coollog: I'm thinking since we are using `Files.list` already, perhaps, we can just make this a recursive method that does `Files.list` over all the subfiles/directories and adds all the subdirectories and subfiles that pass the filter, and avoid performing a `list` on each directory while simultaneously walking the directories (resulting in essentially double listing).

chanseokoh: Good idea. Let me try and see how it works.

chanseokoh: Did the recursion version, but not sure I like it. The double listing seems more concise. In this case, I'd prefer simple code that's easy to digest and understand. I don't really think the optimization is worth.

```java
  private static void addFilesToLayer(
      Path sourceRoot,
      Predicate<Path> pathFilter,
      AbsoluteUnixPath basePathInContainer,
      FileToLayerAdder addFileToLayer)
      throws IOException {
    addFilesToLayer(sourceRoot, sourceRoot, pathFilter, basePathInContainer, addFileToLayer);
  }

  @VisibleForTesting
  static void addFilesToLayer(
      Path sourceFile,
      Path sourceRoot,
      Predicate<Path> pathFilter,
      AbsoluteUnixPath basePathInContainer,
      FileToLayerAdder addFileToLayer)
      throws IOException {
    AbsoluteUnixPath finalPath = basePathInContainer.resolve(sourceRoot.relativize(sourceFile));

    if (!Files.isDirectory(sourceFile)) {
      if (pathFilter.test(sourceFile)) {
        addFileToLayer.add(sourceFile, finalPath);
      }
      return;
    }

    List<Path> children = listDirectory(sourceFile);
    if (children.isEmpty()) { // always add empty directory
      addFileToLayer.add(sourceFile, finalPath);
    }
    for (Path child : children) {
      addFilesToLayer(child, sourceRoot, pathFilter, basePathInContainer, addFileToLayer);
    }
  }
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
```diff
@@ -89,27 +91,39 @@ static JavaLayerConfigurations getForProject(
...
+    if (Files.isDirectory(path)) {
+      try (Stream<Path> stream = Files.list(path)) {
+        return !stream.findAny().isPresent();
+      }
+    }
+    return false;
+  }
+
+  @VisibleForTesting
+  static void addFilesToLayer(
```
coollog: It'd be good to have a javadoc here to explain what each of the parameters refer to.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurationsTest.java
```diff
@@ -66,10 +68,13 @@ private static void assertExtractionPathsUnordered(
   }
 
   @Rule public final TestRepository testRepository = new TestRepository();
+  @Rule public final TemporaryFolder tempFolder = new TemporaryFolder();
```
coollog: nit: avoid abbreviations s/tempFolder/temporaryFolder`

chanseokoh:
> have for filtering is to walk all the files we want to add to the container and filter in an if-elseif manner such that files that pass a filter will be added to the layer for that filter or otherwise continue to the next filter until reaching a default layer to place the file into.

Interesting idea that we can try.

PR #1063, opened 2018-09-28
Support stringifying ImageReference with digest

### LINKED ISSUE #981: Support stringifying ImageReference with digest.
Currently, digests parsed by `ImageReference` are stringified as tags.

For example:
`gcr.io/distroless/java@sha256:b430543bea1d8326e767058bdab3a2482ea45f59d7af5c5c61334cd29ede88a1`, once parsed into `ImageReference`, should stringify as the same, not as `gcr.io/distroless/java:sha256:b430543bea1d8326e767058bdab3a2482ea45f59d7af5c5c61334cd29ede88a1`

TadCordle:
Fixes #981

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
@@ -290,7 +290,8 @@ public class ImageReference {
 
     // Use tag if not the default tag.
     if (!DEFAULT_TAG.equals(tag)) {
-      referenceString.append(':').append(tag);
+      // Append with "@tag" instead of ":tag" if tag is a digest
+      referenceString.append(tag.matches(DescriptorDigest.DIGEST_REGEX) ? '@' : ':').append(tag);
     }
 
     return referenceString.toString();
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/image/ImageReferenceTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/image/ImageReferenceTest.java
@@ -121,7 +121,7 @@ public class ImageReferenceTest {
   }
 
   @Test
-  public void testToString() {
+  public void testToString() throws InvalidImageReferenceException {
     Assert.assertEquals("someimage", ImageReference.of(null, "someimage", null).toString());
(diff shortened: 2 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -290,7 +290,8 @@ public String toString() {
 
     // Use tag if not the default tag.
     if (!DEFAULT_TAG.equals(tag)) {
-      referenceString.append(':').append(tag);
+      // Append with "@tag" instead of ":tag" if tag is a digest
+      referenceString.append(tag.matches(DescriptorDigest.DIGEST_REGEX) ? '@' : ':').append(tag);
```
coollog: Should prob add this to `toStringWithTag` as well (perhaps with a `getTagString` method).

TadCordle: Shouldn't need to, `getTagString` only does things differently if the tag is default (which it won't be if the tag is a digest).

coollog: I think adding a `TagType getTagType` (alternatively, `boolean isTagDigest`) would be useful for users of `ImageReference`

coollog: Ah right :+1:

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageReference.java
```diff
@@ -256,6 +256,16 @@ public boolean usesDefaultTag() {
     return isDefaultTag(tag);
   }
 
+  /**
+   * Returns {@code true} if the {@link ImageReference} uses a SHA-256 digest as its tag; {@code
+   * false} if not.
```
coollog: Maybe also add `When the tag is a digest, in string form, it is prepended with the at {@code @} symbol instead of a colon (@code :}.`

TadCordle:
> Should probably also change isValidTag to accept a digest as well.

Doesn't the tag regex already match digests?

PR #1066, opened 2018-09-28
Add jib version to layer history

### LINKED ISSUE #1041: Add version of Jib for generated history entry.

TadCordle:
Fixes #1041

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/ProjectInfo.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/ProjectInfo.java
@@ -25,5 +25,8 @@ public class ProjectInfo {
   /** Link to file an issue against the GitHub repository. */
   public static final String GITHUB_NEW_ISSUE_URL = GITHUB_URL + "/issues/new";
 
+  /** The project version. */
+  public static final String VERSION = ProjectInfo.class.getPackage().getImplementationVersion();
+
   private ProjectInfo() {}
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.jib.builder.steps;
 
+import com.google.cloud.tools.jib.ProjectInfo;
 import com.google.cloud.tools.jib.async.AsyncStep;
 import com.google.cloud.tools.jib.async.NonBlockingSteps;
 import com.google.cloud.tools.jib.blob.Blob;
@@ -160,6 +161,8 @@ class BuildImageStep
       }
 
       // Add built layers/configuration
+      // Note: ProjectInfo.VERSION may be null in tests
+      String version = ProjectInfo.VERSION == null ? "" : (":" + ProjectInfo.VERSION);
       for (BuildAndCacheApplicationLayerStep buildAndCacheApplicationLayerStep :
           buildAndCacheApplicationLayerSteps) {
         imageBuilder.addLayer(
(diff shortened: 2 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -160,6 +161,8 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
       }
 
       // Add built layers/configuration
+      // Note: ProjectInfo.VERSION may be null in tests
+      String version = ProjectInfo.VERSION == null ? "" : (":" + ProjectInfo.VERSION);
```
chanseokoh: When null, how about ":unknown-version" or such? That way, we can distinguish between older versions and failing to get version info in newer versions.

TadCordle: Good point

jib-core/src/main/java/com/google/cloud/tools/jib/ProjectInfo.java
```diff
@@ -25,5 +25,8 @@
   /** Link to file an issue against the GitHub repository. */
   public static final String GITHUB_NEW_ISSUE_URL = GITHUB_URL + "/issues/new";
 
+  /** The project version. */
+  public static final String VERSION = ProjectInfo.class.getPackage().getImplementationVersion();
```
coollog: I think we should error if this is null.

TadCordle: It's null during testing though?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -160,6 +161,8 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
       }
 
       // Add built layers/configuration
+      // Note: ProjectInfo.VERSION may be null in tests
+      String version = ":" + ProjectInfo.VERSION;
```
coollog: This variable looks like it's only used on line 174

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -115,7 +116,7 @@ private String makeUserAgent() {
         return "";
       }
 
-      String version = RegistryClient.class.getPackage().getImplementationVersion();
+      String version = ProjectInfo.VERSION;
       StringBuilder userAgentBuilder = new StringBuilder();
       userAgentBuilder.append("jib");
       if (version != null) {
```
coollog: Can remove this null check too

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -115,12 +116,9 @@ private String makeUserAgent() {
         return "";
       }
 
-      String version = RegistryClient.class.getPackage().getImplementationVersion();
       StringBuilder userAgentBuilder = new StringBuilder();
```
chanseokoh: Unrelated, but this looks like this User-Agent string doesn't really conform to the spec or customs. Assume the following doc is up to date, https://en.wikipedia.org/wiki/User_agent#Format_for_human-operated_web_browsers

But I admit it can be a pain to update the string, as backend systems may be relying on it.

TadCordle: Should we create an issue for this?

chanseokoh: We use the string when interactive with registries, right? If so, let's just create it to keep track of it, even though it may take forever to get to this.

TadCordle: #1067

chanseokoh: Thanks!

PR #1069, opened 2018-09-30
Make ImageParameters interface public

anuraaga:
`ImageParameters` is exposed through methods `from` and `to`, but if the interface isn't `public`, then consumers can't actually configure the parameters. This doesn't effect Groovy builds where visibility is ignored, but Java and Kotlin users can't configure image parameters right now.


anuraaga:
Realized #1070 is probably the better fix.

PR #1070, opened 2018-09-30
Return concrete types in from and to

anuraaga:
`from` and `to` currently return `ImageParameters` to consumers, but that is a private type and cannot be used in static-typed scripts (Java, Kotlin). Since `getFrom` and `getTo` return the concrete types, there doesn't seem to be a good reason not to do that in the action versions too.


loosebazooka:
This is covered in #1061.

Issue #924, opened 2018-09-03
Add Support for use CMD to start container


swift1911:
I use supervise like s6-overlay to handle PID 1 for container, jib plugin use entrypoint to start container and overwrite my baseimage entrypoint, so i want to use CMD to start jib generated container 

like this:
```
<container>
            <jvmFlags>
                  <jvmFlag>-javaagent:/data0/java/deploy/pinpoint.agent/pinpoint-bootstrap.jar</jvmFlag>
                  <jvmFlag>-Dpinpoint.agentId=docker</jvmFlag>
                  <jvmFlag>-Dpinpoint.applicationName=docker-test</jvmFlag>
            </jvmFlags>
      <mainClass>com.fangdd.bwfs.server.container.Application</mainClass>
      <exec>CMD</exec>
</container>
```

chanseokoh:
Hi @swift1911,

Let me check if I understood you correctly: you don't want Jib to overwrite the entrypoint and CMD, and you just Jib to just inherit and use the entrypoint of CMD that whatever base image has, right?

swift1911:
@chanseokoh Yes, this is the meaning

chanseokoh:
Now the question is, where/how to indicate Jib to not overwrite the entrypoint and CMD, e.g., a new configuration parameter or a system property?

swift1911:
@chanseokoh i think that a a new configuration parameter is better

swift1911:
@coollog this is my Dockerile, i want to use s6-overlay to init and handle pid 1
```
FROM alpine
RUN apk add busybox-extras bash vim lsof strace pstree git curl tzdata
ADD https://github.com/just-containers/s6-overlay/releases/download/v1.21.4.0/s6-overlay-amd64.tar.gz /tmp/
RUN gunzip -c /tmp/s6-overlay-amd64.tar.gz | tar -xf - -C /
ENV TZ Asia/Shanghai
ENTRYPOINT ["/init"]
```

loosebazooka:
@swift1911 so you're expected execution is something like?

```
/init java -cp <some-classpath> <some-main-class> ...
```

swift1911:
@loosebazooka yes

coollog:
Hi @swift1911, have you tried using the `<entrypoint>` configuration?

```xml
<container>
  <entrypoint>/init</entrypoint>
</container>
```

swift1911:
[...]

PR #1068, opened 2018-09-29
Maven War support

CyrilleH:
Towards #431, here's my proposal to support "WAR" project with Jib and Maven.

~~This PR is based from #923.
Once #923 is merged, I will rebase on master. So, please review only the last commit.~~

PR #1071, opened 2018-09-30
Inherit ENTRYPOINT and CMD from base image for WAR projects

CyrilleH:
Towards #431, here's my proposal to support inferring "entrypoint" from base images, for WAR project.

These configurations are no longer needed for Tomcat base images :
```xml
<entrypoint>
   <arg>catalina.sh</arg>
   <arg>run</arg>
</entrypoint>
```

```groovy
entrypoint = ['catalina.sh', 'run']
```

~~This PR is based from #1068.
Please review only the last commit.~~

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
@@ -158,12 +158,15 @@ public class BuildSteps {
 
     if (buildConfiguration.getContainerConfiguration() != null) {
       buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(""));
-      buildConfiguration
-          .getEventDispatcher()
-          .dispatch(
-              LogEvent.lifecycle(
-                  "Container entrypoint set to "
-                      + buildConfiguration.getContainerConfiguration().getEntrypoint()));
+      String eventMessage;
+      if (buildConfiguration.getContainerConfiguration().isEntrypointInferredFromBaseImage()) {
+        eventMessage = "Container entrypoint inferred from base image";
+      } else {
+        eventMessage =
+            "Container entrypoint set to "
+                + buildConfiguration.getContainerConfiguration().getEntrypoint();
+      }
+      buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(eventMessage));
     }
 
     return imageDigest;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
@@ -158,7 +158,7 @@ public class BuildSteps {
 
     if (buildConfiguration.getContainerConfiguration() != null) {
       buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(""));
-
+      // TODO refactor below
       if (buildConfiguration.getContainerConfiguration().getEntrypoint() != null) {
         buildConfiguration
             .getEventDispatcher()
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
@@ -158,15 +158,15 @@ public class BuildSteps {
 
     if (buildConfiguration.getContainerConfiguration() != null) {
       buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(""));
-      String eventMessage;
-      if (buildConfiguration.getContainerConfiguration().isEntrypointInferredFromBaseImage()) {
-        eventMessage = "Container entrypoint inferred from base image";
-      } else {
-        eventMessage =
-            "Container entrypoint set to "
-                + buildConfiguration.getContainerConfiguration().getEntrypoint();
+
+      if (!buildConfiguration.getContainerConfiguration().isEntrypointInferredFromBaseImage()) {
+        buildConfiguration
+            .getEventDispatcher()
+            .dispatch(
+                LogEvent.lifecycle(
+                    "Container entrypoint set to "
+                        + buildConfiguration.getContainerConfiguration().getEntrypoint()));
       }
-      buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(eventMessage));
     }
 
     return imageDigest;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
@@ -158,7 +158,8 @@ public class BuildSteps {
 
     if (buildConfiguration.getContainerConfiguration() != null) {
       buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(""));
-      // TODO refactor below
+      // TODO refactor code to also log ENTRYPOINT and CMD when inheriting them in this code,
+      // instead of logging them elsewhere.
       if (buildConfiguration.getContainerConfiguration().getEntrypoint() != null) {
         buildConfiguration
             .getEventDispatcher()
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
@@ -159,7 +159,7 @@ public class BuildSteps {
     if (buildConfiguration.getContainerConfiguration() != null) {
       buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(""));
 
-      if (!buildConfiguration.getContainerConfiguration().isEntrypointInferredFromBaseImage()) {
+      if (buildConfiguration.getContainerConfiguration().getEntrypoint() != null) {
         buildConfiguration
             .getEventDispatcher()
             .dispatch(
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
@@ -179,7 +179,8 @@ class BuildImageStep
         imageBuilder.setCreated(containerConfiguration.getCreationTime());
         imageBuilder.setUser(containerConfiguration.getUser());
         imageBuilder.setEntrypoint(computeEntrypoint(baseImage, containerConfiguration));
-        imageBuilder.setProgramArguments(computeProgramArguments(baseImage, containerConfiguration));
+        imageBuilder.setProgramArguments(
+            computeProgramArguments(baseImage, containerConfiguration));
         imageBuilder.setExposedPorts(containerConfiguration.getExposedPorts());
         imageBuilder.addLabels(containerConfiguration.getLabels());
       }
@@ -216,18 +217,18 @@ class BuildImageStep
         containerConfiguration.getEntrypoint() == null
             && containerConfiguration.getProgramArguments() == null;
 
-    if (shouldInheritProgramArguments && baseImage.getJavaArguments() != null) {
+    if (shouldInheritProgramArguments && baseImage.getProgramArguments() != null) {
       buildConfiguration
           .getEventDispatcher()
           .dispatch(
               LogEvent.lifecycle(
                   "Container program arguments set to "
-                      + baseImage.getJavaArguments()
+                      + baseImage.getProgramArguments()
                       + " (inherited from base image)"));
     }
 
     return shouldInheritProgramArguments
-        ? baseImage.getJavaArguments()
+        ? baseImage.getProgramArguments()
         : containerConfiguration.getProgramArguments();
   }
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
@@ -214,8 +214,7 @@ class BuildImageStep
       Image<Layer> baseImage, ContainerConfiguration containerConfiguration) {
     boolean shouldInheritProgramArguments =
         containerConfiguration.getEntrypoint() == null
-            && (containerConfiguration.getProgramArguments() == null
-                || containerConfiguration.getProgramArguments().isEmpty());
+            && containerConfiguration.getProgramArguments() == null;
 
     if (shouldInheritProgramArguments && baseImage.getJavaArguments() != null) {
       buildConfiguration
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
@@ -25,6 +25,7 @@ import com.google.cloud.tools.jib.builder.TimerEventDispatcher;
 import com.google.cloud.tools.jib.cache.CacheEntry;
 import com.google.cloud.tools.jib.configuration.BuildConfiguration;
 import com.google.cloud.tools.jib.configuration.ContainerConfiguration;
+import com.google.cloud.tools.jib.event.events.LogEvent;
 import com.google.cloud.tools.jib.image.DescriptorDigest;
 import com.google.cloud.tools.jib.image.Image;
 import com.google.cloud.tools.jib.image.Layer;
@@ -186,6 +187,27 @@ class BuildImageStep
                 : containerConfiguration.getProgramArguments());
         imageBuilder.setExposedPorts(containerConfiguration.getExposedPorts());
         imageBuilder.addLabels(containerConfiguration.getLabels());
+
+        if (containerConfiguration.isEntrypointInferredFromBaseImage()
+            && baseImage.getEntrypoint() != null) {
+          buildConfiguration
+              .getEventDispatcher()
+              .dispatch(
+                  LogEvent.lifecycle(
+                      "Container entrypoint set to "
+                          + baseImage.getEntrypoint()
+                          + " (inherited from base image)"));
+        }
+        if (containerConfiguration.isProgramArgumentsInferredFromBaseImage()
+            && baseImage.getJavaArguments() != null) {
+          buildConfiguration
+              .getEventDispatcher()
+              .dispatch(
+                  LogEvent.lifecycle(
+                      "Container program arguments set to "
+                          + baseImage.getJavaArguments()
+                          + " (inherited from base image)"));
+        }
       }
 
       // Gets the container configuration content descriptor.
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
@@ -174,22 +174,27 @@ class BuildImageStep
                 .build());
       }
       if (containerConfiguration != null) {
+        boolean shouldInheritEntrypoint = containerConfiguration.getEntrypoint() == null;
+        boolean shouldInheritProgramArguments =
+            containerConfiguration.getEntrypoint() == null
+                && (containerConfiguration.getProgramArguments() == null
+                    || containerConfiguration.getProgramArguments().isEmpty());
+
         imageBuilder.addEnvironment(containerConfiguration.getEnvironmentMap());
         imageBuilder.setCreated(containerConfiguration.getCreationTime());
         imageBuilder.setUser(containerConfiguration.getUser());
         imageBuilder.setEntrypoint(
-            containerConfiguration.isEntrypointInferredFromBaseImage()
+            shouldInheritEntrypoint
                 ? baseImage.getEntrypoint()
                 : containerConfiguration.getEntrypoint());
         imageBuilder.setProgramArguments(
-            containerConfiguration.isProgramArgumentsInferredFromBaseImage()
+            shouldInheritProgramArguments
                 ? baseImage.getJavaArguments()
                 : containerConfiguration.getProgramArguments());
         imageBuilder.setExposedPorts(containerConfiguration.getExposedPorts());
         imageBuilder.addLabels(containerConfiguration.getLabels());
 
-        if (containerConfiguration.isEntrypointInferredFromBaseImage()
-            && baseImage.getEntrypoint() != null) {
+        if (shouldInheritEntrypoint && baseImage.getEntrypoint() != null) {
           buildConfiguration
               .getEventDispatcher()
               .dispatch(
@@ -198,8 +203,7 @@ class BuildImageStep
                           + baseImage.getEntrypoint()
                           + " (inherited from base image)"));
         }
-        if (containerConfiguration.isProgramArgumentsInferredFromBaseImage()
-            && baseImage.getJavaArguments() != null) {
+        if (shouldInheritProgramArguments && baseImage.getJavaArguments() != null) {
           buildConfiguration
               .getEventDispatcher()
               .dispatch(
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
@@ -190,45 +190,58 @@ class BuildImageStep
     }
   }
 
+  /**
+   * Computes the image entrypoint. If {@link ContainerConfiguration#getEntrypoint()} is null, the
+   * entrypoint is inherited from the base image. Otherwise {@link
+   * ContainerConfiguration#getEntrypoint()} is returned.
+   *
+   * @param baseImage the base image
+   * @param containerConfiguration the container configuration
+   * @return the container entrypoint
+   */
   @Nullable
   private ImmutableList<String> computeEntrypoint(
       Image<Layer> baseImage, ContainerConfiguration containerConfiguration) {
-    boolean shouldInheritEntrypoint = containerConfiguration.getEntrypoint() == null;
-
-    if (shouldInheritEntrypoint && baseImage.getEntrypoint() != null) {
-      buildConfiguration
-          .getEventDispatcher()
-          .dispatch(
-              LogEvent.lifecycle(
-                  "Container entrypoint set to "
-                      + baseImage.getEntrypoint()
-                      + " (inherited from base image)"));
+    if (baseImage.getEntrypoint() == null || containerConfiguration.getEntrypoint() != null) {
+      return containerConfiguration.getEntrypoint();
     }
 
-    return shouldInheritEntrypoint
-        ? baseImage.getEntrypoint()
-        : containerConfiguration.getEntrypoint();
+    buildConfiguration
+        .getEventDispatcher()
+        .dispatch(
+            LogEvent.lifecycle(
+                "Container entrypoint set to "
+                    + baseImage.getEntrypoint()
+                    + " (inherited from base image)"));
+    return baseImage.getEntrypoint();
   }
 
+  /**
+   * Computes the image program arguments. If {@link ContainerConfiguration#getEntrypoint()} and
+   * {@link ContainerConfiguration#getProgramArguments()} are null, the program arguments are
+   * inherited from the base image. Otherwise {@link ContainerConfiguration#getProgramArguments()}
+   * is returned.
+   *
+   * @param baseImage the base image
+   * @param containerConfiguration the container configuration
+   * @return the container program arguments
+   */
   @Nullable
   private ImmutableList<String> computeProgramArguments(
       Image<Layer> baseImage, ContainerConfiguration containerConfiguration) {
-    boolean shouldInheritProgramArguments =
-        containerConfiguration.getEntrypoint() == null
-            && containerConfiguration.getProgramArguments() == null;
-
-    if (shouldInheritProgramArguments && baseImage.getProgramArguments() != null) {
-      buildConfiguration
-          .getEventDispatcher()
-          .dispatch(
-              LogEvent.lifecycle(
-                  "Container program arguments set to "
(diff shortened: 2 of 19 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
```diff
@@ -137,12 +137,15 @@ public void run() throws InterruptedException, ExecutionException {
...
       buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(""));
-      buildConfiguration
-          .getEventDispatcher()
-          .dispatch(
-              LogEvent.lifecycle(
-                  "Container entrypoint set to "
-                      + buildConfiguration.getContainerConfiguration().getEntrypoint()));
+      String eventMessage;
+      if (buildConfiguration.getContainerConfiguration().isEntrypointInferredFromBaseImage()) {
+        eventMessage = "Container entrypoint inferred from base image";
```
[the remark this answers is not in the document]
CyrilleH: What do you thing about moving this code into `BuildImageSteps.afterCacheEntrySteps()` method ?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -209,6 +247,14 @@ public Instant getCreationTime() {
     return labels;
   }
 
+  public boolean isEntrypointInferredFromBaseImage() {
```
coollog: Wording nit: Since this is the configuration, the flag is more of like "should entrypoint be inferred from base image?" rather than "is entrypoint inferred from base image?". Perhaps this could be named `shouldInheritEntrypoint` (base image is implied)?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -209,6 +247,14 @@ public Instant getCreationTime() {
...
+    return isEntrypointInferredFromBaseImage;
```
coollog: I'm also thinking perhaps we should have the field `entrypoint` as `null` to indicate that it should be inherited. The API for `ContainerConfiguration` would be the same, but the builder's setter could be called `setInheritEntrypoint` and it would just set the entrypoint to null. This would make it so that we never have a state where `getEntrypoint` returns something while `shouldInheritEntrypoint` is true.

coollog: @GoogleContainerTools/java-tools ?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
CyrilleH: Yes it's better.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
chanseokoh: I am also leaning toward to not having this extra flag but interpreting the `null` value of `entrypoint` as inheriting it from the base image. And I think we can apply this policy consistently across other fields.

Does it need an additional `setInheritEntrypoint`? The builder's `setEntrypoint` already accepts `null`.

chanseokoh: And if we go for the null-means-inherit, I think we also need to consider the validity of possible `Entrypoint` and `Cmd` combinations. For example, with `Dockerfile`, if you define a new `Entypoint` but no `Cmd`, `Cmd` is not inherited but set to `null` explicitly in the image metadata. On the other hand, if you define a new `Cmd` but no `Entrypoint`, you inherit `Entrypoint`. So, if `Entrypoint` is non-null, we shouldn't inherit `Cmd` even if `Cmd` is null.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -173,12 +174,43 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
                 .build());
       }
       if (containerConfiguration != null) {
+        boolean shouldInheritEntrypoint = containerConfiguration.getEntrypoint() == null;
+        boolean shouldInheritProgramArguments =
+            containerConfiguration.getEntrypoint() == null
+                && (containerConfiguration.getProgramArguments() == null
+                    || containerConfiguration.getProgramArguments().isEmpty());
```
coollog: Hmm, I think it should not inherit the base image program arguments if the program arguments configuration was set to an empty list.

CyrilleH: But it does not work since by default it's set to an empty list.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -173,12 +174,43 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
                 .build());
       }
       if (containerConfiguration != null) {
+        boolean shouldInheritEntrypoint = containerConfiguration.getEntrypoint() == null;
+        boolean shouldInheritProgramArguments =
```
loosebazooka: What do you think about breaking this out for readability.
```
imageBuilder.setEntryPoint(calculateEntrypoint(containerConfiguration, baseImage))
imageBuilder.setJavaArguments(calculateProgramArguments(containerConfiguration, baseImage))
```
and implement the `calculateX methods`

CyrilleH: we need these boolean for these logs : 
```
if (shouldInheritEntrypoint && baseImage.getEntrypoint() != null) {
          buildConfiguration
              .getEventDispatcher()
              .dispatch(
                  LogEvent.lifecycle(
                      "Container entrypoint set to "
                          + baseImage.getEntrypoint()
                          + " (inherited from base image)"));
        }
        if (shouldInheritProgramArguments && baseImage.getJavaArguments() != null) {
          buildConfiguration
              .getEventDispatcher()
              .dispatch(
                  LogEvent.lifecycle(
                      "Container program arguments set to "
                          + baseImage.getJavaArguments()
                          + " (inherited from base image)"));
        }
```

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/DockerContextMojoTest.java
```diff
@@ -259,4 +273,10 @@ private String getBaseImage() throws IOException {
     List<String> lines = Files.readAllLines(dockerfile);
     return lines.stream().filter(line -> line.startsWith("FROM")).findFirst().get();
   }
+
+  private String getCmd() throws IOException {
+    Path dockerfile = projectRoot.getRoot().toPath().resolve("target/Dockerfile");
+    List<String> lines = Files.readAllLines(dockerfile);
+    return lines.stream().filter(line -> line.startsWith("CMD")).findFirst().get();
+  }
```
chanseokoh: How about removing code duplication?

```java
private String getEntrypoint() {
  getDockerfileLine("ENTRYPOINT");
}

private String getBaseImage() {
  getDockerfileLine("FROM");
}

private String getCmd() {
  getDockerfileLine("CMD");
}

private String getDockerfileLine(String command) {
  Path dockerfile = projectRoot.getRoot().toPath().resolve("target/Dockerfile");
  List<String> lines = Files.readAllLines(dockerfile);
  return lines.stream().filter(line -> line.startsWith(command)).findFirst().get();
}
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -185,4 +187,47 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
       return imageBuilder.build();
     }
   }
+
+  @Nullable
+  private ImmutableList<String> computeEntrypoint(
+      Image<Layer> baseImage, ContainerConfiguration containerConfiguration) {
+    boolean shouldInheritEntrypoint = containerConfiguration.getEntrypoint() == null;
```
chanseokoh: This is not consistent with program arguments in that it doesn't inherit if it is an empty list. Reason?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -185,4 +187,47 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
...
+        : containerConfiguration.getEntrypoint();
+  }
+
+  @Nullable
+  private ImmutableList<String> computeProgramArguments(
+      Image<Layer> baseImage, ContainerConfiguration containerConfiguration) {
+    boolean shouldInheritProgramArguments =
+        containerConfiguration.getEntrypoint() == null
+            && (containerConfiguration.getProgramArguments() == null
+                || containerConfiguration.getProgramArguments().isEmpty());
```
chanseokoh: `Strings.isNullOrEmpty` is handy here. But then, this is not consistent with ENTRYPOINT, as I said.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
chanseokoh: Yeah, I think we should decide whether to inherit CMD or not in the same way as ENTRYPOINT.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -185,4 +187,47 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
...
+
+    if (shouldInheritProgramArguments && baseImage.getJavaArguments() != null) {
+      buildConfiguration
+          .getEventDispatcher()
+          .dispatch(
+              LogEvent.lifecycle(
```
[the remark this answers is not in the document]
CyrilleH: It was a copy-paste. But I can change that.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -185,4 +187,47 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
...
+  @Nullable
+  private ImmutableList<String> computeEntrypoint(
+      Image<Layer> baseImage, ContainerConfiguration containerConfiguration) {
+    boolean shouldInheritEntrypoint = containerConfiguration.getEntrypoint() == null;
+
+    if (shouldInheritEntrypoint && baseImage.getEntrypoint() != null) {
+      buildConfiguration
+          .getEventDispatcher()
+          .dispatch(
+              LogEvent.lifecycle(
```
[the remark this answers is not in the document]
CyrilleH: A copy-paste from BuildSteps.

Do you want me to change BuildSteps as well (for consistency) ?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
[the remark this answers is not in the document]
CyrilleH: If I change that, before : 
```
Retrieving registry credentials for localhost:5000...
Getting base image localhost:5000/tomcat_light:8.5-jre8-alpine...
Building resources layer...
Building classes layer...
Finalizing...
Container program arguments set to [catalina.sh, run] (inherited from base image)


Built and pushed image as localhost:5000/my-image/image
```

after : 
```
Retrieving registry credentials for localhost:5000...
Getting base image localhost:5000/tomcat_light:8.5-jre8-alpine...
Building resources layer...
Building classes layer...
Finalizing...


Built and pushed image as localhost:5000/my-image/image
```

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -326,11 +325,12 @@ String makeDockerfile() throws JsonProcessingException {
...
-    dockerfile
-        .append("\nENTRYPOINT ")
-        .append(objectMapper.writeValueAsString(entrypoint))
-        .append("\nCMD ")
-        .append(objectMapper.writeValueAsString(javaArguments));
+    if (entrypoint != null) {
+      dockerfile.append("\nENTRYPOINT ").append(objectMapper.writeValueAsString(entrypoint));
+    }
+    if (entrypoint != null || !javaArguments.isEmpty()) {
+      dockerfile.append("\nCMD ").append(objectMapper.writeValueAsString(javaArguments));
```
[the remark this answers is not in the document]
CyrilleH: I think it does not work with tomcat.
I you look in docker hub (and I tested it), the entrypoint is null and cmd isn't.

> How to use this image.
> Run the default Tomcat server (CMD ["catalina.sh", "run"]):

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
CyrilleH: The type is `ImmutableList<String>` not `String`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
CyrilleH: Tomcat have no entrypoing but cmd. I changed how to deal with it. Tell me if it suits you now.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/DockerContextMojoTest.java
CyrilleH: done

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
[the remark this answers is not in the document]
CyrilleH: I moved the logic into plugins.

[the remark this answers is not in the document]
CyrilleH: See `PluginConfigurationProcessor.computeProgramArguments`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/PluginConfigurationProcessor.java
```diff
@@ -249,14 +253,32 @@ static ImageReference parseImageReference(String image, String type) {
...
     }
 
     String mainClass = projectProperties.getMainClass(jibPluginConfiguration);
     return JavaEntrypointConstructor.makeDefaultEntrypoint(
         getAppRootChecked(jibPluginConfiguration), jibPluginConfiguration.getJvmFlags(), mainClass);
   }
 
+  /**
+   * Compute the container program arguments. If entrypoint is inherited (null), program arguments
+   * must be inherited (null) if empty (used for Tomcat base images).
```
chanseokoh: This special handling for Tomcat and the inconsistency between the entrypoint being null and the program arguments being empty bother me. Ideally, I think the entrypoint and the program arguments could be nullable and start with null at the beginning instead of the current defaulting to empty, indicating Jib will by default inherit entrypoint and cmd. Then for non-WAR projects, we are already setting the entrypoint and the cmd explicitly, so we still won't inherit them for the non-WAR projects as usual. That is, the non-WAR is a special case that overrides the default. On the other hand, for WAR, we would not touch the null entrypoint and cmd, and inheriting them would be just the default behavior. Also, I imagine fixing #924 fits naturally in this model, since we can later add an option to not override them in the non-WAR case (through some configuration or so). @GoogleContainerTools/java-tools Thoughts?

coollog: I agree - we should have the default be `null` and set to empty if needed (so non-WAR would set entrypoint as generated and program arguments to empty).

CyrilleH: I have a problem. If the user set the ENTRYPOINT to `[catalina.sh, run]` the CMD is null so it will be inherited and we will ENTRYPOINT and CMD equal to `[catalina.sh, run]`.

Can I set CMD to an empty list if the ENTRYPOINT is not null ?

chanseokoh: I think it's right to not inherit CMD when ENTRYPOINT is redefined: https://github.com/GoogleContainerTools/jib/pull/1071#discussion_r223431180

I'm just wondering if `CMD []` in `Dockerfile` will set the value of the CMD metadata to null or an empty string. I'll have to double-check.

chanseokoh: > Can I set CMD to an empty list if the ENTRYPOINT is not null ?

Hmm... this actually is tricky.

The value of an empty list is supposed to mean `CMD []`, which will specifically define CMD (and override the base value) to be an empty string. And we are trying to set the behavior here that the null value means inheriting CMD from the base image. That said, there is way to explicitly set CMD to null in the container metadata. Sure, setting to an empty list might be a feasible choice, but this diverges from how Docker build actually works, although I'm not saying we always have to stick to how Docker build works.

chanseokoh: > there is way to explicitly set CMD to null

Mistyped: there is *no* way to ...

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
```diff
@@ -188,14 +192,32 @@ static PluginConfigurationProcessor processCommonConfiguration(
...
+   * must not be inherited .
+   *
+   * @param entrypoint the container entrypoint
+   * @param jibExtension the {@link JibExtension} providing the configuration data
+   * @return the program arguments
+   */
+  @Nullable
+  static List<String> computeProgramArguments(
+      @Nullable List<String> entrypoint, JibExtension jibExtension) {
+    if (entrypoint != null && jibExtension.getContainer().getArgs() == null) {
```
coollog: I believe this is duplicating the logic in `BuildImageSteps` - I think we could just use `jibExtension.getContainer().getArgs()` directly.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/DockerContextTaskTest.java
```diff
@@ -83,16 +84,29 @@ public void testEntrypoint_nonDefaultAppRoot() throws IOException {
...
+    Mockito.when(containerParameters.getArgs()).thenCallRealMethod();
     project.getPluginManager().apply("war");
 
     task.generateDockerContext();
 
-    Assert.assertEquals("ENTRYPOINT [\"java\",\"-jar\",\"/jetty/start.jar\"]", getEntrypoint());
+    try {
+      getEntrypoint();
+      Assert.fail();
+    } catch (NoSuchElementException e) {
```
coollog: nit: s/`e`/`ex`

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/DockerContextTaskTest.java
```diff
@@ -83,16 +84,29 @@ public void testEntrypoint_nonDefaultAppRoot() throws IOException {
...
+      Assert.assertEquals("No value present", e.getMessage());
```
coollog: If this is the generic message for `NoSuchElementException`, we can just replace this with `// pass`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -128,7 +128,7 @@ public void set(String image) {
 
     @Nullable @Parameter private String mainClass;
 
-    @Parameter private List<String> args = Collections.emptyList();
+    @Nullable @Parameter private List<String> args = null;
```
coollog: s/` = null`//

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/PluginConfigurationProcessor.java
```diff
@@ -249,14 +254,32 @@ static ImageReference parseImageReference(String image, String type) {
...
+  /**
+   * Compute the container program arguments. If the entrypoint is not inherited, program arguments
+   * must not be inherited .
+   *
+   * @param entrypoint the container entrypoint
+   * @param containerArgs the container arguments
+   * @return the program arguments
+   */
+  @Nullable
+  static List<String> computeProgramArguments(
```
coollog: ditto this

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/DockerContextMojoTest.java
```diff
@@ -205,7 +208,18 @@ public void testEntrypoint_defaultWarPackaging() throws MojoExecutionException,
     Mockito.doReturn(projectRoot.getRoot().toString()).when(build).getDirectory();
     mojo.execute();
 
-    Assert.assertEquals("ENTRYPOINT [\"java\",\"-jar\",\"/jetty/start.jar\"]", getEntrypoint());
+    try {
+      getEntrypoint();
+      Assert.fail();
+    } catch (NoSuchElementException e) {
+      Assert.assertEquals("No value present", e.getMessage());
```
coollog: ditto

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/DockerContextMojoTest.java
coollog: Let's do this for `DockerContextTaskTest` as well.

CyrilleH: yes obviously.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
chanseokoh: > But I have to change `JavaDockerContextGenerator#makeDockerfile` to add `CMD []` if `ENTRYPOINT` is not null.

The correct way should be to not print anything about `CMD`, because if there exists `ENTRYPOINT` which will override the base value, Docker build will automatically reset CMD to null. (And null CMD is different from an empty list CMD.) But the problem we have here, I think (correct me if I'm getting something wrong, haven't looked into the overall code in detail), is that for the image built from our generated `Dockerfile` to be consistent with the image we build directly, we also have to reset CMD in the container metadata instead of setting it to an empty list.  And that seems a bit tricky for now: https://github.com/GoogleContainerTools/jib/pull/1071#discussion_r223760143 Not sure what would be the right/best way.

CyrilleH: But I didn't change this behaviour. 
`CMD []` was already there before this PR if we set only the entrypoint.

chanseokoh: > But I haven't change this behaviour. CMD [] was already there before this PR if we set only the entrypoint.

Sure, that was and will be the behavior that we explicitly override our `ENTRYPOINT` and `CMD` for non-WAR projects. In this case, we will always override `CMD` to `[]`. I'm talking about the case (currently this can apply to WAR projects) where the user is overriding `ENTRYPOINT` and left `CMD` undefined, in which case, `CMD` should be set to null instead of `[]`.

chanseokoh: Actually, since we are enabling Jib to inherit ENTRYPOINT and CMD, I think we should eventually change the behavior for the non-WAR case too, when we find a way to fix #924 which will allow users to inherit either ENTRYPOINT or CMD for non-WAR. After fixing #924, I imagine that the same rule for inheriting ENTRYPOINT and CMD will apply to both non-WAR and WAR.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -186,4 +189,59 @@ public DescriptorDigest getDiffId() throws LayerPropertyNotFoundException {
...
+   * @param baseImage the base image
+   * @param containerConfiguration the container configuration
+   * @return the container program arguments
+   */
+  @Nullable
+  private ImmutableList<String> computeProgramArguments(
+      Image<Layer> baseImage, ContainerConfiguration containerConfiguration) {
+    if (baseImage.getProgramArguments() == null
+        || containerConfiguration.getEntrypoint() != null
+        || containerConfiguration.getProgramArguments() != null) {
```
chanseokoh: Looks like `ContainerConfiguration.getProgramArguments()` can never be null.

- `ContainerConfiguration.programArguments` is set to `JibContainerBuilder.programArguments` in `JibContainerBuilder.toContainerConfiguration()`.
- `JibContainerBuilder.programArguments` is in turn set to `jibExtension.getContainer().getArgs()` in `PluginConfigurationProcessor.processCommonConfiguration()` for Gradle.
- `jibExtension.getContainer().getArgs()` is not `@Nullable`, so never null.

That means, we will never inherit program arguments, right? I think `PluginConfigurationProcessor.processCommonConfiguration()` should set `JibContainerBuilder.programArguments` to `null` if WAR, similar to what this PR does with Entrypoint in `computeEntrypoint`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -339,11 +338,15 @@ String makeDockerfile() throws JsonProcessingException {
...
-        .append("\nCMD ")
-        .append(objectMapper.writeValueAsString(programArguments));
+    if (entrypoint != null) {
+      dockerfile.append("\nENTRYPOINT ").append(objectMapper.writeValueAsString(entrypoint));
+    }
+    if (programArguments != null) {
+      dockerfile.append("\nCMD ").append(objectMapper.writeValueAsString(programArguments));
+    } else if (entrypoint != null) {
+      // Do not inherit CMD from base image if ENTRYPOINT is not inherited
+      dockerfile.append("\nCMD []");
```
chanseokoh: I think we don't need this. if `entrypoint` is not null, `Dockefile` will have the line `ENTRYPOINT` but no `CMD`, and Docker build won't inherit CMD.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
chanseokoh: Wait, I do see the program arguments are being inherited in my actual small testing. Looking into what's happening.

chanseokoh: Sorry, never mind. I was looking at a wrong branch.

chanseokoh:
Just making a note: integration tests will fail, but I think it will be difficult for an external contributor to fix them. It's probably more productive for us to fix them later.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
chanseokoh: I looked into the code, and updated my previous comments in this particular conversation about the CMD being null vs. an empty list.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
```diff
@@ -158,12 +158,15 @@ public DescriptorDigest run() throws InterruptedException, ExecutionException {
...
     if (buildConfiguration.getContainerConfiguration() != null) {
       buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle(""));
-      buildConfiguration
-          .getEventDispatcher()
-          .dispatch(
-              LogEvent.lifecycle(
-                  "Container entrypoint set to "
-                      + buildConfiguration.getContainerConfiguration().getEntrypoint()));
+
+      if (buildConfiguration.getContainerConfiguration().getEntrypoint() != null) {
```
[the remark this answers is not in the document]
CyrilleH: done

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
CyrilleH: Ok, I've removed this.

chanseokoh:
FTR: we will stop adding `CMD []` to the generated `Dockerfile` after #1071. Our `CHANGELOG` files already have an entry saying "`jib:exportDockerContext` generates different directory layout and `Dockerfile` ...", so I think we don't have to add another "Changed" entry for this.

chanseokoh:
Thanks, @CyrilleH. The PR is merged.

coollog:
Thanks again for your contribution!

PR #1075, opened 2018-10-01
Removes CacheConfiguration in favor of setting directories directly on BuildConfiguration.

coollog:
This helps with implementing `JibContainerBuilder#containerize`. `CacheConfiguration` seems like an unnecessary and clumsy extra class to use to just configure basic cache directory paths.


jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -28,6 +32,14 @@
 // TODO: Add tests once JibContainerBuilder#containerize() is added.
 public class Containerizer {
 
+  /**
+   * The default directory for caching the base image layers, in {@code [user cache
+   * home]/google-cloud-tools-java/jib}.
+   */
+  // TODO: Reduce scope once plugins are migrated to use the new Jib Core API.
```
briandealwis: Seems an odd comment when this is a newly-introduced field?

coollog: This is currently used to set the default in `PluginsConfigurationProcessor`. However, once the plugins are migrated to use the new API (and this class), it won't need to use this field to set defaults as this class sets the default itself.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -106,8 +139,23 @@ TargetImage getTargetImage() {
     return Optional.ofNullable(executorService);
   }
 
-  Optional<CacheConfiguration> getCacheConfiguration() {
-    return Optional.ofNullable(cacheConfiguration);
+  Path getBaseImageLayersCacheDirectory() {
+    return baseImageLayersCacheDirectory;
+  }
+
+  Path getApplicationLayersCacheDirectory() throws CacheDirectoryCreationException {
```
chanseokoh: Who's calling this? Can we guarantee this is not called concurrently?

coollog: This is to-be-called by `JibContainerBuilder#containerize`. None of the API classes are thread-safe and should not be called concurrently.

PR #1076, opened 2018-10-01
Adds JibContainerBuilder#containerize.

### LINKED ISSUE #950: Add JibContainer and JibContainerBuilder#containerize.
Continuing from #949 ...

`JibContainer` is the build result of `JibContainerBuilder#containerize`. API:
- `DescriptorDigest getDigest()`
- `ImageReference getImageReference`?

`JibContainerBuilder#containerize` launches the Jib execution. API:
- `JibContainer containerize(Containerizer containerizer)`

coollog:
Fixes #950 

This currently has a single basic integration test. Follow-ups will add more integration tests.


kokoro/presubmit.sh
```diff
@@ -12,6 +12,8 @@ docker kill $(docker ps --all --quiet) || true
 
 cd github/jib
 
+rm /usr/local/bin/docker-credential-osxkeychain || true
```
TadCordle: What's this for? If we still need it, maybe add a comment why.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainer.java
```diff
@@ -0,0 +1,38 @@
...
+  JibContainer(DescriptorDigest imageDigest) {
+    this.imageDigest = imageDigest;
+  }
+
+  /**
+   * Gets the digest of the container image built by Jib.
+   *
+   * @return the image digest
+   */
+  DescriptorDigest getDigest() {
```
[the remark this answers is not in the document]
coollog: Yes

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -131,6 +132,17 @@ public Containerizer setEventHandlers(EventHandlers eventHandlers) {
     return this;
   }
 
+  /**
+   * Sets whether or not to allow communication over HTTP (as opposed to HTTPS by default).
```
chanseokoh: Not just about HTTP; it allows to communicate over insecure HTTPS too.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -301,20 +305,63 @@ public JibContainerBuilder addLabel(String key, String value) {
...
+   * Builds a {@link BuildConfiguration} using this and a {@link Containerizer}.
+   *
+   * @param buildConfigurationBuilder the {@link BuildConfiguration.Builder} to use
+   * @param containerizer the {@link Containerizer}
+   * @return the {@link BuildConfiguration}
+   * @throws CacheDirectoryCreationException if a cache directory could not be created
+   * @throws IOException if an I/O exception occurs
+   */
+  @VisibleForTesting
+  BuildConfiguration toBuildConfiguration(
```
chanseokoh: This is a private method used only be `containerize()`. Do we have to accept `buildConfigurationBuilder`? Can't this method just create it with `BuildConfiguration.builder()`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -301,20 +305,63 @@ public JibContainerBuilder addLabel(String key, String value) {
...
-        .setTargetImageConfiguration(targetImage.toImageConfiguration())
+        .setTargetImageConfiguration(containerizer.getTargetImage().toImageConfiguration())
+        .setBaseImageLayersCacheDirectory(containerizer.getBaseImageLayersCacheDirectory())
+        .setApplicationLayersCacheDirectory(containerizer.getApplicationLayersCacheDirectory())
         .setContainerConfiguration(toContainerConfiguration())
-        .setLayerConfigurations(layerConfigurations);
+        .setLayerConfigurations(layerConfigurations)
+        .setAllowInsecureRegistries(containerizer.getAllowInsecureRegistries());
+
+    if (containerizer.getExecutorService().isPresent()) {
```
chanseokoh: Maybe `containerizer.getExecutorService.ifPresnet(buildConfigurationBuilder::setExecutorService)` will work.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
```diff
@@ -64,19 +66,29 @@ public static CredentialRetrieverFactory forImage(
...
 
-  private final EventDispatcher eventDispatcher;
+  /**
+   * Creates a new {@link CredentialRetrieverFactory} for an image.
+   *
+   * @param imageReference the image the credential are for
+   * @return a new {@link CredentialRetrieverFactory}
+   */
+  public static CredentialRetrieverFactory forImage(ImageReference imageReference) {
+    return new CredentialRetrieverFactory(imageReference, null);
```
chanseokoh: Pass a no-op consumer and remove `@Nullable` and the extra method?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: This is mostly for the test to be able to pass a `BuildConfiguration.Builder` to verify.

chanseokoh: I see, that makes sense. Just wondering if you can't directly examine the built `BuildConfiguration` using getters.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
```diff
@@ -70,16 +73,17 @@ public DockerDaemonImage setDockerExecutable(Path dockerExecutable) {
...
-   *
-   * @return an {@link ImageConfiguration}
-   */
   @Override
   public ImageConfiguration toImageConfiguration() {
     return ImageConfiguration.builder(imageReference).build();
   }
 
+  @Override
+  public BuildSteps toBuildSteps(BuildConfiguration buildConfiguration) {
```
loosebazooka: This feels a little weird. I would expect image doesn't need to know about the BuildSteps class?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
coollog: Yea, it was mostly for the two caches that get generated on `BuildConfiguration.Builder#build` and didn't want to add a getter and new field to `Cache` just for this test.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
coollog: The `BuildSteps` only vary based on the image target (we should probably rename the interfaces to `ImageTarget` and `ImageSource`) - and generating the `BuildSteps` in `JibContainerBuilder`, for example, would involve logic around checking for being instance of `DockerDaemonImage`.

PR #1078, opened 2018-10-01
Makes BuildSteps#run return the built image digest.

coollog:
This is useful for `JibContainerBuilder#containerize` since it needs to return the built image digest.


jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -85,8 +92,28 @@ public Void handleHttpResponseException(HttpResponseException httpResponseExcept
...
+        if (expectedDigest.equals(receivedDigest)) {
+          return expectedDigest;
+        }
+
+      } catch (DigestException ex) {
+        // Invalid digest.
+      }
+    }
+
+    throw new UnexpectedImageDigestException(expectedDigest, receivedDigests);
```
chanseokoh: Just be cautious here, since this changes the current behavior in that Jib will start to fail for docker registries that do not return this header (probably not conforming to the spec). Do you think this will be OK? Like, what if we just show a warning and proceed with returning `expectedDigest`?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -85,8 +92,28 @@ public Void handleHttpResponseException(HttpResponseException httpResponseExcept
...
+  public DescriptorDigest handleResponse(Response response)
+      throws IOException, UnexpectedImageDigestException {
+    // Checks if the image digest is as expected.
+    DescriptorDigest expectedDigest =
+        JsonTemplateMapper.toBlob(manifestTemplate)
+            .writeTo(ByteStreams.nullOutputStream())
+            .getDigest();
+
+    List<String> receivedDigests = response.getHeader(RESPONSE_DIGEST_HEADER);
+    if (receivedDigests.size() == 1) {
```
chanseokoh: Any chance the server returns multiple headers for some reason and one of the values is the expected match?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
coollog: That's a good idea, I'll make this just a warning.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
coollog: I would not think so, at least the spec indicates that only one header would be returned

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPusher.java
```diff
@@ -18,30 +18,65 @@
...
+   * Makes the warning for when the registry responds with an image digest that is not the expected
+   * digest of the image.
+   *
+   * @param expectedDigest the expected image digest
+   * @param receivedDigests the received image digests
+   * @return the warning message
+   */
+  private static String makeUnexpectedImageDigestWarning(
+      DescriptorDigest expectedDigest, List<String> receivedDigests) {
+    if (receivedDigests.size() == 0) {
```
chanseokoh: `isEmpty()`?

PR #1080, opened 2018-10-01
Add system properties for remaining maven config

### LINKED ISSUE #728: Add clear and non-ambiguous maven properties to the maven plugin
When using Maven, often one wants to pass some of the plugin configuration via properties (`-DmyProp`).

Currently it seems some of the configuration are available via properties but:
- it's not documented anywhere (usually plugins have each configuration mapped to an `expression` that is resolved by maven, and it appears in plugin doc as well as in IDE)
- it's not clear if things like `credHelper`, if it is available as a property, is going to be applied to `to`, `from` or both.

It would be great if all the configuration could be specified via properties named like `jib.to.credHelper` and so on.

TadCordle:
Fixes #728 (~~although we may want to do this for Gradle as well~~ Filed #1083 for Gradle)

Adds clear command-line properties for all maven parameters that were missing one. Each one follows the convention of `jib.[object].[config]` and matches the configuration parameter names (e.g. `jib.from.image`, `jib.container.entrypoint`, `jib.extraDirectory`, etc.)

`<to><image>` has two corresponding properties: the existing shorthand `image`, and a full `jib.to.image` to match the other properties.

List/set parameters can be passed using a comma-separated list, where comma values can be escaped using a backslash `\`
e.g. `-Djib.to.tags=tag1,tag2,ta\,g3`

Map parameters can be passed using comma-separated entries with separated key/value pairs separated with `=`
e.g. `-Djib.container.environment=key1=value1,key2=value2`


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +134,59 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+  /**
+   * Parses a string in the form of "key1=value1,key2=value2,..." into a map.
+   *
+   * @param property the map string to parse, with entries separated by "," and key-value pairs
+   *     separated by "="
+   * @return the map of parsed values
+   */
+  public static Map<String, String> parseMapProperty(String property) {
+    // Matches key-value pairs in the form of "key=value"
+    Pattern environmentPattern = Pattern.compile("(?<name>[^=]+)=(?<value>.*)");
```
coollog: nit: Making this static would avoid multiple compilations.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +134,59 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+
+  /**
+   * Parses a comma-separated string into a list. Ignores commas escaped with "\".
+   *
+   * @param property the comma-separated string
+   * @return the list of parsed values
+   */
+  public static List<String> parseListProperty(String property) {
+    List<String> result = new ArrayList<>();
+    boolean evenConsecutiveEscapes = true;
```
coollog: I think this may be clearer if called `shouldEscape`? (and have it be inverted)

And then the loop would be like:
```
loop {
  if shouldEscape, then shouldEscape=false and skip
  if character is comma, do the logic
  if character is backslash, shouldEscape=true
}
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +134,59 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+    return result;
+  }
+
+  /**
+   * Parses a comma-separated string into a list. Ignores commas escaped with "\".
+   *
+   * @param property the comma-separated string
+   * @return the list of parsed values
+   */
+  public static List<String> parseListProperty(String property) {
```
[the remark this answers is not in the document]
TadCordle: I think everywhere else we parse lists, we can just do it automatically with a json translator. But we can't really do that since the format here is slightly different. I was going off of the format I saw [in this answer](https://stackoverflow.com/questions/31620967/any-way-to-specify-a-fileset-as-command-line-parameter), and assumed commas should be escaped using `\` since `\` is such a common escape character.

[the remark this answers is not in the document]
chanseokoh: Not that this is unacceptable. Just giving more thoughts, and I kinda like I have seen this key-value parsing somewhere.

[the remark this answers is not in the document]
TadCordle: We do have the same regex pattern for environment variables in `JsonToImageTranslator`, but in that case the list is already split up, so we only need to parse the keys/values from each entry.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -35,6 +35,9 @@
 /** Validator for plugin configuration parameters and system properties. */
 public class ConfigurationPropertyValidator {
 
+  /** Matches key-value pairs in the form of "key=value" */
+  private static Pattern environmentPattern = Pattern.compile("(?<name>[^=]+)=(?<value>.*)");
```
coollog: final

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +137,58 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+      if (property.charAt(endIndex) == ',') {
+        // Split on non-escaped comma
+        result.add(property.substring(startIndex, endIndex));
+        startIndex = endIndex + 1;
+      } else if (property.charAt(endIndex) == '\\') {
+        // Found a backslash, ignore next character
+        shouldEscape = true;
+      }
+    }
+    result.add(property.substring(startIndex, endIndex));
```
coollog: nit: `result.add(property.substring(startIndex));`

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +137,58 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+  }
+
+  /**
+   * Parses a comma-separated string into a list. Ignores commas escaped with "\".
+   *
+   * @param property the comma-separated string
+   * @return the list of parsed values
+   */
+  public static List<String> parseListProperty(String property) {
+    List<String> result = new ArrayList<>();
```
coollog: nit: Maybe `items` would be a more descriptive name?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidatorTest.java
```diff
@@ -144,4 +146,29 @@ public void testGetGeneratedTargetDockerTag() throws InvalidImageReferenceExcept
...
+    Assert.assertEquals(
+        ImmutableMap.of("abc", "def"), ConfigurationPropertyValidator.parseMapProperty("abc=def"));
+    Assert.assertEquals(
+        ImmutableMap.of("abc", "def", "gh\\,i", "j\\\\\\,kl", "mno", "pqr"),
+        ConfigurationPropertyValidator.parseMapProperty("abc=def,gh\\,i=j\\\\\\,kl,mno=pqr"));
+    try {
+      ConfigurationPropertyValidator.parseMapProperty("not valid");
+      Assert.fail();
+    } catch (IllegalArgumentException ignored) {
+    }
```
coollog: nit: add comment `// pass`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -152,20 +174,22 @@ public void set(String image) {
...
   @Parameter private FromConfiguration from = new FromConfiguration();
 
-  @Parameter(property = "image")
-  private ToConfiguration to = new ToConfiguration();
+  @Parameter private ToConfiguration to = new ToConfiguration();
 
   @Parameter private ContainerParameters container = new ContainerParameters();
 
-  @Parameter(defaultValue = "false", required = true)
+  @Parameter(defaultValue = "false", required = true, property = "jib.useOnlyProjectCache")
```
coollog: These should probably be in `PropertyNames` too

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -187,76 +211,209 @@ MavenProject getProject() {
...
   @Nullable
   String getBaseImageCredentialHelperName() {
+    if (System.getProperty(PropertyNames.fromCredHelper) != null) {
+      return System.getProperty(PropertyNames.fromCredHelper);
+    }
     return Preconditions.checkNotNull(from).credHelper;
   }
 
   AuthConfiguration getBaseImageAuth() {
+    // System properties are handled in ConfigurationPropertyValidator
```
coollog: We should probably move the property names passed to `ConfigurationPropertyValidator#getImageCredential` to `PropertyNames`

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +137,58 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+        continue;
+      }
+
+      if (property.charAt(endIndex) == ',') {
+        // Split on non-escaped comma
+        items.add(property.substring(startIndex, endIndex));
+        startIndex = endIndex + 1;
+      } else if (property.charAt(endIndex) == '\\') {
+        // Found a backslash, ignore next character
+        shouldEscape = true;
```
chanseokoh: Probably we can just do `endIndex++` here and remove the `shouldEscape` variable?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +137,58 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+   *
+   * @param property the comma-separated string
+   * @return the list of parsed values
+   */
+  public static List<String> parseListProperty(String property) {
+    List<String> items = new ArrayList<>();
+    boolean shouldEscape = false;
+    int startIndex = 0;
+    int endIndex = 0;
+    for (; endIndex < property.length(); endIndex++) {
```
chanseokoh: `for (int endIndex = 0; ...` since it no longer has to be visible outside the `for` loop?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -128,5 +137,58 @@ public static ImageReference getGeneratedTargetDockerTag(
...
+    Map<String, String> result = new HashMap<>();
+
+    // Split on non-escaped commas
+    List<String> entries = parseListProperty(property);
+    for (String entry : entries) {
+      Matcher matcher = environmentPattern.matcher(entry);
+      if (!matcher.matches()) {
+        throw new IllegalArgumentException("'" + entry + "' is not a valid key-value pair");
+      }
+      result.put(matcher.group("name"), matcher.group("value"));
```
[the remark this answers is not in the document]
TadCordle: Added a test.

PR #1087, opened 2018-10-02
 Add system properties for gradle config

### LINKED ISSUE #1083: Add system properties for gradle command-line config
The Gradle equivalent for #728

TadCordle:
Fixes #1083.

Adds system properties to allow configuring gradle parameters via command-line. Same format as maven in #1080.


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PropertyNames.java
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.gradle;
+
+/** Names of system properties used to set configuration via commandline. */
+class PropertyNames {
```
coollog: These are the same as the Maven ones right? Let's put them in `plugins-common`?

coollog: Also, what do you think about using an enum for this?

TadCordle: I think they could go in `plugins-common`, although I thought we were keeping configuration naming out of `plugins-common` for some reason (maybe I'm misremembering a conversation we had a while ago, might have been something else).

I don't think we should use an enum here since this is specifically a collection of string constants, not a special data type.

PR #1092, opened 2018-10-03
Add FAQ for how to set up file system structure with copy-resources

### LINKED ISSUE #1085: Maybe add FAQ for how to set up file system structure with copy-resources
From https://github.com/GoogleContainerTools/jib/issues/1035#issuecomment-425931530

TadCordle:
Fixes #1085. Also cleans up a couple other outdated FAQ explanations.


docs/faq.md
```diff
@@ -202,6 +202,60 @@ jib.from.image = 'custom-base-image'
 
 We currently support adding a custom directory with an **incubating** feature. This feature may change in later versions. If your application needs to use custom files, place them into the `src/main/jib` folder. Files placed here will be added to the filesystem of the container. For example, `src/main/jib/foo/bar` would add `/foo/bar` into the container filesystem.
 
+### <a id="i-want-more-control-over-which-files-to-place-in-the-extra-directory"></a>I want more control over which files to place in the extra directory/where they end up in the container
```
chanseokoh: Is the anchor tag required for some reason?

Missing an ending period.

docs/faq.md
```diff
@@ -202,6 +202,60 @@ jib.from.image = 'custom-base-image'
...
+#### Maven
+
+In Maven, you can use the `maven-resources-plugin` to copy files to your extra directory. In your `pom.xml`:
+
+```xml
+<plugins>
+  ...
+  <plugin>
+    <artifact>maven-resources-plugin</artifact>
+    <version>3.1.0</version>
```
chanseokoh: Not the most recent version. Any reason?

docs/faq.md
```diff
@@ -165,6 +164,7 @@ Jib packages your Java application into the following paths on the image:
 * `/app/libs/` contains all the dependency artifacts
 * `/app/resources/` contains all the resource files
 * `/app/classes/` contains all the classes files
+* the contents of the extra directory (default `src/main/jib`) are placed in the container's root directory (`/`)
```
coollog: are placed *relative to* the container...?

docs/faq.md
coollog: Is the anchor here necessary? I think we should be fine to use a shorter bookmark link if GitHub generates a shorter one.

docs/faq.md
```diff
@@ -202,6 +202,60 @@ jib.from.image = 'custom-base-image'
...
+In Maven, you can use the `maven-resources-plugin` to copy files to your extra directory. In your `pom.xml`:
+
+```xml
+<plugins>
+  ...
+  <plugin>
+    <artifact>maven-resources-plugin</artifact>
+    <version>3.1.0</version>
+    <configuration>
+      <outputDirectory>${basedir}/src/main/jib/</outputDirectory>
```
chanseokoh: I think the output of a build should not go into the source repo. It means the files become source to be checked into a VCS.

docs/faq.md
```diff
@@ -202,6 +202,60 @@ jib.from.image = 'custom-base-image'
 
 We currently support adding a custom directory with an **incubating** feature. This feature may change in later versions. If your application needs to use custom files, place them into the `src/main/jib` folder. Files placed here will be added to the filesystem of the container. For example, `src/main/jib/foo/bar` would add `/foo/bar` into the container filesystem.
 
+### <a id="i-want-more-control-over-which-files-to-place-in-the-extra-directory"></a>I want more control over which files to place in the extra directory/where they end up in the container
+
+If the current extra directory design doesn't meet your needs (e.g. the extra files you want are generated outside the extra directory, and you want a different directory structure in the container), you can use additional goals/tasks to move the files to the configured extra directory between compile time and package time.
```
coollog: Maybe this can be more about `if you need to set up the extra files directory with files generated during build...`?

docs/faq.md
```diff
@@ -202,6 +202,60 @@ jib.from.image = 'custom-base-image'
...
+    </configuration>
+  </plugin>
+  ...
+</plugins>
+```
+
+Then run the goal between compiling and building the image to prepare the directory for building, either by setting the lifecycle phase, or by running the goal via command-line:
+
+```sh
+mvn compile resources:copy-resources jib:build
```
chanseokoh: AFAIK, the `generate-resources` and `process-resources` lifecycle run before `compile`. Is it that `resources:copy-resources` does not run as part of `generate-resources` or `process-resources` and we need to explicitly call `resources:copy-resources`?

docs/faq.md
```diff
@@ -202,6 +202,60 @@ jib.from.image = 'custom-base-image'
...
+```
+
+#### Gradle
+
+The same can be accomplished in Gradle by using a `Copy` task. In your `build.gradle`:
+
+```groovy
+task setupExtraDir(type: Copy) {
+  from file('files/you/want/in/extra/directory')
+  into file('src/main/jib/')
```
chanseokoh: ditto.

docs/faq.md
coollog: If you need to set up the extra files directory structure with files generated during the build process, you can create the extra files directory as part of your build:

docs/faq.md
```diff
@@ -148,11 +149,9 @@ See [Extended Usage](../jib-maven-plugin#extended-usage) for the `<container><fo
 
 See [Extended Usage](../jib-gradle-plugin#extended-usage) for the `container.format` configuration.
 
-### Can I define a custom entrypoint?
+### Can I define a custom entrypoint at runtime?
 
-The plugin attaches a default entrypoint that will run your application automatically.
-
-When running the image, you can override this default entrypoint with your own custom command.
+Normally, the plugin sets a default entrypoint for java applications, or lets you configure a custom entrypoint using the `container.entrypoint` configuration parameter. You can also override the default/configured entrypoint by defining a custom entrypoint when running the container.
```
chanseokoh: I feel like the last statement should be in the second paragraph? Or just have everything in a single paragraph.

docs/faq.md
coollog: I don't think we should have users be modifying source directories during the build process. The better practice is to construct a custom directory in the build output directory instead. (and then show how to configure the location of the extra files directory to point to that new directory)

docs/faq.md
TadCordle: I couldn't get the inferred link to work for some reason (maybe because of the '/' or title length?) so I just made it explicit.

docs/faq.md
TadCordle: It was what I saw [here](https://maven.apache.org/plugins/maven-resources-plugin/examples/copy-resources.html)

chanseokoh: OK, not a big deal. :)

docs/faq.md
TadCordle: Oh hmm yeah it looks like it runs automatically before compile.

docs/faq.md
TadCordle: https://github.com/GoogleContainerTools/jib/pull/1092#discussion_r222408669

chanseokoh: Try `directory-where` (i.e., use `-` instead of `/`).

chanseokoh: Sorry, seems like `directorywhere` (nothing between "directory" and "where").

docs/faq.md
```diff
@@ -202,6 +198,70 @@ jib.from.image = 'custom-base-image'
...
+```
+
+#### Gradle
+
+The same can be accomplished in Gradle by using a `Copy` task. In your `build.gradle`:
+
+```groovy
+jib.extraDirectory = file('build/extra-directory')
+
+task setupExtraDir(type: Copy) {
```
chanseokoh: Question: suppose the user wants to make sure this task runs after the normal build is complete. Is that achievable?

TadCordle: Probably, using `setupExtraDir.dependsOn`?

PR #1094, opened 2018-10-03
Set up integration test for Gradle WAR image building

chanseokoh:
Created `JibRunHelper` to share some code.

You'll notice `assertDockerInspect` did not migrate and excluded from `JibRunHelper`. That's why calls to it were added in some places in `JibPluginIntegrationTest`.

Will add more tests in later PRs.


jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/WarTest.java
```diff
@@ -0,0 +1,71 @@
...
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+/** Integration tests for building WAR images. */
+public class WarTest {
```
coollog: `WarIntegrationTest`? I think this should just belong in `JibPluginIntegrationTest` since the tests in there are for non-wars. Or, we can set a convention of naming like `JibPluginIntegrationTestWar.java` to indicate this class is just an extension onto integration tests for `JibPlugin`.

chanseokoh: I'm not really getting the part "integration tests for `JibPlugin`." The test looks more like an end-to-end test, actually running a Gradle build (albeit programmatically) with

```java
GradleRunner.create().withProjectDir(projectRoot.toFile()).withPluginClasspath()
```

and not like you directly instantiate `JibPlugin` or testing each method in `JibPlugin`.

BTW, I think `WarIntegrationTest` sounds good. I plan to add many tests, so it doesn't sound bad to classify tests with more classes. I think we don't need to put everything into a single class.

coollog: The `JibPlugin` class is the class that is directly applied to the Gradle build script and executed on the project build. The `JibPluginIntegrationTest` is testing the integration of `JibPlugin` with external dependencies (like GCR). I think if we break out `JibPluginIntegrationTest` into multiple classes, we should group them by the test project, so we'd have something like `SimpleProjectIntegrationTest`, `MultiProjectIntegrationTest`, `EmptyProjectIntegrationTest`, and `DefaultTargetProjectIntegrationTest`, and `WarProjectIntegrationTest`.

jib-gradle-plugin/src/integration-test/resources/projects/war_jetty_servlet25/build.gradle
```diff
@@ -12,18 +12,18 @@ repositories {
 }
 
 configurations {
-  moreLibs
+    moreLibs
```
coollog: nit: I think we should actually change all these test projects' build file indentation to 2 spaces for consistency with our other build files that use 2 spaces for indentation.

chanseokoh: Took this opportunity to fix other files too.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/WarTest.java
chanseokoh: OK, will try breaking up `JibPluginIntegrationTest`.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/WarIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+            + "/war_jetty_servlet25:gradle"
+            + System.nanoTime();
+
+    String containerName =
+        JibRunHelper.buildAndRun(jettyServlet25Project, targetImage, "--detach", "-p8080:8080");
+    try {
+      String content = getContent(new URL("http://localhost:8080/hello"));
+      Assert.assertEquals("Hello world", content);
+    } finally {
+      new Command("docker", "stop", containerName.trim()).run();
```
coollog: Doesn't have to be in this PR, but if we're adding many tests that may require this `try-finally`, we should prob use a separate closeable class that performs the container stop upon close so we can just use a try-with-resource. This separate class can also probably encapsulate the `getContent` waiting for ready logic to be shared with more tests.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/WarTest.java
coollog: Can probably leave that to a follow-up PR too.

chanseokoh: OK, I'm almost done with breaking it up, but I'll actually put that up in a later PR.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/WarIntegrationTest.java
chanseokoh: I have a follow-up PR with one more test, which refactors this. We can discuss this further in that PR.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/WarIntegrationTest.java
```diff
@@ -0,0 +1,71 @@
...
+        int code = connection.getResponseCode();
+        if (code == HttpURLConnection.HTTP_OK) {
+          ByteArrayOutputStream out = new ByteArrayOutputStream();
+          ByteStreams.copy(connection.getInputStream(), out);
+          return new String(out.toByteArray(), StandardCharsets.UTF_8);
+        }
+      } catch (IOException ex) {
+      }
+    }
+    return null;
```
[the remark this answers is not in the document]
chanseokoh: Since this is just a helper in a test, I'll go for the simplicity with `@Nullable`.

PR #1095, opened 2018-10-04
Upgrade / add Gradle wrappers

sschuberth:
Please have a look at the individual commit messages for the details.


coollog:
Hi @sschuberth thanks for your contribution! Could you share with us the commands you ran to generate these new wrappers? I'm currently seeing that the `.bat` wrappers have `LF` line endings rather than the `CRLF` required by Windows, which may cause the wrappers to not work on Windows.

sschuberth:
@coollog, interesting that you're indeed committing `.bat` files with Windows line endings in the repository. So far my belief was it's best Git practice to always commit Unix line endings, and add a `.gitattributes` file to eventually change how files appear in the working tree. I was just running the wrapper task, but my Git for Windows installation is configured to normalize to Unix line endings on commit. So, do you want me to commit with Windows line endings, or add a `.gitattributes` file?

coollog:
Hi @sschuberth , adding a `.gitattributes` file would be great - feel free to include `.bat` and `.cmd` files as `eol=crlf`.

sschuberth:
@coollog, I've added new commits to this PR accordingly.

coollog:
@sschuberth Thanks! It looks like the `gradle.properties` files got committed with CRLF line endings as well though.

jib-core/build.gradle
```diff
@@ -1,126 +1,122 @@
...
-  def googleChecksNoJava = googleChecks.replace(fileExtensionsBefore, fileExtensionsAfter)
-  assert !googleChecks.equals(googleChecksNoJava)
-
-  config = resources.text.fromString(googleChecksNoJava)
-
-  maxErrors = 0
-  maxWarnings = 0
-}
-/* CHECKSTYLE */
+plugins {
```
coollog: It looks like the `build.gradle` files got CRLF endings too.

sschuberth:
Doh, thanks again, fixed!

chanseokoh:
Can we merge this?

PR #1097, opened 2018-10-04
Add WAR integration test using Tomcat

chanseokoh:


chanseokoh:
The version Servlet 2.5 that is very old. However, good for testing a legacy web app, and it forces you to put some settings into `web.xml`.

I plan to add a Servlet 3.0 or 3.1 example too, which may not need `web.xml` at all.

PR #1106, opened 2018-10-05
Add WAR integration tests for exportDockerContext

chanseokoh:


jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/HttpGetVerifier.java
```diff
@@ -0,0 +1,38 @@
+package com.google.cloud.tools.jib.maven;
```
coollog: I'm a bit late here but copyright.

PR #1107, opened 2018-10-05
Include last modified time in cache selector digests instead of metadata

### LINKED ISSUE #1073: Consider having cache selector include last modified time as well.
Currently, the last modified time is stored as metadata in each cache entry. We could actually have that be a part of the data that generates the selector (along with the layer entries). This will save having to retrieve and check metadata content.

TadCordle:
Includes the source file modification time in the generated selector digests. As a result, we no longer need the cache metadata file, which ends up simplifying a lot of stuff.

Fixes #1073


jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -63,10 +64,11 @@ public void testLayerEntryTemplate_compareTo() {
   }
 
   @Test
-  public void testToSortedJsonTemplates() {
+  public void testToSortedJsonTemplates() throws IOException {
     Assert.assertEquals(
         toLayerEntryTemplates(IN_ORDER_LAYER_ENTRIES),
-        LayerEntriesSelector.toSortedJsonTemplates(OUT_OF_ORDER_LAYER_ENTRIES));
+        ImmutableList.copyOf(
```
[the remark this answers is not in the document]
TadCordle: iirc this test was failing because the `ArrayList` returned by `LayerEntriesSelector#toSortedJsonTemplates` wasn't equal to the `ImmutableList` returned by `toLayerEntryTemplates`, even though the contents were the same.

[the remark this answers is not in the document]
chanseokoh: Maybe we should just make `toLayerEntryTemplates` return `List`. Returning `ImmutableList` is too limiting without much value in tests.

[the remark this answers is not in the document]
coollog: Hmm, that's weird - `ArrayList` should be equal to `ImmutableList` if contents are the same.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
```diff
@@ -82,24 +96,26 @@ public boolean equals(Object other) {
...
       return sourceFile.equals(otherLayerEntryTemplate.sourceFile)
-          && extractionPath.equals(otherLayerEntryTemplate.extractionPath);
+          && extractionPath.equals(otherLayerEntryTemplate.extractionPath)
+          && lastModifiedTime.equals(otherLayerEntryTemplate.lastModifiedTime);
     }
 
     @Override
     public int hashCode() {
-      return Objects.hash(sourceFile, extractionPath);
+      return Objects.hash(sourceFile, extractionPath, extractionPath);
```
briandealwis: Why is `extractionPath`  here twice?  Should the second be `lastModifiedTime`?

jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -63,6 +67,29 @@
         .collect(ImmutableList.toImmutableList());
   }
 
+  @Test
+  public void testLayerEntryTemplate_fileModified() throws IOException {
+    TemporaryFolder temporaryFolder = new TemporaryFolder();
```
coollog: I would recommend using `@Rule` since it would run the cleanup of the created folder automatically.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -63,6 +67,29 @@
...
+    temporaryFolder.create();
+    Path layerFile = temporaryFolder.newFolder("testFolder").toPath().resolve("file");
+    LayerEntry layerEntry = new LayerEntry(layerFile, AbsoluteUnixPath.get("/extraction/path"));
+
+    LayerEntryTemplate layerEntryTemplate = new LayerEntryTemplate(layerEntry);
+    DescriptorDigest expectedSelector =
```
coollog: This seems to be testing `LayerEntryTemplate` and duplicating the logic in `LayerEntriesSelector`. Should probably test `LayerEntriesSelector#generateSelector` instead.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -63,6 +67,29 @@
         .collect(ImmutableList.toImmutableList());
   }
 
+  @Test
+  public void testLayerEntryTemplate_fileModified() throws IOException {
+    TemporaryFolder temporaryFolder = new TemporaryFolder();
+    temporaryFolder.create();
+    Path layerFile = temporaryFolder.newFolder("testFolder").toPath().resolve("file");
```
coollog: This file should probably exist - we wouldn't ever be adding a layer entry for a file that doesn't exist.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
TadCordle: oops

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
```diff
@@ -56,11 +60,16 @@
...
     private final String extractionPath;
+    private final Instant lastModifiedTime;
 
     @VisibleForTesting
-    LayerEntryTemplate(LayerEntry layerEntry) {
+    LayerEntryTemplate(LayerEntry layerEntry) throws IOException {
       sourceFile = layerEntry.getAbsoluteSourceFileString();
       extractionPath = layerEntry.getAbsoluteExtractionPathString();
+      lastModifiedTime =
+          Files.exists(layerEntry.getSourceFile())
```
coollog: Hmm, I don't think any of `LayerEntry` files should not exist. This should probably just let `Files.getLastModifiedTime` throw its exception.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -52,7 +51,15 @@
       ImmutableList<LayerEntry> layerEntries) {
     return layerEntries
         .stream()
-        .map(layerEntry -> new LayerEntryTemplate(layerEntry, Instant.EPOCH.toString()))
+        .map(
+            layerEntry -> {
+              try {
```
coollog: For this kind of method, should probably just use an iterative approach over stream to avoid the try-catch here.

TadCordle:
I think we're covering a lot of cases already, but if you can think of more specific edge cases we should test for let me know.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -82,10 +93,25 @@ public void testGenerateSelector_empty() throws IOException {
...
+  @Test
+  public void testGenerateSelector_fileModified() throws IOException, InterruptedException {
+    Path layerFile = temporaryFolder.newFolder("testFolder").toPath().resolve("file");
+    Files.write(layerFile, "hello".getBytes(StandardCharsets.UTF_8));
+    LayerEntry layerEntry = new LayerEntry(layerFile, AbsoluteUnixPath.get("/extraction/path"));
+    DescriptorDigest expectedSelector =
+        LayerEntriesSelector.generateSelector(ImmutableList.of(layerEntry));
+
+    // Verify that changing modified time generates a different selector
+    Thread.sleep(1000);
```
coollog: `Files.setLastModified` to set the last modified time explicitly.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -82,10 +93,25 @@ public void testGenerateSelector_empty() throws IOException {
...
+    Files.write(layerFile, "goodbye".getBytes(StandardCharsets.UTF_8));
+    Assert.assertNotEquals(
+        expectedSelector, LayerEntriesSelector.generateSelector(ImmutableList.of(layerEntry)));
```
coollog: Could also just check that setting last modified back to the original generates the same selector.

jib-maven-plugin/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- Last modified time of layer entries is now stored in selectors instead of metadata file in cache ([#1073](https://github.com/GoogleContainerTools/jib/issues/1073))
```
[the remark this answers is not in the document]
coollog: Yea also this feature wasn't present in the previous version so nothing really changed yet haha.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
```diff
@@ -82,24 +93,27 @@ public boolean equals(Object other) {
...
    *
-   * @param layerEntries
+   * @param layerEntries the list of {@link LayerEntry} to convert
    * @return list of {@link LayerEntryTemplate} after sorting
+   * @throws IOException if checking the file creation time of a layer entry fails
    */
   @VisibleForTesting
-  static List<LayerEntryTemplate> toSortedJsonTemplates(List<LayerEntry> layerEntries) {
+  static List<LayerEntryTemplate> toSortedJsonTemplates(List<LayerEntry> layerEntries)
+      throws IOException {
```
coollog: stray exception?

TadCordle: Nope, `LayerEntryTemplate`'s constructor can throw `IOException` when getting last modified time.

Issue #1029, opened 2018-09-24
Create images with a built-in non-root user


cjvogel1972:
**Description of the issue**:
Best practices for Docker containers is to run them as a non-root user.  As Jib works now, there is no way to specify a built-in non-root user.  Google distroless gives an example of how to do this using [Bazel](https://github.com/GoogleContainerTools/distroless/tree/master/examples/nonroot). 

I saw your answer for issue #631, but that I don't think that limits the security risk.  In Docker, you have to remember to add the --user in the run command.  You Kubernetes security context answer helps, but it still requires you to remember to set the configuration.

I just think it would be good if the tool gives the option to build the user ID into the image.  Just as an example, where I work, neither solution would work because we have security software that inspects the image and prevents its use if there is not a user ID built into the image.  I would love to use Jib at work, but until the user ID is baked into the image, it won't be possible.

**Expected behavior**:
It would be good to add to the jib/to closure an option for specifying a user and group to run the container as.

coollog:
Hi @cjvogel1972 , thanks for suggesting this and providing the detailed explanation! We'll look into adding this option @GoogleContainerTools/java-tools 

We will probably add a `<container><user>`/`jib.container.user` parameter that will accept the user/group in the forms Dockerfiles accept currently: `user`, `uid`, `user:group`, `uid:gid`, `uid:group`, `user:gid`

patflynn:
@mattmoor any thoughts on this?

chanseokoh:
I suggest initially we support only numbers, unlike the Dockerfile spec. Almost all the time, supporting names would require modifying `/etc/password` and `/etc/group` for Unix images, and this also raises the question of which number Jib should assign to. For example,

```Dockerfile
FROM ubuntu
USER chanseok
CMD ls
```

```
$ docker build -t user-test .
$ docker run --rm user-test
docker: Error response from daemon: linux spec user: unable to find user chanseok: no matching entries in passwd file.
```

One way to make this work with `Dockerfile` is
```Dockerfile
FROM ubuntu
RUN groupadd --non-unique --gid 23456 awesomegroup
RUN useradd --non-unique --system --uid 12345 --gid 23456 chanseok
USER chanseok
ENTRYPOINT ["cat", "/etc/group", "/etc/passwd"]
```

cjvogel1972:
[...]

PR #1109, opened 2018-10-05
Adds container.user configuration

CyrilleH:
Towards #1029, this PR adds the possibility to sets the username or UID which the process in the container should run as.

For security reason, we prefer to start a tomcat container with non-root user. We have a generic base image with this (in our private registry) : 
```
RUN addgroup -S tomcat && adduser -S -G tomcat tomcat && chown -R tomcat:tomcat /usr/local
```


coollog:
Thanks for the contribution! We'll take a look at this soon. For the future, we'd recommend commenting on the issue thread with the design/solution you will be implementing before starting work to help save time potentially changing large portions of the implementation.

CyrilleH:
Ok I understand. It was very easy to implement.
If you want I can close this PR.

coollog:
We can definitely work with most of this :) It's mostly that we need to decide what we would support for the user configuration. Currently, as proposed in https://github.com/GoogleContainerTools/jib/issues/1029#issuecomment-426405346, we are intending only to support user id and group id and not username, but we might decide to allow username as well. @GoogleContainerTools/java-tools

chanseokoh:
Yeah, looks like we can work on this. There are a few things to think about though.

And thinking about it for a while, it might be possible to accept any string as the parameter value, not just numbers, if what we end up doing is to just set the value in the container config metadata. In that case, we may always be able to build an image (needs to check), which may error at runtime due to non-existing username. However, on the other hand, this may result in a `Dockerfile`  that is not buildable.

chanseokoh:
> However, on the other hand, this may result in a `Dockerfile`  that is not buildable.

Never mind. I think it is always buildable. Now I think there is no problem supporting usernames and it is the user's responsibility to use the base image that has the right user database. Same for usual Docker build.

CyrilleH:
If we set the wrong username, we have this error at runtime : 
```
docker: Error response from daemon: linux spec user: unable to find user tomcat2: no matching entries in passwd file.
ERRO[0000] error waiting for container: context canceled 
```

chanseokoh:
> If we set the wrong username, we have this error at runtime

Yeah, it's the same story for `Dockerfile`, so perhaps we should say it's WAI. The big difference is, however, you can easily do `adduser` or `addgroup` with `Dockerfile` to make it work. But with Jib, it's a bit cumbersome in that you need to prepare a base image with the user or have `/etc/passwd` and `/etc/group` under `/src/main/jib`.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -136,14 +137,25 @@ public Builder setEntrypoint(@Nullable List<String> entrypoint) {
       return this;
     }
 
+    /**
+     * Sets the username or UID which the process in the container should run as.
```
chanseokoh: I believe this can be a mix of user and group.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -299,6 +311,7 @@ public void generate(Path targetDirectory) throws IOException {
    * LABEL [key1]="[value1]" \
    *     [key2]="[value2]" \
    *     [...]
+   * USER [user or UID]
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -89,6 +90,17 @@
       return this;
     }
 
+    /**
+     * Sets the username or UID which the process in the container should run as.
+     *
+     * @param user the username or UID
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -191,6 +192,17 @@ public JavaDockerContextGenerator setEntrypoint(List<String> entrypoint) {
     return this;
   }
 
+  /**
+   * Sets the username or UID which the process in the container should run as.
+   *
+   * @param user the username or UID.
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -141,6 +141,9 @@ public Blob getContainerConfigurationBlob() {
     // Sets the working directory.
     template.setContainerWorkingDir(image.getWorkingDirectory());
 
+    // Sets the user.
+    template.setContainerUser(image.getUser());
```
chanseokoh: Should there be a counterpart for `JsonToImageTranslator`?

CyrilleH: Yes, for future use

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
CyrilleH: You're right, I changed the javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
CyrilleH: Javadoc modified.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -136,14 +137,26 @@ public Builder setEntrypoint(@Nullable List<String> entrypoint) {
       return this;
     }
 
+    /**
+     * Sets the user name (or UID) and optionally the user group (or GID) which the process in the
```
coollog: Wording suggestion: `Sets the user and group to run the container as. {@code user} can be a username or UID along with an optional groupname or GID. The following are all valid: {@code user}, {@code uid}, {@code user:group}, {@code uid:gid}, {@code uid:group}, {@code user:gid}.`

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -191,6 +192,18 @@ public JavaDockerContextGenerator setEntrypoint(List<String> entrypoint) {
     return this;
   }
 
+  /**
+   * Sets the user name (or UID) and optionally the user group (or GID) which the process in the
```
coollog: Wording suggestion: `Sets the user for the {@code USER} directive.`

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -89,6 +90,18 @@
       return this;
     }
 
+    /**
+     * Sets the user name (or UID) and optionally the user group (or GID) which the process in the
```
coollog: Wording suggestion: `Sets the user/group to run the container as.`

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -89,6 +90,18 @@
...
+     * container should run as.
+     *
+     * @param user the username and optionally the user group
```
coollog: `the username/UID and optionally the groupname/GID`

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -136,14 +137,26 @@ public Builder setEntrypoint(@Nullable List<String> entrypoint) {
...
+     * container should run as.
+     *
+     * @param user the username and optionally the user group
```
coollog: `the username/UID and optionally the groupname/GID`

chanseokoh:
Looks good to me. Updating some wordings, and I think we're good to go.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
CyrilleH: done

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
CyrilleH: done

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
CyrilleH: done

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
CyrilleH: done

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
CyrilleH: done

CyrilleH:
Great. I've rebased on master and fixed wordings.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/DockerContextTaskTest.java
```diff
@@ -95,6 +96,25 @@ public void testEntrypoint_defaultWebAppRoot() throws IOException {
...
+
+  @Test
+  public void testUser_null() throws IOException {
+    Mockito.when(containerParameters.getUser()).thenReturn(null);
+    task.generateDockerContext();
+    try {
+      getUser();
+      Assert.fail();
+    } catch (NoSuchElementException ex) {
+    }
```
coollog: nit: just add a `// pass` comment to the empty block here

CyrilleH: done

Issue #1118, opened 2018-10-09
Use Jib Core in jib-gradle-plugin and jib-maven-plugin.


coollog:
- [ ] Use Jib Core API in `jib-gradle-plugin`
- [ ] Use Jib Core API in `jib-maven-plugin`

I will work on the `jib-gradle-plugin` side and it'd be great if someone else takes the `jib-maven-plugin` side to get a feel of the new library.

TadCordle:
I could take maven

loosebazooka:
Assign yourself or edit the first comment to indicate ownership.

coollog:
Actually, I was working on the Gradle side and there might be some overlap in changes that might affect Maven side (changes in `plugins-common`). It might be better to wait until I push the Gradle changes and work on top of those changes for the Maven side - unless you've gotten far in it already?

TadCordle:
Nope, haven't started yet; still working on the cache stuff. I can wait until you've made progress.

coollog:
@TadCordle I put up #1119, so working off of that should be good. The main thing to note is to switch maven to use `NBuildStepsRunner`, which will replace `BuildStepsRunner` once everything is changed over.

PR #1119, opened 2018-10-09
 Changes jib-gradle-plugin to use new Jib Core API.

coollog:
For #1118

This adds `NBuildStepsRunner` that should replace `BuildStepsRunner` once `jib-maven-plugin` is changed to use Jib Core as well.


coollog:
Merged master.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/api/JibContainerBuilderTestHelper.java
```diff
@@ -0,0 +1,35 @@
+/*
```
TadCordle: Should probably move this to core or plugins-common, I think we'll need it for Maven as well.

TadCordle: Or possibly use an alternative design that doesn't involve wrapping a single method, but I'm not sure the best way of doing that.

coollog: Will move to `plugins-common`. This was mostly necessary because didn't want to put an `@VisibleForTesting` method in the user-exposed `JibContainerBuilder` class.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/NBuildStepsRunner.java
```diff
@@ -0,0 +1,265 @@
...
+
+        eventDispatcher.dispatch(
+            LogEvent.info("\t" + capitalizeFirstLetter(layerConfiguration.getName()) + ":"));
+
+        for (LayerEntry layerEntry : layerConfiguration.getLayerEntries()) {
+          eventDispatcher.dispatch(LogEvent.info("\t\t" + layerEntry.getSourceFile()));
+        }
+      }
+
+      JibContainer jibContainer = jibContainerBuilder.containerize(containerizer);
```
TadCordle: Unused, should we be returning this?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -105,37 +111,35 @@ public void buildImage()
...
-    BuildConfiguration buildConfiguration =
-        pluginConfigurationProcessor
-            .getBuildConfigurationBuilder()
-            .setBaseImageConfiguration(
-                pluginConfigurationProcessor.getBaseImageConfigurationBuilder().build())
-            .setTargetImageConfiguration(targetImageConfiguration)
-            .setAdditionalTargetImageTags(jibExtension.getTo().getTags())
-            .setContainerConfiguration(
-                pluginConfigurationProcessor.getContainerConfigurationBuilder().build())
-            .setTargetFormat(jibExtension.getContainer().getFormat())
```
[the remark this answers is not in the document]
coollog: Ah right I'll change that back.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessorTest.java
```diff
@@ -46,65 +62,82 @@
   @Before
   public void setUp() {
     Mockito.doReturn("gcr.io/distroless/java").when(mockBaseImageParameters).getImage();
```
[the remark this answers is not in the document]
coollog: Yea I'll go and change over everything in this test.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/NBuildStepsRunner.java
coollog: Yes, will make this method return that.

coollog:
Gonna wait for one more approval since this touches a large portion of the critical code path. @GoogleContainerTools/java-tools

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
```diff
@@ -122,45 +128,44 @@ static PluginConfigurationProcessor processCommonConfiguration(
...
-                projectProperties.getJavaLayerConfigurations().getLayerConfigurations());
-    buildConfigurationBuilder.setApplicationLayersCacheDirectory(
-        projectProperties.getCacheDirectory());
-    buildConfigurationBuilder.setBaseImageLayersCacheDirectory(
-        Containerizer.DEFAULT_BASE_CACHE_DIRECTORY);
-    if (jibExtension.getUseOnlyProjectCache()) {
-      buildConfigurationBuilder.setBaseImageLayersCacheDirectory(
-          projectProperties.getCacheDirectory());
-    }
+    Consumer<Containerizer> containerizerConsumer =
```
chanseokoh: This seems like the only place where you pass a `containerizerConsumer` to the `PluginConfigurationProcessor`. Do we need to parameterize the consumer? (That is, can't we skip passing a consumer and let `PluginConfigurationProcess.configureContainerizer()` just do the setup there?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/NBuildStepsRunner.java
```diff
@@ -0,0 +1,270 @@
...
+
+  private static final String STARTUP_MESSAGE_PREFIX_FOR_DOCKER_DAEMON =
+      "Containerizing application to Docker daemon as ";
+  private static final String SUCCESS_MESSAGE_PREFIX_FOR_DOCKER_DAEMON =
+      "Built image to Docker daemon as ";
+
+  private static final String STARTUP_MESSAGE_FORMAT_FOR_TARBALL =
+      "Containerizing application to file at '%s'...";
+  private static final String SUCCESS_MESSAGE_FORMAT_FOR_TARBALL =
+      "Built image tarball at \u001B[36m%s\u001B[0m";
```
chanseokoh: Just asking: I feel like I've seen this class or most part of the code. Is this class just copied or a completely new one? If copied, is it possible to remove the original one (basically moving the class)?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PluginConfigurationProcessor.java
coollog: Yes, I'll move it to a static `configureContainerizer` method. I think the `PluginConfigurationProcessor` will need to be restructured at some point since the logic is a bit unwieldy and interdependent on the plugin goal/task logic.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/NBuildStepsRunner.java
coollog: Yes it is copied. #1124 will remove the original.

PR #1127, opened 2018-10-10
Add example of downloading and configuring the Stackdriver Debugger Java Agent

### LINKED ISSUE #135: Add example for running built image with Java agent in README.

briandealwis:
Fixes #135


examples/README.md
```diff
@@ -23,3 +23,7 @@ See [spring-boot-kubernetes](spring-boot-kubernetes) for containerizing a [Sprin
 ### Micronaut example
 
 See [micronaut](micronaut) for containerizing a [Micronaut framework](https://micronaut.io/) Groovy/Java application.
+
+### Java Agents example
```
coollog: tiny nit: lowercase `agents`

examples/README.md
```diff
@@ -23,3 +23,7 @@ See [spring-boot-kubernetes](spring-boot-kubernetes) for containerizing a [Sprin
...
+
+See [java-agent](java-agent) for launching with a Java agent.
```
coollog: Might be good to include the specific one used:

`...launching with a Java agent (Stackdriver Debugger).`

examples/java-agent/build.gradle
```diff
@@ -0,0 +1,54 @@
+plugins {
+  id 'java'
+  id 'com.google.cloud.tools.jib' version '0.9.11'
```
coollog: Oh let's update all these gradle versions to `0.9.12`. Can be a separate PR though.

examples/java-agent/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+}
+
+dependencies {
+  compile 'com.google.guava:guava:23.6-jre'
+}
+
+// Download and extract the Cloud Debugger Java Agent
+task downloadAgent(type: Download) {
+  src stackdriverDebuggerAgentUrl
+  dest new File(buildDir, 'cdbg_java_agent_gce.tar.gz')
```
coollog: s/`new File`/`file`/

examples/java-agent/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+        '-Dcom.google.cdbg.module=' + project.name,
+        '-Dcom.google.cdbg.version=' + version]
+        // if not running on Compute Engine, must download and copy a service key
+        // (https://cloud.google.com/debugger/docs/setup/java#local)
+        // -Dcom.google.cdbg.auth.serviceaccount.enable=true
+        // -Dcom.google.cdbg.auth.serviceaccount.jsonfile=/opt/cdbg/gcp-svc.json
+  }
+}
+
+defaultTasks 'extractAgent'
```
coollog: I think it might be better practice to just have `jib` depend on `extractAgent`?

examples/java-agent/README.md
```diff
@@ -0,0 +1,14 @@
+Builds a container image that includes the [Stackdriver Debugger Java Agent](https://cloud.google.com/debugger/docs/).
+
+To build the image:
```
coollog: Let's add the Gradle instructions too.

examples/java-agent/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+1. In `pom.xml`, replace `REPLACE-WITH-YOUR-GCP-PROJECT` with your GCP project.
+
+1. If running outside of Google Cloud Platform, you must:
+     1. [Create and download service account credentials](https://cloud.google.com/debugger/docs/setup/java#local).
+     2. Copy the credentials file into the jib `extraDirectory` location so that it is copied into the container.
```
chanseokoh: I think the service account credential part deserves some more attention:

1. The service account should be kept safely. It's easy to put an image in a registry with public access, then anyone can download and use it.
1. I see `extraDirectory` is set to the `agent` directory in a build output in this example. So with this setup, perhaps users need to do `mvn compile` first, copy the JSON key to `${build}/agent/opt/cdbg/gcp-svc.json` (the `agent` part already feels weird, maybe just another name), and then do `mvn package`? Also note `mvn clean` will clear up the JSON.

examples/java-agent/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <jib-maven-plugin.version>0.9.11</jib-maven-plugin.version>
+    <maven-compiler-plugin.version>3.8.0</maven-compiler-plugin.version>
+    <download-maven-plugin.version>1.4.1</download-maven-plugin.version>
+
+    <!-- agents will be extracted relative to this build location -->
+    <agent-extraction-root>${project.build.directory}/agents</agent-extraction-root>
+    
+    <!-- location of Stackdriver Debugger Java Agent https://cloud.google.com/debugger/docs/setup/java -->
+    <stackdriver-debugger-agent-url>https://storage.googleapis.com/cloud-debugger/compute-java/debian-wheezy/cdbg_java_agent_gce.tar.gz</stackdriver-debugger-agent-url>
```
chanseokoh: I wonder if it makes sense to pin down to a specific version or just embed the binary for build reproducibility.

examples/java-agent/README.md
saturnism: +100.. a service account should _never_ be included in the container image.
if the cluster was created with the proper machine credential/scope, the service account is not needed.

examples/java-agent/build.gradle
```diff
@@ -0,0 +1,54 @@
+plugins {
+  id 'java'
+  id 'com.google.cloud.tools.jib' version '0.9.11'
+  id 'de.undercouch.download' version '3.4.0'
+}
+
+ext {
+  // where to download the Stackdriver Debugger agent https://cloud.google.com/debugger/docs/setup/java
+  stackdriverDebuggerAgentUrl = 'https://storage.googleapis.com/cloud-debugger/compute-java/debian-wheezy/cdbg_java_agent_gce.tar.gz'
```
saturnism: if we use gce agent, we don't need to generate the service account at all. just make sure the cluster has the proper scope.

examples/java-agent/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+            <image>gcr.io/REPLACE-WITH-YOUR-GCP-PROJECT/image-built-with-jib</image>
+          </to>
+          <extraDirectory>${agent-extraction-root}</extraDirectory>
+          <container>
+            <jvmFlags>
+              <jvmFlag>-agentpath:${stackdriver-debugger-agent-location}/cdbg_java_agent.so</jvmFlag>
+              <jvmFlag>-Dcom.google.cdbg.module=${project.artifactId}</jvmFlag>
+              <jvmFlag>-Dcom.google.cdbg.version=${project.version}</jvmFlag>
+              <!--
+                 if not running on Compute Engine, must download and copy a service key
```
saturnism: let's not encourage users copying service keys into a container image. use secrets instead.

examples/java-agent/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>23.6-jre</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
```
saturnism: add the git commit plugin if using a debugger agent.. it'll generate a git.properties file that has the metadata necessary for debugger agent to get the source.

examples/java-agent/README.md
briandealwis: I think it's fair to say that properly obtaining and managing the service account credential is outside the scope of this example (and hence why it doesn't do it :-).  The text above does say: _If running outside of Google Cloud Platform_.  I'm happy to take suggestions for improvements.

examples/java-agent/pom.xml
briandealwis: They don't do versioned releases, from what I could find.

examples/java-agent/pom.xml
briandealwis: Suggestions welcome!

examples/java-agent/build.gradle
briandealwis: How do you do this?  My gradle-fu isn't well-developed.

examples/java-agent/README.md
briandealwis: I'll remove this bit and link to the Stackdriver Debugger page.

examples/java-agent/src/main/java/example/HelloWorld.java
```diff
@@ -0,0 +1,36 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class HelloWorld {
+
+  public static void main(String[] args) throws URISyntaxException, IOException {
+    Path worldFile = Paths.get(Resources.getResource("world").toURI());
+    String world = new String(Files.readAllBytes(worldFile), StandardCharsets.UTF_8);
```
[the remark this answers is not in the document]
briandealwis: ok, I've set up as a little SparkJava server and enabled cdbg logging to stderr.

briandealwis:
Stackdriver Debugger connects to a Google service so it doesn't need any ports to be exposed.

examples/java-agent/pom.xml
briandealwis: I've dropped this text and will rely on the pointers in README.md instead.

examples/java-agent/build.gradle
coollog: I believe here you would do `tasks.jib.dependsOn extractAgent`

examples/java-agent/README.md
```diff
@@ -0,0 +1,14 @@
+This SparkJava-based example builds a container image that includes the [Stackdriver Debugger Java Agent](https://cloud.google.com/debugger/docs/).
+This project assumes the resulting image will be run inside Google CLoud Platform. To run outside of
+GCP, follow the Stackdrive Debugger instructions to [create and download service account
+credentials](https://cloud.google.com/debugger/docs/setup/java#local).
```
chanseokoh: This change definitely seems better, but after hearing what @saturnism said, I think we should get rid of anything that can make people even remotely think about putting a credential into a container. Such idea seems really dangerous. The Stackdriver debugger instructions are under "Local and Elsewhere", and I'm not sure if the "Elsewhere" part includes running containers on prod outside GCP, but now I'd like to avoid making the practice of embedding a key in a container look like a viable solution, even remotely. Maybe we can append at the end of the paragraph "and inject the credentials in the secure way" or "never embed a key into the container, since it becomes easy to make it accessible from others" or "the credentials are secret that should be kept secure".

briandealwis: I'll remove that text, and will submit feedback on the Stackdriver Debugger doc page.

examples/java-agent/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+	compile "com.sparkjava:spark-core:2.7.2"
+	compile "org.slf4j:slf4j-simple:1.7.21"
```
chanseokoh: nit: indentation

examples/java-agent/build.gradle
chanseokoh: `0.9.12` sounds good.

briandealwis: #1139

PR #1129, opened 2018-10-11
Updates jib-core release script to accept version suffixes.

coollog:
This is modeled after https://github.com/GoogleCloudPlatform/app-gradle-plugin/blob/master/scripts/prepare_release.sh

~~Planning to do a `v0.1.0-rc1` release of Jib Core first.~~
We'll actually just do a `v0.1.0` release first. This change can still go in though.


jib-core/scripts/prepare_release.sh
```diff
@@ -4,39 +4,43 @@
...
 }
 
 # Usage: CheckVersion <version>
 CheckVersion() {
-    [[ $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || Die "Version not in ###.###.### format."
+  [[ $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z]+)?$ ]] || Die "Version: $1 not in ###.###.###[-XXX] format."
 }
 
-[ $# -ne 2 ] || DieUsage
+[ $# -ne 2 ] || [ $# -ne 3 ] || DieUsage
```
chanseokoh: Is this right? I think the original code was also wrong, if the intention was to die if the number of args isn't 2. It should have been `[ $# -ne 2] && DieUsage` or `[ $# -eq 2 ] || DieUsage`.

jib-core/scripts/prepare_release.sh
```diff
@@ -4,39 +4,43 @@
 set -e
 
 Colorize() {
```
chanseokoh: Is Colorize used anywhere?

jib-core/scripts/prepare_release.sh
```diff
@@ -4,39 +4,43 @@
...
 }
 
-[ $# -ne 2 ] || DieUsage
+[ $# -ne 2 ] || [ $# -ne 3 ] || DieUsage
 
 EchoGreen '===== RELEASE SETUP SCRIPT ====='
 
 VERSION=$1
 CheckVersion ${VERSION}
+if [ $2 ]; then
```
chanseokoh: It's good to wrap it in quotes `"$2"`, since otherwise if `$2` is undefined, the `if` syntax becomes broken as `if [  ]; then`.  Maybe `if [ -n "$2" ]` (testing if length is non-zero).

jib-core/scripts/prepare_release.sh
coollog: Nope, will remove.

jib-core/scripts/prepare_release.sh
coollog: Ah yes this logic is just wrong.

chanseokoh:
Probably fix https://github.com/GoogleCloudPlatform/app-gradle-plugin/blob/master/scripts/prepare_release.sh too?

loosebazooka:
classic bash :\

PR #1132, opened 2018-10-11
Allow using path or suffix in cred helper config

### LINKED ISSUE #925: [Gradle] Ability to specify absolute path of credential helper
**Description of the issue**:

Currently it is only possible to specify the credential helper suffix (e.g., gcr). This requires an executable `docker-credential-gcr` to be on the `PATH`. This manipulation of the `PATH` may not be ideal, i.e., to allow builds to be reproducible on multiple environments.

**Expected behavior**:
It is possible to specify the absolute path to a credential helper. I believe`jib-core` already supports passing a `java.nio.Path` and it just needs to be exposed.

TadCordle:
Fixes #925


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -112,10 +114,15 @@ public DefaultCredentialRetrievers setCredentialHelperSuffix(
...
       credentialRetrievers.add(knownCredentialRetriever);
     }
-    if (credentialHelperSuffix != null) {
-      credentialRetrievers.add(
-          credentialRetrieverFactory.dockerCredentialHelper(
-              DockerCredentialHelper.CREDENTIAL_HELPER_PREFIX + credentialHelperSuffix));
+    if (credentialHelper != null) {
+      if (Files.exists(Paths.get(credentialHelper))) {
+        credentialRetrievers.add(
+            credentialRetrieverFactory.dockerCredentialHelper(Paths.get(credentialHelper)));
```
chanseokoh: hit: there is a `String`-version, so you don't have to create a `Path` here.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrieversTest.java
```diff
@@ -91,4 +100,21 @@ public void testInitAsList_all() {
     Mockito.verify(mockCredentialRetrieverFactory)
         .dockerCredentialHelper("docker-credential-credentialHelperSuffix");
   }
+
+  @Test
+  public void testInitAsList_credentialHelperPath() throws IOException {
```
chanseokoh: Is it worth adding a test when the file does not exist?

TadCordle: `testInitAsList_all` covers this, I think.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -90,15 +92,15 @@ public DefaultCredentialRetrievers setInferredCredential(
...
-   * @param credentialHelperSuffix the known credential helper suffix (following {@code
-   *     docker-credential-})
+   * @param credentialHelper the path to a credential helper, or a credential helper suffix
+   *     (following {@code docker-credential-}) if the path doesn't exist.
    * @return this
    */
-  public DefaultCredentialRetrievers setCredentialHelperSuffix(
-      @Nullable String credentialHelperSuffix) {
-    this.credentialHelperSuffix = credentialHelperSuffix;
+  public DefaultCredentialRetrievers setCredentialHelper(@Nullable String credentialHelper) {
```
coollog: I think we should have two methods, one for setting a suffix and one for specifying a path rather than overloading this single `credentialHelper` parameter.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -112,10 +114,15 @@ public DefaultCredentialRetrievers setCredentialHelperSuffix(
     if (knownCredentialRetriever != null) {
       credentialRetrievers.add(knownCredentialRetriever);
     }
-    if (credentialHelperSuffix != null) {
-      credentialRetrievers.add(
-          credentialRetrieverFactory.dockerCredentialHelper(
-              DockerCredentialHelper.CREDENTIAL_HELPER_PREFIX + credentialHelperSuffix));
+    if (credentialHelper != null) {
+      if (Files.exists(Paths.get(credentialHelper))) {
```
coollog: I think we might want to be careful about a credential helper suffix matching a file in the cwd. For example, if there's a file called `gcr` in the directory, this would try to use that as the credential helper executable. I think we should limit this to only check as a path if `credentialHelper` contains at least one file separator.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
coollog: I think it's better to keep the logic for parsing user input (into the separate credential helper suffix vs. credential helper path) as close to the input as possible, rather than hiding the logic within this class. I think we will need to refactor more of the common plugin code later, so this can be here for now.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -90,15 +93,15 @@ public DefaultCredentialRetrievers setInferredCredential(
...
 
   /**
-   * Sets the suffix for a known credential helper.
+   * Sets the known credential helper. May either be a path to a credential helper executable, or a
+   * credential helper suffix (following {@code docker-credential-}).
    *
-   * @param credentialHelperSuffix the known credential helper suffix (following {@code
-   *     docker-credential-})
+   * @param credentialHelper the path to a credential helper, or a credential helper suffix
+   *     (following {@code docker-credential-}) if the path doesn't exist.
```
coollog: I think the `if the path doesn't exist` can be removed?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -112,10 +115,16 @@ public DefaultCredentialRetrievers setCredentialHelperSuffix(
     if (knownCredentialRetriever != null) {
       credentialRetrievers.add(knownCredentialRetriever);
     }
-    if (credentialHelperSuffix != null) {
-      credentialRetrievers.add(
-          credentialRetrieverFactory.dockerCredentialHelper(
-              DockerCredentialHelper.CREDENTIAL_HELPER_PREFIX + credentialHelperSuffix));
+    if (credentialHelper != null) {
+      if (credentialHelper.contains(FileSystems.getDefault().getSeparator())
+          && Files.exists(Paths.get(credentialHelper))) {
```
coollog: I think we should throw an error if the credential helper path doesn't exist or else authentication may fail invisibly.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -112,10 +115,16 @@ public DefaultCredentialRetrievers setCredentialHelperSuffix(
...
+        credentialRetrievers.add(
+            credentialRetrieverFactory.dockerCredentialHelper(credentialHelper));
+      } else {
+        credentialRetrievers.add(
```
coollog: I think we shouldn't use the `credentialHelper` as a suffix if it contains a file separator. `docker-credential-some/path/to/file` doesn't seem like something we should try to use as a credential helper.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
chanseokoh: Can't remember exactly, but isn't our strategy trying the next credential if the current one fails? I wonder if we just print a warning and ignore the non-existing helper path (just skip adding it the return list) and proceed. It's basically a question whether we want to crash right away if the file doesn't exist. (Currently, I think if a credential helper doesn't exist, it doesn't crash?)

coollog: Yea, we should definitely try other strategies if certain strategies like inferring the credential helper fails. However, for the case where a path is specified and the file doesn't exist, this should be an invalid input error for the user since the user defined it themselves.

Issue #1079, opened 2018-10-01
File timestamp 0 breaks Spring WebMVCs static resource resolving


mbonato:
To create reproducible builds Jib sets a modTime of 0 for every file it includes in the Docker image layers https://github.com/GoogleContainerTools/jib/blob/f1c1bb4b081969d164d5522933a06042a53d7211/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java#L117

This unfortunately breaks resolving static resources in Spring WebApplications. One might argue, that this is bug in Spring WebMVC and in fact I've also filed a bug there (https://jira.spring.io/browse/SPR-17320) - however, the java.io.File.lastModified() uses 0L also to indicate non-existing files and I/O errors. 

In Java 6 and below (https://docs.oracle.com/javase/6/docs/api/java/io/File.html#lastModified()), 0L did not seem to be considered as a valid value. In later versions the JavaDoc also talks about 0L as a possible non-error value. "Where it is required to distinguish an I/O exception from the case where 0L is returned, or where several attributes of the same file are required at the same time, or where the time of last access or the creation time are required, then the Files.readAttributes method may be used." (https://docs.oracle.com/javase/7/docs/api/java/io/File.html#lastModified()).

Considering the contract of the java.io.File.lastModified() method, IMHO the choice of 0L for the modification date is quite unfortunate. Any fixed value other than 0 would also ensure reproducible builds but avoid the problematic error handling of java.io.File.lastModified() returning 0L also for non-existing files and I/O errors. Therefore, I would suggest to either set an unproblematic fixed value of e.g. 1L or make it configurable (with a default value other then 0L).

loosebazooka:
@mattmoor do you use 0 in bazel builds? I have a vague memory that it used some other date.

ST-DDT:
I would appreciate a configurable timestamp. Because that way I won't be restricted with my choice of `"x' is newer than x"` http checks. I would then use the timestamp of the last commit to build that image.

ST-DDT:
During the start of the application, I read some config/files from my app and publish it in a cache, but only if it is newer than the old. If during my rolling update an old container crashes unexpectedly it might restart after the last instance of my new version, thus I cannot rely on the config hash alone. Usually there would be build metadata in my jar, but jib does not build a jar, thats why I have to get the date somewhere else.

[...]

PR #1133, opened 2018-10-11
Set file timestamps in image to epoch + 1 second

chanseokoh:
Short-term solution for #1079.

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -11,7 +11,7 @@ All notable changes to this project will be documented in this file.
 
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
 - `jibExportDockerContext` generates different directory layout and `Dockerfile`.
-- File timestamps in the build image are set to 1 second since the epoch (hence 1970-01-01T00:00:01Z); previously they were set to the epoch. ([#1079](https://github.com/GoogleContainerTools/jib/issues/1079))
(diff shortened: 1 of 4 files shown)


jib-gradle-plugin/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
 - `jibExportDockerContext` generates different directory layout and `Dockerfile`.
+- File timestamps in the build image are set to 1 second since the epoch (hence 1970-01-01T00:00:01Z); previously they were set to the epoch. ([#1079](https://github.com/GoogleContainerTools/jib/issues/1079))
```
coollog: maybe add "to resolve compatibility with <Java 6 applications where file time of 0 means nonexistent"?

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
 - `jibExportDockerContext` generates different directory layout and `Dockerfile`.
+- File timestamps in the build image are set to 1 second since the epoch (hence 1970-01-01T00:00:01Z) to resolve compatibility with applications on Java 6 or below where the epoch means nonexistent or I/O errors; previously they were set to the epoch. ([#1079](https://github.com/GoogleContainerTools/jib/issues/1079))
```
loosebazooka: I think this I/O exception on 0 happens on all versions of java, not just 6- ?

chanseokoh: Yeah, my interpretation is that 0 can be a valid timestamp but may also mean an error, so to be sure, you have to use some other means to double-check, if you have to.

loosebazooka:
lgtm

PR #1136, opened 2018-10-12
Add integration tests for container.user parameter / update CHANGELOG

### LINKED ISSUE #1125: Add CHANGELOG entry, documentation updates, and integration test for `container.user`.
- [x] CHANGELOG entry
- [x] documentation updates (merge after releasing `v0.10.0`)
  - [x] `container.user`
  - [x] `extraDirectory.path` / `extraDirectory.permissions`
- [x] integration test to use `container.user` parameter

chanseokoh:
Toward #1125.


chanseokoh:
Ah, I shouldn't update `README.md` yet. Will remove that.

jib-gradle-plugin/src/integration-test/resources/projects/broken_user/src/main/java/com/test/Empty.java
```diff
@@ -0,0 +1,22 @@
+/*
```
TadCordle: We can probably just reuse the existing `empty` test project and test multiple different build configurations, like we do under the `simple` project with `build.gradle` and `complex-build.gradle`.

TadCordle: (We can do that with maven as well)

TadCordle: Right, but I think we can just create a new build.gradle (called `brokenuser-build.gradle` or something), and pass the name of that into the `TestProject#build` call with the "-b" flag. This way we can make one build that succeeds with the correct parameters and a build the fails with different parameters, without having to duplicate this Empty.java.

chanseokoh: Ah, yeah, we can have another build file.

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -10,8 +10,9 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
-- `jibExportDockerContext` generates different directory layout and `Dockerfile`.
+- `jibExportDockerContext` generates different directory layout and `Dockerfile`
```
TadCordle: I know it's not related to this PR, but what does this refer to again? Maybe we should be more specific.

chanseokoh: Makes sense. I'll put up another PR to make it more specific.

PR #1137, opened 2018-10-12
Update CHANGELOG

chanseokoh:
About https://github.com/GoogleContainerTools/jib/pull/1136#discussion_r224894466

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
-- `jibExportDockerContext` generates different directory layout in `build/jib-docker-context` than before, and the content of the generated `Dockerfile` has also been changed accordingly
+- `jibExportDockerContext` generates different directory layout and `Dockerfile` to enable WAR support ([#1007](https://github.com/GoogleContainerTools/jib/pull/1007))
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
-- `jibExportDockerContext` generates different directory layout and `Dockerfile`.
-- File timestamps in the build image are set to 1 second since the epoch (hence 1970-01-01T00:00:01Z) to resolve compatibility with applications on Java 6 or below where the epoch means nonexistent or I/O errors; previously they were set to the epoch. ([#1079](https://github.com/GoogleContainerTools/jib/issues/1079))
+- `jibExportDockerContext` generates different directory layout in `build/jib-docker-context` than before, and the content of the generated `Dockerfile` has also been changed accordingly
```
TadCordle: Sorry, I meant more along the lines of how they've changed, not what's changed. What's different about the directory layout/Dockerfile? Was it for WAR support or something?

chanseokoh: It's kind of difficult to explain the details in a couple sentences. The change is to add support for WAR, but it affects all builds, whether or not WAR. Previously, we had

```
jib-docker-context/
    classes/
        com/
            example/
                HelloWorld.class
```
and
```Dockerfile
COPY classes /app/classes/
```

Now, we have
```
jib-docker-context/
    classes/
        app/
            classes/
                com/
                    example/
                        HelloWorld.class
```
and
```Dockerfile
COPY classes /
```

I don't think we need to explain to this level of detail. Or, if you have some wording suggestions, let me know.

TadCordle: Yeah, we don't need to explain in that level of detail. But maybe we can say something like "`jibExportDockerContext` generates different directory layout and `Dockerfile` to prepare for WAR support", or something along the lines of "to add more layering flexibility". Just the intent without all the technical details, since as is it sounds like we just changed the directory structure for no reason.

chanseokoh: Ah, good point. I think "... and `Dockerfile` to enable WAR support"  works.

PR #1146, opened 2018-10-15
Log CMD and move code that logs ENTRYPOINT

chanseokoh:
Changes:
1. Log CMD (in addition to the current ENTRYPOINT)
2. Move the ENTYPOINT-logging code from `BuildSteps` to `BuildImageStep`.

Logging CMD means in the some cases (including the default case), it will say the "program arguments set to null", which is saying the truth:
```
Finalizing...

Container entrypoint set to [java, -cp, /app/resources:/app/classes:/app/libs/*, example.HelloWorld]
Container program arguments set to null

Built and pushed image as ...
```

Moving the ENTRYPOINT-logging code means now the log may appear before "Finalizing..."
```
[INFO] Container entrypoint set to [/usr/bin/java, -jar] (inherited from base image)
[INFO] Container program arguments set to [/jetty/start.jar] (inherited from base image)
[INFO] Finalizing...
```

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
@@ -31,20 +31,18 @@ public class BuildSteps {
   private static final String DESCRIPTION_FOR_DOCKER_DAEMON = "Building image to Docker daemon";
   private static final String DESCRIPTION_FOR_TARBALL = "Building image tarball";
 
-  /** Accepts {@link StepsRunner} by running the appropriate steps. */
+  /** Runs appropriate steps to build an image. */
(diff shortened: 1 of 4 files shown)


coollog:
Hmm I think we should just not display anything for program arguments if they were set to `null`. Another option is to just tack the program arguments onto the end of the entrypoint list.

briandealwis:
I don't much like `null` — generally outputting `null` makes it look like a coding error.  I think it's better to show an explicit message (_No container program arguments were configured_). Given that we report values when they're inherited, I'd support @coollog's suggestion to not show anything.

chanseokoh:
> I think we should just not display anything for program arguments if they were set to `null`

Yeah, I'll skip it if it's set to `null`.

Issue #1138, opened 2018-10-12
Updating docs about container.args parameter?


chanseokoh:
Property | Type | Default | Description
-- | -- | -- | --
`args` | list | *None* | Default main method arguments to run your application with.

The description that it's the Java `main` method arguments made sense when Jib did not have the feature to customize ENTRYPOINT. I think it makes more sense to update the description to

"Additional program arguments appended to the command to start the container (similar to Docker's `CMD` instruction in relation with `ENTRYPOINT`)". In the default case where you do not set a custom `container.entrypoint`, this parameter is effectively the arguments to the `main` method of your Java application."

But still this will not be complete, especially after #1071 when we will have different behaviors of inheriting ENTRYPOINT and CMD for WAR vs. non-WAR.

(Don't forget to also update the Javadoc of `JibPluginConfiguration.getArgs()`.)

PR #1147, opened 2018-10-16
Update README for 0.10.0

### LINKED ISSUE #1125: Add CHANGELOG entry, documentation updates, and integration test for `container.user`.

chanseokoh:
Fixes #1125.

Should be merged after the next release, so labeling with "PR: Not Ready".


jib-gradle-plugin/README.md
```diff
@@ -202,6 +202,8 @@ Property | Type | Default | Description
 `mainClass` | `String` | *Inferred\** | The main class to launch your application from.
 `ports` | `List<String>` | *None* | Ports that the container exposes at runtime (similar to Docker's [EXPOSE](https://docs.docker.com/engine/reference/builder/#expose) instruction).
 `useCurrentTimestamp` | `boolean` | `false` | By default, Jib wipes all timestamps to guarantee reproducibility. If this parameter is set to `true`, Jib will set the image's creation timestamp to the time of the build, which sacrifices reproducibility for easily being able to tell when your image was created.
+`user` | `String` | `0:0` | The user and group to run the container as. {@code user} can be a username or UID along with an optional groupname or GID. The following are all valid: `user`, `uid`, `user:group`, `uid:gid`, `uid:group`, `user:gid`.
```
coollog: Does `@code` render into somehting?

coollog: Isn't the default `*None*`? (We don't set the user so the container will run with whatever default user the runtime chooses)

chanseokoh: Uh... this Stack Overflow answer says if it's not configured, it will be the root: https://stackoverflow.com/a/52072782/1701388 The answer seems right to me. I think it's not like the user changes depending on where you run the container. This article also says "By default docker containers run as `root`": https://www.projectatomic.io/docs/docker-image-author-guidance/

chanseokoh: Oops, catch-paste error from a Javadoc.

coollog: I think we should keep our documentation more about what we build into the image and not what the runtime implementation chooses as the default user. Since we don't explicitly set the user to anything if not specified, it is up to the runtime to choose which user to use, where Docker chooses root, but other runtimes could potentially choose other values.

chanseokoh: Makes sense. I was thinking our doc was about defaults is more about describing what would be the end result of the image built by Jib, but sounds like rather we're shaping the doc as more about what Jib would simply do.

jib-gradle-plugin/README.md
```diff
@@ -250,15 +257,27 @@ jib {
 
 You can add arbitrary, non-classpath files to the image by placing them in a `src/main/jib` directory. This will copy all files within the `jib` folder to the image's root directory, maintaining the same structure (e.g. if you have a text file at `src/main/jib/dir/hello.txt`, then your image will contain `/dir/hello.txt` after being built with Jib).
 
-You can configure a different directory by using the `extraDirectory` parameter in your `build.gradle`:
+You can configure a different directory by using the `extraDirectory` parameter in your `build.gradle`, either by using the `jib.extraDirectory.path` parameter:
```
coollog: I think this would be clearer if the two options were first presented and then the examples below were shown with the two options, rather than separating the options descriptions with the examples.

jib-gradle-plugin/README.md
```diff
@@ -250,15 +257,27 @@ jib {
...
 
 ```groovy
-jib {
-  ...
-  // Copies files from 'src/main/custom-extra-dir' instead of 'src/main/jib'
-  extraDirectory = file('src/main/custom-extra-dir')
-  ...
-}
+// Copies files from 'src/main/custom-extra-dir' instead of 'src/main/jib'
+jib.extraDirectory.path = file('src/main/custom-extra-dir')
```
coollog: Should we just encourage them to keep the same usage `jib.extraDirectory =` rather than showing this `.path` alternative, which may be confusing?

jib-gradle-plugin/README.md
```diff
@@ -248,15 +257,27 @@ jib {
...
+
+```groovy
+// Copies files from 'src/main/custom-extra-dir' instead of 'src/main/jib'
+jib.extraDirectory = file('src/main/custom-extra-dir')
+```
+
+The `extraDirectory` parameter can also be used to set the file permissions of files added to the container by specifying the absolute path of the file on the container as well as the octal representation of its file permission bits in the `jib.extraDirectory.permissions` field:
+
+```groovy
+jib.extraDirectory.permissions = [
```
chanseokoh: When using `.permissions`, it is still possible to directly set `jib.extraDirectory = file(...)` without `.path`? If not, we should clarity this.

jib-maven-plugin/README.md
```diff
@@ -306,15 +315,37 @@ In this configuration, the image:
...
+
+```xml
+<!-- Copies files from 'src/main/custom-extra-dir' instead of 'src/main/jib' -->
+<extraDirectory>${project.basedir}/src/main/custom-extra-dir</extraDirectory>
+```
+
+The `extraDirectory` parameter can also be used to set the file permissions of files added to the container by specifying the absolute path of the file on the container as well as the octal representation of its file permission bits in the `<extraDirectory><permissions>` field:
+
+```xml
+<extraDirectory>
```
chanseokoh: Similarly, when using `<permissions>`, it is still possible to set the path without `<path> like below?

```xml
<extraDirectory>
   /some/path
  <permissions>...</permissions>
</extraDirectory>
```

jib-gradle-plugin/README.md
```diff
@@ -161,6 +161,7 @@ Field | Type | Default | Description
 `to` | [`to`](#to-closure) | *Required* | Configures the target image to build your application to.
 `from` | [`from`](#from-closure) | See [`from`](#from-closure) | Configures the base image to build your application on top of.
 `container` | [`container`](#container-closure) | See [`container`](#container-closure) | Configures the container that is run from your built image.
+`extraDirectory` | [`extraDirectory`](#extradirectory-closure) / `File` | See `extraDirectory` | Configures the directory used to add arbitrary files to the image.
```
chanseokoh: Isn't default always project_dir/src/main/jib?

jib-maven-plugin/README.md
```diff
@@ -202,6 +202,7 @@ Field | Type | Default | Description
 `to` | [`to`](#to-object) | *Required* | Configures the target image to build your application to.
 `from` | [`from`](#from-object) | See [`from`](#from-object) | Configures the base image to build your application on top of.
 `container` | [`container`](#container-object) | See [`container`](#container-object) | Configures the container that is run from your image.
+`extraDirectory` | [`extraDirectory`](#extradirectory-object) / string | See `extraDirectory` | Configures the directory used to add arbitrary files to the image.
```
chanseokoh: Isn't default always project_dir/src/main/jib?

jib-gradle-plugin/README.md
TadCordle: Oh yeah, I meant to specify that in the `extraDirectory` table but deleted it in favor of the minimal table layout.

jib-gradle-plugin/README.md
TadCordle: I wonder if it'd be clearer if, instead of talking about 1 way to configure 1 parameter + 2 ways to configure another parameter, we frame it as two possible options.

- a minimal configuration: 
```
extraDirectory = file('...')
```
- and a full configuration: 
```
extraDirectory {
  path = file('...')
  permissions = [...]
}
```

I think this would also help with the other ambiguities you and @chanseokoh mentioned.

jib-gradle-plugin/README.md
```diff
@@ -257,27 +257,26 @@ jib {
...
-jib.extraDirectory.path = file('src/main/custom-extra-dir')
-```
-
-or by setting the path on the `extraDirectory` parameter directly:
-
+You can configure a different directory by using the `jib.extraDirectory` parameter in your `build.gradle`:
 ```groovy
-// Copies files from 'src/main/custom-extra-dir' instead of 'src/main/jib'
-jib.extraDirectory = file('src/main/custom-extra-dir')
+jib {
```
coollog: I think this is an extra layer of `jib`?

chanseokoh:
I will do the WAR part.

TadCordle:
Oh, we should probably address #1138 as well.

jib-gradle-plugin/README.md
```diff
@@ -350,6 +350,26 @@ jib {
 }
 ```
 
+### WAR Projects
+
+Jib supports creating images for projects that build a WAR. If the Gradle project uses the [WAR Plugin](https://docs.gradle.org/current/userguide/war_plugin.html), Jib will by default use the [distroless Jetty]https://github.com/GoogleContainerTools/distroless/tree/master/java/jetty) as a base image to deploy the project WAR. No extra configuration is necessary than using the WAR Plugin to make Jib build WAR images.
+
+Note that Jib will work slightly differently for WAR projects from plain Java projects:
```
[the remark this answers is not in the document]
chanseokoh: Not really the right terminology. I've been using non-WAR projects, but I remember @briandealwis said (in a different context) "maybe more accurate to say JAR". It'd be good to pick one terminology and stick to it.

chanseokoh:
I'm not so familiar with the image ID and digest honestly. We don't have to add it in this PR, as long as we don't forget to add it soon.

chanseokoh:
LGTM for the parts not from me. Since it's me who put up the PR, this PR needs an approval from others than me.

PR #1152, opened 2018-10-16
Fix missing env in maven docker context generator

TadCordle:
Again, changes for Maven 0.9.13 patch (skipping 0.9.12 to stay even with the Gradle plugin). Will close and re-open with the release PR after approval.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
@@ -139,7 +139,7 @@ public class JavaDockerContextGenerator {
   @Nullable private String baseImage;
   private List<String> entrypoint = Collections.emptyList();
   private List<String> javaArguments = Collections.emptyList();
-  private Map<String, String> environment = Collections.emptyMap();
+  @Nullable private Map<String, String> environment;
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -139,7 +139,7 @@ private static String mapToDockerfileString(Map<String, String> map, String comm
   @Nullable private String baseImage;
   private List<String> entrypoint = Collections.emptyList();
   private List<String> javaArguments = Collections.emptyList();
-  private Map<String, String> environment = Collections.emptyMap();
+  @Nullable private Map<String, String> environment;
```
coollog: Should we have this be consistent with the others and not be nullable?

TadCordle: entrypoint and javaArguments actually become nullable in a future change, I think.

coollog: Ah right. I think if we make environment nullable, we should make exposedPorts and labels nullable too but that change can go in the post-release PR.

TadCordle:
#1153

PR #1153, opened 2018-10-16
Maven release v0.9.13

TadCordle:


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGenerator.java
```diff
@@ -336,7 +336,9 @@ String makeDockerfile() throws JsonProcessingException {
       dockerfile.append("\nEXPOSE ").append(port);
     }
 
-    dockerfile.append(mapToDockerfileString(environment, "ENV"));
+    if (environment != null) {
+      dockerfile.append(mapToDockerfileString(environment, "ENV"));
```
chanseokoh: I think `environment` and `labels` are similar in various ways. Can we be consistent?

TadCordle: Yeah, wasn't sure if we wanted to make labels and exposed ports nullable in this PR or a followup? I guess we can do this one.

PR #1155, opened 2018-10-16
Write out image digest to <target>/jib-image.digest

### LINKED ISSUE #933: Give access to Image ID
Docker v1.10 introduced a content/digest-based image ID to uniquely identify docker images. It would be great if this image ID would be available from the gradle (and maybe maven) plugin after the image has been built or is deemed being already up-to-date. 

Having the image ID would, for example, let the subsequent deployment processes skip containers that do not change (without having to resolve tags to image ids or workaround with manual mechanisms). This is of particular importance for larger multi-project gradle setups / single repository approaches where a single build manages multiple images.

briandealwis:
Fixed #933


briandealwis:
I'm not enamoured by this solution which has a lot of repetition.  I'll look into instead leveraging the eventing system.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -88,16 +90,26 @@ public FileCollection getInputFiles() {
...
+   * Returns the location for the tarball. By default, it is {@code build/jib-image.tar}.
+   *
+   * @return the tarball location
+   */
+  private Path getTarPath() {
+    return getProject().getBuildDir().toPath().resolve("jib-image.tar");
+  }
+
+  /**
+   * Returns the output directory for the image digest. By default, it is {@code
```
coollog: output file

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -88,16 +90,26 @@ public FileCollection getInputFiles() {
...
+   * build/jib-image.digest}.
    *
    * @return the output directory
    */
-  private String getTargetPath() {
```
coollog: nit: looks like this was in the wrong location before - let's move these private methods to the bottom.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -117,6 +132,8 @@ private static String buildAndRunAdditionalTag(
     verifier.executeGoals(Arrays.asList("clean", "compile", "jib:build"));
     verifier.verifyErrorFreeLog();
 
+    assertImageDigest(projectRoot);
```
coollog: For the test, we can actually pull and build using the digest that was generated to assert that the image was actually pushed with that digest.

briandealwis:
So I tried various other approaches and creating an `ImageCreatedEvent` seemed the cleanest approach IMHO.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -61,21 +62,30 @@ static GradleProjectProperties getForProject(
...
-            new TimerEventHandler(message -> logEventHandler.accept(LogEvent.debug(message))));
+    EventHandlers handlers =
+        new EventHandlers()
+            .add(JibEventType.LOGGING, logEventHandler)
+            .add(
+                JibEventType.TIMING,
+                new TimerEventHandler(message -> logEventHandler.accept(LogEvent.debug(message))));
+    
+    Path buildOutputPath = project.getBuildDir().toPath();
+    Path digestOutputPath = buildOutputPath.resolve("jib-image.digest");
```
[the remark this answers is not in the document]
coollog: Oh interesting, I thought we had that as configurable - nevermind then.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibRunHelper.java
```diff
@@ -130,15 +130,11 @@ static void assertCreationTimeEpoch(String imageReference)
         new Command("docker", "inspect", "-f", "{{.Created}}", imageReference).run().trim());
   }
 
-  static void assertImageDigest(Path projectRoot) {
+  static String assertImageDigest(Path projectRoot) throws IOException, DigestException {
```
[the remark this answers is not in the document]
briandealwis: Hmm no.  We're testing the result in the maven tests, so I'll drop it from here.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainer.java
```diff
@@ -17,22 +17,63 @@
...
 public class JibContainer {
 
+  /**
+   * Create a container.
+   *
+   * @param imageDigest the digest of the registry image manifest
+   * @param imageId digest of the container configuration
+   * @return the new container
+   */
+  public static JibContainer create(DescriptorDigest imageDigest, DescriptorDigest imageId) {
```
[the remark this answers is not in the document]
coollog: I think we should keep the package-hierarchy free of cyclical dependencies such that `.api` depends on `.build.steps` and there isn't a reference in the other direction - so an alternative here would be to have the build steps return the image digest and id (via some class like `BuildSteps.Result` or something) and have `JibContainerBuilder` create the `JibContainer` using those return values.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainer.java
```diff
@@ -66,14 +60,24 @@ public int hashCode() {
...
+    return imageDigest.equals(otherContainer.imageDigest) && imageId.equals(otherContainer.imageId);
+  }
+
+  /**
+   * Write out the image digest to the given location.
+   *
+   * @param location the location to write the digest
+   * @throws IOException a problem occurred when writing the digest
+   */
+  public void writeImageDigest(Path location) throws IOException {
```
coollog: Hmm, I'm not sure if this would be that useful of a method to expose as part of the user-facing API for `JibContainer`. I think the general rule for including something in our API should be that by including it, we are encouraging its use. I'm not sure if we should be encouraging writing out an image digest to a file.

For the `BuildStepsRunner` chaining, I was referring to something like `BuildStepsRunner.ImageDigestWriter` that was the return class and it has the `writeImageDigest` method, rather than `JibContainer`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -56,6 +57,16 @@
  */
 // TODO: Add tests once containerize() is added.
 public class JibContainerBuilder {
+  /**
```
coollog: nit: new line after class definition

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -57,6 +57,16 @@
...
+   * Create a container for a given build result. For internal use only.
+   *
+   * @param imageDigest the digest of the registry image manifest
+   * @param imageId digest of the container configuration
+   * @return the new container
+   */
+  public static JibContainer created(DescriptorDigest imageDigest, DescriptorDigest imageId) {
```
coollog: Yea, so for example related to https://github.com/GoogleContainerTools/jib/pull/1155#discussion_r227173561, here, it feels a bit off that `JibContainerBuilder` calls into `BuildSteps` to run the build and `BuildSteps` calls `JibContainerBuilder` to create its result.

coollog:
Let's add a CHANGLOG entry as well.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildResult.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(imageDigest, imageId);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) return true;
```
TadCordle: nit: I think we generally still use brackets for single-line if-statements.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildResult.java
```diff
@@ -36,4 +37,19 @@ public DescriptorDigest getImageDigest() {
...
+    if (obj == null) return false;
+    if (getClass() != obj.getClass()) return false;
```
coollog: I think we generally use `instanceof` even though we don't use inheritance, but it does save the extra null check.

briandealwis: I meant to look up Bloch previously and he discusses at length in _Effective Java_ how `.equals()` is required to be symmetric and thus `instanceof` should never be used.

briandealwis: Filed #1202

Issue #794, opened 2018-08-03
executable permission is never set


tbutter:
**Description of the issue**:

If you add a executable file to src/main/resources or src/main/jib the resulting docker image will include the files only with mode 0644.

**Expected behavior**:

The files should have a mode of 0744

**Additional Information**: 

https://github.com/GoogleContainerTools/jib/blob/8fbdb2e1b1018565388c3036540125b852ddc142/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java#L68

Creates a new TarArchiveEntry for each File. The TarArchiveEntry will always use `TarArchiveEntry. DEFAULT_FILE_MODE` for all files (0100644).

A possible change could be something like this:

```java
if(java.nio.file.Files.isExecutable(sourceFile))
  tarArchiveEntry.setMode(tarArchiveEntry.getMode() | 0100);
```

coollog:
Hi @tbutter , thanks for filing this issue and providing the recommended fix! This is related to #523 and #727. Will defer to @GoogleContainerTools/java-tools for some comments about the approach.

coollog:
We will most likely implement this fix by **reading the file permissions set in `src/main/jib` and using those permissions explicitly** over the defaults provided by `TarArchiveEntry`. We won't be doing any changes for classes/resources/dependencies files since those can be part of build output and vary in file permissions between different environments (and thus may break reproducibility if used). Therefore, this would only support executable files placed under `src/main/jib` such that the file permissions for those files would be generally committed into source control and reproducible across environments.

chanseokoh:
- Was going to say this is related to "reproducibility".
- If implemented, may need to consider setting the executable bit for group and others, as needed in #523 and #727.
- I'm not super clear about executable permission on Windows. It has always been confusing.

chanseokoh:
> `src/main/jib` such that the file permissions for those files would be generally committed into source control and reproducible across environments.

Generally speaking, this is not true. For example,

```
$ umask
0022
$ git clone https://github.com/GoogleContainerTools/jib.git
$ ls -l jib/README.md 
-rw-r--r-- 1 chanseok chanseok 4691 Aug  3 14:38 jib/README.md
$ umask 077
$ rm -rf jib
$ git clone https://github.com/GoogleContainerTools/jib.git
$ ls -l jib/README.md 
-rw------- 1 chanseok chanseok 4691 Aug  3 14:38 jib/README.md  
```

[...]

PR #1156, opened 2018-10-16
Core changes for setting file permissions

TadCordle:
Towards #794.

Adds a `FilePermissions` field to `LayerEntry` and applies the permissions to the `TarArchiveEntry` the `LayerEntry`s are added to. Also adds a couple methods to `LayerConfiguration` and `JavaLayerConfiguration` to take a map of file permissions.


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -70,6 +70,21 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
...
+     * Adds an entry to the layer. Only adds the single source file to the exact path in the
+     * container file system with the given permissions.
+     *
+     * @param sourceFile the source file to add to the layer
+     * @param pathInContainer the path in the container file system corresponding to the {@code
+     *     sourceFile}
+     * @param permissions the file permissions on the container
+     * @return this
+     */
+    public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer, int permissions) {
```
coollog: Hmm, maybe we should use `PosixFilePermission` instead since that is the class that represents file permissions.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -70,6 +70,21 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
       return this;
```
coollog: We could probably have just one constructor for `LayerEntry` and just call it ehre with like `null` for the third parameter (would need to use a permissions class).

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -126,6 +126,44 @@ public Builder addDirectoryContents(
...
+     *
+     * @param sourceRoot root directory whose contents will be added
+     * @param basePathInContainer directory in the layer into which the source contents are added
+     * @param permissionsMap the map from absolute path on container to file permission
+     * @return this
+     * @throws IOException error while listing directories
+     * @throws NotDirectoryException if {@code sourceRoot} is not a directory
+     */
+    // TODO: Use in plugins
+    public Builder addExtraDirectoryContents(
```
coollog: Same logic here as in the comment for the permissions-vs.-no-permissions methods. I think this can just be `addDirectoryContents` as well (we'll only use it for extra directories), and then the other `addDirectoryContents` would just call this one with no permissions to set, so that the code duplication can be reduced.

coollog: I think for this method, we should also be setting the same permissions for all subfiles of a directory if a permission was specified for that directory?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -56,6 +59,25 @@
   public LayerEntry(Path sourceFile, AbsoluteUnixPath extractionPath) {
     this.sourceFile = sourceFile;
     this.extractionPath = extractionPath;
+    this.permissions = Files.isDirectory(sourceFile) ? 0755 : 0644;
```
coollog: Let's have these as constants

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -100,9 +100,11 @@ public Blob build() {
     for (LayerEntry layerEntry : layerEntries) {
       // Adds the entries to uniqueTarArchiveEntries, which makes sure all entries are unique and
       // adds parent directories for each extraction path.
-      uniqueTarArchiveEntries.add(
+      TarArchiveEntry entry =
           new TarArchiveEntry(
-              layerEntry.getSourceFile().toFile(), layerEntry.getAbsoluteExtractionPathString()));
+              layerEntry.getSourceFile().toFile(), layerEntry.getAbsoluteExtractionPathString());
+      entry.setMode((entry.getMode() & ~0777) | layerEntry.getPermissions());
```
chanseokoh: `entry.getMode() & ~0777`

This looks weird. I think you are always resetting the right-most three bits to 0, effectively making the number to 0.

Also, are you sure `TarArchiveEntry.setMode(int)` expects an octal number? The Javadoc doesn't say anything about it, but I think it's very possible that `setMode()` just expects a decimal number.

TadCordle: I'll change it for clarity, but to clear things up:

>are you sure TarArchiveEntry.setMode(int) expects an octal number

Numbers starting with 0 are treated as octal in java; i.e. `0777` and `511` are treated as the same thing. So you're right that `TarArchiveEntry.setMode` expects a decimal number, but passing in an octal representation doesn't make a difference as long as it's specified as octal.

>This looks weird. I think you are always resetting the right-most three digits to 0, effectively making the number to 0.

The reason for the inversion in `~0777` is to make the lower 9 bits `0` and the remaining upper bits `1` so only the permission bits are masked out, because the `mode` int contains more than just permissions.

chanseokoh: Okay, let me rephrase my question: so, in order to set the executable, write, and read bits to everything (i.e., the usual "777" numeric notation), should it be `TarArchiveEntry.setMode(777)` or `TarArchiveEntry.setMode(511)`? I thought it should be the former, but I think you are passing the latter.

> This looks weird. I think you are always resetting the right-most three digits to 0, effectively making the number to 0.

Sorry, never mind. I confused `entry.getMode()` as `layerEntry.getPermissions()`.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
TadCordle: `PosixFilePermission` doesn't seem all that easy to use in our case. In the end we're going to be parsing a number from the build config, converting it into a `Set<PosixFilePermission>` to store, and then converting that set back to a number because there doesn't seem to be any way to set posix file permissions directly on `TarArchiveEntry`, the only place we actually need to use permissions.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
TadCordle: >I think this can just be addDirectoryContents as well ...

I felt the need to specify "Extra" in the name since I got rid of the `LayerType` parameter. But yeah I think we can just make it an overload that can take an empty permissions map.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
TadCordle: Judging by the [source of `TarArchiveEntry`](https://commons.apache.org/proper/commons-compress/jacoco/org.apache.commons.compress.archivers.tar/TarArchiveEntry.java.html), it would take `setMode(511)`; it has a couple default modes that show the layout it's expecting.

L217-221:
```
/** Default permissions bits for directories */
public static final int DEFAULT_DIR_MODE = 040755;

/** Default permissions bits for files */
public static final int DEFAULT_FILE_MODE = 0100644;
```

chanseokoh: Ah, interesting. Thanks for the clarification.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
chanseokoh: Just wondering if this (i.e., explicitly setting directories to 0755 and files to 0644) changes the current behavior, because it seems like we haven't done anything about permissions.

TadCordle: 0755 and 0644 were the existing defaults from `TarArchiveEntry`, so it shouldn't have changed anything.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Converts a 3-digit octal integer to an equivalent set of {@link PosixFilePermission}.
+   *
+   * @param permissions the 3-digit octal integer
+   * @return the equivalent {@link PosixFilePermission} set
+   */
+  public static Set<PosixFilePermission> toSet(int permissions) {
+    Preconditions.checkArgument(
+        permissions >= 0 && permissions <= 511, "Permissions must be between 000 and 777 octal");
```
chanseokoh: Does it make sense to make the check complete? (For example, the number 99 will pass.)

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -34,8 +39,14 @@
  */
 public class LayerEntry {
 
+  private static final Set<PosixFilePermission> DEFAULT_FILE_PERMISSIONS =
```
chanseokoh: This is actually mutable, so the all caps don't look right.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,61 @@
...
+    ImmutableSet.Builder<PosixFilePermission> result = ImmutableSet.builder();
+    for (int bit = 0; bit < 9; bit++) {
+      if ((permissions & (1 << bit)) != 0) {
+        result.add(PosixFilePermission.values()[8 - bit]);
```
chanseokoh: I personally prefer

```java
if ((permissions & 0100) == 0100) {
   result.add(PosixFilePermission.OWNER_EXECUTE);
}
```
@GoogleContainerTools/java-tools what do you think?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -115,14 +117,49 @@ public Builder addDirectoryContents(
...
+     * @param sourceRoot root directory whose contents will be added
+     * @param pathFilter filter that determines which files (not directories) should be added
+     * @param basePathInContainer directory in the layer into which the source contents are added
+     * @param permissionsMap the map from absolute path on container to file permission
+     * @return this
+     * @throws IOException error while listing directories
+     * @throws NotDirectoryException if {@code sourceRoot} is not a directory
+     */
+    // TODO: Use in plugins
+    public Builder addDirectoryContents(
```
chanseokoh: I think we need the permission version for `addFile()` too (the single-file addition method). It's being used.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
chanseokoh: Oops, never mind.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
TadCordle: I was thinking about this, but I didn't like the repetition that came out of it.

TadCordle:
>I think for this method, we should also be setting the same permissions for all subfiles of a directory if a permission was specified for that directory?

I'm thinking I'm going to leave this for another PR.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,61 @@
...
+      }
+    }
+    return result.build();
```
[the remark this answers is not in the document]
chanseokoh: I wonder `Collections.unmodifiableSet()` is enough, getting rid of `ImmutableSet`.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,79 @@
...
+    result.put(PosixFilePermission.OWNER_WRITE, 0b010000000);
+    result.put(PosixFilePermission.OWNER_EXECUTE, 0b001000000);
+    result.put(PosixFilePermission.GROUP_READ, 0b000100000);
+    result.put(PosixFilePermission.GROUP_WRITE, 0b000010000);
+    result.put(PosixFilePermission.GROUP_EXECUTE, 0b000001000);
+    result.put(PosixFilePermission.OTHERS_READ, 0b000000100);
+    result.put(PosixFilePermission.OTHERS_WRITE, 0b000000010);
+    result.put(PosixFilePermission.OTHERS_EXECUTE, 0b000000001);
+    return result;
+  }
```
chanseokoh: If you prefer, you can used the `static` initialization block and remove the method. I believe you can also do `PERMISSION_MAP = Collections.unmodifiedSet(result)`, which will effectively be an immutable set.

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurationsTest.java
```diff
@@ -379,4 +392,65 @@ public void testAddDirectoryContents_complex() throws IOException {
...
         configurations.getExtraFilesLayerEntries());
   }
+
+  @Test
+  public void testAddExtraDirectoryContents() throws IOException {
+    temporaryFolder.newFolder("src", "main", "jib");
+    temporaryFolder.newFile("src/main/jib/fileA");
+    temporaryFolder.newFile("src/main/jib/fileB");
+    temporaryFolder.newFile("src/main/jib/fileC");
+    temporaryFolder.newFolder("src", "main", "jib", "folder");
```
chanseokoh: Call this once, and you don't need the first `newFolder("src", "main", "jib")`.

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurationsTest.java
```diff
@@ -379,4 +392,65 @@ public void testAddDirectoryContents_complex() throws IOException {
...
+    temporaryFolder.newFile("src/main/jib/folder/fileD");
+    temporaryFolder.newFile("src/main/jib/folder/fileE");
+    temporaryFolder.newFile("src/main/jib/folder/folder2");
```
chanseokoh: Perhaps `newFile` --> `newFolder`. In that case, calling `newFolder("src", "main", "jib", "folder", "folder2")` will be sufficient (no need for multiple `newFolder()` calls).

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+/** Helpers for converting file permissions. */
+public class PermissionsHelper {
+
+  private static final Map<PosixFilePermission, Integer> permissionMap;
```
coollog: An alternative for this would be to just create an `ImmutableMap` using `ImmutableMap.of` directly.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+
+  static {
+    Map<PosixFilePermission, Integer> result = new HashMap<>();
+    result.put(PosixFilePermission.OWNER_READ, 0b100000000);
```
coollog: I think it might make more sense to use octal here since we're using octal everywhere else - so reads will be 4, writes will be 2 and executes will be 1.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Converts a set of {@link PosixFilePermission} to its 3-digit octal integer representation.
+   *
+   * @param permissions the set of {@link PosixFilePermission}
+   * @return the 3-digit octal integer representation of the set
+   */
+  public static int toInt(Set<PosixFilePermission> permissions) {
+    int result = 0;
```
coollog: nit: more descriptive name like `octalPermissions`?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
TadCordle: I think `ImmutableMap.of` takes at most 5 entries.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -66,7 +69,27 @@ public Builder setName(String name) {
...
+
+    /**
+     * Adds an entry to the layer with the given permissions. Only adds the single source file to
+     * the exact path in the container file system. See {@link Builder#addEntry(Path,
+     * AbsoluteUnixPath)} for more information.
+     *
+     * @param sourceFile the source file to add to the layer
+     * @param pathInContainer the path in the container file system corresponding to the {@code
+     *     sourceFile}
+     * @param permissions the file permissions on the container
```
briandealwis: If null then…

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.common.base.Preconditions;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+/** Helpers for converting file permissions. */
```
briandealwis: Nit: in [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html) the integer representation is called the _file mode_ or _file permission bits_.  The integer isn't really in octal — we represent the permissions in octal.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -115,14 +142,49 @@ public Builder addDirectoryContents(
...
+        LayerType layerType,
+        Path sourceRoot,
+        Predicate<Path> pathFilter,
+        AbsoluteUnixPath basePathInContainer,
+        Map<AbsoluteUnixPath, Set<PosixFilePermission>> permissionsMap)
```
briandealwis: Wondering if it might not be more flexible to encode this instead as a `Function<AbsoluteUnixPath,Set<PosixFilePermission>>`?  (Or a `BiFunction` with the actual on-disk path?)

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Converts a 3-digit octal integer to an equivalent set of {@link PosixFilePermission}.
+   *
+   * @param permissions the 3-digit octal integer
+   * @return the equivalent {@link PosixFilePermission} set
+   */
+  public static Set<PosixFilePermission> toSet(int permissions) {
+    Preconditions.checkArgument(
+        permissions >= 0 && permissions <= 0777, "Permissions must be between 000 and 777 octal");
+    HashSet<PosixFilePermission> result = new HashSet<>();
```
briandealwis: EnumSet

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -100,9 +101,16 @@ public Blob build() {
...
-      uniqueTarArchiveEntries.add(
+      TarArchiveEntry entry =
           new TarArchiveEntry(
-              layerEntry.getSourceFile().toFile(), layerEntry.getAbsoluteExtractionPathString()));
+              layerEntry.getSourceFile().toFile(), layerEntry.getAbsoluteExtractionPathString());
+
+      // Sets the entry's permissions by masking out the permission bits from the entry's mode (the
+      // lowest 9 bits) then using a bitwise OR to set them to the layerEntry's permissions.
+      entry.setMode(
+          (entry.getMode() & ~0777) | PermissionsHelper.toInt(layerEntry.getPermissions()));
```
briandealwis: Have you seen the upper bits being used?

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGeneratorTest.java
```diff
@@ -73,7 +73,7 @@ private static void assertSameFiles(Path directory1, Path directory2) throws IOE
         .map(
             sourceFile ->
                 new LayerEntry(
-                    sourceFile, extractionRoot.resolve(directory.relativize(sourceFile))))
+                    sourceFile, extractionRoot.resolve(directory.relativize(sourceFile)), null))
```
briandealwis: I wonder if this Dockerfile generator should be writing out `RUN chmod xxx` for the various files…

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -52,10 +63,21 @@
...
    * @param extractionPath the path in the container file system corresponding to the {@code
    *     sourceFile}
+   * @param permissions the file permissions on the container. Use {@code null} to use defaults (644
+   *     for files, 755 for directories)
    */
-  public LayerEntry(Path sourceFile, AbsoluteUnixPath extractionPath) {
+  public LayerEntry(
+      Path sourceFile,
+      AbsoluteUnixPath extractionPath,
+      @Nullable Set<PosixFilePermission> permissions) {
```
briandealwis: Given all the changes to add a `null` parameter, it might be best to retain the original 2-argument contractor.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -52,10 +63,21 @@
...
     this.sourceFile = sourceFile;
     this.extractionPath = extractionPath;
+    if (permissions == null) {
+      this.permissions =
```
briandealwis: I'm wondering if this logic might not be better put into the tar-building so that we can differentiate between files that had explicit permissions and those that didn't.  This would be important if we modify the Dockerfile generator to do a `RUN chmod xxx` for the specific files.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
TadCordle: Almost all of the changes were in tests, and `LayerEntry` isn't supposed to be used by jib-core users, so I think it's fine.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
TadCordle: They're used to specify file type (e.g. in `TarArchiveEntry`, files start with `0100` and directories start with `040`)

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGeneratorTest.java
chanseokoh: Hmm... that's a good point. But probably distroless doesn't have `chmod`, so it may not work in the default case. Not sure what we can actually do.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
TadCordle: Looks like this requires a custom implementation?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
coollog: Ah okay, then it could be `ImmutableMap.builder().put(...).put(...)....build()`

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
TadCordle: That's true, I think I'll move it.

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/JavaDockerContextGeneratorTest.java
coollog: I think we should actually give a warning that file permissions are not supported  when the user tries to export a docker context with file permissions configured.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
TadCordle: Not sure what you mean, can you be more specific?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -17,38 +17,34 @@
...
-    result.put(PosixFilePermission.OWNER_EXECUTE, 0100);
-    result.put(PosixFilePermission.GROUP_READ, 040);
-    result.put(PosixFilePermission.GROUP_WRITE, 020);
-    result.put(PosixFilePermission.GROUP_EXECUTE, 010);
-    result.put(PosixFilePermission.OTHERS_READ, 04);
-    result.put(PosixFilePermission.OTHERS_WRITE, 02);
-    result.put(PosixFilePermission.OTHERS_EXECUTE, 01);
-    permissionMap = Collections.unmodifiableMap(result);
-  }
+  private static final Map<PosixFilePermission, Integer> permissionMap =
```
coollog: type `ImmutableMap`

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -39,14 +37,9 @@
...
 
-  private static final Set<PosixFilePermission> defaultFilePermissions =
-      PermissionsHelper.toSet(0644);
-  private static final Set<PosixFilePermission> defaultFolderPermissions =
-      PermissionsHelper.toSet(0755);
-
   private final Path sourceFile;
   private final AbsoluteUnixPath extractionPath;
-  private final Set<PosixFilePermission> permissions;
+  @Nullable private final Set<PosixFilePermission> permissions;
```
coollog: We should keep `LayerEntry` immutable and make this an `ImmutableSet`

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -108,6 +96,7 @@ public AbsoluteUnixPath getExtractionPath() {
    *
    * @return the file permissions on the container
    */
+  @Nullable
   public Set<PosixFilePermission> getPermissions() {
```
coollog: I think we should just use `Optional` from now on for nullable returns.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
chanseokoh: I believe what @briandealwis said is like a map is basically a function. That is, `map.put(path, permissions)` is like `Function<...> f = key -> key == path ? permissions : ...`. Conceptually they are somewhat interchangeable, but I'm not opinionated in which one is better here. But if you have a map, wrapping it as a function is probably very easy, which is just `Function f = map::get`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -111,12 +111,16 @@ public Blob build() {
...
+
+      Set<PosixFilePermission> permissions;
+      if (layerEntry.getPermissions().isPresent()) {
+        permissions = layerEntry.getPermissions().get();
+      } else {
+        if (Files.isDirectory(layerEntry.getSourceFile())) {
+          permissions = defaultFolderPermissions;
+        } else {
+          permissions = defaultFilePermissions;
+        }
```
chanseokoh: Just saying `permissions = layerEntry.getPermissions().orElse(Files.isDirectory(...) ? defaultFolderPermissions : defaultFilePermissions)` can also work (I think), but not that it's better.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -52,10 +58,16 @@
...
    * @param extractionPath the path in the container file system corresponding to the {@code
    *     sourceFile}
+   * @param permissions the file permissions on the container. Use {@code null} to use defaults (644
+   *     for files, 755 for directories)
    */
-  public LayerEntry(Path sourceFile, AbsoluteUnixPath extractionPath) {
+  public LayerEntry(
+      Path sourceFile,
+      AbsoluteUnixPath extractionPath,
+      @Nullable Set<PosixFilePermission> permissions) {
```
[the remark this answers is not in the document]
TadCordle: Would rather just convert it to an `ImmutableSet` within the constructor instead of wrapping the parameter in an `ImmutableSet.copyOf` every time we call the constructor throughout the codebase.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -40,10 +41,10 @@
  */
 public class ReproducibleLayerBuilder {
 
-  private static final Set<PosixFilePermission> defaultFilePermissions =
-      PermissionsHelper.toSet(0644);
+  private static final ImmutableSet<PosixFilePermission> defaultFilePermissions =
+      ImmutableSet.copyOf(PermissionsHelper.toSet(0644));
```
coollog: I think `PermissionsHelper.toSet` could just return an `ImmutableSet`

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -116,9 +117,10 @@ public Blob build() {
           layerEntry
               .getPermissions()
               .orElse(
-                  Files.isDirectory(layerEntry.getSourceFile())
-                      ? defaultFolderPermissions
-                      : defaultFilePermissions);
+                  ImmutableSet.copyOf(
```
coollog: Can remove `ImmutableSet.copyOf`

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
coollog: Probably call it `toImmutableSet`

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -42,9 +42,9 @@
 public class ReproducibleLayerBuilder {
 
   private static final ImmutableSet<PosixFilePermission> defaultFilePermissions =
-      ImmutableSet.copyOf(PermissionsHelper.toSet(0644));
-  private static final Set<PosixFilePermission> defaultFolderPermissions =
-      ImmutableSet.copyOf(PermissionsHelper.toSet(0755));
+      ImmutableSet.copyOf(PermissionsHelper.toImmutableSet(0644));
```
coollog: can remove all the `copyOf`s

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -18,8 +18,8 @@
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import java.nio.file.attribute.PosixFilePermission;
-import java.util.HashSet;
 import java.util.Map.Entry;
 import java.util.Set;
```
coollog: nit: `permissionMap` to include a short javadoc like `Maps from {@link PosixFilePermission} to their intended file permissions bit.`

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -59,16 +59,16 @@ public static int toInt(Set<PosixFilePermission> permissions) {
    * @param permissions the file permission bits
    * @return the equivalent {@link PosixFilePermission} set
    */
-  public static Set<PosixFilePermission> toSet(int permissions) {
+  public static ImmutableSet<PosixFilePermission> toImmutableSet(int permissions) {
     Preconditions.checkArgument(
         permissions >= 0 && permissions <= 0777, "Permissions must be between 000 and 777 octal");
-    HashSet<PosixFilePermission> result = new HashSet<>();
+    ImmutableSet.Builder<PosixFilePermission> result = ImmutableSet.builder();
```
coollog: more descriptive name like `permissionsSetBuilder`?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,76 @@
...
+          .put(PosixFilePermission.OTHERS_EXECUTE, 01)
+          .build();
+
+  /**
+   * Converts a set of {@link PosixFilePermission} to the equivalent file permission bits.
+   *
+   * @param permissions the set of {@link PosixFilePermission}
+   * @return the equivalent file permission bits
+   */
+  public static int toInt(Set<PosixFilePermission> permissions) {
```
coollog: Hmm, I was thinking maybe this can be called `toPermissionBits` instead and s/`octalPermissions`/`permissionBits`/, and maybe the other method to be called `toPermissionSet`

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -37,6 +41,11 @@
  */
 public class ReproducibleLayerBuilder {
 
+  private static final ImmutableSet<PosixFilePermission> defaultFilePermissions =
```
coollog: can be capitalized now since its immutable

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
coollog: Also, since this is only used internally, we could also make this just take an `ImmutableSet`

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurationsTest.java
```diff
@@ -379,4 +393,63 @@ public void testAddDirectoryContents_complex() throws IOException {
...
+            .addDirectoryContents(
+                LayerType.EXTRA_FILES, extraDirectory, path -> true, basePath, permissions)
+            .build();
+
+    Assert.assertEquals(
+        Arrays.asList(
+            new LayerEntry(
+                extraDirectory.resolve("fileA"),
+                basePath.resolve("fileA"),
+                ImmutableSet.copyOf(PermissionsHelper.toImmutableSet(0123))),
```
coollog: There's a few more stray `copyOf`s

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,75 @@
...
+          .put(PosixFilePermission.OTHERS_EXECUTE, 01)
+          .build();
+
+  /**
+   * Converts a set of {@link PosixFilePermission} to the equivalent file permission bits.
+   *
+   * @param permissions the set of {@link PosixFilePermission}
+   * @return the equivalent file permission bits
+   */
+  public static int toPermissionBits(ImmutableSet<PosixFilePermission> permissions) {
```
briandealwis: This should take `Set`, not `ImmutableSet`.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -0,0 +1,75 @@
...
+    return permissionBits;
+  }
+
+  /**
+   * Converts file permission bits to an equivalent set of {@link PosixFilePermission}.
+   *
+   * @param permissions the file permission bits
+   * @return the equivalent {@link PosixFilePermission} set
+   */
+  public static ImmutableSet<PosixFilePermission> toPermissionSet(int permissions) {
```
briandealwis: IMHO this should be `Set`, and should actually return an `EnumSet`.  Wrap it with `Collections.unmodifiableSet()`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
briandealwis: It adds flexibility for the future.  I'm thinking that if we may already have a Maven or Gradle model object representing the mapping of files to permissions.  Forcing the building of a map may be unnecessary.  And it simplifies adding rules like _all .class files should always be 0444_.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -36,6 +40,7 @@
 
   private final Path sourceFile;
   private final AbsoluteUnixPath extractionPath;
+  @Nullable private final ImmutableSet<PosixFilePermission> permissions;
```
briandealwis: Record this as the integer file mode permissions — it's more compact, and saves the recreation of immutable sets, etc.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -37,6 +41,11 @@
  */
 public class ReproducibleLayerBuilder {
 
+  private static final ImmutableSet<PosixFilePermission> DEFAULT_FILE_PERMISSIONS =
```
briandealwis: Leave these as integers — there's no point continuously switching between integer representations to non-integer representations.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
TadCordle: I'll change this when I get to the plugins if need be.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
TadCordle: If we store this as an integer I think we lose the ability to differentiate between using default and explicit permissions without making additional changes (unless we use a nullable `Integer`, which seems weird?)

chanseokoh: We should be cautious with `-1`, because all the bits are set in the 2's complement binary notation, and it may well be possible that `TarArchiveEntry` only does bit manipulation.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/PermissionsHelper.java
```diff
@@ -62,13 +64,13 @@ public static int toPermissionBits(ImmutableSet<PosixFilePermission> permissions
...
         permissions >= 0 && permissions <= 0777, "Permissions must be between 000 and 777 octal");
-    ImmutableSet.Builder<PosixFilePermission> permissionsSetBuilder = ImmutableSet.builder();
+    EnumSet<PosixFilePermission> permissionsSetBuilder = EnumSet.noneOf(PosixFilePermission.class);
     for (Entry<PosixFilePermission, Integer> entry : permissionMap.entrySet()) {
       if ((permissions & entry.getValue()) != 0) {
         permissionsSetBuilder.add(entry.getKey());
       }
     }
-    return permissionsSetBuilder.build();
+    return ImmutableSet.copyOf(permissionsSetBuilder);
```
briandealwis: Why does this need to be immutable? It's not a reference that we hold on to or care about.  If the caller wants to modify it to add or remove permissions now we force them to copy it. And it should just be an `EnumSet`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
briandealwis: -1 indicates that it uses the defaults.   We're not storing TarEntry's special bits here.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -100,9 +102,23 @@ public Blob build() {
     for (LayerEntry layerEntry : layerEntries) {
       // Adds the entries to uniqueTarArchiveEntries, which makes sure all entries are unique and
       // adds parent directories for each extraction path.
-      uniqueTarArchiveEntries.add(
+      TarArchiveEntry entry =
           new TarArchiveEntry(
-              layerEntry.getSourceFile().toFile(), layerEntry.getAbsoluteExtractionPathString()));
+              layerEntry.getSourceFile().toFile(), layerEntry.getAbsoluteExtractionPathString());
+
+      FilePermissions permissions =
```
loosebazooka: I wonder if this should be in the LayerEntry. If someone does a getPermissions on a layer entry, it would make sense that this is what's returned.

We could do this directly in the `LayerEntry` constructor `if permissions == null`? and then 
`public Optional<FilePermissions> getPermissions()` -> `public FilePermissions getPermissions()`

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -52,10 +56,14 @@
    * @param sourceFile the source file to add to the layer
    * @param extractionPath the path in the container file system corresponding to the {@code
    *     sourceFile}
+   * @param permissions the file permissions on the container. Use {@code null} to use defaults (644
+   *     for files, 755 for directories)
    */
-  public LayerEntry(Path sourceFile, AbsoluteUnixPath extractionPath) {
+  public LayerEntry(
```
loosebazooka: We might want to consider in this or a follow up PR: but we seem to now have a lot of code that is LayerEntry(file, path, null) where maybe static initializers could be a nice entry point? My naming is a little off, but I dunno.
```
LayerEntry.asEntry(sourceFile, extractionPath)
LayerEntry.asEntryWithPermissions(sourceFile, extractionPath, permissions)
```
and this constructor with optional args then becomes private/package private

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -86,8 +88,32 @@ private Builder() {
      * @see LayerConfiguration.Builder#addEntry(Path, AbsoluteUnixPath)
      */
     public Builder addFile(LayerType layerType, Path sourceFile, AbsoluteUnixPath pathInContainer) {
+      addFile(layerType, sourceFile, pathInContainer, null);
```
chanseokoh: `return addFile(...)`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
chanseokoh: But those a lot of `new LayerEntry(file, path, null)` are only in testing. I vaguely remember the reason we did not add the 2-arg version was that internally there will be only a couple places for instantiating `LayerEntry`? (I may be wrong.)

TadCordle: Yeah, `LayerEntry` isn't intended to be a user-facing class, so I don't think this matters that much at the moment.

loosebazooka: Oh, well if that's the case, lets just leave this as is.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
chanseokoh: I think @TadCordle initially had that in `LayerEntry` but move it out of there to here so that you know if the permission was explicitly configured or not: https://github.com/GoogleContainerTools/jib/pull/1156#discussion_r226067367

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
TadCordle: Filed an issue in case we want to go back to it later: https://github.com/GoogleContainerTools/jib/issues/1168

PR #1159, opened 2018-10-18
Set tag to latest if version is unspecified in jibDockerBuild

### LINKED ISSUE #1096: If no version is specified, the tag should default to "latest" instead of "unspecified"
**Description of the issue**:

When running `./gradlew jibDockerBuild` the Docke rimage is always also tagged with "unspecified" even if a tag is specified via `jib.to.tags`.

**Expected behavior**:

Only the tags specified via `jib.to.tags` are used to tag the Docker image.

**Steps to reproduce**:

Run `./gradlew jibDockerBuild` with the configuration as mentioned below. See https://github.com/heremaps/oss-review-toolkit/blob/jib/cli/build.gradle to reproduce.

**Environment**:

Ubuntu 18.04

```groovy
plugins {
    id 'com.google.cloud.tools.jib' version '0.9.11'
}

jib.to.tags = ['latest']
```

**Log output**:

```
Parallel execution with configuration on demand is an incubating feature.
Tagging image with generated image reference cli:unspecified. If you'd like to specify a different tag, you can set the jib.to.image parameter in your build.gradle, or use the --image=<MY IMAGE> commandline flag.

Containerizing application to Docker daemon as cli:unspecified, cli...

Building classes layer...
Finalizing...
Loading to Docker daemon...

Container entrypoint set to [java, -cp, /app/resources:/app/classes:/app/libs/*, com.here.ort.Main]

Built image to Docker daemon as cli:unspecified, cli

BUILD SUCCESSFUL in 9s
68 actionable tasks: 1 executed, 67 up-to-date
```

TadCordle:
Fixes #1096

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
 - `jibExportDockerContext` generates different directory layout and `Dockerfile` to enable WAR support ([#1007](https://github.com/GoogleContainerTools/jib/pull/1007))
 - File timestamps in the built image are set to 1 second since the epoch (hence 1970-01-01T00:00:01Z) to resolve compatibility with applications on Java 6 or below where the epoch means nonexistent or I/O errors; previously they were set to the epoch ([#1079](https://github.com/GoogleContainerTools/jib/issues/1079))
+- Sets tag to "latest" instead of "unspecified" if `jib.to.image` and project version are both unspecified when running `jibDockerBuild` ([#1096](https://github.com/GoogleContainerTools/jib/issues/1096))
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -96,7 +96,9 @@ public void buildDocker()
             jibExtension.getTo().getImage(),
             eventDispatcher,
             getProject().getName(),
-            getProject().getVersion().toString(),
+            getProject().getVersion().toString().equals("unspecified")
+                ? "latest"
+                : getProject().getVersion().toString(),
```
chanseokoh: This works, but the argument in the Javadoc is
```
   * @param projectVersion the project version, as determined by the plugin
```
so "latest" as a `projectVersion` might not fit well (especially the Javadoc says "as determined by the plugin"). I wonder if it makes sense to either push the logic down to `getGeneratedTargetDockerTag()` or change the argument name to ... something else.

TadCordle: If the version is unspecified in Maven, you get `[ERROR] Some problems were encountered while processing the POMs: [FATAL] 'version' is missing. @ line 4, column 104`, so I decided to just make the change gradle-side.

I will change the argument name.

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 - Removed deprecated `jib.jvmFlags`, `jib.mainClass`, `jib.args`, and `jib.format` in favor of the equivalents under `jib.container` ([#461](https://github.com/GoogleContainerTools/jib/issues/461))
 - `jibExportDockerContext` generates different directory layout and `Dockerfile` to enable WAR support ([#1007](https://github.com/GoogleContainerTools/jib/pull/1007))
 - File timestamps in the built image are set to 1 second since the epoch (hence 1970-01-01T00:00:01Z) to resolve compatibility with applications on Java 6 or below where the epoch means nonexistent or I/O errors; previously they were set to the epoch ([#1079](https://github.com/GoogleContainerTools/jib/issues/1079))
+- Sets tag to "latest" instead of "unspecified" if `jib.to.image` and project version are both unspecified when running `jibDockerBuild` ([#1096](https://github.com/GoogleContainerTools/jib/issues/1096))
```
chanseokoh: nit: `jibDockerBuild` and `jibBuildTar`? And my last question: why it already works with `jib` (the registry push)?

TadCordle: `jib` requires `jib.to.image` to be set

PR #1166, opened 2018-10-19
Add FilePermissions class

### LINKED ISSUE #1165: Create a Permissions class
To make #1156 less of a mess, create a `Permissions` class to allow for
- Type safety - can be parsed from a `String` instead of an `int` to avoid octal vs. decimal confusion
- Compactness - internal representation can still be an `int`
- Ease of use - relevant methods on this class instead of using a helper and doing multiple back-and-forth conversions
- Extendability - should be easy to add operations if/as we need them

Right now we only really need two methods:
```
/* Parse/validate an octal String (e.g. "755"). */
static Permissions fromString(String octalString);

/* Set the permissions on a TarArchiveEntry. */
void setOnTarArchiveEntry(TarArchiveEntry entry);
```

but we can decide to add stuff depending on how people want to use the library.
```
static Permissions fromSet(Set<PosixFilePermission> octalString);

Permissions addPermission(PosixFilePermission permission);

Set<PosixFilePermission> getPermissionSet();

int getPermissionBits();

// etc.
```

TadCordle:
Fixes #1165 

Adds a `FilePermissions` class that can be instantiated from either an octal string or a set of `PosixFilePermissions`. Also has a getter for the file permission bits.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/FilePermissions.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/FilePermissions.java
@@ -22,14 +22,17 @@ import com.google.common.collect.ImmutableMap;
 import java.nio.file.attribute.PosixFilePermission;
 import java.util.Set;
 
-/** Represents read/write/execute file permissions. */
+/** Represents read/write/execute file permissions for owner, group, and others. */
 public class FilePermissions {
 
-  /** Matches an octal string representation of file permissions. */
-  private static final String octalPattern = "[0-7][0-7][0-7]";
+  /**
+   * Matches an octal string representation of file permissions. From left to right, each digit
+   * represents permissions for owner, group, and other.
+   */
+  private static final String OCTAL_PATTERN = "[0-7][0-7][0-7]";
 
   /** Maps from a {@link PosixFilePermission} to its corresponding file permission bit. */
-  private static final ImmutableMap<PosixFilePermission, Integer> permissionMap =
+  private static final ImmutableMap<PosixFilePermission, Integer> PERMISSION_MAP =
       ImmutableMap.<PosixFilePermission, Integer>builder()
           .put(PosixFilePermission.OWNER_READ, 0400)
           .put(PosixFilePermission.OWNER_WRITE, 0200)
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/FilePermissions.java
```diff
@@ -0,0 +1,108 @@
...
+
+package com.google.cloud.tools.jib.configuration;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.Set;
+
+/** Represents read/write/execute file permissions. */
```
coollog: ... for owner, group and others...

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/FilePermissions.java
```diff
@@ -0,0 +1,108 @@
...
+public class FilePermissions {
+
+  /** Matches an octal string representation of file permissions. */
```
coollog: nit: just mention which digit refers to which

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/FilePermissions.java
```diff
@@ -0,0 +1,106 @@
...
+  private static final String octalPattern = "[0-7][0-7][0-7]";
+
+  /** Maps from a {@link PosixFilePermission} to its corresponding file permission bit. */
+  private static final ImmutableMap<PosixFilePermission, Integer> permissionMap =
```
coollog: nit: capitalize for constant (and above)

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/FilePermissions.java
```diff
@@ -0,0 +1,106 @@
...
+    if (!(other instanceof FilePermissions)) {
+      return false;
+    }
+    FilePermissions otherFilePermissions = (FilePermissions) other;
+    return permissionBits == otherFilePermissions.permissionBits;
+  }
+
+  @Override
+  public int hashCode() {
+    return Integer.hashCode(permissionBits);
```
coollog: this could also just be `permissionBits`

PR #1171, opened 2018-10-19
Make CacheEntry extend Layer

### LINKED ISSUE #1055: Consider having CacheEntry implement Layer.
From https://github.com/GoogleContainerTools/jib/pull/1047#discussion_r221054886

This way, `BuildImageStep` would not need a `cacheEntryToLayer` method.

TadCordle:
Made an attempt to fix #1055 

The issue now is that by extending `Layer` and using its methods, we lose a bit of context with the method names (e.g. `getLayerBlob` is now `getBlob`, `getLayerDiffId` is now `getDiffId`). So I think we should rename `CacheEntry` to something with "layer" in the name (like `CachedLayer`, or `CacheLayerEntry` or something). Or maybe we don't even need `CacheEntry` anymore since it's identical to `Layer` now. WDYT?

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
@@ -111,7 +111,8 @@ class BuildAndCacheApplicationLayerStep implements AsyncStep<CacheEntry>, Callab
 
       buildConfiguration
           .getEventDispatcher()
-          .dispatch(LogEvent.debug(description + " built " + cacheEntry.getLayerDigest()));
+          .dispatch(
+              LogEvent.debug(description + " built " + cacheEntry.getBlobDescriptor().getDigest()));
 
       return cacheEntry;
     }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
@@ -99,10 +99,10 @@ class BuildAndCacheApplicationLayerStep implements AsyncStep<CachedLayer>, Calla
       Cache cache = buildConfiguration.getApplicationLayersCache();
 
       // Don't build the layer if it exists already.
-      Optional<CachedLayer> optionalCacheEntry =
+      Optional<CachedLayer> optionalCachedLayer =
           cache.retrieve(layerConfiguration.getLayerEntries());
-      if (optionalCacheEntry.isPresent()) {
-        return optionalCacheEntry.get();
+      if (optionalCachedLayer.isPresent()) {
+        return optionalCachedLayer.get();
       }
 
       Blob layerBlob = new ReproducibleLayerBuilder(layerConfiguration.getLayerEntries()).build();
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
@@ -111,9 +111,7 @@ class BuildAndCacheApplicationLayerStep implements AsyncStep<CachedLayer>, Calla
 
       buildConfiguration
           .getEventDispatcher()
-          .dispatch(
-              LogEvent.debug(
-                  description + " built " + cachedLayer.getBlobDescriptor().getDigest()));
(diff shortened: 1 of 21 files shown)


coollog:
Yea, renaming sounds good. Maybe even LayerInCache? Or even just CachedLayer like we had before. 

I also think we should have digest and size separate and not in a getBlobDescriptor since most of the time they are used separately and we end up calling getBlobDescriptor().getDigest() everywhere.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
```diff
@@ -99,22 +99,21 @@ public CacheEntry call() throws IOException, CacheCorruptedException {
       Cache cache = buildConfiguration.getApplicationLayersCache();
 
       // Don't build the layer if it exists already.
-      Optional<CacheEntry> optionalCacheEntry =
+      Optional<CachedLayer> optionalCacheEntry =
```
coollog: nit: optionalCachedLayer

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
```diff
@@ -56,19 +56,19 @@
...
-  public CacheEntry call() throws IOException, CacheCorruptedException {
+  public CachedLayer call() throws IOException, CacheCorruptedException {
     try (TimerEventDispatcher ignored =
         new TimerEventDispatcher(
             buildConfiguration.getEventDispatcher(), String.format(DESCRIPTION, layerDigest))) {
       Cache cache = buildConfiguration.getBaseImageLayersCache();
 
       // Checks if the layer already exists in the cache.
-      Optional<CacheEntry> optionalCacheEntry = cache.retrieve(layerDigest);
+      Optional<CachedLayer> optionalCacheEntry = cache.retrieve(layerDigest);
```
coollog: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
```diff
@@ -83,14 +83,14 @@
     }
   }
 
-  private PushBlobStep makePushBlobStep(AsyncStep<? extends CacheEntry> cacheEntryStep)
+  private PushBlobStep makePushBlobStep(AsyncStep<? extends CachedLayer> cacheEntryStep)
```
coollog: cachedLayerStep

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCachedLayer.java
```diff
@@ -68,44 +69,47 @@ CacheEntry build() {
...
   }
 
   @Override
-  public Blob getLayerBlob() {
+  public Blob getBlob() {
     return layerBlob;
   }
+
+  @Override
+  public BlobDescriptor getBlobDescriptor() {
```
coollog: Hmm, can we remove this from Layer interface?

jib-core/src/test/java/com/google/cloud/tools/jib/cache/DefaultCachedLayerTest.java
```diff
@@ -65,18 +65,18 @@ public void testBuilder_fail() {
...
             .setLayerSize(1337);
     Assert.assertFalse(cacheEntryBuilder.hasLayerBlob());
     cacheEntryBuilder.setLayerBlob(Blobs.from("layerBlob"));
     Assert.assertTrue(cacheEntryBuilder.hasLayerBlob());
-    CacheEntry cacheEntry = cacheEntryBuilder.build();
-    Assert.assertEquals(mockLayerDigest, cacheEntry.getLayerDigest());
-    Assert.assertEquals(mockLayerDiffId, cacheEntry.getLayerDiffId());
-    Assert.assertEquals(1337, cacheEntry.getLayerSize());
-    Assert.assertEquals("layerBlob", Blobs.writeToString(cacheEntry.getLayerBlob()));
+    CachedLayer cachedLayer = cacheEntryBuilder.build();
```
coollog: cachedLayerBuilder

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCachedLayer.java
TadCordle: I tried and it looks like there may end up being lots of changes, maybe not worth it for now.

coollog: Okay let's add a TODO to remove this later then.

Issue #468, opened 2018-06-28
specifying target docker daemon for jibDockerBuild 


gintautassulskus:
Section [Build to Docker daemon](https://github.com/GoogleContainerTools/jib/tree/v0.9.1-gradle/jib-gradle-plugin#build-to-docker-daemon) provides an example on how to build and push a docker image to minikube docker daemon:

```
eval $(minikube docker-env)
gradle jibDockerBuild
```

Is there a gradle-native way to configure target docker daemon for the jib plugin?
I have multiple minikube instances running at the same time and must use minikube `--profile` option to retrieve specific minikube's docker-envs.

gintautassulskus:
At first glance, the variable approach seems a bit neater, but the support for environmnt variable ingestion is more future-proof and flexible.
Whatever the decision, the end solution should integrate well with the [minikube plugin](https://github.com/GoogleContainerTools/minikube-build-tools-for-java).

I have opened an [issue](https://github.com/GoogleContainerTools/minikube-build-tools-for-java/issues/141) which I believe is somewaht relevant to this one.

coollog:
Since it might be confusing to have `environment` on the `jib` extension (which configures the entire build), we may have the `jibDockerBuild` task configurable with environment variables, so it would look something like:

```
tasks.jibDockerBuild.environment 'NAME', 'VALUE'
```

Similarly, for Maven, only the `jib:dockerBuild` goal will be configurable with `dockerBuildEnvironment`.

gintautassulskus:
I am thinking of adding `jib.dockerClient` to configure docker executable specific parameters, such as environment and path.

gintautassulskus:
In response to https://github.com/GoogleContainerTools/jib/pull/1173#issuecomment-431891842

We have a number of multi-project gradle builds that require separate development enviromments. Hence the need for Minikube profiles.

If the following PR gets accepted (https://github.com/GoogleContainerTools/minikube-build-tools-for-java/pull/146), then the push of docker images would look something like:

```
    jibDockerBuild {
        doLast {
            dockerExecutable = "/usr/local/bin/docker"
            dockerEnvironment = minikube.getDockerEnv(projectID)
        }
    }
```
Where `minikube.getDockerEnv(profileID)` retrieves docker ENV configuration from the specified minikube instance and configures the `jibDockerBuild` task to deploy the image to the very same minikube instance.

[...]

PR #1173, opened 2018-10-21
added executable and environment parameters for jibDockerBuild

gintautassulskus:
Addresses https://github.com/GoogleContainerTools/jib/issues/468

jibDockerBuild can now be configured with two optional properties:
```
jibDockerBuild {
    dockerExecutable = "/optional/path/to/docker/"

    def envVars = [:]
        envVars["DOCKER_TLS_VERIFY"] = "1"
        envVars["DOCKER_HOST"] = "tcp://IP:PORT"
        envVars["DOCKER_CERT_PATH"] = "/path/to/.minikube/certs"
        envVars["DOCKER_API_VERSION"] = "1.35"
    dockerEnvironment = envVars
}
```

Also, reformatted source code to follow the suggested [Oracle conventions](https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html) (4-space indentations). Hence so many changed lines. Now not sure any more whether this was a good idea.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
@@ -23,10 +23,6 @@ import com.google.cloud.tools.jib.configuration.ImageConfiguration;
 import com.google.cloud.tools.jib.docker.DockerClient;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.InvalidImageReferenceException;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.Collections;
-import java.util.Map;
 
 /** Builds to the Docker daemon. */
 // TODO: Add tests once JibContainerBuilder#containerize() is added.
@@ -57,23 +53,12 @@ public class DockerDaemonImage implements TargetImage {
   }
 
   private final ImageReference imageReference;
-  private Path dockerExecutable = Paths.get("docker");
-  private Map<String, String> dockerEnvironment = Collections.emptyMap();
+
+  private DockerClient dockerClient = DockerClient.newClient();
 
   /** Instantiate with {@link #named}. */
   private DockerDaemonImage(ImageReference imageReference) {
     this.imageReference = imageReference;
-  }
-
-  /**
-   * Sets the path to the {@code docker} CLI. This is {@code docker} by default.
-   *
-   * @param dockerExecutable the path to the {@code docker} CLI
-   * @return this
-   */
-  public DockerDaemonImage setDockerExecutable(Path dockerExecutable) {
-    this.dockerExecutable = dockerExecutable;
-    return this;
   }
 
   @Override
@@ -83,34 +68,15 @@ public class DockerDaemonImage implements TargetImage {
 
   @Override
   public BuildSteps toBuildSteps(BuildConfiguration buildConfiguration) {
-    return BuildSteps.forBuildToDockerDaemon(
-        DockerClient.newClient(dockerExecutable, dockerEnvironment), buildConfiguration);
+    return BuildSteps.forBuildToDockerDaemon(dockerClient, buildConfiguration);
   }
 
-  /**
-   * Gets the path to the {@code docker} CLI.
-   *
-   * @return the path to the {@code docker} CLI
-   */
-  Path getDockerExecutable() {
-    return dockerExecutable;
+  public DockerClient getDockerClient() {
+    return dockerClient;
   }
 
-  /**
-   * Gets environment variables for the {@code docker} CLI.
-   *
-   * @return the path to the {@code docker} CLI
-   */
-  public Map<String, String> getDockerEnvironment() {
-    return dockerEnvironment;
-  }
-
-  /**
-   * Sets environment variables for the {@code docker} CLI.
-   *
-   * @param dockerEnvironment a map of docker environment variables
-   */
-  public void setDockerEnvironment(Map<String, String> dockerEnvironment) {
-    this.dockerEnvironment = dockerEnvironment;
+  public DockerDaemonImage setDockerClient(DockerClient dockerClient) {
+    this.dockerClient = dockerClient;
+    return this;
   }
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
@@ -23,6 +23,7 @@ import com.google.cloud.tools.jib.configuration.ImageConfiguration;
 import com.google.cloud.tools.jib.docker.DockerClient;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.InvalidImageReferenceException;
+import com.google.common.annotations.VisibleForTesting;
 
 /** Builds to the Docker daemon. */
 // TODO: Add tests once JibContainerBuilder#containerize() is added.
@@ -71,6 +72,7 @@ public class DockerDaemonImage implements TargetImage {
     return BuildSteps.forBuildToDockerDaemon(dockerClient, buildConfiguration);
   }
 
+  @VisibleForTesting
   public DockerClient getDockerClient() {
     return dockerClient;
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
@@ -29,7 +29,9 @@ import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Map;
 
-/** Builds to the Docker daemon. */
+/**
+ * Builds to the Docker daemon.
+ */
 // TODO: Add tests once JibContainerBuilder#containerize() is added.
 public class DockerDaemonImage implements TargetImage {
 
@@ -50,7 +52,8 @@ public class DockerDaemonImage implements TargetImage {
    *
    * @param imageReference the image reference
    * @return a new {@link DockerDaemonImage}
-   * @throws InvalidImageReferenceException if {@code imageReference} is not a valid image reference
+   * @throws InvalidImageReferenceException if {@code imageReference} is not a valid image
+   * reference
    */
   public static DockerDaemonImage named(String imageReference)
       throws InvalidImageReferenceException {
@@ -61,7 +64,9 @@ public class DockerDaemonImage implements TargetImage {
   private Path dockerExecutable = Paths.get("docker");
   private Map<String, String> dockerEnvironment = Collections.emptyMap();
 
-  /** Instantiate with {@link #named}. */
+  /**
+   * Instantiate with {@link #named}.
+   */
   private DockerDaemonImage(ImageReference imageReference) {
     this.imageReference = imageReference;
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
@@ -23,8 +23,11 @@ import com.google.cloud.tools.jib.configuration.ImageConfiguration;
 import com.google.cloud.tools.jib.docker.DockerClient;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.InvalidImageReferenceException;
+
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.Map;
 
 /** Builds to the Docker daemon. */
 // TODO: Add tests once JibContainerBuilder#containerize() is added.
@@ -56,6 +59,7 @@ public class DockerDaemonImage implements TargetImage {
 
   private final ImageReference imageReference;
   private Path dockerExecutable = Paths.get("docker");
+  private Map<String, String> dockerEnvironment = Collections.emptyMap();
 
   /** Instantiate with {@link #named}. */
   private DockerDaemonImage(ImageReference imageReference) {
@@ -81,7 +85,7 @@ public class DockerDaemonImage implements TargetImage {
   @Override
   public BuildSteps toBuildSteps(BuildConfiguration buildConfiguration) {
     return BuildSteps.forBuildToDockerDaemon(
-        DockerClient.newClient(dockerExecutable), buildConfiguration);
+        DockerClient.newClient(dockerExecutable, dockerEnvironment), buildConfiguration);
   }
 
   /**
@@ -91,5 +95,23 @@ public class DockerDaemonImage implements TargetImage {
    */
   Path getDockerExecutable() {
     return dockerExecutable;
+  }
+
+  /**
+   * Gets environment variables for the {@code docker} CLI.
+   *
+   * @return the path to the {@code docker} CLI
+   */
+  public Map<String, String> getDockerEnvironment() {
+    return dockerEnvironment;
+  }
+
+  /**
+   * Sets environment variables for the {@code docker} CLI.
+   *
+   * @return the path to the {@code docker} CLI
+   */
+  public void setDockerEnvironment(Map<String, String> dockerEnvironment) {
+    this.dockerEnvironment = dockerEnvironment;
   }
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
@@ -23,15 +23,12 @@ import com.google.cloud.tools.jib.configuration.ImageConfiguration;
 import com.google.cloud.tools.jib.docker.DockerClient;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.InvalidImageReferenceException;
-
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Map;
 
-/**
- * Builds to the Docker daemon.
- */
+/** Builds to the Docker daemon. */
 // TODO: Add tests once JibContainerBuilder#containerize() is added.
 public class DockerDaemonImage implements TargetImage {
 
@@ -52,8 +49,7 @@ public class DockerDaemonImage implements TargetImage {
    *
    * @param imageReference the image reference
    * @return a new {@link DockerDaemonImage}
-   * @throws InvalidImageReferenceException if {@code imageReference} is not a valid image
-   * reference
+   * @throws InvalidImageReferenceException if {@code imageReference} is not a valid image reference
    */
   public static DockerDaemonImage named(String imageReference)
       throws InvalidImageReferenceException {
@@ -64,9 +60,7 @@ public class DockerDaemonImage implements TargetImage {
   private Path dockerExecutable = Paths.get("docker");
   private Map<String, String> dockerEnvironment = Collections.emptyMap();
 
-  /**
-   * Instantiate with {@link #named}.
-   */
+  /** Instantiate with {@link #named}. */
   private DockerDaemonImage(ImageReference imageReference) {
     this.imageReference = imageReference;
   }
@@ -115,7 +109,6 @@ public class DockerDaemonImage implements TargetImage {
    * Sets environment variables for the {@code docker} CLI.
    *
    * @param dockerEnvironment a map of docker environment variables
-   * 
    */
   public void setDockerEnvironment(Map<String, String> dockerEnvironment) {
     this.dockerEnvironment = dockerEnvironment;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
@@ -114,7 +114,8 @@ public class DockerDaemonImage implements TargetImage {
   /**
    * Sets environment variables for the {@code docker} CLI.
    *
-   * @return the path to the {@code docker} CLI
+   * @param dockerEnvironment a map of docker environment variables
+   * 
    */
   public void setDockerEnvironment(Map<String, String> dockerEnvironment) {
     this.dockerEnvironment = dockerEnvironment;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
@@ -25,7 +25,11 @@ import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
 import java.util.function.Function;
 
 /** Calls out to the {@code docker} CLI. */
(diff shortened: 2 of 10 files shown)


briandealwis:
> Also, reformatted source code to follow the suggested Oracle conventions (4-space indentations). Hence so many changed lines. Now not sure any more whether this was a good idea.

It's not documented in the CONTRIBUTING.md, but we use the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).

gintautassulskus:
@briandealwis, my bad. Addressed. 
The update should come up shortly.

briandealwis:
Thanks @gintautassulskus — if you run `./build.sh`, it will reformat and run basic tests.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -20,28 +20,31 @@
...
 import java.util.function.Function;
 
-/** Calls out to the {@code docker} CLI. */
+/**
+ * Calls out to the {@code docker} CLI.
+ */
 public class DockerClient {
 
   private static final String DEFAULT_DOCKER_CLIENT = "docker";
+  private static final Map<String, String> DEFAULT_DOCKER_ENVIRONMENT = Collections.emptyMap();
```
briandealwis: So this results in a change in behaviour as previously the docker command would have inherited the current environment.  This may actually break existing users who are using `minikube docker-env`.  I this should should default to `null` and the `defaultProcessBuilderFactory` should only set the environment if non-null.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
```diff
@@ -56,8 +62,11 @@ public static DockerDaemonImage named(String imageReference)
 
   private final ImageReference imageReference;
   private Path dockerExecutable = Paths.get("docker");
+  private Map<String, String> dockerEnvironment = Collections.emptyMap();
```
briandealwis: As I mention in the `DockerClient` change, this is a change in behaviour as it will cause the environment to be replaced.

briandealwis:
It would be helpful to add a test to `DockerClientTest` too.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
gintautassulskus: The default `DEFAULT_DOCKER_ENVIRONMENT` value was not there before. If no environment variable is passed, this is then used as a default one. Then, I simply add `dockerEnvironment` to the map of inherited ENV variables:
```
      ProcessBuilder pb = new ProcessBuilder(dockerCommand);

      Map<String, String> env = pb.environment();
      env.putAll(dockerEnvironment);
```

Therefore, I do not think this will introduce any breaking changes. 

You could argue, however, whether there is a point of passing an empty list as a default value. In this context, I kept consistent with the `dockerExecutable` configuration.

Ideally, I think, `DockerDaemonImage` should have no configuration defaults and call `newClient()` or `newClient(null, null)`. `DockerClient` would then check the parameters and would set defaults in case of `null`'s.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
gintautassulskus: Same as above. In my opinion, `dockerExecutable` and `dockerEnvironment` should be removed from this class and left in `DockerClient` in order to eliminate duplication.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
briandealwis: > Therefore, I do not think this will introduce any breaking changes.

You're right — I misread. My apologies.

briandealwis:
@gintautassulskus it would be helpful for the team to better understand your context.  Would you mind describing how this change and your minikube PR fit into your processes — perhaps on #468?  Have you looked at [Skaffold](https://github.com/GoogleContainerTools/skaffold)?

gintautassulskus:
> It would be helpful to add a test to `DockerClientTest` too.

Please check the latest commit. The most straightforward approach to test DockerClient that I have come up requires exposing the `processBuilderFactory`:
```
  @VisibleForTesting
  Function<List<String>, ProcessBuilder> getProcessBuilderFactory() {
    return processBuilderFactory;
  }
```
Let me know if this is acceptable. I have noticed a similar approach throughout the project.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
gintautassulskus: No prob. Thanks for the feedback!

gintautassulskus:
> @gintautassulskus it would be helpful for the team to better understand your context. Would you mind describing how this change and your minikube PR fit into your processes — perhaps on #468? Have you looked at [Skaffold](https://github.com/GoogleContainerTools/skaffold)?

Thanks, @briandealwis, Skaffold is something new to me. Sounds very promising.

chanseokoh:
@gintautassulskus thanks for putting up this. Our team had an internal discussion on this topic of how best this feature can be supported. We will finalize the design after doing some more discussions and research, but it may take some time as our team members are busy with various events. Please be patient, and we will get back to you later.

chanseokoh:
We had an internal discussion, and basically, the configurable-environment-variable approach makes sense. We just decided to use a slightly different config names:

```
jibDockerBuild.dockerClient.executable = ...
jibDockerBuild.dockerClient.environment = [key, 'value', ...]
```

It'd be nice if you can update the `CHANGLOG.md` in this PR too, but we can do it too if that slows you down.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -25,23 +25,23 @@
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
+import java.util.*;
```
chanseokoh: Should be expanded per our style guide.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -62,12 +76,18 @@ public static DockerClient newClient(Path dockerExecutable) {
...
   private static Function<List<String>, ProcessBuilder> defaultProcessBuilderFactory(
-      String dockerExecutable) {
+      String dockerExecutable, Map<String, String> dockerEnvironment) {
     return dockerSubCommand -> {
       List<String> dockerCommand = new ArrayList<>(1 + dockerSubCommand.size());
       dockerCommand.add(dockerExecutable);
       dockerCommand.addAll(dockerSubCommand);
-      return new ProcessBuilder(dockerCommand);
+
+      ProcessBuilder pb = new ProcessBuilder(dockerCommand);
```
chanseokoh: nit: let's say `ProcessBuilder processBuilder = new ...`.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -62,12 +76,18 @@ public static DockerClient newClient(Path dockerExecutable) {
...
+
+      Map<String, String> env = pb.environment();
```
chanseokoh: nit: `... environment = ...`

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.jib.docker;
 
+import static org.mockito.Mockito.when;
```
chanseokoh: nit: in our repo, we just do `Mockito.when(...)` without static import.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -176,11 +181,37 @@ public void testTag() throws InterruptedException, IOException, InvalidImageRefe
...
             });
-    Mockito.when(mockProcess.waitFor()).thenReturn(0);
+    when(mockProcess.waitFor()).thenReturn(0);
 
     testDockerClient.tag(ImageReference.of(null, "original", null), ImageReference.parse("new"));
   }
 
+  @Test
+  public void testDefaultClient() {
+    DockerClient dc = DockerClient.newClient();
```
chanseokoh: nit: `DockerClient client = ...`

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -176,11 +181,37 @@ public void testTag() throws InterruptedException, IOException, InvalidImageRefe
...
+    Assert.assertEquals(
+        dc.getProcessBuilderFactory().apply(Collections.emptyList()).command(),
+        Collections.singletonList("docker"));
```
chanseokoh: In our repo, we reverse the order so that `assertEquals(expected, actual)`.

jib-gradle-plugin/README.md
```diff
@@ -108,6 +108,21 @@ eval $(minikube docker-env)
 gradle jibDockerBuild
 ```
 
+Alternatively, the environment can be configured in the task with two optional parameters:
```
chanseokoh: We don't update `README` until we actually make a release, because it may confuse the users of the current latest version. Should be removed here. It'd be nice to put up a separate PR.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -135,4 +148,24 @@ public BuildDockerTask setJibExtension(JibExtension jibExtension) {
     this.jibExtension = jibExtension;
     return this;
   }
+
+  @Nullable
+  public Map<String, String> getDockerEnvironment() {
```
chanseokoh: Is this method necessary?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -135,4 +148,24 @@ public BuildDockerTask setJibExtension(JibExtension jibExtension) {
...
+    return dockerEnvironment;
+  }
+
+  @Input
+  public void setDockerEnvironment(@Nullable Map<String, String> dockerEnvironment) {
+    this.dockerEnvironment = dockerEnvironment;
+  }
+
+  @Nullable
+  public Path getDockerExecutable() {
```
chanseokoh: Is this method necessary?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -135,4 +148,24 @@ public BuildDockerTask setJibExtension(JibExtension jibExtension) {
...
     return this;
   }
+
+  @Nullable
+  public Map<String, String> getDockerEnvironment() {
+    return dockerEnvironment;
+  }
+
+  @Input
+  public void setDockerEnvironment(@Nullable Map<String, String> dockerEnvironment) {
```
chanseokoh: Just a reminder: as I said, we will change the config name/structure.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
chanseokoh: Yeah, I think it makes sense to remove both `setDockerExecutable` and `setDockerEnvironment` and instead have `setDockerClient()`.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -79,6 +99,11 @@ public static DockerClient newClient(Path dockerExecutable) {
     this.processBuilderFactory = processBuilderFactory;
   }
 
+  @VisibleForTesting
+  Function<List<String>, ProcessBuilder> getProcessBuilderFactory() {
```
chanseokoh: nit: non-public methods go below public in code.

BTW, instead of having this new getter, is it possible to make `defaultProcessBuilderFactory()` `@VisibleForTesting`?

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -176,11 +181,37 @@ public void testTag() throws InterruptedException, IOException, InvalidImageRefe
...
+  @Test
+  public void testCustomClient() {
+    Path path = Paths.get("/path.to/docker");
+    Map<String, String> environment = new HashMap<String, String>();
+    environment.putAll(System.getenv());
+    environment.putAll(Collections.singletonMap("Key1", "Value1"));
+
+    DockerClient dc = DockerClient.newClient(path, environment);
+    Assert.assertEquals(
+        dc.getProcessBuilderFactory().apply(Collections.emptyList()).command(),
```
chanseokoh: nit: the `dc` name, and the order of assert arguments.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -176,11 +181,37 @@ public void testTag() throws InterruptedException, IOException, InvalidImageRefe
...
+        dc.getProcessBuilderFactory().apply(Collections.emptyList()).environment(),
+        System.getenv());
+  }
+
+  @Test
+  public void testCustomClient() {
+    Path path = Paths.get("/path.to/docker");
+    Map<String, String> environment = new HashMap<String, String>();
+    environment.putAll(System.getenv());
+    environment.putAll(Collections.singletonMap("Key1", "Value1"));
```
chanseokoh: Why not just `put("Key1", "Value1")`?

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
```diff
@@ -176,11 +181,37 @@ public void testTag() throws InterruptedException, IOException, InvalidImageRefe
...
+        Collections.singletonList("docker"));
+    Assert.assertEquals(
+        dc.getProcessBuilderFactory().apply(Collections.emptyList()).environment(),
+        System.getenv());
+  }
+
+  @Test
+  public void testCustomClient() {
+    Path path = Paths.get("/path.to/docker");
+    Map<String, String> environment = new HashMap<String, String>();
```
chanseokoh: `... = HashMap<>(System.getenv())`?

At least remove `HashMap<String, String>()` --> `HashMap<>()`.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
gintautassulskus: expanded

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
gintautassulskus: addressed

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
gintautassulskus: addressed

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
gintautassulskus: addressed

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
gintautassulskus: addressed

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
gintautassulskus: addressed

jib-gradle-plugin/README.md
gintautassulskus: will do

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
gintautassulskus: nope

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
gintautassulskus: nope

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
gintautassulskus: Yes, but I do not think that this is a good solution. 

We want to test the actual use case, which is creating the process builder using one of the defined `newClient(*)` methods. Then, however, the only way to access the private process builder object is using the `getProcessBuilderFactory` getter.

If we are going to create the object directly using `defaultProcessBuilderFactory()`, we will not test the whole execution path.

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
gintautassulskus: addressed

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
gintautassulskus: addressed

jib-core/src/test/java/com/google/cloud/tools/jib/docker/DockerClientTest.java
gintautassulskus: addressed

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
chanseokoh: What you said makes sense, but I think this is basically a matter of the scope of a unit test, like a small unit test vs. medium or large unit test (more toward an end-to-end integration test). It's OK and meaningful to have narrowly scoped unit tests that test only one method, and if it's possible, it's also meaningful to have another larger tests that cover a  longer execution path at a higher level. I just feel it doesn't seem worth adding a new method (especially a simple getter) just to enable a test. After all, what you get from the getter is just an object built once from `defaultProcessBuilderFactory` at construction.

chanseokoh: To add, I'm trying to avoid spreading the practice like, "oh, I cannot access these fields in tests, but I want to directly examine the values in tests. I'll just make them visible by adding getters." It feels like, you may be more focused on testing internal implementation or design that is meant to be encapsulated.

jib-maven-plugin/src/test/resources/projects/war_servlet25/src/main/java/example/HelloWorld.java
```diff
@@ -16,16 +16,16 @@
 
 package example;
 
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
```
chanseokoh: nit: unnecessary changes in test data (and not in the Google style, actually).

jib-maven-plugin/src/test/resources/projects/simple/src/main/java/com/test/HelloWorld.java
```diff
@@ -17,6 +17,7 @@
 package com.test;
 
 import dependency.Greeting;
+
```
chanseokoh: nit: unnecessary changes in test data (and not in the Google style, actually).

jib-maven-plugin/src/test/resources/projects/multi/lib/src/test/java/com/lib/LibTest.java
```diff
@@ -16,8 +16,6 @@
 
 package com.lib;
 
-import static org.junit.Assert.assertTrue;
```
chanseokoh: nit: unnecessary changes in test data (and not in the Google style, actually).

jib-maven-plugin/src/test/resources/projects/default-target/src/main/java/com/test/HelloWorld.java
```diff
@@ -17,6 +17,7 @@
 package com.test;
 
 import dependency.Greeting;
+
```
chanseokoh: nit: unnecessary changes in test data (and not in the Google style, actually).

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
```diff
@@ -55,41 +53,30 @@ public static DockerDaemonImage named(String imageReference)
...
   }
 
-  /**
-   * Gets the path to the {@code docker} CLI.
-   *
-   * @return the path to the {@code docker} CLI
-   */
-  Path getDockerExecutable() {
-    return dockerExecutable;
+  public DockerClient getDockerClient() {
```
chanseokoh: Looks like this exists only for testing. It'd be better to avoid this, but if you think you can't, at least add `@VisibleForTesting` for indication (even though it is already public).

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
```diff
@@ -55,41 +53,30 @@ public static DockerDaemonImage named(String imageReference)
...
+    return dockerClient;
+  }
+
+  public DockerDaemonImage setDockerClient(DockerClient dockerClient) {
```
chanseokoh: This is a public-facing jib-core API, so let's add a Javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
gintautassulskus: I can use `defaultProcessBuilderFactory` but then the last assertion in `DockerDaemonImageTest` is no longer feasible.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
gintautassulskus: For now, I left it with @ VisibleForTesting

gintautassulskus:
I replaced `setDockerExecutable` and `setDockerEnvironment` with `setDockerClient()` but now the user must set it like this:
`jibDockerBuild.dockerClient = com.google.cloud.tools.jib.docker.DockerClient.newClient()`

This needs to be changed to a closure. Until then, it's better not to merge this PR.

Unfortunately, I do not have time to work on this at the moment.

chanseokoh:
> I replaced setDockerExecutable and setDockerEnvironment with setDockerClient() but now the user must set it like this:
> jibDockerBuild.dockerClient = com.google.cloud.tools.jib.docker.DockerClient.newClient()

I think we can retain `setDockerExecutable` and `setDockerEnvironment` in `DockerBuildTask`.

But anyway, now you said it, I just realized that `DockerClient` is not a public user-facing API (i.e., not under the `.api` package). Sorry for the turnaround. I should have been more careful.

And since you said you'll not have time for this, I think we can take over from now.

Thanks again for your great contribution. Much appreciated.

gintautassulskus:
No prob. Thanks for your patience and attention.

chanseokoh:
@GoogleContainerTools/java-tools I'll create a branch from this and put up a PR.

chanseokoh:
@gintautassulskus we've merged #1214, which will make the Docker client parameters available in the next 0.10.0 release. Thanks a ton for your contribution!

gintautassulskus:
Thanks for taking over and finishing it.

PR #1174, opened 2018-10-21
maven-surefire-plugin 2.22.1

sullis:

### DIFF
--- a/jib-maven-plugin/pom.xml
+++ b/jib-maven-plugin/pom.xml
@@ -147,7 +147,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-failsafe-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


sullis:
issue:  https://github.com/GoogleContainerTools/jib/issues/1175

briandealwis:
Thanks @sullis — did you have a particular reason to update surefire?  Would you mind updating `maven-failsafe-plugin` too?

loosebazooka:
@briandealwis should we merge this?

TadCordle:
I think we might as well. [Patch notes here](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12317927&version=12343425) in case anyone wants to double check.

briandealwis:
I think we should, but with the failsafe plugin too.  No point letting them diverge.

PR #1178, opened 2018-10-22
Update gradle extraDirectory config parameter to include path and permissions field

TadCordle:
Toward #794 

`extraDirectory` configuration parameter is now an object with a `path` and `permissions` field, instead of just being a file
  - `extraDirectory.path` covers the original purpose of `extraDirectory`
  - `extraDirectory.permissions` is a map from path on container to an octal string representation of the file's permission bits

For example, if your extra directory is `(project dir)/src/main/custom-jib`, and you want to give executable permissions to `path/to/file1` and `path/to/file2` within, you would add this configuration to your `build.gradle`:

```groovy
jib {
  ...
  extraDirectory {
    path = file('src/main/custom-jib')
    permissions = [
      '/path/to/file1': '755',
      '/path/to/file2': '755'
    ]
  }
}
```

The old method of configuring the extra directory is also still supported:
```groovy
jib {
  ...
  extraDirectory = file('src/main/custom-jib')
}
```

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
@@ -30,12 +30,12 @@ import org.gradle.api.tasks.Internal;
 /** Object in {@link JibExtension} that configures the extra directory. */
 public class ExtraDirectoryParameters {
 
-  private Path path;
-  private Map<String, String> permissions = Collections.emptyMap();
-
   private static Path resolveDefaultExtraDirectory(Path projectDirectory) {
     return projectDirectory.resolve("src").resolve("main").resolve("jib");
   }
+
+  private Path path;
+  private Map<String, String> permissions = Collections.emptyMap();
 
   @Inject
   public ExtraDirectoryParameters(Path projectDirectory) {
@@ -66,6 +66,13 @@ public class ExtraDirectoryParameters {
     this.path = path.toPath();
   }
 
+  /**
+   * Gets the permissions for files in the extra layer on the container. Maps from absolute path on
+   * the container to a 3-digit octal string representation of the file permission bits (e.g. {@code
+   * "/path/on/container" -> "755"}).
+   *
+   * @return the permissions map from path on container to file permissions
+   */
   @Input
   public Map<String, String> getPermissions() {
     if (System.getProperty(PropertyNames.EXTRA_DIRECTORY_PERMISSIONS) != null) {
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.gradle;
+
+import com.google.cloud.tools.jib.plugins.common.ConfigurationPropertyValidator;
+import com.google.cloud.tools.jib.plugins.common.PropertyNames;
+import java.io.File;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.Map;
+import javax.inject.Inject;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Internal;
+
+/** Object in {@link JibExtension} that configures the extra directory. */
+public class ExtraDirectoryParameters {
+
+  private Path path;
+  private Map<String, String> permissions = Collections.emptyMap();
+
+  private static Path resolveDefaultExtraDirectory(Path projectDirectory) {
+    return projectDirectory.resolve("src").resolve("main").resolve("jib");
+  }
+
+  @Inject
+  public ExtraDirectoryParameters(Path projectDirectory) {
+    path = resolveDefaultExtraDirectory(projectDirectory);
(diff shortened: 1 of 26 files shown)


chanseokoh:
I wonder how this will look like in Maven, because currently for other maps, an HTML tag name works as a key to the maps. But obviously we can't do `</path/to/file2>777<//path/to/file2>`.

Assuming that this will break the existing users who configured `extraDirectory`, another thing I'd like to make sure is that existing users will get a good error message so that they have no friction in updating their config, not like a native Java type mismatch error or such.

TadCordle:
> I wonder how this will look like in Maven, because currently for other maps, an HTML tag name works as a key to the maps. But obviously we can't do `</path/to/file2>777<//path/to/file2>`.

Judging by the original comment, I think we can make the map an alternating list style sort of thing:

```xml
<extraDirectory>
  <path>extra/dir/path</path>
  <permissions>
    <file>/path/file1</file>
    <mode>755</mode>
    <file>/path/file2</file>
    <mode>755</mode>
  </permissions>
</extraDirectory>
```

> Assuming that this will break the existing users who configured `extraDirectory`, another thing I'd like to make sure is that existing users will get a good error message so that they have no friction in updating their config, not like a native Java type mismatch error or such.

Yeah, I think I would have liked to deprecate the parameter first, but considering we're keeping the parameter and just changing its type I wasn't sure how the best way to do this.

coollog:
A slight correction on the Maven config, but otherwise looks good:

```xml
<extraDirectory>
  <path>extra/dir/path</path>
  <permissions>
    <permission>
      <file>/path/file1</file>
      <mode>755</mode>
    </permission>
    <permission>
      <file>/path/file2</file>
      <mode>755</mode>
    </permission>
  </permissions>
</extraDirectory>
```

I think for Gradle, we could actually support both the use of `extraDirectory` as a closure with the path and permissions fields or as a file to just be the path.
I believe in Maven, there's some way to make it so that if the user just uses `<extraDirectory>some/path</extraDirectory>`, it can take `some/path` as the `path` nested input, but I'll need to find that again.

TadCordle:
Hmmm I just noticed that changing permissions in the build file without changing anything in the extra directory itself causes the extra layer to not be updated because file permissions aren't cached.

TadCordle:
Have updated to support both `jib.extraDirectory { ... }` and `jib.extraDirectory = file(...)`.

coollog:
Good catch, we'll need to update the selector to include permissions then.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
```diff
@@ -0,0 +1,81 @@
...
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Internal;
+
+/** Object in {@link JibExtension} that configures the extra directory. */
+public class ExtraDirectoryParameters {
+
+  private Path path;
+  private Map<String, String> permissions = Collections.emptyMap();
+
+  private static Path resolveDefaultExtraDirectory(Path projectDirectory) {
```
coollog: nit: static above non-static

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+    return path;
+  }
+
+  public void setPath(File path) {
+    this.path = path.toPath();
+  }
+
+  @Input
+  public Map<String, String> getPermissions() {
```
coollog: Let's add a short javadoc here (or the setter) to explain what the map represents.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurations.java
```diff
@@ -49,13 +51,17 @@
    * @throws IOException if an I/O exception occurred during resolution
    */
   static JavaLayerConfigurations getForProject(
-      Project project, Logger logger, Path extraDirectory, AbsoluteUnixPath appRoot)
+      Project project,
+      Logger logger,
+      Path extraDirectory,
+      Map<AbsoluteUnixPath, FilePermissions> permissions,
```
chanseokoh: Javadoc about `permissions`. Also, it should be obvious from name or Javadoc that these file permissions apply only to the files from `extraDirectory`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurations.java
```diff
@@ -70,7 +76,11 @@ static JavaLayerConfigurations getForProject(
    * @throws IOException if an I/O exception occurred during resolution
    */
   private static JavaLayerConfigurations getForNonWarProject(
-      Project project, Logger logger, Path extraDirectory, AbsoluteUnixPath appRoot)
+      Project project,
+      Logger logger,
+      Path extraDirectory,
+      Map<AbsoluteUnixPath, FilePermissions> permissions,
```
chanseokoh: ditto

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurations.java
```diff
@@ -137,17 +151,21 @@ private static JavaLayerConfigurations getForNonWarProject(
...
+   * @param permissions map from path on container to file permissions
    * @param appRoot root directory in the image where the app will be placed
    * @return {@link JavaLayerConfigurations} for the layers for the Gradle {@link Project}
    * @throws IOException if an I/O exception occurred during resolution
    */
   private static JavaLayerConfigurations getForWarProject(
-      Project project, Logger logger, Path extraDirectory, AbsoluteUnixPath appRoot)
+      Project project,
+      Path extraDirectory,
+      Map<AbsoluteUnixPath, FilePermissions> permissions,
```
chanseokoh: ditto

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurations.java
```diff
@@ -137,17 +151,21 @@ private static JavaLayerConfigurations getForNonWarProject(
    * Resolves the {@link JavaLayerConfigurations} for a WAR Gradle {@link Project}.
    *
    * @param project the Gradle {@link Project}
-   * @param logger the build logger for providing feedback about the resolution
    * @param extraDirectory path to the source directory for the extra files layer
+   * @param permissions map from path on container to file permissions
```
chanseokoh: Let's mention this applies only to the files from `extraDirectory`.

PR #1182, opened 2018-10-23
Add file permissions to cache selector

### LINKED ISSUE #1180: Update cache selector to include file permissions
[Original comment](https://github.com/GoogleContainerTools/jib/pull/1178#issuecomment-432282847)

>Hmmm I just noticed that changing permissions in the build file without changing anything in the extra directory itself causes the extra layer to not be updated because file permissions aren't cached.

TadCordle:
Fixes #1180

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
@@ -43,11 +43,13 @@ import java.util.Objects;
  *     "sourceFile": "source/file/for/layer/entry/1",
  *     "extractionPath": "/extraction/path/for/layer/entry/1"
  *     "lastModifiedTime": "2018-10-03T15:48:32.416152Z"
+ *     "permissions": 511
  *   },
  *   {
  *     "sourceFile": "source/file/for/layer/entry/2",
  *     "extractionPath": "/extraction/path/for/layer/entry/2"
  *     "lastModifiedTime": "2018-10-03T15:48:32.416152Z"
+ *     "permissions": 511
  *   }
  * ]
  * }</pre>
@@ -61,12 +63,14 @@ class LayerEntriesSelector {
     private final String sourceFile;
     private final String extractionPath;
     private final Instant lastModifiedTime;
+    private final int permissions;
 
     @VisibleForTesting
     LayerEntryTemplate(LayerEntry layerEntry) throws IOException {
       sourceFile = layerEntry.getAbsoluteSourceFileString();
       extractionPath = layerEntry.getAbsoluteExtractionPathString();
       lastModifiedTime = Files.getLastModifiedTime(layerEntry.getSourceFile()).toInstant();
+      permissions = layerEntry.getPermissions().getPermissionBits();
     }
 
     @Override
@@ -80,7 +84,12 @@ class LayerEntriesSelector {
       if (extractionPathComparison != 0) {
         return extractionPathComparison;
       }
-      return lastModifiedTime.compareTo(otherLayerEntryTemplate.lastModifiedTime);
+      int lastModifiedTimeComparison =
+          lastModifiedTime.compareTo(otherLayerEntryTemplate.lastModifiedTime);
+      if (lastModifiedTimeComparison != 0) {
+        return lastModifiedTimeComparison;
+      }
+      return Integer.compare(permissions, otherLayerEntryTemplate.permissions);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
```diff
@@ -43,11 +43,13 @@
  *     "sourceFile": "source/file/for/layer/entry/1",
  *     "extractionPath": "/extraction/path/for/layer/entry/1"
  *     "lastModifiedTime": "2018-10-03T15:48:32.416152Z"
+ *     "permissions": 511
```
coollog: What do you think about using the string here instead?

TadCordle: I don't think this json is ever visible to anyone, so I didn't think it was worth it to add the extra step to make it more readable. But on the other hand, maybe we should add a method to `FilePermissions` to get the permission bits as a string anyway.

TadCordle: I'd still say at this point it's fine as is, but I don't feel strongly about it.

briandealwis: Strings might be better since 0511 would really be 329 here.

TadCordle: Right, but it's 511 not 0511, so it would actually be equal to 0777. And again, since this json is never actually readable (I think it just get converted straight into a digest) I wasn't too worried about people mixing up decimal/octal. The only thing that really matters is whether or not the contents change, not what they specifically are.

chanseokoh: 511 here, you probably meant the usual 777.

> it was worth it to add the extra step to make it more readable

It's very easy to write a string in octal with the static `Integer.toString(num, radix)` (the other way is easy too), so I vote for string.

TadCordle: I didn't mean it's difficult to code, I meant the code is performing one less step this way, since all we're doing is writing out the internal representation of `FilePermissions` to json. It's an extra operation to convert it to a string, which I didn't think was useful since this json is never actually useful to a user.

>511 here, you probably meant the usual 777.

If `layerEntry.getPermissions().getPermissionBits()` is 0777, then the json will be 511, won't it?

TadCordle: It seems strange to me to perform the extra step to make json that's immediately turned into a digest more readable, but I guess we can go with that if people feel strongly about it.

chanseokoh: > If layerEntry.getPermissions().getPermissionBits() is 0777, then the json will be 511, won't it?

Oh, sorry. I didn't really mean something is wrong. Was saying to @briandealwis just to say 511 is from 777.

> writing out the internal representation of `FilePermissions` to json

Ah, I didn't go over the code yet and didn't notice this is the result of a direct serialization.

chanseokoh: Looking at the `LayerEntrySelector.permissions` field, so the question is about whether to have the field of type `String` instead of `int`? I think `String` will work fine. Any issue?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/FilePermissions.java
```diff
@@ -86,14 +86,23 @@ public static FilePermissions fromPosixFilePermissions(
...
   public int getPermissionBits() {
     return permissionBits;
   }
 
+  /**
+   * Gets the octal string representation of the permissions.
+   *
+   * @return the octal string representation of the permissions
+   */
+  public String toOctalString() {
```
TadCordle: Maybe this should just be `toString`?

chanseokoh: I think `toOctalString` is better. There can be other string representation, like binary bits or just a human readable "rwxr--r--".

jib-core/src/test/java/com/google/cloud/tools/jib/cache/LayerEntriesSelectorTest.java
```diff
@@ -126,4 +150,26 @@ public void testGenerateSelector_fileModified() throws IOException {
...
+            AbsoluteUnixPath.get("/extraction/path"),
+            FilePermissions.fromOctalString("111"));
+    LayerEntry layerEntry222 =
+        new LayerEntry(
+            layerFile,
+            AbsoluteUnixPath.get("/extraction/path"),
+            FilePermissions.fromOctalString("222"));
+
+    // Verify that changing permissions generates a different selector
+    DescriptorDigest expectedSelector =
```
chanseokoh: The `expected` naming makes me to think that the actual should be equal to this. `entry111Selector` or just `selector`?

TadCordle: Ah right, I think that was the result of some copy-pasting. Actually I don't think the variable is even really necessary.

PR #1190, opened 2018-10-24
Remove LayerEntry#getAbsolute...String methods

### LINKED ISSUE #1008: Cache metadata source files needs to store windows path for windows.

TadCordle:
Fixes #1008

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
@@ -72,12 +72,9 @@ public class LayerEntry {
   }
 
   /**
-   * Gets the source file.
-   *
-   * <p>Do <b>not</b> call {@link Path#toString} on this - use {@link #getAbsoluteSourceFileString}
-   * instead. This path can be relative or absolute, but {@link #getAbsoluteSourceFileString} can
-   * only be absolute. Callers should rely on {@link #getAbsoluteSourceFileString} for the
-   * serialized form since the serialization could change independently of the path representation.
+   * Gets the source file. The source file may be relative or absolute, so the caller should use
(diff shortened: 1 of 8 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -103,26 +100,6 @@ public FilePermissions getPermissions() {
...
-  public String getAbsoluteSourceFileString() {
-    return AbsoluteUnixPath.fromPath(sourceFile.toAbsolutePath()).toString();
-  }
-
-  /**
-   * Gets the extraction path as an absolute path in Unix form.
-   *
-   * @return the extraction path
-   */
-  public String getAbsoluteExtractionPathString() {
```
coollog: Hmm, I dont think we should remove `getAbsoluteExtractionPathString` since it ends up just changing all calls to `getAbsoluteExtractionPathString()` into `getExtractionPath().toString()`.

TadCordle: I don't really see that as a problem. I'd rather remove a method that wraps one method call to remove the extra level of indirection than avoid typing an extra dot.

coollog: Yea, removing a one of the extraction path getters make sense. I feel like it makes more sense to remove the one that returns `AbsoluteUnixPath` but I guess `AbsoluteUnixPath` is straightforward enough to return.

PR #1195, opened 2018-10-25
Update maven extraDirectory config parameter to include path and permissions field

### LINKED ISSUE #794: executable permission is never set

TadCordle:
Fixes #794.

Adds/changes the `<extraDirectory>` configuration:
- Can still configure extra directory using `<extraDirectory>${project.basedir}/path/to/extra/directory</extraDirectory>`
- Can also configure the extra directory/permissions using new `<extraDirector><path>` and `<extraDirectory><permissions>` fields

e.g.

```xml
<configuration>
  ...
  <extraDirectory>
    <path>${project.basedir}/src/main/jib-custom</path>
    <permissions>
      <permission>
        <file>/foo</file>
        <mode>755</mode>
      </permission>
      <permission>
        <file>/bar/cat</file>
        <mode>777</mode>
      </permission>
    </permissions>
  </extraDirectory>
</configuration>
```


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -417,6 +458,37 @@ String getFormat() {
...
+  /**
+   * Gets the configured extra directory path.
+   *
+   * @return the configured extra directory path
+   */
+  @Nullable
+  Path getExtraDirectoryPath() {
+    // TODO: Should inform user about nonexistent directory if using custom directory.
+    if (System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH) != null) {
+      return Paths.get(System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH));
```
briandealwis: Should check the Maven settings and pom properties too to keep with the semantics of Maven's `@Parameter(property)` values?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -417,6 +458,37 @@ String getFormat() {
...
+    return extraDirectory.path == null ? null : extraDirectory.path.toPath();
+  }
+
+  /**
+   * Gets the configured extra layer file permissions.
+   *
+   * @return the configured extra layer file permissions
+   */
+  List<PermissionConfiguration> getExtraDirectoryPermissions() {
+    if (System.getProperty(PropertyNames.EXTRA_DIRECTORY_PERMISSIONS) != null) {
```
briandealwis: Do you really expect this to happen?

TadCordle: Not particularly, but every other parameter has a corresponding system property, so I think it's best to keep it consistent.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
TadCordle: Is this a trivial change? It sounds like something we may want to do for all of the parameters, so maybe it should wait until another PR.

TadCordle: Speaking of which, I think supporting globs would be a good idea, but I think that should wait until another PR as well.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -193,4 +191,23 @@ static FileCollection getInputFiles(File extraDirectory, Project project) {
...
+   * Validates and converts a {@code String->String} file-path-to-file-permissions map to an
+   * equivalent {@code AbsoluteUnixPath->FilePermission} map.
+   *
+   * @param inputMap the map to convert
+   * @return the converted map
+   */
+  @VisibleForTesting
+  static Map<AbsoluteUnixPath, FilePermissions> convertPermissionsMap(
+      Map<String, String> inputMap) {
+    ImmutableMap.Builder<AbsoluteUnixPath, FilePermissions> permissionsMap = ImmutableMap.builder();
```
chanseokoh: nit: I think we can just return a `HashMap`. I see no need to prevent callers from modifying the returned map.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/PluginConfigurationProcessor.java
```diff
@@ -102,6 +108,50 @@ static String getBaseImage(JibPluginConfiguration jibPluginConfiguration) {
...
+   * Validates and converts a list of {@link PermissionConfiguration} to an equivalent {@code
+   * AbsoluteUnixPath->FilePermission} map.
+   *
+   * @param inputList the list to convert
+   * @return the resulting map
+   */
+  @VisibleForTesting
+  static Map<AbsoluteUnixPath, FilePermissions> convertPermissionsList(
+      List<PermissionConfiguration> inputList) {
+    ImmutableMap.Builder<AbsoluteUnixPath, FilePermissions> permissionsMap = ImmutableMap.builder();
```
chanseokoh: ditto, `HashMap`?

jib-maven-plugin/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ All notable changes to this project will be documented in this file.
 - `<to><credHelper>` and `<from><credHelper>` can be used to specify a credential helper suffix or a full path to a credential helper executable ([#925](https://github.com/GoogleContainerTools/jib/issues/925))
 - `container.user` configuration parameter to configure the user and group to run the container as ([#1029](https://github.com/GoogleContainerTools/jib/issues/1029))
 - Preliminary support for building images for WAR projects ([#431](https://github.com/GoogleContainerTools/jib/issues/431))
+- `<extraDirectory>` object with a `<path>` and `<permissions>` field ([#794](https://github.com/GoogleContainerTools/jib/issues/794))
```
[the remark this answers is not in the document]
coollog: I think object is the right term here since this is part of the project object model.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
coollog: Oh yea, and for cases where we were to return `ImmutableMap`, we should be making the return type be `ImmutableMap`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
TadCordle: Filed #1200 and #1201 for these 2 comments.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -169,14 +218,6 @@ public void set(String image) {
       property = PropertyNames.ALLOW_INSECURE_REGISTRIES)
   private boolean allowInsecureRegistries;
 
-  // this parameter is cloned in FilesMojo
-  @Nullable
-  @Parameter(
-      defaultValue = "${project.basedir}/src/main/jib",
```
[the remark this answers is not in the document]
coollog: Hmm, would passing `project.basedir/src/main/jib` in via a constructor work? And call `new ExtraDirectoryParameters(project.basedir/src/main/jib)` (not right syntax) on line 266.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -202,12 +202,12 @@ static FileCollection getInputFiles(File extraDirectory, Project project) {
   @VisibleForTesting
   static Map<AbsoluteUnixPath, FilePermissions> convertPermissionsMap(
       Map<String, String> inputMap) {
-    ImmutableMap.Builder<AbsoluteUnixPath, FilePermissions> permissionsMap = ImmutableMap.builder();
+    HashMap<AbsoluteUnixPath, FilePermissions> permissionsMap = new HashMap<>();
```
coollog: tiny nit: prefer `Map` for type

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -193,4 +191,23 @@ static FileCollection getInputFiles(File extraDirectory, Project project) {
       return project.files(dependencyFileCollections);
     }
   }
+
+  /**
+   * Validates and converts a {@code String->String} file-path-to-file-permissions map to an
+   * equivalent {@code AbsoluteUnixPath->FilePermission} map.
+   *
+   * @param inputMap the map to convert
```
coollog: nit: `inputMap` should probably be more descriptive - might be good to even include an example of an entry in this map.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -87,6 +90,31 @@ private void setPropertyDescriptors(String descriptorPrefix) {
     }
   }
 
+  /** Used to configure {@code extraDirectory.permissions} parameter. */
+  public static class PermissionConfiguration {
+
+    @Nullable @Parameter private String file;
+
```
coollog: tiny nit: don't need to separate these with newline

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -143,6 +171,23 @@ public void set(String image) {
...
   }
 
+  /** Configuration for the {@code extraDirectory} parameter. */
+  public static class ExtraDirectoryParameters {
+
+    @Nullable @Parameter private File path;
+
+    @Parameter private List<PermissionConfiguration> permissions = Collections.emptyList();
+
+    public void set(File path) {
```
coollog: nit: maybe add a javadoc about how this allows for setting a path string directly as the `extraDirectory`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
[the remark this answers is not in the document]
coollog: I thought that was just for objects that Maven has to create (like objects in a list), but `extraDirectory` we are just constructing that explicitly?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/PluginConfigurationProcessor.java
```diff
@@ -102,6 +108,49 @@ static String getBaseImage(JibPluginConfiguration jibPluginConfiguration) {
...
+                .resolve("src")
+                .resolve("main")
+                .resolve("jib"));
+  }
+
+  /**
+   * Validates and converts a list of {@link PermissionConfiguration} to an equivalent {@code
+   * AbsoluteUnixPath->FilePermission} map.
+   *
+   * @param inputList the list to convert
```
coollog: ditto the variable naming nit

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/PluginConfigurationProcessor.java
```diff
@@ -102,6 +108,49 @@ static String getBaseImage(JibPluginConfiguration jibPluginConfiguration) {
...
+   * @return the resulting map
+   */
+  @VisibleForTesting
+  static Map<AbsoluteUnixPath, FilePermissions> convertPermissionsList(
+      List<PermissionConfiguration> inputList) {
+    HashMap<AbsoluteUnixPath, FilePermissions> permissionsMap = new HashMap<>();
```
coollog: ditto type nit

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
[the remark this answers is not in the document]
TadCordle: From the looks of it, `project` is still `null` at the time we explicitly construct `extraDirectory`.

Issue #1179, opened 2018-10-22
Add layer description to history


loosebazooka:
Feedback from @anthonydahanne after code one talk. Take #840 further.

Instead of generic entries, add description of what the layer is holding. "Dependencies", "Snapshot Dependencies", "Classes", "Resources"

anthonydahanne:
thanks for adding this issue to the backlog !

coollog:
@anthonydahanne Thanks for suggesting this!

TadCordle:
Fixed with #1198

PR #1198, opened 2018-10-28
Add layer description history

paivagustavo:
This closes the issue #1179 adding the descriptions ('Classes', 'Dependencies', 'Resources' and 'Extra files') to the layer history. 

Example:
```
e4f32e363ba1   45 minutes ago      Extra files created by jib-maven-plugin:0.9.14-SNAPSHOT    6B                  
<missing>           45 minutes ago      Classes created by jib-maven-plugin:0.9.14-SNAPSHOT        2.62kB              
<missing>           45 minutes ago      Resources created by jib-maven-plugin:0.9.14-SNAPSHOT      5B                  
<missing>           45 minutes ago      Dependencies created by jib-maven-plugin:0.9.14-SNAPSHOT   770B                
<missing>           48 years ago        bazel build ...                                            99.8MB              
<missing>           48 years ago        bazel build ...                                            1.93MB              
<missing>           48 years ago        bazel build ...                                            16.8MB              
```


coollog:
Thanks for the contribution! @GoogleContainerTools/java-tools What do you think about having the layer type be the comment instead, so the `created by` still is as before?

chanseokoh:
Adding the layer type looks good.

GuustavoPaiva:
> Thanks for the contribution! @GoogleContainerTools/java-tools What do you think about having the layer type be the `comment` instead, so the `created by` still is the same as before?

I think this makes more sense, I'll change this to work like that.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -162,6 +167,13 @@
     }
   }
 
+  private static String capitalizeFirstLetter(String string) {
```
loosebazooka: Capitalization is probably unnecessary, would prefer we just use the value in `getLayerType()` directly.

GuustavoPaiva:
Changed to add the layer type on the comments and without capitalizing it, so the updated example is: 
```
IMAGE               CREATED             CREATED BY                         SIZE                COMMENT
cad97fa5d3a0        21 minutes ago      jib-maven-plugin:0.9.14-SNAPSHOT   6B                  extra files
<missing>           21 minutes ago      jib-maven-plugin:0.9.14-SNAPSHOT   2.62kB              classes
<missing>           21 minutes ago      jib-maven-plugin:0.9.14-SNAPSHOT   5B                  resources
<missing>           21 minutes ago      jib-maven-plugin:0.9.14-SNAPSHOT   770B                dependencies
<missing>           48 years ago        bazel build ...                    99.8MB              
<missing>           48 years ago        bazel build ...                    1.93MB              
<missing>           48 years ago        bazel build ...                    16.8MB              

```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/BuildImageStepTest.java
```diff
@@ -297,11 +344,12 @@ public void test_generateHistoryObjects() throws ExecutionException, Interrupted
     Assert.assertEquals(expectedAddedBaseLayerHistory, image.getHistory().get(4));
 
     // Application layers (3 generated)
```
TadCordle: This is 4 with the addition of the extra layer.

PR #1203, opened 2018-10-29
Make equals() symmetric

### LINKED ISSUE #1202: Make equals() symmetric
As discussed in _Effective Java_, equals() should be symmetric and hence cannot use `instanceof`.

briandealwis:
Fixes #1202

Oops, hit "Enter" by mistake — this needs to be rebased once #1155 is committed.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainer.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainer.java
@@ -59,7 +59,10 @@ public class JibContainer {
     if (this == other) {
       return true;
     }
-    if (!(other instanceof JibContainer)) {
+    if (other == null) {
+      return false;
+    }
+    if (getClass() != other.getClass()) {
       return false;
     }
     JibContainer otherContainer = (JibContainer) other;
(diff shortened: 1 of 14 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainer.java
```diff
@@ -59,7 +59,10 @@ public boolean equals(Object other) {
     if (this == other) {
       return true;
     }
-    if (!(other instanceof JibContainer)) {
+    if (other == null) {
```
chanseokoh: I think we can combine these two into `if (other == null || getClass() != other.getClass())`.

jib-core/src/main/java/com/google/cloud/tools/jib/blob/BlobDescriptor.java
```diff
@@ -84,16 +84,21 @@ public long getSize() {
...
+    if (other == null) {
       return false;
     }
-
-    BlobDescriptor other = (BlobDescriptor) obj;
-    return size == other.getSize() && digest.equals(other.getDigest());
+    if (getClass() != other.getClass()) {
+      return false;
+    }
+    if (size < 0) {
```
chanseokoh: This looks weird. It will be right to get rid of it, but I'm unsure how this will affect the current code. Perhaps there was a reason, so until we figure that out, let's leave it there.

briandealwis:
Closing: I clearly misread my copy where he in fact advocates for the `instanceof` test.  Sorry for the noise.

PR #1205, opened 2018-10-29
Share more code between tasks, Maven and Gradle

chanseokoh:
Trying to share more code while retaining the current structure. I don't intend to radically change the current design in this PR, so this is largely moving around code.

1. `PluginConfigurationProcessor` configures both `Containerizer` and `JibContainerBuilder`.
   - Previously, tasks will need to create a `TargetImage` and a `Containizer` after calling PCP, and then call another PCP method to configure `Containerizer`.
   - Now, each task will call one of `process...ForDockerDaemonImage`, `...ForTarImage`, or `...ForRegistryImage`.
1. Shares the target image ref generation code.
1. Shares the code to add credential retrievers for adding the `<from>` and `<to>` auth info. (See `configureCredentialRetrievers`)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -204,8 +230,105 @@ public static PluginConfigurationProcessor processCommonConfiguration(
...
+    }
+    try {
+      return AbsoluteUnixPath.get(appRoot);
+    } catch (IllegalArgumentException ex) {
+      throw new AppRootInvalidException(appRoot, appRoot, ex);
+    }
+  }
+
+  @FunctionalInterface
+  private static interface InferredAuthProvider {
```
coollog: `static` unnecessary

coollog: interfaces and classes should go above methods at top of class

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -204,8 +230,105 @@ public static PluginConfigurationProcessor processCommonConfiguration(
...
+
+    Optional<AuthProperty> getInferredAuth(String registry) throws InferredAuthRetrievalException;
+  };
+
+  private static boolean configureCredentialRetrievers(
```
coollog: This would be good to refactor later so that it doesn't have such a large parameter list.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -204,8 +230,105 @@ public static PluginConfigurationProcessor processCommonConfiguration(
...
+      EventDispatcher eventDispatcher,
+      RegistryImage registryImage,
+      ImageReference imageReference,
+      String usernamePropertyName,
+      String passwordPropertyName,
+      AuthProperty knownAuth,
+      String knownAuthSource,
+      InferredAuthProvider inferredAuthProvider,
+      Optional<String> credHelper)
```
coollog: `Optional` should only be used as return types

coollog:
Looks great! This removed a lot of redundancy. I'll have someone else do the second pass and approval.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: Agreed, I'll add a TODO. The long list of parameters is a consequence of just trying to technically share some section of code. At least we can remove `knownAuthSource` once the NPE problem is fixed.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: Oh, I think I have not been following the repo style. I've been putting priority `public`/`private`.. over anything (even if they are classes).

TadCordle: I think interfaces are just implicitly static?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
coollog: Ah yea, I've been meaning to add the style for that so we can reference it.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -133,13 +118,82 @@ public static String getBaseImage(
...
+        rawConfiguration, projectProperties, containerizer, targetImageReference, false);
+  }
+
+  public static PluginConfigurationProcessor processCommonConfigurationForRegistryImage(
       RawConfiguration rawConfiguration, ProjectProperties projectProperties)
-      throws InvalidImageReferenceException, NumberFormatException, FileNotFoundException,
-          MainClassInferenceException, AppRootInvalidException, InferredAuthRetrievalException {
+      throws InferredAuthRetrievalException, InvalidImageReferenceException,
+          MainClassInferenceException, AppRootInvalidException, IOException {
+    Preconditions.checkArgument(rawConfiguration.getToImage().isPresent());
```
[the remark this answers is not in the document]
chanseokoh: We validate them in the image building Mojo and Gradle task:

https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java#L90

I agree at the moment, validation on various values happen here and there at different times. It'd be nice to have validation in a more coherent and consistent way.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -204,8 +235,100 @@ public static PluginConfigurationProcessor processCommonConfiguration(
...
+      throws FileNotFoundException, InferredAuthRetrievalException {
+    DefaultCredentialRetrievers defaultCredentialRetrievers =
+        DefaultCredentialRetrievers.init(
+            CredentialRetrieverFactory.forImage(imageReference, eventDispatcher));
+    Optional<Credential> optionalToCredential =
+        ConfigurationPropertyValidator.getImageCredential(
+            eventDispatcher, usernamePropertyName, passwordPropertyName, knownAuth);
+    boolean credentialPresent = optionalToCredential.isPresent();
+    if (optionalToCredential.isPresent()) {
+      // TODO: fix https://github.com/GoogleContainerTools/jib/issues/1177
```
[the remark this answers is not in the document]
chanseokoh: I looked into this, but it did not seem trivial for some reason I don't remember now. That's why I put up an issue in the past. I agree we should fix this sooner than later.

PR #1208, opened 2018-10-30
Adds initial style guide.

coollog:
Adds an initial style guide as rules that extend the Google Java style guide for our codebase. We will add more rules/amend existing rules as necessary.

Preview at: https://github.com/GoogleContainerTools/jib/blob/style-guide/STYLE_GUIDE.md

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -1,0 +1,48 @@
+# Style guide
+
+This style guide and defines specific coding standards and advice for this Java codebase. The rules here are extensions to the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).
+
+Please see the [contributing guide](CONTRIBUTING.md) for general guidance for contributing to this project.
+
+### Automatic formatting
+
+Automatic formatting should be performed with `./gradlew goJF` or `./mvnw fmt:format`.
+
+### Class member order
+
+*Extends [3.4.2](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents)*
+
+Class members should in the following order in decreasing priority:
+
+1. Static before non-static
+1. Nested classes/interfaces before fields before constructors before methods
+1. Public before private
+1. Final before non-final
+
+### Public APIs
+
+User-facing methods (such as those in Jib Core) should not have types in their signature that are not standard JDK classes. For example, a parameter should take type `List` rather than Guava's `ImmutableList`.
+
+### Package hierarchy
+
+Packages should depend on each other without cycles.
+
+The following is a list of current `jib-core` packages (under `com.google.cloud.tools.jib`) and their immediate dependencies. These can be amended as code changes, but there should not be cyclical dependencies.
+
+- `api`
+- `async`
+- `blob` - `filesystem`, `hash`, `image` (cycle - should fix)
+- `builder` - `async`, `blob`, `builder`, `cache` `configuration`, `docker`, `event`, `filesystem`, `global`, `http`, `image`, `json`, `registry`
+- `cache` - `blob`, `filesystem`, `hash`, `image`, `json`
+- `configuration` - `cache`, `filesystem`, `event`, `image`, `registry`
+- `docker` - `blob`, `cache`, `image`, `json`, `tar`
+- `event`
+- `filesystem`
+- `frontend` - `configuration`, `event`, `filesystem`, `image`, `registry`
+- `global`
+- `hash` - `blob`, `image`
+- `http` - `blob`
+- `image` - `blob`, `configuration` (cycle - should fix - `ImageToJsonTranslator`), `filesystem`, `json`, `tar`
+- `json` - `blob`
+- `registry` - `blob`, `builder` (cycle - should fix - `RegistryClient`), `configuration` (cycle - should fix - `DockerConfigCredentialRetriever`), `event`, `global`, `http`, `image`, `json`
+- `tar` - `blob`
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -1,18 +1,18 @@
 # Style guide
 
-This style guide and defines specific coding standards and advice for this Java codebase. The rules here are extensions to the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).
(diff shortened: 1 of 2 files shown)


STYLE_GUIDE.md
```diff
@@ -0,0 +1,48 @@
+# Style guide
+
+This style guide and defines specific coding standards and advice for this Java codebase. The rules here are extensions to the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html).
```
TadCordle: "This style guide ~~and~~ defines"

STYLE_GUIDE.md
```diff
@@ -0,0 +1,48 @@
...
+
+Please see the [contributing guide](CONTRIBUTING.md) for general guidance for contributing to this project.
+
+### Automatic formatting
+
+Automatic formatting should be performed with `./gradlew goJF` or `./mvnw fmt:format`.
```
TadCordle: Should we mention `build.sh` so people can format all the modules at once?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Class member order
+
+*Extends [3.4.2](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents)*
+
+Class members should in the following order in decreasing priority:
```
TadCordle: Missing a word ("Class members should **appear/be?** in the following order")

STYLE_GUIDE.md
```diff
@@ -0,0 +1,48 @@
...
+
+User-facing methods (such as those in Jib Core) should not have types in their signature that are not standard JDK classes. For example, a parameter should take type `List` rather than Guava's `ImmutableList`.
+
+### Package hierarchy
+
+Packages should depend on each other without cycles.
+
+The following is a list of current `jib-core` packages (under `com.google.cloud.tools.jib`) and their immediate dependencies. These can be amended as code changes, but there should not be cyclical dependencies.
+
+- `api`
```
TadCordle: This section is a bit hard to read, I wonder if we could somehow arrange this in a tree/pyramid structure to make the relationships more clear? Maybe not possible until we fix the cyclical dependencies.

coollog: Yea, I think we can iterate on this to make it more readable as we fix the cycles.

TadCordle:
LGTM for now, will let someone else approve.

STYLE_GUIDE.md
briandealwis: Could you use `dot` or something to express this as a graph?

briandealwis: It's a graph generator from [GraphViz](https://www.graphviz.org).  There's a [web generator](http://www.webgraphviz.com).

coollog: Oh cool, I'll try that out.

coollog: Oh man I tried it out but the graph is quite messy:

```
digraph G {
"api"
"async"
"blob" -> "filesystem"
"blob" -> "hash"
"blob" -> "image" [label = "cycle - should fix" ]
"builder" -> "async"
"builder" -> "blob"
"builder" -> "cache"
"builder" -> "configuration"
"builder" -> "docker"
"builder" -> "event"
"builder" -> "filesystem"
"builder" -> "global"
"builder" -> "http"
"builder" -> "image"
"builder" -> "json"
"builder" -> "registry"
"cache" -> "blob"
"cache" -> "filesystem"
"cache" -> "hash"
"cache" -> "image"
"cache" -> "json"
"configuration" -> "cache"
"configuration" -> "filesystem"
"configuration" -> "event"
"configuration" -> "image"
"configuration" -> "registry"
"docker" -> "blob"
"docker" -> "cache"
"docker" -> "image"
"docker" -> "json"
"docker" -> "tar"
"event"
"filesystem"
"frontend" -> "configuration"
"frontend" -> "event"
"frontend" -> "filesystem"
"frontend" -> "image"
"frontend" -> "registry"
"global"
"hash" -> "blob"
"hash" -> "image"
"http" -> "blob"
"image" -> "blob"
"image" -> "configuration" [ label = "cycle - should fix - ImageToJsonTranslator" ]
"image" -> "filesystem"
"image" -> "json"
"image" -> "tar"
"json" -> "blob"
"registry" -> "blob"
"registry" -> "builder" [ label = "cycle - should fix - RegistryClient" ]
"registry" -> "configuration" [ label = "cycle - should fix - DockerConfigCredentialRetriever" ]
"registry" -> "event"
"registry" -> "global"
"registry" -> "http"
"registry" -> "image"
"registry" -> "json"
"tar" -> "blob"
}
```
resulting graph:
graph

STYLE_GUIDE.md
```diff
@@ -0,0 +1,48 @@
...
+Class members should be in the following order, in decreasing priority:
+
+1. Static before non-static
+1. Nested classes/interfaces before fields before constructors before methods
+1. Public before private
+1. Final before non-final
+
+### Public APIs
+
+User-facing methods (such as those in Jib Core) should not have types in their signature that are not standard JDK classes. For example, a parameter should take type `List` rather than Guava's `ImmutableList`.
```
patflynn: presumably you could have jib core types.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Package hierarchy
+
+Packages should depend on each other without cycles.
```
patflynn: you mean java packages? what about sub-packages? This is a little bit more stringent than what good build systems enforce which is no-cycles between build level modules.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,48 @@
...
+
+Automatic formatting should be performed with `./gradlew goJF` or `./mvnw fmt:format`. Formatting all projects can be done with `./build.sh format`.
+
+### Class member order
+
+*Extends [3.4.2](https://google.github.io/styleguide/javaguide.html#s3.4.2-ordering-class-contents)*
+
+Class members should be in the following order, in decreasing priority:
+
+1. Static before non-static
```
patflynn: While I generally agree with tending to order things in more or less this way, I've found the more relaxed logical ordering requirement in the linked google style guide to work pretty well.  In a way, this section goes against the advice in the Google style-guide.

Defining symbols close to first use is a common practice at Google that often helps readability but would be disallowed by this strict ordering.

PR #1214, opened 2018-11-02
Allow configuration docker executable path and additional environment variables

chanseokoh:
#468 for Gradle. Branched off of #1173.

Allow configuration docker executable path and additional environment variables.

```gradle
jibDockerBuild.dockerClient.executable = '/usr/bin/docker'
jibDockerBuild.dockerClient.environment = [KEY: 'value', DOCKER_HOST: '...']

jibDockerBuild {
  dockerClient {
    executable = ...
    environment = ...
  }
}
```


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Optional;
+
+/**
+ * Object in {@link BuildDockerTask} that configures the Docker executable and the additional
+ * environment variables to use when executing the executable.
+ */
+class DockerClientParameters {
```
coollog: I believe this needs to be public for the `build.gradle` to use it correctly.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public void setExecutable(String executable) {
+    this.executable = Paths.get(executable);
+  }
+
+  @Input
+  @Nullable
+  @Optional
+  public Map<String, String> getEnvironment() {
```
coollog: I think `setEnvironment` is missing

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleRawConfiguration.java
```diff
@@ -136,4 +143,14 @@ public boolean getUseOnlyProjectCache() {
   public ImageFormat getImageFormat() {
     return jibExtension.getContainer().getFormat();
   }
+
+  @Override
+  public Optional<Path> getDockerClientExecutable() {
```
coollog: Just an alternative suggestion: I'm thinking maybe since these are only used when calling `processCommonConfigurationForDockerDaemonImage`, perhaps we can just have them be passed in via parameters to `processCommonConfigurationForDockerDaemonImage` like how the `tarImagePath` is a parameter to `processCommonConfigurationForTarImage`, rather than add them to the `RawConfiguration ?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
chanseokoh: I noticed `build.gradle` works without it (don't know why), but do you think the setter should actually be there?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
chanseokoh: It worked with `build.gradle`, but I'll change it to public anyway.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -41,17 +44,22 @@
    * @return a new {@link DockerClient}
    */
   public static DockerClient newClient() {
-    return new DockerClient(defaultProcessBuilderFactory(DEFAULT_DOCKER_CLIENT));
+    return newClient(null, null);
```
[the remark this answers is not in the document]
chanseokoh: That works, but the downside is that `BuildDockerTask` will have to do

```java
if (dockerExecutable != null && dockerEnvironment != null) {
  ... = DockerClient.newClient(dockerExecutable, dockerEnvironment);
} else if (dockerExecutable == null && dockerEnvironment == null ) {
  ... = DockerClient.newClient();
} else {
  ... ???
}
```

or make internal defaults of `DockerClient` public, so that

```java
... = DockerClient.newClient(
    dockerClient == null ? DockerClient.DEFAULT_DOCKER_CLIENT : dockerClient.toString(),
    dockerEnvironment == null ? ... : dockerEnvironment);
```

or define `DockerBuildTask`s own defaults, duplicating defaults in multiple places.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleRawConfiguration.java
chanseokoh: Thought about that too, and since you said it, let's do it that way.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
coollog: Hmm that's interesting. How does it know what the set.. @loosebazooka Is no setter a valid construct?

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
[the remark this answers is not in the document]
coollog: Hmm, maybe this warrants a simple `Builder` to make the construct clearer - using `null` to mean default where the default is not `null` feels a bit off

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
chanseokoh: One thing I also noticed is that, `JibExtension` doesn't have a setter for `useOnlyProjectCache` or `allowInsecureRegistries` either. (And interestingly, the getters for them are not public but package-private.)

But perhaps il'll be just safer to add a setter. Who knows if older Gradle versions won't work if there is no setter. I'll add a setter.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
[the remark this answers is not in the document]
chanseokoh: Good idea. Let me try a builder.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
coollog: I might just be remembering things wrong and perhaps a setter is only necessary if we need to accept a different type than what's declared on the getter.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
```diff
@@ -216,7 +216,7 @@ public void testSteps_forBuildToDockerDaemon()
                 ImageReference.of("gcr.io", "distroless/java", "latest"),
                 ImageReference.of(null, imageReference, null))
             .build();
-    BuildSteps.forBuildToDockerDaemon(DockerClient.newClient(), buildConfiguration).run();
+    BuildSteps.forBuildToDockerDaemon(new DockerClient.Builder().build(), buildConfiguration).run();
```
coollog: I'd recommend having a static factory for the fully-default case (like `DockerClient.newDefaultClient()`)

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
```diff
@@ -70,7 +71,7 @@ private DockerDaemonImage(ImageReference imageReference) {
    * @param dockerExecutable the path to the {@code docker} CLI
    * @return this
    */
-  public DockerDaemonImage setDockerExecutable(Path dockerExecutable) {
+  public DockerDaemonImage setDockerExecutable(@Nullable Path dockerExecutable) {
```
coollog: I think javadoc should say something about `null` meaning default, though I lean towards not having `null` mean default and having the caller do the null check.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -88,9 +88,12 @@ public void buildDocker()
       throws InvalidImageReferenceException, IOException, BuildStepsExecutionException,
           CacheDirectoryCreationException, MainClassInferenceException,
           InferredAuthRetrievalException {
-    DockerClient dockerClient =
-        DockerClient.newClient(dockerClientParameters.getExecutable(), null);
-    if (!dockerClient.isDockerInstalled()) {
+    Path dockerExecutable = dockerClientParameters.getExecutable();
+    DockerClient.Builder dockerClientBuilder = new DockerClient.Builder();
```
coollog: I lean towards having a static factory to create a builder (`DockerClient.builder()`) over using the inner class constructor.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -88,9 +88,12 @@ public void buildDocker()
...
+    if (dockerExecutable != null) {
+      dockerClientBuilder.setDockerExecutable(dockerExecutable);
+    }
+    if (!dockerClientBuilder.build().isDockerInstalled()) {
```
coollog: I just realized - is this creating a `DockerClient` solely for checking the executable exists? I think we can change this to a static method then.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -61,13 +76,19 @@ public static DockerClient newClient(Path dockerExecutable) {
    * @param dockerExecutable path to {@code docker}
    * @return the default {@link ProcessBuilder} factory for running a {@code docker} subcommand
    */
-  private static Function<List<String>, ProcessBuilder> defaultProcessBuilderFactory(
-      String dockerExecutable) {
+  @VisibleForTesting
+  static Function<List<String>, ProcessBuilder> defaultProcessBuilderFactory(
```
coollog: We probably want to make `dockerEnvironment` immutable so there's no expected behavior if the user changes the map they passed into the builder since the map can change and the environment used by the `ProcessBuilder` would be different.

chanseokoh: Ah, good catch. I think we can copy the map at the beginning.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
chanseokoh: Which one do you prefer,
```java
Path executable = dockerClientParameters.getExecutable();
boolean isDockerInstalled =
    executable == null
        ? DockerClient.isDefaultDockerInstalled()
        : DockerClient.isDockerInstalled(executable);
```
or
```java
  DockerClient.isDockerInstalled(dockerClientParameters.getExecutable());
```
?
I guess you don't like passing `null` to mean using a default value, so the former which needs adding an additional static method?

coollog: I think since ideally `getExecutable` should return `Optional` and perform the logic based on `Optional.isPresent`. Passing `null` down in general feels like tailoring an invoked method's semantics to the semantics of the input (which we happen to do in `BuildConfiguration.Builder` - for example, by allowing `null` to mean default, we lose the ability to actually set something to `null`). I think in this case, both are fine mostly because this is a static method and doesn't pass the nullable value down further (to instantiate instances).

chanseokoh: Will opt for the first (with two static methods).

> by allowing `null` to mean default, we lose the ability to actually set something to `null`

On a side note, regarding this topic, here's my 2 cents: we shouldn't confuse the semantics of "something missing" (often coded by the `null` value of the Java language or `Optional.isEmpty()` in implementation detail) vs "setting a special, concrete value, e.g., setting a container metadata property to the JSON null value." A component can be designed to act reasonably or properly when something is missing, and the involved parties should agree upon the contract of what will happen in that case. (If "something missing" should translate to setting a property in JSON to the JSON null value, it can work like that by that design.) And if someone does intent to set a property to the JSON null value, then I think it is actually directing to set an explicit value, which is different from saying "I have no information". The confusion may just be that this JSON null value happens to look similar to the `null` keyword in Java code, which may be used to represent that something is missing as one possible Java implementation detail.

coollog: Much agreed

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -79,19 +95,22 @@ public DockerClient build() {
   @VisibleForTesting
   static Function<List<String>, ProcessBuilder> defaultProcessBuilderFactory(
       String dockerExecutable, Map<String, String> dockerEnvironment) {
+    Map<String, String> environmentCopy = new HashMap<>(dockerEnvironment);
```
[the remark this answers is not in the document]
chanseokoh: I'm not using `ImmutableMap.copyOf` and I want to believe `copyOf()` does the operation safely (good to know this anyway), but I'll change the argument to `ImmutableMap`, as this is not part of the public API.

TadCordle:
We should probably add a changelog entry for this. Also are we waiting until 0.10.1 before merging this so we have more time to add the maven side?

coollog:
I think we can actually get this in for `0.10.0`.

chanseokoh:
Yeah, I think we can get this in for `0.10.0`. We can do the Maven part later. I will update `CHANGELOG` in another PR very soon.

Issue #1121, opened 2018-10-09
Is there a way to add a volume in image ?


laurentleseigneur:
**Image don't allows to have a volume**:

my project is a CLI that uses a folder as arg to run. I would like to run produced docker image with a volume to bind to some host folder

**Expected behavior**:

add some *volume* parameter when configuring maven/gradle plugin

coollog:
Hi @laurentleseigneur , thanks for your suggestions! Currently we do not provide configuration for setting the volume, but we'll look into this as a potential feature.

GuustavoPaiva:
Hey @coollog wanna some help with these? This seems cool.

coollog:
Hey @GuustavoPaiva , we’d be glad to have your help on this! Let us know if you have any questions. One thing is that this is probably better done as multiple PRs, like one for the `jib-core` changes, one for the `jib-maven-plugin` changes, and one for the `jib-gradle-plugin` changes.

coollog:
@GuustavoPaiva , let us know if you'd like to handle the `jib-maven-plugin`/`jib-gradle-plugin` changes as well!

GuustavoPaiva:
Hey @coollog I plan to do this by the end of this week, is it fine?

chanseokoh:
@GuustavoPaiva sure, we welcome your help.

chanseokoh:
Fixed by #1278. @GuustavoPaiva thanks again for your help!

laurentleseigneur:
thanks you all for adding this feature in Jib!

PR #1215, opened 2018-11-02
Add volume support to jib-core.

GuustavoPaiva:
First part of #1121, this adds support to docker volumes on the jib-core project.
I tried to keep the comments and the same style to other fields, more specifically the exposedPorts, as these two have the same syntax.

I've also chained some methods call of `Image.Builder` on the `BuildImageStep` class.


GuustavoPaiva:
@coollog as you were the one to present the proposal, would like your review. 

Oh, I think I forgot to add support on the Jib api for AbsoluteUnixPath, if you want I'll add this as well.

coollog:
Hi @GuustavoPaiva , thanks for contributing this - we'll review it soon!

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -239,6 +239,43 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
     return this;
   }
 
+  /**
+   * Sets the directories that the container will create as a volume. Volumes allows persistent
```
coollog: Just some wording suggestions: I believe the volume definition in more in lines of "Sets the directories that may hold externally mounted volumes."

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -239,6 +239,43 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
+   * data.
+   *
+   * <p>This is similar to <a href="https://docs.docker.com/engine/reference/builder/#volume">{@code
+   * VOLUME} in Dockerfiles</a>.
+   *
+   * @param volumes the list of volumes
```
coollog: "the list of directories to set as volumes"

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -239,6 +239,43 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
+   * @return this
+   */
+  public JibContainerBuilder setVolumes(List<String> volumes) {
```
coollog: I think we should have this take `List<AbsoluteUnixPath>`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -239,6 +239,43 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
+    containerConfigurationBuilder.setVolumes(volumes);
+    return this;
+  }
+
+  /**
+   * Sets directories where the process is likely to write. This replaces any previously-set volume.
```
coollog: This javadoc should be the same as the one that takes the `List`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -239,6 +239,43 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
+   * @param volumes the directory to be create as a volume
+   * @return this
+   * @see #setVolumes(List) for more details
+   */
+  public JibContainerBuilder setVolumes(String... volumes) {
+    return setVolumes(Arrays.asList(volumes));
+  }
+
+  /**
+   * Adds a directory where the process is likely to write.
```
coollog: Same here - javadoc should be the same except with `Adds` rather than `Sets`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -239,6 +239,43 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
+   *
+   * @param volume a directory to represent a volume.
+   * @return this
+   * @see #setVolumes(List) for more details
+   */
+  public JibContainerBuilder addVolumes(String volume) {
```
coollog: Should be singular - `addVolume`

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -120,6 +121,29 @@ public void addExposedPort(Port port) {
       exposedPorts.add(port);
     }
 
+    /**
+     * Sets the container's volumes.
+     *
+     * @param volumes the list of volumes
+     * @return this
+     */
+    public Builder setVolumes(@Nullable List<String> volumes) {
```
coollog: `AbsoluteUnixPath` here as well

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -124,6 +125,17 @@
       return this;
     }
 
+    /**
+     * Sets the items in the "Volumes" field in the container configuration.
+     *
+     * @param volumes the list of directories to create a volume.
+     * @return this
+     */
+    public Builder<T> setVolumes(@Nullable List<String> volumes) {
```
coollog: `AbsoluteUnixPath`

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -223,6 +236,9 @@
   /** Ports that the container listens on. */
   @Nullable private final ImmutableList<Port> exposedPorts;
 
+  /** List of directories to mount as volume. */
```
coollog: `volumes`

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -114,6 +115,9 @@
 
     /** User. */
     @Nullable private String User;
+
+    /** Volumes */
+    @Nullable private Map<String, Map<?, ?>> Volumes;
```
coollog: Everywhere else should be `AbsoluteUnixPath` for the volumes except here - this should be `String` - following how exposed ports work in terms of when and where to convert over / check and validate should be the right way to go.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -61,11 +62,22 @@
     if (exposedPorts == null) {
       return null;
     }
+
+    return listToMap(
+        exposedPorts
+            .stream()
+            .map(port -> port.getPort() + "/" + port.getProtocol())
```
coollog: I think an alternative way to do this is to have a list to map like:

```
static <E> Map<String, Map<?, ?>> listToMap(@Nullable List<E> list, Function<E, String> listElementToString)
```

where this method will directory turn the list of E into a map of string to `{}`, where each element E converts to String using the `listElementToString` function.

This can avoid what is going on here where the list is turned into a stream, then mapped, then collected into a list, then iterated over with foreach to place into an `ImmutableSortedMap`. ([`ImmutableSortedMap.toImmutableSortedMap`](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/ImmutableSortedMap.html#toImmutableSortedMap-java.util.Comparator-java.util.function.Function-java.util.function.Function-) may be useful here.)

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
GuustavoPaiva: Yep, that's is better. changed.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
GuustavoPaiva: Changed.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
GuustavoPaiva: Done, changed everywhere to `AbsoluteUnixPath` except here, doing so I've found that I missed the `JsonToImageTranslator.java` which is used for pulling the base image. 

Also, when doing this, I was wondering if exposed ports and volumes from the base image should be propagated to the final image, maybe this should be another issue?  I think volumes makes more sense than ports, but maybe we should have support for both. 

What do you think @coollog

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
GuustavoPaiva: Thanks for the suggestion. I'm not familiarized with Immutable collections specific methods, that's nice to know.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
coollog: Yep! We actually have another issue tracking the base image configuration propagation: #595 If you would like, feel free to mention this on that issue!

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
coollog: I honestly wasn't aware of `ImmutableSortedMap.toImmutableSortedMap` till I looked it up haha - learning new things everyday.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -240,39 +240,38 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
    */
-  public JibContainerBuilder setVolumes(String... volumes) {
-    return setVolumes(Arrays.asList(volumes));
+  public JibContainerBuilder setVolumes(AbsoluteUnixPath... volumes) {
+    return setVolumes(new ArrayList<>(Arrays.asList(volumes)));
   }
 
   /**
-   * Adds a directory where the process is likely to write.
+   * Adds a directory that may hold externally mounted volume.
```
coollog: tiny nit: hold *an* externally..

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -240,39 +240,38 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
   /**
-   * Sets the directories that the container will create as a volume. Volumes allows persistent
-   * data.
+   * Sets the directories that may hold externally mounted volumes.
    *
    * <p>This is similar to <a href="https://docs.docker.com/engine/reference/builder/#volume">{@code
    * VOLUME} in Dockerfiles</a>.
    *
-   * @param volumes the list of volumes
+   * @param volumes the list of directories to set as volumes
```
coollog: Actually, how about changing "directories" to "directory paths on the container filesystem"

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -59,26 +60,34 @@
...
+  static <E> Map<String, Map<?, ?>> listToMap(
+      @Nullable List<E> list, Function<E, String> elementMapper) {
     if (list == null) {
       return null;
     }
-    ImmutableSortedMap.Builder<String, Map<?, ?>> result =
-        new ImmutableSortedMap.Builder<>(String::compareTo);
-    list.forEach(value -> result.put(value, Collections.emptyMap()));
-    return result.build();
+    return list.stream()
```
coollog: Let's add a short comment like:

```
// Turns the list into a sorted map where each element of the list maps to an empty map.
```

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -59,26 +60,34 @@
...
+   */
+  @VisibleForTesting
+  @Nullable
+  static Map<String, Map<?, ?>> volumesListToMap(@Nullable List<AbsoluteUnixPath> volumes) {
+    return listToMap(volumes, AbsoluteUnixPath::toString);
   }
 
   @Nullable
-  static Map<String, Map<?, ?>> listToMap(@Nullable List<String> list) {
+  static <E> Map<String, Map<?, ?>> listToMap(
```
coollog: Let's make this private unless we use it in tests.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -59,26 +60,34 @@
...
       return null;
     }
-    ImmutableSortedMap.Builder<String, Map<?, ?>> result =
-        new ImmutableSortedMap.Builder<>(String::compareTo);
-    list.forEach(value -> result.put(value, Collections.emptyMap()));
-    return result.build();
+    return list.stream()
+        .collect(
+            ImmutableSortedMap.toImmutableSortedMap(
+                String::compareTo, elementMapper, key -> Collections.emptyMap()));
```
coollog: In general, for unused variables, we name them `ignored` to avoid any confusion - s/`key`/`ignored`/

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
coollog: Might want to make a note that this method exists because the JSON form ["is a direct JSON serialization of the Go type map[string]struct{} and is represented in JSON as an object mapping its keys to an empty object"](https://github.com/opencontainers/image-spec/blob/master/config.md).

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -196,5 +202,25 @@
     return ports.build();
   }
 
+  /**
+   * Converts a map of volumes strings to a list of {@link AbsoluteUnixPath}s * (e.g. {@code {@code
```
coollog: Looks like there's an extra `*` in this line.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -196,5 +202,25 @@
...
+  @VisibleForTesting
+  static ImmutableList<AbsoluteUnixPath> volumeMapToList(
+      @Nullable Map<String, Map<?, ?>> volumeMap) {
+    if (volumeMap == null) {
+      return ImmutableList.of();
+    }
+    return volumeMap
+        .keySet()
+        .stream()
+        .map(AbsoluteUnixPath::get)
```
coollog: I think we might want to catch the `IllegalArgumentException` thrown by `AbsoluteUnixPath.get` and wrap that with a `BadContainerConfigurationFormatException` like how `portMapToList` does it.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
GuustavoPaiva: changed.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
GuustavoPaiva: That's sounds better indeed.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
GuustavoPaiva: I've added this info in the javadoc, also added the info that the keys of the maps are generated by the `keyMapper` parameter.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
GuustavoPaiva: sure.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
GuustavoPaiva: changed.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
GuustavoPaiva: Added this on the javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
GuustavoPaiva: removed.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
GuustavoPaiva: I've added this verification and some unit tests.

jib-core/src/test/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslatorTest.java
```diff
@@ -136,6 +141,26 @@ public void testPortListToMap() {
...
+        ImmutableList.of(
+            AbsoluteUnixPath.get("/var/job-result-data"),
+            AbsoluteUnixPath.get("/var/log/my-app-logs"));
+    ImmutableSortedMap<String, Map<?, ?>> expected =
+        ImmutableSortedMap.of(
+            "/var/job-result-data", ImmutableMap.of(), "/var/log/my-app-logs", ImmutableMap.of());
+    Assert.assertEquals(expected, ImageToJsonTranslator.volumesListToMap(input));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
coollog: In our codebase, we try to avoid using `expected` and manually catch the exception in the test method body so that in most cases, we can assert the message and/or cause of the exception is as expected and that the exception was thrown from the expected location.

jib-core/src/test/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslatorTest.java
```diff
@@ -136,6 +141,26 @@ public void testPortListToMap() {
...
+  public void testInvalidVolumeListToMap() {
+    ImmutableList<AbsoluteUnixPath> input =
+        ImmutableList.of(AbsoluteUnixPath.get("'/var/job-result-data"));
+
+    ImageToJsonTranslator.volumesListToMap(input);
```
coollog: So for example, here, we actually have a line that would never be reached since the test actually throws the `IllegalArgumentException` on line 159.

GuustavoPaiva: Actually this test doesn't even make sense. There is no way to get an invalid `AbsoluteUnixPath` instance. Don't know what I tried to do here, removing this method.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -196,5 +202,34 @@
...
+  @VisibleForTesting
+  static ImmutableList<AbsoluteUnixPath> volumeMapToList(@Nullable Map<String, Map<?, ?>> volumeMap)
+      throws BadContainerConfigurationFormatException {
+    if (volumeMap == null) {
+      return ImmutableList.of();
+    }
+
+    for (String volume : volumeMap.keySet()) {
+      try {
+        AbsoluteUnixPath.get(volume);
```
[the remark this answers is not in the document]
GuustavoPaiva: Changed to `ImmutableList.Builder`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -58,15 +60,52 @@
...
+   * maps.
+   *
+   * <p>This method is needed because the volume object is a direct JSON serialization of the Go
+   * type map[string]struct{} and is represented in JSON as an object mapping its keys to an empty
+   * object.
+   *
+   * <p>Further read at the <a
+   * href="https://github.com/opencontainers/image-spec/blob/master/config.md">image specs.</a>
+   *
+   * @param list the list of elements to be transformed.
```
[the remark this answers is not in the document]
GuustavoPaiva: Gotcha.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -58,15 +60,52 @@
...
+  @VisibleForTesting
+  @Nullable
+  static Map<String, Map<?, ?>> volumesListToMap(@Nullable List<AbsoluteUnixPath> volumes) {
+    return listToMap(volumes, AbsoluteUnixPath::toString);
+  }
+
+  /**
+   * Turns a list into a sorted map where each element of the list is mapped to an entry composed by
+   * the key generated with {@code Function<E, String> elementMapper} and an empty map as value.
+   * maps.
```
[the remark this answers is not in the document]
GuustavoPaiva: Removed.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -216,19 +216,16 @@
       return ImmutableList.of();
     }
 
+    ImmutableList.Builder<AbsoluteUnixPath> listBuilder = ImmutableList.builder();
```
coollog: tiny nit: Generally, would prefer a more descriptive name here like `volumeList`. Less descriptive names may be fine, but this variable spans 10 lines, so would help readers remember what the list contained by line 228.

GuustavoPaiva: Thanks for the tips @coollog and the review! 

I'll try my max to practice more these conventions.

Issue #676, opened 2018-07-20
support configuring a Docker HEALTHCHECK


jack-kerouac:
**Description of the issue**:
Support configuring a [Docker HEALTHCHECK](https://docs.docker.com/engine/reference/builder/#healthcheck) for the resulting Docker image.

**Expected behavior**:
The Jib configuration should allow adding a Docker `HEALTHCHECK`, including `interval`, `timeout`, `start-period`, `retries`, and `CMD` to run.

werty1st:
Hello @coollog, i would like to question your assumtion, that by design, a shell is needed to run a HEALTHCHECK.

The docs state that 
> The command after the CMD keyword can be either a shell command (e.g. HEALTHCHECK CMD /bin/check-running) or an exec array (as with other Dockerfile commands; see e.g. ENTRYPOINT for details).

So why should 
`HEALTHCHECK ["java", "-cp", "/app/resources/:/app/classes/:/app/libs/*", "com.company.content.feed.consumer.ContentFeedConsumerApplication", "health"]`
not work?

swift1911:
HEALTHCHECK is very important for check container is healthy or not , especially for java service

tburch:
This would be super useful to have with [ECS health checks](https://aws.amazon.com/about-aws/whats-new/2018/03/amazon-ecs-supports-container-health-checks-and-task-health-mana/). I'd love to see support added.

PR #1217, opened 2018-11-05
Add Healthcheck to ContainerConfigurationTemplate

TadCordle:
This doesn't change any behavior yet, just prepares for #676.

Adds `Healthcheck` fields to the container configuration json template + `Image` class. This field is left empty for OCI images since OCI does not support healthcheck currently.


coollog:
I think we could get this in even if we decide to not support it in the plugins since it still gives the Jib Core users control over this portion of the Docker container config.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -113,6 +117,51 @@
       return this;
     }
 
+    /**
+     * Sets the command used to check the container's health.
+     *
+     * @param healthTest the list of healthcheck command tokens
+     * @return this
+     */
+    public Builder<T> setHealthTest(@Nullable List<String> healthTest) {
```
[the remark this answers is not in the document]
coollog: Yea I agree - sort of like how we have it for `Port` so that we can encapsulate it in this object.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -129,24 +130,36 @@
...
      *
-     * @param healthInterval the number of nanoseconds to wait between healthcheck attempts
+     * @param healthInterval the time duration in the given {@code sourceUnit} to wait between
+     *     healthcheck attempts
+     * @param timeUnit the unit of the {@code healthInterval} argument
      * @return this
      */
-    public Builder<T> setHealthInterval(@Nullable Long healthInterval) {
-      this.healthInterval = healthInterval;
+    public Builder<T> setHealthInterval(@Nullable Long healthInterval, TimeUnit timeUnit) {
```
coollog: Not sure what the best practice is here but would `Duration` make more sense here? (It's in the `java.time` package as opposed to the `java.util.concurrent` package for `TimeUnit`)

coollog: @elharo Would you recommend `(long, TimeUnit)` for the signature or `(Duration)`?

TadCordle: Yeah I like this more.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -44,11 +47,16 @@
     @Nullable private Instant created;
     @Nullable private ImmutableList<String> entrypoint;
     @Nullable private ImmutableList<String> programArguments;
+    @Nullable private DockerHealthCheck healthCheck;
```
coollog: Should healthcheck be nullable still considering there's now two ways to say inherit healthcheck (set to `null` or build with `DockerHealthCheck.builderWithInheritedCommand().build()`)?

TadCordle: Yeah I'm trying to decide between making it nullable in place of either `disabled()` or `builderWithInheritedCommand().build()` vs. making it not nullable and instantiating with a default value. I think you're right though and using a default will be simpler.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -158,33 +159,31 @@ public Blob getContainerConfigurationBlob() {
...
-    // Sets the labels.
     template.setContainerLabels(image.getLabels());
-
-    // Sets the working directory.
     template.setContainerWorkingDir(image.getWorkingDirectory());
-
-    // Sets the user.
     template.setContainerUser(image.getUser());
 
+    // Healthcheck is not supported by OCI
```
coollog: I think we should have this be `== V22ManifestTemplate.class` since healthcheck is only supported by the Docker V2.2 format.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -158,33 +159,31 @@ public Blob getContainerConfigurationBlob() {
...
+    if (image.getImageFormat() != OCIManifestTemplate.class) {
+      DockerHealthCheck healthCheck = image.getHealthCheck();
+      template.setContainerHealthTest(healthCheck.getCommand().orElse(ImmutableList.of()));
+      template.setContainerHealthInterval(
```
coollog:
```
healthCheck.getInterval().ifPresent(interval -> template.setContainerHealthInterval(interval.toNanos()));
```

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -158,33 +159,31 @@ public Blob getContainerConfigurationBlob() {
...
-
-    // Sets the working directory.
     template.setContainerWorkingDir(image.getWorkingDirectory());
-
-    // Sets the user.
     template.setContainerUser(image.getUser());
 
+    // Healthcheck is not supported by OCI
+    if (image.getImageFormat() != OCIManifestTemplate.class) {
+      DockerHealthCheck healthCheck = image.getHealthCheck();
```
coollog: This seems to always add a healthcheck even if the user did not configure one - is this the intended behavior?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -152,6 +184,41 @@ public void setContainerCmd(@Nullable List<String> cmd) {
     config.Cmd = cmd;
   }
 
+  public void setContainerHealthTest(@Nullable List<String> test) {
+    if (config.Healthcheck == null) {
+      config.Healthcheck = new HealthCheckObjectTemplate();
```
coollog: This looks like you can never go back to the `null` `config.Healthcheck` state. Might be fine though.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
TadCordle: Ah, this was from when `healthCheck` was still `@Nullable`, I'll fix

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
TadCordle: Yeah I think this is fine given our use case for this class.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/DockerHealthCheck.java
```diff
@@ -211,4 +211,15 @@ private DockerHealthCheck(
   public Optional<Integer> getRetries() {
     return Optional.ofNullable(retries);
   }
+
+  /**
+   * Returns {@code true} if the entire health check is inherited from the base image, {@code false}
```
coollog: javadoc seems outdated

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/DockerHealthCheck.java
```diff
@@ -211,4 +211,15 @@ private DockerHealthCheck(
...
+   * if not.
+   *
+   * @return {@code true} if the entire health check is inherited from the base image, {@code false}
+   *     if not
+   */
+  public boolean hasCommand() {
+    return command != null && !command.isEmpty();
```
coollog: I think a `DockerHealthcheck` not having a command should not be a valid `DockerHealthcheck`. For example, if someone decided to create a `DockerHealthcheck` with no command but other values set, rather than not doing anything (which would be different from what the user expected), it should error. I'm thinking that defining an inherited healthcheck shouldn't be a healthcheck with no command, but rather just not having a healthcheck (`null`), since in a `Dockerfile`, `HEALTHCHECK` can't be defined with no command.

TadCordle: I personally feel like we should avoid using a `null` healthcheck instead of an explicit way of configuring whether or not to inherit, since it's a bit ambiguous as to whether or not `null` means to inherit or disable healthcheck. Will update the proposal doc or make a new one to get team consensus about what to do here.

TadCordle:
I'm actually realizing now that this API design is probably not the best if we aren't exposing this to library users; if all we're doing is propagating from base image, using different builders instead of just a setter to propagate the command is going to be a pain. So I'm going to pause on this until we figure that out.

TadCordle:
Actually, we can probably move forward with this so we can implement the base image propagation, and simplify later if necessary when we have a final decision on jib-core support.

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -44,11 +47,16 @@
     @Nullable private Instant created;
     @Nullable private ImmutableList<String> entrypoint;
     @Nullable private ImmutableList<String> programArguments;
+    private DockerHealthCheck healthCheck = DockerHealthCheck.disabled();
```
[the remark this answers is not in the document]
TadCordle: Yeah, this is from when the initial design which disabled it by default, I'll change it to inherit.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ImageToJsonTranslator.java
```diff
@@ -79,8 +80,32 @@
   }
 
   /**
-   * Turns a set into a sorted map where each element of the set is mapped to an entry composed by
-   * the key generated with {@code Function<E, String> elementMapper} and an empty map as value.
+   * <<<<<<< HEAD Converts the map of environment variables to a list with items in the format
```
coollog: Looks like a merge conflict left here.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/DockerHealthCheck.java
```diff
@@ -29,13 +30,16 @@
...
     @Nullable private Duration interval;
     @Nullable private Duration timeout;
     @Nullable private Duration startPeriod;
     @Nullable private Integer retries;
 
-    private Builder(@Nullable ImmutableList<String> command) {
+    private Builder(ImmutableList<String> command) {
+      Preconditions.checkArgument(
+          command.size() > 1 && !command.contains(""),
+          "command must not be empty and must not contain empty tokens");
```
chanseokoh: For conditions joined with `&&`, prefer separating them. It helps identify which condition has been violated.

And maybe this should be `!command.isEmpty()` or the error message needs to change?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -152,6 +184,41 @@ public void setContainerCmd(@Nullable List<String> cmd) {
     config.Cmd = cmd;
   }
 
+  public void setContainerHealthCheckTest(@Nullable List<String> test) {
+    if (config.Healthcheck == null) {
+      config.Healthcheck = new HealthCheckObjectTemplate();
+    }
+    Preconditions.checkNotNull(config.Healthcheck).Test = test;
```
chanseokoh: Do we need these `checkNotNull`s? You already checked it above.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -152,6 +184,41 @@ public void setContainerCmd(@Nullable List<String> cmd) {
     config.Cmd = cmd;
   }
 
+  public void setContainerHealthCheckTest(@Nullable List<String> test) {
```
chanseokoh: Just wondering if it's safe to accept null, because I see you had the precondition checks for the `DockerHealthCheck` command.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -68,7 +68,7 @@
    */
   public static Image<Layer> toImage(V21ManifestTemplate manifestTemplate)
       throws LayerPropertyNotFoundException {
-    Image.Builder<Layer> imageBuilder = Image.builder();
+    Image.Builder<Layer> imageBuilder = Image.builder(V22ManifestTemplate.class);
```
chanseokoh: Just checking: `manifestTemplate` is `V21ManifestTemplate`, but here you are using `V22ManifestTemplate`. Correct?

TadCordle: Yeah, the parameter passed into the image builder is the target image format, and we don't build v21 images.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
TadCordle: I think NullAway only checks the control flow, so it still gives me a compile error because I don't return after the null check (it doesn't realize `Healthcheck` is being instantiated if it's null).

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/DockerHealthCheck.java
TadCordle: It needs to check if size is >1, not isEmpty, because "CMD" or "CMD-SHELL" is added before it, and a command containing "CMD"/"CMD-SHELL" with no other arguments isn't valid.

loosebazooka: Should we be checking if `command.stream().anyMatch(str -> str.trim().isEmpty)` or something? Not just against the empty string but the whitespace string?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/DockerHealthCheck.java
```diff
@@ -29,13 +30,15 @@
...
+    private final ImmutableList<String> command;
     @Nullable private Duration interval;
     @Nullable private Duration timeout;
     @Nullable private Duration startPeriod;
     @Nullable private Integer retries;
 
-    private Builder(@Nullable ImmutableList<String> command) {
+    private Builder(ImmutableList<String> command) {
+      Preconditions.checkArgument(command.size() > 1, "command must not be empty");
+      Preconditions.checkArgument(!command.contains(""), "command must not contain empty tokens");
```
loosebazooka: Should we be checking if command.stream().anyMatch(str -> str.trim().isEmpty) or something? Not just against the empty string but the whitespace string?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/DockerHealthCheck.java
TadCordle: Docker allows it.

Dockerfile:
```Dockerfile
HEALTHCHECK CMD [" "]
```

`docker inspect`:

```
"Healthcheck": {
  "Test": [
    "CMD",
    " "
  ]
},
```

TadCordle: Ah wait, it takes empty strings too for CMD, just not CMD-SHELL. I think I have to move the precondition checks.

PR #1221, opened 2018-11-07
Disables Apache HTTP logging at registry client level.

coollog:
This moves the disableHttpLogging logic to the registry client level under `registry.Connection`, where the Apache `HttpTransport` is set up. The implications of this are:

- Jib Core does not show up with annoying logs anymore
- Removes all the disable logging logic from Maven side
- Gradle still needs its own disable logging logic due to how the logging is handled in Gradle

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
@@ -78,6 +78,13 @@ public class Connection implements Closeable {
     return url -> new Connection(url, transport);
   }
 
+  static {
+    // Disables annoying Apache HTTP client logging.
+    System.setProperty(
+        "org.apache.commons.logging.Log", "org.apache.commons.logging.impl.SimpleLog");
+    System.setProperty("org.apache.commons.logging.simplelog.defaultlog", "error");
+  }
+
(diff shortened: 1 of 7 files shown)


chanseokoh:
From the API perspective, it feels weird that just loading or instantiating a util class will define some external system properties. It will have the effect at the JVM level for all jib-core users. Some systems may actually want to use different values for the properties? And generally speaking, as static initialization, it is almost impossible to know when the properties will be defined.

coollog:
Hmm, perhaps we should only set these properties if the properties are `null`? In terms of when the properties are defined, I think this should be okay since they will definitely be defined before `get*ConnectionFactory` is called.

chanseokoh:
> In terms of when the properties are defined,

Yeah, we'll be fine, but I was talking in a more general context. For example, some application or system using jib-core may define the properties at some point themselves. They may be null for a while but defined at some later point, which is like what we did before in plugin tasks and MOJOs. So even if we set the properties when they are null, this is unpredictable in general.

It just feels weird that using a library will override some properties that are widely used to some specific values, by just using the library and that it's unpredictable when it will override them. Setting the commons-logging properties to certain values seems commonplace in practice. Not sure what is the best way for us though.

coollog:
Filed #1224 to address the above comments.

PR #1224, opened 2018-11-07
Disables Apache Commons Logging with properties file.

coollog:
From https://github.com/GoogleContainerTools/jib/pull/1221#issuecomment-436794824

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
@@ -78,13 +78,6 @@ public class Connection implements Closeable {
     return url -> new Connection(url, transport);
   }
 
-  static {
-    // Disables annoying Apache HTTP client logging.
-    System.setProperty(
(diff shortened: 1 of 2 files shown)


coollog:
@chanseokoh 

I tried setting up a project that uses `jib-core` and it disables the logging by default if I don't use my own logging. However, when I add a call to log with commons logging, it seems to override the `jib-core` `commons-logging.properties` and displays all the log messages.

From https://commons.apache.org/proper/commons-logging/guide.html#Configuration:

```
Configuration attributes can be set explicitly by java code, but they are more commonly set by placing a file named commons-logging.properties in the classpath. When such a file exists, every entry in the properties file becomes an "attribute" of the LogFactory. When there is more than one such file in the classpath, releases of commons-logging prior to 1.1 simply use the first one found. From release 1.1, each file may define a priority key, and the file with the highest priority is used (no priority definition implies priority of zero). When multiple files have the same priority, the first one found is used.
```

It looks like if the user wants to explicitly override ours, they can specify their own `commons-logging.properties` with a higher priority.

PR #1243, opened 2018-11-12
Define run.main for the main class to launch

### LINKED ISSUE #1242: Micronaut sample won't run locally without main class arg
When running `./gradlew run` locally, it won't figure which class to run.
Adding a run.main argument in `build.gradle` will solve this.

glaforge:
When running the Micronaut application locally, with `./gradlew run`, a main class should be specified so the JavaExec task knows which main class to run.

This should fix #1242


TadCordle:
Thanks for the fix!

PR #1244, opened 2018-11-12
Fix NullPointerException with incomplete auth config

### LINKED ISSUE #1177: NPE on Maven if only one of username or password is set
```xml
        <configuration>
          <to>
            <image>...</image>
            <auth><username>abc</username></auth>
          </to>
        </configuration>
```
We have code to give a nice error message If either username or password is set only:
```java
    if (Strings.isNullOrEmpty(auth.getUsername())) {
      eventDispatcher.dispatch(
          LogEvent.warn(
              auth.getUsernamePropertyDescriptor()
                  + " is missing from build configuration; ignoring auth section."));
      return Optional.empty();
    }
```
However, the `checkNotNull` inside `auth.getUsernamePropertyDescriptor()` is failing with the sample config above. Note we set the property descriptor in the `JibPluginConfiguration` constructor.
```java
  JibPluginConfiguration() {
    to.auth.setPropertyDescriptor("<to><auth>");
    from.auth.setPropertyDescriptor("<from><auth>");
  }
```
Therefore, when `JibPluginConfiguration` is created, the property descriptors cannot possible by null. So, I think it may be that if the `<auth>` section is defined, a new `AuthConfiguration` instance is created, replacing the one that has property descriptors set by the `JibPluginConfiguration`?

```
Caused by: java.lang.NullPointerException
        at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:871)
        at com.google.cloud.tools.jib.maven.JibPluginConfiguration$AuthConfiguration.getPasswordPropertyDescriptor(JibPluginConfiguration.java:59)
        at com.google.cloud.tools.jib.plugins.common.ConfigurationPropertyValidator.getImageCredential(ConfigurationPropertyValidator.java:99)
        at com.google.cloud.tools.jib.maven.BuildImageMojo.execute(BuildImageMojo.java:107)
        at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
        ... 21 more
```

Haven't tested Gradle.

TadCordle:
Fixes #1177.

`PluginConfigurationProcessor` is getting messy so I think we should figure out how to reduce the number of parameters in each method soon.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/AuthParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/AuthParameters.java
@@ -17,11 +17,8 @@
 package com.google.cloud.tools.jib.gradle;
 
 import com.google.cloud.tools.jib.plugins.common.AuthProperty;
-import com.google.common.base.Preconditions;
 import javax.annotation.Nullable;
-import javax.inject.Inject;
 import org.gradle.api.tasks.Input;
-import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.Optional;
 
 /**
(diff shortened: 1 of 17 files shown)


TadCordle:
Sounds good, filed #1245 for that.

Issue #1251, opened 2018-11-15
Better "bootstrap experience"


dgageot:
This issue sums up some feedback shared by @arun-gupta:

He likes Jib a lot. He also thinks the “bootstrap experience” can be improved a lot.
What can improve to make jib work better the first time someone’s using it.

Issues he shared are:
 + Why do we have to install a credential helper if all we do is push to the docker hub and everything is properly setup in ~/.docker/config.json?
 + Can we print some feedback on progress when an image is being pushed? Especially the first time since it’s obviously going to take longer than subsequent times.
 + Should the image really be pushed by default? Let’s say you are trying Jib without the auth being configured, it’s really annoying that the build is failing.

arun-gupta:
That summarizes the experience very well. Thanks @dgageot 

Wondering if @chanezon has played with this and have ideas on how to simplify the `docker login` experience?

chanseokoh:
Thanks for the feedback.

You don't need to install a cred helper if everything is set up in ~/.docker/config.json. If it is not working, it is a bug.

Regarding the push, we should consider that a lot of people are using Jib primiarly because there is no docker installed locally (or in a CI/CD pipeline).

briandealwis:
You shouldn't need to do anything for Docker Hub.  I tried from a dummy account:

```
$ docker logout
Warning: failed to get default registry endpoint from daemon (Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get http://%2Fvar%2Frun%2Fdocker.sock/v1.38/info: dial unix /var/run/docker.sock: connect: permission denied). Using system default: https://index.docker.io/v1/
Not logged in to https://index.docker.io/v1/

$ ls ~/.docker
ls: /Users/dummy/.docker: No such file or directory

$ cd examples/hellowworld

$ git diff
diff --git a/examples/helloworld/pom.xml b/examples/helloworld/pom.xml
index 22e8e3c..60bd911 100644
--- a/examples/helloworld/pom.xml
+++ b/examples/helloworld/pom.xml
@@ -37,6 +37,7 @@
         <artifactId>jib-maven-plugin</artifactId>
         <version>${jib-maven-plugin.version}</version>
         <configuration>
+         <from><image>openjdk:11-slim</image></from>
           <to>
             <image>gcr.io/REPLACE-WITH-YOUR-GCP-PROJECT/image-built-with-jib</image>
           </to>

[...]

Issue #806, opened 2018-08-07
 mvn compile jib:build freezing after Finalizing 


mushtaqck:
**Description of the issue**: mvn compile jib:build freezing after Finalizing 58mb container takes 11 min to upload to ECS.

**Expected behavior**: Some message is excepted while upload is in progress. 

**Steps to reproduce**: 
mvn clean
mvn compile jib:build

**Environment**: maven

**`jib-maven-plugin` Configuration**:<plugin>
				<groupId>com.google.cloud.tools</groupId>
				<artifactId>jib-maven-plugin</artifactId>
				<version>0.9.7</version>
				<configuration>
			               <to>
						<image>XXXXXXX.dkr.ecr.eu-west-1.amazonaws.com/name-jib</image>
						<credHelper>ecr-login</credHelper>
					</to>
					<mainClass>com.aws.transformation.Main</mainClass>
				</configuration>
			</plugin>

**Log output**: 
[INFO] Finalizing...
[DEBUG] TIMING  Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@fa0bad6
[DEBUG] TIMING  Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@64a82d82
[DEBUG] TIMING  Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@a44292c0
[DEBUG] RUNNING Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@64a82d82
[DEBUG] RUNNING Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@fa0bad6
[DEBUG] TIMING  Pushing container configuration
[DEBUG] RUNNING Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@a44292c0
[DEBUG] RUNNING Pushing container configuration
[DEBUG] TIMED   Pushing container configuration : 31.345 ms
[DEBUG] TIMING  Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@c2d7a77f
[DEBUG] RUNNING Pushing BLOB com.google.cloud.tools.jib.blob.BlobDescriptor@c2d7a77f
[DEBUG] Pushing container configuration : 31.345 ms
[DEBUG] BLOB : com.google.cloud.tools.jib.blob.BlobDescriptor@fa0bad6 already exists on registry
[...]

PR #1262, opened 2018-11-20
Proposal: Provide feedback on push progress.

coollog:
Regards issues: #1251, #806 

Project tracker at #1297

Preview at: https://github.com/GoogleContainerTools/jib/blob/proposal-progress-output/proposals/progress_output.md

Brief recording of the current state of the POC: https://asciinema.org/a/H4BIcjEYsSa8Rtyy0ayZt01Xh

### DIFF
--- a/proposals/progress_output.md
+++ b/proposals/progress_output.md
@@ -14,7 +14,7 @@ This is especially prevalent for first-time users whose first builds would need 
 
 ## Current output
 
-An example of the current output (as of versoin `0.10.0`) looks like:
+An example of the current output (as of version `0.10.0`) looks like:
 
 ```
 Containerizing application to <image>...
@@ -67,9 +67,14 @@ Display an overall progress bar along with the tasks currently being executed.
 ### Example
 
 ```
-Executing tasks: Pushing classes layer, pulling base image layer 50501d3b88f7, pushing dependencies layer, pushing base image layer 8b106a18283f
-[========================         ] 80% complete
+[=====================            ] 60% complete
+> Pushing classes layer
+> Pulling base image layer 50501d3b88f7
+> Pushing dependencies layer
+> Pushing base image layer 8b106a18283f
 ```
+
+The currently executing tasks are displayed below the overall progress bar.
 
 *Note that this would replace the logs messages that are currently outputted as those may corrupt the progress bar display.*
 
@@ -131,6 +136,8 @@ totalProgressAmount += progressUnits / progressTotal * allocationFraction;
 The progress monitor can display a progress bar to the console based on the `totalProgressAmount` upon each update. Using `\r` could work, but results in the console cursor overlapping with the progress bar line. The proposal is to, for each update, move the cursor up a line (`\033[1A`) and print the progress bar line with a newline at the end.
 
 The progress monitor can also keep track of which allocation nodes have completed to display the currently executing tasks.
+
+For Gradle, in order to keep the log messages in sync with Gradle's own log messages (such as the Gradle progress bar), the progress bar display should effectively become a "footer" to the normal log messages Jib outputs and be displayed via the same Gradle Logger interface as the rest of the log messages.
 
 ## Alternative rejected proposal
 
--- a/proposals/progress_output.md
+++ b/proposals/progress_output.md
@@ -4,7 +4,13 @@ Relevant issues: [#806](https://github.com/GoogleContainerTools/jib/issues/806),
 
 ## Motivation
 
-Currently, Jib lacks any progress feedback for layer pushes. In cases where layers take a long time to push, this lack of progress indication is a bad user experience. This is especially prevalent for first-time users whose first builds would need to push all the layers, including the large base image layers.
+Currently, Jib lacks any progress feedback for layer pushes. In cases where layers take a long time to push, this lack of progress indication:
+
+- makes it hard to estimate how long the build will take
+- makes it unclear as to whether the build is stuck or will complete in some reasonable amount of time
+- may deter first-time users from continuing to use Jib 
+
+This is especially prevalent for first-time users whose first builds would need to push all the layers, including the large base image layers.
 
 ## Current output
 
--- a/proposals/progress_output.md
+++ b/proposals/progress_output.md
@@ -101,9 +101,9 @@ Each node in the DAT is immutable and initialized with a count of allocation uni
 
 ##### Example
 
-Consider a root node (`Node A`) with 10 allocation units. A child node (`Node B`) is added with 3 allocation units. Each of these allocation units represents `10% x 30% = 3%` of the entire DAT progress. Completion of all the 3 child node allocation units means that the root node completed 1 allocation unit, or `10%`.
+Consider a root node (`Node A`) with 10 allocation units. A child node (`Node B`) is added with 4 allocation units. Each of these allocation units represents `10% x 25% = 2.5%` of the entire DAT progress. Completion of all the 4 child node allocation units means that the root node completed 1 allocation unit, or `10%`.
 
-New allocations can be added on-the-fly in a decentralized manner. Let's say a new download started for `123000` bytes and that download should be represented by one of the allocation units of `Node B`. We would create a new node (`Node C`) with `123000` allocation units and have its parent be `Node B`. This establishes the `123000` bytes that would represent `3%` of the overall progress without needing to modify other nodes or synchronize with other tasks that may add their own sub-allocations. This also limits the creation of suballocations to only children tasks of the task that created that allocation, making it fully-compatible with a DAG task pipeline.
+New allocations can be added on-the-fly in a decentralized manner. Let's say a new download started for `123000` bytes and that download should be represented by one of the allocation units of `Node B`. We would create a new node (`Node C`) with `123000` allocation units and have its parent be `Node B`. This establishes the `123000` bytes that would represent `2.5%` of the overall progress without needing to modify other nodes or synchronize with other tasks that may add their own sub-allocations. This also limits the creation of suballocations to only children tasks of the task that created that allocation, making it fully-compatible with a DAG task pipeline.
 
 #### Benefits of DAT approach
 
--- a/proposals/progress_output.md
+++ b/proposals/progress_output.md
@@ -1,0 +1,101 @@
+# Proposal: Provide feedback on push progress
+
+Relevant issues: #806, #1251
+
+
+## Motivation
+
+Currently, Jib lacks any progress feedback for layer pushes. In cases where layers take a long time to push, this lack of progress indication is a bad user experience. This is especially prevalent for first-time users whose first builds would need to push all the layers, including the large base image layers.
+
+## Current output
+
+An example of the current output (as of versoin `0.10.0`) looks like:
+
+```
+Containerizing application to <image>...
+warning: Base image 'gcr.io/distroless/java' does not use a specific image digest - build may not be reproducible
+Retrieving registry credentials for gcr.io...
+Getting base image gcr.io/distroless/java...
+Building dependencies layer...
+Building resources layer...
+Building classes layer...
+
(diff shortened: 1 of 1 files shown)


proposals/progress_output.md
```diff
@@ -0,0 +1,106 @@
...
+Finalizing...
+
+Built and pushed image as <image>
+```
+
+Note that `100%` will always be displayed to indicate completion of each layer.
+
+### Alternative considerations
+
+- display byte-completion (eg. `10MB/100MB`)
```
chanseokoh: I like this. Some layers are tiny and others huge. You get the feeling of how much it will take when you know the entire size.

coollog: What do you think about including time elapsed and/or a progress bar as well?

chanseokoh: Adding time sounds fine, but I think the process bar won't help much unless you are doing in-place ncurses-like line replacement. I think we can start simple.

TadCordle: Don't we need to do in-place replacement anyway if we're showing byte completion?

coollog: It'll be just like the percentage completion except rather than a percent, it will show the absolute byte counts.

coollog:
1. Yes, we should have the progress be emitted via progress events.
2. Not sure how we would be able to find which line to update in this case since different lines may print after other lines return to the beginning?
3. So as in we would perform synchronization among the tasks? I feel like we should avoid that since it might unnecessarily add a bit to build times and increase complexity.

loosebazooka:
2. Yeah I'm not sure how clean this would be, but ending a line with `\r` should just move the cursor so any new log message will overwrite it and our next progress will just be on the latest line.
3. Would not be synchronization of the tasks, but smarter handling of the progress on the plugin side.

chanseokoh:
> 3. So as in we would perform synchronization among the tasks? I feel like we should avoid that since it might unnecessarily add a bit to build times and increase complexity.

One way is to do something our internal build tool does: start with known total bytes to process, and as we discover more bytes to process, re-scale the whole progress workload. For example, at first it might look "10MB/100MB", but as you start processing more layers, it may become "11MB/254MB", "12MB/318MB", and so on.

coollog:
Yea, a variable progress bar would work, though it would still mean that there would have to be synchronization of some running total across the tasks though, whereas currently we don't have any explicit shared mutable memory between the concurrent tasks.

loosebazooka:
They all just send their progress over to the progress handler? And it does the math? They don't have to know about eachother.

coollog:
The tasks won't need to know about each other, but the progress handler would need to have synchronized memory to handle the concurrent mutations of its progress/total counters, which would lock the tasks against each other during progress updates.

loosebazooka:
Oh I see, so the event manager isn't dispatching events on a single thread? A *UI Thread*, like swing?

coollog:
The event manager dispatches to a method call on the thread it is dispatched from, although that method could be implemented to execute logic on another thread if needed. However, we probably should avoid having another thread purely for progress monitoring since that could incur a significant cost in context switching for all the progress updates.

chanseokoh:
For updating a global counter, I *think* `AtomicInteger.getAndAdd()` may work, which atomically adds a number in a lock-free and thread-safe manner. No explicit synchronization is required.

From the [Javadoc](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/package-summary.html),

> The specifications of these methods enable implementations to employ efficient machine-level atomic instructions that are available on contemporary processors. However on some platforms, support may entail some form of internal locking. Thus the methods are not strictly guaranteed to be non-blocking -- a thread may block transiently before performing the operation.

I think nowadays with most users using contemporary processors, it will work efficiently without blocking. Even if it does, it should be transient, which I don't think will actually happen in practice. And given the very low frequency of updates (10-second interval or so?), the chances of contention seems extremely low.

chanseokoh:
> Oh I see, so the event manager isn't dispatching events on a single thread? A UI Thread, like swing?

AFAIK, Gradle uses a separate dedicated thread (via `Executors.newSingleThreadExecutor()`) to print logs, while on the Maven side, logs are printed by the thread firing the log event.

proposals/progress_output.md
coollog: Let's take a poll:

*Place your name after the option(s) you would like to vote for*

- byte completion | votes:
- time elapsed | votes:
- progress bar | votes:
- throughput | votes:

coollog:
There's also the issue that we would not be able to log anything else after the progress bar starts and before it finishes, so this would not be compatible with other messages like retrying with auth token. The original "Building..." messages would all be replaced too.

chanseokoh:
Yeah, so I'm actually skeptical with the `\r` approach. (On the other hand, the single, uber progress "bar" can still work without `\r`.)

loosebazooka:
Yeah using `\r` is susceptible to corruption, but the idea is basically something like this

```sh
#!/bin/bash

printf "progress [----    ]\r"
sleep 1
printf "some log statement here\n"
printf "progress [----   ]\r"
sleep 1
printf "some other log statment\n"
printf "progress [----   ]\r"
```

TadCordle:
I think I'm with @loosebazooka on this one, `\r` may be susceptible to corruption but I think we can make it work if we're careful. It may be a little extra work on our end but I think it ends up being a better user experience than just printing more log messages.

coollog:
Updated with design towards the progress bar approach. See [Proposal section](https://github.com/GoogleContainerTools/jib/blob/proposal-progress-output/proposals/progress_output.md#proposal). If looks good, I will start implementing the DAT-based progress events first.

proposals/progress_output.md
```diff
@@ -0,0 +1,177 @@
...
+
+There are a few issues to address for the progress event design:
+
+- the total amount of work is not known beforehand, so there is no static max progress known at any time
+- the builder steps are asynchronous and therefore progress events should not share a single progress state
+- rather, the progress event receiver should be the only potentially stateful entity
+
+These issues can be resolved with a *decentralized allocation tree*.
+
+#### Decentralized allocation tree (DAT)
```
chanseokoh: I read through it, and still not clear if this will work as intended, if what I understood is correct, which makes me think probably I'm missing something.

So, what I can only think of for the intention of a tree is that, basically, the root node corresponds to the top-level entity that kicks off multiple second-level async "steps". For example, if you look at `BuildSteps.forBuildToDockerRegistry`, it spawns 12 steps:

```java
  public static BuildSteps forBuildToDockerRegistry(BuildConfiguration buildConfiguration) {
    return new BuildSteps(
        DESCRIPTION_FOR_DOCKER_REGISTRY,
        buildConfiguration,
        () ->
            new StepsRunner(buildConfiguration)
                .runRetrieveTargetRegistryCredentialsStep()
                .runAuthenticatePushStep()
                .runPullBaseImageStep()
                .runPullAndCacheBaseImageLayersStep()
                .runPushBaseImageLayersStep()
                .runBuildAndCacheApplicationLayerSteps()
                .runBuildImageStep()
                .runPushContainerConfigurationStep()
                .runPushApplicationLayersStep()
                .runFinalizingPushStep()
                .runPushImageStep()
                .waitOnPushImageStep());
  }
```
So you intend that you create a root node here with 12 allocation units, pass the root node to the 12 `Step` instances, and each step will create a single child node (with some allocation units) soon, resulting in 13 nodes (1 root + 12 children) at some point? And each `Step` passes down its child node down to methods and instances they own in the similar manner?

coollog: Yes, that is how it is intended to work. The only thing that will need to known prior to an allocation node being created is the number of allocation units that node should have.

coollog: Oh one thing to note is that the root node with 12 allocation units does not necessarily need to have 12 children - allocation units can be occupied by non-node-based updates (like received bytes in the download progress case).

chanseokoh: If so (passing down the root node to multiple async step instances), and if the root and the children are physically connected (I assume so, because otherwise, it isn't a tree at all), isn't it that you are sharing the tree (at least the root node) concurrently? It sounds like basically you are sharing a single progress state. I thought it'd be cool if anyone could just fire progress events in an isolated manner, not having to establish any connection in relation to others. With no communications with other, it may be impossible to avoid the "backward-moving" progress bar, but I think it's worth the sacrifice for simplicity.

coollog: The node(s) are passed down, but they are all immutable with no progress state. The state is only kept at the receiving end (the progress monitor that receives the progress events). The DAT is essentially just a representation of what fraction of the total progress a progress event actually accounts for. This allows for progress events to be emitted asynchronously and in a decentralized manner, only to be collected by the progress monitor and summed into an overall current progress.

briandealwis:
https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/master/src/main/java/com/google/cloud/tools/managedcloudsdk/ChildProgressListener.java

proposals/progress_output.md
```diff
@@ -0,0 +1,177 @@
...
+- know ahead of time how many layers will be pulled/pushed
+
+## Proposal
+
+Display an overall progress bar along with the tasks currently being executed.
+
+### Example
+
+```
+Executing tasks: Pushing classes layer, pulling base image layer 50501d3b88f7, pushing dependencies layer, pushing base image layer 8b106a18283f
```
TadCordle: It may end up being difficult reading these log messages if we put them on one line/constantly update them; it seems like this layout would be great for serial operations, but when we have a bunch of steps executing at once, showing the log messages like this may not add a lot of value.

proposals/progress_output.md
```diff
@@ -0,0 +1,177 @@
...
+
+- monitor the number of bytes sent (and percentage completion)
+
+And the actions we cannot perform include:
+
+- know ahead of time how many layers will be pulled/pushed
+
+## Proposal
+
+Display an overall progress bar along with the tasks currently being executed.
```
TadCordle: How will this work at different logging levels?

proposals/progress_output.md
coollog: Update: working on prototyping having these on multiple lines - it makes things look a lot clearer

proposals/progress_output.md
coollog: I think we can disable the progress bar for logging levels debug and higher since there will be many log messages to indicate progress anyways, but the progress bar itself should be implemented such that it does not interfere with the other log messages Jib emits.

coollog:
A recording of the WIP POC: https://asciinema.org/a/H4BIcjEYsSa8Rtyy0ayZt01Xh

coollog:
Updated the example and made a note about the things that need to be handled for Gradle logging.

proposals/progress_output.md
```diff
@@ -0,0 +1,184 @@
...
+
+These issues can be resolved with a *decentralized allocation tree*.
+
+#### Decentralized allocation tree (DAT)
+
+Each node in the DAT is immutable and initialized with a count of allocation units. These allocation units are to be claimed by progress made on that node or child nodes. Each child node claims 1 allocation unit of the parent node, meaning that completion of all allocation units on that child node means completion of 1 allocation unit on the parent node.
+
+##### Example
+
+Consider a root node (`Node A`) with 10 allocation units. A child node (`Node B`) is added with 3 allocation units. Each of these allocation units represents `10% x 30% = 3%` of the entire DAT progress. Completion of all the 3 child node allocation units means that the root node completed 1 allocation unit, or `10%`.
```
chanseokoh: Just a nit that I didn't really point out the other day: I think it is `10% x 33.333% = 3.3333%` instead of `10% x 30% = 3%`.

coollog: Fixed and chose 2.5% to keep it non-repeating.

coollog: Fixed and chose 2.5% to keep it non-repeating.

PR #1263, opened 2018-11-20
Allow inheriting ENTRYPOINT and CMD with INHERIT keyword

### LINKED ISSUE #924: Add Support for use CMD to start container

chanseokoh:
Fixes #924.

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Image ID is now written to `build/jib-image.id` ([#1204](https://github.com/GoogleContainerTools/jib/issues/1204))
-- `jib.container.entrypoint = ['INHERIT']` allows inheriting `ENTRYPOINT` and `CMD` from the base image. While inheriting `ENTRYPOINT`, you also override `CMD` using `jib.container.args`.
+- `jib.container.entrypoint = ['INHERIT']` allows inheriting `ENTRYPOINT` and `CMD` from the base image. While inheriting `ENTRYPOINT`, you can also override `CMD` using `jib.container.args`.
 
 ### Changed
 
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Image ID is now written to `build/jib-image.id` ([#1204](https://github.com/GoogleContainerTools/jib/issues/1204))
-- `jib.container.entrypoint = ['INHERIT']` allows inheriting `ENTRYPOINT` and `CMD` from the base image. While inheriting `ENTRYPOINT`, you can also override `CMD` using `jib.container.args`.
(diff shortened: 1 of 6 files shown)


jib-maven-plugin/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Image ID is now written to `target/jib-image.id` ([#1204](https://github.com/GoogleContainerTools/jib/issues/1204))
+- `<container><entrypoint>INHERIT</entrypoint>` allows inheriting `ENTRYPOINT` and `CMD` from the base image. While inheriting `ENTRYPOINT`, you can also override `CMD` using `<container><args>`.
```
coollog: Include `</container>` as well maybe?

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Image ID is now written to `build/jib-image.id` ([#1204](https://github.com/GoogleContainerTools/jib/issues/1204))
+- `jib.container.entrypoint = ['INHERIT']` allows inheriting `ENTRYPOINT` and `CMD` from the base image. While inheriting `ENTRYPOINT`, you can also override `CMD` using `jib.container.args`.
```
coollog: Just realized maybe this should take just a string `'INHERIT'` to indicate inherit rather than the string in a list?

dzmitryh:
Great job guys! Waiting for when it will be merged ;)

jib-gradle-plugin/CHANGELOG.md
chanseokoh: Now `container.entrypoint` can accept a `String`, in which case it will be auto-translated into a singleton list.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -73,6 +73,10 @@ public void setEntrypoint(List<String> entrypoint) {
     this.entrypoint = entrypoint;
   }
 
+  public void setEntrypoint(String entrypoint) {
```
loosebazooka: Maybe this doesn't matter, but `INHERIT` and [`INHERIT`] will both trigger the keyword path?

chanseokoh: Yeah, that's the way it works, which I thought OK. But if you do think `['INHERIT']` shouldn't inherit, I can block it, but I think that'll require larger code changes.

PR #1265, opened 2018-11-20
 Update documentation for 0.10.1

TadCordle:
- Removes docker context generator
- Adds FAQ entries for what a Jib-like Dockerfile looks like and suggestions for inspecting container images

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -15,6 +15,8 @@ If a question you have is not answered below, please [submit an issue](/../../is
 [Can I ADD a custom directory to the image?](#can-i-add-a-custom-directory-to-the-image)\
 [I need to add files generated during the build process to a custom directory on the image.](#i-need-to-add-files-generated-during-the-build-process-to-a-custom-directory-on-the-image)\
 [Can I build to a local Docker daemon?](#can-i-build-to-a-local-docker-daemon)\
+[What would a Dockerfile for a Jib-built image look like?](#what-would-a-dockerfile-for-a-jib-built-image-look-like)\
+[How can I inspect the image Jib built?](#how-can-i-inspect-the-image-jib-built)\
 [I am seeing `ImagePullBackoff` on my pods.](#i-am-seeing-imagepullbackoff-on-my-pods-in-minikube)\
 [How do I configure a proxy?](#how-do-i-configure-a-proxy)\
 [How can I diagnose problems pulling or pushing from remote registries?](#how-can-i-diagnose-problems-pulling-or-pushing-from-remote-registries)\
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -271,6 +273,55 @@ There are several ways of doing this:
...
+A Dockerfile that performs a Jib-like build is shown below:
+
+```Dockerfile
+FROM baseImage
+
+COPY libs /
+COPY snapshot-libs /
+COPY resources /
+COPY classes /
+COPY root /
```
loosebazooka: What's root?

TadCordle: Oh, I could rename this to be more clear. We originally copied the extra directory into a folder called "root" in `jibExportDockerContext`.

docs/faq.md
```diff
@@ -271,6 +273,103 @@ There are several ways of doing this:
...
 - Alternatively, instead of using a Docker daemon, you can run a local container registry, such as [Docker registry](https://docs.docker.com/registry/deploying/) or other repository managers, and point Jib to push to the local registry.
 
+### What would a Dockerfile for a Jib-built image look like?
+
+A Dockerfile that performs a Jib-like build is shown below:
+
+```Dockerfile
+FROM baseImage
+
+COPY libs /
```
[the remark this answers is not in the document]
TadCordle: Hmm, or maybe it'd just be more clear to explicitly reference everything from the project directory. i.e. 

```Dockerfile
COPY build/classes /app/classes
COPY build/libs /app/libs
...
```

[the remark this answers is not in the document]
coollog: +1 that works too - maybe even more explicit directory names?

```
COPY classFiles /app/classes
COPY dependencyJars /app/libs
...
```

Since directories like `build/libs` do not automatically contain all the dependency jars, for example.

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+
+A Dockerfile that performs a Jib-like build is shown below:
+
+```Dockerfile
+# Jib uses distroless java as the default base image
+FROM gcr.io/distroless/java:latest
+
+# Multiple copy statements are used to break the app into layers, allowing for faster rebuilds after 
+# small changes
+COPY dependencyJars /app/libs
```
briandealwis: Should we include a layer for snapshots and another for releases?

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+COPY resources /app/resources
+COPY classFiles /app/classes
+COPY src/main/extra-dir /
```
briandealwis: `src/main/jib` and reference `extraDirectory` in a comment?

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+
+# Sets the app as the container entrypoint with 2 JVM flags and 2 main arguments. Also adds the
+# appropriate files to the classpath.
+ENTRYPOINT ["java", "-Xms512m", "-Xdebug", "-cp", "/app/resources:/app/classes:/app/libs/*", "SomeMainClass"]
```
briandealwis: This implies that we always configure `-Xms512 -Xdebug` How about `["java", container.jvmFlags, "-cp", "/app/...etc", container.mainClass]`

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+CMD ["some", "args"]
```
briandealwis: And `CMD [container.args]`?

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+FROM gcr.io/distroless/java:latest
+
+# Multiple copy statements are used to break the app into layers, allowing for faster rebuilds after 
+# small changes
+COPY dependencyJars /app/libs
+COPY resources /app/resources
+COPY classFiles /app/classes
+COPY src/main/extra-dir /
+
+# Sets the app as the container entrypoint with 2 JVM flags and 2 main arguments. Also adds the
```
briandealwis: Perhaps `# Jib's default entrypoint when container.entrypoint is not set`

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+# appropriate files to the classpath.
+ENTRYPOINT ["java", "-Xms512m", "-Xdebug", "-cp", "/app/resources:/app/classes:/app/libs/*", "SomeMainClass"]
+CMD ["some", "args"]
+```
+
+To achieve similar results with jib, you could use the following configuration in your build configuration. Note that the `container.mainClass` configuration is not actually necessary, since Jib can automatically infer the main class while constructing the container entrypoint.
```
briandealwis: I'm not sure the rest of this is actually helpful.

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+COPY resources /app/resources
+COPY classFiles /app/classes
+COPY src/main/extra-dir /
+
+# Sets the app as the container entrypoint with 2 JVM flags and 2 main arguments. Also adds the
+# appropriate files to the classpath.
+ENTRYPOINT ["java", "-Xms512m", "-Xdebug", "-cp", "/app/resources:/app/classes:/app/libs/*", "SomeMainClass"]
+CMD ["some", "args"]
+```
+
```
briandealwis: _When unset, Jib will infer the value for `container.mainClass`_

docs/faq.md
```diff
@@ -271,6 +273,80 @@ There are several ways of doing this:
...
+</p>
+</details>
+<br>
+
+Some plugins, such as the [Docker Prepare Gradle Plugin](https://github.com/gclayburg/dockerPreparePlugin), will even automatically generate a Docker context for your project, including a Dockerfile.
+
+### How can I inspect the image Jib built?
+
+To inspect the image that is produced from the build using Docker, you can use commands such as `docker inspect your/image:tag` to view the image configuration, or you can also download the image using `docker save` to manually inspect the container image. You can also use tools such as [dive](https://github.com/wagoodman/dive) to analyze the efficiency of your container images.
```
briandealwis: Dive does more than analyze the efficiency, and TBH not sure if that's really a concern of most.  `Other tools, such as dive, provide nicer UI to inspect the image.`?

PR #1266, opened 2018-11-20
Allow configuring working directory

### LINKED ISSUE #1225: add WORKDIR config option
Now there is no config option for the working directory for the container. Why not support it ?

chanseokoh:
Fixes #1225.

Need to update CHANGELOG,

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
@@ -203,8 +203,8 @@ public class ContainerParameters {
   @Nullable
   @Optional
   public String getWorkingDirectory() {
-    if (System.getProperty(PropertyNames.CONTAINER_USER) != null) {
-      return System.getProperty(PropertyNames.CONTAINER_USER);
+    if (System.getProperty(PropertyNames.CONTAINER_WORKING_DIRECTORY) != null) {
+      return System.getProperty(PropertyNames.CONTAINER_WORKING_DIRECTORY);
     }
     return workingDirectory;
   }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
@@ -44,6 +44,7 @@ public class ContainerParameters {
   private Map<String, String> labels = Collections.emptyMap();
   private String appRoot = "";
   @Nullable private String user;
+  @Nullable private String workingDirectory;
 
   @Input
   @Optional
@@ -196,5 +197,19 @@ public class ContainerParameters {
 
   public void setUser(String user) {
     this.user = user;
+  }
+
+  @Input
+  @Nullable
+  @Optional
+  public String getWorkingDirectory() {
+    if (System.getProperty(PropertyNames.CONTAINER_USER) != null) {
+      return System.getProperty(PropertyNames.CONTAINER_USER);
+    }
+    return workingDirectory;
+  }
+
+  public void setWorkingDirectory(String workingDirectory) {
+    this.workingDirectory = workingDirectory;
   }
 }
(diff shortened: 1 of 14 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -197,4 +198,18 @@ public String getUser() {
   public void setUser(String user) {
     this.user = user;
   }
+
+  @Input
+  @Nullable
```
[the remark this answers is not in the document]
chanseokoh: I think it's infeasible to have some concrete default value. Actually, I think will be wrong to set such a value when not configured; we shouldn't touch `WorkDir` in that case. So, I think it's correct to have the default as "no value given" (internally implemented as the `null` value). `BuildImageStep` checks if the value is "not given" (checking `null`), and if so, it doesn't touch the `WorkDir` config. See [line 162 in `BuildImageStep`](https://github.com/GoogleContainerTools/jib/pull/1266/files#diff-156496d3d6522f5171db61c813e2abe7R162).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -243,6 +243,11 @@ static PluginConfigurationProcessor processCommonConfiguration(
             .setExposedPorts(ExposedPortsParser.parse(rawConfiguration.getPorts()))
             .setLabels(rawConfiguration.getLabels())
             .setUser(rawConfiguration.getUser().orElse(null));
+    rawConfiguration
+        .getWorkingDirectory()
+        .ifPresent(
+            workingDirectory ->
+                jibContainerBuilder.setWorkingDirectory(AbsoluteUnixPath.get(workingDirectory)));
```
coollog: We should probably be consistent as to the validation behavior of all the configurations that are `AbsoluteUnixPath` (`workingDirectory`, `appRoot`, `volumes` (#1278)) - currently for `appRoot`, an invalid absolute Unix-style path is thrown as `AppRootInvalidException`, but here it would be thrown as the `IllegalArgumentException`.

chanseokoh: Makes sense. Added `WorkingDirectoryInvalidException`, which follows the pattern of `AppRootInvalidException`. (However, I see `InvalidContainerVolumeException` in #1278 is a bit different.)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -146,7 +147,11 @@ public void buildDocker()
 
     } catch (AppRootInvalidException ex) {
       throw new GradleException(
-          "container.appRoot is not an absolute Unix-style path: " + ex.getInvalidAppRoot());
+          "container.appRoot is not an absolute Unix-style path: " + ex.getInvalidPathValue());
```
coollog: Just noticed - should we include cause `ex` in these?

coollog: Also, looked over `AppRootInvalidException` and it seems that in the constructor, `message` and `invalidAppRoot` are always set to the invalid `appRoot`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
coollog: Hmm, maybe the naming should start with `Invalid...` since in most exception names, the adjective comes first? (ie. `IllegalArgumentException`, `NullPointerException`, `UnsupportedOperationException`)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
chanseokoh: +1 to include cause `ex`.

It is actually intended that the message is set to `appRoot` (like `FileNotFoundException`, whose `message` is simply the missing file path).

coollog: Yea, is the `message` parameter necessary then? I couldn't find a call to the constructor that wasn't `appRoot` repeated twice (`new AppRootInvalidException(appRoot, appRoot, ex)`).

chanseokoh: Not really necessary in our codebase, but I think it makes sense that, conceptually, the general exception message is set to `appRoot` rather than leaving it undefined. If anyone is doing a very general exception handling of displaying a message, it would tell the invalid value.

PR #1278, opened 2018-11-26
Support for volumes on plugins

GuustavoPaiva:
This adds support for both maven and gradle plugins:

Still missing some tests and changelog entries.

This is the final part of #1121.


jib-maven-plugin/src/test/resources/projects/simple/pom.xml
```diff
@@ -52,6 +52,9 @@
               <key1>value1</key1>
               <key2>value2</key2>
             </labels>
+            <volumes>
+              <volume>/var/log</volume>
```
coollog: I would recommend adding another one just to sanity-check the list works as intended.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -262,6 +265,29 @@ static PluginConfigurationProcessor processCommonConfiguration(
...
+  @VisibleForTesting
+  static List<AbsoluteUnixPath> getVolumesList(RawConfiguration rawConfiguration)
+      throws InvalidContainerVolumesException {
+    List<AbsoluteUnixPath> volumes = new ArrayList<>();
+    for (String directory : rawConfiguration.getVolumes()) {
+      try {
+        AbsoluteUnixPath absoluteUnixPath = AbsoluteUnixPath.get(directory);
+        volumes.add(absoluteUnixPath);
+      } catch (IllegalArgumentException exception) {
+        throw new InvalidContainerVolumesException("Invalid volume path: " + directory);
```
coollog: Let's put this message in the constructor for `InvalidContainerVolumeException` (also rename to singular?). Also, the message should probably also include some indication of what a valid volume path is (ie. "(must be an absolute Unix-style path)").

coollog: Should also include the cause (the `IllegalArgumentException`).

coollog:
Thanks for the pull request!

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
GuustavoPaiva: Changed the name, message and added the cause to this exception.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidContainerVolumeException.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.common;
+
+/** Thrown when list of volume's directories contains invalid paths */
+public class InvalidContainerVolumeException extends Exception {
```
coollog: Hi, do you mind making this in the same format as the other exceptions regarding invalid paths? Like https://github.com/GoogleContainerTools/jib/pull/1266/files#diff-a90e603b657d4bf5aca92fe07322a32e - where there is a `getInvalidPathValue` method and the exception is caught at the plugin level so that the parameter name can be in the exception message (`container.volume` or `<container><volume>`).

GuustavoPaiva: No problem at all, I think it looks better now following the same format.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -405,4 +415,26 @@ public void testGetBaseImage_nonDefault() {
...
+  }
+
+  @Test
+  public void testGetInvalidVolumesList() {
+    Mockito.when(rawConfiguration.getVolumes()).thenReturn(Collections.singletonList("`some/root"));
+
+    try {
+      PluginConfigurationProcessor.getVolumesList(rawConfiguration);
+      Assert.fail();
+    } catch (InvalidContainerVolumeException e) {
```
coollog: tiny nit: avoid single-letter variables (use `ex` here)

GuustavoPaiva: done

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -133,6 +134,9 @@ public void buildImage()
           "container.workingDirectory is not an absolute Unix-style path: "
               + ex.getInvalidPathValue(),
           ex);
+    } catch (InvalidContainerVolumeException ex) {
+      throw new GradleException(
+          "container.volumes is not an absolute Unix-style path: " + ex.getInvalidVolume());
```
chanseokoh: Sorry for catching this late. Let's propagate the exception. Here and the other 5 places.

```java
throw new GradleException("..." + ex.getInvalidVolume(), ex);
```

GuustavoPaiva: No problem at all. Added the cause to all places that wrap this exception.

PR #1284, opened 2018-11-28
Change exposed port/volumes API to use sets instead of lists

TadCordle:
[Original comment suggesting this change.](https://github.com/GoogleContainerTools/jib/issues/595#issuecomment-441745538)

Changes API from using lists to using sets for exposed ports and volumes, since these should not contain duplicate values.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -32,8 +32,10 @@ import java.nio.file.Path;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import javax.annotation.Nullable;
 
@@ -245,10 +247,10 @@ public class JibContainerBuilder {
    * <p>This is similar to <a href="https://docs.docker.com/engine/reference/builder/#volume">{@code
    * VOLUME} in Dockerfiles</a>.
    *
-   * @param volumes the list of directory paths on the container filesystem to set as volumes
+   * @param volumes the set of directory paths on the container filesystem to set as volumes
    * @return this
    */
-  public JibContainerBuilder setVolumes(List<AbsoluteUnixPath> volumes) {
+  public JibContainerBuilder setVolumes(Set<AbsoluteUnixPath> volumes) {
     containerConfigurationBuilder.setVolumes(volumes);
     return this;
   }
@@ -258,10 +260,10 @@ public class JibContainerBuilder {
    *
    * @param volumes the list of directory paths on the container filesystem to set as volumes
    * @return this
-   * @see #setVolumes(List) for more details
+   * @see #setVolumes(Set) for more details
    */
   public JibContainerBuilder setVolumes(AbsoluteUnixPath... volumes) {
-    return setVolumes(new ArrayList<>(Arrays.asList(volumes)));
+    return setVolumes(new HashSet<>(Arrays.asList(volumes)));
   }
 
   /**
@@ -269,7 +271,7 @@ public class JibContainerBuilder {
    *
    * @param volume a directory path on the container filesystem to represent a volume.
    * @return this
-   * @see #setVolumes(List) for more details
+   * @see #setVolumes(Set) for more details
    */
   public JibContainerBuilder addVolume(AbsoluteUnixPath volume) {
     containerConfigurationBuilder.addVolume(volume);
@@ -287,10 +289,10 @@ public class JibContainerBuilder {
    * href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#container-v1-core">Kubernetes
    * Container spec</a>.
    *
-   * @param ports the list of ports to expose
+   * @param ports the set of ports to expose
    * @return this
    */
-  public JibContainerBuilder setExposedPorts(List<Port> ports) {
+  public JibContainerBuilder setExposedPorts(Set<Port> ports) {
     containerConfigurationBuilder.setExposedPorts(ports);
     return this;
   }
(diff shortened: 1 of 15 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -251,11 +252,11 @@ static PluginConfigurationProcessor processCommonConfiguration(
             .setProgramArguments(rawConfiguration.getProgramArguments().orElse(null))
             .setEnvironment(rawConfiguration.getEnvironment())
             .setExposedPorts(ExposedPortsParser.parse(rawConfiguration.getPorts()))
-            .setVolumes(getVolumesList(rawConfiguration))
+            .setVolumes(new HashSet<>(getVolumesList(rawConfiguration)))
```
chanseokoh: Why do you need `new HashSet()`, especially given that you didn't need it for `setExposedPorts()`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -125,8 +127,8 @@
      * @param exposedPorts the list of exposed ports to add
      * @return this
      */
-    public Builder<T> setExposedPorts(@Nullable List<Port> exposedPorts) {
-      this.exposedPorts = (exposedPorts == null) ? null : ImmutableList.copyOf(exposedPorts);
+    public Builder<T> setExposedPorts(@Nullable Set<Port> exposedPorts) {
```
chanseokoh: Just saying here, I think eventually this and `setVolumes()` would need to change to `add...()`, like other methods such as `addLabels()`, `addEnvironment()` so that we inherit the base settings and only add new values.

TadCordle: Yeah, was gonna do this when I add/change the inheritance stuff. Is the plan to replace the set methods with add methods, or just add new add methods alongside the set ones?

briandealwis: IMHO, we should keep set and add.  The build plugins should strive to mimic Docker behaviour, but the library should keep an open mind.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -245,10 +247,10 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
    * <p>This is similar to <a href="https://docs.docker.com/engine/reference/builder/#volume">{@code
    * VOLUME} in Dockerfiles</a>.
    *
-   * @param volumes the list of directory paths on the container filesystem to set as volumes
+   * @param volumes the set of directory paths on the container filesystem to set as volumes
```
briandealwis: you could remove the collection type and leave it as _the directory paths on the …_

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -258,18 +260,18 @@ public JibContainerBuilder setVolumes(List<AbsoluteUnixPath> volumes) {
    *
    * @param volumes the list of directory paths on the container filesystem to set as volumes
```
briandealwis: and either change to set or drop the collection type as per my previous comment

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -287,10 +289,10 @@ public JibContainerBuilder addVolume(AbsoluteUnixPath volume) {
    * href="https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#container-v1-core">Kubernetes
    * Container spec</a>.
    *
-   * @param ports the list of ports to expose
+   * @param ports the set of ports to expose
```
briandealwis: _the ports to expose_

jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
chanseokoh: Isn't the `Image` class basically internal? (It's not under `com.google.cloud.tools.jib.api`.) Looks like we won't be using `setExportedPorts()`.

chanseokoh: And since it's the builder method, and you always start with an empty port, even if some library user decided to use it directly. That is, `addExportedPorts()` will effectively be `setExportedPorts()` when called for the first time.

chanseokoh:
> Would it be friendlier to have the method arguments be of type Collection, and have the implementations record the elements in a Set?

Hmm... it can be convenient, but I personally prefer accepting a `Set`. Otherwise, you have to document in the Javadocs that internally duplication will be removed, but even so, people don't usually read Javadocs for the methods that look like simple setters.

TadCordle:
Good point, it's definitely more predictable when it takes a Set instead of a Collection. I'll change it back.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -245,31 +247,31 @@ public JibContainerBuilder addEnvironmentVariable(String name, String value) {
...
    * VOLUME} in Dockerfiles</a>.
    *
-   * @param volumes the list of directory paths on the container filesystem to set as volumes
+   * @param volumes the directory paths on the container filesystem to set as volumes
    * @return this
    */
-  public JibContainerBuilder setVolumes(List<AbsoluteUnixPath> volumes) {
-    containerConfigurationBuilder.setVolumes(volumes);
+  public JibContainerBuilder setVolumes(Set<AbsoluteUnixPath> volumes) {
+    containerConfigurationBuilder.setVolumes(new HashSet<>(volumes));
```
coollog: `new HashSet` isn't necessary since `ContainerConfiguration.Builder#setVolumes` already does the `new HashSet`.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -87,11 +88,6 @@ public void setUp() {
         .thenReturn(containerizer);
   }
 
-  /**
-   * Test with our default mocks, which try to mimic the bare Gradle configuration.
```
coollog: Why was this removed?

PR #1286, opened 2018-11-28
 Propagate exposed ports and volumes from base image

TadCordle:
After #1284.
Part of #595.

Changes `Image#setExposedPorts` and `Image#setVolumes` to `addExposedPorts` and `addVolumes`, and uses this to add configured ports/volumes to existing ports/volumes in the base image.


jib-core/src/main/java/com/google/cloud/tools/jib/image/Image.java
```diff
@@ -138,8 +141,10 @@
      * @param volumes the list of directories to create a volume.
      * @return this
      */
-    public Builder<T> setVolumes(@Nullable Set<AbsoluteUnixPath> volumes) {
-      this.volumes = (volumes == null) ? null : ImmutableSet.copyOf(volumes);
+    public Builder<T> addVolumes(@Nullable Set<AbsoluteUnixPath> volumes) {
```
GuustavoPaiva: Maybe update the javadoc, since these two methods are not setters anymore.

TadCordle: Good point

Issue #1281, opened 2018-11-27
Containerize an executable WAR


danielpetisme:
**Description of the issue**: Impossible to run an executable WAR

**Expected behavior**: Override the default WAR containerization behavior to permit to run executable WAR

**Steps to reproduce**: 
* Generate a jhipster application: https://www.jhipster.tech/creating-an-app/
* Build the container: https://www.jhipster.tech/docker-compose/#3
* Try to run the container, it should fail
* Run a shell on the container and check `/jetty/webapps/ROOT` has the WAR content 

**Environment**: 

**`jib-maven-plugin` Configuration**: 
```xml
...
<plugin>
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>jib-maven-plugin</artifactId>
   <version>${jib-maven-plugin.version}</version>
   <configuration>
      <from>
         <image>openjdk:8-jre-alpine</image>
      </from>
      <to>
         <image>jhipster:latest</image>
      </to>
      <container>
         <entrypoint>
            <script>/entrypoint.sh</script>
         </entrypoint>
         <ports>
            <port>8080</port>
         </ports>
         <environment>
            <SPRING_OUTPUT_ANSI_ENABLED>ALWAYS</SPRING_OUTPUT_ANSI_ENABLED>
            <JHIPSTER_SLEEP>0</JHIPSTER_SLEEP>
         </environment>
         <useCurrentTimestamp>true</useCurrentTimestamp>
      </container>
      <extraDirectory>
         <permissions>
            <permission>
               <file>/entrypoint.sh</file>
               <mode>700</mode>
            </permission>
         </permissions>
      </extraDirectory>
   </configuration>
</plugin>    
...
```

**`jib-gradle-plugin` Configuration**: 
```groovy
jib {
    from {
        image = 'openjdk:8-jre-alpine'
    }
    to {
        image = 'jhipster:latest'
    }
    container {
        entrypoint = ['/entrypoint.sh']
        ports = ['8080']
        environment = [
            SPRING_OUTPUT_ANSI_ENABLED: 'ALWAYS',
            JHIPSTER_SLEEP: '0'
        ]
        useCurrentTimestamp = true
    }
    extraDirectory {
        permissions = ['/entrypoint.sh': '700']
    }
}
```

**Log output**: 

**Additional Information**: 

The presented issue (executable WAR) seems orthogonal to the `0.10.0` WAR behavior but is a neat feature provided by Spring Boot.

[...]

PR #1287, opened 2018-11-30
Add FAQ entry for containerizing executable JAR

TadCordle:
In response to https://github.com/GoogleContainerTools/jib/issues/1281#issuecomment-442964122

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -10,6 +10,7 @@ If a question you have is not answered below, please [submit an issue](/../../is
 [How do I set parameters for my image at runtime?](#how-do-i-set-parameters-for-my-image-at-runtime)\
 [What image format does Jib use?](#what-image-format-does-jib-use)\
 [Can I define a custom entrypoint?](#can-i-define-a-custom-entrypoint)\
+[I want to containerize an executable JAR/WAR.](#i-want-to-containerize-an-executable-jar-war)\
 [Where is the application in the container filesystem?](#where-is-the-application-in-the-container-filesystem)\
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -153,6 +154,10 @@ See [Extended Usage](../jib-gradle-plugin#extended-usage) for the `container.for
...
 
+### I want to containerize an executable JAR.
+
+Although Jib supports the ability to add an executable JAR to the container and run it using a custom `java -jar` entrypoint command, the intention of Jib is to add individual class files and dependency JARs into the container instead of putting a runnable JAR into the container. This lets Jib choose an opinionated, optimal layout for the application on the container image, which also allows it to skip the extra JAR-packaging step.
```
chanseokoh: How about "Although it is possible to configure Jib to add an ..."? I think it will not be so trivial to make that work, because you need to to first package a JAR, copy the JAR into an isolated location, which may need some time to research (you don't want to have the entire `target` directory as `<extraDirectory>`), and configure `<extraDirectory>` and `<entrypoint>`.

chanseokoh: And basically the same argument applies to a non-executable JAR (in which case you'd have to do `java -cp ...`). Can we somehow mention this too?

Oh, and this sort of applies to WAR too (we put exploded WAR contents), but I think I don't want to mention WAR here for now, because although many popular Servlet engines support placing an exploded WAR for deployment, I wonder if there can be some server that doesn't support it. We will watch that.

chanseokoh:
Looks good. I'll give time for others for comments, if any.

Issue #347, opened 2018-06-01
Could be nice to identify top 3-5 frameworks and provide tutorials for how to use Jib with them.


coollog:
What should be cool is that we simplify their instructions a lot because they don’t need to setup uberjar.

Possible frameworks:

- [x] Spring boot
- [x] Dropwizard (#1290)
- [x] [ktor](https://github.com/ktorio/ktor) (#1291)
- [x] Vert.x
- [ ] Spark
- [ ] [Play framework](https://www.playframework.com/)
- [x] [micronaut](http://micronaut.io/) - move https://github.com/coollog/micronaut-jib over here? #893

coollog:
Any help would be greatly appreciated!

dansiviter:
Thorntail? Although v2 branch uses uberjar packaging which I believe Jib doesn't support.

coollog:
@dansiviter Are there any special files in the uberjar packaging that is not available prior to packaging? If not, Jib should be able to package the container just like the uberjar packaging packages the jar.

dansiviter:
Yea, Thorntail uses JBoss modules which are resolved when the `thorntail-maven-plugin` runs. The maven dependency graph primarily only has the JEE spec libraries (plus a few other bits) so doesn't have the implementations which mean Jib can't bundle the libraries correctly and fails at runtime. The Thorntail maintainers are flattening the classpath via their v4 branch but that's way off yet.

cricket007:
Any news, or has anyone come across a Dropwizard example recently?

Would like to know if there are any gotchas before we begin.

coollog:
@cricket007 We have a few examples up in the [`examples` folder](https://github.com/GoogleContainerTools/jib/tree/master/examples) (Spring Boot, Micronaut, Vert.x). We don't have a Dropwizard example yet, but we could definitely add one! (We welcome anyone to contribute an example too)

chanseokoh:
We've covered 5 frameworks in the examples. Seems good enough to close this now.

ffluk3:
Hey no need to re-open this issue necessarily, but has anyone encountered an example of leveraging the Play Framework with Jib? I am running into pathing issues trying to get things set up.

OneCricketeer:
I recall seeing someone had posted about Play, or maybe in the playfrmework repo about Jib.

I vaguely recall someone mentioning that they recommended using sbt docker plugin instead

PR #1290, opened 2018-12-02
[examples] Dropwizard 

cricket007:
Contributing to #347 

Looking forward to reviews!


cricket007:
I signed it!

examples/dropwizard/pom.xml
```diff
@@ -0,0 +1,235 @@
...
+                        <argument>server</argument>
+                        <argument>${dropwizard.server.config}</argument>
+                    </arguments>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>com.diffplug.spotless</groupId>
+                <artifactId>spotless-maven-plugin</artifactId>
+                <version>${spotless-maven-plugin.version}</version>
```
cricket007: Probably isn't needed anymore. Just used it once to do code-formatting and add the license-header

examples/dropwizard/.gitignore
```diff
@@ -0,0 +1,8 @@
+src/main/jib/config.yml
+
+target/
+dependency-reduced-pom.xml
```
coollog: What is this file? `dependency-reduced-pom.xml`

examples/dropwizard/README.md
```diff
@@ -0,0 +1,50 @@
+# Containerize a [Dropwizard](https://dropwizard.io) application with Jib
+
+Starter Maven template generated with [`dropwizard-archetypes`](https://github.com/dropwizard/dropwizard/tree/master/dropwizard-archetypes)
```
coollog: Is this first section explaining how the example was generated rather than what the user should do to follow the guide? If so, I would recommend placing a header for that like "## How this example was created" or something similar.

examples/dropwizard/README.md
```diff
@@ -0,0 +1,50 @@
...
+  -DarchetypeGroupId=io.dropwizard.archetypes \
+  -DarchetypeArtifactId=example \
+  -DarchetypeVersion=[REPLACE ME WITH A VALID DROPWIZARD VERSION]
+```
+
+Ref. [Dropwizard - Getting Started, Setting up With Maven](https://www.dropwizard.io/1.3.5/docs/getting-started.html#setting-up-using-maven)
+
+Remainder of archetype code was filled in following the above guide.
+
+How to start the Dropwizard application
```
coollog: We prefer the use of markdown headers - `## How to start...`

examples/dropwizard/pom.xml
```diff
@@ -0,0 +1,235 @@
...
+                        <!-- NOTE: Incubating feature. Defaulting to src/main/jib -->
+                        <path>${jib.extraContext}</path>
+                    </extraDirectory>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>dockerBuild</id>
+                        <goals>
+                            <goal>dockerBuild</goal>
+                        </goals>
```
coollog: Looks like this is attaching `dockerBuild` to `package` and `build` to `deploy` - is it the intention to build to both the Docker daemon and the docker registry upon each `mvn deploy`?

examples/dropwizard/pom.xml
```diff
@@ -0,0 +1,235 @@
...
+                        <goals>
+                            <goal>build</goal>
+                        </goals>
+                        <phase>deploy</phase>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <artifactId>maven-shade-plugin</artifactId>
```
coollog: Since Jib doesn't use a JAR or WAR, I believe we can remove the `maven-shade-plugin` and `maven-jar-plugin`?

examples/dropwizard/pom.xml
coollog: Yea, I would recommend removing anything that isn't necessary as it makes the example clearer for the reader.

examples/dropwizard/.gitignore
briandealwis: It's created by the maven-shade-plugin.

examples/dropwizard/config.yml
```diff
@@ -0,0 +1,16 @@
+hello:
```
chanseokoh: Is it possible to move this file to `src/main/jib`? Even if that requires some custom path configuration on the Dropwizard size, I think it will reduce the complexity originated mainly from it: no need for copy-rename-maven-plugin, maven-clean-plugin, the weird copying into the src directory in the generate-source phase while preventing committing the generated file into the repo, and no need to configure `extraDirectory`. And I think it will allow removing the `package` and `deploy` binding altogether. I just hope Dropwizard is flexible enough to place `config.yml` in a non-project root. The example feels a bit heavy on the config side overall as a getting started experience (perhaps the maven archetype is not intended to be a minimal example?), so it'd be nice if we can simplify it.

briandealwis: It does seem odd to copy this `config.yaml` into the root of the container.

briandealwis:
There may be some tension here between applying Jib with minimal changes to a DropWizard example, vs using Jib in a more optimal way to containerize a DropWizard application.

examples/dropwizard/config.yml
chanseokoh: If so, it sounds like we can move `config.yml` to `src/main/jib/app/config.yml` (or `dropwizard-config.yml`) and have `<args><arg>server</arg><arg>/app/config.yml</arg></args>`. Then we can remove copy-rename-maven-plugin, maven-clean-plugin, `config.yml` from `.gitignore`, and the `<extraContents>` property. Probably at this point, I think we can also remove the phase binding, since `mvn compile jib:build` or `mvn compile jib:dockerBuild` will work out of the box.

cricket007: That's a good point about just having the config be in `src/main/jib/app` rather than mucking with copying/cleaning it. 

Can do 👍

examples/dropwizard/pom.xml
```diff
@@ -0,0 +1,235 @@
...
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>jib-maven-plugin</artifactId>
+                <version>${jib-maven-plugin.version}</version>
+                <configuration>
+                    <to>
+                        <image>${jib.to.image}</image>
+                        <tags>
+                            <tag>${project.version}</tag>
+                            <tag>latest</tag>
+                        </tags>
```
chanseokoh: If `latest` is not required, I think we can just git rid of the entire `<tags>` part to further simplify it.

examples/dropwizard/README.md
```diff
@@ -0,0 +1,40 @@
...
+    - **With Docker**: `docker run --rm -p 8080:8080 dropwizard-jib-example:1`
+1. Check that your application is running at http://localhost:8080/hello-world
+
+## Health Check
+
+See your application's health at http://localhost:8080/admin/healthcheck
+
+## Extras
+
+Freemaker templating is setup for [`config.yml`](src/main/jib/app/config/dropwizard.yml) through [`tkrille/dropwizard-template-config`](https://github.com/tkrille/dropwizard-template-config); this allows one to heavily customize the properties file via the container environment with FTL conditional checks and for loops, for example.
```
chanseokoh: Freemaker -->
Freemarker

Maybe FreeMarker.

`config.yml` --> `dropwizard.yml`

examples/dropwizard/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <version>1.6.0</version>
+                <executions>
```
chanseokoh: I think the `<executions>` part is unnecessary.

examples/dropwizard/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+                            <goal>java</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <executable>java</executable>
+                    <mainClass>${mainClass}</mainClass>
+                    <arguments>
+                        <argument>server</argument>
+                        <argument>${dropwizard.server.config}</argument>
```
chanseokoh: Since the `jib` profile is activated by default, this is overridden to `/app/config/dropwizard.yml`, which makes `./mvnw exec:java` fail.

examples/dropwizard/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+        <dropwizard-template-config.version>1.5.0</dropwizard-template-config.version>
+
+        <jib.to.image>gcr.io/REPLACE-WITH-YOUR-GCP-PROJECT/image-built-with-jib</jib.to.image>
+        <jib.container.appRoot>/app</jib.container.appRoot>
+        <jib-maven-plugin.version>0.10.1</jib-maven-plugin.version>
+    </properties>
+
+    <profiles>
+        <profile>
+            <!-- Isolating Jib as a Maven Profile so that the config file
```
chanseokoh: Not really isolated in some sense this this is active by default. If it's active by default, I wonder what's really good to have a profile. Because of overriding `dropwizard.server.config`, `./mvnw exec:java` fails.

examples/dropwizard/README.md
cricket007: Whoops. 

Hmm. IntelliJ caught the path rename, just not the link text.

examples/dropwizard/pom.xml
cricket007: Remove the `exec:java` entirely, then?

examples/dropwizard/pom.xml
cricket007: I meant more like "separating". For example, if `mvn -Pjar` were to be added. 

Maybe that should be an excercize left up to the reader, though

chanseokoh: Ah, yeah, makes sense. Whatever you do, just make sure the commands in `README.md` work.

BTW, is the profile necessary? Configure the right paths to `dropwizard.yml` to `exec-maven-plugin` and `jib-maven-plugin`, and they should work together, right?

examples/dropwizard/pom.xml
cricket007: Oh, you are saying that `<id>java</id>. <goal>java</goal>` is redundant?

I can play around with it, if I understand the context is just to get `mvnw exec:java` working alongside the `jibDockerBuild`

chanseokoh: > Oh, you are saying that `<id>java</id>`. `<goal>java</goal>` is redundant?

Yeah, including the opening and closing `<executions>` part. From my testing, it works without them. You could also check the effective pom before and after removing the section (`mvnw help:effective-pom`) to see if there is any difference.

briandealwis: An [`<execution>`](https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Using_the_executions_Tag) is necessary to either (1) bind the the specific goals to a phase or (2) to be able to specialize the configuration by referencing the execution ID like `mvn exec:java@executionID`.  Maven will use an execution with an ID `default-cli` when invoked from the command-line.

examples/dropwizard/pom.xml
briandealwis: You could put this file in `src/main/resources` and then reference it in the Jib `<args>` as `/app/resources/dropwizard.yml`.

cricket007: That makes more sense, actually, because it's resources _for the application_, not for Jib.

examples/dropwizard/pom.xml
cricket007: @briandealwis So, what are your recommendations here?

examples/dropwizard/pom.xml
cricket007: >they should work together, right?

_Individually_ the commands in the README should work. Not _together_, unless you're saying something else.

chanseokoh: Yeah, the individual commands should work with jib-maven-plugin and exec-maven-plugin configs working together in the default profile.

examples/dropwizard/pom.xml
chanseokoh: My interpretation is that we should remove it, since we don't want to care about the execution ID here. (We won't do `mvnw exec:java@java`). Just verify that `mvnw exec:java` works great.

cricket007:
```bash

docker run --rm -d -p 8081:8080 dropwizard-jib-example:1 && ./mvnw exec:java
3e8b4c9e882f72df0e172979b819116ad7a546e4002c124ba482d52f571ba65e

[INFO] 
[INFO] -------------------------< example:dropwizard >-------------------------
[INFO] Building Dropwizard Jib Example 1
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] --- exec-maven-plugin:1.6.0:java (default-cli) @ dropwizard ---
INFO  [2018-12-07 07:43:44,361] io.dropwizard.server.SimpleServerFactory: Registering jersey handler with root path prefix: /
INFO  [2018-12-07 07:43:44,364] io.dropwizard.server.SimpleServerFactory: Registering admin handler with root path prefix: /admin
INFO  [2018-12-07 07:43:44,368] io.dropwizard.server.SimpleServerFactory: Registering jersey handler with root path prefix: /
INFO  [2018-12-07 07:43:44,369] io.dropwizard.server.SimpleServerFactory: Registering admin handler with root path prefix: /admin
INFO  [2018-12-07 07:43:44,378] io.dropwizard.server.ServerFactory: Starting Dropwizard Jib Example
================================================================================

                              Dropwizard Jib Example

================================================================================


INFO  [2018-12-07 07:43:44,560] org.eclipse.jetty.setuid.SetUIDListener: Opened Dropwizard Jib Example@4d380b90{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
INFO  [2018-12-07 07:43:44,562] org.eclipse.jetty.server.Server: jetty-9.4.11.v20180605; built: 2018-06-05T18:24:03.829Z; git: d5fc0523cfa96bfebfbda19606cad384d772f04c; jvm 1.8.0_192-b12
INFO  [2018-12-07 07:43:45,026] io.dropwizard.jersey.DropwizardResourceConfig: The following paths were found for the configured resources:

    GET     /hello-world (example.resources.HelloWorldResource)

INFO  [2018-12-07 07:43:45,030] org.eclipse.jetty.server.handler.ContextHandler: Started i.d.j.MutableServletContextHandler@75f11cec{/,null,AVAILABLE}
INFO  [2018-12-07 07:43:45,034] io.dropwizard.setup.AdminEnvironment: tasks = 

    POST    /tasks/log-level (io.dropwizard.servlets.tasks.LogConfigurationTask)
    POST    /tasks/gc (io.dropwizard.servlets.tasks.GarbageCollectionTask)

INFO  [2018-12-07 07:43:45,039] org.eclipse.jetty.server.handler.ContextHandler: Started i.d.j.MutableServletContextHandler@4d347ded{/admin,null,AVAILABLE}
INFO  [2018-12-07 07:43:45,049] org.eclipse.jetty.server.AbstractConnector: Started Dropwizard Jib Example@4d380b90{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
INFO  [2018-12-07 07:43:45,049] org.eclipse.jetty.server.Server: Started @4208ms
0:0:0:0:0:0:0:1 - - [07/Dec/2018:07:43:58 +0000] "GET / HTTP/1.1" 404 43 "-" "curl/7.54.0" 81
0:0:0:0:0:0:0:1 - - [07/Dec/2018:07:44:35 +0000] "GET /hello-world HTTP/1.1" 200 37 "-" "curl/7.54.0" 13

^C

INFO  [2018-12-07 07:44:54,896] org.eclipse.jetty.server.AbstractConnector: Stopped Dropwizard Jib Example@4d380b90{HTTP/1.1,[http/1.1]}{0.0.0.0:8080}
INFO  [2018-12-07 07:44:54,897] org.eclipse.jetty.server.handler.ContextHandler: Stopped i.d.j.MutableServletContextHandler@4d347ded{/admin,null,UNAVAILABLE}
INFO  [2018-12-07 07:44:54,902] org.eclipse.jetty.server.handler.ContextHandler: Stopped i.d.j.MutableServletContextHandler@75f11cec{/,null,UNAVAILABLE}

$ docker ps 
CONTAINER ID        IMAGE                      COMMAND                  CREATED              STATUS              PORTS                    NAMES
3e8b4c9e882f        dropwizard-jib-example:1   "java -server -Djava…"   About a minute ago   Up About a minute   0.0.0.0:8081->8080/tcp   modest_hodgkin
```

PR #1291, opened 2018-12-03
[examples] Ktor

cricket007:
Contributing to #347 

Hot off the heals of Gradle 5.0 Kotlin's DSL and [Jetbrain's 1.0 release of Ktor](https://blog.jetbrains.com/kotlin/2018/11/ktor-1-0/), and showcasing Jib building Kotlin applications as  well. 

And much less ceremony than [Ktor's Docker docs](https://ktor.io/quickstart/quickstart/docker.html)


patflynn:
@cricket007 sweet!

examples/ktor/README.md
```diff
@@ -0,0 +1,32 @@
+# Containerize a [Ktor] application with Jib
+
+This is an example of how to easily build a Docker image for a [Ktor] application with Jib.
+
+```shell
+./gradlew jibDockerBuild
+
+docker run -d --rm -p 8080:8080 ktor-jib-example
```
coollog: Since the version is automatically set to `1` - I believe this should be `ktor-jib-example:1`

chanseokoh: > It seems to also tag `latest`.

I don't think this is true here. In the other Dropwizard PR, it had explicit configuration to have the additional `latest` tag, but here, I think it is not creating `latest`.

coollog: +1 to @chanseokoh comment

For group, `example` would be fine. Only an example :)

examples/ktor/README.md
```diff
@@ -0,0 +1,32 @@
+# Containerize a [Ktor] application with Jib
+
+This is an example of how to easily build a Docker image for a [Ktor] application with Jib.
+
+```shell
+./gradlew jibDockerBuild
+
+docker run -d --rm -p 8080:8080 ktor-jib-example:1
```
chanseokoh: Let's remove `-d`. Will show logs in the terminal and the user can do Ctrl+C to quit.

cricket007: Whoops. Must've copied that line from me testing other things

Issue #1297, opened 2018-12-04
Provide feedback on containerization progress.


coollog:
This issue tracks the tasks needed to implement the progress feedback system intended to address issues #1251 and #806.

(# represents the estimated amount of work)

Design:
- [x] (5) Proposal #1262

Core system:
- [x] (2) Add `Progress.Allocation` class to represent the [decentralized allocation tree](https://github.com/GoogleContainerTools/jib/blob/master/proposals/progress_output.md#decentralized-allocation-tree-dat) #1298
- [x] (1) Add `Progress.Event` class to represent a `JibEvent` indicating progress on some `Progress.Allocation` #1305
- [x] (5) Add `ProgressEventHandler` for handling `Progress.Event`s and managing an overall progress amount #1309

Integration:
- [x] (1) Change `StepsRunner` to be more of a builder to prepare for assigning `Allocation`s #1312
- [x] (4) Assign `Progress.Allocation` to each builder step #1321
- [x] (4) Change `RegistryClient` push and pull to take callback to listen on the progress of the push/pull #1323
  - [x] (3) Add `ListenableCountingOutputStream` to count sent/received bytes #1300
- [x] (3) Add `ProgressEventDispatcher` to dispatch `ProgressEvent`s efficiently #1330
- [x] (2) Emit progress of push/pull as `ProgressEvent`s #1334
- [x] (3) Add integration test(s) to make sure progress reaches 100% after each of the builds #1339

Display:
- [x] (3) Add `AnsiLoggerWithFooter` to be able to keep a footer for where progress indication will go #1326
- [x] (3) Add `ProgressDisplayGenerator` to generate the displayed text for progress feedback #1333
- [x] (2) Add `AnsiLoggerWithFooter` to Maven/Gradle #1341
- [x] (2) Change `LogEventHandler` to use `AnsiLoggerWithFooter` #1341
- [x] (3) Add ability to switch on/off the old logging and the footer #1344
- [x] (2) Change `ProgressEventHandler` to provide the callback with updates #1343
- [x] (1) Attach `ProgressEventHandler` as an event handler via the plugins #1350
- [x] (1) Change `jib.showProgress` to `true` by default #1354

Cleanup:
- [x] Refactor progress callbacks in `RegistryClient` classes #1334 
- [x] (2) Improve `AnsiLoggerWithFooter` logic #1340
[...]

PR #1298, opened 2018-12-04
Adds progress Allocation to represent Decentralized Allocation Tree node.

coollog:
Part of #1297. The `Allocation` class will be used by the `ProgressEvent`.

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/AllocationTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/AllocationTest.java
@@ -27,9 +27,9 @@ public class AllocationTest {
 
   @Test
   public void testSmoke_linear() {
-    Allocation root = Allocation.newRoot("root").allocate(1);
-    Allocation node1 = root.newChild("node1").allocate(2);
-    Allocation node2 = node1.newChild("node2").allocate(3);
+    Allocation root = Allocation.newRoot("root", 1);
+    Allocation node1 = root.newChild("node1", 2);
+    Allocation node2 = node1.newChild("node2", 3);
 
     Assert.assertEquals("node2", node2.getDescription());
     Assert.assertEquals(3, node2.getAllocationUnits());
@@ -51,13 +51,13 @@ public class AllocationTest {
 
   @Test
   public void testFractionOfRoot_tree_partial() {
-    Allocation root = Allocation.newRoot("ignored").allocate(10);
-    Allocation left = root.newChild("ignored").allocate(2);
-    Allocation right = root.newChild("ignored").allocate(4);
-    Allocation leftDown = left.newChild("ignored").allocate(20);
-    Allocation rightLeft = right.newChild("ignored").allocate(20);
-    Allocation rightRight = right.newChild("ignored").allocate(100);
-    Allocation rightRightDown = rightRight.newChild("ignored").allocate(200);
+    Allocation root = Allocation.newRoot("ignored", 10);
+    Allocation left = root.newChild("ignored", 2);
+    Allocation right = root.newChild("ignored", 4);
+    Allocation leftDown = left.newChild("ignored", 20);
+    Allocation rightLeft = right.newChild("ignored", 20);
+    Allocation rightRight = right.newChild("ignored", 100);
+    Allocation rightRightDown = rightRight.newChild("ignored", 200);
 
     Assert.assertEquals(1.0, root.getFractionOfRoot(), DOUBLE_ERROR_MARGIN);
     Assert.assertEquals(1.0 / 10, left.getFractionOfRoot(), DOUBLE_ERROR_MARGIN);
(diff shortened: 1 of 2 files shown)


loosebazooka:
Not specifically about this, but autobuilder is a great alternative to writing these out by hand.

coollog:
Hmm, I'm not sure if AutoValue Builder is able to work for the single-method builder we have here?

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/AllocationTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /** Error margin for checking equality of two doubles. */
+  private static final double DOUBLE_ERROR_MARGIN = 1e-10;
+
+  /**
+   * Checks that the {@code allocations} add up to a full {@code 1.0}.
+   *
+   * @param allocations the leaf allocations
+   */
+  private static void checkAllocationComplete(Allocation... allocations) {
```
TadCordle: This function is only called once, so maybe it can just go straight in the test.

chanseokoh:
I see this will have no problem constructing the allocation tree.

However, I think I don't fully understand the intended usage in practice yet. I thought about some scenarios from the perspective of a developer trying to construct the whole tree with multiple depth of method calls and am left with some questions. For example, it isn't yet clear if whether it would be a caller or callee that should call newChild, if it is OK to pass the root or some child node to multiple, sequential method calls and/or pass the node down multiple call depth, if we can enforce matching the allocation units with the minimum number of newChild calls (or if it is necessary), what a callee should do if it just wants to mark completion (i.e., what info to emit as an event) when the allocation the callee got is a root node, etc. Perhaps better to ask these offline.

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/AllocationTest.java
```diff
@@ -0,0 +1,93 @@
...
+    Assert.assertEquals(1.0 / 10, right.getFractionOfRoot(), DOUBLE_ERROR_MARGIN);
+    Assert.assertEquals(1.0 / 10 / 2, leftDown.getFractionOfRoot(), DOUBLE_ERROR_MARGIN);
+    Assert.assertEquals(1.0 / 10 / 4, rightLeft.getFractionOfRoot(), DOUBLE_ERROR_MARGIN);
+    Assert.assertEquals(1.0 / 10 / 4, rightRight.getFractionOfRoot(), DOUBLE_ERROR_MARGIN);
+    Assert.assertEquals(
+        1.0 / 10 / 4 / 100, rightRightDown.getFractionOfRoot(), DOUBLE_ERROR_MARGIN);
+  }
+
+  @Test
+  public void testFractionOfRoot_tree_complete() {
```
chanseokoh: So I see this example shows that the expect normal usage is to match the number of children and the parent allocation units. As I said in my [previous comment](https://github.com/GoogleContainerTools/jib/pull/1298#issuecomment-444320149), I am really curious how this can pan out or be enforced efficiently in practice. But I think we can move forward with this now, and I'd be able to see if the API will work out to minimize errors.

coollog: Yea, the user of the `Allocation` class will be responsible for sub-allocating the correct number of allocations. For example, for a task that knows it will launch 3 subtasks, it will first create its own allocation as 3 allocation units, and then launch the 3 subtasks, passing down its own allocation as the "parent" allocation, knowing that each of the subtasks will thus make a suballocation, resulting in 3 total suballocations.

chanseokoh: Yeah, in that example, what if the writer of one of the subtasks (think of as separate methods) that get the "parent" allocation doesn't have anything to report progress or if the writer simply wants to mark completion before returning using the given parent allocation node? Should it emit an event with the parent node and 1 allocation done? But, in turn, what if the subtask called other methods and propagated down the parent node as-is? In that case, if the sub-subtasks did the same thing as what the subtask did (emitting the event with the parent node and 1 allocation done), the progress will go over the allocated amount.

coollog: Yes, emitted an event with the parent node and 1 allocation done can work, or the subtask can create a suballocation and emit an event with all progress units in the suballocation. The allocation(s) are only to represent the subdivision of the overall progress and it will be up to the tasks to coordinate how to emit progress on those allocations (with the premise that parent tasks have power over child tasks) and up to the progress event listener to process the emitted progress in whichever way it chooses to do so (allow or disallow going over allocated amount, for eg.).

chanseokoh: > with the premise that parent tasks have power over child tasks)

Yeah, I think this premise is required for this work well. When the premise is applied recursively, I think it implies having power over all descendants. For example, if you are some subtask at some point, and you call several long-running external methods in sequence, you don't really know about whose internals. So, you just pass the allocation node you have, but you basically have no idea how those will make use up the node. They may or may not consume 1 allocation, or more. And those external methods may in the same situation recursively. So, basically, it looks to me that we should have the premise that parent tasks should have an idea about all descendants.

coollog: Yeah, in our use case in the builder steps, this premise will be met by a constraint that 

1. A task may only create at most one suballocation of the parent allocation passed to it, and
1. A task cannot pass the parent allocation on to subtasks

PR #1300, opened 2018-12-04
Adds ListenableCountingOutputStream.

coollog:
Part of #1297 

This will be used to count sent/received bytes when pushing and pulling to/from registries.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
@@ -17,85 +17,18 @@
 package com.google.cloud.tools.jib.http;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-import javax.annotation.Nullable;
 
 /**
  * Counts the number of bytes written and reports the count to a callback. The count is reported
  * with certain time delays to avoid calling the callback too often.
  */
 class ListenableCountingOutputStream extends OutputStream {
-
-  /** Partial {@link Builder} interface for setting the {@link #delayBetweenCallbacks}. */
-  interface DelayBuilder {
-
-    /**
-     * Sets the delay between each call to {@link #byteCountConsumer}.
-     *
-     * @param delayBetweenCallbacks the delay between calls
-     * @return a {@link CallbackBuilder}
-     */
-    CallbackBuilder every(Duration delayBetweenCallbacks);
-  }
-
-  /** Partial {@link Builder} interface for setting the {@link #byteCountConsumer}. */
-  interface CallbackBuilder {
-
-    /**
-     * Sets the {@link #byteCountConsumer} that is with a count of bytes written.
-     *
-     * @param byteCountConsumer the byte count {@link Consumer}
-     * @return a {@link ListenableCountingOutputStream}
-     */
-    ListenableCountingOutputStream forEachByteCount(Consumer<Long> byteCountConsumer);
-  }
-
-  /** Builds a {@link ListenableCountingOutputStream}. */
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
```diff
@@ -0,0 +1,127 @@
...
+    callByteCountConsumer(instantSupplier.get());
+  }
+
+  @Override
+  public void close() throws IOException {
+    underlyingOutputStream.close();
+    callByteCountConsumer(instantSupplier.get());
+  }
+
+  private void wroteBytes(int byteCount) {
```
TadCordle: Naming this in the past tense makes it sound like it should return a boolean.

jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
```diff
@@ -0,0 +1,127 @@
...
+
+  @Override
+  public void write(int singleByte) throws IOException {
+    underlyingOutputStream.write(singleByte);
+    wroteBytes(1);
+  }
+
+  @Override
+  public void write(byte[] byteArray) throws IOException {
+    underlyingOutputStream.write(byteArray);
```
TadCordle: Could just call the overload `write(byteArray, 0, byteArray.length)` I think?

jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
```diff
@@ -0,0 +1,127 @@
...
+    this.underlyingOutputStream = underlyingOutputStream;
+    this.byteCountConsumer = byteCountConsumer;
+    this.delayBetweenCallbacks = delayBetweenCallbacks;
+    this.instantSupplier = instantSupplier;
+
+    previousCallback = instantSupplier.get();
+  }
+
+  @Override
+  public void write(int singleByte) throws IOException {
```
TadCordle: This should take a `byte` instead of an `int`.

jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
TadCordle: And maybe even for `write(byte singleByte)` you can call `write(new byte[] { singleByte }, 0, 1)`, but maybe that hurts readability too much.

coollog: This is just deferring the logic to the underlying outputstream to handle. Calling `write(byteArray, 0, byteArray.length)` would assume that the underlying `write(byte[], offset, length)` is the same implementation as `write(byte[])`, which it may not be (ie. `write(byte[], offset, length)` could be less efficient).

`write(new byte[] { singleByte }, 0, 1)` is also most likely going to be less efficient than `write(singleByte)` especially with the additional array creation.

jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
coollog: This is just overriding the `OutputStream#write(int)` method.

TadCordle: In that case it should be called something other than `singleByte` and should call `wroteBytes(4)`?

TadCordle: Nevermind, apparently the upper 24 bits are ignored: https://docs.oracle.com/javase/7/docs/api/java/io/OutputStream.html#write(int)

coollog: The `OutputStream#write(int)` method is not the best-designed method signature, but the int does only represent 1 byte (only the 8 low-order bits are taken into account).

PR #1301, opened 2018-12-05
[chore] Simplify ZipUtil.unzip

samlanning:
There is no associated issue for this, as it is a minor thing I noticed when going through the LGTM issues for this project. (I'm currently looking for instances of ZipSlip, which this code currently protects against). I'm unsure if this is the correct thing to do here as it's not issue worthy. I'll understand if it gets rejected. :)

You can simplify checking that destination files remain within the destination folder using `Path.startsWith()`, rather than doing a string prefix comparison. `Path.toRealPath()` resolves symlinks like `File.getCanonicalPath()`, so behaviour should remain identical.

There are a [couple of other alerts](https://lgtm.com/projects/g/GoogleContainerTools/jib/alerts) for this project on LGTM.com that would probably be good to fix too.

*(Full disclosure: I'm part of the team behind LGTM.com)*

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java
@@ -38,7 +38,6 @@ public class ZipUtil {
    * @throws IOException when I/O error occurs
    */
   public static void unzip(Path archive, Path destination) throws IOException {
-    String canonicalDestination = destination.toFile().getCanonicalPath();
 
     try (InputStream fileIn = Files.newInputStream(archive);
         ZipInputStream zipIn = new ZipInputStream(new BufferedInputStream(fileIn))) {
@@ -46,8 +45,7 @@ public class ZipUtil {
       for (ZipEntry entry = zipIn.getNextEntry(); entry != null; entry = zipIn.getNextEntry()) {
         Path entryPath = destination.resolve(entry.getName());
 
-        String canonicalTarget = entryPath.toFile().getCanonicalPath();
-        if (!canonicalTarget.startsWith(canonicalDestination + File.separator)) {
+        if (!entryPath.toRealPath().startsWith(destination.toRealPath())) {
           String offender = entry.getName() + " from " + archive;
           throw new IOException("Blocked unzipping files outside destination: " + offender);
         }


chanseokoh:
Oh, you said the method resolves symlinks, so I think the symlink is not a good example. But I do remember there were other things that you really need to compute the canonical path.

samlanning:
Oh, interesting! Well I'd definitely be interested in knowing the difference between toRealPath and getCanonicalPath.

As far as this particular issue goes, if we're only trying to detect unzipping to outside the target directory, completely avoiding filesystem accesses (resolving synlinks) and just using `normalize()` instead of `toRealPath()` should be sufficient right?

I mean the only case where this would be likely to have bad behaviour is if the target directory itself contained synlinks to folders matching the name of directory entries in the zip file?

I'd be really interested in learning how that could be abused.

chanseokoh:
Now I can recall.

First of all, `normalize()` does not access the filesystem, so this is out of the option. You already said the example where the target directory contains symlinks, and I can imagine that this can be very easy to exploit. For example, on my system, `/var/run` is a symlink pointing to the `/run` directory. Since anyone can expect a lot of systems will have `/var/run` as a symlink, you can easily think of having ZIP entries like `/var/run/evil.txt`. Then, by unpacking a zip with this entry under `/var/run`, you can actually create a file under `/run`. I am not a security person, but it doesn't make sense to let it happen even though we know we can definitely prevent it.

For `toRealPath()`, there are two issues, one of which is particularly problematic:
1. It is for returning a path for an *existing* file; it throws an exception if the file does not exist. That's why the tests are currently failing in this PR. This makes `toRealPath()` out of the option.
1. `getCanonicalPath()` guarantees a unique (one and only one) path for any existing and non-existing pathname, which is what we exact here. According to its Javadoc, computing a canonical path "typical involves removing ..., resolving symbolic links (on UNIX platforms), and converting drive letters to a standard case (on Microsoft Windows platforms)." Even if `toRealPath()` worked for a non-existing pathname, it is not clear if it would do the exactly same computation as `getCanonicalPath()` does. So, I think it is the safest bet we stick to `getCanonicalPath()`. It is also the actual recommended Java code as a fix from the team who announced Zip-Slip.

samlanning:
Allrighty then, that sounds good! I'll close. Thanks for the clarification!

PR #1305, opened 2018-12-05
Adds ProgressEvent.

coollog:
Part of #1297 

Represents a `JibEvent` indicating progress on some `Allocation`.

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/event/events/ProgressEventTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/event/events/ProgressEventTest.java
@@ -23,55 +23,108 @@ import com.google.cloud.tools.jib.event.JibEventType;
 import com.google.cloud.tools.jib.event.progress.Allocation;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Consumer;
 import org.junit.Assert;
 import org.junit.Test;
 
 /** Tests for {@link ProgressEvent}. */
 public class ProgressEventTest {
 
+  /** {@link Allocation} tree for testing. */
+  private static class AllocationTree {
+
+    /** The root node. */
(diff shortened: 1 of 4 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/event/events/ProgressEventTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+    allocationCompletionMap.put(allocation, units);
+
+    if (allocation.getAllocationUnits() == units) {
+      allocation
+          .getParent()
+          .ifPresent(parentAllocation -> updateCompletionMap(parentAllocation, 1));
+    }
+  }
+}
```
chanseokoh: How about adding a simple test that doesn't make use of `updateCompletionMap` but just accumulates the overall progress in a single double variable utilizing `getFractionOfRoot()`? You could add a new event listener.

jib-core/src/test/java/com/google/cloud/tools/jib/event/events/ProgressEventTest.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  private void handleProgressEvent(ProgressEvent progressEvent) {
+    Allocation allocation = progressEvent.getAllocation();
+    long units = progressEvent.getUnits();
+
+    updateCompletionMap(allocation, units);
+  }
+
+  private void updateCompletionMap(Allocation allocation, long units) {
```
chanseokoh: Just saying, I guess we need to think a bit more about this in a concurrent environment, but this works in tests.

coollog: Yes, in a concurrent environment this would need to synchronize at the get-and-increment level.

PR #1306, opened 2018-12-05
Add system properties for setting cache directories

### LINKED ISSUE #1238: [gradle] Allow programatically setting base layer cache directory
**Description of the issue**: It would be convenient to be able to set the base layer cache directory programatically, for example to set it to `project.getGradle().getGradleUserHomeDir()` to keep it together with other gradle caches if desired.

**Expected behavior**: Currently, it seems to only be possible to either say yes/no to project-level cache, or use the environment variable XDG_CACHE_HOME. It is often difficult to set an environment variable programatically or in a way that allows reproducible builds (e.g. applies to both CI and dev machines).

**Environment**: gradle-jib-plugin

TadCordle:
Fixes #1238.

Adds two system properties for setting cache directories:
* `jib.baseImageCache`
* `jib.applicationCache`

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -450,11 +450,9 @@ public class PluginConfigurationProcessor {
    */
   private static Path getCheckedCacheDirectory(String property, Path defaultPath) {
     if (System.getProperty(property) != null) {
-      Path path = Paths.get(System.getProperty(property));
-      return path.isAbsolute() ? path : Paths.get(System.getProperty("user.dir")).resolve(path);
-    } else {
-      return defaultPath;
+      return Paths.get(System.getProperty(property));
     }
+    return defaultPath;
   }
 
   private final JibContainerBuilder jibContainerBuilder;
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -40,6 +40,7 @@ import com.google.common.base.Verify;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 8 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -421,13 +422,38 @@ private static void configureContainerizer(
...
+   *
+   * @param property the name of the system property to check
+   * @param defaultPath the path to return if the system property isn't set
+   * @return the value of a cache directory system property if it is set, otherwise returns {@code
+   *     defaultPath}
+   */
+  private static Path getCheckedCacheDirectory(String property, Path defaultPath) {
+    if (System.getProperty(property) != null) {
+      Path path = Paths.get(System.getProperty(property));
+      return path.isAbsolute() ? path : Paths.get(System.getProperty("user.dir")).resolve(path);
```
coollog: Hmm, why is this resolving relative to the user dir (how is it different from just using the path directly?)?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -421,13 +422,38 @@ private static void configureContainerizer(
...
+    } else {
```
coollog: nit: else unnecessary

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
TadCordle: Because I don't know how paths work, I guess.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PropertyNames.java
```diff
@@ -42,6 +42,8 @@
   public static final String CONTAINER_PORTS = "jib.container.ports";
   public static final String CONTAINER_USE_CURRENT_TIMESTAMP = "jib.container.useCurrentTimestamp";
   public static final String USE_ONLY_PROJECT_CACHE = "jib.useOnlyProjectCache";
+  public static final String BASE_IMAGE_LAYERS_CACHE = "jib.baseImageCache";
+  public static final String APPLICATION_IMAGE_LAYERS_CACHE = "jib.applicationImageCache";
```
loosebazooka: I'm being nitty, but the application is not an "image" right? Just curious if `applicationCache` makes sense?

TadCordle: Isn't the concept of layering supposed to be hidden from plugin users though?

coollog: Yea, either way would be fine.

TadCordle: Since this is in plugins common, I think I'll keep it this way then (`jib.baseImageCache` and `jib.applicationCache`). Plus it's less verbose which is nice.

coollog:
We might want to consider deprecating `useOnlyProjectCache` to remove that from the plugin configuration and have that settable via the new system property. It didn't make sense to define the cache location via the plugin configuration anyways.

TadCordle:
Filed an issue #1308

PR #1307, opened 2018-12-05
Consolidate digest generation in LoadDockerStep and WriteTarFileStep

TadCordle:
Fixes a todo that has been around for a while. It may be possible to consolidate a bit more in `PushImageStep`, but that has a bit of extra code for pushing to multiple tags, so I ignored it for now.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildResult.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildResult.java
@@ -16,11 +16,46 @@
 
 package com.google.cloud.tools.jib.builder.steps;
 
+import com.google.cloud.tools.jib.blob.BlobDescriptor;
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildResult.java
```diff
@@ -16,12 +16,47 @@
...
 
+  /**
+   * Gets a {@link BuildResult} from an {@link Image}.
+   *
+   * @param image the image
+   * @param targetFormat the target format of the image
+   * @return a new {@link BuildResult} with the image's digest and id
+   * @throws IOException if writing the digest or container configuration fails
+   */
+  static BuildResult fromImage(
```
coollog: Let's add a simple unit test now that this is separated out nicely into its own method.

PR #1309, opened 2018-12-05
 Adds ProgressEventHandler.

coollog:
Part of #1297 

`ProgressEventHandler` keeps an overall progress state as well as tracks unfinished allocations via the `AllocationCompletionTracker`. `ProgressEventHandler` will be the only concurrently-modifiable class in Jib (as of yet).

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.event.progress;
+
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.PriorityQueue;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Keeps track of the progress for {@link Allocation}s as well as their order in which they appear.
+ *
+ * <p>This implementation is thread-safe.
+ */
+class AllocationCompletionTracker {
+
+  /**
+   * Holds the progress units completed along with a creation order. This is used as the value of
+   * the {@link #completionMap}.
+   */
+  private static class InsertionOrderUnits implements Comparable<InsertionOrderUnits> {
+
+    /** Monotonically-increasing source for {@link #index}. */
+    private static final AtomicInteger currentIndex = new AtomicInteger();
+
+    /** The creation order that monotonically increases. */
+    private final int index = currentIndex.getAndIncrement();
+
+    /**
+     * Progress units completed. This can be shared across multiple threads and should be updated
+     * atomically.
+     */
+    private final AtomicLong units = new AtomicLong();
+
+    /** The {@link Allocation} this is for. */
+    private final Allocation allocation;
+
+    private InsertionOrderUnits(Allocation allocation) {
+      this.allocation = allocation;
+    }
+
+    @Override
+    public int compareTo(InsertionOrderUnits otherInsertionOrderUnits) {
+      return index - otherInsertionOrderUnits.index;
+    }
+  }
+
+  /**
+   * Maps from {@link Allocation} to the number of progress units completed in that {@link
+   * Allocation}, while keeping track of the insertion order of the {@link Allocation}s with {@link
+   * InsertionOrderUnits}.
+   */
+  private final ConcurrentHashMap<Allocation, InsertionOrderUnits> completionMap =
+      new ConcurrentHashMap<>();
+
+  /**
+   * Puts an entry for {@code allocation} in the map if not present, with progress initialized to
+   * {@code 0}.
+   *
+   * @param allocation the {@link Allocation}
+   * @return {@code true} if the map was updated; {@code false} if {@code allocation} was already
+   *     present
+   */
+  boolean putIfAbsent(Allocation allocation) {
+    // Note: Could have false positives.
+    boolean alreadyPresent = completionMap.containsKey(allocation);
+    completionMap.computeIfAbsent(allocation, InsertionOrderUnits::new);
+    return alreadyPresent;
+  }
+
+  /**
+   * Updates the progress for {@link Allocation} atomically relative to the {@code allocation}.
+   *
+   * @param allocation the {@link Allocation} to update progress for
+   * @param units the units of progress
+   */
+  void updateProgress(Allocation allocation, long units) {
+    completionMap.compute(
+        allocation,
+        (ignored, insertionOrderUnits) -> {
+          if (insertionOrderUnits == null) {
+            insertionOrderUnits = new InsertionOrderUnits(allocation);
+          }
+
+          updateProgress(insertionOrderUnits, units);
+
+          return insertionOrderUnits;
+        });
+  }
+
+  /**
+   * Gets a list of the unfinished {@link Allocation}s in the order in which those {@link
+   * Allocation}s were encountered.
+   *
+   * @return a list of unfinished {@link Allocation}s
+   */
+  List<Allocation> getUnfinishedAllocations() {
+    Queue<InsertionOrderUnits> unfinishedInsertionOrderUnits = new PriorityQueue<>();
+
+    for (InsertionOrderUnits insertionOrderUnits : completionMap.values()) {
+      if (insertionOrderUnits.units.get() == insertionOrderUnits.allocation.getAllocationUnits()) {
+        unfinishedInsertionOrderUnits.add(insertionOrderUnits);
+      }
+    }
+
+    List<Allocation> unfinishedAllocations = new ArrayList<>();
+    for (InsertionOrderUnits insertionOrderUnits : unfinishedInsertionOrderUnits) {
+      unfinishedAllocations.add(insertionOrderUnits.allocation);
+    }
+    return unfinishedAllocations;
+  }
+
+  /**
+   * Helper method for {@link #updateProgress(Allocation, long)}. This method is <em>not</em>
+   * thread-safe for the {@code insertionOrderUnits} and should be called atomically relative to the
+   * {@code insertionOrderUnits}.
+   *
+   * @param insertionOrderUnits the {@link InsertionOrderUnits} to update progress for
+   * @param units the units of progress
+   */
+  private void updateProgress(InsertionOrderUnits insertionOrderUnits, long units) {
+    Allocation allocation = insertionOrderUnits.allocation;
+
+    long newUnits = insertionOrderUnits.units.addAndGet(units);
+    if (newUnits > allocation.getAllocationUnits()) {
+      throw new IllegalStateException(
+          "Progress exceeds max for '"
+              + allocation.getDescription()
+              + "': "
+              + newUnits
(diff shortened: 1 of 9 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Updates the progress for {@link Allocation} atomically relative to the {@code allocation}.
+   *
+   * <p>For any {@link Allocation}, this method <em>must</em> have been called on all of its parents
+   * beforehand.
+   *
+   * @param allocation the {@link Allocation} to update progress for
+   * @param units the units of progress
+   * @return {@code true} if the map was updated; {@code false} if {@code allocation} was already
```
chanseokoh: I think it's easier to understand to say something along the line "it may return true even if the map was not updated if called concurrently, but never false if the map was updated."

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
```diff
@@ -0,0 +1,162 @@
...
+    return true;
+  }
+
+  /**
+   * Gets a list of the unfinished {@link Allocation}s in the order in which those {@link
+   * Allocation}s were encountered.
+   *
+   * @return a list of unfinished {@link Allocation}s
+   */
+  List<Allocation> getUnfinishedAllocations() {
```
chanseokoh: It's not revealed in this PR why we need a list ordered by the time at which events arrived to the event consumer. Can you explain it briefly here? I think it'd be nice and things will become a lot simpler if whoever displays the multiple progress bars in the end would not have to depend on the order of the list, sorted by the time the events arrived.

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/ProgressEventHandlerTest.java
```diff
@@ -0,0 +1,124 @@
...
+  @Test
+  public void testAccept() throws ExecutionException, InterruptedException, IOException {
+    try (MultithreadedExecutor multithreadedExecutor = new MultithreadedExecutor()) {
+      ProgressEventHandler progressEventHandler = new ProgressEventHandler(() -> {});
+      EventDispatcher eventDispatcher =
+          new DefaultEventDispatcher(
+              new EventHandlers().add(JibEventType.PROGRESS, progressEventHandler));
+
+      // Adds root, child1, and child1Child.
+      multithreadedExecutor.invokeAll(
```
chanseokoh: I see most of the usage of `invokeAll()` is `invokveAll(signletonList)`. How about just adding `invoke(callable)`?

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/ProgressEventHandlerTest.java
```diff
@@ -0,0 +1,124 @@
...
+          Collections.singletonList(
+              () -> {
+                eventDispatcher.dispatch(new ProgressEvent(AllocationTree.root, 0L));
+
+                multithreadedExecutor.invokeAll(
+                    Collections.singletonList(
+                        () -> {
+                          eventDispatcher.dispatch(new ProgressEvent(AllocationTree.child1, 0L));
+
+                          multithreadedExecutor.invokeAll(
```
chanseokoh: `invokeAll()` is a blocking call, so this whole chain just emits events in this deterministic sequence:

```java
() -> new ProgressEvent(AllocationTree.child1Child, 0L)
eventDispatcher.dispatch(new ProgressEvent(AllocationTree.child1, 0L))
eventDispatcher.dispatch(new ProgressEvent(AllocationTree.root, 0L))
```

Why not

```
invoke(() -> new ProgressEvent(AllocationTree.child1Child, 0L));
invoke(() -> eventDispatcher.dispatch(new ProgressEvent(AllocationTree.child1, 0L)));
invoke(() -> eventDispatcher.dispatch(new ProgressEvent(AllocationTree.root, 0L)));
```
? This looks more concise and easier to get.

Or if you want some random concurrency

```
invokeAll(
  () -> new ProgressEvent(AllocationTree.child1Child, 0L),
  () -> eventDispatcher.dispatch(new ProgressEvent(AllocationTree.child1, 0L)),
  () -> eventDispatcher.dispatch(new ProgressEvent(AllocationTree.root, 0L)))?
```
? But maybe not, because you're doing the concurrent calls in the other test.

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/ProgressEventHandlerTest.java
```diff
@@ -0,0 +1,124 @@
...
+                              Collections.singletonList(
+                                  () -> new ProgressEvent(AllocationTree.child1Child, 0L)));
```
chanseokoh: Did you mean `eventDispatcher.dispatch(new ProgressEvent(AllocationTree.child1Child, 0L)))`? This doesn't seem to do anything really.

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/ProgressEventHandlerTest.java
chanseokoh: Oh, actually, the order of invocation is the opposite.

coollog: Hmm, yeah, that's true, having it one-after-another is the same and is a lot cleaner.

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/MultithreadedExecutor.java
```diff
@@ -0,0 +1,63 @@
...
+  private static final Duration MULTITHREADED_TEST_TIMEOUT = Duration.ofSeconds(1);
+  private static final int THREAD_COUNT = 20;
+
+  private final ExecutorService executorService = Executors.newFixedThreadPool(THREAD_COUNT);
+
+  <E> List<E> invokeAll(List<Callable<E>> callables)
+      throws InterruptedException, ExecutionException {
+    List<Future<E>> futures =
+        executorService.invokeAll(
+            callables, MULTITHREADED_TEST_TIMEOUT.getSeconds(), TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
coollog: The `invokeAll` is the method that blocks and `MULTITHREADED_TEST_TIMEOUT` avoids having our test run forever in the case we have a deadlock.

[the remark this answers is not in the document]
chanseokoh: Yeah, but `invokeAll()` returns all future regardless of if some tasks timed out or not. That means, in the next line, you call `Future.get()` on all of them, so the whole method will block until all of the given tasks are completed. Did I miss something?

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
```diff
@@ -0,0 +1,162 @@
...
+   * <p>For any {@link Allocation}, this method <em>must</em> have been called on all of its parents
+   * beforehand.
+   *
+   * @param allocation the {@link Allocation} to update progress for
+   * @param units the units of progress
+   * @return {@code true} if the map was updated; {@code false} if {@code allocation} was already
+   *     present. Note that there could be false positives if called concurrently, but never false
+   *     negatives.
+   */
+  boolean updateProgress(Allocation allocation, long units) {
```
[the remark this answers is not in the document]
coollog: The 0-unit progress event will actually be sent whenever we start a new task to tell the progress handler about that new task that is in progress.

`updateProgress(nsertionOrderUnits, units)` changes the `InsertionOrderUnits` but doesn't actually update the `ConcurrentHashMap`. Synchronizing the method would prevent other `InsertionOrderUnits` from being updated at the same time and reduce the efficiency of concurrency here.

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/MultithreadedExecutor.java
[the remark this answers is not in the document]
coollog: Ah that's true, I will have the timeout applied to `get` as well then.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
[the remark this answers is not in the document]
chanseokoh: > `updateProgress(nsertionOrderUnits, units)` changes the `InsertionOrderUnits` but doesn't actually update the `ConcurrentHashMap`.

Ah, I see. I was confused by the same method name. Having the helper name same as the original name seems a bit confusing. How about `updateInsertionOrderUnits()` or something else?

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
```diff
@@ -107,7 +107,9 @@ boolean updateProgress(Allocation allocation, long units) {
 
   /**
    * Gets a list of the unfinished {@link Allocation}s in the order in which those {@link
-   * Allocation}s were encountered.
+   * Allocation}s were encountered. This can be used to display, for example, currently executing
+   * tasks. The order helps to keep the displayed tasks in a deterministic order (new subtasks
```
chanseokoh: > new subtasks appear below older ones

I personally think it doesn't really matter if a new subtask appear above older ones, as long as the displayer does not dynamically change the order once it displayed some tasks. Not having to think about the order here will make the code simpler, and the displayer just needs to manage the order consistently on its side. What do you think?

coollog: One of the things I've noticed in the prototype is that since a normal map doesn't give a deterministic order, when a new subtask is added, it could change the relative orders of tasks that existed before (when getting the unfinished tasks in some order from the map). This makes it sometimes hard to keep track of which is the new task that started vs an old task that is still running. We could manage the order from the displayer side but that would mean that the displayer would need to keep some kind of map/association from allocation to an order index and sort all unfinished ones by the order index, which results in concurrent code that would basically be the same. I think we can explore moving the order tracking to the display side once we add that and see which way is cleaner?

chanseokoh: > which results in concurrent code that would basically be the same.

Yeah, I can imagine that too. Let's move forward with managing the order here.

jib-core/src/test/java/com/google/cloud/tools/jib/event/progress/MultithreadedExecutor.java
```diff
@@ -0,0 +1,71 @@
...
+
+  <E> List<E> invokeAll(List<Callable<E>> callables)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    List<Future<E>> futures =
+        executorService.invokeAll(
+            callables, MULTITHREADED_TEST_TIMEOUT.getSeconds(), TimeUnit.SECONDS);
+
+    List<E> returnValues = new ArrayList<>();
+    for (Future<E> future : futures) {
+      returnValues.add(future.get(MULTITHREADED_TEST_TIMEOUT.getSeconds(), TimeUnit.SECONDS));
```
chanseokoh: nit: actually, it makes more sense to have a timeout of being close to 0 here. (0 might mean infinity.) Otherwise, you are basically giving twice of the intended timeout, since you already waited at `invokeAll`.

coollog: Hmm, I'll actually just assert that they are done then.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/ProgressEventHandler.java
```diff
@@ -0,0 +1,86 @@
...
+
+  @Override
+  public void accept(ProgressEvent progressEvent) {
+    Allocation allocation = progressEvent.getAllocation();
+    long progressUnits = progressEvent.getUnits();
+    double allocationFraction = allocation.getFractionOfRoot();
+    long allocationUnits = allocation.getAllocationUnits();
+
+    if (progressUnits != 0) {
+      progress.add(progressUnits * allocationFraction / allocationUnits);
```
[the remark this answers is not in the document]
coollog: Yea, that sounds good, I'll change it in this PR.

PR #1312, opened 2018-12-06
Change StepsRunner to be more of a builder to prepare for assigning Allocations.

coollog:
Part of #1297 

This defers running the steps till the end and allows for `StepsRunner` to know how many steps there are before running them all.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
@@ -54,8 +54,7 @@ public class BuildSteps {
     return new BuildSteps(
         DESCRIPTION_FOR_DOCKER_REGISTRY,
         buildConfiguration,
-        () ->
-            StepsRunner.begin(buildConfiguration)
+        StepsRunner.begin(buildConfiguration)
                 .retrieveTargetRegistryCredentials()
                 .authenticatePush()
                 .pullBaseImage()
@@ -67,7 +66,7 @@ public class BuildSteps {
                 .pushApplicationLayers()
                 .finalizingPush()
                 .pushImage()
-                .waitOnPushImage());
+            ::waitOnPushImage);
   }
 
   /**
@@ -82,15 +81,14 @@ public class BuildSteps {
     return new BuildSteps(
         DESCRIPTION_FOR_DOCKER_DAEMON,
         buildConfiguration,
-        () ->
-            StepsRunner.begin(buildConfiguration)
+        StepsRunner.begin(buildConfiguration)
                 .pullBaseImage()
                 .pullAndCacheBaseImageLayers()
                 .buildAndCacheApplicationLayers()
                 .buildImage()
                 .finalizingBuild()
                 .loadDocker(dockerClient)
-                .waitOnLoadDocker());
+            ::waitOnLoadDocker);
   }
 
   /**
@@ -104,15 +102,14 @@ public class BuildSteps {
     return new BuildSteps(
         DESCRIPTION_FOR_TARBALL,
         buildConfiguration,
-        () ->
-            StepsRunner.begin(buildConfiguration)
+        StepsRunner.begin(buildConfiguration)
                 .pullBaseImage()
                 .pullAndCacheBaseImageLayers()
                 .buildAndCacheApplicationLayers()
                 .buildImage()
                 .finalizingBuild()
                 .writeTarFile(outputPath)
-                .waitOnWriteTarFile());
+            ::waitOnWriteTarFile);
   }
 
   private final String description;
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -25,192 +25,256 @@
...
+    @Nullable private WriteTarFileStep writeTarFileStep;
+  }
 
+  /**
+   * Starts building the steps to run.
+   *
+   * @param buildConfiguration the {@link BuildConfiguration}
+   * @return a new {@link StepsRunner}
+   */
+  public static StepsRunner begin(BuildConfiguration buildConfiguration) {
```
chanseokoh: As a mere static instance creator, wouldn't below make more sense?
```java
public static StepsRunner begin(BuildConfiguration buildConfiguration) {
  return new StepsRunner(buildConfiguration);
}
```
(While just making the `StepsRunner` constructor `private` from `public`, without really touching it.) Not getting why taking out part of the existing, internal constructor code into the static initializer.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
```diff
@@ -55,19 +55,19 @@ public static BuildSteps forBuildToDockerRegistry(BuildConfiguration buildConfig
         DESCRIPTION_FOR_DOCKER_REGISTRY,
         buildConfiguration,
         () ->
```
chanseokoh: Looks like the sole reason we were accepting a lambda here was not to run `StepsRunner` right away. With the change to `StepsRunner` in this PR, just instantiating `StepsRunner` with `.begin()` won't run the steps right away (if we did not call the `wait...Image()` methods.) Previously, `BuildSteps.run()` would kick the steps. So, what if we remove the lambda wrapping that now seems unnecessary? That is,

1. Revert `wait...Image()` and remove the call to `runStepsRunners()`.
1. Change the `BuildSteps` constructor to accept a `StepsRunner` instance, instead of a lambda wrapping that.
1. Change `BuildSteps.run()` to call `StepsRunner.runStepsRunner()` there.

Haven't really tested out this, but I think this will work?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -25,192 +25,256 @@
...
+    this.listeningExecutorService = listeningExecutorService;
+    this.buildConfiguration = buildConfiguration;
   }
 
-  public StepsRunner runRetrieveTargetRegistryCredentialsStep() {
-    retrieveTargetRegistryCredentialsStep =
-        RetrieveRegistryCredentialsStep.forTargetImage(
-            listeningExecutorService, buildConfiguration);
+  public StepsRunner retrieveTargetRegistryCredentials() {
+    stepsRunnables.add(
```
chanseokoh: If all you need is a number of steps added, another way to achieve this is, instead of maintaining a list of runnables, you have a single runnable field, and do

```java
stepsAdded++;
Runnable previousRunnable = stepsRunnable;
stepsRunnable = () -> {
  previousRunnable.run();
  steps.retrieveTargetRegistryCredentialsStep = ...
}
```
And `runStepsRunners()` just becomes `stepsRunnable.run()`. But I think the list is OK too. I was just thinking maintaining a list might make readers think you really need to maintain a list of these for some reason.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
coollog: This is mostly following the idea that potentially costly logic should go in a static initializer rather than a constructor, which should not have any costly logic. This might not be necessary here but it's under the assumption that creating an `ExecutorService` could potentially be a costly operation.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildSteps.java
coollog: Yep that would work! Will change it to that

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
coollog: Sounds good! Will do.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -34,183 +34,238 @@
...
+    stepsRunnable.run();
+    return Preconditions.checkNotNull(steps.writeTarFileStep).getFuture().get();
   }
 
-  public BuildResult waitOnWriteTarFileStep() throws ExecutionException, InterruptedException {
-    return Preconditions.checkNotNull(writeTarFileStep).getFuture().get();
+  private StepsRunner enqueueStep(Runnable stepRunnable) {
+    stepsRunnable =
+        () -> {
+          stepsRunnable.run();
```
chanseokoh: Hmm... not sure if this will work, because it's defined recursively using itself. Needs verification.

coollog: Ah right, needs to finalize in another variable.

PR #1319, opened 2018-12-07
Minor updates to Dropwizard example

chanseokoh:
Some simplification after #1290.

It's tempting to remove the `package` phase binding too (and the command becomes `./mvnw clean compile jib:dockerBuild`), but I guess that's fine.


chanseokoh:
@cricket007 for your interest.

PR #1320, opened 2018-12-07
 Adds AsyncDependencies to reduce boilerplate code.

### LINKED ISSUE #1317: Refactor AsyncSteps to reduce dependency boilerplate.
Suggested by @GuustavoPaiva :

For example, in the `LoadDockerStep` these are defined in the following way:

```java
  @Override
  public BuildResult call() throws ExecutionException, InterruptedException {
    ImmutableList.Builder<ListenableFuture<?>> dependenciesBuilder = ImmutableList.builder();
    for (PullAndCacheBaseImageLayerStep pullAndCacheBaseImageLayerStep :
        NonBlockingSteps.get(pullAndCacheBaseImageLayersStep)) {
      dependenciesBuilder.add(pullAndCacheBaseImageLayerStep.getFuture());
    }
    for (BuildAndCacheApplicationLayerStep buildAndCacheApplicationLayerStep :
        buildAndCacheApplicationLayerSteps) {
      dependenciesBuilder.add(buildAndCacheApplicationLayerStep.getFuture());
    }
    dependenciesBuilder.add(NonBlockingSteps.get(buildImageStep).getFuture());
    return Futures.whenAllSucceed(dependenciesBuilder.build())
        .call(this::afterPushBaseImageLayerFuturesFuture, listeningExecutorService)
        .get();
  }
```

What if we wrap this in a clearer way, like:

```java
  @Override
  public BuildResult call() throws ExecutionException, InterruptedException {
    return new StepDependencies()
        .ofStepResult(pullAndCacheBaseImageLayersStep)
        .ofStepResult(buildImageStep)
        .ofStep(buildAndCacheApplicationLayerSteps)
        .whenAllSucceed()
        .call(this::afterPushBaseImageLayerFuturesFuture, listeningExecutorService)
        .get();
  }
```

coollog:
Fixes #1317

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.async;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+
+public class AsyncDependencies {
+
+  public static AsyncDependencies using(ListeningExecutorService listeningExecutorService) {
+    return new AsyncDependencies(listeningExecutorService);
+  }
+
+  private final ListeningExecutorService listeningExecutorService;
+  private final List<ListenableFuture<?>> futures = new ArrayList<>();
+
+  private AsyncDependencies(ListeningExecutorService listeningExecutorService) {
+    this.listeningExecutorService = listeningExecutorService;
+  }
+
+  public AsyncDependencies addStep(AsyncStep<?> asyncStep) {
+    futures.add(asyncStep.getFuture());
+    return this;
+  }
+
+  public AsyncDependencies addStepOfStep(AsyncStep<? extends AsyncStep<?>> asyncStepOfAsyncStep)
+      throws ExecutionException {
+    return addStep(NonBlockingSteps.get(asyncStepOfAsyncStep));
+  }
+
+  public AsyncDependencies addStepOfStepOfStep(
+      AsyncStep<? extends AsyncStep<? extends AsyncStep<?>>> asyncStepOfAsyncStepOfAsyncStep)
+      throws ExecutionException {
+    return addStepOfStep(NonBlockingSteps.get(asyncStepOfAsyncStepOfAsyncStep));
+  }
+
+  public AsyncDependencies addListOfSteps(
+      AsyncStep<? extends List<? extends AsyncStep<?>>> asyncStepOfAsyncSteps)
+      throws ExecutionException {
+    for (AsyncStep<?> asyncStep : NonBlockingSteps.get(asyncStepOfAsyncSteps)) {
+      addStep(asyncStep);
+    }
+    return this;
+  }
+
+  public AsyncDependencies addListOfStepOfSteps(
+      AsyncStep<? extends ImmutableList<? extends AsyncStep<? extends AsyncStep<?>>>>
+          asyncStepOfAsyncStepsOfAsyncStep)
+      throws ExecutionException {
+    for (AsyncStep<? extends AsyncStep<?>> asyncStepOfAsyncStep :
+        NonBlockingSteps.get(asyncStepOfAsyncStepsOfAsyncStep)) {
+      addStepOfStep(asyncStepOfAsyncStep);
+    }
+    return this;
+  }
+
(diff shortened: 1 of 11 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
```diff
@@ -0,0 +1,137 @@
...
+   * Adds the future of an {@link AsyncStep} wrapped in an outer {@link AsyncStep}.
+   *
+   * @param asyncStepOfAsyncStep the outer {@link AsyncStep}
+   * @return this
+   * @throws ExecutionException if an exception occurs during execution of {@code
+   *     asyncStepOfAsyncStep}
+   */
+  public AsyncDependencies addStepOfStep(AsyncStep<? extends AsyncStep<?>> asyncStepOfAsyncStep)
+      throws ExecutionException {
+    return addStep(NonBlockingSteps.get(asyncStepOfAsyncStep));
```
chanseokoh: IMO, I think it is safer to not call `NonBlockingSteps.get()` in this helper class. Using it may simplify caller's code in a mechanical way, but I think assuming that the given outermost step must have been completed may become a bit danger when people start using this without knowing the details. As a async dependency handler, I'd prefer that whatever asyn step you throw to the handler, whether completed or now, it should work. After all, I think it'd be awkward put in the Javadoc of this method that "the given `asyncStepOfAsyncStep` must have been completed."

That said, you may lose the extra utility methods, but I think it might actually make this class easiler to understand. These additional methods seem like too case-specific with the assumption that the outer step is completed. Things like `asyncStepOfAsyncStepsOfAsyncStep` or `asyncStepOfAsyncStepOfAsyncStep` sound a bit confusing too. At least it seems better to not call `NonBlockingSteps.get()`.

coollog: Agreed, will try to move `NonBlockingSteps.get()` out of this class.

jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.concurrent.Callable;
+
+/**
+ * Builds a list of dependency {@link ListenableFuture}s to wait on before calling a {@link
+ * Callable}.
+ */
+public class AsyncDependencies {
+
+  /**
+   * Initialize with a {@link ListeningExecutorService}.
```
chanseokoh: See if this can just be `ExecutorService`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
```diff
@@ -77,19 +77,12 @@
...
-      dependencies.add(pullAndCacheBaseImageLayerStep.getFuture());
-    }
-    for (BuildAndCacheApplicationLayerStep buildAndCacheApplicationLayerStep :
-        buildAndCacheApplicationLayerSteps) {
-      dependencies.add(buildAndCacheApplicationLayerStep.getFuture());
-    }
+    AsyncDependencies dependencies =
+        AsyncDependencies.using(listeningExecutorService)
+            .addSteps(NonBlockingSteps.get(pullAndCacheBaseImageLayersStep));
+    buildAndCacheApplicationLayerSteps.forEach(dependencies::addStep);
```
chanseokoh: Instead of `forEach`, can this be chained with `.addSteps(buildAndCacheApplicationLayerSteps)`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/FinalizingStep.java
```diff
@@ -53,16 +52,10 @@
...
-    for (AsyncStep<?> dependency : wrappedDependencyLists) {
-      dependenciesFutures.add(dependency.getFuture());
-    }
-    for (AsyncStep<?> dependency : dependencyList) {
-      dependenciesFutures.add(dependency.getFuture());
-    }
-    listenableFuture =
-        Futures.whenAllSucceed(dependenciesFutures).call(this, listeningExecutorService);
+    AsyncDependencies dependencies = AsyncDependencies.using(listeningExecutorService);
+    wrappedDependencyLists.forEach(dependencies::addStep);
```
chanseokoh: ditto, I see a lot of `forEach`, and looks like you can chain `addSteps()`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
```diff
@@ -86,38 +86,32 @@
...
+    AsyncDependencies dependencies = AsyncDependencies.using(listeningExecutorService);
+    for (AsyncStep<PushBlobStep> pushBaseImageLayerStep :
+        NonBlockingSteps.get(pushBaseImageLayersStep)) {
+      dependencies.addStep(NonBlockingSteps.get(pushBaseImageLayerStep));
     }
-    for (AsyncStep<PushBlobStep> pushBlobStepStep :
+    for (AsyncStep<PushBlobStep> pushApplicationLayerStep :
         NonBlockingSteps.get(pushApplicationLayersStep)) {
-      dependencies.add(NonBlockingSteps.get(pushBlobStepStep).getFuture());
+      dependencies.addStep(NonBlockingSteps.get(pushApplicationLayerStep));
```
chanseokoh: I wonder `NonBlockingSteps.get(pushApplicationLayersStep).stream().map(NonBlockingSteps::get).forEach(dependencies::addStep)` or something like may look better. Not that I insist using `stream()` though.

jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
coollog: The `ListeningExecutorService` is necessary for returning the `ListenableFuture` in `whenAllSucceed`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
coollog: The reason for not using `stream` is that `NonBlockingSteps.get` may throw an `ExecutionException` that can't be propagated directly from the lambda.

PR #1321, opened 2018-12-07
Assigns Allocation to each builder step.

coollog:
Part of #1297 

Every builder step now emits `ProgressEvent`s. This will be refactored once `ProgressEventDispatcher` is in.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -24,6 +24,8 @@ import com.google.cloud.tools.jib.builder.steps.PullBaseImageStep.BaseImageWithA
 import com.google.cloud.tools.jib.configuration.BuildConfiguration;
 import com.google.cloud.tools.jib.configuration.credentials.Credential;
 import com.google.cloud.tools.jib.event.events.LogEvent;
+import com.google.cloud.tools.jib.event.events.ProgressEvent;
+import com.google.cloud.tools.jib.event.progress.Allocation;
 import com.google.cloud.tools.jib.http.Authorization;
 import com.google.cloud.tools.jib.http.Authorizations;
 import com.google.cloud.tools.jib.image.Image;
@@ -82,12 +84,16 @@ class PullBaseImageStep
   }
 
   private final BuildConfiguration buildConfiguration;
+  private final Allocation parentProgressAllocation;
 
   private final ListenableFuture<BaseImageWithAuthorization> listenableFuture;
 
   PullBaseImageStep(
-      ListeningExecutorService listeningExecutorService, BuildConfiguration buildConfiguration) {
+      ListeningExecutorService listeningExecutorService,
+      BuildConfiguration buildConfiguration,
+      Allocation parentProgressAllocation) {
     this.buildConfiguration = buildConfiguration;
+    this.parentProgressAllocation = parentProgressAllocation;
 
     listenableFuture = listeningExecutorService.submit(this);
   }
@@ -109,12 +115,19 @@ class PullBaseImageStep
                 "Getting base image "
                     + buildConfiguration.getBaseImageConfiguration().getImage()
                     + "..."));
+    Allocation progressAllocation = parentProgressAllocation.newChild("pull base image", 2);
+    buildConfiguration.getEventDispatcher().dispatch(new ProgressEvent(progressAllocation, 0));
 
     try (TimerEventDispatcher ignored =
         new TimerEventDispatcher(buildConfiguration.getEventDispatcher(), DESCRIPTION)) {
       // First, try with no credentials.
       try {
-        return new BaseImageWithAuthorization(pullBaseImage(null), null);
+        BaseImageWithAuthorization baseImageWithAuthorization =
+            new BaseImageWithAuthorization(pullBaseImage(null), null);
+
+        buildConfiguration.getEventDispatcher().dispatch(new ProgressEvent(progressAllocation, 2));
+
+        return baseImageWithAuthorization;
 
       } catch (RegistryUnauthorizedException ex) {
         buildConfiguration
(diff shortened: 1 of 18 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -87,20 +98,30 @@ static RetrieveRegistryCredentialsStep forTargetImage(
...
+    buildConfiguration.getEventDispatcher().dispatch(new ProgressEvent(progressAllocation, 0));
+
+    try (TimerEventDispatcher ignored =
+        new TimerEventDispatcher(buildConfiguration.getEventDispatcher(), description)) {
       for (CredentialRetriever credentialRetriever : credentialRetrievers) {
         Optional<Credential> optionalCredential = credentialRetriever.retrieve();
         if (optionalCredential.isPresent()) {
+          buildConfiguration
+              .getEventDispatcher()
+              .dispatch(new ProgressEvent(progressAllocation, 1));
```
chanseokoh: Another way is to put this is to add a `finally` block to the existing `try`.

If there is no existing `try`, you can add one, but another way would be to wrap it in another method, like

```java
Object call() {
  try {
   send 0-unit event
   return realCall();
  } finally {
    send completion event
  }
}

Object realCall() {
  // the original method
}
```

But sometimes, just putting the final event in two places can be simpler, and I have no object to it.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
```diff
@@ -86,9 +92,15 @@
   private PushBlobStep makePushBlobStep(AsyncStep<? extends CachedLayer> cachedLayerStep)
       throws ExecutionException {
     CachedLayer cachedLayer = NonBlockingSteps.get(cachedLayerStep);
+
+    Allocation progressAllocation =
+        parentProgressAllocation.newChild("push layer " + cachedLayer.getDigest(), 1);
+    buildConfiguration.getEventDispatcher().dispatch(new ProgressEvent(progressAllocation, 0));
```
chanseokoh: Ah, I thought why it did not have the completion event, but now I think `PushBlobStep` will do it (and we need that kind of premise).

If so, I can see that passing down `parentProgressAllocation` to `PushBlobStep` will do the same if we only think about the entire allocation units, while the difference is that you'll have one more depth in the tree, right?

So, it looks like for this child, the allocation should always be `1`, right?

So these have been the questions I had from the start, like whether to pass down the parent or create a child unit with a single allocation, what if you wanted to pass down a node to multiple sequential calls, what if we also want to do some progress report at this node, etc.

Anyway, we are here at this point, and I guess we'll move forward with this.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -140,8 +154,15 @@ public BaseImageWithAuthorization call()
...
         try {
-          return new BaseImageWithAuthorization(
-              pullBaseImage(registryAuthorization), registryAuthorization);
+          BaseImageWithAuthorization baseImageWithAuthorization =
+              new BaseImageWithAuthorization(
+                  pullBaseImage(registryAuthorization), registryAuthorization);
+
+          buildConfiguration
+              .getEventDispatcher()
+              .dispatch(new ProgressEvent(progressAllocation, 1));
```
chanseokoh: I think this should be `2` here. At this point, you always return successfully. `finally` may work?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -164,8 +185,15 @@ public BaseImageWithAuthorization call()
...
 
-          return new BaseImageWithAuthorization(
-              pullBaseImage(registryAuthorization), registryAuthorization);
+          BaseImageWithAuthorization baseImageWithAuthorization =
+              new BaseImageWithAuthorization(
+                  pullBaseImage(registryAuthorization), registryAuthorization);
+
+          buildConfiguration
+              .getEventDispatcher()
+              .dispatch(new ProgressEvent(progressAllocation, 1));
```
chanseokoh: ditto, I think this should be `2`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
coollog: Yea, this will be completely refactored to not include so much duplication after `ProgressEventDispatcher` is added.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
coollog: Yea, basically the constraint is that each step creates its `Allocation` knowing how many children it will have, and passes down the `Allocation` to the children, which each parent does not know how many allocation units each of its children's `Allocation` have.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
coollog: `RetrieveRegistryCredentialsStep` takes up another unit.

coollog:
Seems to have broken master: https://fusion.corp.google.com/runanalysis/info/prod%3Ajib%2Fgcp_ubuntu%2Fcontinuous/prod%3Ajib%2Fgcp_ubuntu%2Fcontinuous/KOKORO/917339b3-38f2-43b8-bade-a76aaa3c1ec3/1544213544048/prod%3Ajib%2Fgcp_ubuntu%2Fcontinuous%20Build%20%23722/Targets

chanseokoh:
Is it this PR?

chanseokoh:
`SingleProjectIntegrationTest` succeeds locally. Could have been flakyness.

chanseokoh:
Triggering the rebuild on the same commit succeeded too.

coollog:
Hmm, yea, it might be flakiness, but we should probably investigate a bit more to make sure there isn't some underlying problem here.

PR #1322, opened 2018-12-07
Use artifact ID as image ref for dockerBuild

### LINKED ISSUE #1318: When to.image is not set, don't use <name> in pom.xml as image reference
Say we have below, and `<to><image>`is undefined.

```xml
    <groupId>example</groupId>
    <artifactId>dropwizard</artifactId>
    <version>1</version>
    <packaging>jar</packaging>

    <name>Dropwizard Jib Example</name>
```

When doing `jib:dockerBuild`, you get an error: `Invalid image reference: Dropwizard Jib Example:1`.

Looks like the value of `<name>` is not good for an image reference: https://maven.apache.org/guides/introduction/introduction-to-the-pom.html

I think we should pick the `<artifactId>`.

chanseokoh:
Fixes #1318.

### DIFF
--- a/jib-maven-plugin/src/test/resources/projects/simple/pom-no-to-image.xml
+++ b/jib-maven-plugin/src/test/resources/projects/simple/pom-no-to-image.xml
@@ -1,0 +1,50 @@
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.test</groupId>
+  <artifactId>my-artifact-id</artifactId>
+  <version>1</version>
+
+  <name>Name not Usable as Image Reference</name>
+  
+  <properties>
+    <java.target.version>1.8</java.target.version>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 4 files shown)


jib-maven-plugin/src/test/resources/projects/simple/pom-no-to-image.xml
```diff
@@ -0,0 +1,50 @@
+<project>
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.test</groupId>
+  <artifactId>my-artifact-id</artifactId>
+  <version>1</version>
+
+  <name>Name not Usable as Image Reference</name>
```
coollog: To not have to have another pom, what about just setting the `simple/pom.xml`'s `<name>`?

chanseokoh: `simple/pom.xml` defines `<to><image>`, so I had to create a new one without it. I think this is simpler.

coollog: Ah okay.

PR #1323, opened 2018-12-07
Change RegistryClient push/pull to take callback to listen on progress.

coollog:
Part of #1297 

This will be used to populate `ProgressEvent`s once `ProgressEventDispatcher` is added.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -243,6 +243,7 @@ class PullBaseImageStep
             Blobs.writeToString(
                 registryClient.pullBlob(
                     v22ManifestTemplate.getContainerConfiguration().getDigest(),
+                    // TODO: Replace with progress updates.
                     ignored -> {},
                     ignored -> {}));
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -214,7 +214,9 @@ class PullBaseImageStep
         String containerConfigurationString =
             Blobs.writeToString(
                 registryClient.pullBlob(
-                    v22ManifestTemplate.getContainerConfiguration().getDigest()));
+                    v22ManifestTemplate.getContainerConfiguration().getDigest(),
+                    ignored -> {},
+                    ignored -> {}));
 
         ContainerConfigurationTemplate containerConfigurationTemplate =
             JsonTemplateMapper.readJson(
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
@@ -20,16 +20,22 @@ import com.google.api.client.http.HttpContent;
 import com.google.cloud.tools.jib.blob.Blob;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.time.Duration;
+import java.util.function.Consumer;
 
 /** {@link Blob}-backed {@link HttpContent}. */
 public class BlobHttpContent implements HttpContent {
 
   private final Blob blob;
   private final String contentType;
+  // TODO: Refactor into BlobPushMonitor or something.
+  private final Consumer<Long> sentByteCountConsumer;
+  private final Duration delayBetweenCallbacks = Duration.ofMillis(100);
 
-  public BlobHttpContent(Blob blob, String contentType) {
+  public BlobHttpContent(Blob blob, String contentType, Consumer<Long> sentByteCountConsumer) {
     this.blob = blob;
     this.contentType = contentType;
+    this.sentByteCountConsumer = sentByteCountConsumer;
   }
 
   @Override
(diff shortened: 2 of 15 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -242,7 +242,9 @@ public BaseImageWithAuthorization call()
         String containerConfigurationString =
             Blobs.writeToString(
                 registryClient.pullBlob(
-                    v22ManifestTemplate.getContainerConfiguration().getDigest()));
+                    v22ManifestTemplate.getContainerConfiguration().getDigest(),
+                    ignored -> {},
```
chanseokoh: Just in case we don't forget: // TODO: Replace ...

jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
```diff
@@ -20,16 +20,22 @@
...
+import java.util.function.Consumer;
 
 /** {@link Blob}-backed {@link HttpContent}. */
 public class BlobHttpContent implements HttpContent {
 
   private final Blob blob;
   private final String contentType;
+  // TODO: Refactor into BlobPushMonitor or something.
+  private final Consumer<Long> sentByteCountConsumer;
+  private final Duration delayBetweenCallbacks = Duration.ofMillis(100);
```
chanseokoh: This is more like a `static final` and all caps (The class `Duration` is immutable.)

jib-core/src/main/java/com/google/cloud/tools/jib/registry/BlobPuller.java
```diff
@@ -42,18 +45,31 @@
    */
   private final OutputStream destinationOutputStream;
 
+  // TODO: Refactor into BlobPullMonitor.
+  private final Consumer<Long> sizeConsumer;
+  private final Consumer<Long> receivedByteCountConsumer;
+  private final Duration delayBetweenCallbacks = Duration.ofMillis(100);
```
chanseokoh: ditto

jib-core/src/test/java/com/google/cloud/tools/jib/http/ConnectionTest.java
```diff
@@ -58,6 +58,8 @@
   private Request fakeRequest;
   private HttpResponse mockHttpResponse;
 
+  private long byteCount = 0;
```
chanseokoh: Use a local `LongAdder` like others and remove this?

jib-core/src/test/java/com/google/cloud/tools/jib/registry/BlobPusherTest.java
```diff
@@ -41,7 +44,9 @@
 @RunWith(MockitoJUnitRunner.class)
 public class BlobPusherTest {
 
-  @Mock private Blob mockBlob;
+  private static final String testBlobContent = "some BLOB content";
```
chanseokoh: Why not all caps?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/BlobPuller.java
```diff
@@ -42,18 +45,31 @@
...
   BlobPuller(
       RegistryEndpointRequestProperties registryEndpointRequestProperties,
       DescriptorDigest blobDigest,
-      OutputStream destinationOutputStream) {
+      OutputStream destinationOutputStream,
+      Consumer<Long> sizeConsumer,
+      Consumer<Long> receivedByteCountConsumer) {
     this.registryEndpointRequestProperties = registryEndpointRequestProperties;
     this.blobDigest = blobDigest;
     this.destinationOutputStream = destinationOutputStream;
```
chanseokoh: What if do `this.destinationOutputStream = new ListenableCountingOutputStream(...)`? Doesn't have to save `receivedByteCountConsumer` in a field.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/BlobPuller.java
```diff
@@ -42,18 +45,31 @@
...
 
   @Override
   public Void handleResponse(Response response) throws IOException, UnexpectedBlobDigestException {
-    try (OutputStream outputStream = destinationOutputStream) {
+    sizeConsumer.accept(response.getContentLength());
+
+    try (OutputStream outputStream =
+        new ListenableCountingOutputStream(
+            destinationOutputStream, receivedByteCountConsumer, delayBetweenCallbacks)) {
       BlobDescriptor receivedBlobDescriptor = response.getBody().writeTo(outputStream);
```
chanseokoh: For `BlobHttpContent`, you create `new ListenableCountingOutputStream` inside `writeTo()`. It may just be that the names are idential, but does it make sense to create the stream inside `Blob.writeTo()` similarly? (Probably not.)

jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
coollog: I intend to actually refactor this and `sentByteCountConsumer` to a `BlobPushMonitor` so that the actual delay can be configured from the outside and not intending `delayBetweenCallbacks` to remain a constant.

jib-core/src/test/java/com/google/cloud/tools/jib/http/ConnectionTest.java
coollog: Using a local var would mean having to pass it between `setUpMocksAndFakes` and `testSend`, which is used by most of the tests.

PR #1326, opened 2018-12-07
Adds SingleThreadedLogger to be able to keep a footer for where progress indication will go.

coollog:
Part of #1297 

This will be gradually improved once it is used in the `LogEventHandler`.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
@@ -41,6 +41,12 @@ class AnsiLoggerWithFooter {
   /** ANSI escape sequence for erasing to end of display. */
   private static final String ERASE_DISPLAY_BELOW = "\033[0J";
 
+  /** ANSI escape sequence for setting all further characters to bold. */
+  private static final String BOLD = "\033[1m";
+
+  /** ANSI escape sequence for setting all further characters to not bold. */
+  private static final String UNBOLD = "\033[0m";
+
   private static final Duration EXECUTOR_SHUTDOWN_WAIT = Duration.ofSeconds(1);
 
   private final ExecutorService executorService = Executors.newSingleThreadExecutor();
@@ -81,6 +87,8 @@ class AnsiLoggerWithFooter {
    * Sets the footer asynchronously. This will replace the previously-printed footer with the new
    * {@code footerLines}.
    *
+   * <p>The footer is printed in <strong>bold</strong>.
+   *
    * @param footerLines the footer, with each line as an element (no newline at end)
    * @return a {@link Future} to track completion
    */
@@ -110,7 +118,16 @@ class AnsiLoggerWithFooter {
           // Writes out logMessage and footer.
           plainLogger.accept(plainLogBuilder.toString());
           messageLogger.run();
-          plainLogger.accept(String.join("\n", newFooterLines));
+
+          if (newFooterLines.size() > 0) {
+            StringBuilder footerBuilder = new StringBuilder();
+            for (String newFooterLine : newFooterLines) {
+              footerBuilder.append(BOLD).append(newFooterLine).append(UNBOLD).append('\n');
+            }
+            // Removes last newline.
+            footerBuilder.setLength(footerBuilder.length() - 1);
+            plainLogger.accept(footerBuilder.toString());
+          }
 
           this.footerLines = newFooterLines;
 
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
@@ -26,9 +26,9 @@ import java.util.function.Consumer;
 
 /**
  * Keeps all log messages in a sequential, deterministic order along with an additional footer that
- * always appears below log messages.
+ * always appears below log messages. This is intended to log both the messages and the footer to the same console.
  */
-class SingleThreadedLogger {
+class AnsiLoggerWithFooter {
 
   /** ANSI escape sequence for moving the cursor up one line. */
   private static final String CURSOR_UP_SEQUENCE = "\033[1A";
@@ -42,7 +42,7 @@ class SingleThreadedLogger {
 
   private List<String> footerLines = Collections.emptyList();
 
-  SingleThreadedLogger(Consumer<String> plainLogger) {
+  AnsiLoggerWithFooter(Consumer<String> plainLogger) {
     this.plainLogger = plainLogger;
   }
 
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
@@ -31,7 +31,7 @@ import java.util.function.Consumer;
  * always appears below log messages. This is intended to log both the messages and the footer to
  * the same console.
  *
- * Make sure to call {@link #shutDown} when finished.
+ * <p>Make sure to call {@link #shutDown} when finished.
  */
 class AnsiLoggerWithFooter {
 
@@ -57,7 +57,8 @@ class AnsiLoggerWithFooter {
   public void shutDown() {
     executorService.shutdown();
     try {
-      if (!executorService.awaitTermination(EXECUTOR_SHUTDOWN_WAIT.getSeconds(), TimeUnit.SECONDS)) {
+      if (!executorService.awaitTermination(
+          EXECUTOR_SHUTDOWN_WAIT.getSeconds(), TimeUnit.SECONDS)) {
         executorService.shutdownNow();
       }
     } catch (InterruptedException ex) {
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
@@ -26,7 +26,8 @@ import java.util.function.Consumer;
 
 /**
  * Keeps all log messages in a sequential, deterministic order along with an additional footer that
- * always appears below log messages. This is intended to log both the messages and the footer to the same console.
+ * always appears below log messages. This is intended to log both the messages and the footer to
+ * the same console.
  */
 class AnsiLoggerWithFooter {
 
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
@@ -62,6 +62,10 @@ class AnsiLoggerWithFooter {
   /** Shuts down the {@link #executorService}. */
   public void shutDown() {
     executorService.shutdown();
+  }
+
+  /** Waits for the {@link #executorService} to terminate. */
+  public void awaitTermination() {
     try {
       if (!executorService.awaitTermination(
           EXECUTOR_SHUTDOWN_WAIT.getSeconds(), TimeUnit.SECONDS)) {
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
@@ -17,17 +17,21 @@
 package com.google.cloud.tools.jib.plugins.common;
 
 import com.google.common.util.concurrent.Futures;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
 /**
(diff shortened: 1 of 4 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,110 @@
...
+  /**
+   * Sets the footer asynchronously.
+   *
+   * @param footer the footer
+   * @param lineCount the number of lines in the footer
+   * @return a {@link Future} to track completion
+   */
+  public Future<Void> setFooter(String footer, int lineCount) {
+    if (footer.equals(this.footer)) {
+      return Futures.immediateFuture(null);
```
chanseokoh: Just saying: it's possible that the `footer` may have changed and no longer equal at this point due to concurrency, but I get that it's OK anyway, in that there should always be losers.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,110 @@
...
+            // Moves cursor up.
+            plainLogBuilder.append(CURSOR_UP_SEQUENCE);
+          }
+
+          // Overwrites the footer with whitespace.
+          // TODO: Optimize to single init.
+          for (int i = 0; i < previousFooter.length(); i++) {
+            plainLogBuilder.append(previousFooter.charAt(i) == '\n' ? '\n' : ' ');
+          }
+          // plainLogBuilder.append('\n');
```
chanseokoh: Remove?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Runs {@code messageLogger} asynchronously.
+   *
+   * @param messageLogger the {@link Runnable} intended to synchronously log a message to the
+   *     console
+   * @return a {@link Future} to track completion
+   */
+  public Future<Void> log(Runnable messageLogger) {
```
chanseokoh: Trying to understand why this needs to be a `Runnable` instead of a `String`. Isn't it that in the end what this `Runnable` will do is to write some string to `plainLogger`? Actually, what does "plain" mean here? Also, is it that this "message logger" (who will log a message to the console according to the Javadoc) is a different logger than the plain logger? (I didn't get what these two loggers will be in the end.)

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,110 @@
...
+    return log(messageLogger, footer, footerLineCount);
+  }
+
+  /**
+   * Sets the footer asynchronously.
```
chanseokoh: It can also output to the `plainLogger` (whatever that means, erasing the previous output?)

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,110 @@
...
+   *
+   * @param footer the footer
+   * @param lineCount the number of lines in the footer
+   * @return a {@link Future} to track completion
+   */
+  public Future<Void> setFooter(String footer, int lineCount) {
```
chanseokoh: Hmm... is `lineCount` actually the line count for `footer`? When I see the bottom of `log()` implementation, I feel like it is the line count of what `mesageLogger` printed plus `footer`? In any case, if `lineCount` is the line count of `footer`, can't we just count that from `footer` instead of accepting it as a param?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
coollog: Yep, if the footer changed, then that footer would be the new footer and this footer that reached this condition would be the old one.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
coollog: So this is for the user to pass like (Maven) `() -> log.info("log message")` or other logging method, so that this class is independent of the actual different logging levels.

`plain` is for a logging method similar to just plain `System.out.print`, so it will be like `lifecycle` for Gradle and probably `System.out.print` for Maven (might need to do some things differently if wanting to use Maven info logging).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
coollog: We could count it from `footer` but I had left that as a parameter since counting it from the footer might be an unnecessary extra computation especially if the footer gets lengthy, but we could leave it out for simplicity of the API.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,110 @@
...
+          // TODO: Optimize to single init.
+          for (int i = 0; i < previousLineCount; i++) {
+            // Moves cursor up.
+            plainLogBuilder.append(CURSOR_UP_SEQUENCE);
+          }
+
+          // Overwrites the footer with whitespace.
+          // TODO: Optimize to single init.
+          for (int i = 0; i < previousFooter.length(); i++) {
+            plainLogBuilder.append(previousFooter.charAt(i) == '\n' ? '\n' : ' ');
```
briandealwis: Since you're assuming this is outputting to an ANSI-capable terminal, you could consider using the ED (erase in display) to clear the end of screen, or EL (erase in line) to clear the line. It would help if there is other output interleaved on the screen that is unbeknownst to us (e.g., java.util.logging output).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.function.Consumer;
+
+/**
+ * Keeps all log messages in a sequential, deterministic order along with an additional footer that
+ * always appears below log messages.
+ */
+class SingleThreadedLogger {
```
briandealwis: This class name doesn't really describe the class well.  This logger is only appropriate for terminals that support ANSI escape codes.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
coollog: Oh ED would be exactly what we want.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
coollog: Hmm what would you recommend as the name?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
chanseokoh: Ah, now I see the intention. Basically, this class is the `LogEvent` handler (or will be used by whatever `LogEvent` handler) to "move" the footer (the multi-line progress display) down by one line after logging the message (if it gets printed to the screen, for example). This is done by calling this `log()` message. And at the same time, this class is the `ProgressEvent` handler (or will be used by whatever `ProgressEvent` handler`) to update the progress display, which is done by `setFooter()`. I see where my confusion came from: first, the potential mix of `LogEvent` and `ProgressEvent`; secondly, the mix of two output channels, one for `messageLogger` and `plainLogger` (for example, one going out to a file and the other going out to a screen); and the `messageLogger` is basically an unrestricted lambda. Not that I am suggesting something concrete here... don't really have an idea.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
chanseokoh: Generally speaking, from the API perspective, removing the unnecessary param eliminates the possibility of misuse, so I think it's better without `lineCount` in the general sense. Even if a footer length is say, 3000, counting from 0 to 3000 even on an old machine should be a blink of an eye. But I also value code simplicity and don't want to introduce more code. I see the tradeoff.

If the caller knows the line count ahead, I wonder if the call can provide `List<String>` from the first place (a `String` element per line). Otherwise, I think I am fine with `lineCounter.`

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
coollog: Yea, this will be used by the `LogEventHandler`. A `ProgressLogger` will be the one that handles `ProgressEvent`s and directly calls `setFooter` on this or set the footer via a `LogEvent` type of `CHANGE_FOOTER` or something.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
coollog: Sounds good, taking the `List<String>` seems like the best way to do it then.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
briandealwis: `SureFootedLogger` :-)

More seriously, the key here is that the logger uses ANSI escapes to maintain a footer element. `AnsiFooterLogger`?

coollog: Hmm, this is more of a logger than logs messages along with a footer that is always at the bottom and `AnsiFooterLogger` seems to point to it being just logging a footer - how about `AnsiLoggerWithFooter`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
```diff
@@ -0,0 +1,101 @@
...
+   */
+  public Future<Void> setFooter(List<String> footerLines) {
+    if (footerLines.equals(this.footerLines)) {
+      return Futures.immediateFuture(null);
+    }
+
+    int previousLineCount = this.footerLines.size();
+    this.footerLines = footerLines;
+
+    return log(() -> {}, previousLineCount);
```
chanseokoh: Do we assume that this class is not used concurrently (i.e., not calling these methods concurrently in different threads)? Otherwise, I think I see some issues.

coollog: Ah right, the previous line count could be different by the time this is called. Lemme fix this.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLoggerTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.junit.Assert;
+import org.junit.Test;
+
+/** Tests for {@link SingleThreadedLogger}. */
+public class SingleThreadedLoggerTest {
```
chanseokoh: Don't forget to change the name and the Javadoc when you rename the class.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SingleThreadedLogger.java
chanseokoh: I think `AnsiLoggerWithFooter` works.

I'd just like to say in the Javadoc that the class is intended for the case where both of the two log destination channels are the same console. (For example, it isn't really for `messageLogger` to log to a file or a database, while the footer is going out to stdout.)

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
```diff
@@ -0,0 +1,137 @@
...
+
+  AnsiLoggerWithFooter(Consumer<String> plainLogger) {
+    this.plainLogger = plainLogger;
+  }
+
+  /** Shuts down the {@link #executorService}. */
+  public void shutDown() {
+    executorService.shutdown();
+    try {
+      if (!executorService.awaitTermination(
```
chanseokoh: I think it's better to have `shutdown()` just call `executorService.shutdown()` and not block. I think in many cases, the client of this class don't want to wait after calling `shutdown()` but just proceed. If you want to give the client an option to be able to wait termination, I'd just create another method named `awaitTermination()`.

PR #1330, opened 2018-12-10
 Adds ProgressEventDispatcher to replace manual Allocation management.

coollog:
Part of #1297 

This gets rid of all the manual calls to dispatch `ProgressEvent`s in the builder steps.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
@@ -25,7 +25,7 @@ import java.io.Closeable;
 
 /**
  * Dispatches {@link ProgressEvent}s associated with a managed {@link Allocation}. Keeps track of
- * the allocation units that are remaining so that it can emits the remaining progress units upon
+ * the allocation units that are remaining so that it can emit the remaining progress units upon
  * {@link #close}.
  *
  * <p>This class is <em>not</em> thread-safe. Only use a single instance per thread and create child
@@ -35,9 +35,7 @@ public class ProgressEventDispatcher implements Closeable {
 
   /**
    * Creates a new {@link ProgressEventDispatcher} based off an existing {@link
-   * ProgressEventDispatcher}.
-   *
-   * <p>Implementations should be thread-safe.
+   * ProgressEventDispatcher}. {@link #create} should only be called once.
    */
   @FunctionalInterface
   public interface Factory {
@@ -101,9 +99,19 @@ public class ProgressEventDispatcher implements Closeable {
    */
   public Factory newChildProducer() {
     decrementRemainingAllocationUnits(1);
-    return (description, allocationUnits) ->
-        newProgressEventDispatcher(
+
+    return new Factory() {
+
+      private boolean used = false;
+
+      @Override
+      public ProgressEventDispatcher create(String description, long allocationUnits) {
+        Preconditions.checkState(!used);
+        used = true;
+        return newProgressEventDispatcher(
             eventDispatcher, allocation.newChild(description, allocationUnits));
+      }
+    };
   }
 
   /** Emits the remaining allocation units as progress units in a {@link ProgressEvent}. */
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
@@ -24,24 +24,41 @@ import com.google.common.base.Verify;
 import java.io.Closeable;
 
 /**
- * Emits the remaining progress units upon {@link #close}.
+ * Dispatches {@link ProgressEvent}s associated with a managed {@link Allocation}. Keeps track of
+ * the allocation units that are remaining so that it can emits the remaining progress units upon
+ * {@link #close}.
  *
  * <p>This class is <em>not</em> thread-safe. Only use a single instance per thread and create child
  * instances with {@link #newChildProducer}.
  */
 public class ProgressEventDispatcher implements Closeable {
 
+  /**
+   * Creates a new {@link ProgressEventDispatcher} based off an existing {@link
+   * ProgressEventDispatcher}.
+   *
+   * <p>Implementations should be thread-safe.
+   */
   @FunctionalInterface
   public interface Factory {
 
+    /**
+     * Creates the {@link ProgressEventDispatcher} with an associated {@link Allocation}.
+     *
+     * @param description user-facing description of what the allocation represents
+     * @param allocationUnits number of allocation units
+     * @return the new {@link ProgressEventDispatcher}
+     */
     ProgressEventDispatcher create(String description, long allocationUnits);
   }
 
   /**
-   * @param eventDispatcher
-   * @param description
-   * @param allocationUnits
-   * @return
+   * Creates a new {@link ProgressEventDispatcher} with a root {@link Allocation}.
+   *
+   * @param eventDispatcher the {@link EventDispatcher}
+   * @param description user-facing description of what the allocation represents
+   * @param allocationUnits number of allocation units
+   * @return a new {@link ProgressEventDispatcher}
    */
   public static ProgressEventDispatcher newRoot(
       EventDispatcher eventDispatcher, String description, long allocationUnits) {
@@ -53,9 +70,9 @@ public class ProgressEventDispatcher implements Closeable {
    * Creates a new {@link ProgressEventDispatcher} and dispatches a new {@link ProgressEvent} with
    * progress 0 for {@code allocation}.
    *
-   * @param eventDispatcher
-   * @param allocation
-   * @return
+   * @param eventDispatcher the {@link EventDispatcher}
+   * @param allocation the {@link Allocation} to manage
+   * @return a new {@link ProgressEventDispatcher}
    */
   private static ProgressEventDispatcher newProgressEventDispatcher(
       EventDispatcher eventDispatcher, Allocation allocation) {
@@ -75,30 +92,43 @@ public class ProgressEventDispatcher implements Closeable {
     remainingAllocationUnits = allocation.getAllocationUnits();
   }
 
+  /**
+   * Creates a new {@link Factory} for a {@link ProgressEventDispatcher} that manages a child {@link
+   * Allocation}. Since each child {@link Allocation} accounts for 1 allocation unit of its parent,
+   * this method decrements the {@link #remainingAllocationUnits} by {@code 1}.
+   *
+   * @return a new {@link Factory}
+   */
   public Factory newChildProducer() {
-    remainingAllocationUnits--;
+    decrementRemainingAllocationUnits(1);
     return (description, allocationUnits) ->
         newProgressEventDispatcher(
             eventDispatcher, allocation.newChild(description, allocationUnits));
   }
 
-  public ProgressEventDispatcher dispatchProgress(long progressUnits) {
-    eventDispatcher.dispatch(new ProgressEvent(allocation, progressUnits));
-    remainingAllocationUnits -= progressUnits;
-    Verify.verify(
-        remainingAllocationUnits > 0,
-        "Remaining allocation units less than 0 for '%s': %s",
-        allocation.getDescription(),
-        remainingAllocationUnits);
-    return this;
-  }
-
+  /** Emits the remaining allocation units as progress units in a {@link ProgressEvent}. */
   @Override
   public void close() {
-    Preconditions.checkState(!closed);
-    closed = true;
     if (remainingAllocationUnits > 0) {
       dispatchProgress(remainingAllocationUnits);
     }
+    closed = true;
+  }
+
+  private ProgressEventDispatcher dispatchProgress(long progressUnits) {
+    decrementRemainingAllocationUnits(progressUnits);
+    eventDispatcher.dispatch(new ProgressEvent(allocation, progressUnits));
+    return this;
+  }
+
+  private void decrementRemainingAllocationUnits(long units) {
+    Preconditions.checkState(!closed);
+
+    remainingAllocationUnits -= units;
+    Verify.verify(
+        remainingAllocationUnits >= 0,
+        "Remaining allocation units less than 0 for '%s': %s",
+        allocation.getDescription(),
+        remainingAllocationUnits);
   }
 }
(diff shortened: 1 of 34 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Creates a new {@link Factory} for a {@link ProgressEventDispatcher} that manages a child {@link
+   * Allocation}. Since each child {@link Allocation} accounts for 1 allocation unit of its parent,
+   * this method decrements the {@link #remainingAllocationUnits} by {@code 1}.
+   *
+   * @return a new {@link Factory}
+   */
+  public Factory newChildProducer() {
+    decrementRemainingAllocationUnits(1);
+    return (description, allocationUnits) ->
```
chanseokoh: Question: the name `Factory` implies that, when I have this "factory" instance, I can use it multiple times to create `ProgressEventDispatcher`s. It this allowed? Or, it is intended that the caller should call the "create()" only once? If the latter, can we prevent creating the dispatcher multiple times?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+    decrementRemainingAllocationUnits(progressUnits);
+    eventDispatcher.dispatch(new ProgressEvent(allocation, progressUnits));
+    return this;
+  }
+
+  private void decrementRemainingAllocationUnits(long units) {
+    Preconditions.checkState(!closed);
+
+    remainingAllocationUnits -= units;
+    Verify.verify(
```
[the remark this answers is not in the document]
coollog: Yea the intention is that we would not be able to make a commit that does not pass this check. That actually reminds me that I need to add an integration test for making sure the progress finishes at 100%.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
```diff
@@ -0,0 +1,134 @@
...
+    return (description, allocationUnits) ->
+        newProgressEventDispatcher(
+            eventDispatcher, allocation.newChild(description, allocationUnits));
+  }
+
+  /** Emits the remaining allocation units as progress units in a {@link ProgressEvent}. */
+  @Override
+  public void close() {
+    if (remainingAllocationUnits > 0) {
+      dispatchProgress(remainingAllocationUnits);
```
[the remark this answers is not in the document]
coollog: The progress should never exceed 100% by either manually creating children (and taking up 1 unit each) or by manually dispatching progress units. However, this class is intended to dispatch the rest of the progress units remaining upon close. This way, each of the builder steps do not need to manually dispatch progress events for the remaining portion of their allocations upon finish, and thus guaranteeing 100% completion.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
coollog: Yes, it is intended to be called only once - will ensure this.

coollog: Looks like this caught some mistakes!

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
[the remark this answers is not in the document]
chanseokoh: > the rest of the progress units remaining upon close

I get the intention that you want the auto-dispatch convenience using the `try` block. However, I believe there are only the following two cases (if I have been following the progress framework correctly):
- You have a factory and call create() with some allocation units, say 10, to get `ProgressEventDispatcher`, and you never call `newChildProducer()` at all. That is, you end up calling `dispatchProgress()` with the entire allocation units (10). However, in this case, I see almost all the time the allocation units are 1.
- You have a factory and call create() some some allocation units, say 10. Then you call `newChildProducer()` exactly 10 times.

That is, there is nothing in between where you call `newChildProducer()` 4 times and you end up calling `dispatchProgress()` with 6 units. If what I understood is right, I think it's better to check if the things fall under those two cases: check if you called `newChildProducer()` the number of times equal to the allocation units (`remaningAllocationUnits == 0`) or you never called `newChildProducer()`?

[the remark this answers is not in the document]
coollog: There are times where there can be a mix of the two. For example, in `PullBaseImageStep`, if the base image required authorization, then one of the original 2 allocations would go to a child allocation for `RetrieveRegistryCredentialsStep` and the second would go to pulling the base image.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
```diff
@@ -0,0 +1,142 @@
...
+  public Factory newChildProducer() {
+    decrementRemainingAllocationUnits(1);
+
+    return new Factory() {
+
+      private boolean used = false;
+
+      @Override
+      public ProgressEventDispatcher create(String description, long allocationUnits) {
+        Preconditions.checkState(!used);
```
chanseokoh: Just keep in mind that generally speaking, it's better if restrictions can be enforced at the API level (the API prevents mis-usage from the get-go), so try to go that way as much as possible if achievable.

Anyways, if `create()` should be called once, is it not possible to somehow remove calling `newChildProducer().create()` in a chain? I see you call `someDispatcher.newChildProducer()` for passing factory to another method and the callee calls `create()` there, and just from the technical viewpoint, you can pass `someDispatcher` to the callee (which I think is basically the previous code before this PR). I know the `close()` method matters with `try` now, but isn't it that any callees should complete before going out of the `try` block anyway? Trying to understand why we need to have the factory and then calling `create()`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -239,41 +239,49 @@ public StepsRunner pushImage() {
...
                     Preconditions.checkNotNull(steps.buildImageStep)));
   }
 
   public BuildResult run() throws ExecutionException, InterruptedException {
-    Preconditions.checkNotNull(rootProgressAllocation);
-    stepsRunnable.run();
-    return Preconditions.checkNotNull(steps.finalStep).getFuture().get();
+    Preconditions.checkNotNull(rootProgressAllocationDescription);
+
+    try (ProgressEventDispatcher progressEventDispatcher =
```
chanseokoh: nit: `try (rootProgressEventDispatcher) = ProgressEventDispatcher.newRoot(...)) {`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -100,29 +99,26 @@ public Credential call() throws CredentialRetrievalException {
...
-        if (optionalCredential.isPresent()) {
-          buildConfiguration
-              .getEventDispatcher()
-              .dispatch(new ProgressEvent(progressAllocation, 1));
-          return optionalCredential.get();
+
+    try (ProgressEventDispatcher ignored =
+        progressEventDispatcherFactory.create("Retrieve registry credentials for " + registry, 1)) {
+      try (TimerEventDispatcher ignored2 =
+          new TimerEventDispatcher(buildConfiguration.getEventDispatcher(), description)) {
```
chanseokoh: nit: a single `try` block probably works. (The difference is negligible.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
```diff
@@ -76,34 +75,40 @@
...
   }
 
-  private PushBlobStep makePushBlobStep(AsyncStep<? extends CachedLayer> cachedLayerStep)
+  private PushBlobStep makePushBlobStep(
+      AsyncStep<? extends CachedLayer> cachedLayerStep,
+      ProgressEventDispatcher.Factory progressEventDispatcherFactory)
       throws ExecutionException {
     CachedLayer cachedLayer = NonBlockingSteps.get(cachedLayerStep);
 
-    Allocation progressAllocation =
```
chanseokoh: Double-checking: was it incorrect to add a new child previously?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
```diff
@@ -76,34 +75,40 @@
...
+        // Constructs a PushBlobStep for each layer.
+        ImmutableList.Builder<AsyncStep<PushBlobStep>> pushBlobStepsBuilder =
+            ImmutableList.builder();
+        for (AsyncStep<? extends CachedLayer> cachedLayerStep : cachedLayers) {
+          ProgressEventDispatcher.Factory childProgressEventDispatcherFactory =
+              progressEventDispatcher.newChildProducer();
+          ListenableFuture<PushBlobStep> pushBlobStepFuture =
+              Futures.whenAllSucceed(cachedLayerStep.getFuture())
+                  .call(
+                      () -> makePushBlobStep(cachedLayerStep, childProgressEventDispatcherFactory),
```
chanseokoh: You can embed `progressEventDispatcher.newChildProducer()` here instead of having `childProgressEventDispatcherFactory`? The timing will be different, but if timing matters, but I wonder if the `try` block is doing the right thing, as I said below.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
```diff
@@ -76,34 +75,40 @@
...
+                      listeningExecutorService);
+          pushBlobStepsBuilder.add(() -> pushBlobStepFuture);
+        }
```
chanseokoh: I *think* this `call()` just returns ` list of `AsyncSteps`, which may not have been completed. Does it make sense we emit a completion event at this point when getting out of the `try` block?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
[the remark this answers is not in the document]
chanseokoh: Hmm... got it. I think I'm increasing loosing confidence in using the progress API correctly if I were to set up progress reporting in our codebase.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
coollog: The factory is there to first decrement the `remainingAllocationUnits` and passing that `Factory` to the child `Step`. The child `Step` does not create its own `ProgressEventDispatcher` with the `Factory` until it actually runs since creation of the `ProgressEventDispatcher` means to also send a `ProgressEvent` with 0 progress to add that step as an unfinished step (we don't want to add it before it starts running).

An alternative could be to not have the `Factory` but rather just pass the child `ProgressEventDispatcher`s and have the `try-with-resource` try on a `ProgressEventDispatcher.Manager` that on creation sends a `ProgressEvent` with 0 progress and on close does what `ProgressEventDispatcher#close` does now.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
coollog: `try-with-resource` requires creation of a new local variable.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
coollog: Yea, because a new child should have been created with the allocation units equal to the number of blobs to push, so that each of the `PushBlobStep` takes up a single unit of that new child rather than the parent.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
coollog: The `newChildProducer` is necessary to be outside because it needs to decrement the `remainingAllocationUnits` of `progressEventDispatcher`, whereas the actual `create` call can be done when the `pushBlobStepFuture` actually runs.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayersStep.java
coollog: Getting out the `try` block doesn't emit a "completion event" but rather just sends a `ProgressEvent` with the rest of the `remainingAllocationUnits`. In this case, it would not send anything because `remainingAllocationUnits` is 0 (all taken up by the child producers).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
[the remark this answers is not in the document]
coollog: Hmm, I'll add a task to potentially make this API clearer to #1297

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
coollog: The `dispatchProgress(0)` is called in `newProgressEventDispatcher` on line 77.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcherTest.java
```diff
@@ -0,0 +1,78 @@
...
+    Assert.assertEquals(0, progressEvents.get(0).getUnits());
+    Assert.assertEquals(0, progressEvents.get(1).getUnits());
+    Assert.assertEquals(20, progressEvents.get(2).getUnits());
+    Assert.assertEquals(9, progressEvents.get(3).getUnits());
+  }
+
+  @Test
+  public void testDispatch_tooManyChildren() {
+    try {
+      try (ProgressEventDispatcher progressEventDispatcher =
```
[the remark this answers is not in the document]
coollog: Oh yea that is true.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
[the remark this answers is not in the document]
coollog: Oops forgot to push this change.

PR #1331, opened 2018-12-10
Ensure internally-created executor service is shutdown on completion

### LINKED ISSUE #1310: jib-core: shut down executor service to allow JVM termination
Was testing jib-core.

```java
  public static void main(String[] args) throws Exception {
    Containerizer containerizer = Containerizer.to(DockerDaemonImage.named("francium25/test"));
    Jib.from("francium25/base").containerize(containerizer);
  }
```

The code does not terminate after the image is built, because we are not shutting down `StepsRunner.listeningExecutorService`. One potential-but-not-really-right way to handle it would be to shut it down in `waitOn...Step()`, which is the last step of the `Steps` we chain.

From the perspective of the above simple code, it makes sense to shut it down. However, it's possible to accept an executor service from the user via `Containerizer.setExecutorService()`, and we shouldn't shut their service down on our discretion at the end. But we could accept another parameter for `setExecutorService()` to tell us if we will shut it down or not.

Also, another consequence of shutting down the service is that, in the usual case where people don't provide their executor service, they cannot reuse `JibContainerBuilder`.
```java
    JibContainerBuilder cb = Jib.from(...);
    cb.containerize(...);
    cb.containerize(...); // exception because the executor service is shut down.
```
This is because `BuildConfiguration` is owned by `JibContainerBuilder`, hence the two `containerize()` calls share `executorService`, which `BuildConfiguration` owns. And `executorService` is a mutable object.

And the problem is that, as long as we allow reusing `JibContainerBuilder`, it is impossible to know when it is OK to shut it down.

briandealwis:
Ensure that internally-created `ExecutorService` is shutdown.

Fixes #1310

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -426,13 +426,13 @@ public class JibContainerBuilder {
 
   @VisibleForTesting
   JibContainer containerize(
-      Containerizer containerizer, Supplier<ExecutorService> executorServiceFactory)
+      Containerizer containerizer, Supplier<ExecutorService> defaultExecutorServiceFactory)
       throws InterruptedException, ExecutionException, IOException,
           CacheDirectoryCreationException {
 
     boolean shutdownExecutorService = !containerizer.getExecutorService().isPresent();
     ExecutorService executorService =
-        containerizer.getExecutorService().orElseGet(executorServiceFactory);
+        containerizer.getExecutorService().orElseGet(defaultExecutorServiceFactory);
 
     BuildConfiguration buildConfiguration = toBuildConfiguration(containerizer, executorService);
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -61,9 +61,6 @@ import javax.annotation.Nullable;
 // TODO: Add tests once containerize() is added.
 public class JibContainerBuilder {
 
-  @VisibleForTesting
-  static Supplier<ExecutorService> executorServiceFactory = Executors::newCachedThreadPool;
-
   private final ContainerConfiguration.Builder containerConfigurationBuilder =
       ContainerConfiguration.builder();
   private final BuildConfiguration.Builder buildConfigurationBuilder;
@@ -421,6 +418,15 @@ public class JibContainerBuilder {
    * @throws IOException if an I/O exception occurs
    */
   public JibContainer containerize(Containerizer containerizer)
+      throws InterruptedException, ExecutionException, IOException,
+          CacheDirectoryCreationException {
+
+    return containerize(containerizer, Executors::newCachedThreadPool);
+  }
+
+  @VisibleForTesting
+  JibContainer containerize(
+      Containerizer containerizer, Supplier<ExecutorService> executorServiceFactory)
       throws InterruptedException, ExecutionException, IOException,
           CacheDirectoryCreationException {
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -444,6 +444,8 @@ public class JibContainerBuilder {
    * Builds a {@link BuildConfiguration} using this and a {@link Containerizer}.
    *
    * @param containerizer the {@link Containerizer}
+   * @param executorService the {@link ExecutorService} to use, overriding the executor in the
+   *     {@link Containerizer}
    * @return the {@link BuildConfiguration}
    * @throws CacheDirectoryCreationException if a cache directory could not be created
    * @throws IOException if an I/O exception occurs
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -37,6 +37,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.function.Supplier;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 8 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -441,9 +462,8 @@ BuildConfiguration toBuildConfiguration(Containerizer containerizer)
         .setContainerConfiguration(containerConfigurationBuilder.build())
         .setLayerConfigurations(layerConfigurations)
         .setAllowInsecureRegistries(containerizer.getAllowInsecureRegistries())
-        .setToolName(containerizer.getToolName());
-
-    containerizer.getExecutorService().ifPresent(buildConfigurationBuilder::setExecutorService);
+        .setToolName(containerizer.getToolName())
+        .setExecutorService(executorService);
```
chanseokoh: Should we remove the code calling `newCachedThreadPool()` in `BuildConfiguration`, and return an error if the executor is not set? (`BuildConfiguration` is not public-facing.)

```
          if (executorService == null) {
            executorService = Executors.newCachedThreadPool();
          }
```

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -58,6 +61,9 @@
 // TODO: Add tests once containerize() is added.
 public class JibContainerBuilder {
 
+  @VisibleForTesting
+  static Supplier<ExecutorService> executorServiceFactory = Executors::newCachedThreadPool;
```
chanseokoh: Personally, I try avoiding swapping out a field for testing particularly for a static field (or it is just me thinking this way?), but I guess there is no other better way.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
briandealwis: Oops, that had been removed, but I reworked the patch and it was restored.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
briandealwis: Just realized a better approach.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -417,21 +420,45 @@ public JibContainerBuilder setWorkingDirectory(@Nullable AbsoluteUnixPath workin
...
-    BuildConfiguration buildConfiguration = toBuildConfiguration(containerizer);
-    BuildResult result = containerizer.getTargetImage().toBuildSteps(buildConfiguration).run();
-    return new JibContainer(result.getImageDigest(), result.getImageId());
+
+    return containerize(containerizer, Executors::newCachedThreadPool);
+  }
+
+  @VisibleForTesting
+  JibContainer containerize(
+      Containerizer containerizer, Supplier<ExecutorService> executorServiceFactory)
```
loosebazooka: Would consider a different name for this `executorServiceFactory` since it's just a backup/default.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -417,21 +420,45 @@ public JibContainerBuilder setWorkingDirectory(@Nullable AbsoluteUnixPath workin
...
+    ExecutorService executorService =
+        containerizer.getExecutorService().orElseGet(executorServiceFactory);
+
+    BuildConfiguration buildConfiguration = toBuildConfiguration(containerizer, executorService);
+
+    try {
+      BuildResult result = containerizer.getTargetImage().toBuildSteps(buildConfiguration).run();
+      return new JibContainer(result.getImageDigest(), result.getImageId());
+    } finally {
+      if (shutdownExecutorService) {
```
loosebazooka: Could we control the lifecycle with an `Consumer<ExecutorService> cleanup` option?

I'm having a hard time visualizing it though, so maybe we shouldn't.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
loosebazooka: maybe the method needs to be renamed to indicate the `Supplier` is an or else kind of situation?

jib-core/src/test/java/com/google/cloud/tools/jib/api/JibContainerBuilderTest.java
```diff
@@ -141,7 +156,9 @@ public void testToBuildConfiguration()
     JibContainerBuilder jibContainerBuilder =
         new JibContainerBuilder(baseImage, spyBuildConfigurationBuilder)
             .setLayers(Arrays.asList(mockLayerConfiguration1, mockLayerConfiguration2));
-    BuildConfiguration buildConfiguration = jibContainerBuilder.toBuildConfiguration(containerizer);
+    BuildConfiguration buildConfiguration =
+        jibContainerBuilder.toBuildConfiguration(
+            containerizer, containerizer.getExecutorService().get());
```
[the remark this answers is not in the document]
briandealwis: The test is checking that the set values are actually propagated into the `BuildConfiguration`.  Using `Optional.get()` here will thrown an exception if the `Containerizer` wasn't configured with an `ExecutorService`, and also ensures that the resulting `BuilkdConfiguration` has the configured `ExecutorService`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
chanseokoh: I think `try ... finally` works. (I'm assuming the `run()` is blocking, which I believe is true.)

briandealwis: The `run()` is somewhat-blocking, in that we [wait on the future from the final step](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java#L276), but `shutdown()` triggers an orderly shutdown that will continue any pending work until completion.

chanseokoh: > somewhat-blocking,

I'd say it's sure-blocking. :)

PR #1333, opened 2018-12-10
Adds ProgressDisplayGenerator.

coollog:
Part of #1297 

`ProgressDisplayGenerator` generates the displayed text for progress feedback.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ProgressDisplayGenerator.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ProgressDisplayGenerator.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.common;
+
+import com.google.cloud.tools.jib.event.progress.Allocation;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+/**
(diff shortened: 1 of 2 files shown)


jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ProgressDisplayGeneratorTest.java
```diff
@@ -0,0 +1,71 @@
...
+    Allocation childRight = root.newChild("childRight", 2);
+
+    Assert.assertEquals(
+        Arrays.asList(
+            "Executing tasks:",
+            "[=========================                         ] 50.0% complete",
+            "> childLeftDown",
+            "> childRight"),
+        ProgressDisplayGenerator.generateProgressDisplay(
+            0.5, Arrays.asList(root, childLeft, childLeftDown, childRight)));
```
chanseokoh: Hmm... looks like a caller enumerate all unfinished allocations. So I think the caller will go over every node in the tree and identify which node is unfinished. Knowing that, is it not easy for the caller to identify leaves and pass only leaves?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ProgressDisplayGenerator.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Generates a display of progress and unfinished tasks.
+ *
+ * <p>Example:
+ *
+ * <p>Executing tasks... [================= ] 72.5% complete > task 1 running > task 3 running
```
chanseokoh: The HEADER, the bar, and the two tasks appear in a single line here. (Not that it cannot be.)

coollog: Oh oops the formatting got messed up

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ProgressDisplayGeneratorTest.java
coollog: The caller would be the one that gets the unfinished allocations from `ProgressEventHandler#getUnfinishedAllocations`, which gets it from `AllocationCompletionTracker`. I believe we could have `AllocationCompletionTracker` just give back the leaf allocations so that the set computation can be optimized away. I'll make a TODO for this optimization.

chanseokoh: Yeah, I think that'd be nice.

PR #1336, opened 2018-12-12
Adds SuccessEvent.

coollog:
Part of #1297 

This is useful to know when to shutdown `AnsiLoggerWithFooter`.


coollog:
Actually, might not be necessary, closing for now.

PR #1340, opened 2018-12-12
 Improves the logic of AnsiLoggerWithFooter.

coollog:
Part of #1297 

To be more compatible with logging methods with newline at end.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
@@ -58,7 +58,8 @@ class AnsiLoggerWithFooter {
   /**
    * Creates a new {@link AnsiLoggerWithFooter}
    *
-   * @param plainPrinter the {@link Consumer} intended to synchronously print the footer and other plain console output. {@code plainPrinter} should print a new line at the end.
+   * @param plainPrinter the {@link Consumer} intended to synchronously print the footer and other
+   *     plain console output. {@code plainPrinter} should print a new line at the end.
    */
   AnsiLoggerWithFooter(Consumer<String> plainPrinter) {
     this.plainPrinter = plainPrinter;
@@ -102,7 +103,7 @@ class AnsiLoggerWithFooter {
           for (String footerLine : footerLines) {
             plainPrinter.accept(BOLD + footerLine + UNBOLD);
           }
-          
+
           return null;
         });
   }
@@ -129,7 +130,7 @@ class AnsiLoggerWithFooter {
           String newFooterPrefix = didErase ? CURSOR_UP_SEQUENCE : "";
 
           for (String newFooterLine : newFooterLines) {
-            plainPrinter.accept(newFooterPrefix+BOLD+newFooterLine+UNBOLD);
+            plainPrinter.accept(newFooterPrefix + BOLD + newFooterLine + UNBOLD);
             newFooterPrefix = "";
           }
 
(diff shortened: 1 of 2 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
```diff
@@ -93,49 +114,57 @@ public void awaitTermination() {
...
-            StringBuilder footerBuilder = new StringBuilder();
-            for (String newFooterLine : newFooterLines) {
-              footerBuilder.append(BOLD).append(newFooterLine).append(UNBOLD).append('\n');
-            }
-            // Removes last newline.
-            footerBuilder.setLength(footerBuilder.length() - 1);
-            plainLogger.accept(footerBuilder.toString());
+          for (String newFooterLine : newFooterLines) {
+            plainPrinter.accept(newFooterPrefix + BOLD + newFooterLine + UNBOLD);
+            newFooterPrefix = "";
```
chanseokoh: Hmm... how unfortunate. What do you think of

```java
plainPrinter.accept(TWO_TIMES_CURSOR_UP_SEQUENCE)?
for (...) {
  plainPrinter.accept(BOLD + newFooterLine + UNBOLD);
}
```
?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
```diff
@@ -79,12 +86,26 @@ public void awaitTermination() {
...
-  public Future<Void> log(Runnable messageLogger) {
-    return log(messageLogger, footerLines);
+  public Future<Void> log(Consumer<String> messageLogger, String message) {
+    return executorService.submit(
+        () -> {
+          boolean didErase = eraseFooter();
+
+          // If a previous footer was erased, the message needs to go up a line.
+          String messagePrefix = didErase ? CURSOR_UP_SEQUENCE : "";
+          messageLogger.accept(messagePrefix + message);
```
chanseokoh: Will this work when `messageLogger` itself is prepending, e.g., "[INFO] "? For example, I wonder if the end string might be "[INFO] + <cursor up> + message", in which case you leave "[INFO]" at the bottom and the actual message line loses "[INFO]".

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
coollog: Yea that would be ideal, though somehow on Gradle, this sometimes causes a prior-logged line to disappear. Not too sure why that happens though...

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
coollog: Yep, it works since cursor up goes up in the same column. However, it does replace other message lines' prepended text like `[DEBUG]` with `[INFO]`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
chanseokoh: Ah, OK. Moving one line seems safter.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/AnsiLoggerWithFooter.java
chanseokoh: I see. I think we can tolerate that.

PR #1341, opened 2018-12-12
Adds AnsiLoggerWithFooter to Maven/Gradle plugins.

coollog:
Part of #1297 

Wraps the original logging mechanisms.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -146,6 +146,8 @@ public class BuildDockerTask extends DefaultTask implements JibTask {
               projectProperties.getJavaLayerConfigurations().getLayerConfigurations(),
               helpfulSuggestions);
 
+      projectProperties.waitForLoggingThread();
+
     } catch (InvalidAppRootException ex) {
       throw new GradleException(
           "container.appRoot is not an absolute Unix-style path: " + ex.getInvalidPathValue(), ex);
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -146,6 +146,7 @@ public class BuildDockerTask extends DefaultTask implements JibTask {
               projectProperties.getJavaLayerConfigurations().getLayerConfigurations(),
               helpfulSuggestions);
 
+      // TODO: This should not be called on projectProperties.
       projectProperties.waitForLoggingThread();
 
     } catch (InvalidAppRootException ex) {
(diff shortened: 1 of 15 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -146,6 +146,8 @@ public void buildDocker()
               projectProperties.getJavaLayerConfigurations().getLayerConfigurations(),
               helpfulSuggestions);
 
+      projectProperties.waitForLoggingThread();
```
loosebazooka: It seems like we might be doing too much in the class. It's not longer supplier of `properties`. It does seem sort of weird that we would call `waitForLoggingThread()` on it.

This is an interface that appears it is feeding some data back to jib-core under normal circumstances? But `waitForLoggingThread` is an action that appears to have nothing to do with jib-core?

loosebazooka: Can this be a part of the `LogEventHandler` instead?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -144,6 +144,9 @@ public void buildTar()
               projectProperties.getJavaLayerConfigurations().getLayerConfigurations(),
               helpfulSuggestions);
 
+      // TODO: This should not be called on projectProperties.
+      projectProperties.waitForLoggingThread();
```
chanseokoh: Is it worth waiting for the logging thread to complete? If it can take some time and we do wait for it, we add that amount of time to the build. Is it because finishing the log with an incomplete percentage may look weird?

But honestly, I do believe the wait time will be close to 0, so I'm actually fine with this. Just asking if you felt the strong need.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/LogEventHandler.java
```diff
@@ -17,40 +17,53 @@
...
-        log.error(logEvent.getMessage());
+        if (log.isErrorEnabled()) {
+          ansiLoggerWithFooter.log(log::error, logEvent.getMessage());
+        }
         break;
 
       case INFO:
         // Use lifecycle for progress-indicating messages.
-        log.debug(logEvent.getMessage());
+        if (log.isDebugEnabled()) {
```
chanseokoh: Huh... interesting comment that we send INFO to debug. Should it really be the case?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
coollog: When not waiting, last few lines get mixed up with whatever comes next in the build system build since the thread can keep running even after Jib finishes.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/LogEventHandler.java
coollog: Yeah, the `INFO` here is at a higher verbosity than `LIFECYCLE`, and in Maven, the lowest verbosity is `.info`, so the next verbosity would be `.debug`.

PR #1342, opened 2018-12-12
Suppresses warning about ignored Future in FinalizingStep.

coollog:
This is fine because we don't care about exceptions from a logging future.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/FinalizingStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/FinalizingStep.java
@@ -27,6 +27,7 @@ import com.google.common.util.concurrent.ListeningExecutorService;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 
 /** Logs the message before finalizing an image build. */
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/FinalizingStep.java
```diff
@@ -74,12 +75,13 @@ public Void call() throws ExecutionException {
       asyncDependencies.addSteps(NonBlockingSteps.get(wrappedDependency));
     }
 
-    // TODO: Don't let future error be suppressed
-    asyncDependencies.whenAllSucceed(
-        () -> {
-          buildConfiguration.getEventDispatcher().dispatch(LogEvent.lifecycle("Finalizing..."));
-          return null;
-        });
+    // This suppresses any exceptions of this future.
```
chanseokoh: You mean "suppresses compile warning"? This reads to me like, by having this, image building exceptions when running the future will not be reported at run time.

coollog: This is in fact suppressing the exceptions that could happen in this future, but I think it's okay since it's just an extraneous logging message.

chanseokoh: Okay, just thought the comment does not match the PR title.

PR #1343, opened 2018-12-12
Changes ProgressEventHandler to provide the callback with updates.

coollog:
Part of #1297

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
@@ -62,6 +62,24 @@ import org.slf4j.LoggerFactory;
 /** Integration tests for {@link BuildSteps}. */
 public class BuildStepsIntegrationTest {
 
+  private static class ProgressChecker {
+
+    private final ProgressEventHandler progressEventHandler =
+        new ProgressEventHandler(
+            update -> {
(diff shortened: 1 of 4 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/BuildStepsIntegrationTest.java
```diff
@@ -62,6 +62,27 @@
...
 
+  /**
+   * Helper class to hold a {@link ProgressEventHandler} and verify that it handles a full progress.
+   */
+  private static class ProgressChecker {
+
+    private final ProgressEventHandler progressEventHandler =
+        new ProgressEventHandler(
+            update -> {
+              lastProgress = update.getProgress();
```
chanseokoh: Hmm... I think the code of `ProgressEventHandler.accept()` (and hence this updater too) runs in the thread that emits the progress event, right? If so, theoretically the order of overwriting `lastProgress` is not deterministic, meaning that it's possible that the 1.0-sum-progress assignment may happen before another less-than-1.0-sum-progress assignment? (But I guess it may never happen in practice if systematically the thread that does the 1.0-sum-progress assignment is always the last thread to run after all the other threads have been completed?)

coollog: Yea, the thread that gets to 1.0 and has no tasks unfinished is always the last update, no matter what order those updates come in.

PR #1344, opened 2018-12-12
 Disables progress footer by default or if console is null.

coollog:
Part of #1297 

This hides enabling footer under system property `jib.showProgress`, which will be changed to `true` by default.

UPDATE:
- Added `LogEventHandlerBuilder` to build a `Consumer<LogEvent>` and removed the two plugin-specific `LogEventHandler`s.
- Separated out `PlainConsoleLogger` from `AnsiLoggerWithFooter` to be the logger with no footer.


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -113,7 +115,14 @@ static Path getExplodedWarDirectory(Project project) {
     this.project = project;
     this.javaLayerConfigurations = javaLayerConfigurations;
 
-    ansiLoggerWithFooter = new AnsiLoggerWithFooter(logger::lifecycle);
+    // TODO: Make SHOW_PROGRESS be true by default.
```
TadCordle: Maybe we could make it `HIDE_PROGRESS` so it can be false by default?

coollog: This system property is currently for hiding the feature behind it for testing before actually enabling it for release, so we can change to `HIDE_PROGRESS` when ready.

briandealwis: It would be worth pulling the enablement into a method. You'll want to check that either we're on Windows (CMD windows support ANSI), or `TERM` matches `xterm.**` or `vt[0-9]*`, or at the very worst `TERM` is not null and `TERM != 'dumb'`.

coollog: Sounds good - moving to separate method and adding this check.

loosebazooka:
Should we also handle `--console=plain` on gradle? Not sure what our behavior is.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/LogEventHandler.java
```diff
@@ -41,6 +41,13 @@ public void accept(LogEvent logEvent) {
         }
         break;
 
+      case PROGRESS:
+        // Progress messages only print if not printing progress in the footer.
```
loosebazooka: I think this might be hard to follow in the future if we haven't been looking at it. There's a lot of branching in a lot of places. 

Is it possible to structure this in a way that we just instantiate the right kind of progress object? (the cool one, and the boring one) or whatever?

Each of them just handles the progress correctly?

coollog:
Currently, for `--console=plain` the behavior is to not show the progress footer and just do the progress messages as before.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/LogEventHandler.java
coollog: Yep, I'll try splitting this class into two classes `RichLogEventHandler` and `PlainLogEventHandler` (or some other better naming).

chanseokoh:
Haven't looked into the code, but I think the progress report should be disabled when the log level is not the default. For example, `mvn -X`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
chanseokoh: If the user is redirecting all stdin and stdout to a file or the underlying logging system is configured to send output to a file, the `TERM` env variable will be meaningless, but I guess we won't worry about that too much.

coollog: Yea, not sure how we could detect if stdout is redirected to a file in Java without calling `isatty`.

coollog:
@chanseokoh I'm not sure if we need to disable footer on debug since, for example, with Gradle, even when `--debug`, the Gradle progress bar still shows.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
briandealwis: Maven also supports a _batch_ mode (`-B`)

coollog: I think we can ignore batch mode since that only affects stdin.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -94,8 +112,22 @@ private static EventHandlers makeEventHandlers(
         .add(JibEventType.TIMING, timerEventHandler);
   }
 
+  private static boolean isProgressFooterEnabled() {
+    // TODO: Make SHOW_PROGRESS be true by default.
```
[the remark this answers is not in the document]
coollog: Yea, the intention is to return false when `SHOW_PROGRESS` is not set.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/SingleThreadedExecutor.java
```diff
@@ -0,0 +1,64 @@
...
+  public void shutDownAndAwaitTermination() {
+    executorService.shutdown();
+
+    try {
+      if (!executorService.awaitTermination(
+          EXECUTOR_SHUTDOWN_WAIT.getSeconds(), TimeUnit.SECONDS)) {
+        executorService.shutdownNow();
+        if (!executorService.awaitTermination(
+            EXECUTOR_SHUTDOWN_WAIT.getSeconds(), TimeUnit.SECONDS)) {
+          throw new RuntimeException("Could not shut down AnsiLoggerWithFooter executor");
```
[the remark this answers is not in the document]
coollog: Hmm, yeah, I think we can just do exactly as recommended in `ExecutorService` and just print a message instead.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -84,8 +88,22 @@ static MavenProjectProperties getForProject(
...
-    LogEventHandler logEventHandler = new LogEventHandler(log, ansiLoggerWithFooter);
+      Log log, SingleThreadedExecutor singleThreadedExecutor) {
+    Consumer<String> noOp = ignored -> {};
+
+    Consumer<LogEvent> logEventHandler =
+        (isProgressFooterEnabled()
+                ? LogEventHandlerBuilder.rich(singleThreadedExecutor).progress(noOp)
+                : LogEventHandlerBuilder.plain(singleThreadedExecutor)
+                    .progress(log.isInfoEnabled() ? log::info : noOp))
+            .lifecycle(log.isInfoEnabled() ? log::info : noOp)
```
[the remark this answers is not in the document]
coollog: Do you mean to just have one method `LogEventHandlerBuilder#addHandler(Level, Consumer<String>)` rather than the 6 separate ones we have now?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
coollog: I tried running Jib with `-B` and although Maven's own colors were disabled, Jib's colors were not.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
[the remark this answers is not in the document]
loosebazooka: I mean sure if it's possible. But now I realize the builder doesn't have access to the logger implementation in the client, so maybe this is the right way?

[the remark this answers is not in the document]
coollog: Yea the logger implementation is specific to the build system.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -102,8 +95,56 @@ static Path getExplodedWarDirectory(Project project) {
...
   }
 
+  private static EventHandlers makeEventHandlers(
+      Project project, Logger logger, SingleThreadedExecutor singleThreadedExecutor) {
+    LogEventHandlerBuilder logEventHandlerBuilder =
+        (isProgressFooterEnabled(project)
+                ? LogEventHandlerBuilder.rich(singleThreadedExecutor)
+                : LogEventHandlerBuilder.plain(singleThreadedExecutor).progress(logger::lifecycle))
+            .lifecycle(logger::lifecycle);
+    if (logger.isDebugEnabled()) {
```
loosebazooka: Sorry I'm stuck on this. And I don't even know if we should handle this. So stop me if I should be stopped.
1. The logger implementations are already filtering the log out based on the log level. So should we do additional filtering here? Presumably what we're doing here can lead to one less call for every log message that needs to be ignored.
2. If someone dynamically changes the log level to a different level, we're kind of stuck in this implementations.

Would it make sense to just pass the supplier in unconditionally and let the log systems's mechanisms handle the filtering according to level?
```
.debug(logger::debug) // also this double debug thing is confusing
.debug(logger::info)
.warn...
.build()
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -102,8 +95,56 @@ static Path getExplodedWarDirectory(Project project) {
...
+      logEventHandlerBuilder.debug(logger::debug);
+    }
+    if (logger.isInfoEnabled()) {
+      logEventHandlerBuilder.debug(logger::info);
```
loosebazooka: should this have been `.info(logger::debug)` ?

coollog: Oh oops, this should have been `.info(logger::info)`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
coollog: Yes, the Gradle `Logger` implementation will filter calls to the individual `Logger` methods based on the verbosity chosen by the user, but this API is on the Jib Core side. Each `LogEventHandlerBuilder` method is attaching a `Logger` method to the corresponding `LogEvent.Level` that should forward to the `Logger` method. `Logger` then can do its logging based on the user choice.

The double debug was a typo - will fix.

loosebazooka: Ah yea right, that makes sense.

loosebazooka: oh yeah okay, I didn't realize it had that effect.

PR #1347, opened 2018-12-14
Add JavaContainerBuilder to jib core

### LINKED ISSUE #1212: Add JavaContainerBuilder as described in Jib Core proposal.
From https://github.com/GoogleContainerTools/jib/blob/master/proposals/archives/jib_core_library.md#for-java-containers

TadCordle:
Fixes #1212.

Adds a `JavaContainerBuilder` class that can be used to build Java-specific containers. Example usage:

```java
JavaContainerBuilder.fromDistroless()
  .addClasses(Paths.get("application/classes"))
  .addResources(Paths.get("application/resources"))
  .addDependencies(
      Paths.get("application/dependencies/dependency.jar"),
      Paths.get("application/dependencies/dependency2.jar"),
      Paths.get("application/dependencies/dependency3-SNAPSHOT.jar)) // Separates snapshots to their own layer
  .addToClasspath(Paths.get("extra/classpath/dir")) // Adds to extra layer, but also adds new directory to classpath
  .setJvmFlags("-Xms512m", "-Xdebug")
  .setMainClass("HelloWorld")
  .toContainerBuilder()
  .containerize(...);
```

Potential followup issues:
- Add a `WarContainerBuilder` for WAR-specific case, or adjust `JavaContainerBuilder` to support WARs
- Use `JavaContainerBuilder` in `Gradle`/`MavenLayerConfigurations`


jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+   * @param dependencyFiles the list of dependencies to add to the image
+   * @return this
+   * @throws IOException if adding the layer fails
+   */
+  public JavaContainerBuilder addDependencies(Path... dependencyFiles) throws IOException {
+    return addDependencies(Arrays.asList(dependencyFiles));
+  }
+
+  /**
+   * Adds resources to /app/resources on the image.
```
coollog: `s/Adds resources/Adds the contents of a resource directory/`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+    layerConfigurationsBuilder.addDirectoryContents(
+        LayerType.CLASSES, classFilesDirectory, path -> true, CLASSES_PATH);
+    if (!classpath.contains(CLASSES_PATH.toString())) {
+      classpath.add(CLASSES_PATH.toString());
+    }
+    return this;
+  }
+
+  /**
+   * Adds additional files to the image's classpath.
```
coollog: I think we can remove `image's` since the classpath is for the java invocation and not the image itself.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+   *
+   * @param otherFiles the list of files to add. Files are added to /app/other on the container
```
coollog: on the container _file system_

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+    return addToClasspath(Arrays.asList(otherFiles));
+  }
+
+  /**
+   * Sets the JVM flags to use when starting the application.
+   *
+   * @param jvmFlags the list of JVM flags
+   * @return this
+   */
+  public JavaContainerBuilder setJvmFlags(List<String> jvmFlags) {
```
coollog: Should we change these to `addJvmFlags` instead?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+   * @param jvmFlags the list of JVM flags
+   * @return this
+   */
+  public JavaContainerBuilder setJvmFlags(String... jvmFlags) {
+    this.jvmFlags = ImmutableList.copyOf(jvmFlags);
+    return this;
+  }
+
+  /**
+   * Sets the main class used to start the application on the image.
```
coollog: I think we can expose `MainClassFinder` to users of Jib Core and point them to use that if they want to have an automatic way of finding the main class among a set of class files.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,291 @@
...
+      } else {
+        layerConfigurationsBuilder.addFile(
+            file.getFileName().toString().contains("SNAPSHOT")
+                ? LayerType.SNAPSHOT_DEPENDENCIES
+                : LayerType.DEPENDENCIES,
+            file,
+            DEPENDENCIES_PATH.resolve(file.getFileName()));
+      }
+    }
+    if (!classpath.contains(DEPENDENCIES_PATH.resolve("*").toString())) {
```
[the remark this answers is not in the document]
TadCordle: Hmm, do we want to maintain the order the user sets the classpath in?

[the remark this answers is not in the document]
coollog: We probably would want to maintain a defined order that doesn't depend on the order in which the user calls the builder methods.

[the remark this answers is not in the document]
loosebazooka: Could we use TreeSet? We'd probably have to implement compareTo or some kind of comparable.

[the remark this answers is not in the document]
loosebazooka: oh just kidding, you actually want it in the fixed order the user defined.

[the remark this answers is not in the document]
loosebazooka: linked hash set?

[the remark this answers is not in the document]
TadCordle: Do we actually want it to be ordered the way the user sets it, or should it be a specific order? User-defined order makes the most sense to me but I don't know how useful that is.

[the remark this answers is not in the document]
coollog: I think either works depending on what we want to guarantee. User-defined order would just mean calling builder methods in different order produces different image (which may be find since `addJvmFlag` would be that case unless we sorted those or something).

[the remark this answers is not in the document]
loosebazooka: And since it's a specific builder, we can be opinionated about how we think java should be ordered?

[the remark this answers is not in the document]
loosebazooka: I think I confused myself, but does the order of setting these layers affect how the entry point is built?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,286 @@
...
+   * Adds dependency JARs to {@code /app/libs} on the image.
+   *
+   * @param dependencyFiles the list of dependency JARs to add to the image
+   * @return this
+   * @throws IOException if adding the layer fails
+   */
+  public JavaContainerBuilder addDependencies(List<Path> dependencyFiles) throws IOException {
+    for (Path file : dependencyFiles) {
+      if (!Files.exists(file)) {
+        throw new IOException("File '" + file + "' does not exist.");
```
coollog: `FileNotFoundException` or `NoSuchFileException`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,286 @@
...
+  private final LinkedHashSet<String> classpath = new LinkedHashSet<>(4);
+
+  @Nullable private String mainClass;
+
+  private JavaContainerBuilder(JibContainerBuilder jibContainerBuilder) {
+    this.jibContainerBuilder = jibContainerBuilder;
+  }
+
+  /**
+   * Adds dependency JARs to {@code /app/libs} on the image.
```
coollog: Perhaps we might not want to indicate where the location is on the image since if we change the constants we would have to update this javadoc and it might be an implementation detail that is not necessary to describe in this `JavaContainerBuilder` API.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,286 @@
...
+   *
+   * @param dependencyFiles the list of dependency JARs to add to the image
+   * @return this
+   * @throws IOException if adding the layer fails
+   */
+  public JavaContainerBuilder addDependencies(List<Path> dependencyFiles) throws IOException {
+    for (Path file : dependencyFiles) {
+      if (!Files.exists(file)) {
```
coollog: I think we should check all existence before adding to the `layerConfigurationsBuilder` since it might be surprising to the user if an exception occurred and files were still added.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,286 @@
...
+      }
+
+      layerConfigurationsBuilder.addFile(
+          file.getFileName().toString().contains("SNAPSHOT")
+              ? LayerType.SNAPSHOT_DEPENDENCIES
+              : LayerType.DEPENDENCIES,
+          file,
+          DEPENDENCIES_PATH.resolve(file.getFileName()));
+    }
+    classpath.add(DEPENDENCIES_PATH.resolve("*").toString());
```
coollog: Suggest adding this as a constant like `DEPENDENCIES_CLASSPATH`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,286 @@
...
+   * @return this
+   * @throws IOException if adding the layer fails
+   */
+  public JavaContainerBuilder addResources(Path resourceFilesDirectory) throws IOException {
+    if (!Files.exists(resourceFilesDirectory)) {
+      throw new IOException("Directory '" + resourceFilesDirectory + "' does not exist.");
+    }
+    if (!Files.isDirectory(resourceFilesDirectory)) {
+      throw new IOException(
+          "Adding resources failed: '" + resourceFilesDirectory + "' is not a directory");
```
coollog: `NotDirectoryException`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,286 @@
...
+    return addToClasspath(Arrays.asList(otherFiles));
+  }
+
+  /**
+   * Adds JVM flags to use when starting the application.
+   *
+   * @param jvmFlags the list of JVM flags to add
+   * @return this
+   */
+  public JavaContainerBuilder addJvmFlags(List<String> jvmFlags) {
```
coollog: I think we should have a singular case too

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,286 @@
...
+   * @param jvmFlags the list of JVM flags to add
+   * @return this
+   */
+  public JavaContainerBuilder addJvmFlags(String... jvmFlags) {
+    this.jvmFlags.addAll(Arrays.asList(jvmFlags));
+    return this;
+  }
+
+  /**
+   * Sets the main class used to start the application on the image. To automatically infer the main
```
coollog: nit: "automatically infer" doesn't really apply here - "To find the main class from {@code .class} files, use..."

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,314 @@
...
+   */
+  public static JavaContainerBuilder from(RegistryImage registryImage) {
+    return new JavaContainerBuilder(Jib.from(registryImage));
+  }
+
+  private final JibContainerBuilder jibContainerBuilder;
+  private final JavaLayerConfigurations.Builder layerConfigurationsBuilder =
+      JavaLayerConfigurations.builder();
+  private final List<String> jvmFlags = new ArrayList<>();
+  private final LinkedHashSet<String> classpath = new LinkedHashSet<>(4);
```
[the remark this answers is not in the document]
loosebazooka: It's nice that there's some indication that the Set is ordered built into the type. We've very specifically chosen this. However I don't know if the general guideline is to do that in the variable name or explicitly in the type.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,314 @@
...
+    return this;
+  }
+
+  /**
+   * Adds JVM flags to use when starting the application.
+   *
+   * @param jvmFlags the list of JVM flags to add
+   * @return this
+   */
+  public JavaContainerBuilder addJvmFlags(List<String> jvmFlags) {
```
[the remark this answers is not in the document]
TadCordle: I think we should keep them, the `(String...)` variant is nice for hardcoded JVM flags, and if we have that we might as well keep the `(List<String>)` overload.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  /**
+   * Adds the contents of a classes directory to the image.
+   *
+   * @param classFilesDirectory the directory containing the class files
+   * @return this
+   * @throws IOException if adding the layer fails
+   */
+  public JavaContainerBuilder addClasses(Path classFilesDirectory) throws IOException {
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -0,0 +1,305 @@
...
+  }
+
+  /**
+   * Adds the contents of a resources directory to the image.
+   *
+   * @param resourceFilesDirectory the directory containing the project's resources
+   * @return this
+   * @throws IOException if adding the layer fails
+   */
+  public JavaContainerBuilder addResources(Path resourceFilesDirectory) throws IOException {
```
chanseokoh: This may work for many jib-core uses, but it is a bit limited in that, e.g., our plugins won't be able to use this public API. For example, class files and resource files will be mixed under `target/classes`.

I'm not saying this has to be addressed in this PR, but keep in mind that if the API is like this, our plugins cannot be the user of this API. But then, I guess you don't intend for our plugins to use this API, since it doesn't have the ability to change the app root or put some files in the app root?

chanseokoh:
> Sorry if this has been discussed, but for all of these, what if there are duplicate file names?

I mean I know what happens (only the last file will win and survive), e.g.,

```java
  .addToClasspath("/loser/messages.properties",
                  "/winner/messages.properties")
```

so my question is if this is what we really intended, and if so, I think we should explain it in the Javadocs.

TadCordle:
@chanseokoh I can't remember where it was discussed, but I think this is intended behavior, and you are correct that I did forget to mention it in the javadocs. Will fix.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
coollog: Maybe we should add overloads with a second parameter that's a path filter? I think for app root we might want to keep that out of this API and just have the user use the `JibContainerBuilder` API to work with that.

coollog:
This LGTM - let's follow up this with having our plugins use this new API and make sure everything works the same as before.

PR #1353, opened 2018-12-17
Fail build if using default base image and Java version >= 9

### LINKED ISSUE #1143: Fail the build if a suitable image isn't available
**Description of the issue**:
I successfully built my sample project but it crashed upon launch. I figured out that Jib uses the Docker image gcr.io/distroless/java which is based on Java 8. My sample project is based on Java 11.

**Expected behavior**:
The Maven plugin could choose an image more suited for my project or fail the build with a helpful error message in case a suitable image isn't available.

**Steps to reproduce**:
```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <version>3.8.0</version>
    <configuration>
        <release>11</release>
    </configuration>
</plugin>
```

**Environment**:
Windows 10, Maven 3.5.0, OpenJDK 18.9 (build 11+28)

```xml
<plugin>
    <version>0.9.11</version>
    <configuration>
        <to>
            <image>sample-project</image>
        </to>
    </configuration>
</plugin>
```

**Log output**:
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/example/business/items/ItemResource has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0

TadCordle:
Fixes #1143


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -101,6 +101,8 @@ public void buildDocker()
 
     // Asserts required @Input parameters are not null.
     Preconditions.checkNotNull(jibExtension);
+    PluginConfigurationProcessor.checkJavaVersion(
+        jibExtension.getFrom().getImage(), "jib.to.image");
```
chanseokoh: "jib.from.image"

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -78,6 +78,8 @@ public void buildImage()
           InferredAuthRetrievalException {
     // Asserts required @Input parameters are not null.
     Preconditions.checkNotNull(jibExtension);
+    PluginConfigurationProcessor.checkJavaVersion(
+        jibExtension.getFrom().getImage(), "jib.to.image");
```
chanseokoh: "jib.from.image"

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -100,6 +100,8 @@ public void buildTar()
           InferredAuthRetrievalException {
     // Asserts required @Input parameters are not null.
     Preconditions.checkNotNull(jibExtension);
+    PluginConfigurationProcessor.checkJavaVersion(
+        jibExtension.getFrom().getImage(), "jib.to.image");
```
chanseokoh: "jib.from.image"

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -102,6 +105,38 @@
...
+   * @throws InvalidImageReferenceException if the project is incompatible with the base image
+   */
+  public static void checkJavaVersion(
+      @Nullable String baseImageConfiguration, String fromImageParameterName)
+      throws InvalidImageReferenceException {
+    // TODO: check for other Java 8 base images?
+    if (baseImageConfiguration == null
+        || baseImageConfiguration.equals(DEFAULT_BASE_IMAGE)
+        || baseImageConfiguration.equals(DEFAULT_BASE_IMAGE_WAR)) {
+      String version = System.getProperty("java.version");
```
chanseokoh: I think this reads the JVM version running Maven/Gradle, which may do the job. I just wonder if we should be reading the target version in the build files. For example, looking at `<maven.compiler.target>` or `<maven.complier.release>` (this property seems new after Java 8) or the `<target>` version of `maven-compiler-plugin`. @GoogleContainerTools/java-tools thoughts?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -66,6 +67,12 @@ public void execute() throws MojoExecutionException {
           HelpfulSuggestions.forDockerNotInstalled(HELPFUL_SUGGESTIONS_PREFIX));
     }
 
+    try {
+      PluginConfigurationProcessor.checkJavaVersion(getBaseImage(), "<from><image>");
+    } catch (InvalidImageReferenceException ex) {
+      throw new MojoFailureException(ex.getMessage());
```
coollog: Set `ex` as cause

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -102,6 +105,38 @@
         appRoot, rawConfiguration.getJvmFlags(), mainClass);
   }
 
+  /**
+   * Checks whether or not current Java version is compatible with the base image.
+   *
+   * @param baseImageConfiguration the configured base image
+   * @param fromImageParameterName the name of the base image configuration parameter
+   * @throws InvalidImageReferenceException if the project is incompatible with the base image
```
coollog: I don't think `InvalidImageReferenceException` is the right exception use here since that refers to an invalid image reference for parsing, not the actual contents of the image in relation to the proper runtime.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -102,6 +105,38 @@
...
+   */
+  public static void checkJavaVersion(
+      @Nullable String baseImageConfiguration, String fromImageParameterName)
+      throws InvalidImageReferenceException {
+    // TODO: check for other Java 8 base images?
+    if (baseImageConfiguration == null
```
coollog: nit: I would invert the condition here and early-return so the rest of the method is not inside this conditional

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
coollog: `getProperty` can return null

coollog: Yea I agree that this should use the target version defined in the build file since that could be completely different from the JVM version used to run Maven/Gradle.

TadCordle: Any idea how to do this?

coollog: It looks like in Maven, this can be determined from:

1. `maven-compiler-plugin` configuration `<release>` or `<target>` or `<compilerVersion>`
1. properties `maven.compiler.release` or `maven.compiler.target`
1. could potentially be inherited from parent pom - could recursively check parent poms or perhaps look at the effective pom (like what `mvn help:effective-pom` shows)

In gradle, I think it's more straightforward - just look at `targetCompatibility` from the `compileJava` extension.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidJavaVersionException.java
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.common;
+
+/** Thrown when attempting to use an incompatible base image. */
+public class InvalidJavaVersionException extends Exception {
```
coollog: It looks like from the javadoc this should be called `IncompatibleBaseImageException`, or maybe `JavaVersionMismatchException`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
coollog: Another alternative that might be easier is to use javassist to read a class file and check its major version.

TadCordle:
I changed a bunch of stuff but haven't tested it yet, just pushing for progress' sake. You can take a look if you want, but keep that in mind.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -219,4 +221,65 @@ public String getName() {
   public String getVersion() {
     return project.getVersion();
   }
+
+  void validateBaseImageVersion(@Nullable String baseImage) throws MojoFailureException {
+    if (!PluginConfigurationProcessor.usingDefaultBaseImage(baseImage)) {
```
briandealwis: Could we anticipate having a JVM version label and change this to be something like:
```
  Optional<JvmVersion> getBaseImageJvmVersion(@Nullable String image);
  Optional<JvmVersion> getBuildJvmVersion();

  public void validateBaseImageVersion(@Nullable String baseImage) {
    …
  }
```
And we can check now if we have `distroless/java:8` or `:11`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -345,6 +359,12 @@ static AbsoluteUnixPath getAppRootChecked(
     }
   }
 
+  public static int getVersionFromString(String versionString) {
```
briandealwis: This is very specific to JVM versions, so I think the name should reflect that.  We also need some Javadoc to list the JVMs for which we know this to work.  This version interpretation is what has tripped up numerous libraries between each JVM release (e.g., [log4j v1 and Java 9](http://mail.openjdk.java.net/pipermail/jigsaw-dev/2016-July/008654.html))

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -63,8 +64,12 @@ private static BuildConfiguration getBuildConfiguration(JibContainerBuilder jibC
   @Mock private AuthProperty authProperty;
   @Mock private Consumer<LogEvent> logger;
 
+  private String originalJavaVersion;
+
   @Before
   public void setUp() {
+    originalJavaVersion = System.getProperty("java.version");
```
briandealwis: Is this necessary?  I don't see anything using `java.version`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -254,6 +255,27 @@ public String getVersion() {
     return project.getVersion().toString();
   }
 
+  void validateBaseImageVersion(@Nullable String baseImage) {
+    if (!PluginConfigurationProcessor.usingDefaultBaseImage(baseImage)) {
+      return;
+    }
+    Map<String, ?> javaCompile = project.getProperties();
```
loosebazooka: Few comments about this section
1. It's possible that we only need to worry about `targetCompatibility` as that is the byte code level of the generated code?
2. The default value when neither is set is = JavaVersion.current() I think.
3. We can probably query the `JavaPluginConvention` to get these values, instead of parsing project properties.
4. I think `kotlin` might have special behavior for this, but I'm not sure.
5. I think you can use `JavaVersion` to do all your math.
    - For parsing: `JavaVersion.toVersion()` or `JavaVersion.valueOf()`
    - For compating: `parsedJavaVersion.isJava8Compatible()`

TadCordle: Thanks, is there an easy Maven equivalent? `JavaVersion` is gradle-only.

loosebazooka: Yeah I don't know, I remember we had to do some shenanigans to get the version from maven. It might be even more complicated now. Let me look around.

loosebazooka: So here's some old code: https://github.com/GoogleCloudPlatform/app-maven-plugin/blob/7dafc0ee6d5798242aa08ee8b1fe9da8c9b45a8f/src/main/java/com/google/cloud/tools/maven/cloudsdk/CloudSdkMojo.java#L81

But maven compiler plugin compatibility for newer version of maven seem to be using a new config parameter `release` which is not accounted for in this code.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -140,6 +142,25 @@ private static boolean isProgressFooterEnabled() {
...
 
+  /**
+   * Gets the major version number from a Java version string.
+   *
+   * <p>Examples: {@code "1.7" -> 7, "1.8.0_161" -> 8, "10" -> 10, "11.0.1" -> 11}
+   *
+   * @param versionString the string to convert
+   * @return the major version number as an integer
+   */
+  private static int getVersionFromString(String versionString) {
```
loosebazooka: What's the consequence of `versionString` being some nonsense instead of a parseable value?

Would it just be `0`?

Would be nice to add unit tests to this function specifically since it's doing something that is non-trivial.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -254,6 +257,26 @@ public String getVersion() {
     return project.getVersion().toString();
   }
 
+  void validateBaseImageVersion(@Nullable String baseImage) {
```
loosebazooka: nit: Naming is a little misleading since we're not explicitly validating all base image version. Would prefer something that matches the meaning of "validate base image version if build uses default base image"

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
TadCordle: Maybe I'm not totally sure what you mean, but is this something that's currently used? If it's not currently in practice I think we can hold off on it for now.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
TadCordle: The way it's written right now, it shouldn't be tied to any specific JVM versions, unless they change the naming scheme again in the future? Can you give examples?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: I was understanding it in the same way, @TadCordle, so I agree we can hold off for now.

PR #1359, opened 2018-12-17
Add Automatic-Module-Name: in MANIFEST.MF

chanseokoh:
Following [Stephen Colebourne's excellent advice on using the super-package as the module name](https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html), I suggest **`com.google.cloud.tools.jib`** as the module name for our `jib-core` library.

Note that once we decided the name, we should never change after the library goes GA.

Note that any traditional JARs prior to Java 9 are automatically turned into modules (the feature called `automatic modules` when specified on the module path. Even now, users can use `jib-core` with JPMS. However, if we do not declare an official name for `jib-core`, [it will become ugly in the JPMS world in the not-too-distant future](http://branchandbound.net/blog/java/2017/12/automatic-module-name/).

I've done a few sanity checks suggested in the above link, and our `jib-core` appears to be safe.
- No internal types from the JDK: `~/jdk-11.0.1/bin/jdeps --jdk-internals build/libs/jib-core-0.1.1-SNAPSHOT.jar` shows nothing.
- No default (unnamed) package.
- No split packages.
- No `META-INF/services`, so nothing to worry.

Some details:

## Status quo ##

- The name of the automatic module is `jib.core`, derived from the JAR name, `jib-core-<version>.jar`:
```
$ ./gradlew build
$ ~/jdk-11.0.1/bin/java --module-path build/libs/jib-core-0.1.1-SNAPSHOT.jar --describe-module jib.core
jib.core@0.1.1-SNAPSHOT file://.../jib/jib-core/build/libs/jib-core-0.1.1-SNAPSHOT.jar automatic
requires java.base mandated
contains com.google.cloud.tools.jib
contains com.google.cloud.tools.jib.api
contains com.google.cloud.tools.jib.async
contains com.google.cloud.tools.jib.blob
...
```

- `MANIFEST.MF` doesn't have the `Automatic-Module-Name` entry:
```
$ unzip -p build/libs/jib-core-0.1.1-SNAPSHOT.jar META-INF/MANIFEST.MF | grep -v '^ ' | cut -d: -f1
Manifest-Version
Bnd-LastModified
Bundle-DocURL
Bundle-License
Bundle-ManifestVersion
Bundle-Name
Bundle-SymbolicName
Bundle-Vendor
Bundle-Version
Created-By
Export-Package
Implementation-Version
Import-Package
Require-Capability
Tool
```

## After the PR ###
- The name of the automatic module is `com.google.cloud.tools.jib`:
```
$ ./gradlew build
$ ~/jdk-11.0.1/bin/java --module-path build/libs/jib-core-0.1.1-SNAPSHOT.jar --describe-module jib.core
jib.core not found

$ ~/jdk-11.0.1/bin/java --module-path build/libs/jib-core-0.1.1-SNAPSHOT.jar --describe-module com.google.cloud.tools.jib
com.google.cloud.tools.jib@0.1.1-SNAPSHOT file://.../jib/jib-core/build/libs/jib-core-0.1.1-SNAPSHOT.jar automatic
requires java.base mandated
contains com.google.cloud.tools.jib
contains com.google.cloud.tools.jib.api
contains com.google.cloud.tools.jib.async
...
```
- `MANIFEST.MF` has the `Automatic-Module-Name` entry:
```
$ unzip -p build/libs/jib-core-0.1.1-SNAPSHOT.jar META-INF/MANIFEST.MF | grep Automatic-Module-Name
Automatic-Module-Name: com.google.cloud.tools.jib
```

### DIFF
--- a/jib-core/build.gradle
+++ b/jib-core/build.gradle
@@ -60,6 +60,7 @@ dependencies {
 jar {
   manifest {
     attributes 'Implementation-Version': version
+    attributes 'Automatic-Module-Name': 'com.google.cloud.tools.jib'
 
     // OSGi metadata
     attributes 'Bundle-SymbolicName': 'com.google.cloud.tools.jib'


chanseokoh:
As we discussed, I think this is the best we can do regarding JPMS support ATM. Any traditional JARs can already work on the module path thanks to the automatic modules feature, and giving the right name is what is required.

There may be a way to add proper module info without Java >8 or `module-info.java`, but it may not be trivial or worth the effort to achieve it. For the proper module declaration, I think moving our build environment to, say, Java 11 should be the first step, which we won't be doing soon.

In the meantime, I'll be doing some more testing with this regarding JPMS.

chanseokoh:
Another related ref is Guava ([Add an Automatic-Module-Name manifest entry](https://github.com/google/guava/pull/2846)).

loosebazooka:
Doesn't using `com.google.cloud.tools.jib` prevent us from using that package name from anything in the future?

Like if we were to release a registry client, we couldn't use `com.google.cloud.tools.jib.registry`?

chanseokoh:
Good question. [Stephen Colebourne's blog post explains sub-packages and sub-modules](https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html) too.

> **Can you have sub-modules?**
> Yes. When a module name is chosen, the developer is taking control of a namespace. That namespace consists of the module name and all sub-names below it - sub-package names and sub-module names.
>
> Ownership of that namespace allows the developer to release one module or many. The main constraint is that there should not be two published modules containing the same package.
>
> As a side effect of this, the practice of larger projects releasing an "all" jar will need to stop. An "all" jar is used when the project has lots of separate jar files, but also wants to allow end-users to depend on a single jar file. These "all" jar files are a pain in Maven dependency trees, but will be a disaster in JPMS ones, as there is no way to override the metadata, unlike in Maven.

So, we fully own and control the `com.google.cloud.tools.jib` namespace, and we will be able to create a module named `com.google.cloud.tools.jib.registry`. However, no two modules should have the same package.

loosebazooka:
Ah so registry couldn't have a package called `com.google.cloud.tools.jib` or for that matter `com.google.cloud.tools.jib.api` or whatever.

does `com.google.cloud.tools.jib` conflict with any of the plugins jars or plugins-common?

chanseokoh:
> Ah so registry couldn't have a package called `com.google.cloud.tools.jib` or for that matter `com.google.cloud.tools.jib.api` or whatever.

Yeah, split packages are evil. [JPMS does not allow split packages in different modules. This restriction is even applied to the packages that are not exported. (Only unnamed modules are exempted from this restriction.)](https://www.logicbig.com/tutorials/core-java-tutorial/modules/split-packages.html)

> does com.google.cloud.tools.jib conflict with any of the plugins jars or plugins-common?

Good question. Fortunately, `plugins-common` starts with `com.google.cloud.tools.jib.plugins.common`. If it had `com.google.cloud.tools.jib`, we would introduce a new sub-package. For the plugins JARs, we have `com.google.cloud.tools.jib.maven` and `com.google.cloud.tools.jib.gradle`.

However, we don't really publish those JARs, and even if they had `com.google.cloud.tools.jib` as the top-level package, I (naively ) think renaming could wait until we claim these "libraries" are GA intended to be used with `jib-core`. Anyways, we have different top-level package names, so we're good.

chanseokoh:
An alternative is to move everything to `com.google.cloud.tools.jib.core` (`com.google.cloud.tools.jib` becomes empty), and have that as the module name for `jib-core`.

loosebazooka:
Yeah the alternative isn't all that great huh. If only one library can occupy the jib package name, it might as well be our core jib offering.

loosebazooka:
I think this LGTM, however i think it might be nice for all of us to catch up on this tomorrow during the bugout or during the design meeting.

briandealwis:
LGTM too.  We need to sort out what is intended to be exposed by Jib Core in terms of API: there a number of types referenced from `JibContainerBuilder` that are either not in `.api` or that can only be created by through classes not in `.api`.

coollog:
Yea, as a side note, I think we should also move the classes in `.api` out onto the root of `com.google.cloud.tools.jib`.

loosebazooka:
Should we add this to the 1.0 milestone?

chanseokoh:
I tested building and running a very simple `jib-core` client application with JDK 11 using the module path way.

`com/example/JibCoreTest.java`:
```java
package com.example;

import com.google.cloud.tools.jib.api.Containerizer;
import com.google.cloud.tools.jib.api.DockerDaemonImage;
import com.google.cloud.tools.jib.api.Jib;

public class JibCoreTest {

  public static void main(String[] args) throws Exception {
    System.out.print("Containerizing... ");
    Containerizer containerizer = Containerizer.to(DockerDaemonImage.named("francium25/test"));
    Jib.from("francium25/base").containerize(containerizer);
    System.out.println("done.");
  }
}
```

1. Have the `jib-core` JAR in the current directory and compile:
   ```
   $ ~/jdk-11.0.1/bin/javac \
   >   --module-path jib-core-0.1.1-SNAPSHOT.jar \
   >   --add-modules com.google.cloud.tools.jib \
   >   com/example/JibCoreTest.java
   ```

1. Have all the dependency JARs in the current directory and run:
   ```
   $ ~/jdk-11.0.1/bin/java \
   >   --module-path $( echo *.jar | tr ' ' : ) \
   >   --add-modules com.google.cloud.tools.jib \
   >   com.example.JibCoreTest
   Containerizing... done.
   ```

The dependency JARs I used are
```
├── animal-sniffer-annotations-1.14.jar
├── checker-qual-2.0.0.jar
├── commons-codec-1.3.jar
├── commons-compress-1.18.jar
├── commons-logging-1.1.1.jar
├── error_prone_annotations-2.0.18.jar
├── google-http-client-1.23.0.jar
├── guava-23.5-jre.jar
├── httpclient-4.0.1.jar
├── httpcore-4.0.1.jar
├── j2objc-annotations-1.1.jar
├── jackson-annotations-2.9.0.jar
├── jackson-core-2.9.6.jar
├── jackson-databind-2.9.6.jar
├── javassist-3.22.0-GA.jar
├── jib-core-0.1.1-SNAPSHOT.jar
└── jsr305-1.3.9.jar
```
There are all automatic modules, some of which have the `Automatic-Module-Name`, some of which don't.

chanseokoh:
Ah, just realized I can specify a directory in `--module-path`.

chanseokoh:
When adding `com/example/module-info.java` in the sample app,
```java
module com.example {
  requires com.google.cloud.tools.jib;
}
```
there is no need for `--add-modules com.google.cloud.tools.jib`, and the app can be run with `--module com.example/com.example.JibCoreTest`.

```
$ ~/jdk-11.0.1/bin/javac -d out --module-path . com/example/*.java
$ ~/jdk-11.0.1/bin/java --module-path .:out --module com.example/com.example.JibCoreTest
Containerizing... done.
```

chanseokoh:
I think we are all good with the module name `com.google.cloud.tools.jib`. Sounds like the right name from both technical and non-technical points of view. Doesn't seem like we will ever change it. Merging it.

Issue #1362, opened 2018-12-18
Wait for logging thread to finish before throwing errors.


coollog:
Currently, when an error occurs, the output gets messed up like:

```
[INFO] --- jib-maven-plugin:0.10.2-SNAPSHOT:build (default-cli) @ hello-world ---
[INFO] 
[INFO] Containerizing application to gcr.io/qingyangc-sandbox/jibtestimage:20181218-180002889...
[INFO] The base image requires auth. Trying again for openjdk:8-jdk...
[INFO] Executing tasks:
[INFO] [==================                                ] 36.4% complete
[INFO] > setting up to push layers
[INFO] 
[INFO] Executing tasks:2 s
[INFO] [==================                                ] 36.4% complete
[INFO] Finished at: 2018-12-18T13:00:07-05:00
[INFO] ------------------------------------------------------------------------
[INFO] > setting up to push layers
[INFO] > checking base image layer sha256:d6341e30912f12f56e18564a3b582853f65376766f5f9d641a68a724ed6db88f
[INFO] > checking base image layer sha256:087a57faf9491b1b82a83e26bc8cc90c90c30e4a4d858b57ddd5b4c2c90095f6
[INFO] > checking base image layer sha256:5d71636fb824265e30ff34bf20737c9cdc4f5af28b6bce86f08215c55b89bfab
[INFO] > checking base image layer sha256:54f7e8ac135a5f502a6ee9537ef3d64b1cd2fa570dc0a40b4d3b6f7ac81e7486
[INFO] > checking base image layer sha256:9da6b28682cfe6db721c143309125728417dca93643b3d8d4c22ae7fbb4eb940
[INFO] > checking base image layer sha256:203f1094a1e2165da6f6ec505e8cffb8853d3c72a8088f41c25218121f883b0a
[ERROR]
[ERROR]Executing tasks:
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[INFO] 
```

PR #1364, opened 2018-12-18
Adds support for rc release to prepare_release.sh.

coollog:
To allow for us to release `v1.0.0-rc1` after #1362 .

This is based on https://github.com/GoogleCloudPlatform/app-gradle-plugin/blob/master/scripts/prepare_release.sh

### DIFF
--- a/jib-gradle-plugin/scripts/prepare_release.sh
+++ b/jib-gradle-plugin/scripts/prepare_release.sh
@@ -1,11 +1,7 @@
 #!/bin/bash -
-# Usage: ./scripts/prepare_release.sh <release version>
+# Usage: ./scripts/prepare_release.sh <release version> [<post-release version>]
 
 set -e
-
-Colorize() {
-	echo "$(tput setff $2)$1$(tput sgr0)"
-}
 
 EchoRed() {
 	echo "$(tput setaf 1; tput bold)$1$(tput sgr0)"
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/scripts/prepare_release.sh
```diff
@@ -20,20 +16,24 @@ Die() {
...
 }
 
-[ $# -ne 2 ] || DieUsage
+[ $# -ne 1 ] && [ $# -ne 2 ] && DieUsage
 
 EchoGreen '===== RELEASE SETUP SCRIPT ====='
 
 VERSION=$1
 CheckVersion ${VERSION}
+if [ $2 ]; then
```
chanseokoh: I think the script will fail if $2 is not given. Try `if [ -n "$2" ];`.

coollog: This sets `POST_RELEASE_VERSION` if it is given in the command line `[ $2 ];`, and doesn't set it if not.

chanseokoh: I meant the script will give you a syntax error if $2 is not given or empty, as it ends up `if [ ]`. You need to guard it with quotes.

chanseokoh: I just realized `[ ]` or `[ -n ]` does not crash, but it's desirable to quote it, as `[ == "something" ]` surely fails: https://stackoverflow.com/questions/53470679/bash-if-statement-with-prepended-character

PR #1366, opened 2018-12-18
Fix issues around adding proxy credentials to Apache HttpClient

### LINKED ISSUE #1365: Finish propagating Maven proxy auth configuration
After #1337. I'll put up a PR to fix several things shortly.

chanseokoh:
Fixes #1365.

Fixes several issues after merging #1337.

- `System.getProperty("http.proxyPort", "8080")`

   [According to the Oracle doc, the default proxy port is 80, not 8080](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html).

- `NumberFormatException`: no validation on `http(s).proxyPort`

- No tests

- `@Nullable` parameter

- Mutable `PROXY_PROPERTIES` (The name cannot be capitalized.)

- Minor doc issues

### DIFF
--- a/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/ProxyProviderTest.java
+++ b/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/ProxyProviderTest.java
@@ -45,7 +45,7 @@ public class ProxyProviderTest {
           "http.nonProxyHosts");
 
   // HashMap to allow saving null values.
-  private HashMap<String, String> savedProperties = new HashMap<>();
+  private final HashMap<String, String> savedProperties = new HashMap<>();
 
   @Before
   public void setUp() {
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/http/ConnectionWithProxyCredentialsTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/http/ConnectionWithProxyCredentialsTest.java
@@ -43,10 +43,10 @@ public class ConnectionWithProxyCredentialsTest {
           "https.proxyUser",
           "https.proxyPassword");
 
-  private final ApacheHttpTransport transport = new ApacheHttpTransport();
-
   // HashMap to allow saving null values.
-  private HashMap<String, String> savedProperties = new HashMap<>();
+  private final HashMap<String, String> savedProperties = new HashMap<>();
+
+  private final ApacheHttpTransport transport = new ApacheHttpTransport();
 
   @Before
   public void setUp() {


jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -87,40 +88,34 @@
...
+
+  private static void addProxyCredentials(ApacheHttpTransport transport, String protocol) {
+    Preconditions.checkArgument(protocol.equals("http") || protocol.equals("https"));
+
+    boolean addCredentials =
+        System.getProperty(protocol + ".proxyHost") != null
+            && System.getProperty(protocol + ".proxyUser") != null
+            && System.getProperty(protocol + ".proxyPassword") != null;
+
+    if (addCredentials) {
```
coollog:
```
if (!addCredentials {
  return;
}
...
```

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -87,40 +88,34 @@
...
+      String defaultProxyPort = protocol.equals("http") ? "80" : "443";
+
+      DefaultHttpClient httpClient = (DefaultHttpClient) transport.getHttpClient();
       httpClient
           .getCredentialsProvider()
           .setCredentials(
               new AuthScope(
-                  System.getProperty("https.proxyHost"),
-                  Integer.parseInt(System.getProperty("https.proxyPort", "443"))),
+                  System.getProperty(protocol + ".proxyHost"),
```
coollog: nit: Perhaps change to single calls to `getProperty` and hold in variables?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -16,59 +16,49 @@
...
-  /**
-   * Set proxy system property if it has a proper value
-   *
-   * @param property property name
-   * @param value property value
-   */
-  private static void setProxyProperty(String property, String value) {
-    if (property != null && value != null) {
-      System.setProperty(property, value);
+    if (!proxyPropertiesSet(protocol)) {
```
coollog:
```
if (areProxyPropertiesSet(protocol)) {
  return;
}
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -79,6 +69,8 @@ private static void setProxyProperty(String property, String value) {
    * @param protocol protocol
    */
   private static boolean proxyPropertiesSet(String protocol) {
```
coollog: maybe rename as `areProxyPropertiesSet`?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -16,60 +16,52 @@
...
     }
-  }
 
-  /**
-   * Set proxy system property if it has a proper value
-   *
-   * @param property property name
-   * @param value property value
-   */
-  private static void setProxyProperty(String property, String value) {
```
chanseokoh: Just realized we need this.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -87,41 +88,32 @@
...
+
+    String proxyHost = System.getProperty(protocol + ".proxyHost");
+    String proxyUser = System.getProperty(protocol + ".proxyUser");
+    String proxyPassword = System.getProperty(protocol + ".proxyPassword");
+    if (proxyHost == null || proxyUser == null || proxyPassword == null) {
+      return;
     }
+
+    String defaultProxyPort = protocol.equals("http") ? "80" : "443";
+    int proxyPort = Integer.parseInt(System.getProperty(protocol + ".proxyPort", defaultProxyPort));
```
coollog: Should we catch the `NumberFormatException`?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -87,41 +88,32 @@
...
+
+    DefaultHttpClient httpClient = (DefaultHttpClient) transport.getHttpClient();
```
coollog: Could we avoid this cast?

chanseokoh: Doesn't seem like we can. Need `DefaultHttpClient` for `getCredentialsProvider()`, which is not part of `HttpClient`. In our codebase, this is always a `DefaultHttpClient` instance.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
chanseokoh: In principle, I think `addProxyCredentials()` and hence `getConnectionFactory()` should throw some custom checked exception as the error is due to an external input, so that a higher-level application can act accordingly if this happens, but from a practical standpoint, I don't feel it is worth disrupting a large part of the codebase for the sake of propagating a checked exception.

So, I'm leaning toward of letting it throw the runtime exception as-is, meaning that if any application (including our Maven/Gradle plugins) ever surfaces the `NumberFormatException`, then it would indicate, righteously, a genuine program bug of the application. That said, our plugins should never let this happen. One way we can achieve this is to validate `http(s).proxyPort` at the higher application level. In fact, we are already doing such a check with `jib.httpTimeout`. I'll add a check in `JibSystemProperties` for `http(s).proxyPort`, and I think it's the best option here.

chanseokoh:
I'm going to add tests.

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -83,18 +84,38 @@ public static boolean isUserAgentEnabled() {
...
+  }
+
+  /**
+   * Checks if {@code http.proxyPort} and {@code https.proxyPort} system properties are in the
+   * [0..65535] range.
+   *
+   * @throws NumberFormatException if invalid values
+   */
+  public static void checkProxyPortProperty() throws NumberFormatException {
+    checkNumericSystemProperty("http.proxyPort", Range.closed(0, 65535));
```
loosebazooka: Does it make any sense to make `http.proxyPort` and `https.proxyPort` constants?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
loosebazooka: A little curious if switching to https://github.com/googleapis/google-http-java-client/tree/master/google-http-client-apache will give us an updated api for this?

chanseokoh: We are using Google HTTP Client. There might be an alternative way to set a proxy and credentials (or not really, don't know), but this is a proven way at least.

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
chanseokoh: Yeah, I think `private` constants work here.

chanseokoh: On second thought, I think direct strings are fine. Used only in this line, and it is not that the strings will ever change. Does not seem worth a name as a field.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
loosebazooka: Yeah, sorry I guess my comment wasn't clear. I tried finding a solution using the current version, but the ApacheHttpTransport (and Builder) objects don't have a good passthrough for setting proxy credentials, which is a shame.

I was hoping, give that they are doing work on this, that we might be able to use new features, but I'm not sure exactly what their plan is.

loosebazooka: Anyway, we can ignore my comments for now.

chanseokoh:
Added tests, and ready for review. But the last commits are just tests.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/ProxyProviderTest.java
```diff
@@ -0,0 +1,270 @@
...
+          "http.proxyPort",
+          "http.proxyUser",
+          "http.proxyPassword",
+          "https.proxyHost",
+          "https.proxyPort",
+          "https.proxyUser",
+          "https.proxyPassword",
+          "http.nonProxyHosts");
+
+  // private static Proxy createProxy(@Nullable protocol, @Nullable )
```
loosebazooka: need to remove this comment?

loosebazooka:
I wonder if we should have an intermediate object for proxy config instead of using system properties as the source of data when configuring proxies? Not saying it needs to be done here in this PR, but the fact that we're translating maven settings to system properties seems odd.

chanseokoh:
I get what you're saying. My 2 cents:

We have simply been piggybacking on the usual system properties for proxy support, which `java.net` JDK stuff and Apache HttpClient honor, except that Apache HttpClient did not handle the username and the password (not sure why). Likewise, currently jib-core will honor those usual system properties (including the auth properties after this PR). In that sense, I think supporting the system properties in the current way is reasonable.

Next, as you said, it may make sense to pass down the proxy info with an interim object. However, it will be useless for Gradle and we will most likely end up setting the system properties in jib-core anyway (unless we use some APIs of the HTTP client library), so I also wonder if it is worth the extra complexity. The system properties will be scoped inside the Maven plugin, so setting them via properties looks OK to me.

PR #1374, opened 2018-12-20
Allow forcing containerization type

chanseokoh:
https://github.com/GoogleContainerTools/jib/issues/1281#issuecomment-444605402

Basically ready for review, except the names of the config and values can change anytime.

- The value `war` forces WAR containerization.
- `java` is intended for non-WAR but currently, anything but `war` will force non-WAR.
- I've been recently thinking `war-archived` and `jar-archived` for containerizaing a single-WAR/JAR.

- `boolean isWarPackaging()`: I could have added an `enum` instead of just returning `boolean` for whether Jib should do "WAR containerization", but at this point, it seemed not necessary. I thought this would also help creating a smaller PR.


chanseokoh:
Basically ready for review, except the names of the config and values can change anytime.

- The value `war` forces WAR containerization.
- `java` is intended for non-WAR but currently, anything but `war` will force non-WAR.
- I've been recently thinking `war-archived` and `jar-archived` for containerizaing a single-WAR/JAR.

- `boolean isWarPackaging()`: I could have added an `enum` instead of just returning `boolean` for whether Jib should do "WAR containerization", but at this point, it seemed not necessary. I thought this would also help creating a smaller PR.

loosebazooka:
I'm going to look again tomorrow, but here are my initial thoughts.
- It's my favorite that we have to pass `containerizeWar` in so many places. It'd be nice if we could set it once and all operations after the fact are either *jar* or *war* operations. I haven't looked at it deep enough to understand how exactly to do that though.
- Would it make sense to have packaging types `app` and `webapp` ?

chanseokoh:
> I don't really like that we have to pass `containerizeWar` in so many places. It'd be nice if we could set it once and all operations after the fact are either jar or war operations.

For the methods in `PluginConfigurationProcessor`, one way that comes into my mind is to refactor the class to be instantiated earlier and have `containerizeWar` as a field. Currently, its constructor is private and the entry static method such as `processCommonConfigurationForTarImage()` is somewhat like a static initializer, but the initializer has becomes very big, calling many other static methods to finally instantiate a `PluginConfigurationProcessor` instance, and because these are static methods, we are passing around many arguments including `containerizeWar`. So I could basically try convert `processCommonConfigurationForTarImage()` into a constructor doing a lot of stuff there. (I know some people try to avoid doing a lot of stuff in a constructor, but we may make this an exception?) Will mess around with that idea.

loosebazooka:
@chanseokoh if we ever wanted to support other containerization types, we could accept an enum instead of a boolean, or use a builder/static initializer of some kind?

chanseokoh:
Yeah, certainly. For now, I opted for `boolean`, as I thought having an enum now will make this PR much longer and we could extend it later in an incremental manner when the time comes.

> `boolean isWarPackaging()`: I could have added an enum instead of just returning boolean for whether Jib should do "WAR containerization", but at this point, it seemed not necessary. I thought this would also help creating a smaller PR.

chanseokoh:
@loosebazooka I tried creating a simple bag of input values `CommonInput` and pass it around static methods. Look at this diff: https://github.com/GoogleContainerTools/jib/compare/i1281-packaging-type...i1281-packaging-type-refactor?expand=1 (not pushed into this PR)

So instead of additionally passing `containerizeWar` to `processCommonConfiguration()`, `getBaseImage()`, and `computeEntrypoint()`, it passes this single bag of `CommonInput`. What do you think?

loosebazooka:
I think I might prefer the current (as opposed to the diff). I guess the enum will follow up. Will take a look with the current impl.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -119,7 +123,8 @@ public void buildTar()
       Path tarOutputPath = getTargetPath();
       PluginConfigurationProcessor pluginConfigurationProcessor =
           PluginConfigurationProcessor.processCommonConfigurationForTarImage(
-              new GradleRawConfiguration(jibExtension),
+              rawConfiguration,
+              containerizeWar,
```
loosebazooka: Can this value be obtained from project properties?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -143,4 +145,21 @@ boolean getAllowInsecureRegistries() {
...
     return allowInsecureRegistries.get();
   }
+
+  @Input
+  @Nullable
+  @Optional
+  // TODO: validate the value somewhere. We don't have a central place for config validation yet.
+  // Or, can this be enforced at the plugin level, e.g., by having this as an enum in both Maven and
+  // Gradle?
+  public String getPackagingOverride() {
```
loosebazooka: should we allow packagingOverride="jar" for "war" projects?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -74,5 +46,12 @@ static void disableHttpLogging() {
     java.util.logging.Logger.getLogger(HttpTransport.class.getName()).setLevel(Level.OFF);
   }
 
+  static boolean isWarContainerization(Project project, RawConfiguration rawConfiguration) {
+    if (!rawConfiguration.getPackagingOverride().isPresent()) {
+      return GradleProjectProperties.getWarTask(project) != null;
+    }
+    return "war".equals(rawConfiguration.getPackagingOverride().get());
```
loosebazooka: So we can assume at this point packagingOverride has already been validated?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
chanseokoh: If you mean just returning if the project is WAR (has the Gradle WAR Plugin applied or `<packaging>war` in `pom.xml`) and not about the final packaging decision after factoring in `packagingOverride`, the value can already be obtained from `GradleProjectProperties.getWarTask()` or directly from `MavenProject`. But not sure if you are referring to this.

So, this `containerizeWar` indicates the final packaging decision which should read the `packagingOverride` value, and knowing that, it may not fit very well into "project properties". But even if we decide to add a method like `getFinalPackagingType()` in project properties, it is not possible in the current design. A `GradleProjectProperites` instance is created via the static initializer `GradleProjectProperties.getForProject()`, and for it to create an instance, it has to know ahead if it should do WAR containerization or not (and the correct `appRoot` value based on the WAR containerization decision). In simple words, you can lay an egg only after having a chicken, but to have a chicken, you need an egg. In the past, I was wondering how our codebase can be changed to instantiate a `GradleProjectProperties` without requiring raw configuration, but haven't really found a good way. Currently, project properties is designed to provide `JavaLayerConfigurations`, and creating correct layers require processed config info.

loosebazooka: Ah, my confusion is because containerizeWar is sent to projectProperties, but that appears to be to do some calculations, not to hold the value.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
chanseokoh: I hope. Or it is technically possible to validate the value here. No decision yet. This PR doesn't do any validation (just the TODO for it as you saw above). But for now, any value than "war" will be considered the "flat" packaging (no error). It is not clear where is the best place to validate raw config values. I see some of them are done here and there, and as such, I haven't added validation yet. I have been thinking it'd be nice to have a global config validation phase in a pipeline, but we are not there yet.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
chanseokoh: Currently, any value other than "war" will be considered the "flat" packaging (no error). I hope an invalid value will return an error when validation is set in place. So, yes, packagingOverride="jar" will force the "flat" packaging on a project with `<packaging>war` in `pom.xml`, for example.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
chanseokoh: > Ah, my confusion is because containerizeWar is sent to projectProperties, but that appears to be to do some calculations, not to hold the value.

OK, I think you got what's happening now, but just in case, `Gradle/MavenProjectProperties` gets `containerizeWar` to actually do some calculations, for example, to compute the correct `JavaLayerConfigurations` which requires whether to know if it should do WAR containerization or not. It doesn't need to hold the `containerizeWar`.

chanseokoh:
> As for naming, was thinking about `standard|webapp` or `standard|exploded-war`?

I like `standard` and `web-app` (not `webapp`, I suppose). We can have other names for "packaged-as-a-single-archive" apps.

chanseokoh:
Making the PR up to date.

loosebazooka:
So there are PRs in the jhipster repo right that might remove our responsibility to hack a solution to this: 

Discussion: https://github.com/jhipster/generator-jhipster/pull/8891
Follow-up: https://github.com/jhipster/generator-jhipster/pull/9034

chanseokoh:
Ah, hopefully the PR to change the packaging to `jar` gets in soon (after taking care of Gradle too).

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurations.java
```diff
@@ -56,12 +57,13 @@
...
       Logger logger,
       Path extraDirectory,
       Map<AbsoluteUnixPath, FilePermissions> extraDirectoryPermissions,
       AbsoluteUnixPath appRoot)
       throws IOException {
-    if (GradleProjectProperties.getWarTask(project) != null) {
+    if (containerizeWar) {
       logger.info("WAR project identified, creating WAR image: " + project.getDisplayName());
       return getForWarProject(project, extraDirectory, extraDirectoryPermissions, appRoot);
     } else {
```
coollog: tiny nit: else not necessary

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -104,11 +104,15 @@ public void buildDocker()
     TaskCommon.disableHttpLogging();
 
     try {
-      AbsoluteUnixPath appRoot = TaskCommon.getAppRootChecked(jibExtension, getProject());
+      GradleRawConfiguration rawConfiguration = new GradleRawConfiguration(jibExtension);
+      boolean containerizeWar = TaskCommon.isWarContainerization(getProject(), rawConfiguration);
```
coollog: naming suggestion: `shouldContainerizeAsWar`/`containerizeAsWar`/`containerizeAsWebApp`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
```diff
@@ -106,7 +106,9 @@ public void apply(Project project) {
           try {
             War warTask = GradleProjectProperties.getWarTask(project);
             Task dependsOnTask;
-            if (warTask != null) {
+            if (warTask != null
+                && (jibExtension.getPackagingOverride() == null
```
coollog: Can this use `TaskCommon.isWarContainerization` or at least share the same logic? If `jib.packagingOverride` system property is set, this would not be affected.

jib-maven-plugin/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Setting proxy credentials (via system properties `http(s).proxyUser` and `http(s).proxyPassword`) is now supported.
 - Maven proxy settings are now supported.
 - Now checks for system properties in pom as well as commandline. ([#1201](https://github.com/GoogleContainerTools/jib/issues/1201))
+- `<packagingOverride>{standard|web-app}` configuration parameter to enforce WAR or non-WAR containerization.
```
coollog: `<packagingOverride>{standard|web-app}</packagingOverride>`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MojoCommon.java
```diff
@@ -33,42 +32,20 @@
 class MojoCommon {
 
   /**
-   * Gets whether or not the given project is a war project. This is the case for projects with
-   * packaging {@code war} and {@code gwt-app}.
+   * Computes whether or not it should containerize WAR. This is the case for projects with
+   * packaging {@code war} and {@code gwt-app}, unless non-WAR containierzation is indicated in
```
coollog: containerization

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MojoCommon.java
```diff
@@ -33,42 +32,20 @@
...
-    try {
-      return AbsoluteUnixPath.get(appRoot);
-    } catch (IllegalArgumentException ex) {
-      throw new InvalidAppRootException(appRoot, appRoot, ex);
+  static boolean isWarContainerization(MavenProject project, RawConfiguration rawConfiguration) {
+    if (!rawConfiguration.getPackagingOverride().isPresent()) {
+      String packaging = project.getPackaging();
+      return "war".equals(packaging) || "gwt-app".equals(packaging);
     }
+    return "web-app".equals(rawConfiguration.getPackagingOverride().get());
```
coollog: Could we share this constant "web-app" with Gradle?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenLayerConfigurations.java
```diff
@@ -51,11 +52,12 @@
...
       MavenProject project,
+      boolean containerizeWar,
       Path extraDirectory,
       Map<AbsoluteUnixPath, FilePermissions> extraDirectoryPermissions,
       AbsoluteUnixPath appRoot)
       throws IOException {
-    if (MojoCommon.isWarProject(project)) {
+    if (containerizeWar) {
       return getForWarProject(project, extraDirectory, extraDirectoryPermissions, appRoot);
     } else {
```
coollog: tiny nit: else not necessary

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -81,10 +81,15 @@ public void buildImage()
     TaskCommon.disableHttpLogging();
 
     try {
-      AbsoluteUnixPath appRoot = TaskCommon.getAppRootChecked(jibExtension, getProject());
+      GradleRawConfiguration rawConfiguration = new GradleRawConfiguration(jibExtension);
+      boolean containerizeWar = TaskCommon.isWarContainerization(getProject(), rawConfiguration);
```
coollog: What if we rather stored `containerizeWar` in `RawConfiguration` rather than pass it around as a separate parameter?

chanseokoh:
https://github.com/jhipster/generator-jhipster/pull/9034 is the PR working on migrating from WAR to JAR for both Maven and Gradle. It seems pretty involved, but eventually, I think it will happen. Closing this PR.

PR #1377, opened 2018-12-26
Add JibContainerBuilder#addLayer overload with String instead of AbsoluteUnixPath

### LINKED ISSUE #1184: Consider adding JibContainerBuilder.addLayer that takes string in place of AbsoluteUnixPath.
From https://github.com/GoogleContainerTools/jib/pull/1160#discussion_r226765744

TadCordle:
Fixes #1184

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -120,6 +120,32 @@ public class JibContainerBuilder {
   }
 
   /**
+   * Adds a new layer to the container with {@code files} as the source files and {@code
+   * pathInContainer} as the path to copy the source files to in the container file system.
+   *
+   * @param files the source files to copy to a new layer in the container
+   * @param pathInContainer the path in the container file system corresponding to the {@code
+   *     sourceFile}
+   * @return this
+   * @throws IOException if an exception occurred when recursively listing any directories
+   * @see #addLayer(List, AbsoluteUnixPath) for more details
+   */
+  public JibContainerBuilder addLayer(List<Path> files, String pathInContainer) throws IOException {
+    return addLayer(files, AbsoluteUnixPath.get(pathInContainer));
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -119,6 +119,32 @@ public JibContainerBuilder addLayer(List<Path> files, AbsoluteUnixPath pathInCon
...
 
+  /**
+   * Adds a new layer to the container with {@code files} as the source files and {@code
+   * pathInContainer} as the path to copy the source files to in the container file system.
+   *
+   * @param files the source files to copy to a new layer in the container
+   * @param pathInContainer the path in the container file system corresponding to the {@code
+   *     sourceFile}
+   * @return this
+   * @throws IOException if an exception occurred when recursively listing any directories
```
chanseokoh: How about adding

`@throws IllegalArgumentException if {@code pathInContainer} is not an absolute Unix-style path`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -119,6 +119,32 @@ public JibContainerBuilder addLayer(List<Path> files, AbsoluteUnixPath pathInCon
...
+   * @see #addLayer(List, AbsoluteUnixPath) for more details
```
chanseokoh: I took this as a chance for the `@see` usage: https://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDIEDI

Looking at **Form 3** in the linked doc, it seems like, this line will generate

> **See Also:**
>
> [for more details]()

That is, you only see "for more details" without the method name, where the text is linked to `addLayer()`. Maybe just `@see #addLayer(List, AbsoluteUnixPath)`?

TadCordle: Looks like we do this in a few places, I'll follow up on this.

PR #1378, opened 2018-12-26
Avoid nested Futures by calling combiners in a blocking manner

chanseokoh:
Added `whenAllSucceedBlocking()`, which waits for `combiner` to complete.

This is mainly to [avoid returning nested futures](https://github.com/google/guava/wiki/ListenableFutureExplained#avoid-nested-futures) in `PushImageStep` (although we are sort of OK with the nested futures because the combiner is meant to be called after the depending futures have been completed.) This will give an added benefit of better readability, as it was a bit complex to reason about the number of consecutive `.get()` calls or digest the meaning of returning `ListenableFuture<ListenableFuture<BuildResult>>`.

That said, `PushImageStep` will see a subtle difference in semantics regarding future execution (for example, an outer future may be scheduled to wait until an inner future completes prior to this change, while now an "outer" task will run explicitly after the "inner" task completes), but in the end, I believe the change doesn't matter because `PushImageStep.afterPushSteps()` is a private method, all those consecutive `.get()` calls just wait for all the tasks to complete in the same sequence after this change, and thus `PushImageStep.call()` is a blocking call ultimately.  I think wrapping tasks in `Future` here is an unnecessary layering, when we know how we chain the calls until blocking.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
@@ -22,6 +22,7 @@ import com.google.common.util.concurrent.ListeningExecutorService;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 
 /**
  * Builds a list of dependency {@link ListenableFuture}s to wait on before calling a {@link
@@ -71,7 +72,8 @@ public class AsyncDependencies {
   }
 
   /**
-   * Calls {@code combiner} when all the added futures succeed.
+   * Creates the {@link ListenableFuture} which will return the result of calling {@code combiner}
+   * when all the added futures succeed.
    *
    * @param combiner the {@link Callable}
    * @param <C> the return type of {@code combiner}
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
```diff
@@ -80,4 +82,19 @@ public AsyncDependencies addSteps(List<? extends AsyncStep<?>> asyncSteps) {
   public <C> ListenableFuture<C> whenAllSucceed(Callable<C> combiner) {
     return Futures.whenAllSucceed(futures).call(combiner, listeningExecutorService);
   }
+
+  /**
+   * Calls {@code combiner} when all the added futures succeed.
+   *
+   * @param combiner the {@link Callable}
+   * @param <C> the return type of {@code combiner}
+   * @return a {@link ListenableFuture} to handle completion of the call to {@code combiner}
```
loosebazooka: I think the return type might be wrong here?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
```diff
@@ -97,14 +97,10 @@ public BuildResult call() throws ExecutionException, InterruptedException {
         .addSteps(NonBlockingSteps.get(pushApplicationLayersStep))
         .addStep(NonBlockingSteps.get(pushContainerConfigurationStep))
         .addStep(NonBlockingSteps.get(buildImageStep))
-        .whenAllSucceed(this::afterPushSteps)
-        .get()
-        .get()
-        .get();
+        .whenAllSucceedBlocking(this::afterPushSteps);
```
loosebazooka: I'm a little lost as to how this triple get() is satisfied by the single get() in whenAllSucceedBlocking

loosebazooka:
I would wait for @coollog to comment, since most of this async stuff was written by him.  I can't remember the original intention of nesting the futures.

jib-core/src/main/java/com/google/cloud/tools/jib/async/AsyncDependencies.java
```diff
@@ -80,4 +82,19 @@ public AsyncDependencies addSteps(List<? extends AsyncStep<?>> asyncSteps) {
...
+   * @param combiner the {@link Callable}
+   * @param <C> the return type of {@code combiner}
+   * @return the result returned from calling {@code combiner}
+   * @throws ExecutionException if {@code combiner} threw an exception
+   * @throws InterruptedException if the current thread was interrupted while waiting for {@code
+   *     combiner}
+   */
+  public <C> C whenAllSucceedBlocking(Callable<C> combiner)
+      throws InterruptedException, ExecutionException {
+    return Futures.whenAllSucceed(futures).call(combiner, listeningExecutorService).get();
```
coollog: I think this looks fine, but this looks like it's just `whenAllSucceed` with a call to `.get`. I'm wondering if we can leave off the extra method and just have the `.get` explicitly in the individual functions so we get rid of the nested `ListenableFuture`s but don't need this extra method?

PR #1380, opened 2018-12-26
Build projects with JDK 11 (targeting Java 11 bytecode)

chanseokoh:
Not for review or something we'd want to do soon. This is manly for testing various things with Java 11.

For now, putting up here as a reference to document what is necessary to build all projects with JDK 11.


chanseokoh:
I see the following warning when running tests on jib-gradle-plugin. (I wonder if this might be due to our examples using Gradle 4.10 wrapper. Haven't checked it.)

```
> Task :test
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access using Lookup on org.gradle.internal.classloader.ClassLoaderUtils$AbstractClassLoaderLookuper (file:/usr/local/google/home/chanseok/.gradle/caches/5.0/generated-gradle-jars/gradle-api-5.0.jar) to class java.lang.ClassLoader
WARNING: Please consider reporting this to the maintainers of org.gradle.internal.classloader.ClassLoaderUtils$AbstractClassLoaderLookuper
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

chanseokoh:
Everything builds successfully with JDK 11. Travis builds on JDK 11 are passing too. (Not yet with Kokoro.)

jib-core/build.gradle
```diff
@@ -13,8 +13,8 @@ plugins {
 
 group 'com.google.cloud.tools'
 
-sourceCompatibility = JavaVersion.VERSION_1_8
-targetCompatibility = JavaVersion.VERSION_1_8
+sourceCompatibility = JavaVersion.VERSION_11
+targetCompatibility = JavaVersion.VERSION_11
```
briandealwis: Um… won't this result in using Java 11 bytecodes and .class file markers, preventing its use on Java 8?

jib-maven-plugin/pom.xml
```diff
@@ -43,9 +43,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <!-- Use <maven.compiler.release> instead when migrating to JDK 11. -->
-    <maven.compiler.source>8</maven.compiler.source>
-    <maven.compiler.target>8</maven.compiler.target>
+    <maven.compiler.release>11</maven.compiler.release>
```
briandealwis: This prevents use of the plugin on Java 8

jib-core/build.gradle
chanseokoh: Yeah, this is for reference only and be merged probably a few years later.

PR #1382, opened 2018-12-26
Fix javadoc 'see'

### LINKED ISSUE #1381: Fix "@see" in javadocs
Looks like there are a few places where we do something along the lines of:
```
/**
 * @see #methodName() for more details
 */
```
which generates:
> **See Also:**
>
> [for more details]()

[Original comment](https://github.com/GoogleContainerTools/jib/pull/1377#discussion_r244052548)

TadCordle:
Fixes #1381

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistryImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistryImage.java
@@ -89,7 +89,8 @@ public class RegistryImage implements SourceImage, TargetImage {
   /**
    * Adds {@link CredentialRetriever} to fetch push/pull credentials for the image. Credential
    * retrievers are attempted in the order in which they are specified until credentials are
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistryImage.java
```diff
@@ -110,7 +110,7 @@ public RegistryImage addCredential(String username, String password) {
    *
    * @param credentialRetriever the {@link CredentialRetriever} to add
    * @return this
-   * @see CredentialRetrieverFactory for useful pre-defined {@link CredentialRetriever}s
+   * @see CredentialRetrieverFactory
```
chanseokoh: This one seems worth adding the text to the main Javadoc too, like `See {@link CredentialRetrieverFactory} for useful pre-defined {@link CredentialRetriever}s`.

Issue #1201, opened 2018-10-29
Check Maven settings and pom properties as well as system properties in maven config


TadCordle:
[Original comment](https://github.com/GoogleContainerTools/jib/pull/1195#discussion_r228575774)

PR #1384, opened 2018-12-27
 Add aliases for nested maven parameters

### LINKED ISSUE #382: Add aliases for Maven configuration.
From https://github.com/GoogleContainerTools/jib/issues/345#issuecomment-394828180

First, confirm if this actually enables IntelliJ quickdocs for nested configuration aliases.

TadCordle:
Fixes #382.

For example, the following:
```xml
<configuration>
  <to>
    <image>target-image</image>
  </to>
</configuration>
```

can now be written as:
```xml
<configuration>
  <to.image>target-image</to.image>
</configuration>
```


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -234,6 +240,102 @@ public File getPath() {
 
   @Nullable @Component protected SettingsDecrypter settingsDecrypter;
 
+  // Aliases
+  @Nullable
+  @Parameter(alias = "from.image")
+  private String fromImageAlias;
```
loosebazooka: I thought this makes this value accessible via `<fromImageAlias>`? Is that not true? 

If that's the case, then could the aliases be directly assigned to the original values?

Maven's behavior here isn't super clear to me.

TadCordle: Oh yeah, it might be `name` I need to set, not `alias`.

Assigning the aliases to the original values is tricky in our case, since we have several repurposed classes (like to vs. from `AuthConfiguration`). Also I'm not even sure if you can set aliases on nested parameters, I remember trying that with properties a while ago and it wasn't possible, which is why we had to do all the `System.getProperty` workarounds.

loosebazooka: Yeah, it's not clear whether maven supports this at all.

Would be worth emailing the maven developer mailing list for ideas: https://maven.apache.org/mailing-lists.html

I don't think the JSR330 and Sisu stuff works really either for this case. Although it might, might be worth asking them about that too.

TadCordle: I'll try the mailing list. If that doesn't work, there are a couple alternatives I can think of.
- Instead of using the `from.image` style alias, we can use camel-case `fromImage` and just use the field name if we can't disable it.
- Or, maybe it's ok to allow both the field name and the alias, but only document the alias. But I'm not sure if we want to risk the complexity/maintenance burden if people still find out about it and start using it anyway.

loosebazooka: Yeah my concern is 2. We're only helping IDEs out here by populating some alternative options in our plugin, but it still wont allow them to autocomplete the normal config.

Personally I don't think it's worth it to make our config so complex, it's a problem with maven that appears to be unsolvable at the moment (unless we get some update on the mailing list).

loosebazooka: I wonder if #1201 can be solved by getting properties from the MavenProject component?
I don't know what the behavior of it is, but it might be worth checking out.

In a mojo
```java
@Component //should be injected by the system
MavenProject mavenProject;

// ...in some method somewhere to access properties, instead of system.getProperty
mavenProject.getProperties()
```

TadCordle:
Closing this; no response on the mailing list, plus the added complexity isn't worth it.

loosebazooka:
Can we still use mavenProject properties to solve the other problem?

TadCordle:
Yup, just put up another PR.

PR #1385, opened 2018-12-28
Rename Maven profile ID

chanseokoh:
I am always confuse by the different names between Maven and Gradle. This will make them consistent.


chanseokoh:
Yeah, I see we took the profile approach, so we can't get away with `-P`.

PR #1386, opened 2018-12-28
Set up Travis build matrix for additional OpenJDK 11 / upgrade Error Prone, NullAway, Javassist

chanseokoh:
Setting up Travis build matrix to include additional build on OpenJDK 11. This required upgrading Error Prone and NullAway and tweaking configurations.

Note the OpenJDK 11 build still targets Java 8. It doesn't allow using Java 11 features in source code or generate Java 11 bytecode.


jib-gradle-plugin/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.0-bin.zip
```
loosebazooka: I think this enables us (developers) to potentially use 5.0 level APIs which then might mean we have to update our minimum gradle supported version in: `JibPlugin`

Can we keep a test that runs against `JibPlugin.GRADLE_MIN_VERSION` or just change that to `5.0` also. Changing it to minimum 5.0 might be annoying to users though.

chanseokoh: Good point. To my surprise, it is [currently not possible to specify the Gradle API dependency version](https://github.com/gradle/gradle/issues/1835).

I've restored the Gradle TestKit test to use the `JibPlugin.GRADLE_MIN_VERSION` (4.6). However, 4.6 just can't run on JDK 11, so I had to add `assumeTrue(Java 8)`. As long as we have a Java 8 build, we should be fine. The test was lacking applying the Jib plugin, so I also enhanced it to load the existing simple `build.gradle`:
```build.gradle
plugins {
    id 'java'
    id 'com.google.cloud.tools.jib'
}
```
Not sure if this will always catch any API usage greater than 4.6 though.

chanseokoh: BTW, although `GRADLE_MIN_VERSION` was 4.6, our build system was using 4.10-2, so we were developing against the 4.10-2 API anyway. I think it is not impossible that we already used API requiring >4.6.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurationsTest.java
```diff
@@ -79,6 +81,17 @@ public String getDisplayName() {
     public Set<File> getFiles() {
       return files;
     }
+
+    @Override
+    public TaskDependency getBuildDependencies() {
```
loosebazooka: I thought this was a little strange, but there isn't really an easy way to Mock this out :disappointed: 
No action required on this comment.

jib-gradle-plugin/gradle/wrapper/gradle-wrapper.properties
chanseokoh: Seems like we don't need Gradle 5.0 if not targeting 11. Restored back to 4.10.2.

PR #1389, opened 2019-01-03
Check for system properties in pom as well as commandline

TadCordle:
Part of #1201. Auth is a special case, so we'll have to figure out how to handle that next.

Rather than using `System.getProperty()` for the system properties, this PR gets the system properties via the Maven project and/or session.

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
@@ -249,7 +249,10 @@ public abstract class JibPluginConfiguration extends AbstractMojo {
   @Nullable
   String getBaseImage() {
     String property = getProperty(PropertyNames.FROM_IMAGE);
-    return property == null ? Preconditions.checkNotNull(from).image : property;
+    if (property != null) {
+      return property;
+    }
+    return Preconditions.checkNotNull(from).image;
   }
 
   /**
@@ -260,12 +263,14 @@ public abstract class JibPluginConfiguration extends AbstractMojo {
   @Nullable
   String getBaseImageCredentialHelperName() {
     String property = getProperty(PropertyNames.FROM_CRED_HELPER);
-    return property == null ? Preconditions.checkNotNull(from).credHelper : property;
+    if (property != null) {
+      return property;
+    }
+    return Preconditions.checkNotNull(from).credHelper;
   }
 
   AuthConfiguration getBaseImageAuth() {
-    // System properties are handled in ConfigurationPropertyValidator
-    // TODO: handle properties here, not in ConfigurationPropertyValidator
+    // TODO: handle properties here instead of in ConfigurationPropertyValidator
     return from.auth;
   }
 
@@ -276,9 +281,15 @@ public abstract class JibPluginConfiguration extends AbstractMojo {
    */
   @Nullable
   String getTargetImage() {
-    String property = getProperty(PropertyNames.TO_IMAGE);
     String propertyAlternate = getProperty(PropertyNames.TO_IMAGE_ALTERNATE);
-    return propertyAlternate == null ? (property == null ? to.image : property) : propertyAlternate;
+    if (propertyAlternate != null) {
+      return propertyAlternate;
+    }
+    String property = getProperty(PropertyNames.TO_IMAGE);
+    if (property != null) {
+      return property;
+    }
+    return to.image;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -262,14 +259,13 @@ String getBaseImage() {
...
-      return System.getProperty(PropertyNames.FROM_CRED_HELPER);
-    }
-    return Preconditions.checkNotNull(from).credHelper;
+    String property = getProperty(PropertyNames.FROM_CRED_HELPER);
+    return property == null ? Preconditions.checkNotNull(from).credHelper : property;
   }
 
   AuthConfiguration getBaseImageAuth() {
     // System properties are handled in ConfigurationPropertyValidator
+    // TODO: handle properties here, not in ConfigurationPropertyValidator
```
loosebazooka: This new comment seems to be in conflict with the previous.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -496,12 +465,12 @@ String getFormat() {
...
   Optional<Path> getExtraDirectoryPath() {
     // TODO: Should inform user about nonexistent directory if using custom directory.
-    if (System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH) != null) {
-      return Optional.of(Paths.get(System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH)));
-    }
-    return extraDirectory.path == null
-        ? Optional.empty()
-        : Optional.of(extraDirectory.path.toPath());
+    String property = getProperty(PropertyNames.EXTRA_DIRECTORY_PATH);
+    return property == null
```
loosebazooka: Might consider rewriting this as `if/else` for readability.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/JibPluginConfigurationTest.java
```diff
@@ -22,47 +22,43 @@
...
-    System.clearProperty("jib.container.jvmFlags");
-    System.clearProperty("jib.container.labels");
-    System.clearProperty("jib.container.mainClass");
-    System.clearProperty("jib.container.ports");
-    System.clearProperty("jib.container.useCurrentTimestamp");
-    System.clearProperty("jib.container.user");
-    System.clearProperty("jib.container.workingDirectory");
-    System.clearProperty("jib.extraDirectory.path");
-    System.clearProperty("jib.extraDirectory.permissions");
+  private void clearProperties() {
```
loosebazooka: I don't think we need this method anymore? `project` and `sessionProperties` should be new on every test that is run in this class.

loosebazooka:
Also, what are your thoughts on adding a helper to remove all the boiler plate? Perhaps with the signatures
```
// simple
public static <T> T getPropertyOrValue(String property, Function<T, String> converter, T configValue)
// for something that is nested (supplier to evaluate later)
public static <T> T getPropertyOrValue(String property, Function<T, String> converter, Supplier<T> configValue)
// for stuff with alternate properties
public static <T> T getPropertyOrValue(String property, String altProperty, Function<T, String> converter, T configValue)
// for stuff with alternate properties and a weird value (supplier to evaluate later)
public static <T> T getPropertyOrValue(String property, String altProperty, Function<T, String> converter, Supplier<T> configValue)
```
There might be so many cases that it's not worth it, but just curious if you think it could help?

TadCordle:
I started out trying something like this, but it did seem like too many cases for it to be worth it (especially for the convert/supplier methods; it was like I was still rewriting 70% of the method within the method's argument list). I think the boilerplate is short enough that it's not that big of a deal.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
loosebazooka: Sorry, I only meant one instance, but I think it's fine. Feel free to change the others back to `? :` if you liked that.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/JibPluginConfigurationTest.java
```diff
@@ -40,18 +39,14 @@
 
   private static final MavenProject project = new MavenProject();
```
loosebazooka: Oh didn't see that it was a static MavenProject before. And I feel like this might be bad practice since now we're responsible for cleaning it up between each test. If we don't have a great reason to keep this static, could we change it to just `private final MavenProject` ?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/JibPluginConfigurationTest.java
```diff
@@ -40,18 +39,14 @@
...
   private JibPluginConfiguration testPluginConfiguration;
 
-  private void clearProperties() {
-    sessionProperties.clear();
-    project.getProperties().clear();
-  }
-
   @Before
   public void setup() {
+    project.getProperties().clear();
```
loosebazooka: not using a static project would also allow us to ignore cleaning up the state here.

PR #1391, opened 2019-01-04
Get maven auth system properties from pom and commandline

### LINKED ISSUE #1201: Check Maven settings and pom properties as well as system properties in maven config

TadCordle:
Follow-up to #1389, fixes #1201. Needed a bit of special work for this since auth properties are a special case and need to be defined together.

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
@@ -141,4 +141,9 @@ class MavenRawConfiguration implements RawConfiguration {
   public ImageFormat getImageFormat() {
     return ImageFormat.valueOf(jibPluginConfiguration.getFormat());
   }
+
+  @Override
+  public Optional<String> getProperty(String propertyName) {
+    return Optional.ofNullable(jibPluginConfiguration.getProperty(propertyName));
+  }
 }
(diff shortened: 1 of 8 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
```diff
@@ -141,4 +141,9 @@ public boolean getAllowInsecureRegistries() {
   public ImageFormat getImageFormat() {
     return ImageFormat.valueOf(jibPluginConfiguration.getFormat());
   }
+
+  @Override
+  public Optional<String> getProperty(String propertyName) {
```
loosebazooka: I'm not sure I follow why we need two different approaches to configuring values with properties. Could we potentially just use this for all config? Why is it only used for Auth?

And maybe I missed the comment on why Auth couldn't be done like in #1389

chanseokoh: AFAIKT, this is another instance of the problem related to validation. For example, following the apporach in #1389, `JibPluginConfiguration.getUsername/getPassword()` may give you the final, effective username and password by looking at all of the system properties, `pom.xml` properties, and the Jib config values. However, then, when it come to validation, callers don't know where the values came, so `ConfigurationPropertyValidator` can't check, e.g., the username is given by the system properties but the password is missing in the system properties. And in that case, `ConfigurationPropertyValidator` will ignore the username from the system properties and try the username/password from the Jib config values.

TadCordle: Yeah, I think we should keep this as is right now, and reorganize the validation stuff later if we agree it's too decentralized. I agree it's sort of messy, but I think we can finish up this functionality and then refactor later.

PR #1398, opened 2019-01-08
Add dockerClient configuration to READMEs

### LINKED ISSUE #1394: Add docker executable/env configuration to READMEs or FAQ
#1214 and #1393 are undocumented at the moment.

TadCordle:
Fixes #1394. To be merged after next RC.

### DIFF
--- a/jib-gradle-plugin/README.md
+++ b/jib-gradle-plugin/README.md
@@ -237,7 +237,7 @@ gradle jib \
     -Djib.to.auth.password=$PASSWORD
 
 gradle jibDockerBuild \
-    -Djib.dockerClient.executable=/path/to/docker/ \
+    -Djib.dockerClient.executable=/path/to/docker \
     -Djib.container.environment=key1="value1",key2="value2" \
     -Djib.container.args=arg1,arg2,arg3
 ```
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/README.md
```diff
@@ -128,6 +128,8 @@ eval $(minikube docker-env)
 gradle jibDockerBuild
 ```
 
+See [`dockerClient`](#dockerclient-closure) for more configuration options.
```
chanseokoh: The doc was talking about doing `eval $(minikube docker-env)`, for which you have an alternative to set `dockerClient.environment`. What I mean is, it may be confusing to users if there are two different ways to achieve the same goal, so I think it's good to clear any potential confusion.

What if we add "Alternatively, you can set environment variables in the Jib configuration. See ... "

jib-maven-plugin/README.md
```diff
@@ -153,6 +153,8 @@ eval $(minikube docker-env)
 mvn compile jib:dockerBuild
 ```
 
+See [`dockerClient`](#dockerclient-object) for more configuration options.
```
chanseokoh: ditto

PR #1415, opened 2019-01-14
Upgrade Google HTTP Client dependency

chanseokoh:
Not really necessary, but the latest 1.27.0 defines `Automatic-Module-Name`.

### DIFF
--- a/jib-core/build.gradle
+++ b/jib-core/build.gradle
@@ -35,7 +35,7 @@ configurations {
 
 dependencies {
   // Make sure these are consistent with jib-maven-plugin.
-  implementation 'com.google.http-client:google-http-client:1.23.0'
+  implementation 'com.google.http-client:google-http-client:1.27.0'
   implementation 'org.apache.commons:commons-compress:1.18'
   implementation 'com.google.guava:guava:23.5-jre'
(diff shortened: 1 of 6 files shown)


chanseokoh:
Turns out 1.27.0 throws `SSHHandshakeException` when attempting secure TLS connection to an insecure server.

chanseokoh:
Now it works. Catching the general `SSLException` makes sense, and this will actually make Jib more stable with respect to `allowInsecureRegistries`.

I noticed this upgrade causes many deprecated warnings, which I will resolve in the next PR.

chanseokoh:
I will hold off merging this until we make a 1.0 release.

chanseokoh:
`https://github.com/GoogleContainerTools/jib/compare/remove-deprecated?expand=1` will remove the deprecated usage introduced by upgrading the library to 1.27.0. However, that doesn't work at runtime due to a [bug in Google HTTP Client](https://github.com/googleapis/google-http-java-client/issues/503), which is fixed just 12 days ago. The latest 1.28.0 release is supposed to fix it, but then [1.28.0 is missing `com.google.api.client.http.apache`](https://github.com/googleapis/google-http-java-client/issues/576), so we will be stuck with the deprecated usage at the moment.

PR #1417, opened 2019-01-15
Upgrade micronaut example

sdelamo:
Close: https://github.com/GoogleContainerTools/jib/issues/1416


coollog:
Hi @sdelamo thanks for the contribution! We originally had `Application.java` to show that Java and Groovy can be inter-used in the example - was there a reason for changing it to `Application.groovy` instead?

sdelamo:
@coollog I replaced Application.java with Groovy because if you used the --lang=groovy that it is the file that would be generated. If you want I can contribute the same sample in Java or Kotlin.

chanseokoh:
I think this is good enough.

PR #1424, opened 2019-01-16
Rename test resource directories to avoid overlaps between plugins and core

### LINKED ISSUE #1349: Test resource directories overlaps between plugins and jib-core
In jib-core, we have a `resources/application` directory. In the plugins, we also have a `resources/application` directory. Travis doesn't seem to like this; I ran into an issue with https://github.com/GoogleContainerTools/jib/pull/1347 where jib-core tests pass by themselves, but fail when they're re-run after building the plugins/plugins-common, because it seems to use the plugins' `resources/application` directory instead of jib-core's.

Example:
- [jib-core passes](https://travis-ci.org/GoogleContainerTools/jib/builds/468172271#L518)
- [but plugins-common fails due to additional files being added](https://travis-ci.org/GoogleContainerTools/jib/builds/468172271#L545)

I avoided this issue in #1347 because of unrelated code changes, but it might make sense to fix this.

TadCordle:
- Moves resources to extra parent folder indicating which module it is part of (e.g. `resources/folder` becomes `resources/core/folder` in jib-core, `resources/plugins-common/folder` in jib-plugins-common, `resources/gradle/folder` in jib-gradle-plugin, and `resources/maven/folder` in jib-maven-plugin)
- `resources/mockito-extensions` was not moved
- Removes unused `resources/layer` directory in maven plugin

Fixes #1349


TadCordle:
Actually, I think I'm going to move everything under an extra parent folder instead of just renaming some directories, so it's easier to prevent the problem from happening again.

e.g. instead of `resources/application-core`, `resources/projects-gradle`, etc. all of the core resources will be under `resources/core/...`, all the gradle resources will be under `resources/gradle/...`, etc.

PR #1444, opened 2019-01-23
Fix progress display tests ignoring locale

### LINKED ISSUE #1442: Build failure on FR environment
**Description of the issue**:
Master branch build fails on FR environment
Test class ProgressDisplayGeneratorTest fails

**Expected behavior**:
Build should not fail

**Steps to reproduce**:
Adding jvm args seems enough to reproduce the issue on en_US.UTF-8 environment
`(cd jib-plugins-common/ && ./gradlew clean build  -Dorg.gradle.jvmargs=-Duser.language=fr)`

**Environment**: 
Original env 
```
$ localectl status
   System Locale: LANG=fr_FR.UTF-8
```

**Log output**:
`java.lang.AssertionError: expected:<[Executing tasks:, [===============               ] 50.0% complete]> but was:<[Executing tasks:, [===============               ] 50,0% complete]>` (dot vs comma)

TadCordle:
Fixes #1442

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/ProgressDisplayGenerator.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/ProgressDisplayGenerator.java
@@ -17,11 +17,9 @@
 package com.google.cloud.tools.jib.plugins.common.logging;
 
 import com.google.cloud.tools.jib.event.progress.Allocation;
-import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


TadCordle:
@chanseokoh It seemed like it was either, touch the non-test code to inject a `Locale`, or calculate the expected string in the test using the default locale. But that meant adding extra logic to the tests that could possibly fail, and the tests checking for different things when it's run on different platforms, which seemed weird to me. But maybe it doesn't matter that much?

chanseokoh:
If it's simple like

```java
assertEquals(
   String.format("Executing tasks:", "[===============               ] %.1f% complete", 50),
   runNonTestCode());
```

I think it's worth doing so. Basically running the same non-test code.

Otherwise, yeah, I am fine with changing the non-test code.

TadCordle:
Ok, I'll change it.

Issue #1054, opened 2018-09-27
Create new events that contain structured information for users of Jib Core for various parts of the build execution.


coollog:
Follow-up from #714

PR #1449, opened 2019-01-25
Add enum for matching progress events to build steps

TadCordle:
Part of #1054. This PR adds a `BuildStepType` enum, which is passed to progress events so that jib-core users may know which progress events correspond to which steps in the build.

I'm still not totally familiar with the progress system, so let me know if this seems like an ok direction to take.

Possible followups:
- Add documentation for what each of the `BuildStepType`s mean
- Add an event for counting the number of layers involved in the push/pull steps, or modify/add progress events that make it possible to identify which layer a progress event was sent from

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildStepType.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildStepType.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2019 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.builder;
+
+/** Types corresponding to steps in the containerization process. */
+public enum BuildStepType {
+  AuthenticatePush,
+  BuildAndCacheApplicationLayer,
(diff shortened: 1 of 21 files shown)


TadCordle:
Actually hang on, going to add some tests.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/BuildStepType.java
```diff
@@ -0,0 +1,33 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.builder;
+
+/** Types corresponding to steps in the containerization process. */
+public enum BuildStepType {
+  AUTHENTICATE_PUSH,
```
coollog: Since this will be user-facing, let's add a short description to each one of these.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
```diff
@@ -61,7 +64,7 @@
   public static ProgressEventDispatcher newRoot(
       EventDispatcher eventDispatcher, String description, long allocationUnits) {
     return newProgressEventDispatcher(
-        eventDispatcher, Allocation.newRoot(description, allocationUnits));
+        eventDispatcher, Allocation.newRoot(description, allocationUnits), null);
```
coollog: maybe we should just not have a `Nullable` by adding a `BuildStepType` for the root, maybe called `ALL` or `COMPLETION` or something?

jib-core/src/test/java/com/google/cloud/tools/jib/event/events/ProgressEventTest.java
```diff
@@ -117,6 +118,22 @@ public void testSmoke() {
     Assert.assertEquals(2, allocationCompletionMap.get(AllocationTree.root).longValue());
   }
 
+  @Test
+  public void testType() {
```
coollog: Note that this test would also need to check if the consumers were actually called.

PR #1457, opened 2019-01-30
Fix main class finder missing main defined with varargs

### LINKED ISSUE #1456: MainClassFinder misses "public static void main(String... args)"
`public static void main(String... args)` is apparently a valid signature for the main method, but Jib doesn't find it.

TadCordle:
Fixes #1456


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/MainClassFinder.java
```diff
@@ -115,7 +115,8 @@ public Type getType() {
         Type.getMethodDescriptor(Type.VOID_TYPE, Type.getType(String[].class));
 
     /** Accessors that main may or may not have. */
-    private static final int OPTIONAL_ACCESS = Opcodes.ACC_FINAL | Opcodes.ACC_DEPRECATED;
+    private static final int OPTIONAL_ACCESS =
+        Opcodes.ACC_FINAL | Opcodes.ACC_DEPRECATED | Opcodes.ACC_VARARGS;
```
chanseokoh: Huh, it's weird that var args fall in the category of method access control.

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/MainClassFinderTest.java
```diff
@@ -126,4 +126,15 @@ public void testFindMainClass_innerClasses() throws URISyntaxException, IOExcept
...
         CoreMatchers.containsString("HelloWorld$InnerClass"));
   }
+
+  @Test
+  public void testMainClass_varargs() throws URISyntaxException, IOException {
+    Path rootDirectory =
+        Paths.get(Resources.getResource("core/class-finder-tests/varargs").toURI());
+    MainClassFinder.Result mainClassFinderResult =
+        new MainClassFinder(new DirectoryWalker(rootDirectory).walk(), mockEventDispatcher).find();
+    Assert.assertSame(Type.MAIN_CLASS_FOUND, mainClassFinderResult.getType());
```
chanseokoh: Can't `assertEquals` work?

TadCordle: It can, but we use "same" in most of the other `MainClassFinder` tests (`assertSame` tests using `==`, `assertEquals` tests using `.equals()`)

PR #1465, opened 2019-02-06
Add event system explanation to jib-core readme

### LINKED ISSUE #1462: Add detailed documentation for the jib-core events

TadCordle:
Fixes #1462. Should be merged after the next core release (0.1.2? 0.2.0?)

### DIFF
--- a/jib-core/README.md
+++ b/jib-core/README.md
@@ -22,7 +22,7 @@ Add Jib Core as a dependency using Maven:
 <dependency>
   <groupId>com.google.cloud.tools</groupId>
   <artifactId>jib-core</artifactId>
-  <version>0.1.1</version>
+  <version>0.1.2</version>
 </dependency>
 ```
 
@@ -30,7 +30,7 @@ Add Jib Core as a dependency using Gradle:
 
 ```groovy
 dependencies {
-  compile 'com.google.cloud.tools:jib-core:0.1.1'
+  compile 'com.google.cloud.tools:jib-core:0.1.2'
 }
 ```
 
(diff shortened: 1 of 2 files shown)


jib-core/README.md
```diff
@@ -98,17 +98,42 @@ The Jib Core system consists 3 main parts:
...
+// Register with Jib
+Jib.from(...)
+    ...
+    .containerize(
+        Containerizer.to(...)
+            ...
+            .setEventHandlers(eventHandlers));
+```
+
+When Jib dispatches events, the event handlers you defined for that event type will be called. The following are the types of events you can listen for in Jib core (see [API reference](http://static.javadoc.io/com.google.cloud.tools/jib-core/0.1.1/com/google/cloud/tools/jib/api/package-summary.html) for more information):
```
loosebazooka: can this reference a *latest* instead of 0.1.1? Seems like we could possibly miss updating this when doing releases?

loosebazooka: Looks like the same thing through the readme.

loosebazooka: There appears to be some provision for it at: https://www.javadoc.io/, but I don't know how we're generating these in the first place.

loosebazooka: okay, so the url just has to be formatted a little weird

```diff
- http://static.javadoc.io/com.google.cloud.tools/jib-core/0.1.1/com/google/cloud/tools/jib/api/package-summary.html
+ https://www.javadoc.io/page/com.google.cloud.tools/jib-core/latest/com/google/cloud/tools/jib/api/package-summary.html
```

TadCordle: Sounds like a good idea to me

TadCordle:
Since we're releasing 0.1.2, we can review this @GoogleContainerTools/java-tools-build

PR #1466, opened 2019-02-07
Rethrow ExecutionException's cause in jib-core

### LINKED ISSUE #1440: Organize jib-core exceptions
At the moment, we wrap most of jib-core's errors and rethrow as `BuildStepsExecutionException`s with custom messages. For library users, it's somewhat vague which exceptions are thrown (since many are blanketed by `ExecutionException`) and why each exception is thrown, so it would be useful to either document this better, or somehow provide an easier way for more explicit exception handling.

TadCordle:
Fixes #1440 

Originally, it was up to the library user to catch an `ExecutionException`, and determine what kind of execution caused it using a series of `if (ex instanceof ...)` statements for more fine-grained exception handling. With this PR, `ExecutionException`s' causes are re-thrown if they are `RegistryException`s, which makes jib-core error handling/documentation much more straightforward.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
@@ -242,7 +242,11 @@ public class BuildStepsRunner {
             new RegistryUnauthorizedException(
                 ex.getServerUrl(), ex.getImageName(), (HttpResponseException) ex.getCause()),
             helpfulSuggestions);
+      } else {
+        // Unknown cause
+        throw new BuildStepsExecutionException(helpfulSuggestions.none(), ex);
       }
+
     } catch (UnknownHostException ex) {
       throw new BuildStepsExecutionException(helpfulSuggestions.forUnknownHost(), ex);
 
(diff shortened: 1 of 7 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
```diff
@@ -226,53 +226,36 @@ public JibContainer build(
...
-            helpfulSuggestions.forCredentialsNotSent(), exceptionDuringBuildSteps);
+    } catch (RegistryCredentialsNotSentException ex) {
+      throw new BuildStepsExecutionException(helpfulSuggestions.forCredentialsNotSent(), ex);
 
-      } else if (exceptionDuringBuildSteps instanceof RegistryAuthenticationFailedException
-          && exceptionDuringBuildSteps.getCause() instanceof HttpResponseException) {
-        RegistryAuthenticationFailedException failureException =
-            (RegistryAuthenticationFailedException) exceptionDuringBuildSteps;
+    } catch (RegistryAuthenticationFailedException ex) {
+      if (ex.getCause() instanceof HttpResponseException) {
```
briandealwis: What happens if the cause isn't a `HttpResponseException`?

TadCordle: Ah right, I'll change it to match previous behavior.

PR #1473, opened 2019-02-11
Mentions sbt-jib and AutoJib in Jib Core examples.

### LINKED ISSUE #1467: Add autojib as a Jib Core example.
Source: https://github.com/coollog/autojib

coollog:
Fixes #1467.

### DIFF
--- a/jib-core/examples/README.md
+++ b/jib-core/examples/README.md
@@ -4,6 +4,10 @@ Please [file an issue](/../../issues/new) if you find any problems with the exam
 
 For examples on using Jib plugins, see [examples](../../examples).
 
+### [sbt](https://www.scala-sbt.org/) plugin
+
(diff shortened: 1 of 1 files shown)


jib-core/examples/README.md
```diff
@@ -4,6 +4,14 @@ Please [file an issue](/../../issues/new) if you find any problems with the exam
 
 For examples on using Jib plugins, see [examples](../../examples).
 
+### [sbt](https://www.scala-sbt.org/) plugin
```
TadCordle: We should probably link directly to the source somewhere since it's intended to be an example for jib-core.

coollog: Switched up the links a bit so that the project name links to the GitHub project.

PR #1475, opened 2019-02-12
Add setAppRoot() and fromDistrolessJetty() to JavaContainerBuilder

### LINKED ISSUE #1464: Allow building WAR containers with JavaContainerBuilder
Currently there is no way of setting the app root in `JavaContainerBuilder`, so it is can't be used for WAR containers, where the default app root is in "/jetty/webapps/ROOT".

There are a couple possible ways we could do this:
- Add `JavaContainerBuilder#setAppRoot()`
  - This is the most flexible, but probably not very user friendly due to this class trying to abstract away the internals of the container being built
- Add more initializers that automatically set the approot
  - In addition to `fromDistroless()`, we could add a `fromDistrolessJetty()` method that sets the app root automatically.
  - In addition to `from()`, we could add `warFrom()` that takes an image reference and an app root. But again, this may not be very user friendly.

@GoogleContainerTools/java-tools thoughts?

TadCordle:
Fixes #1464.

Adds `JavaContainerBuilder#setAppRoot()` to support building WARs using `JavaContainerBuilder`. Also adds `JavaContainerBuilder#fromDistrolessJetty()` that uses `gcr.io/distroless/java/jetty` as the base image and `/jetty/webapps/ROOT` as the app root.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
@@ -46,11 +46,14 @@ public class JavaContainerBuilder {
    * href="https://github.com/GoogleContainerTools/distroless">the distroless repository</a>.
    *
    * @return a new {@link JavaContainerBuilder}
-   * @throws InvalidImageReferenceException if creating the base image reference fails
    * @see <a href="https://github.com/GoogleContainerTools/distroless">The distroless repository</a>
    */
-  public static JavaContainerBuilder fromDistroless() throws InvalidImageReferenceException {
-    return from(RegistryImage.named("gcr.io/distroless/java"));
+  public static JavaContainerBuilder fromDistroless() {
+    try {
+      return from(RegistryImage.named("gcr.io/distroless/java"));
+    } catch (InvalidImageReferenceException ignored) {
+      throw new IllegalStateException("Unreachable");
+    }
   }
 
   /**
@@ -59,12 +62,15 @@ public class JavaContainerBuilder {
    * href="https://github.com/GoogleContainerTools/distroless">the distroless repository</a>.
    *
    * @return a new {@link JavaContainerBuilder}
-   * @throws InvalidImageReferenceException if creating the base image reference fails
    * @see <a href="https://github.com/GoogleContainerTools/distroless">The distroless repository</a>
    */
-  public static JavaContainerBuilder fromDistrolessJetty() throws InvalidImageReferenceException {
-    return from(RegistryImage.named("gcr.io/distroless/java/jetty"))
-        .setAppRoot(AbsoluteUnixPath.get(JavaLayerConfigurations.DEFAULT_WEB_APP_ROOT));
+  public static JavaContainerBuilder fromDistrolessJetty() {
+    try {
+      return from(RegistryImage.named("gcr.io/distroless/java/jetty"))
+          .setAppRoot(AbsoluteUnixPath.get(JavaLayerConfigurations.DEFAULT_WEB_APP_ROOT));
+    } catch (InvalidImageReferenceException ignored) {
+      throw new IllegalStateException("Unreachable");
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -74,6 +53,20 @@ public static JavaContainerBuilder fromDistroless() throws InvalidImageReference
     return from(RegistryImage.named("gcr.io/distroless/java"));
   }
 
+  /**
+   * Creates a new {@link JavaContainerBuilder} that uses distroless jetty as the base image. For
+   * more information on {@code gcr.io/distroless/java}, see <a
+   * href="https://github.com/GoogleContainerTools/distroless">the distroless repository</a>.
+   *
+   * @return a new {@link JavaContainerBuilder}
+   * @throws InvalidImageReferenceException if creating the base image reference fails
```
chanseokoh: Maybe we can shallow this exception. It will never be thrown, because `gcr.io/distroless/java/jetty` is always a valid reference.

jib-core/src/test/java/com/google/cloud/tools/jib/api/JavaContainerBuilderTest.java
```diff
@@ -192,6 +193,21 @@ public void testToJibContainerBuilder_missingAndMultipleAdds()
...
   }
 
+  @Test
+  public void testToJibContainerBuilder_setAppRootLate()
+      throws URISyntaxException, InvalidImageReferenceException, IOException {
+    try {
+      JavaContainerBuilder.fromDistroless()
+          .addClasses(getResource("core/application/classes"))
+          .setAppRoot("/oh no")
+          .toContainerBuilder();
```
chanseokoh: It's not clear who's throwing the exception. And `.toContainerBuilder()` isn't necessary, I think. Do

```java
try {
   containerBuilder.setAppRoot("/oh no");
```

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -113,10 +106,64 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
...
+  }
+
+  /**
+   * Sets the app root of the container image (useful for building WAR containers).
+   *
+   * @param appRoot the absolute path of the app on the container ({@code /app} by default)
+   * @return this
+   */
+  public JavaContainerBuilder setAppRoot(AbsoluteUnixPath appRoot) {
+    if (!classpath.isEmpty()) {
```
chanseokoh: Looking at the exception message "App root must be set before adding files", I don't really get this check. We are given `appRoot` by the parameter and setting this now and this methods doesn't add files.

TadCordle: Do you mean the exception message is unclear, or you don't understand why we're doing the check in the first place? The check is there because right now, setting the app root doesn't change the app root for previously added files. e.g. if you add class files, change app root, then add dependencies, now the dependencies are in a different root directory than the classes that were added earlier. I added the check so people don't assume calling `setAppRoot` updates previously added layers.

coollog: Hmm, we might want to make this declarative and be able to allow the user to set the app root whenever, including after adding the layers.

chanseokoh: Ah, now I get it. I think the error message is a bit confusing. Something like "you cannot set a new app root (or change the app root) after files are added" seems clearer.

TadCordle: @coollog I think I'll just file an issue for that since it seems like it'd involve changing a lot of internals of the class.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -113,10 +112,64 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
...
+      throw new IllegalStateException(
+          "You cannot set a new app root (or change the app root) after files are added");
```
chanseokoh: "You cannot change the app root after files are added" seems enough.

PR #1479, opened 2019-02-13
Adds Google Cloud Build example.

coollog:
Preview at https://github.com/GoogleContainerTools/jib/tree/cloud-build-example/examples/google-cloud-build

### DIFF
--- a/examples/google-cloud-build/README.md
+++ b/examples/google-cloud-build/README.md
@@ -1,0 +1,81 @@
+# Use [Google Cloud Build](https://cloud.google.com/cloud-build) with Jib
+
+This example shows how to use [Google Cloud Build](https://cloud.google.com/cloud-build) and Jib to build your Java project into a container image. 
+
+## Prerequisites
+
+Make sure you enable the Cloud Build API and have the Cloud SDK installed. The `gcloud` binary should be on your `PATH`.
+
+See the [Cloud Build Quickstart](https://cloud.google.com/cloud-build/docs/quickstart-docker#before-you-begin) for instructions.
+
+## Build a Maven project
+
+To build your Maven project into a container image, use [`cloudbuild-jib-maven.yaml`](cloudbuild-jib-maven.yaml) as the Cloud Build build configuration file:
+
+```bash
+gcloud builds submit --config cloudbuild-jib-maven.yaml <Maven project directory>
+```
+
+Make sure to change the `_IMAGE_NAME` substitution to the name of the image you wish to build.
+
+## Build a Gradle project
+
+To build your Gradle project into a container image, first add Jib to your `build.gradle`:
+
+```groovy
+plugins {
+  id 'com.google.cloud.tools.jib' version '1.0.0'
+}
+```
+
+Then, use [`cloudbuild-jib-gradle.yaml`](cloudbuild-jib-gradle.yaml) as the Cloud Build build configuration file:
+
+```bash
+gcloud builds submit --config cloudbuild-jib-gradle.yaml <Gradle project directory>
+```
+
+Make sure to change the `_IMAGE_NAME` substitution to the name of the image you wish to build.
+
+### Build a Gradle project without applying the Jib plugin
+
+You can also build any Gradle project without having to add Jib to the `build.gradle`. First, copy [`auto-jib.gradle`](auto-jib.gradle) to your Gradle project directory. Then, uncomment the `init-script` arg in [`cloudbuild-jib-gradle.yaml`](cloudbuild-jib-gradle-auto.yaml) and submit the build with:
+
+```bash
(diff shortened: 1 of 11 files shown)


examples/google-cloud-build/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+Uncomment the `init-script` arg in [`cloudbuild-jib-gradle.yaml`](cloudbuild-jib-gradle.yaml) and submit the build with:
+
+```bash
+gcloud builds submit --config cloudbuild-jib-gradle.yaml helloworld-sample
+```
+
+## Speeding up your build
+
+By default, Google Cloud Build does not cache anything across builds. However, you can [use Google Cloud Storage to cache files](https://cloud.google.com/cloud-build/docs/speeding-up-builds) across builds to speed up subsequent builds.
```
loosebazooka: Is this strictly true? I remember vaguely that GCS sometimes didn't speed anything up?

examples/google-cloud-build/cloudbuild-jib-gradle+cache.yaml
```diff
@@ -0,0 +1,69 @@
...
+# First, loads the cached files from GCS if they exist.
+- waitFor:
+  - '-'
+  name: gcr.io/cloud-builders/gsutil
+  dir: /builder/home
+  entrypoint: /bin/bash
+  args:
+  - -c
+  - |
+    mkdir /builder/home/.gradle &&
```
chanseokoh: Just to confirm: is everything under `/builder/home` persistent across all the steps?

examples/google-cloud-build/README.md
ImJasonH: GCS isn't guaranteed to speed anything up, it depends a lot on your use case, what you write back to GCS at the end of a build and how much you can reuse on the next build (and file sizes/counts, since writing 1000000 1B files to GCS is tons slower than one 1MB file)

Similarly, how you upload stuff to GCS and how you read it back can matter a lot to how concurrency-safe and incremental it is.

examples/google-cloud-build/cloudbuild-jib-gradle+cache.yaml
ImJasonH: Yes

examples/google-cloud-build/cloudbuild-jib-maven+cache.yaml
```diff
@@ -0,0 +1,66 @@
...
+      tar -xzf jib-cache.tar.gz
+    ) || echo 'Cache not found'
+
+# Runs the Jib build.
+- id: jib-build
+  name: gcr.io/cloud-builders/mvn
+  args:
+  - compile
+  - com.google.cloud.tools:jib-maven-plugin:build
+  - -Duser.home=/builder/home
```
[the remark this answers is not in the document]
ImJasonH: We don't do that. We could start if that's helpful enough to you, but it might break existing users.

examples/google-cloud-build/README.md
coollog: Yea, probably not strictly true, but redownloading lots of Maven dependencies each time seems slower than downloading the cache tar and extracting. Based on my runs for the helloworld example, the caching brings the total build time down from 40+ seconds to 30+ seconds. As for concurrency, since single object writes are transactional, concurrent writes would not create corrupt caches, and for our use case, stale writes should be fine since it's just a cache. Incrementality would not be present though since the caches are bundled up as single objects.

chanseokoh:
This ad-hoc example works, but I'm wary of spreading the practice of overriding the Java system property `user.home` as a workaround to make Jib pick up `.docker/config.json` that happens to be under `/builder/home`. Changing the user home to an arbitrary location (`/builder/home` is just a random directory considering that the actual home is `/root`) may affect various tools and programs; obviously, Maven and Jib are such examples here. Particularly, Maven will not make use of the .m2 repository in the actual user home (I mean `/root/.m2`), which GCB might decide to auto-cache in the future. I think we should wait until GCB fixes this issue.

briandealwis:
Should we make the docker config.json location be configurable?

coollog:
@chanseokoh Maybe I should just add a step that links the docker config to `/root/.docker/config.json` first.

chanseokoh:
@coollog that's another way to make it work, but my point is that, looks like GCB will fix the issue and people won't need to do it themselves. If we put up an "official" doc here to do that, people will think copying/linking the file will be a necessary step for GCB, and making it kind of part of "official" instructions is what I'd like to avoid. At least we should add a comment saying linking the file is just a temporary workaround until GCB fixes it.

chanseokoh:
(Linking the file seems better than overriding `user.home`, which will affect everyone on the JVM.)

coollog:
Moving this to a separate repo for now. Will add official Jib example once GCB updates to remove the need for the workarounds here.

PR #1483, opened 2019-02-14
Choose Base Distroless Java version based on project version

### LINKED ISSUE #1279: Intelligently pick Java 8 vs Java 11 distroless base image
From various sources and accounts, it seems that running applications compiled into Java 8 bytecode may fail to run on JVM 11.

Once we have the distroless Java 11, it may be desirable to intelligently match distroless 8 vs 11 based on the binary target version, instead of trying to use the latest JVM all the time.

chanseokoh:
Fixes #1279.

I've spent a lot of time contemplating where would be the ideal place to do the validation if the user is using an incompatible base image. I ended up doing the check in the place where we have the logic to choose the base image, because otherwise, I thought we would be prone to miss updating both the image-choosing logic and the validation logic in a synchronized manner in the future when, e.g., Java 15 is introduced. That resulted in throwing an "incompatible base image" exception from `PluginConfigurationProgressor.getBaseImage()`.

### DIFF
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -19,10 +19,10 @@ package com.google.cloud.tools.jib.gradle;
 import com.google.cloud.tools.jib.Command;
 import com.google.cloud.tools.jib.IntegrationTestingConfiguration;
 import com.google.cloud.tools.jib.registry.LocalRegistry;
+import com.google.common.base.Splitter;
 import java.io.IOException;
 import java.security.DigestException;
 import java.time.Instant;
-import org.gradle.internal.impldep.com.google.api.client.repackaged.com.google.common.base.Splitter;
 import org.gradle.testkit.runner.BuildResult;
 import org.gradle.testkit.runner.UnexpectedBuildFailure;
 import org.hamcrest.CoreMatchers;
@@ -183,7 +183,7 @@ public class SingleProjectIntegrationTest {
       throws DigestException, IOException, InterruptedException {
     Assume.assumeTrue(isJava11RuntimeOrHigher());
 
-    String targetImage = "localhost:6000/compleximage:gradle" + System.nanoTime();
+    String targetImage = "localhost:6000/simpleimage:gradle" + System.nanoTime();
     Assert.assertEquals(
         "Hello, world. \n",
         JibRunHelper.buildToDockerDaemonAndRun(
@@ -204,7 +204,7 @@ public class SingleProjectIntegrationTest {
       Assert.assertThat(
           ex.getMessage(),
           CoreMatchers.containsString(
-              "The base image uses Java 8, but project is using Java 11; perhaps you should "
+              "The base image uses Java 8, but project is using Java 11, perhaps you should "
                   + "configure a Java 11-compatible base image using the 'jib.from.image' "
                   + "parameter, or set targetCompatibility = 8 or below in your build "
                   + "configuration"));
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -204,7 +204,7 @@ public class SingleProjectIntegrationTest {
       Assert.assertThat(
           ex.getMessage(),
           CoreMatchers.containsString(
-              "The base image uses Java 8, but project is using Java 11, perhaps you should "
+              "Your project is using Java 11 but the base image is for Java 8, perhaps you should "
                   + "configure a Java 11-compatible base image using the 'jib.from.image' "
                   + "parameter, or set targetCompatibility = 8 or below in your build "
                   + "configuration"));
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -22,6 +22,7 @@ import com.google.cloud.tools.jib.registry.LocalRegistry;
 import java.io.IOException;
 import java.security.DigestException;
 import java.time.Instant;
+import org.gradle.internal.impldep.com.google.api.client.repackaged.com.google.common.base.Splitter;
 import org.gradle.testkit.runner.BuildResult;
 import org.gradle.testkit.runner.UnexpectedBuildFailure;
 import org.hamcrest.CoreMatchers;
@@ -43,6 +44,11 @@ public class SingleProjectIntegrationTest {
       new LocalRegistry(6000, "testuser2", "testpassword2");
 
   @ClassRule public static final TestProject simpleTestProject = new TestProject("simple");
+
+  private static boolean isJava11RuntimeOrHigher() {
+    Iterable<String> split = Splitter.on(".").split(System.getProperty("java.version"));
+    return Integer.valueOf(split.iterator().next()) >= 11;
+  }
 
   /**
    * Asserts that the creation time of the simple test project is set. If the time parsed from the
@@ -114,6 +120,7 @@ public class SingleProjectIntegrationTest {
             "clean",
             "jib",
             "-Djib.useOnlyProjectCache=true",
+            "-Djib.console=plain",
             "-D_TARGET_IMAGE=" + imageReference,
             "-D_TARGET_USERNAME=" + username,
             "-D_TARGET_PASSWORD=" + password,
@@ -150,6 +157,7 @@ public class SingleProjectIntegrationTest {
           "clean",
           "jib",
           "-Djib.useOnlyProjectCache=true",
+          "-Djib.console=plain",
           "-x=classes",
           "-D_TARGET_IMAGE=" + targetImage);
       Assert.fail();
@@ -168,6 +176,39 @@ public class SingleProjectIntegrationTest {
     assertDockerInspect(targetImage);
     assertSimpleCreationTimeIsAfter(beforeBuild, targetImage);
     assertWorkingDirectory("/home", targetImage);
+  }
+
+  @Test
+  public void testDockerDaemon_simpleOnJava11()
+      throws DigestException, IOException, InterruptedException {
+    Assume.assumeTrue(isJava11RuntimeOrHigher());
+
+    String targetImage = "localhost:6000/compleximage:gradle" + System.nanoTime();
+    Assert.assertEquals(
+        "Hello, world. \n",
+        JibRunHelper.buildToDockerDaemonAndRun(
+            simpleTestProject, targetImage, "build-java11.gradle"));
+  }
+
+  @Test
+  public void testDockerDaemon_simpleWithIncompatibleJava11()
+      throws DigestException, IOException, InterruptedException {
+    Assume.assumeTrue(isJava11RuntimeOrHigher());
+
+    try {
+      JibRunHelper.buildToDockerDaemonAndRun(
+          simpleTestProject, "willnotbuild", "build-java11-incompatible.gradle");
+      Assert.fail();
+
+    } catch (UnexpectedBuildFailure ex) {
+      Assert.assertThat(
+          ex.getMessage(),
+          CoreMatchers.containsString(
+              "The base image uses Java 8, but project is using Java 11; perhaps you should "
+                  + "configure a Java 11-compatible base image using the 'jib.from.image' "
+                  + "parameter, or set targetCompatibility = 8 or below in your build "
+                  + "configuration"));
+    }
   }
 
   @Test
@@ -198,7 +239,7 @@ public class SingleProjectIntegrationTest {
     Instant beforeBuild = Instant.now();
     Assert.assertEquals(
         "Hello, world. An argument.\nrw-r--r--\nrw-r--r--\nfoo\ncat\n",
-        JibRunHelper.buildToDockerDaemonAndRun(simpleTestProject, targetImage));
+        JibRunHelper.buildToDockerDaemonAndRun(simpleTestProject, targetImage, "build.gradle"));
     assertSimpleCreationTimeIsAfter(beforeBuild, targetImage);
     assertDockerInspect(targetImage);
     assertWorkingDirectory("/home", targetImage);
@@ -218,6 +259,7 @@ public class SingleProjectIntegrationTest {
             "clean",
             "jibDockerBuild",
             "-Djib.useOnlyProjectCache=true",
+            "-Djib.console=plain",
             "-D_TARGET_IMAGE=" + targetImage,
             "-b=build-dockerclient.gradle",
             "--debug");
@@ -241,6 +283,7 @@ public class SingleProjectIntegrationTest {
             "clean",
             "jibBuildTar",
             "-Djib.useOnlyProjectCache=true",
+            "-Djib.console=plain",
             "-D_TARGET_IMAGE=" + targetImage);
 
     JibRunHelper.assertBuildSuccess(buildResult, "jibBuildTar", "Built image tarball at ");
--- a/jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-java11-incompatible.gradle
+++ b/jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-java11-incompatible.gradle
@@ -1,0 +1,18 @@
+plugins {
+  id 'java'
+  id 'com.google.cloud.tools.jib'
+}
+
+sourceCompatibility = 11
+targetCompatibility = 11
+
+repositories {
+  mavenCentral()
+}
+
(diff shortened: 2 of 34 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -289,16 +273,34 @@ static PluginConfigurationProcessor processCommonConfiguration(
    * @param rawConfiguration raw configuration data
    * @param projectProperties used for providing additional information
    * @return the base image
+   * @throws IncompatibleBaseImageJavaVersionException
```
TadCordle: Missing description (although doesn't matter that much since it's package private).

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -170,6 +178,39 @@ public void testBuild_simple() throws IOException, InterruptedException, DigestE
     assertWorkingDirectory("/home", targetImage);
   }
 
+  @Test
+  public void testDockerDaemon_simpleOnJava11()
+      throws DigestException, IOException, InterruptedException {
+    Assume.assumeTrue(isJava11RuntimeOrHigher());
+
+    String targetImage = "localhost:6000/compleximage:gradle" + System.nanoTime();
```
coollog: nit: `simpleimage`?

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-java11-incompatible.gradle
```diff
@@ -0,0 +1,18 @@
...
+
+repositories {
+  mavenCentral()
+}
+
+dependencies {
+  compile files('libs/dependency-1.0.0.jar')
+}
+
+jib.from.image = 'gcr.io/distroless/java'
```
coollog: Should we qualify this with the `8` tag? (In case we change `latest` to be Java 11 in the future)

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/IncompatibleBaseImageJavaVersionException.java
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.common;
+
+public class IncompatibleBaseImageJavaVersionException extends Exception {
```
coollog: nit: short javadoc

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -289,16 +273,34 @@ static PluginConfigurationProcessor processCommonConfiguration(
...
+
+    // Base image not configured; auto-pick Distroless.
+    if (projectProperties.isWarProject()) {
+      return "gcr.io/distroless/java/jetty";
+    } else if (javaVersion <= 8) {
+      return "gcr.io/distroless/java:8";
+    } else if (javaVersion <= 11) {
+      return "gcr.io/distroless/java:11";
+    } else {
+      throw new IncompatibleBaseImageJavaVersionException(11, javaVersion);
```
coollog: Should this use `11` as the `baseImageMajorJavaVersion`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -289,16 +273,34 @@ static PluginConfigurationProcessor processCommonConfiguration(
...
+  static String getBaseImage(RawConfiguration rawConfiguration, ProjectProperties projectProperties)
+      throws IncompatibleBaseImageJavaVersionException {
+    int javaVersion = projectProperties.getMajorJavaVersion();
+
+    if (rawConfiguration.getFromImage().isPresent()) {
+      String baseImage = rawConfiguration.getFromImage().get();
+
+      if (knownDistrolessJava8Image(baseImage) && javaVersion > 8) {
+        throw new IncompatibleBaseImageJavaVersionException(8, javaVersion);
+      } else if (knownDistrolessJava11Image(baseImage) && javaVersion > 11) {
```
coollog: nit: else's in this method not necessary

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -471,6 +473,45 @@ private static Path getCheckedCacheDirectory(String property, Path defaultPath)
...
+  /**
+   * Checks whether or not the known default Java 8 distroless base image is being used. Checking
+   * against only images known to Java 8, the method may to return {@code false} for Java 8
+   * distroless unknown to it.
+   *
+   * @param baseImageConfiguration the configured base image
+   * @return {@code true} if the base image is equal to one of the known Java 8 distroless images,
+   *     else {@code false}
+   */
+  private static boolean knownDistrolessJava8Image(String baseImageConfiguration) {
```
coollog: nit: `isKnownDistrolessJava8Image`?

coollog: nit: just `baseImage` since it's just a string?

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -22,6 +22,7 @@
 import java.io.IOException;
 import java.security.DigestException;
 import java.time.Instant;
+import org.gradle.internal.impldep.com.google.api.client.repackaged.com.google.common.base.Splitter;
```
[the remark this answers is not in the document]
chanseokoh: Nice catch.

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-java11-incompatible.gradle
chanseokoh: My idea was to leave the tag out so that if `latest` becomes Java 11, then builds will break and we will be immediately notified. When the switch happens, we will fix the test by qualifying this with the `8` tag, giving the chance to update other places that need attention.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: We are auto-picking a base image and the highest available now is 11, so I thought saying 11 works here.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: Will add. The exception class is not package-private, so makes sense to add it.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -170,6 +178,39 @@ public void testBuild_simple() throws IOException, InterruptedException, DigestE
...
+    try {
+      JibRunHelper.buildToDockerDaemonAndRun(
+          simpleTestProject, "willnotbuild", "build-java11-incompatible.gradle");
+      Assert.fail();
+
+    } catch (UnexpectedBuildFailure ex) {
+      Assert.assertThat(
+          ex.getMessage(),
+          CoreMatchers.containsString(
+              "The base image uses Java 8, but project is using Java 11, perhaps you should "
```
briandealwis: Trivial tweak:
> *your* project

But I wonder if flipping the order helps identify the mismatch:
> _Your project is using Java 11 but the base image is for Java 8.  Perhaps…_

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -171,16 +171,7 @@ public void buildDocker()
...
-              + ex.getBaseImageMajorJavaVersion()
-              + ", but project is using Java "
-              + ex.getProjectMajorJavaVersion()
-              + "; perhaps you should configure a Java "
-              + ex.getProjectMajorJavaVersion()
-              + "-compatible base image using the 'jib.from.image' parameter, or set "
-              + "targetCompatibility = "
-              + ex.getBaseImageMajorJavaVersion()
-              + " or below in your build configuration");
+          HelpfulSuggestions.forIncompatibleBaseImageJavaVesionForGradle(8, 11));
```
coollog: Should this just take in the exception instead?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -471,6 +476,45 @@ private static Path getCheckedCacheDirectory(String property, Path defaultPath)
...
+        || baseImage.equals("gcr.io/distroless/java:8")
+        || baseImage.equals("gcr.io/distroless/java:8-debug")
+        || baseImage.equals("gcr.io/distroless/java/jetty")
+        || baseImage.equals("gcr.io/distroless/java/jetty:debug")
+        || baseImage.equals("gcr.io/distroless/java/jetty:java8")
+        || baseImage.equals("gcr.io/distroless/java/jetty:java8-debug");
+  }
+
+  /**
+   * Checks whether or not the known default Java 11 distroless base image is being used. Checking
```
briandealwis: _Checks if the base image is a known Java 11 distroless image._?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -471,6 +476,45 @@ private static Path getCheckedCacheDirectory(String property, Path defaultPath)
...
+   * Checks whether or not the known default Java 8 distroless base image is being used. Checking
+   * against only images known to Java 8, the method may to return {@code false} for Java 8
+   * distroless unknown to it.
+   *
+   * @param baseImage the configured base image
+   * @return {@code true} if the base image is equal to one of the known Java 8 distroless images,
+   *     else {@code false}
+   */
+  private static boolean isKnownDistrolessJava8Image(String baseImage) {
+    // TODO: drop "latest" and the likes once it no longer points to Java 8.
```
briandealwis: Since there's just `:debug` may as well list it explicitly (I think _like_ would be singular: _the like once they no longer_?)

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -471,6 +476,45 @@ private static Path getCheckedCacheDirectory(String property, Path defaultPath)
     return defaultPath;
   }
 
+  /**
+   * Checks whether or not the known default Java 8 distroless base image is being used. Checking
```
briandealwis: _Checks if the base image uses a known Java 8 distroless image._ is shorter.  I wasn't sure what "default" meant, especially since Jib doesn't use the `:debug` images ever.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -471,6 +476,45 @@ private static Path getCheckedCacheDirectory(String property, Path defaultPath)
...
+   * Checks whether or not the known default Java 11 distroless base image is being used. Checking
+   * against only images known to Java 11, the method may to return {@code false} for Java 11
+   * distroless unknown to it.
+   *
+   * @param baseImage the configured base image
+   * @return {@code true} if the base image is equal to one of the known Java 11 distroless images,
+   *     else {@code false}
+   */
+  private static boolean isKnownDistrolessJava11Image(String baseImage) {
+    // TODO: add "latest" and the likes once it points to Java 11.
```
briandealwis: There is no `latest` here

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
chanseokoh: Sounds better. Unfortunately, `, perhaps you should` is the part automatically added by `HelpfulSuggestions.suggest()`, so can't fix here to use a dot.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: Actually, I noticed this is a general method checking if the given image reference is a known distroless image. It doesn't have to be a base image. Reworded.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: There are "jetty:debug" too. Noticed I missed "jetty:latest".

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: Done. Also noticed the image doesn't have to be a base image. Fixed other minor things too.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
chanseokoh: I didn't like making this class depend on the exception class.

coollog: Should this at least take the versions from the exception instead of the hard-coded 8 and 11? Also, should prob add `ex` as the cause.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -262,20 +260,13 @@ public String getVersion() {
...
-    // maven-compiler-plugin default is 1.6
-    int version = 6;
-
+  @Override
+  public int getMajorJavaVersion() {
     // Check properties for version
     if (project.getProperties().getProperty("maven.compiler.target") != null) {
-      version = getVersionFromString(project.getProperties().getProperty("maven.compiler.target"));
+      return getVersionFromString(project.getProperties().getProperty("maven.compiler.target"));
     } else if (project.getProperties().getProperty("maven.compiler.release") != null) {
```
coollog: nit: These else's are not necessary as well

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
chanseokoh: Gosh. Nice catch.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -165,6 +165,11 @@ public void buildTar()
     } catch (InvalidContainerVolumeException ex) {
       throw new GradleException(
           "container.volumes is not an absolute Unix-style path: " + ex.getInvalidVolume(), ex);
+
+    } catch (IncompatibleBaseImageJavaVersionException ex) {
+      throw new GradleException(
+          HelpfulSuggestions.forIncompatibleBaseImageJavaVesionForGradle(
+              ex.getBaseImageMajorJavaVersion(), ex.getProjectMajorJavaVersion()));
```
coollog: Also, maybe include `ex` as the cause for the `GradleException` for the stacktrace?

chanseokoh: We should. I missed it. Thanks.

chanseokoh:
Will update CHANGELOG shortly.

Issue #1484, opened 2019-02-15
Add new file watching task/goal for more optimal Skaffold usage


TadCordle:
Rather than outputting all files at once, Jib will output a JSON string categorizing the different files into the following:
- `build`: build definitions that may indicate a change in project structure when changed
- `inputs`: directories containing source files/resources to watch for changes
- `ignore`: specific files within `inputs` that should be ignored

Steps:
- [x] Add JSON template for task/goal output
- [x] Add Gradle task
- [ ] Add Maven goal

PR #1485, opened 2019-02-15
Add JSON template for Skaffold file-watching

TadCordle:
Part of #1484. Adds a class that outputs the JSON string to be passed to Skaffold.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
@@ -14,7 +14,7 @@
  * the License.
  */
 
-package com.google.cloud.tools.jib.skaffold.json;
+package com.google.cloud.tools.jib.frontend;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.google.cloud.tools.jib.blob.Blob;
@@ -27,29 +27,34 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 
-@JsonIgnoreProperties(ignoreUnknown = true)
-public class SkaffoldFilesTemplate implements JsonTemplate {
+public class SkaffoldFilesOutput {
 
-  private final List<String> buildFiles = new ArrayList<>();
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  private static class SkaffoldFilesTemplate implements JsonTemplate {
 
-  private final List<String> inputs = new ArrayList<>();
+    private final List<String> buildFiles = new ArrayList<>();
 
-  private final List<String> ignore = new ArrayList<>();
+    private final List<String> inputs = new ArrayList<>();
+
+    private final List<String> ignore = new ArrayList<>();
+  }
+
+  private final SkaffoldFilesTemplate skaffoldFilesTemplate = new SkaffoldFilesTemplate();
 
   public void addBuildFile(Path buildFile) {
-    buildFiles.add(buildFile.toString());
+    skaffoldFilesTemplate.buildFiles.add(buildFile.toString());
   }
 
   public void addInput(Path inputFile) {
-    inputs.add(inputFile.toString());
+    skaffoldFilesTemplate.inputs.add(inputFile.toString());
   }
 
   public void addIgnore(Path ignoreFile) {
-    ignore.add(ignoreFile.toString());
+    skaffoldFilesTemplate.ignore.add(ignoreFile.toString());
   }
 
   public String getJsonString() throws IOException {
-    Blob blob = JsonTemplateMapper.toBlob(this);
+    Blob blob = JsonTemplateMapper.toBlob(skaffoldFilesTemplate);
     try (OutputStream outputStream = new ByteArrayOutputStream()) {
       blob.writeTo(outputStream);
       return outputStream.toString();
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
@@ -17,19 +17,45 @@
 package com.google.cloud.tools.jib.frontend;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.google.cloud.tools.jib.blob.Blob;
+import com.google.cloud.tools.jib.blob.Blobs;
 import com.google.cloud.tools.jib.json.JsonTemplate;
 import com.google.cloud.tools.jib.json.JsonTemplateMapper;
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
- * Builds a JSON string containing files and directories for <a
- * href="https://github.com/GoogleContainerTools/skaffold">Skaffold</a> to watch.
+ * Builds a JSON string containing files and directories that <a
+ * href="https://github.com/GoogleContainerTools/skaffold">Skaffold</a> should watch for changes
+ * (and consequently trigger rebuilds).
+ *
+ * <p>{@code buildFiles} consist of build definitions. Changes in these files indicate that the
+ * project structure may have changed, so Skaffold will refresh the file watch list when this
+ * happens.
+ *
+ * <p>{@code inputs} consist of source/resource directories. Skaffold will trigger a rebuild when
+ * changes are detected in these files.
+ *
+ * <p>{@code ignore} consists of files that the Skaffold file watcher should not watch.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * {
+ *   "buildFiles": [
+ *     "buildFile1",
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
```diff
@@ -0,0 +1,88 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Builds a JSON string containing files and directories for <a
+ * href="https://github.com/GoogleContainerTools/skaffold">Skaffold</a> to watch.
+ */
```
briandealwis: It would be helpful to define what this string covers.  I skipped down to see addBuildFiles and realized I didn't know what a build file was.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
```diff
@@ -0,0 +1,88 @@
...
+  public void addBuildFile(Path buildFile) {
+    skaffoldFilesTemplate.buildFiles.add(buildFile.toString());
+  }
+
+  /**
+   * Adds an input directory.
+   *
+   * @param inputFile the path to the directory
+   */
+  public void addInput(Path inputFile) {
```
briandealwis: Should we validate that this path is a directory?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
```diff
@@ -0,0 +1,88 @@
...
+
+  private final SkaffoldFilesTemplate skaffoldFilesTemplate = new SkaffoldFilesTemplate();
+
+  /**
+   * Adds a build file.
+   *
+   * @param buildFile the path to the file
+   */
+  public void addBuildFile(Path buildFile) {
+    skaffoldFilesTemplate.buildFiles.add(buildFile.toString());
```
briandealwis: Should we validate that this is a file?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
TadCordle: Saw your comment on the design doc, don't think we need to do this.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
TadCordle: Yeah, I was a bit lazy with this javadoc, I'll improve it.

TadCordle:
@coollog I wanted to keep this inside jib-core since it makes use of `Blobs` and `JsonTemplateMapper`.

chanseokoh:
Hmm... @coollog has a point, but what @TadCordle said makes sense. This is an interesting decision.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/SkaffoldFilesOutput.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Gets the added files in JSON format.
+   *
+   * @return the files in a JSON string
+   * @throws IOException if writing out the JSON fails
+   */
+  public String getJsonString() throws IOException {
+    return Blobs.writeToString(JsonTemplateMapper.toBlob(skaffoldFilesTemplate));
```
loosebazooka: Are we only using Blobs and JsonTemplateManager here? Would it make sense to use the underlying library to write out the string, and have this in plugins-common?

TadCordle: We could for now, but I just wasn't sure if we wanted to use Blobs and JsonTemplateManager from outside of jib-core, since they're low-level/not part of the public api.

loosebazooka: Sorry I meant the underlying library we are using for json. With stuff from the jackson library (using `ObjectMapper`) directly.

TadCordle: Hmm, yeah, I wanted to avoid duplicate code, but I guess it won't be as much as I thought.

loosebazooka: cool

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutput.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private final List<String> inputs = new ArrayList<>();
+
+    private final List<String> ignore = new ArrayList<>();
+  }
+
+  private final SkaffoldFilesTemplate skaffoldFilesTemplate = new SkaffoldFilesTemplate();
+
+  /**
+   * Adds a build file/directory.
```
loosebazooka: Is this actually build file/**directory**? If so, should we then rename `addBuildFile` -> `addBuild` ? Or are we mandating that this is all only files (and not directories?)

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutputTest.java
```diff
@@ -0,0 +1,40 @@
...
+    skaffoldFilesOutput.addBuildFile(Paths.get("buildFile1"));
+    skaffoldFilesOutput.addBuildFile(Paths.get("buildFile2"));
+    skaffoldFilesOutput.addInput(Paths.get("input1"));
+    skaffoldFilesOutput.addInput(Paths.get("input2"));
+    skaffoldFilesOutput.addIgnore(Paths.get("ignore1"));
+    skaffoldFilesOutput.addIgnore(Paths.get("ignore2"));
+    Assert.assertEquals(
+        "{\"buildFiles\":[\"buildFile1\",\"buildFile2\"],\"inputs\":[\"input1\",\"input2\"],\"ignore\":[\"ignore1\",\"ignore2\"]}",
+        skaffoldFilesOutput.getJsonString());
+  }
```
loosebazooka: would like a test to see what happens when it's all empty (or something like that)? Expectation is `{}` ?

TadCordle: Should it omit empty fields? Right each field is there, just as an empty list. i.e. `{"build":[],"inputs":[],"ignore":[]}`

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutput.java
TadCordle: @briandealwis left a comment on the design doc suggesting that we may want to include directories here as well as files:

>Do we need to be specific as to files-only, directories-only? Maven supports a .mvn/ directory where extensions can be put — these would be buildFiles. I could see -ignore- wildcards being useful for both input and ignore.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutput.java
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param build the path to the file
```
loosebazooka: Last thing: lines 77,86,95 should probably all indicate file/directory?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutputTest.java
loosebazooka: I didn't know what the behavior was, but having the test like you added in now defines whatever behavior it does have.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutput.java
```diff
@@ -0,0 +1,113 @@
...
+ * <p>{@code inputs} consist of source/resource files/directories. Skaffold will trigger a rebuild
+ * when changes are detected in these files.
+ *
+ * <p>{@code ignore} consists of files/directories that the Skaffold file watcher should not watch.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * {
+ *   "buildFiles": [
```
loosebazooka: oops, I missed this one, doc needs to be updated in "build"

TadCordle: Yeah I just realized that myself, haha

PR #1492, opened 2019-02-20
Add _jibSkaffoldFilesV2 task

TadCordle:
Part of #1484. The logic is very similar to the original `_jibSkaffoldFiles` (actually almost completely copy-pasted), but it categorizes the files and prints them out in a JSON string instead of just a list.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.gradle;
+
+import com.google.cloud.tools.jib.plugins.common.SkaffoldFilesOutput;
+import com.google.common.base.Preconditions;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.Nullable;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.artifacts.ProjectDependency;
+import org.gradle.api.artifacts.PublishArtifact;
+import org.gradle.api.initialization.Settings;
+import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.SourceSet;
+import org.gradle.api.tasks.TaskAction;
+
+/**
+ * Print out changing source dependencies on a project.
+ *
+ * <p>Expected use: {@code ./gradlew _jibSkaffoldFiles -q} or {@code ./gradlew
+ * :<subproject>:_jibSkaffoldFiles -q}
+ */
+public class FilesTaskV2 extends DefaultTask {
+
+  private final SkaffoldFilesOutput skaffoldFilesOutput = new SkaffoldFilesOutput();
+
+  @Nullable private JibExtension jibExtension;
+
+  public FilesTaskV2 setJibExtension(JibExtension jibExtension) {
+    this.jibExtension = jibExtension;
+    return this;
+  }
+
+  @TaskAction
+  public void listFiles() throws IOException {
+    Preconditions.checkNotNull(jibExtension);
+    Project project = getProject();
+
+    // If this is not the root project, add the root project's build.gradle and settings.gradle
+    if (project != project.getRootProject()) {
+      addGradleFiles(project.getRootProject());
+    }
+
+    addProjectFiles(project);
+
+    // Add extra layer
+    if (Files.exists(jibExtension.getExtraDirectory().getPath())) {
+      skaffoldFilesOutput.addInput(jibExtension.getExtraDirectory().getPath());
+    }
+
+    // Find project dependencies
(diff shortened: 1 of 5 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
```diff
@@ -0,0 +1,193 @@
...
+   * Prints build files, sources, and resources associated with a project.
+   *
+   * @param project the project
+   */
+  private void addProjectFiles(Project project) {
+    JavaPluginConvention javaConvention =
+        project.getConvention().getPlugin(JavaPluginConvention.class);
+    SourceSet mainSourceSet =
+        javaConvention.getSourceSets().findByName(SourceSet.MAIN_SOURCE_SET_NAME);
+    if (mainSourceSet == null) {
```
loosebazooka: Should we add gradle files whether or not it has a valid main source set? Presumably the gradle config is doing something?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
```diff
@@ -0,0 +1,193 @@
...
+      return;
+    }
+
+    // Print build config, settings, etc.
+    addGradleFiles(project);
```
loosebazooka: I was a little confused by us calling `addGradleFiles` here and in the main task action (https://github.com/GoogleContainerTools/jib/pull/1492/files#diff-6459724c2453bb9a6d621ab67dd3c866R65). What do you think about making this method just deal with sources and resources and calling both `addGradlefile` and `addProjectSources` directly in the main taskaction method?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
```diff
@@ -0,0 +1,193 @@
...
+    // Print files
+    System.out.println(skaffoldFilesOutput.getJsonString());
+  }
+
+  /**
+   * Prints the locations of a project's build.gradle, settings.gradle, and gradle.properties.
+   *
+   * @param project the project
+   */
+  private void addGradleFiles(Project project) {
```
loosebazooka: we may also need to deal with script plugins here (https://docs.gradle.org/current/userguide/plugins.html#sec:script_plugins)

I think there's a way to query it, but I'm not sure.

loosebazooka: maybe a TODO, presumably this isn't available in v1 either

loosebazooka: filed: https://discuss.gradle.org/t/programmatically-access-script-plugin-references/30668 for now, lets see what they say.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
TadCordle: I think the one in the main task action adds the build files for the root project in case you run the task on a sub-project, and the one in `addProjectFiles` adds the sub-project's build files.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
TadCordle: Hmm, yeah I'll open an issue for this.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
```diff
@@ -0,0 +1,191 @@
+/*
+ * Copyright 2018 Google LLC.
```
loosebazooka: 2019

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/FilesTaskV2Test.java
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright 2018 Google LLC.
```
loosebazooka: 2019

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
TadCordle: https://github.com/GoogleContainerTools/jib/issues/1494

loosebazooka:
we probably also want to track deprecating the v1 of the task - once we update skaffold.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutput.java
```diff
@@ -69,7 +70,23 @@
...
+
+  /**
+   * Creates a {@link SkaffoldFilesOutput} from a JSON string.
+   *
+   * @param json the JSON string
+   * @throws IOException if reading the JSON string fails
+   */
+  @VisibleForTesting
+  public SkaffoldFilesOutput(String json) throws IOException {
+    this.skaffoldFilesTemplate = new ObjectMapper().readValue(json, SkaffoldFilesTemplate.class);
```
coollog: nit: no need for `this`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTaskV2.java
```diff
@@ -0,0 +1,191 @@
...
+
+/**
+ * Prints out changing source dependencies on a project.
+ *
+ * <p>Expected use: {@code ./gradlew _jibSkaffoldFiles -q} or {@code ./gradlew
+ * :<subproject>:_jibSkaffoldFiles -q}
+ */
+public class FilesTaskV2 extends DefaultTask {
+
+  private final SkaffoldFilesOutput skaffoldFilesOutput = new SkaffoldFilesOutput();
```
coollog: I think we should not have the task be stateful like this, since if `listFiles` is called twice there would be two JSONs printed out.

loosebazooka: I'm not sure I understand, this task will only be called once. The gradle system wont add it to the task tree twice?

coollog: Ah okay, it just felt like the variable could be local to the `listFiles` method to avoid any potential misuse of `listFiles` by calling it more than once.

PR #1506, opened 2019-02-22
Restart docker as workaround to fix clock skew

chanseokoh:
Recently, the degree of flakiness has well crossed the tolerance threshold.

selection_001

Then I just noticed we weren't applying the same workaround to restart Docker on presubmit builds. Hopefully this resolves the issue.

### DIFF
--- a/kokoro/presubmit.sh
+++ b/kokoro/presubmit.sh
@@ -10,6 +10,17 @@ export PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/b
 docker stop $(docker ps --all --quiet) || true
 docker kill $(docker ps --all --quiet) || true
 
+# Restarting Docker for Mac to get around the certificate expiration issue:
+# b/112707824
+# https://github.com/GoogleContainerTools/jib/issues/730#issuecomment-413603874
+# https://github.com/moby/moby/issues/11534
+# TODO: remove this temporary fix once b/112707824 is permanently fixed.
+if [ "${KOKORO_JOB_CLUSTER}" = "MACOS_EXTERNAL" ]; then
+  osascript -e 'quit app "Docker"'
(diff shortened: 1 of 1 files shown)


kokoro/presubmit.sh
```diff
@@ -10,6 +10,17 @@ export PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/b
...
 
+# Restarting Docker for Mac to get around the certificate expiration issue:
+# b/112707824
+# https://github.com/GoogleContainerTools/jib/issues/730#issuecomment-413603874
+# https://github.com/moby/moby/issues/11534
+# TODO: remove this temporary fix once b/112707824 is permanently fixed.
+if [ "${KOKORO_JOB_CLUSTER}" = "MACOS_EXTERNAL" ]; then
+  osascript -e 'quit app "Docker"'
+  open -a Docker
+  while ! docker info > /dev/null 2>&1; do sleep 1; done
```
loosebazooka: should this timeout not be infinite? Or will kokoro just timeout if this fails?

chanseokoh: Kokoro has its own timeout, although it is rather large. But I opt for keeping things simple, since it is extremely rare that this fails.

loosebazooka: does `open -a Docker` return a success/error code?

chanseokoh: Can't verify myself, but as a core command, I am pretty sure it will return proper codes: https://ss64.com/osx/open.html

briandealwis: `open` returns non-zero if the requested application was not found.  The implementation launches the application in the background and then returns, so the wait is still required.

loosebazooka: sounds good to me.

Issue #1513, opened 2019-02-27
Maven --batch-mode not supressing jib progress report 


fralalonde:
When captured by non-interactive terminal, Jib's maven plugin output of transfer progress makes job logs much longer / dirtier than they should be. This is a minor, but very real pain point when trying to keep Continuous Integration logs meaningful and readable.

Maven provides the `--batch-mode` CLI option to suppress such progress reports. Jib should use the same mechanism to suppress it's output when specified. 

Here's a _partial_ sample output of jib when running in non-interactive terminal:

```
...
[INFO] Executing tasks:
[INFO] [==================            ] 60.7% complete
[INFO] > building dependencies layer
[INFO] > setting up to push layers
[INFO] > setting up to push layers
[INFO] > pulling base image layer sha256:6ac6f91f019c1...
[INFO] 
[INFO] Executing tasks:
[INFO] [==================            ] 60.8% complete
[INFO] > building dependencies layer
[INFO] > setting up to push layers
[INFO] > setting up to push layers
[INFO] > pulling base image layer sha256:6ac6f91f019c1...
[INFO] 
[INFO] Executing tasks:
[INFO] [==================            ] 60.9% complete
[INFO] > building dependencies layer
[INFO] > setting up to push layers
[INFO] > setting up to push layers
[INFO] > pulling base image layer sha256:6ac6f91f019c1...
[INFO] 
[INFO] Executing tasks:
[INFO] [==================            ] 61.0% complete
...
```

chanseokoh:
Thanks for the PR. I also suggested in the past the batch mode should suppress the progress. Perhaps we should seriously consider it.

For now, you can set `-Djib.console=plain` to suppress the progress.

fralalonde:
Thanks, I'll try that, still better than filtering.

chanseokoh:
The previous discussion around disabling the progress for the batch mode: https://github.com/GoogleContainerTools/jib/pull/1344#discussion_r241636041

chanseokoh:
Fixed by #1514.

chanseokoh:
@fralalonde v1.0.2 released with the fix. Thanks for your contribution!

PR #1514, opened 2019-02-27
Stop displaying progress when Maven --batch-mode is specified

fralalonde:
Fix for #1513

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
@@ -61,6 +61,7 @@ public class MavenProjectProperties implements ProjectProperties {
 
   /**
    * @param project the {@link MavenProject} for the plugin.
+   * @param session the {@link MavenSession} for the plugin.
    * @param log the Maven {@link Log} to log messages during Jib execution
    * @param extraDirectory path to the directory for the extra files layer
    * @param permissions map from path on container to file permissions for extra-layer files
@@ -70,6 +71,7 @@ public class MavenProjectProperties implements ProjectProperties {
    */
   static MavenProjectProperties getForProject(
       MavenProject project,
+      MavenSession session,
       Log log,
       Path extraDirectory,
       Map<AbsoluteUnixPath, FilePermissions> permissions,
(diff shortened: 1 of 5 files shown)


fralalonde:
CLA: `I signed it!`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -130,20 +133,26 @@ private static EventHandlers makeEventHandlers(
...
+    EventHandlers eventHandlers =
+        new EventHandlers()
+            .add(
+                JibEventType.LOGGING,
+                logEvent -> consoleLogger.log(logEvent.getLevel(), logEvent.getMessage()))
+            .add(
+                JibEventType.TIMING,
+                new TimerEventHandler(message -> consoleLogger.log(LogEvent.Level.DEBUG, message)));
+
+    if (session.getRequest().isInteractiveMode()) {
```
chanseokoh: We should still add the dumb plain progress messages (just a few lines of message when not using the fancy progress bars).

```java
    ConsoleLoggerBuilder logEventHandlerBuilder =
        (isProgressFooterEnabled()
                ? ConsoleLoggerBuilder.rich(singleThreadedExecutor)
                : ConsoleLoggerBuilder.plain(singleThreadedExecutor).progress(log::info))
```

So, the check should be done inside `isProgressFooterEnabled()`. Will need passing `session` to the method.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -146,7 +149,11 @@ private static EventHandlers makeEventHandlers(
                             update.getProgress(), update.getUnfinishedAllocations()))));
   }
 
-  private static boolean isProgressFooterEnabled() {
+  private static boolean isProgressFooterEnabled(MavenSession session) {
```
chanseokoh: Now is a good time to add a test for this method. You'll need to make it package-private and add the `@VisibleForTesting`.

```java
@VisibleForTesting
tatic boolean isProgressFooterEnabled(...) {
```

chanseokoh: Taken care of in #1525.

PR #1541, opened 2019-03-08
Allow setting approot any time during JavaContainerBuilder construction

### LINKED ISSUE #1476: Allow changing app root in JavaContainerBuilder even after files are added
With the current way JavaContainerBuilder is written, the app root needs to be set before any files are added. Allowing app root to be set at any time before `toContainerBuilder` is called would fit the declarative style a bit better.

TadCordle:
This changes `JavaContainerBuilder` to not build the layer configuration until `toContainerBuilder()` is called, allowing people who use `JavaContainerBuilder` to change the app root even after files have been added.

Fixes #1476.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
@@ -363,10 +363,7 @@ public class JavaContainerBuilder {
               + "JavaContainerBuilder#setMainClass(String), or consider using a "
               + "jib.frontend.MainClassFinder to infer the main class");
     }
-    if (addedClasses.isEmpty()
-        && addedResources.isEmpty()
-        && addedDependencies.isEmpty()
-        && addedOthers.isEmpty()) {
+    if (classpath.isEmpty()) {
       throw new IllegalStateException(
           "Failed to construct entrypoint because no files were added to the JavaContainerBuilder");
     }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
@@ -29,7 +29,6 @@ import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.NotDirectoryException;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
@@ -128,6 +127,8 @@ public class JavaContainerBuilder {
   private final JibContainerBuilder jibContainerBuilder;
   private final List<String> jvmFlags = new ArrayList<>();
   private final LinkedHashSet<RelativeUnixPath> classpath = new LinkedHashSet<>(4);
+
+  // Keeps track of files to add to the image, by system path
   private final List<PathPredicatePair> addedResources = new ArrayList<>();
   private final List<PathPredicatePair> addedClasses = new ArrayList<>();
   private final List<Path> addedDependencies = new ArrayList<>();
@@ -177,27 +178,7 @@ public class JavaContainerBuilder {
         throw new NoSuchFileException(file.toString());
       }
     }
-
-    // Detect duplicate filenames and rename with filesize to avoid collisions
-    List<String> duplicates =
-        dependencyFiles
-            .stream()
-            .map(Path::getFileName)
-            .map(Path::toString)
-            .collect(Collectors.groupingBy(filename -> filename, Collectors.counting()))
-            .entrySet()
-            .stream()
-            .filter(entry -> entry.getValue() > 1)
-            .map(Entry::getKey)
-            .collect(Collectors.toList());
-    for (Path file : dependencyFiles) {
-      addedDependencies.add(
-          Paths.get(
-              duplicates.contains(file.getFileName().toString())
-                  ? file.getFileName().toString().replaceFirst("\\.jar$", "-" + Files.size(file))
-                      + ".jar"
-                  : file.getFileName().toString()));
-    }
+    addedDependencies.addAll(dependencyFiles);
     classpath.add(DEPENDENCIES_CLASSPATH);
     return this;
   }
@@ -389,6 +370,18 @@ public class JavaContainerBuilder {
     }
 
     // Add dependencies to layer configuration
+    // Detect duplicate filenames and rename with filesize to avoid collisions
+    List<String> duplicates =
+        addedDependencies
+            .stream()
+            .map(Path::getFileName)
+            .map(Path::toString)
+            .collect(Collectors.groupingBy(filename -> filename, Collectors.counting()))
+            .entrySet()
+            .stream()
+            .filter(entry -> entry.getValue() > 1)
+            .map(Entry::getKey)
+            .collect(Collectors.toList());
     for (Path file : addedDependencies) {
       layerConfigurationsBuilder.addFile(
           file.getFileName().toString().contains("SNAPSHOT")
@@ -397,7 +390,13 @@ public class JavaContainerBuilder {
           file,
           appRoot
               .resolve(JavaEntrypointConstructor.DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE)
-              .resolve(file.getFileName().toString()));
+              .resolve(
+                  duplicates.contains(file.getFileName().toString())
+                      ? file.getFileName()
+                              .toString()
+                              .replaceFirst("\\.jar$", "-" + Files.size(file))
+                          + ".jar"
+                      : file.getFileName().toString()));
     }
 
     // Add others to layer configuration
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -107,38 +126,19 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
...
-  /** The entrypoint classpath element corresponding to dependencies. */
-  private AbsoluteUnixPath dependenciesClasspath;
-
-  /** Absolute path of directory containing additional classpath files on container. */
-  private AbsoluteUnixPath othersPath;
+  private final LinkedHashSet<RelativeUnixPath> classpath = new LinkedHashSet<>(4);
+  private final List<PathPredicatePair> addedResources = new ArrayList<>();
+  private final List<PathPredicatePair> addedClasses = new ArrayList<>();
+  private final List<Path> addedDependencies = new ArrayList<>();
+  private final List<Path> addedOthers = new ArrayList<>();
```
chanseokoh: Previously these held `AbsoluteUnixPath` for path inside an image. Should this `RelativeUnixPath` instead of `Path`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -107,38 +126,19 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
...
-  /** Absolute path of directory containing dependencies on container. */
-  private AbsoluteUnixPath dependenciesPath;
-
-  /** The entrypoint classpath element corresponding to dependencies. */
-  private AbsoluteUnixPath dependenciesClasspath;
-
-  /** Absolute path of directory containing additional classpath files on container. */
-  private AbsoluteUnixPath othersPath;
+  private final LinkedHashSet<RelativeUnixPath> classpath = new LinkedHashSet<>(4);
+  private final List<PathPredicatePair> addedResources = new ArrayList<>();
```
chanseokoh: How about adding a single comment saying these fields are the file paths on the system rather than the paths inside images?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -40,6 +42,23 @@
 /** Creates a {@link JibContainerBuilder} for containerizing Java applications. */
 public class JavaContainerBuilder {
 
+  /** Holds a directory and a filter. */
+  private static class PathPredicatePair {
+
+    private final Path path;
```
chanseokoh: Ditto. I wonder if `RelativeUnixPath` is better.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -200,18 +191,14 @@ public JavaContainerBuilder addDependencies(List<Path> dependencyFiles) throws I
...
-      layerConfigurationsBuilder.addFile(
-          file.getFileName().toString().contains("SNAPSHOT")
-              ? LayerType.SNAPSHOT_DEPENDENCIES
-              : LayerType.DEPENDENCIES,
-          file,
-          dependenciesPath.resolve(
+      addedDependencies.add(
+          Paths.get(
               duplicates.contains(file.getFileName().toString())
                   ? file.getFileName().toString().replaceFirst("\\.jar$", "-" + Files.size(file))
```
chanseokoh: I think this new approach will fail because of this. Later, you will try to fetch the actual files by this name, which won't exist.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -248,7 +235,8 @@ public JavaContainerBuilder addResources(Path resourceFilesDirectory) throws IOE
    */
   public JavaContainerBuilder addResources(Path resourceFilesDirectory, Predicate<Path> pathFilter)
       throws IOException {
-    return addDirectory(resourceFilesDirectory, resourcesPath, LayerType.RESOURCES, pathFilter);
+    classpath.add(RESOURCES_CLASSPATH);
```
chanseokoh: Look like this constant path will be added multiple times if this method is called multiple times. Doesn't seem necessary to have duplication?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -272,7 +260,8 @@ public JavaContainerBuilder addClasses(Path classFilesDirectory) throws IOExcept
    */
   public JavaContainerBuilder addClasses(Path classFilesDirectory, Predicate<Path> pathFilter)
       throws IOException {
-    return addDirectory(classFilesDirectory, classesPath, LayerType.CLASSES, pathFilter);
+    classpath.add(CLASSES_CLASSPATH);
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -200,18 +191,14 @@ public JavaContainerBuilder addDependencies(List<Path> dependencyFiles) throws I
...
-          dependenciesPath.resolve(
+      addedDependencies.add(
+          Paths.get(
               duplicates.contains(file.getFileName().toString())
                   ? file.getFileName().toString().replaceFirst("\\.jar$", "-" + Files.size(file))
                       + ".jar"
                   : file.getFileName().toString()));
     }
-    classpath.add(dependenciesClasspath.toString());
+    classpath.add(DEPENDENCIES_CLASSPATH);
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -293,17 +282,8 @@ public JavaContainerBuilder addToClasspath(List<Path> otherFiles) throws IOExcep
...
-      if (Files.isDirectory(file)) {
-        layerConfigurationsBuilder.addDirectoryContents(
-            LayerType.EXTRA_FILES, file, path -> true, othersPath);
-      } else {
-        layerConfigurationsBuilder.addFile(
-            LayerType.EXTRA_FILES, file, othersPath.resolve(file.getFileName()));
-      }
-    }
-    classpath.add(othersPath.toString());
+    classpath.add(OTHERS_CLASSPATH);
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
chanseokoh: Oops. Never mind. Stale comment.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
TadCordle: `classpath` is a `LinkedHashSet`, won't contain duplicates.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -107,38 +125,21 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
...
-
-  /** Absolute path of directory containing application resources on container. */
-  private AbsoluteUnixPath resourcesPath;
-
-  /** Absolute path of directory containing classes on container. */
-  private AbsoluteUnixPath classesPath;
-
-  /** Absolute path of directory containing dependencies on container. */
-  private AbsoluteUnixPath dependenciesPath;
+  private final LinkedHashSet<RelativeUnixPath> classpath = new LinkedHashSet<>(4);
```
chanseokoh: How about `relativeClasspaths` (plurar)?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -387,23 +349,90 @@ public JibContainerBuilder toContainerBuilder() {
           "Failed to construct entrypoint because no files were added to the JavaContainerBuilder");
     }
 
+    JavaLayerConfigurations.Builder layerConfigurationsBuilder = JavaLayerConfigurations.builder();
+
+    // Add classes to layer configuration
+    for (PathPredicatePair file : addedClasses) {
```
chanseokoh: I think `directory` fits better than `file`. (Code below too.)

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -387,23 +349,90 @@ public JibContainerBuilder toContainerBuilder() {
...
+      throws NoSuchFileException, NotDirectoryException {
     if (!Files.exists(directory)) {
       throw new NoSuchFileException(directory.toString());
     }
     if (!Files.isDirectory(directory)) {
       throw new NotDirectoryException(directory.toString());
     }
-    layerConfigurationsBuilder.addDirectoryContents(layerType, directory, pathFilter, destination);
-    classpath.add(destination.toString());
+    addedFiles.add(new PathPredicatePair(directory, filter));
```
chanseokoh: `addedPaths` instead of `addedFiles`?

PR #1546, opened 2019-03-11
Send client_id for OAuth2 registry auth

### LINKED ISSUE #1545: OAuth2 flow fails against docker trusted registry (missing client_id)
**Description of the issue**:

When pulling/pushing to a docker trusted registry, authentication fails with error: 
400 Bad Request {"details":"invalid client ID"}

Docker registry (DTR) requires a client_id parameter to be passed with the refresh_token when making a login request.

From
https://docs.docker.com/registry/spec/auth/oauth/

> client_id
>(REQUIRED) String identifying the client. This client_id does not need to be registered with the authorization server but should be set to a meaningful value in order to allow auditing keys created by unregistered clients. Accepted syntax is defined in [RFC6749 Appendix A.1](https://tools.ietf.org/html/rfc6749#appendix-A.1)

**Expected behavior**:
It should be possible to pull a base image from an internal docker registry (DTR) using OAuth2 authentication flow.

**Steps to reproduce**: 
* Modify base image to pull from a docker trusted registry (https://docs.docker.com/datacenter/dtr/2.1/guides/)
* Login to registry
Observe that the login is stored as an OAuth token
```
docker-credential-wincred list
{"internalregistry":"\u003ctoken\u003e"}
```

* Run ```mvn jib:dockerBuild```
Build fails with error:
Unauthorized for internalregistry/base/custom-base-image: 400 Bad Request
[ERROR] {"details":"invalid client ID"}

**Environment**:
jib-maven-plugin 1.0.2
maven 3.5.0
jdk 1.8.0_131

```xml
<from>
   <image>internalregistry/base/custom-base-image</image>
</from>
```

**Log output**: 
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.0.2:dockerBuild (default-cli) on project my-project Build to Docker daemon failed, perhaps you should set a credential helper name with the configuration '<from><credHelper>' or set credentials for 'internalregistry' in your Maven settings: Unauthorized for internalregistry/base/custom-base-image: 400 Bad Request
[ERROR] {"details":"invalid client ID"}

chanseokoh:
Fixes #1545. `client_id` is a required parameter.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
@@ -269,7 +269,7 @@ public class RegistryAuthenticator {
     String serviceScope = getServiceScopeRequestParameters(scope);
     return isOAuth2Auth()
         ? serviceScope
-            + "&grant_type=refresh_token&refresh_token="
+            + "&client_id=jib&grant_type=refresh_token&refresh_token="
             // If OAuth2, credential.getPassword() is a refresh token.
             + Verify.verifyNotNull(credential).getPassword()
         : serviceScope;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
@@ -270,7 +270,7 @@ public class RegistryAuthenticator {
     return isOAuth2Auth()
         ? serviceScope
             // https://github.com/GoogleContainerTools/jib/pull/1545
-            + "&client_id=da031fe481a93ac107a95a96462358f9"
+            + "&client_id=jib.da031fe481a93ac107a95a96462358f9"
             + "&grant_type=refresh_token&refresh_token="
             // If OAuth2, credential.getPassword() is a refresh token.
             + Verify.verifyNotNull(credential).getPassword()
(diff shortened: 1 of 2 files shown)


chanseokoh:
My understanding is that this OAuth2 Client ID is a public identifier for an app. In our case, the app is Jib. Basically, I think we can [generate whatever value we want](https://stackoverflow.com/a/23671977/1701388). The [spec doc](https://docs.docker.com/registry/spec/auth/oauth/) says the auth server does not need to have ID registered, so I think any value will work.

My only concern is that, although the ID is public, this doc says it's a good practice to [use an ID that is not easily guessable](https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/). I think I'll generate some longer string.

loosebazooka:
It's be hard to make an unguessable client ID of we hardcore it. If we need to make it configurable in the future maybe we can deal with that then.

chanseokoh:
Let me use an MD5 hash `da031fe481a93ac107a95a96462358f9` generated from a string `jib`. (https://www.md5hashgenerator.com/) The ID is public, so it really shouldn't matter whether it's known. It is just that, it will be a slight deterrent if using a non-trivial ID. Like, if someone already knows an email address of someone less, it is slightly easier to attempt logging in by trying a brute-force password attack. And I think it's not really something that users want to configure.

briandealwis:
Since it's likely the client-id will appear in the registry logs, IMHO we should use something that clearly indicates Jib — the `TOOL_NAME` would be best, wouldn't it?

chanseokoh:
Good idea. The class is in jib-core, so the `TOOL_NAME` is `jib-core`. And I think it's not worth passing down `jib-maven-plugin` or `jib-gradle-plugin` depending on the execution context. So I think I'd just have `jib` as a prefix: `jib.da031fe481a93ac107a95a96462358f9`. It's clear from the ID that the application is Jib, and as suggested in the doc above, the entire ID is not able to guess without any context.

briandealwis:
`Containerizer` and `BuildConfiguration` provide `getToolName()`, but that isn't propagated into `RegistryAuthenticator`. We can revisit this if a need arises.

Issue #1471, opened 2019-02-10
Support scratch image


rmannibucau:
Hi guys,

with jlink, java will also use scratch image in FROM clause but ImageReference does not enable that. Is it possible to support it somehow?

loosebazooka:
Can you provide some more detail? Maybe an example of a workflow?

rmannibucau:
yes, basically I build with maven a native java image with jlink plugin so I get a custom java distro (of 17M in my case), then I use jib-core in a groovy script to create the image but scratch image is not supported and I have to use forks to workaround the limitation of jib.
The jib setup is simply to add the jlink zip exploded and set the right entry point.

loosebazooka:
Ah yes, I guess we force the user to select a FROM. Just curious: does the final jre generated by jlink really require nothing? I suspect it would at least need glibc?

Do you have an equivalent Dockerfile that works here?

Otherwise I would think that you would at least need to use `FROM gcr.io/distroless/base`?

rmannibucau:
Hmm, this can be the short term impl, if scratch -> switch to that. I didn't check if libc is needed or not, that's a good point!

rmannibucau:
A quick test with distroless/base ends with:

> bin/java: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory

coollog:
As a side note, here's a way to make your own scratch image to use:

`Dockerfile`:
```
FROM bash as base
RUN mkdir /empty
FROM scratch
COPY --from=base /empty /
```

```bash
docker build -t <image on your own registry> .
docker push <image on your own registry>
```

rmannibucau:
I created https://hub.docker.com/r/rmannibucau/jlink-base/tags as a workaround

coollog:
We will support using `scratch` as the base image in Jib. Using `scratch` will skip any base image steps in the Jib build.

loosebazooka:
I don't know if we should use a keyword like `scratch` in the core if that's what you're suggesting. Maybe just a mechanism to do this pro grammatically by augmenting the api.

coollog:
Yea, another option is to add a `Jib.fromScratch` method for this. The support for keyword `scratch` is just to match the expected behavior of using the official `scratch` "image" so that if users use it they don't just break.

TadCordle:
We added a `Jib.fromScratch` method, which will be available in the next release of jib-core.

PR #1550, opened 2019-03-11
Allow setting scratch base image in jib-core

TadCordle:
Adds `Jib.fromScratch()`, which signals most of `PullBaseImageStep` to be skipped (and effectively skips `PullAndCacheBaseImageLayersStep` and `PushLayersStep` for the base image as a result, since there are no layers to process).

Part of #1471. Should we support this in the plugins, or does that start to stray too much from being a "Java" image builder?


TadCordle:
Yeah, I can some tests. And the naming is an extension of `ImageReference.of()`, maybe something like `ImageReference.scratch()` would be less awkward..

Issue #1373, opened 2018-12-20
Use JavaContainerBuilder in plugins


TadCordle:
Followup to #1347

chanseokoh:
Hmm... if we really want to do this, I think we will need more changes to `JavaContianerBuilder`? For example, we cannot create an extra files layer with it.

TadCordle:
We can call `toContainerBuilder()` and addLayers to that, can't we?

chanseokoh:
Ah, it returns `JibContainerBuilder`. I missed that.

PR #1552, opened 2019-03-13
Allow setting destinations on JavaContainerBuilder

TadCordle:
For the WAR case, classes and dependencies need to be added to a non-default directory on the container. So this PR adds the ability to set the destination to send the added files to on the container.

This is a small step towards #1373.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.jib.api;
 
+import com.google.cloud.tools.jib.ProjectInfo;
 import com.google.cloud.tools.jib.filesystem.AbsoluteUnixPath;
 import com.google.cloud.tools.jib.filesystem.RelativeUnixPath;
 import com.google.cloud.tools.jib.frontend.JavaEntrypointConstructor;
@@ -136,7 +137,7 @@ public class JavaContainerBuilder {
       JavaEntrypointConstructor.DEFAULT_RELATIVE_RESOURCES_PATH_ON_IMAGE;
   private RelativeUnixPath dependenciesDestination =
       JavaEntrypointConstructor.DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE;
-  private RelativeUnixPath classpathDestination = RelativeUnixPath.get("classpath");
+  private RelativeUnixPath othersDestination = RelativeUnixPath.get("classpath");
   @Nullable private String mainClass;
 
   private JavaContainerBuilder(JibContainerBuilder jibContainerBuilder) {
@@ -204,11 +205,11 @@ public class JavaContainerBuilder {
    * Sets the destination directory of additional classpath files added to the container (relative
    * to the app root).
    *
-   * @param classpathDestination the additional classpath directory, relative to the app root
+   * @param othersDestination the additional classpath directory, relative to the app root
    * @return this
    */
-  public JavaContainerBuilder setClasspathDestination(RelativeUnixPath classpathDestination) {
-    this.classpathDestination = classpathDestination;
+  public JavaContainerBuilder setOthersDestination(RelativeUnixPath othersDestination) {
+    this.othersDestination = othersDestination;
     return this;
   }
 
@@ -452,12 +453,12 @@ public class JavaContainerBuilder {
     for (Path path : addedOthers) {
       if (Files.isDirectory(path)) {
         layerConfigurationsBuilder.addDirectoryContents(
-            LayerType.EXTRA_FILES, path, path1 -> true, appRoot.resolve(classpathDestination));
+            LayerType.EXTRA_FILES, path, path1 -> true, appRoot.resolve(othersDestination));
       } else {
         layerConfigurationsBuilder.addFile(
             LayerType.EXTRA_FILES,
             path,
-            appRoot.resolve(classpathDestination).resolve(path.getFileName()));
+            appRoot.resolve(othersDestination).resolve(path.getFileName()));
       }
     }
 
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -126,20 +121,26 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
...
   private final List<Path> addedOthers = new ArrayList<>();
 
-  private AbsoluteUnixPath appRoot;
+  private AbsoluteUnixPath appRoot = AbsoluteUnixPath.get("/app");
+  private RelativeUnixPath classesDestination =
+      JavaEntrypointConstructor.DEFAULT_RELATIVE_CLASSES_PATH_ON_IMAGE;
+  private RelativeUnixPath resourcesDestination =
+      JavaEntrypointConstructor.DEFAULT_RELATIVE_RESOURCES_PATH_ON_IMAGE;
+  private RelativeUnixPath dependenciesDestination =
+      JavaEntrypointConstructor.DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE;
```
chanseokoh: In the end, it may be possible to remove `JavaEntrypointConstructor` and move these constants to here. We will see.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -163,6 +164,54 @@ public JavaContainerBuilder setAppRoot(AbsoluteUnixPath appRoot) {
...
+  }
+
+  /**
+   * Sets the destination directory of additional classpath files added to the container (relative
+   * to the app root).
+   *
+   * @param classpathDestination the additional classpath directory, relative to the app root
+   * @return this
+   */
+  public JavaContainerBuilder setClasspathDestination(RelativeUnixPath classpathDestination) {
```
chanseokoh: `setOthersDestination`, and update the Javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -126,20 +121,26 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
...
 
-  private AbsoluteUnixPath appRoot;
+  private AbsoluteUnixPath appRoot = AbsoluteUnixPath.get("/app");
+  private RelativeUnixPath classesDestination =
+      JavaEntrypointConstructor.DEFAULT_RELATIVE_CLASSES_PATH_ON_IMAGE;
+  private RelativeUnixPath resourcesDestination =
+      JavaEntrypointConstructor.DEFAULT_RELATIVE_RESOURCES_PATH_ON_IMAGE;
+  private RelativeUnixPath dependenciesDestination =
+      JavaEntrypointConstructor.DEFAULT_RELATIVE_DEPENDENCIES_PATH_ON_IMAGE;
+  private RelativeUnixPath classpathDestination = RelativeUnixPath.get("classpath");
```
chanseokoh: `classpathDestination` sounds confusing. Probably `othersDestination`, since that's what we call currently.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -263,7 +312,7 @@ public JavaContainerBuilder addToClasspath(List<Path> otherFiles) throws IOExcep
         throw new NoSuchFileException(file.toString());
       }
     }
-    relativeClasspaths.add(OTHERS_CLASSPATH);
+    classpathOrder.add(LayerType.EXTRA_FILES);
```
chanseokoh: Huh, now I see we've been putting these "others" files into the `EXTRA_FILES` layer. When we get to use this class in our plugins, this won't fit. 1) Extra files in the plugins will go into the root, not `/app/classpath`; and 2) Plugins won't add a classpath for the extra files root (`/`). You will need to think about this divergence too.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -403,19 +452,33 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
...
+        case CLASSES:
+          classpathElements.add(appRoot.resolve(classesDestination).toString());
+          break;
+        case RESOURCES:
+          classpathElements.add(appRoot.resolve(resourcesDestination).toString());
+          break;
+        case DEPENDENCIES:
+          classpathElements.add(appRoot.resolve(dependenciesDestination).resolve("*").toString());
+          break;
+        default:
```
chanseokoh: I think below is safer:
```java
case EXTRA_FILES:
  ...
default:
  throw new RuntimeException("Bug in jib-core. Please report bug ...");
}
```

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
TadCordle: These aren't actually being added to the extra files layer, I'm just using the enum to keep track of the classpath order. Maybe this is a bit too hacky?

chanseokoh: I think we add "others" to the "extra files" layer.

```java
        layerConfigurationsBuilder.addDirectoryContents(
            LayerType.EXTRA_FILES, path, path1 -> true, appRoot.resolve(OTHERS_CLASSPATH));
```

chanseokoh: Let's keep this in mind. This will come to us soon.

TadCordle: Oh yeah, you're right. They do get added to the extra layer, but `addToClasspath` isn't meant to be the same mechanism as the extra files configuration. So yeah, we'll probably have to find some other way to handle this.

PR #1553, opened 2019-03-13
Allow skipping main class in JavaContainerBuilder

TadCordle:
After #1552.

Discussed with @chanseokoh and he mentioned that the WAR case does not want to set the entrypoint using a main class. So in order for this class to be useful for building WAR containers, we need a way to skip setting the entrypoint. So, this PR will skip constructing the entrypoint if the user never calls `JavaContainerBuilder#setMainClass()`.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
@@ -387,11 +387,12 @@ public class JavaContainerBuilder {
    * @throws IOException if building the {@link JibContainerBuilder} fails.
    */
   public JibContainerBuilder toContainerBuilder() throws IOException {
-    if (mainClass == null) {
+    if (mainClass == null && !jvmFlags.isEmpty()) {
       throw new IllegalStateException(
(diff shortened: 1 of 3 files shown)


chanseokoh:
This is also needed in the case when someone or we want to inherit the entrypoint in a non-WAR case too.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -461,32 +461,32 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
...
+          case CLASSES:
+            classpathElements.add(appRoot.resolve(classesDestination).toString());
+            break;
+          case RESOURCES:
+            classpathElements.add(appRoot.resolve(resourcesDestination).toString());
+            break;
+          case DEPENDENCIES:
+            classpathElements.add(appRoot.resolve(dependenciesDestination).resolve("*").toString());
+            break;
+          default:
```
chanseokoh: Regression. Restore `case EXTRA_FILES`.

TadCordle: Oh oops.

PR #1557, opened 2019-03-14
Add instructions for importing source into Eclipse

briandealwis:
When building a new workspace, I had forgotten what the steps were to properly bring Jib into Eclipse. I also recall seeing someone asking for the instructions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,6 +49,39 @@ To build, use the provided `build.sh` which builds and tests each of the compone
 
 # Development Tips
 
+## Configuring Eclipse
+
+Jib is a mix of Gradle and Maven projects, and the projects reference
+source from their dependencies directly using `sourceSets` and
+`build-helper-maven-plugin`.  When importing these projects into
+Eclipse, the Gradle and Maven tooling faithfully replicate the
+direct source referencs, but this has a side-effect that classes
+appear multiple times, in the originating project and in the
+subprojects.  These instructions first import the projects, and
+then replace the source-set-style references to use project references
+instead.
+
+  1. Ensure you have installed the Gradle tooling for Eclipse, called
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -49,6 +49,46 @@ To build, use the provided `build.sh` which builds and tests each of the compone
...
+     to use the Eclipe Smart Import feature and specify the root of the Jib project.
+     Import the `jib-maven-plugin` project.
+  3. **Import the Gradle projects:** Buildship does [not yet support
+     Eclipse Smart Import](https://github.com/eclipse/buildship/issues/356).
+     Use _File &rarr; Import &rarr; Gradle &rarr; Existing Gradle Project_
+     and import each of `jib-core`, `jib-plugins-common`, and `jib-gradle-plugin`.
+  4. **Turn source-set references to project references:** For each of 
+     `jib-plugins-common`, `jib-maven-plugin`, and `jib-gradle-plugin`:
+     - Right-click on the correponding project and select _Properties &rarr; Java Build Path_
+     - Open the _Source_ panel and remove all _linked_ source folders: these are like
```
chanseokoh: FYI:

Sometimes (but rarely), you'll have to refresh/update these projects, e.g., when we upgrade dependencies. From my experience, Buildship restores all of these, which is annoying. So I just do minimal changes to `jib-gradle-plugin` only (excluding `src/test/resources/gradle/projects` and `src/integration-test/resources/gradle/projects` from source), which makes Eclipse happy. I even automated this with a small script.

I haven't had any major problem with classes appearing in multiple places.

One problem (may not be true anymore) I remember was that, if I remove all linked source folders, I couldn't run unit tests in Eclipse in many cases, where they call `getResource()` to get test resources.

PR #1559, opened 2019-03-15
Rename from Credential.basic() to Credential.from()

### LINKED ISSUE #1526: Rename Credential.basic(username, password)
After the support of OAuth2 in #1511, we now have an OAuth2 flow in addition to the basic auth flow. (FTR, in OAuth2 case, a `Credential` instance holds the fixed constant `<token>` as a username and the OAuth2 refresh token as a password.) The name `Credential.basic()` does not fit well with the OAuth2 flow.

Maybe just `Credential.of(username, password)`? Or, we could turn `Credential` from a simple holder to a more high-level abstraction for both basic auth and OAuth. But since there's no need to directly instantiate an OAuth2 `Credential` instance (the Docker spec returns the explicit username `<token>` and that's how we instantiate the simple holder in practice), so I'm leaning toward the current simple holder.

chanseokoh:
Fixes #1526. Simple renaming.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/credentials/Credential.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/credentials/Credential.java
@@ -18,7 +18,6 @@ package com.google.cloud.tools.jib.configuration.credentials;
 
 import java.util.Objects;
 
-// TODO: Move to lower-level package - probably at same level as Authorization.
 /** Holds credentials (username and password). */
 public class Credential {
   // If the username is set to <token>, the secret would be a refresh token.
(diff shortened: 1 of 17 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/credentials/Credential.java
```diff
@@ -21,7 +21,7 @@
 // TODO: Move to lower-level package - probably at same level as Authorization.
```
TadCordle: Curious, is there a reason we haven't done this yet?

chanseokoh: No idea what this means. This looks like an old comment. `Authorization` is under the `.jib.http` package (which doesn't look wrong), and this class does not fit into that package. I'll remove the comment.

PR #1564, opened 2019-03-19
Propagate architecture and os from base to target images

### LINKED ISSUE #1547: Inherit OS/Architecture from base image
We set [OS to "amd64" and Architecture "linux"](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java#L87-L91), but we should probably inherit them from the base image. I expect this will enable, for example, creating an ARM image.

A lot of official Docker Hub images support multiple architectures. [OpenJDK](https://hub.docker.com/_/openjdk) for example have the following images.

> **Supported architectures: (more info)**
> amd64, arm32v5, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x, windows-amd64

If you have to, you can force referencing/pulling architecture-specific images by, e.g., [`amd64/openjdk`](https://hub.docker.com/r/amd64/openjdk/) or [`arm64v8/openjdk`](https://hub.docker.com/r/arm64v8/openjdk/). However, `docker pull openjdk` will just pull the matching architecture on all supported platforms, and it should work.

briandealwis:
- Adds `architecture` and `os` fields to `Image`/`ImageBuilder` and `ContainerConfigurationTemplate`, defaulting to current `amd64`/`linux`
- Modifies `ImageToJsonTranslator` and `JsonToImageTranslator` to propagate `architecture`/`os`
- Modifies `BuildImageStep` to propagate `architecture`/`os` from the base image to the built image
- Adds some explicit tests, and modifies existing tests to us `wasm/js` to ensure we're not just picking up the default `amd64/linux`
- Verified by hand by building against image `s390x/openjdk:11-stretch`

Note that this PR does not expose any facilities for changing the arch/os.  I'll leave that to a separate PR.

Fixes #1547


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -172,6 +172,28 @@ public void setCreated(@Nullable String created) {
     this.created = created;
   }
 
+  /**
+   * Set the architecture for which this container was built. See the <a
```
TadCordle: nit: Sets (same for `setOs` javadoc)

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ContainerConfigurationTemplate.java
```diff
@@ -260,6 +282,28 @@ String getCreated() {
     return created;
   }
 
+  /**
+   * Return the architecture for which this container was built. See the <a
```
TadCordle: nit: Returns (same for `getOs` javadoc)

jib-core/src/test/java/com/google/cloud/tools/jib/image/ImageTest.java
```diff
@@ -76,4 +77,23 @@ public void test_smokeTest() throws LayerPropertyNotFoundException {
...
+    Assert.assertEquals(Collections.emptyList(), image.getHistory());
+  }
+
+  @Test
+  public void testOsArch() {
+    Image<Layer> image =
+        Image.builder(V22ManifestTemplate.class).setArchitecture("wasm").setOs("js").build();
+    Assert.assertEquals("wasm", image.getArchitecture());
+    Assert.assertEquals("js", image.getOs());
+    Assert.assertEquals(Collections.emptyList(), image.getLayers());
```
TadCordle: What's the reason for testing the layers/history size in the os/arch test?

briandealwis: Oops.  I originally had this test the must-set items.

chanseokoh:
> Note that this PR does not expose any facilities for changing the arch/os. I'll leave that to a separate PR.

I don't think there is a need to change the arch/os. If they wanted a different arch/os, they should use the image with the arch/os they want. For example, `arm64v8/openjdk` or `winmad64/openjdk`.

Now that said, I think now this PR undermines reproducibility, just a bit. That is, when the base image is `openjdk`, Jib will build an AMD image on AMD and an ARM image on ARM. We need to think about this.

chanseokoh:
> Jib will build an AMD image on AMD and an ARM image on ARM.

Actually, I'm not sure if this is the case with Jib. OTOH, I believe Docker build automatically pulls in the matching os/arch.

briandealwis:
I don't think it undermines reproducibility: we propagate the os/arch from the base image.  Docker Hub doesn't silently remap between architectures, so we should get the same base image for the same image ref.

I can totally get that `docker build` may silently swap os/arch.  I don't see how it can work otherwise since it will be executing binaries for the current system.

But we should document the difference.

chanseokoh:
> I don't think it undermines reproducibility: we propagate the os/arch from the base image. Docker Hub doesn't silently remap between architectures, so we should get the same base image for the same image ref.

Ah, OK. That makes sense.

Issue #1562, opened 2019-03-18
Create time of building docker images using Jib


tankdeper:
**Description of the issue**:  I used `jib` to build up my java application as docker image, when I pull request from docker, I found that the `creatime` is extremely ridiculous, it shows `49 years ago`

**Expected behavior**: it should be the same or similar to the time as I built

**Steps to reproduce**: 

**Environment**: windows10, Maven

**Log output**:  
```zhengtank/hello-world                                  latest              5fc7a163adf7        49 years ago        102MB```

**Additional Information**:  You all could pull request this image from registry.hub.docker.com, and you'll see `49 years ago`

TadCordle:
Jib strips timestamps in order to maintain reproducibility. You can bypass this behavior by setting the `container.useCurrentTimestamp` configuration parameter to `true`. See the [FAQ](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#why-is-my-image-created-48-years-ago) for more details.

tankdeper:
@TadCordle Excuse me! Could you please explain what reproducibility means and what kind of situation we need this?

chanseokoh:
@tankdeper say that you ran `mvn jib:dockerBuild` with `<container><userCurrentTimestamp>true`. It will build an image whose creation time is the current timestamp. Now, a few seconds later, you run `mvn jib:dockerBuild` again without changing anything in your project. That is, the image created by the second `mvn jib:dockerBuild` build is identical to the first image except the timestamp. However, because the two images have different timestamps, tools, systems, and infrastructures (Docker daemon, Docker Hub, registries, ECR, Kubernetes, etc) will see the two as different images. For example, now `docker images` will list two images, whereas without `<useCurrentTimestamp>`, the Docker daemon would have thought that the image already existed and resulted in no-op. Likewise, when you push the second image to a remote registry, the registry will see that as a new image that doesn't exist. Tools will think it is a different image and download it again and cache it as a different image, even if it is the exactly same image except the timestamp. So, running `mvn jib:dockerBuild` multiple times is not a reproducible build in the sense that it doesn't create the same image each time.

tankdeper:
Thanks a lots! Running `mvn jib:build` is the same to this, right?

chanseokoh:
[...]

PR #1565, opened 2019-03-19
Add FAQ item on reproducible builds

briandealwis:
Questions around timestamps has come up a few times, so I took @chanseokoh's answer to #1562 and turned it into a FAQ.  Too wordy?


chanseokoh:
Umm... maybe too colloquial? And I said something about downloading and caching, but I think this could be misinterpreted, because I believe the file layers are same so the only data that you'll download is a small metadata (and this claim may not be 100% accurate either)?

chanseokoh:
Oh, on second thought, this affects timestamps of files in the image too, so I guess often the classes and resources layers would be different too.

chanseokoh:
> Oh, on second thought, this affects timestamps of files in the image too

Actually, `<useCurrentTimestamp>` doesn't affect the file timestamps. We always set them to [1+epoch](https://github.com/GoogleContainerTools/jib/blob/v1.0.2-maven/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java#L124).

docs/faq.md
```diff
@@ -413,6 +413,34 @@ For reproducibility purposes, Jib sets the creation time of the container images
 jib.container.useCurrentTimestamp = true
 ```
 
+#### Please tell me more about reproducibility!
+
+_Reproducible_ means that building and rebuilding with Jib should
```
TadCordle: If we do want this (I think it looks good, it's a nice explanation) I think we should remove all these extra newlines, or it could potentially mess with the formatting for different screen sizes.

chanseokoh: OK. I'll leave @TadCordle for wording suggestion.

docs/faq.md
```diff
@@ -413,6 +413,34 @@ For reproducibility purposes, Jib sets the creation time of the container images
...
+`<container><userCurrentTimestamp>true`. Jib will build an image
+whose creation time is the current timestamp. Now, a few seconds
+later, you run `mvn jib:dockerBuild` again without changing anything
+in your project. That is, the image created by the second `mvn
+jib:dockerBuild` build is identical to the first image except for
+the timestamp. But since the two images have different timestamps,
+the images will appear to be different to tools and infrastructure
+such Docker daemon, Docker Hub, registries, Kubernetes, etc). For
+example, now `docker images` will list two images, whereas without
+<useCurrentTimestamp>, the Docker daemon would have thought that
```
chanseokoh: Use back ticks on `<useCurrentTimestamp>`.

docs/faq.md
```diff
@@ -413,6 +413,34 @@ For reproducibility purposes, Jib sets the creation time of the container images
...
+the image already existed and resulted in no-op. Likewise, when you
+push the second image to a remote registry, the registry will treat
+it as a new image that doesn't exist. Tools will think it is a
+different image and download it again and cache it as a different
```
chanseokoh: As I said, this may give the impression that it will download an entire image again, while in reality I think it will be just a manifest. How about removing "download it again"?

briandealwis:
Did a quick rewrite.

docs/faq.md
```diff
@@ -413,6 +413,15 @@ For reproducibility purposes, Jib sets the creation time of the container images
...
+
+To ensure that a Jib build is reproducible — that the rebuilt container image has the same digest — Jib adds files and directories in a consistent order, and sets consistent creation- and modification-times and permissions for all files and directories.  Jib also ensures that the image metadata is recorded in a consistent order, and that the container image has a consistent creation time.  To ensure consistent times, files and directories are recorded as having a creation and modification time of 1 second past the Unix Epoch (1970-01-01 00:00:01.000 UTC), and the container image is recorded as being created on the Unix Epoch.  By setting `container.useCurrentTimestamp=true` then rebuilding an image will produce a different timestamp for the image creation time, and so the container images will have different digests and appear to be different.
```
TadCordle: "By setting `container.useCurrentTimestamp=true`..." should probably just be "Setting `container.useCurrentTimestamp=true`..."

PR #1569, opened 2019-03-20
Don't add empty layers in JavaLayerConfigurations

TadCordle:
Before this PR, in `JavaLayerConfigurations`, all 5 java layers (classes, resources, dependencies, snapshot dependencies, and extra files) are added automatically, whether there are files in the layers or not. This PR skips the empty layers.

Small bit towards #1373.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
@@ -164,7 +164,6 @@ public class JavaLayerConfigurations {
         AbsoluteUnixPath basePathInContainer,
         Map<AbsoluteUnixPath, FilePermissions> permissionsMap)
         throws IOException {
-
       if (layerBuilders.get(layerType) == null) {
         layerBuilders.put(layerType, LayerConfiguration.builder());
       }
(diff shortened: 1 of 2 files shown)


TadCordle:
I only realized this was creating empty layers while fixing tests working on #1373; realized `JavaContainerBuilder.toContainerBuilder` returns a `JibContainerBuilder` that already has the 5 java layers, including `extra files` even if I wasn't adding extra files to the classpath. So when I tried to add extra directory contents to the `JibContainerBuilder` I got after creating a `JavaContainerBuilder`, it ended up with 2 `extra files` layers (one empty classpath one, and one with the extra directory files). So, can't say I've considered other places yet.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -111,8 +106,10 @@ public Builder addFile(
         Path sourceFile,
         AbsoluteUnixPath pathInContainer,
         @Nullable FilePermissions permissions) {
-      Preconditions.checkNotNull(layerBuilders.get(layerType))
-          .addEntry(sourceFile, pathInContainer, permissions);
+      if (!layerBuilders.containsKey(layerType)) {
+        layerBuilders.put(layerType, LayerConfiguration.builder());
+      }
+      layerBuilders.get(layerType).addEntry(sourceFile, pathInContainer, permissions);
```
briandealwis: These could be simplified using `layerBuilders.computeIfAbsent()`

PR #1571, opened 2019-03-21
add lasttimestamp to LayerEntry to ensure the timestamp can be controlled for key files

rmannibucau:
I assume this is close to the issue https://github.com/GoogleContainerTools/jib/issues/1079 so I didn't recreate a ticket but my use case is a bit different.
I have a dockerized app using libraries relying on file lastModifiedDate to create a kind of "testable timestamp" to check if some caching should be invalidated or not.
Current impl of jib-core setModTime to 1s after epoch which breaks that usage.
To ensure it is reproducable I propose to add the timestamp in the API of addEntry(file), with a minimum of 1s after the epoch. 
This way people can choose to hardcode it or use a custom version if they know - like using maven metadata if the file comes from a maven release.

Hope it makes sense.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
@@ -75,7 +75,7 @@ public class LayerEntry {
                 ? FilePermissions.DEFAULT_FOLDER_PERMISSIONS
                 : FilePermissions.DEFAULT_FILE_PERMISSIONS
             : permissions;
-    this.lastModifiedTime = Math.max(lastModifiedTime, 1000);
+    this.lastModifiedTime = lastModifiedTime;
   }
 
   /**
@@ -90,7 +90,7 @@ public class LayerEntry {
    */
   public LayerEntry(
       Path sourceFile, AbsoluteUnixPath extractionPath, @Nullable FilePermissions permissions) {
-    this(sourceFile, extractionPath, permissions, -1);
+    this(sourceFile, extractionPath, permissions, 1000);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


rmannibucau:
I signed it!

chanseokoh:
Thanks @rmannibucau. I think this is a step required toward #1079. Much appreciated.

We will look into it. Expect some minor change requests.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -69,6 +75,30 @@ public LayerEntry(
                 ? FilePermissions.DEFAULT_FOLDER_PERMISSIONS
                 : FilePermissions.DEFAULT_FILE_PERMISSIONS
             : permissions;
+    this.lastModifiedTime = Math.max(lastModifiedTime, 1000);
```
chanseokoh: No need for `Max.max()` after passing 1000 in the `LayerEntry` constructor.

rmannibucau: It is still needed because calling addEntry you can pass your own value and what I understood if we don't want to enable 0 (or < 0) so think it is better to keep it here and it does not really hurt anyway, wdyt?

rmannibucau:
Applied most of the comments except the max() one - added a question on this one but think it is worth to keep it.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
chanseokoh: From the plugin usage perspective, this can be meaningful, but from the API design standpoint, I think this library should not incorporate such a high level processing. I think the API should just do what it is told to do so.

rmannibucau: Works for me, did the change

Issue #1065, opened 2018-09-28
Try alternative mini framework for constructing Java layers


chanseokoh:
From https://github.com/GoogleContainerTools/jib/pull/1062#pullrequestreview-159975210

PR #1572, opened 2019-03-21
Use JavaContainerBuilder in plugins

### LINKED ISSUE #1373: Use JavaContainerBuilder in plugins

TadCordle:
Fixes #1373.

Notable changes:

* Replaced all uses of `JavaLayerConfigurations` in plugins with `JavaContainerBuilder` (aside from a couple global constants; should these be moved?)
* Added `ProjectProperties.getClassFiles()`; since project properties no longer hold layer configurations, we need another way to pass a list of class files to `MainClassResolver`
* `GradleLayerConfigurations` and `MavenLayerConfigurations` were consolidated enough that I decided to just move their logic straight into `GradleProjectProperties`/`MavenProjectProperties`
* As such, `GradleLayerConfigurationsTest` and `MavenLayerConfigurationsTest` were moved into their respective `ProjectProperties` tests
* The extra files layer is added in `PluginConfigurationProcessor` instead of the `ProjectProperties`, to avoid passing around too many parameters/share code between the plugins
* Some redundant info-level log messages were removed; they showed which files were added to the container, but most of the same information is logged on `containerize()`


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -0,0 +1,116 @@
...
+            .setDependenciesDestination(RelativeUnixPath.get("WEB-INF/lib"));
+
+    if (Files.exists(explodedWar)) {
+      javaContainerBuilder.addResources(explodedWar, isResource);
+    }
+    if (Files.exists(webInfClasses)) {
+      javaContainerBuilder.addClasses(webInfClasses, isClassFile);
+    }
+    if (Files.exists(webInfLib)) {
+      javaContainerBuilder.addDependencies(webInfLib, isDependency);
```
chanseokoh: `isDependency`, which checks if the given path starts with `WEB-INF/lib`, seems unnecessary to pass, since the directory root we pass here is `webInfLib`. Then, I guess you don't need to introduce the new `addDependencies()` method?

chanseokoh: Oh, I see you still need to pass a list of JARs. But the filter is not necessary.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/JavaLayerConfigurations.java
```diff
@@ -56,7 +56,7 @@
     }
 
     @VisibleForTesting
```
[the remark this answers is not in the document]
TadCordle: Yup

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -525,7 +525,7 @@ JibContainer containerize(
    * @throws IOException if an I/O exception occurs
    */
   @VisibleForTesting
```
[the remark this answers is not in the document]
TadCordle: Yeah, only went public because it's used in plugin tests.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
TadCordle: The filter is used earlier to define the resources filter (https://github.com/GoogleContainerTools/jib/pull/1572/files/5c0b824d12984cefefa3f2dd90b567a9869f95e5#diff-945af34eb6dbffe51159846ba6bfa9ccR85), but yeah I guess it's not necessary here.

I think having an `addDependencies` method that takes a filter is still nice for consistency among the other methods. Maybe I should even add another method that takes no filter and walks a directory, much like the classes/resources methods?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -127,7 +128,7 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
   // Keeps track of files to add to the image, by system path
   private final List<PathPredicatePair> addedResources = new ArrayList<>();
   private final List<PathPredicatePair> addedClasses = new ArrayList<>();
-  private final List<Path> addedDependencies = new ArrayList<>();
+  private final List<PathPredicatePair> addedDependencies = new ArrayList<>();
```
chanseokoh: I think this can be left with `Path`. You use `predicate` only in the new `addDependencies` method and not later.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -64,23 +71,18 @@
...
-          project,
-          logger,
-          GradleLayerConfigurations.getForProject(
-              project, logger, extraDirectory, convertPermissionsMap(permissions), appRoot));
-
-    } catch (IOException ex) {
-      throw new GradleException("Obtaining project build output files failed", ex);
-    }
+    return new GradleProjectProperties(
+        project, logger, extraDirectory, convertPermissionsMap(permissions), appRoot);
```
chanseokoh: I have an idea. What if we don't pass `extraDirectory` and `permissions` but add the layer outside `ProjectProperties`? I think `extraDirectory` can be not project-specific and independent of Gradle/Mavan. I've not been happy that we have to pass these when instantiating `ProjectProperties`. Also, it's that you add the layer outside `JavaContainerBuilder`, so if it doesn't have a lot of code duplication, it is natual to add the layer separately later?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
briandealwis: This approach is mildly dangerous in that the user might provide a set of filters that misses some files, such that some dependencies don't make it through.  We're trying to intelligently place the file contents in appropriate layers, rather than filter out files.  As it's hard to prove our filters don't miss anything, I wonder if it might not be safer to walk the files with a function that returns a layer-indicator for each file.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -127,7 +129,7 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
   // Keeps track of files to add to the image, by system path
   private final List<PathPredicatePair> addedResources = new ArrayList<>();
```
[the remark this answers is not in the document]
TadCordle: The implementation basically is its purpose, though; it's just intended to be a tuple containing a path and a predicate. The path contents aren't actually filtered yet at this point, so I think something like `FilteredPathContents` might be misleading.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
chanseokoh: @briandealwis I remember a similar, if not same, idea: https://github.com/GoogleContainerTools/jib/issues/1065

From what I remember, I tried the idea, but there were some problems which I don't remember exactly. Things might have changed though.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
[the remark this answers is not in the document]
briandealwis: Well it carries a path and a predicate... :-)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
TadCordle: It looks like resolving these is still a bit plugin specific; the project directory is needed to figure out the extra directory (which is different between maven and gradle), and each of the plugins parses the permissions map differently. So it seems like we'll still need to pass these parameters around a bit. But maybe it'd make more sense to do it using `RawConfiguration` instead of `ProjectProperties`, and still handle adding the extra directory in plugins-common.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -205,6 +205,14 @@ static PluginConfigurationProcessor processCommonConfiguration(
       jibContainerBuilder.setCreationTime(Instant.now());
     }
 
+    // Adds all the extra files.
+    if (Files.exists(rawConfiguration.getExtraDirectory())) {
+      jibContainerBuilder.addLayer(
+          JavaContainerBuilderHelper.extraDirectoryLayerConfiguration(
+              rawConfiguration.getExtraDirectory(),
+              rawConfiguration.getExtraDirectoryPermissions()));
```
chanseokoh: What do you think of doing `TaskCommon.convertPermissionsMap()` here instead of `RawConfiguration` calling it? I have a desire to have `RawConfiguration` as a very think adapter that merely returns config values without any processing or validation.

TadCordle: Makes sense.

TadCordle: Ah wait, I remember why it's like this, it's because Gradle and Maven define the permissions map differently; Maven uses a `List<PermissionConfiguration>`, and Gradle uses a `Map<String, String>`. So it doesn't seem like we can just directly return these from `RawConfiguration` without any processing, since the actual raw type is different between the two plugins.

chanseokoh: Ah, okay. Since they have different types, I think this conversion is necessary to be able to unify them.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,87 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+    }
+  }
+
+  @Override
+  public ImmutableList<Path> getClassFiles() throws IOException {
+    // TODO: Consolidate with getContainerBuilderWithLayers
+    JavaPluginConvention javaPluginConvention =
+        project.getConvention().getPlugin(JavaPluginConvention.class);
+    SourceSet mainSourceSet = javaPluginConvention.getSourceSets().getByName(MAIN_SOURCE_SET_NAME);
+    FileCollection classesOutputDirectories = mainSourceSet.getOutput().getClassesDirs();
```
chanseokoh: From the code in `getContainerBuilderWithLayers()`, seems like this may return directories that don't exist. Will this (`DirectoryWalker(dir.toPath()).walk()`, I guess) work without any problem?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -200,23 +184,55 @@ static int getVersionFromString(String versionString) {
...
+          "Obtaining project build output files failed; make sure you have compiled your project "
+              + "before trying to build the image. (Did you accidentally run \"mvn clean "
+              + "jib:build\" instead of \"mvn clean compile jib:build\"?)",
+          ex);
+    }
+  }
+
+  @Override
+  public ImmutableList<Path> getClassFiles() throws IOException {
+    return new DirectoryWalker(Paths.get(project.getBuild().getOutputDirectory())).walk().asList();
```
chanseokoh: This will include non-`.class` files (e.g., resource files). Will this be OK?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,87 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+
+      return javaContainerBuilder.toContainerBuilder();
+
+    } catch (IOException ex) {
+      throw new GradleException("Obtaining project build output files failed", ex);
+    }
+  }
+
+  @Override
+  public ImmutableList<Path> getClassFiles() throws IOException {
```
chanseokoh: Unless there's a reason, I prefer `List<>`, returning a a new list instance. The caller doesn't have to be constrained.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
TadCordle: For our use case, it should be, since the main class finder already filters for the ".class" extension. But I suppose the method name is misleading.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
TadCordle: I'll add a test.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -245,6 +247,40 @@ public JavaContainerBuilder addDependencies(Path... dependencyFiles) throws IOEx
     return addDependencies(Arrays.asList(dependencyFiles));
   }
 
+  /**
+   * Adds the contents of a dependency JAR directory to the image. Duplicate JAR filenames are
```
briandealwis: This seems confusing: we now have 4 variants of `addDependencies`, but some handle directories whereas others don't.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -245,6 +247,40 @@ public JavaContainerBuilder addDependencies(Path... dependencyFiles) throws IOEx
...
+   * renamed with the filesize in order to avoid collisions.
+   *
+   * @param dependenciesDirectory the directory containing dependency JARs to add to the image
+   * @param pathFilter filter that determines which files (not directories) should be added
+   * @return this
+   * @throws IOException if adding the layer fails
+   */
+  public JavaContainerBuilder addDependencies(
```
briandealwis: Hmm, exposing the predicate as an API method seems unnecessary.  I wonder if this shouldn't just be an internal method.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
TadCordle: Actually yeah, this'll fail, I'll change it.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,87 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+          JavaContainerBuilder.from(baseImage).setAppRoot(appRoot);
+
+      // Adds resource files
+      if (Files.exists(resourcesOutputDirectory)) {
+        javaContainerBuilder.addResources(resourcesOutputDirectory);
+      }
+
+      // Adds class files
+      for (File classesOutputDirectory : classesOutputDirectories.filter(File::exists)) {
+        if (classesOutputDirectory.exists()) {
```
chanseokoh: Probably this `exist()` is unnecessary.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,87 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+          javaContainerBuilder.addClasses(classesOutputDirectory.toPath());
+        }
+      }
+      if (classesOutputDirectories.filter(File::exists).isEmpty()) {
```
chanseokoh: Perhaps just `classesOutputDirectories.filter.isEmpty()` because it is already filtered?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,87 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+      JavaContainerBuilder javaContainerBuilder =
+          JavaContainerBuilder.from(baseImage).setAppRoot(appRoot);
+
+      // Adds resource files
+      if (Files.exists(resourcesOutputDirectory)) {
+        javaContainerBuilder.addResources(resourcesOutputDirectory);
+      }
+
+      // Adds class files
+      for (File classesOutputDirectory : classesOutputDirectories.filter(File::exists)) {
```
chanseokoh: Perhaps just `for (File ... : classesOutputDirectories)` because it is already filtered?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -200,23 +184,55 @@ static int getVersionFromString(String versionString) {
...
     eventHandlers = makeEventHandlers(session, log, singleThreadedExecutor);
   }
 
   @Override
-  public JavaLayerConfigurations getJavaLayerConfigurations() {
-    return javaLayerConfigurations;
+  public JibContainerBuilder getContainerBuilderWithLayers(RegistryImage baseImage)
+      throws IOException {
+    try {
+      if (MojoCommon.isWarProject(project)) {
```
chanseokoh: `if (isWarProject())`, I think, like the Gradle counterpart.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  public static LayerConfiguration extraDirectoryLayerConfiguration(
+      Path extraDirectory, Map<AbsoluteUnixPath, FilePermissions> extraDirectoryPermissions)
+      throws IOException {
+    LayerConfiguration.Builder builder =
+        LayerConfiguration.builder().setName(LayerType.EXTRA_FILES.getName());
+    new DirectoryWalker(extraDirectory)
+        .filterRoot()
+        .walk(
+            path -> {
```
briandealwis: lotta paths here.  Maybe `path` -> `localPath`?  Or `extrasPath`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,90 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+      throw new GradleException("Obtaining project build output files failed", ex);
+    }
+  }
+
+  @Override
+  public List<Path> getClassFiles() throws IOException {
+    // TODO: Consolidate with getContainerBuilderWithLayers
+    JavaPluginConvention javaPluginConvention =
+        project.getConvention().getPlugin(JavaPluginConvention.class);
+    SourceSet mainSourceSet = javaPluginConvention.getSourceSets().getByName(MAIN_SOURCE_SET_NAME);
```
briandealwis: I'm pretty ignorant of sourceSets — we don't walk any other sourceSets?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,90 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+    FileCollection classesOutputDirectories =
+        mainSourceSet.getOutput().getClassesDirs().filter(File::exists);
+    ImmutableList.Builder<Path> classFiles = ImmutableList.builder();
```
chanseokoh: nit: new `ArrayList()`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -148,20 +141,90 @@ private static boolean isProgressFooterEnabled(Project project) {
...
+    for (File classesOutputDirectory : classesOutputDirectories) {
+      classFiles.addAll(
+          new DirectoryWalker(classesOutputDirectory.toPath())
+              .filter(path -> path.getFileName().endsWith(".class"))
```
chanseokoh: I think this is not necessary for Gradle, but no harm to have it.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelperTest.java
```diff
@@ -58,10 +65,38 @@ private static void assertExtractionPathsUnordered(
         expectedPaths, entries, layerEntry -> layerEntry.getExtractionPath().toString());
   }
 
+  private static List<LayerConfiguration> getLayerConfigurationByName(
```
briandealwis: Is this `getLayerConfigurationsByName()`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -0,0 +1,100 @@
...
+              AbsoluteUnixPath pathOnContainer =
+                  AbsoluteUnixPath.get("/").resolve(extraDirectory.relativize(path));
+              builder.addEntry(
+                  path, pathOnContainer, extraDirectoryPermissions.get(pathOnContainer));
+            });
```
chanseokoh: Can't this be the code in the previous `Maven/GradleLayerConfigurations`?

```java
    if (Files.exists(extraDirectory)) {	
      layerBuilder.addDirectoryContents(	
          LayerType.EXTRA_FILES,	
          extraDirectory,	
          path -> true,	
          AbsoluteUnixPath.get("/"),	
          extraDirectoryPermissions);	
    }
```

TadCordle: `addDirectoryContents` is a method on `JavaLayerConfigurations`, not `LayerConfiguration.Builder`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
TadCordle: Yeah, I put it in so "`getClassFiles`" makes sense.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
chanseokoh: Ah, right.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
TadCordle: Nope, I think this matches previous behavior.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
TadCordle: Hmm, yeah I think I'll remove the two new methods. This can be done just taking a list of files.

chanseokoh:
LGTM for what I can check. I'll leave @briandealwis to approve this.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -200,23 +184,55 @@ static int getVersionFromString(String versionString) {
...
     eventHandlers = makeEventHandlers(session, log, singleThreadedExecutor);
   }
 
   @Override
-  public JavaLayerConfigurations getJavaLayerConfigurations() {
-    return javaLayerConfigurations;
+  public JibContainerBuilder getContainerBuilderWithLayers(RegistryImage baseImage)
+      throws IOException {
+    try {
+      if (isWarProject()) {
```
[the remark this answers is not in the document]
TadCordle: Yeah, I'll file an issue for this.

Issue #1369, opened 2018-12-19
Decrypt encrypted Maven proxy auth info in settings.xml


chanseokoh:
https://github.com/GoogleContainerTools/jib/pull/1337#discussion_r241465937

Related: #1304.

PR #1573, opened 2019-03-21
No need for checking null Maven settings.xml Decryptor

chanseokoh:
Toward #1369.

Probably we initially considered the possibility that the decryptor could be null, but so far, no one has reported the violation of this:

```java
  SettingsDecrypter getSettingsDecrypter() {
    return Preconditions.checkNotNull(settingsDecrypter);
  }
```

After all, `getSettingsDecrypter()` has not been `@Nullable`. I don't think we need to take out the `checkNotNull()`.

### DIFF
--- a/jib-maven-plugin/pom.xml
+++ b/jib-maven-plugin/pom.xml
@@ -368,7 +368,7 @@
               </annotationProcessorPaths>
               <compilerArgs>
                 <arg>-XDcompilePolicy=simple</arg>
-                <arg>-Xplugin:ErrorProne -Xep:NullAway:ERROR -XepOpt:NullAway:AnnotatedPackages=com.google.cloud.tools</arg>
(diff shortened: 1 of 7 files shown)


briandealwis:
Although I found some locations that do explicitly check (like the [maven-site-plugin](https://github.com/apache/maven-site-plugin/blob/master/src/main/java/org/apache/maven/plugins/site/deploy/AbstractDeployMojo.java#L334)), many others don't do it.

You do need to remove `@Nullable` from the field injection site though.

https://github.com/GoogleContainerTools/jib/blob/450cf6b7f83fb7545ebd9b97190b60850bfe6257/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java#L235

chanseokoh:
From the test failure, I can see why we needed `@Nullable`.

```
[ERROR] /home/travis/build/GoogleContainerTools/jib/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java:[235,41] error: [NullAway] @NonNull field settingsDecrypter not initialized
```

I'll add the following comment: `// @Nullable to suppress NullAway complaining a non-null field uninitialized; not null in practice`

chanseokoh:
I think this is good to go. The plugins always call `getSettingsDecrypter()`, and no one reported the precondition error. I think we can assume it's never null.

PR #1579, opened 2019-03-25
Added note about debugging on JDK 9+

buckett:
When debugging you need to have the JVM listen on all addresses when using JDK 9+.

https://bugs.openjdk.java.net/browse/JDK-8175050

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -382,9 +382,13 @@ Gradle: use `grade --debug -DjibSerialize=true` to enable more detailed logging 
 
 ### How do I enable debugging?
 
-If using the `distroless/java` base image, then use the [`JAVA_TOOL_OPTIONS`](#how-do-i-set-parameters-for-my-image-at-runtime) to pass along debugging configuration arguments.  For example, to have the remote VM accept debug connections on port 5005, but not suspend:
+If using the `distroless/java` base image, then use the [`JAVA_TOOL_OPTIONS`](#how-do-i-set-parameters-for-my-image-at-runtime) to pass along debugging configuration arguments.  For example, to have the remote VM accept debug connections on port 5005, but not suspend for JDK 5-8:
 ```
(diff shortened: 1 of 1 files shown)


buckett:
Will create an issue.

briandealwis:
@buckett can you check if using `0:5005` works?

chanseokoh:
IMO, normally you shouldn't be using `*`. I thought the behavior in JDK 9+ is to allow localhost when not using `*`?

briandealwis:
I suspect @buckett is trying to debug from a different machine or in a container (with a different IP)?

chanseokoh:
Yeah, I suspect so, and we need to understand the exact setup @buckett has. I think most of time, local port forwarding is enough. Oracle says using `*` is [not secure and not recommended](https://www.oracle.com/technetwork/java/javase/9-notes-3745703.html#JDK-8041435). If @buckett is allowing binding from a different remote machine, they should clearly be aware of the insecure nature.

buckett:
Yeah, I was running the created image using Docker for Mac (so on a Linux VM on my Mac) and as a result having the JVM only listen on localhost means the debugger can't connect to it.

In other setups (ways of deploying containers) are you able to connect to a port that is only listening on localhost?

chanseokoh:
> In other setups (ways of deploying containers) are you able to connect to a port that is only listening on localhost?

Yeah, from what I heard from @briandealwis, for example, on Kubernetes, you can set up port forwarding and connecting to `https://localhost:5005` on your local machine will get you connect to the JWDP agent running on a Kubernetes pod that allowing binding only from `localhost` (a local connection originating from the pod).

buckett:
> @buckett can you check if using `0:5005` works?

Yep, that works a treat.

briandealwis:
Probably best to change the example to use `localhost:5005`.

PR #1588, opened 2019-03-28
Remove JavaLayerConfigurations

### LINKED ISSUE #1586: Move JavaLayerConfigurations logic into JavaContainerBuilder and remove original class
`JavaLayerConfigurations` has little use anymore outside of `JavaContainerBuilder`, so we should transfer the logic and remove the class.

TadCordle:
This moves necessary code from `JavaLayerConfigurations` into `JavaContainerBuilder` and removes the original class. Fixes #1586.


jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -454,16 +498,28 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
...
+            layerBuilders,
             LayerType.EXTRA_FILES,
             path,
             appRoot.resolve(othersDestination).resolve(path.getFileName()));
       }
     }
-    jibContainerBuilder.setLayers(layerConfigurationsBuilder.build().getLayerConfigurations());
+
+    // Add layer configurations to container builder
+    ImmutableMap.Builder<LayerType, LayerConfiguration> layerConfigurationsMap =
```
chanseokoh: We no longer need a map, so probably this can be
```java
      List<LayerConfiguration> layers = new ArrayList<>();
      layerBuilders.forEach(
          (type, builder) -> layers.add(builder.setName(type.getName()).build()));
```
I think we live without `ImmutableList`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -401,11 +442,12 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
...
     }
 
-    JavaLayerConfigurations.Builder layerConfigurationsBuilder = JavaLayerConfigurations.builder();
+    Map<LayerType, LayerConfiguration.Builder> layerBuilders = new EnumMap<>(LayerType.class);
 
     // Add classes to layer configuration
     for (PathPredicatePair directory : addedClasses) {
-      layerConfigurationsBuilder.addDirectoryContents(
+      addDirectoryContentsToLayer(
+          layerBuilders,
```
chanseokoh: How about having `layerBuilders` as a `final` field and not pass it around?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -454,16 +498,28 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
     // Add others to layer configuration
     for (Path path : addedOthers) {
       if (Files.isDirectory(path)) {
-        layerConfigurationsBuilder.addDirectoryContents(
-            LayerType.EXTRA_FILES, path, path1 -> true, appRoot.resolve(othersDestination));
+        addDirectoryContentsToLayer(
+            layerBuilders,
+            LayerType.EXTRA_FILES,
+            path,
+            path1 -> true,
```
chanseokoh: I'd go with `ignored -> true`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
TadCordle: Was worried about what happens if people call `toContainerBuilder()` more than once, but I guess I can just clear the field at the beginning of the method.

chanseokoh: Ah, that's a good point. Then I think I like the way it is now.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
chanseokoh: Don't forget this.

TadCordle: Oh weird, I definitely changed this. I guess I hit undo too many times before committing or something.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -454,16 +497,25 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
     // Add others to layer configuration
     for (Path path : addedOthers) {
       if (Files.isDirectory(path)) {
-        layerConfigurationsBuilder.addDirectoryContents(
-            LayerType.EXTRA_FILES, path, path1 -> true, appRoot.resolve(othersDestination));
+        addDirectoryContentsToLayer(
+            layerBuilders,
+            LayerType.EXTRA_FILES,
+            path,
+            ignored -> true,
```
briandealwis: Even better: `Predicates.alwaysTrue()`

PR #1593, opened 2019-04-02
Log warnings for docker config cred helper exceptions

### LINKED ISSUE #1529: jib should warn when docker-credential-gcloud is not found for gcr.io
@etanshaul hit a situation where he was unable to push to a repository on `gcr.io`. Although he had configured docker to authenticate using `gcloud auth configure-docker` (such that his `~/.docker/config.json` had credhelpers registered to `gcloud`), his `PATH` was wrong and `docker-credential-gcloud` was not found by Jib.  Jib silently moved on, and his push failed with an `UNAUTHORIZED`.  Jib should instead warn when a configured credential helper in the `config.js` cannot be found.

As an aside, Jib does warn if `docker-credential-gcr` is not found, regardless of whether it was configured as a credential helper.

TadCordle:
Should fix #1529.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
@@ -121,9 +121,9 @@ public class DockerConfigCredentialRetriever {
             | CredentialHelperNotFoundException ex) {
           // Warns the user that the specified (or inferred) credential helper cannot be used.
           if (eventDispatcher != null && ex.getMessage() != null) {
-            eventDispatcher.dispatch(LogEvent.info(ex.getMessage()));
+            eventDispatcher.dispatch(LogEvent.warn(ex.getMessage()));
             if (ex.getCause() != null && ex.getCause().getMessage() != null) {
-              eventDispatcher.dispatch(LogEvent.info("  Caused by: " + ex.getCause().getMessage()));
+              eventDispatcher.dispatch(LogEvent.warn("  Caused by: " + ex.getCause().getMessage()));
             }
           }
         }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
@@ -119,7 +119,7 @@ public class DockerConfigCredentialRetriever {
         } catch (IOException
             | CredentialHelperUnhandledServerUrlException
             | CredentialHelperNotFoundException ex) {
-          // Warns the user that the specified (or inferred) credential helper cannot be used.
+          // Warns the user that the specified credential helper cannot be used.
           if (eventDispatcher != null && ex.getMessage() != null) {
             eventDispatcher.dispatch(LogEvent.warn(ex.getMessage()));
             if (ex.getCause() != null && ex.getCause().getMessage() != null) {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
@@ -18,6 +18,8 @@ package com.google.cloud.tools.jib.registry.credentials;
 
 import com.google.api.client.util.Base64;
 import com.google.cloud.tools.jib.configuration.credentials.Credential;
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
```diff
@@ -112,7 +119,13 @@ public DockerConfigCredentialRetriever(String registry, Path dockerConfigFile) {
         } catch (IOException
             | CredentialHelperUnhandledServerUrlException
             | CredentialHelperNotFoundException ex) {
-          // Ignores credential helper retrieval exceptions.
+          // Warns the user that the specified (or inferred) credential helper cannot be used.
+          if (eventDispatcher != null && ex.getMessage() != null) {
+            eventDispatcher.dispatch(LogEvent.info(ex.getMessage()));
```
chanseokoh: I think this should be `.warn()` as mentioned in #1529. This means misconfiguration in `.docker/config.json`: a credential helper binary is configured, but the binary couldn't be found or the helper did not return credentials unexpectedly.

You said

> It looks like we do log about missing inferred/configured credential helpers at the info level already

Where is this being done? Is this related to what @briandealwis said about

> As an aside, Jib does warn if docker-credential-gcr is not found, regardless of whether it was configured as a credential helper.

? I vaguely remember Jib warns if a usual cred helper is missing regardless of whether it is configured in `config.json`.

TadCordle: `CredentialRetrieverFactory#inferCredentialHelper()` logs `CredentialHelperNotFoundException`/`CredentialHelperUnhandledServerUrlException` at the info level. Explicitly configured credential helpers (`CredentialRetrieverFactory#dockerCredentialHelper()`) seems to rethrow a related `IOException` as a `CredentialHelperNotFoundException()`. I'm not finding anything related that's logged at the `warn` level, maybe @briandealwis knows.

briandealwis: I remember seeing complaints about not finding `docker-credential-gcr` though I can't reproduce.  I wonder if I was using an example using an older version of Jib.

briandealwis: I agree with @chanseokoh: I tried pushing to gcr.io but without `docker-credential-gcloud` in my `PATH`, where `gcloud` is configured within my `.docker/config.json` for `gcr.io`.  With this patch as is, I saw no helpful messages.  Changing the `info` to `warn`, I get the more helpful:

```
[INFO] --- jib-maven-plugin:1.0.3-SNAPSHOT:build (default) @ helloworld ---
[INFO] 
[INFO] Containerizing application to gcr.io/gcp-testing-29114/foo...
[INFO]    The system does not have docker-credential-gcloud CLI
```

TadCordle: Alright, I originally had it at info to be consistent with the other cred helper warnings, but I'll change it to warn for this case.

briandealwis: Odd that `eventDispatcher.dispatch(LogEvent.warn(xxx))` is  logged as `[INFO]`.  Hmm so is `.error`.

TadCordle: That shouldn't be the case. Gradle/MavenProjectProperties should setup the loggers to log at the correct level.

chanseokoh: > Alright, I originally had it at info to be consistent with the other cred helper warnings, but I'll change it to warn for this case.

In the other case of `CredentialRetrieverFactory#inferCredentialHelper()`, we blindly attempt running `docker-credential-gcr` if the registry is `gcr.io`, regardless of whether the helper is configured in `.docker/config.json`. It's for the convenience sake, e.g., for people who forgot to configure Docker but has the binary on `PATH` for popular registries like `gcr.io`. So, it makes sense that we don't log at the warn (although I actually think it should do warn on `CredentialHelperUnhandledServerUrlException` to be more helpful while it makes sense to do info on `CredentialHelperNotFoundException`.)

briandealwis:
I must have had a stale artifact as it works fine after a rebuild. 🤷‍♂️

Issue #1020, opened 2018-09-21
Make layering of extraDirectory configurable 


remmeier:
Files in the extra layer can be:
- quickly changing project resources 
- rather static project resources 
- third-party resources
- potentially have a substantial size

So it is thinkable to place extraDirectories inbetween any other layer for optimial layering. A configuration flag for this would be great. Or maybe even the possiblity to have multiple extra layers.

coollog:
Hi @remmeier , thanks for the suggestion. We will definitely consider supporting multiple extra layers though the order will most likely not be configurable since each layer is pushed independently (and thus ordering does not matter). I think for multiple extra directories, we can support something like `src/main/jib2`, `src/main/jib3`, etc. as conventions for extra extra directories.

coollog:
@GoogleContainerTools/java-tools Any comments on supporting extra extra directories (`src/main/jib2`)?

TadCordle:
@coollog Would the same convention be used for custom extra directories? 

e.g. `extraDirectory = file('custom/extra/dir')`
Layered structure is `custom/extra/dir`, `custom/extra/dir2`, `custom/extra/dir3`, etc.

coollog:
Yep, that could be added too, so any number 2 and onwards (consecutively) appended to the configured extra directory path will be used as extra directory layers in the numbered order.

chanseokoh:
I just thought about a (highly unlikely) edge case of `extraDirectory = file('F:\')`. Might need to be careful not to fail in that case.

And another edge case: what if `src/main/jib` doesn't exist but `src/main/jib2` does? Or, `src/main/jib` and `src/main/jib4` exist but no `jib2` or `jib3`.

Might just be cleaner to get a list of extra directory paths in the config. However, these are edge cases that may work in one way or another anyway, so I also see some good things about the simple approach of `jib`, `jib2`, etc.

coollog:
For `F:`, Jib would try to look for `F:2`, but that doesn't exist.
For `src/main/jib2` exists but `src/main/jib` doesn't, Jib looks for `src/main/jib`, doesn't find it, and then stops. Essentially, it should go in sequence.

briandealwis:
This seems to me to be a case where we abandon convention for configuration.  @remmeier's examples don't cleanly map to `2`, `3`, `4`, ... and I would guess that some of those files would be things generated or best copies into `target/` rather than into the source tree.

[...]

PR #1602, opened 2019-04-04
Enable configuring multiple extra directories

chanseokoh:
#1020. (This actually fixes it, but I'll let #1020 open, since we need to update docs and CHANGELOG.)

This PR allows specifying multiple extra directories, retaining backward compatibility. Note that this does NOT have an actual implementation; only the first directory is in effect, which is the current status. This just allows the syntax for specifying multiple directories.

The only exception is the `_skaffold-files(-v2)` goals; I made it print all the given extra directories, which shouldn't be a problem.

All of these are accepted:
```
-Djib.extraDirectory.path=/some/path
-Djib.extraDirectory.path=/some/path,/this/is/ignored
```

```xml
<extraDirectory>/some/path</extraDirectory>

<extraDirectory>
  <path>/some/path</path>
</extraDirectory>

<extraDirectory>
  <paths>
    <path>/some/path</path>
    <path>/this/is/ignored</path>
  </paths>
  <permissions> may or may not have this <permissions>
</extraDirectory>
```

```gradle
jib.extraDirectory = 'some/path'  // wan't possible before
jib.extraDirectory = file('some/path')
jib.extraDirectory = ['some/path', 'this/is/ignored']
jib.extraDirectory = files('some/path', 'this/is/ignored')
jib.extraDirectory.path = 'some/path'  // wasn't possible before
jib.extraDirectory.path = file('some/path')
jib.extraDirectory.path = ['some/path', 'this/is/ignored']
jib.extraDirectory.path = files('some/path', 'this/is/ignored')
```


chanseokoh:
Ah, wait. I just found something wrong.

briandealwis:
> -Djib.extraDirectory.path=/some/path,/this/is/ignored

What do you think about instead using the platform path separator (`:` on Unix systems, `;` on Windows)?  Commas, although uncommon, are allowed in directory names.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -107,6 +106,10 @@ public void setExtraDirectory(File extraDirectory) {
     this.extraDirectory.setPath(extraDirectory);
   }
 
+  public void setExtraDirectory(Object extraDirectories) {
```
briandealwis: I think this needs some javadoc — why is it necessary?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
```diff
@@ -22,48 +22,52 @@
...
+      List<String> pathStrings = ConfigurationPropertyValidator.parseListProperty(property);
+      return pathStrings.stream().map(Paths::get).collect(Collectors.toList());
     }
-    return path;
+    return paths;
   }
 
-  public void setPath(File path) {
-    this.path = path.toPath();
+  public void setPath(Object paths) {
```
briandealwis: I think this needs some javadoc — why is it necessary?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
```diff
@@ -149,9 +150,8 @@ public void listFiles() {
     printProjectFiles(project);
 
     // Print extra layer
-    if (Files.exists(jibExtension.getExtraDirectory().getPath())) {
-      System.out.println(jibExtension.getExtraDirectory().getPath());
-    }
+    List<Path> extraDirectories = jibExtension.getExtraDirectory().getPaths();
```
briandealwis: Do you think it's worth renaming `getExtraDirectory()` to `getExtras()`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleRawConfiguration.java
```diff
@@ -146,8 +146,8 @@ public ImageFormat getImageFormat() {
   }
 
   @Override
-  public Path getExtraDirectory() {
-    return jibExtension.getExtraDirectory().getPath();
+  public List<Path> getExtraDirectory() {
```
briandealwis: Pluralize

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -197,11 +197,11 @@ public void set(String image) {
      * @param path the value to set {@code path} to
      */
     public void set(File path) {
-      this.path = path;
+      this.path = Collections.singletonList(path);
     }
 
     @Nullable
-    public File getPath() {
+    public List<File> getPath() {
```
briandealwis: Pluralize?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
```diff
@@ -74,7 +74,7 @@
 
   Optional<String> getProperty(String propertyName);
 
-  Path getExtraDirectory();
+  List<Path> getExtraDirectory();
```
briandealwis: Pluralize?

TadCordle:
>What do you think about instead using the platform path separator (: on Unix systems, ; on Windows)? Commas, although uncommon, are allowed in directory names.

I will point out that all our other list system properties are separated by commas, but commas themselves can be escaped with `\`. So maybe we should just stay consistent.

briandealwis:
Volumes provide directories too.  Ok, consistency is good.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
chanseokoh: I saw your [comment](https://github.com/GoogleContainerTools/jib/issues/1020#issuecomment-437902030) before, and I thought `jib.extras` sounded too general and also `getExtras()` in our code could be confusing. And `getExtraDirectory()` inside `JibExtension` seems consistent with that it returns the `ExtraDirectoryParameters` object, which matches the config parameter name.

chanseokoh:
I realized the initial PR was very close to making things work end-to-end (placing files from multiple directories in different layers). I'm still fixing things that I did wrong.

chanseokoh:
Maven works end-to-end. Each directory goes into its own layer in order. All below are valid:

```xml
<extraDirectory>/some/path</extraDirectory>

<extraDirectory>
  <path>/some/path</path>
</extraDirectory>

<extraDirectory>
  <!-- if <path> or <paths> is missing, the default "src/main/jib" is in effect -->
  <permissions> ... </permissions>
</extraDirectory>

<extraDirectory>
  <paths>
    <path>/some/path</path>
    <path>/this/is/ignored</path>
  </paths>
  <permissions> may or may not have this </permissions>
</extraDirectory>
```

The property `jib.extraDirectory.path` in Maven works too.

Also fixed a separate issue in `FilesMojoV2` where it throws NPE if `<extraDirectory>` is defined but `<path>` (or `<paths>`) is missing.

However, Gradle isn't working. Will continue to work on that.

chanseokoh:
Gradle working. The property works on Gradle too.

```gradle
jib.extraDirectory = 'some/path'  // wan't possible before
jib.extraDirectory = file('some/path')
jib.extraDirectory = ['some/path', 'this/is/ignored']
jib.extraDirectory = files('some/path', 'this/is/ignored')
jib.extraDirectory.path = 'some/path'  // wasn't possible before
jib.extraDirectory.path = file('some/path')
jib.extraDirectory.path = ['some/path', 'this/is/ignored']
jib.extraDirectory.path = files('some/path', 'this/is/ignored')
```

I'll add more integration tests.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/FilesTask.java
chanseokoh: We may rename this later, but to reduce complexity and potential mistakes, I suggest stop changing names in this PR as it is in working order now.

chanseokoh:
This should be working now, essentially fixing #1020. For doc changes and CHANGLOG, I'll create another PR after having LGTM.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
```diff
@@ -22,48 +22,59 @@
...
   }
 
   @Internal
-  public Path getPath() {
+  public List<Path> getPaths() {
     // Gradle warns about @Input annotations on File objects, so we have to expose a getter for a
     // String to make them go away.
-    if (System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH) != null) {
-      return Paths.get(System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH));
+    String property = System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH);
```
TadCordle: Should we rename this system property to `jib.extraDirectory.paths`? Or maybe introduce a new one to maintain backwards compatibility?

loosebazooka:
I think it might make sense to support `extraDirectory` and a plural version `extraDirectories/extraLayers` (or something). We can say `extraDirectory` will be deprecated in 2.0. This will also allow you to design the extra directories substructure however you want.

chanseokoh:
@loosebazooka I gave a thought about that too, and here's what I thought as downsides:

- Increased pollution in the config space.
- Our docs will look more complicated. Users might be confused. Especially our code can look confusing by having `getExtraDiretory` and `getExtraDirectories` in many places for example.
- Confusion arises when both `extraDirectory` and `extraDirectories` are given. Need more code to examine values of both configs and print errors.
- If we are ever going to deprecate `extraDirectory`, we could just rename `extraDirectory` to `extraDirectories` at that time, instead of carrying both config forms until then.

> This will also allow you to design the extra directories substructure however you want.

I think it isn't the case that the extra directory structures for multiple layers in this PR are not ideal because of the backward compatibility. If we rename `extraDirectory` to `extraDirectories`, I think things will look good enough. For Maven, `<paths>` as a list don't interfere with the current `<path>`. We could simply drop the old `<path>` when we deprecate `extraDirectory`

chanseokoh:
@loosebazooka forgot to ask, what do you think?

loosebazooka:
Okay, so I think I understand and I can be sold with the structural changes to the config being minimal and perhaps I'm being overly zealous with my config deprecation philosophy, I'm just kind of concerned that if we don't try to move to the "right" naming, we will end up with names that don't match their function exactly.

So what about this suggestion: The PR remains as-is, except naming, we're doing a naming change of `extraDirectory`(keep till 2.0, but deprecated) -> `extraDirectories`(documented now in docs)
- I think all the changes we make in `jib-core` can still be breaking (given we are pre-1.0), so we should just be able to call `getExtraDirectories` globally in the codebase.
- The configuration complication is limited at the plugin level. All code still calls `getExtraDirectories()`. `getExtraDirectories()` correctly handles delegating to the users config (possible providing a deprecation message if they happened to use `extraDirectory` instead of `extraDirectories`. I believe this to be an ideal deprecation pattern/policy. The problem with renaming at an instant, is that it will immediately break everyone with no warning (and I feel like I made this mistake with the appengine plugins). 
- From the next release on our docs will only reference `extraDirectories`?

chanseokoh:
Talked with @loosebazooka. Supporting both the old structure and the new structure sounds like a good deprecation strategy. Basically, This is already achieved in this PR; it's merely that the new structure has the same config name as `extraDirectory` instead of `extraDirectories`. We decided we'll rename the new structure in a later PR, which will also contain a deprecation warning message when the old structure is used. So, this PR is still ready for review as-is.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
chanseokoh: We will rename this in a later PR: https://github.com/GoogleContainerTools/jib/pull/1602#issuecomment-482704832

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -197,12 +200,11 @@ public void set(String image) {
...
     public void set(File path) {
-      this.path = path;
+      this.paths = Collections.singletonList(path);
     }
 
-    @Nullable
-    public File getPath() {
-      return path;
+    public List<File> getPaths() {
+      return path != null ? Collections.singletonList(path) : paths;
```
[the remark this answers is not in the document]
chanseokoh: Good point. I expect this will be handled properly in a later PR to have a deprecation warning. Having both `<path>` and `<paths>` is not a correct usage.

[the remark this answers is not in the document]
loosebazooka: Also, maybe we add a deprecation log here for `<path>` (unless we cover this in the followup).

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -507,15 +509,14 @@ String getFormat() {
    *
    * @return the configured extra directory path
```
loosebazooka: a list of configured extra directory paths

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -507,15 +509,14 @@ String getFormat() {
...
    */
-  Optional<Path> getExtraDirectoryPath() {
+  List<Path> getExtraDirectoryPaths() {
```
loosebazooka: can this just be getExtraDirectories?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MojoCommon.java
```diff
@@ -43,22 +44,21 @@ static boolean isWarProject(MavenProject project) {
...
-    return jibPluginConfiguration
-        .getExtraDirectoryPath()
-        .orElse(
-            Preconditions.checkNotNull(jibPluginConfiguration.getProject())
-                .getBasedir()
-                .toPath()
-                .resolve("src")
-                .resolve("main")
-                .resolve("jib"));
+  static List<Path> getExtraDirectoryPaths(JibPluginConfiguration jibPluginConfiguration) {
```
loosebazooka: getExtraDirectories? Is the return type should be sufficient for us to understand it is `java.nio.Path` ?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
```diff
@@ -105,7 +108,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     // and is expected to run on all projects irrespective of their configuring of the jib plugin).
     if (project.getPlugin(MavenProjectProperties.PLUGIN_KEY) != null) {
       // print out extra directory
-      System.out.println(resolveExtraDirectory());
+      resolveExtraDirectoris().stream().forEach(System.out::println);
```
loosebazooka: spelling? 
`Directoris` vs
`Directories`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
```diff
@@ -180,24 +182,31 @@ private Path resolveExtraDirectory(MavenProject project) {
       if (pluginConfiguration != null) {
         Xpp3Dom extraDirectoryConfiguration = pluginConfiguration.getChild("extraDirectory");
         if (extraDirectoryConfiguration != null) {
-          Xpp3Dom child = extraDirectoryConfiguration.getChild("path");
-          if (child != null) {
+          Xpp3Dom pathChild = extraDirectoryConfiguration.getChild("path");
```
loosebazooka: this seems really intense, doesn't maven auto populate complex objects? Or is this because we support `<extraDirectories>/path</extraDirectories>`?

Should we phase/deprecate out the option to configure that?

loosebazooka:
One more thought, sorry `:)`

Should just leave `<path>` as is, and mark it deprecated. Our final state is the only way to specify extra directories would be through the complex object block `<extraDirectories><paths>...`?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
chanseokoh: I believe you are talking about the way `FilesMojo` does with the `extraDirectory` field in it. I remember @TadCordle saying the reason for manual DOM traversal is because there was an issue in the multi-module case.

chanseokoh:
> Should just leave `<path>` as is, and mark it deprecated. 

Good idea. Adding `@Deprecated` can certainly be done in this PR.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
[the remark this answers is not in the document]
chanseokoh: Yeah, will cover the deprecation warning in the followup.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
loosebazooka: Oh yeah I didn't realize this was FilesMojo.

PR #1613, opened 2019-04-09
Add methods for setting file timestamps

### LINKED ISSUE #1079: File timestamp 0 breaks Spring WebMVCs static resource resolving

### LINKED ISSUE #1607: Add methods to jib-core to allow setting timestamps of files on container
We should probably have a method that just takes a timestamp and applies it to all the files, e.g.:

`JibContainerBuilder#setFileCreationTime(Instant creationTime)`

as well as method that takes a function that can be used to calculate the timestamp on a per-file basis:

`JibContainerBuilder#setFileCreationTime(Function<Path, Instant> creationTime)`

From #1079

TadCordle:
Adds overloads for `LayerConfiguration.Builder#addEntryRecursive` that take providers for file permissions/file modification timestamps to apply permissions and timestamps on a per-file basis.

Fixes #1607.


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -58,6 +61,8 @@
     @Nullable private Path baseImageLayersCacheDirectory;
     private boolean allowInsecureRegistries = false;
     private ImmutableList<LayerConfiguration> layerConfigurations = ImmutableList.of();
+    private Function<AbsoluteUnixPath, Instant> fileTimestampProvider =
+        ignored -> Instant.ofEpochMilli(1000);
```
loosebazooka: Wonder if this should be an explicit provider implementation? I can't remember who mentioned this, but it was that the default was somewhere in code rather than set in the api? But maybe it doesn't matter so much if it's configurable (and documented).

jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
```diff
@@ -106,7 +106,8 @@ public void testBuild() throws URISyntaxException, IOException {
                 .addEntry(blobA, AbsoluteUnixPath.get("/extract/here/apple/blobA"))
                 .addEntry(blobA, AbsoluteUnixPath.get("/extract/here/banana/blobA"))
                 .build()
-                .getLayerEntries());
+                .getLayerEntries(),
```
loosebazooka: Would be nice to have a test where we set it to something that is not the default?

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/TimestampProvider.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.google.cloud.tools.jib.frontend;
+
+import com.google.cloud.tools.jib.filesystem.AbsoluteUnixPath;
+import java.time.Instant;
+
+/** Provides a timestamp given a file's extraction path. */
+@FunctionalInterface
+public interface TimestampProvider {
```
chanseokoh: This accepts `AbsoluateUnixPath`, so how about `FileTimestampProvider`?

chanseokoh:
> I think we want to use a provider rather than single value, so `Function<Path,Instant>`

The PR uses a provider (effectively `Function<AbsoluteUnixPath, Instant`), if you haven't noticed.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -70,29 +70,6 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
...
-     * @param sourceFile the source file to add to the layer
-     * @param pathInContainer the path in the container file system corresponding to the {@code
-     *     sourceFile}
-     * @param permissions the file permissions on the container. If null, then default permissions
-     *     are used (644 for files, 755 for directories)
-     * @param lastModified the modification time of the file
-     * @return this
-     * @see Builder#addEntry(Path, AbsoluteUnixPath)
-     */
-    public Builder addEntry(
```
briandealwis: I don't think we should get rid of this.

briandealwis:
Weird — I must have somehow looking from an earlier commit.  This current state looks way better.  But I think we should use this provider to provide defaults which can still be explicitly over-ridden in addEntry(...).

TadCordle:
If we do allow setting a timestamp provider on a per-`LayerEntry` basis, I wonder if we even need the `AbsoluteUnixPath` argument in the provider anymore?

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
TadCordle: Yeah, I think I'll change things from the current `BuildConfiguration` approach to allow configuring different providers on `LayerConfiguration`. That said, I assume this should probably take a `FileTimestampProvider` instead of a `long`.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -103,12 +99,17 @@ public Builder addEntry(
...
      * @return this
      * @see Builder#addEntry(Path, AbsoluteUnixPath)
      */
     public Builder addEntry(
-        Path sourceFile, AbsoluteUnixPath pathInContainer, @Nullable FilePermissions permissions) {
-      return addEntry(sourceFile, pathInContainer, permissions, 1000);
+        Path sourceFile,
+        AbsoluteUnixPath pathInContainer,
+        @Nullable FilePermissions permissions,
+        FileTimestampProvider lastModified) {
```
chanseokoh: As @TadCordle pointed out, considering that `addEntry()` is to add a single file to the layer, it can be confusing from the user perspective that `lastModified` is a function from `Path` to `Instant`. We should clearly document the function is to take the source path. However, it still seems unintuitive given that you are adding a single file and you can always determine the timestamp at the time of using this API. In the end, the function will always get the `sourceFile` parameter being passed here; the function is not for determining timestamps for many different files. Such function will make sense at a higher level when the user is passing multiple files. @briandealwis WDYT?

TadCordle: I kinda agree, having a provider overcomplicates things a bit in this case since the timestamp can just be determined when the files are added to the `LayerConfiguration`. I could see it being more useful for something like `addEntryRecursive` where you can't set the timestamp of each individual file; maybe an overload or two for `addEntryRecursive` that take a timestamp/file permission provider would be more useful, and the non-recursive `addEntry` should just take an `Instant`.

briandealwis: I agree: a provider seems a burden.  But I do like @TadCordle's use of `Instant` over a long, though maybe we should take both to avoid unnecessary API breakage.  Maybe deprecate the variant with a long?

TadCordle: It would only change `LayerEntry`, which isn't intended to be an API class. So I don't think we need to deprecate anything for now.

briandealwis: We have an `Builder.addEntry()` that takes a long that [you're proposing to remove a bit above](https://github.com/GoogleContainerTools/jib/pull/1613/files#diff-fcb2b167686b2e075cb5149f1862e28eL87)

briandealwis: I wonder if we wouldn't be better extracting `addEntryRecursive` into a separate walker that calls `addEntry`.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/FileTimestampProvider.java
```diff
@@ -0,0 +1,36 @@
...
+  /** A provider that returns Epoch + 1 second for all files. */
+  FileTimestampProvider DEFAULT = ignored -> Instant.ofEpochSecond(1);
+
+  /**
+   * Returns the modification timestamp to apply to a file on the container.
+   *
+   * @param sourcePath the source path of the file to be added to the container.
+   * @return the timestamp
+   */
+  Instant generateTimestamp(Path sourcePath);
```
briandealwis: Is _generate_ the right word here?  _get_?  Maybe even just _timestamp(Path)_?

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -40,7 +41,7 @@
   private final Path sourceFile;
   private final AbsoluteUnixPath extractionPath;
   private final FilePermissions permissions;
-  private final long lastModifiedTime;
+  private final FileTimestampProvider modifiedTimeProvider;
```
briandealwis: This should remain either `long` or (better) `Instant` — no need to have a provider here.

TadCordle: If we only have file timestamp specification on individual files, I don't even think we need this interface anymore. If we use it for `addEntryRecursive`, I think just using `Function` would be more self-documenting.

TadCordle:
Sorry to keep changing this so much. Trying this now:

* `addEntry` may take a single `FilePermission` and `Instant`
* `addEntryRecursive` may take a `Function` that determines the permissions/timestamp on a per-file basis

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -128,13 +137,67 @@ public Builder addEntry(
...
+    /**
+     * Adds an entry to the layer. If the source file is a directory, the directory and its contents
+     * will be added recursively.
+     *
+     * @param sourceFile the source file to add to the layer recursively
+     * @param pathInContainer the path in the container file system corresponding to the {@code
+     *     sourceFile}
+     * @param filePermissionProvider a provider that takes a source file and returns the file
+     *     permissions that should be set when it is added to the container
+     * @param fileTimestampProvider a provider that takes a source file and returns the file
```
chanseokoh: I think these functions should act on pathInContainer, not sourceFile. For example, you want to put a lot files in a source directory recursively and you want to set an executable bit only on a few script files. Our plugin configurations take the same approach. With the per-sourceFile way, you end up having to set an executable bit on every file under the directory.

TadCordle: Decided to take both source and destination, since both may be useful (e.g. our default permissions provider needs to check if the source is a file or folder to give 644 and 755 respectively).

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -80,17 +82,12 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
...
      * @see Builder#addEntry(Path, AbsoluteUnixPath)
      */
     public Builder addEntry(
-        Path sourceFile,
-        AbsoluteUnixPath pathInContainer,
-        @Nullable FilePermissions permissions,
-        long lastModified) {
-      layerEntries.add(new LayerEntry(sourceFile, pathInContainer, permissions, lastModified));
-      return this;
+        Path sourceFile, AbsoluteUnixPath pathInContainer, @Nullable FilePermissions permissions) {
```
chanseokoh: This leaves us

`.addEntry(Path, AbsoluteUnixPath, @Nullable FilePermissions)`
`.addEntry(Path, AbsoluteUnixPath, @Nullable FilePermissions, Instant)`

where `@Nullable` means using the default. OTOH, you removed `@Nullable` from `LayerEntry`, leaving us

`LayerEntry(Path, AbsoluteUnixPath)`
`LayerEntry(Path, AbsoluteUnixPath, FilePermissions)`
`LayerEntry(Path, AbsoluteUnixPath, FilePermissions, Instant)`.

That is, the pattern is not consistent, so how about removing `@Nullable` and add `addEntry(Path, AbsoluteUnixPath)`? One downside of removing `@Nullable` is that the user will continue to pass a null after upgrading jib-core and get into trouble, but since we are not GA, I think this is fine.

That said, we should be conscious to have a consistent pattern involving `@Nullable` for default values.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -53,8 +53,12 @@ public static LayerConfiguration extraDirectoryLayerConfiguration(
             localPath -> {
               AbsoluteUnixPath pathOnContainer =
                   AbsoluteUnixPath.get("/").resolve(extraDirectory.relativize(localPath));
-              builder.addEntry(
-                  localPath, pathOnContainer, extraDirectoryPermissions.get(pathOnContainer));
+              FilePermissions permissions = extraDirectoryPermissions.get(pathOnContainer);
```
chanseokoh: See if it is trivial to make this return `Optional` in this PR.

jib-core/src/main/java/com/google/cloud/tools/jib/image/LayerEntry.java
```diff
@@ -88,18 +88,43 @@ public LayerEntry(
...
 
   /**
    * Returns the modification time of the file in the entry.
    *
    * @return the modification time
    */
-  public long getLastModifiedTime() {
-    return lastModifiedTime;
+  public Instant getLastModifiedTime() {
+    return modifiedTime;
```
chanseokoh: nit: `modifiedTime`, `lastModified`, and `lastModifiedTime` are used throughout, so let's try to be consistent. I think "modification time" in Javadocs is OK though.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -80,17 +84,14 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
...
      */
     public Builder addEntry(
-        Path sourceFile,
-        AbsoluteUnixPath pathInContainer,
-        @Nullable FilePermissions permissions,
-        long lastModified) {
-      layerEntries.add(new LayerEntry(sourceFile, pathInContainer, permissions, lastModified));
-      return this;
+        Path sourceFile, AbsoluteUnixPath pathInContainer, FilePermissions permissions) {
+      return addEntry(sourceFile, pathInContainer, permissions, Instant.ofEpochSecond(1));
```
chanseokoh: I think `LayerEntry.DEFAULT_FILE_TIMESTAMP_PROVIDER.apply(sourceFile, pathInContainer)` is better than `Instant.ofEpochSecond(1)` in the long run.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
TadCordle: Do you mean this line should be `Optional<FilePermissions> permissions = ...`, or do you mean something else?

chanseokoh: Yeah, sorry, I meant `Optional<FilePermissions> permissions = ...`.

chanseokoh: Never mind. I didn't realize this is just a map.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -78,19 +82,20 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
      * @param sourceFile the source file to add to the layer
      * @param pathInContainer the path in the container file system corresponding to the {@code
      *     sourceFile}
-     * @param permissions the file permissions on the container. If null, then default permissions
-     *     are used (644 for files, 755 for directories)
-     * @param lastModified the modification time of the file
+     * @param permissions the file permissions on the container (Defaults to 644 for files, 755 for
```
chanseokoh: nit: sorry, I should have said the entire part of "Defaults to 644 ..." doesn't make much sense, because callers are always giving explicit `permissions`. There's no way to use defaults since null is not acceptable. Perhaps move this Javadoc about default permissions to the method above that doesn't take `permissions`.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/LayerConfiguration.java
```diff
@@ -101,14 +106,21 @@ public Builder addEntry(
      * @param sourceFile the source file to add to the layer
      * @param pathInContainer the path in the container file system corresponding to the {@code
      *     sourceFile}
-     * @param permissions the file permissions on the container. If null, then default permissions
-     *     are used (644 for files, 755 for directories)
+     * @param permissions the file permissions on the container (Defaults to 644 for files, 755 for
```
chanseokoh: ditto

PR #1622, opened 2019-04-12
Hide internal classes and methods

chanseokoh:
Up for discussion.

I've looked into the classes in `.api`, and the situation is not so grim if we get rid of the two internal methods `toImageConfiguration()` and `toBuildSteps()`. It would then hide `BuildConfiguration`, `ImageConfiguration`, and `BuildSteps`, the biggest offenders. I tried to explore what can be done to remove them at this stage with little refactoring.

The idea here is to have `TarImage`, `DockerDaemonImage`, etc as interfaces, not mandating them to implement the two internal methods. Instead, internal classes implementing the interfaces will carry them. (Note `SourceImage` and `TargetImage` are already invisible.)

If you have other ideas, throw them here.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
@@ -53,7 +53,8 @@ public class Containerizer {
    * @return a new {@link Containerizer}
    */
   public static Containerizer to(RegistryImage registryImage) {
-    return new Containerizer(registryImage);
+    Preconditions.checkArgument(registryImage instanceof TargetImage);
+    return new Containerizer((TargetImage) registryImage);
   }
 
   /**
@@ -63,7 +64,8 @@ public class Containerizer {
    * @return a new {@link Containerizer}
    */
   public static Containerizer to(DockerDaemonImage dockerDaemonImage) {
-    return new Containerizer(dockerDaemonImage);
+    Preconditions.checkArgument(dockerDaemonImage instanceof TargetImage);
+    return new Containerizer((TargetImage) dockerDaemonImage);
   }
 
   /**
@@ -73,7 +75,8 @@ public class Containerizer {
    * @return a new {@link Containerizer}
    */
   public static Containerizer to(TarImage tarImage) {
-    return new Containerizer(tarImage);
+    Preconditions.checkArgument(tarImage instanceof TargetImage);
+    return new Containerizer((TargetImage) tarImage);
   }
 
   private final TargetImage targetImage;
@@ -133,6 +136,7 @@ public class Containerizer {
     baseImageLayersCacheDirectory = cacheDirectory;
     return this;
   }
+
   /**
    * Sets the directory to use for caching application layers. This cache can be shared between
    * multiple images. If not set, a temporary directory will be used as the application layers
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
@@ -1,4 +1,21 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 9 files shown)


chanseokoh:
What do you think of this guys?

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -53,7 +53,8 @@
    * @return a new {@link Containerizer}
    */
   public static Containerizer to(RegistryImage registryImage) {
-    return new Containerizer(registryImage);
+    Preconditions.checkArgument(registryImage instanceof TargetImage);
```
loosebazooka: This seems like an unnecessary test? Doesn't it extend TargetImage?

loosebazooka:
I think this does a good job of isolating the issues while keeping the user's interface to jib-core exactly the same. I'm not fully onboard with having to use code like
```
Preconditions.checkArgument(registryImage instanceof TargetImage);
```

I wonder if one way to handle that is to just add getters to all our ImageTypes (RegistryImage, etc) so they end up being simple value types and then having hidden (either package private, or non api) classes be responsible for transforming something from ImageType -> `ImageReference` or `BuildSteps`.

For example:
```
// in api package
public class TarImage {
  // or package private?
  public getImageReference()
  public getOutputFile()
}

// in internal package, or package private in api (maybe not static)
public static class TarImages { // or some name
  public static toImageConfiguration(TarImage);
  public static toBuildSteps(TarImage, BuildConfiguration);
}
```

The weird part about this is Containerizer now does some weird stuff:
```
public static Containerizer to(TarImage tarImage) {
  return new Containerizer(tarImage);
}
```
now becomes
```
public static Containerizer to(TarImage tarImage) {
  return new Containerizer(tarImage, TarImages::toImageConfiguration, TarImages::toBuildConfiguration);
}

// where the Containerizer constructor is now templated...
```
since we need to tell it how to convert to what we want inside containerizer. I don't know if this is exactly best practice...


So an alternative to this is to have TarImages be some sort of instantiable class that implements some kind of TargetImageConverter interface

```
public static class TarImageConverter implement TargetImageConverter { // or some name
  public toImageConfiguration(); 
  public toBuildSteps(BuildConfiguration); 
}
```
with this static `to` method:
```
public static Containerizer to(TarImage tarImage) {
  return new Containerizer(tarImage, TarImageConverter.for(tarImage)); // or whatever
}
```

So I haven't really looked into how making these changes will affect our code, and I've only started looking at the problem, so I may have missed things and provided a completely silly solution.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
```diff
@@ -17,20 +17,12 @@
...
 import com.google.cloud.tools.jib.image.InvalidImageReferenceException;
-import com.google.common.collect.ImmutableMap;
 import java.nio.file.Path;
 import java.util.Map;
-import javax.annotation.Nullable;
 
-/** Builds to the Docker daemon. */
-// TODO: Add tests once JibContainerBuilder#containerize() is added.
-public class DockerDaemonImage implements TargetImage {
+public interface DockerDaemonImage {
```
briandealwis: So I kinda like this change except that it would be more in keeping if `DockerDaemonImage` used the builder pattern as used with `TarImage`.  Then we could avoid the oddity of the object supporting `set*` methods.
```java
DockerDaemonImage.builder("image/ref")
  .setDockerEnvironment(map...)
  .build();
```

jib-core/src/main/java/com/google/cloud/tools/jib/api/RegistrySourceTargetImage.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.cloud.tools.jib.configuration.BuildConfiguration;
+import com.google.cloud.tools.jib.configuration.ImageConfiguration;
+import com.google.cloud.tools.jib.configuration.credentials.Credential;
+import com.google.cloud.tools.jib.configuration.credentials.CredentialRetriever;
+import com.google.cloud.tools.jib.image.ImageReference;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+class RegistrySourceTargetImage implements RegistryImage, SourceImage, TargetImage {
```
briandealwis: This `SourceTarget` in the name is a bit odd, but I guess we can name it `RegistryImage` when we move it to a different package.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
chanseokoh: Unrelated to this: looking at `TarImage` again, I think the `Builder` is not a right choice as a name used for the fluent interface pattern.

chanseokoh:
Closing after the follow-up PR #1659.

PR #1626, opened 2019-04-12
Deprecate <extraDirectory> config in Maven

### LINKED ISSUE #1625: FilesMojo.resolveExtraDirectory() doesn't read EXTRA_DIRECTORY_PATH property
It should read the property first. Basically do the same thing as in `JibPluginConfiguration.getExtraDirectory()`, but note that it should resort to the default "src/main/jib" additionally.

chanseokoh:
Note `<extraDirectories>` is a genuine list like `<args>`, so below will no longer be allowed.
```xml
<extraDirectories>path</extraDirectories>
```

Also fixes #1625.


chanseokoh:
Ready for review. Being a Maven list, the new syntax requires always having `<paths>` and `<path>`.

```xml
<extraDirectories>
  <paths>
    <path>/awesome/directory</path>
  </paths>
</extraDirectories>
```

People can continue to use the old syntax. They will see warnings.

I should put up another PR for Gradle.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -512,12 +525,30 @@ String getFormat() {
...
+    String deprecatedProperty = getProperty(PropertyNames.EXTRA_DIRECTORY_PATH);
+    String newProperty = getProperty(PropertyNames.EXTRA_DIRECTORIES_PATHS);
+
+    List<File> deprecatedPaths = extraDirectory.getPaths();
+    List<File> newPaths = extraDirectories.getPaths();
+
+    if (deprecatedProperty != null) {
+      getLog()
+          .warn(
+              "The property 'jib.extraDirectory.path' is deprecated; "
```
loosebazooka: I think this error message can look a little more maven-y?
```c
"The configuration <jib><extraDirectory><path> is deprecated" // or something?
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -512,12 +525,30 @@ String getFormat() {
...
+                  + "use 'jib.extraDirectories.paths' instead");
+    }
+    if (!deprecatedPaths.isEmpty()) {
+      getLog().warn("<extraDirectory> is deprecated; use <extraDirectories> with <paths><path>.");
+    }
+
+    String property = newProperty != null ? newProperty : deprecatedProperty;
```
loosebazooka: should we display a warning when they use both?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -526,15 +557,27 @@ String getFormat() {
    * @return the configured extra layer file permissions
    */
   List<PermissionConfiguration> getExtraDirectoryPermissions() {
-    String property = getProperty(PropertyNames.EXTRA_DIRECTORY_PERMISSIONS);
+    String deprecatedProperty = getProperty(PropertyNames.EXTRA_DIRECTORY_PERMISSIONS);
+    String newProperty = getProperty(PropertyNames.EXTRA_DIRECTORIES_PERMISSIONS);
+    if (deprecatedProperty != null) {
+      getLog()
+          .warn(
+              "The property 'jib.extraDirectory.permissions' is deprecated; "
```
loosebazooka: Perhaps more maven-y message? and similar log warning when both properties used?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
```diff
@@ -154,7 +162,32 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+    // TODO: Should inform user about nonexistent directory if using custom directory.
+    String deprecatedProperty =
+        MavenProjectProperties.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH, project, session);
+    String newProperty =
+        MavenProjectProperties.getProperty(PropertyNames.EXTRA_DIRECTORIES_PATHS, project, session);
+
+    List<File> deprecatedPaths = extraDirectory.getPaths();
+    List<File> newPaths = extraDirectories.getPaths();
+
+    if (deprecatedProperty != null) {
```
loosebazooka: You may not need the add the deprecation warning here as this is mostly an internal task?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
briandealwis: This is specifically about this property though.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
briandealwis: That should fail the build

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
chanseokoh: Yeah, the usage is `-Djib.extraDirectory.path=...` or `<jib.extraDirectory.path>`. So they will understand.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
chanseokoh: Good idea. Lets' fail the build.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
chanseokoh: Resolved (https://github.com/GoogleContainerTools/jib/pull/1626#discussion_r276320616 and
https://github.com/GoogleContainerTools/jib/pull/1626#discussion_r276321076)

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojo.java
chanseokoh: Ah, the output is not visible, right? I'll remove the warnings.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -512,12 +525,40 @@ String getFormat() {
...
+    }
+    if (deprecatedProperty != null && newProperty != null) {
+      throw new IllegalArgumentException(
+          "You should not use the deprecated property 'jib.extraDirectory.path' while using "
+              + "the new 'jib.extraDirectories.paths'");
+    }
+    if (!deprecatedPaths.isEmpty() && !newPaths.isEmpty()) {
+      throw new IllegalArgumentException(
+          "You should not use the deprecated <extraDirectory> config while using "
+              + "the new <extraDirectories>");
```
loosebazooka: I don't know if we should be using "new". How does something like:
```java
"You cannot configure both <extraDirectory> and <extraDirectories>." // or something?
```

loosebazooka: similarly above with `paths`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
```diff
@@ -167,34 +167,53 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+        MavenProjectProperties.getProperty(PropertyNames.EXTRA_DIRECTORIES_PATHS, project, session);
+
+    if (deprecatedProperty != null && newProperty != null) {
+      throw new MojoExecutionException(
+          "You cannot configure both 'jib.extraDirectory.path' and 'jib.extraDirectories.paths'");
+    }
+
+    String property = newProperty != null ? newProperty : deprecatedProperty;
+    if (property != null) {
+      return Collections.singletonList(Paths.get(property));
```
chanseokoh: Hmm... looks like I missed this part. The new property can be a list.

chanseokoh: So I missed handling this as list in the previous PR at this location: https://github.com/GoogleContainerTools/jib/pull/1602/files#diff-1aee6421b98ba15074ed3f0da4755bc5R175

Fixed with the latest commit.

chanseokoh:
I noticed that `FilesMojo` and `FileMojoV2` may not be handling relative paths correctly (#1660), which is independent of this PR. Ready for review again.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -512,12 +525,38 @@ String getFormat() {
...
+    List<File> deprecatedPaths = extraDirectory.getPaths();
+    List<File> newPaths = extraDirectories.getPaths();
+
+    if (deprecatedProperty != null) {
+      getLog()
+          .warn(
+              "The property 'jib.extraDirectory.path' is deprecated; "
+                  + "use 'jib.extraDirectories.paths' instead");
+    }
+    if (!deprecatedPaths.isEmpty()) {
```
loosebazooka: I think there's some logic in here that we don't fully check in tests? I think we should add those in before we merge?

Issue #894, opened 2018-08-28
Expose classpath, main class, and other variables for reference from an explicit entrypoint


briandealwis:
Following on to #579 where it would be useful to be able to somehow reference the computed classpath from within an entrypoint definition.

For example, on Gradle:
```groovy
jib {
  container {
    entrypoint = ['/opt/startup.sh', '-classpath', classpath.join(':'), '-jvmflags', '"' + jvmFlags.join(' ') + '"']
  }
}
```

Would we want users to be able to change the classpath?
Should we expose the computed main-class?

remmeier:
in Gradle the correct behavior would be to set custom Configuration, e.g.

```
configurations {
    something
}

dependencies{
  something '...'
}

jib.configuration = configurations.something.
```

or with source sets like:

```
jib.sourceset = ...
```

The configuration can then be customized, inherited from other configurations, etc. 

but have not seen something suggesting for this to currently be possibe. In GradleLayerConfigurations things is hardcoded to go for the main source set and its classpath. For our use case we would like to change that completely, not just extending it because the image is only a "by-product" of something else and currently we are forced to create a setup an additional project next to it to build the image.

briandealwis:
As I understand them, @remmeier, source-sets are for collecting groups of Java source files, whereas this PR is at least to simplify adding other jars or directories, for things like JVM agents, to the classpath in the container image.  Which is not to say that supporting source-sets are not applicable.  I'm a relative newbie to Gradle, so would you mind describing what you want to do, and why you want to do it that way?

loosebazooka:
@remmeier .. to further what @briandealwis said: we are just allowing further customization of the classpath part of the final entrypoint we generate for the container image.

remmeier:
@briandealwis : sourceSets is not just about sources. It is also backed by configurations that specify how to compile and run the sources, like "test" sourcesets having  "testCompile" and "testRuntime" configurations. Together they make up the application. That is what is used to do the packaging in Jib as far as I have seen in the GradleLayerConfigurations. So it would be a good step to make that configurable. This way one can influence what JARs get packaged and added to the classpath. That is probably sufficient for 95% of the classpath manipulation use cases.

Then there are of course other JVM things like adding agents, etc. That is something not covered by the sourcesets/configurations. JavaExecSpec might be a good inspiration for that area. Or may even make use of it.

chanseokoh:
[...]

PR #1642, opened 2019-04-17
Add ability to add paths to the computed classpath

briandealwis:
Adds a new `extraClasspath` container element for Maven and Gradle to allow specifying additional paths for the container classpath.  This is part of #894.

Gradle:
```
jib {
  container {
    extraClasspath = ['/d1', '/d2']
  }
}
```

Maven:
```xml
<jib>
  <container>
    <extraClasspath>
      <path>/d1</path>
      <path>/d2</path>
    </extraClasspath>
  </container>
</jib>
```

Settings can also be set via `jib.container.extraClasspath` system and project properties.


jib-gradle-plugin/README.md
```diff
@@ -208,6 +208,7 @@ Property | Type | Default | Description
 --- | --- | --- | ---
 `appRoot` | `String` | `/app` | The root directory on the container where the app's contents are placed. Particularly useful for WAR-packaging projects to work with different Servlet engine base images by designating where to put exploded WAR contents; see [WAR usage](#war-projects) as an example.
 `args` | `List<String>` | *None* | Additional program arguments appended to the command to start the container (similar to Docker's [CMD](https://docs.docker.com/engine/reference/builder/#cmd) instruction in relation with [ENTRYPOINT](https://docs.docker.com/engine/reference/builder/#entrypoint)). In the default case where you do not set a custom `entrypoint`, this parameter is effectively the arguments to the main method of your Java application.
+`classpath` | `List<String>` | `/app/resources`, `/app/classes`, `/app/libs/* | The Java classpath to use.
```
chanseokoh: Should this be `extraClasspath`? Also the explanation seems too vague. Closing tick missing for `/app/libs/*` too.

 BTW, we update the doc after making a release.

jib-gradle-plugin/README.md
```diff
@@ -208,6 +208,7 @@ Property | Type | Default | Description
...
 `entrypoint` | `List<String>` | *None* | The command to start the container with (similar to Docker's [ENTRYPOINT](https://docs.docker.com/engine/reference/builder/#entrypoint) instruction). If set, then `jvmFlags` and `mainClass` are ignored. You may also set `jib.container.entrypoint = 'INHERIT'` to indicate that the `entrypoint` and `args` should be inherited from the base image.\*
```
chanseokoh: > If set, then `jvmFlags` and `mainClass` are ignored.

Should add `classpath` to above to state it'll be ignored.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -445,6 +454,9 @@ public void testExecute_complex()
         buildAndRunComplex(
             targetImage, "testuser2", "testpassword2", localRegistry2, "pom-complex.xml"));
     assertWorkingDirectory("", targetImage);
+    assertEntrypoint(
+        "[java -Xms512m -Xdebug -cp /app/resources:/app/classes:/app/libs/* com.test.HelloWorld]",
```
chanseokoh: Can we have a test where `extraClasspath` is in action?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -266,13 +274,13 @@ static PluginConfigurationProcessor processCommonConfiguration(
       return null;
     }
 
-    AbsoluteUnixPath appRoot =
-        getAppRootChecked(rawConfiguration, projectProperties.isWarProject());
+    List<String> classpath = new ArrayList<>(rawExtraClasspath.orElseGet(Collections::emptyList));
```
chanseokoh: nit: nothing wrong, but about `orElse(Collections.emptyList())`?

jib-gradle-plugin/README.md
briandealwis: Good catch.  I'll submit separately.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
briandealwis: 🤦‍♂️ I clearly did something wrong here as this is clearly missing the extraClasspath.

jib-gradle-plugin/src/test/resources/gradle/projects/simple/build.gradle
```diff
@@ -21,4 +21,7 @@ jib {
   extraDirectory {
     path = file('src/main/custom-extra-dir')
   }
+  container {
+    extraClasspath = ['/d1','/d2']
```
chanseokoh: Just realized `extraClasspath = ['/d1:/d2']` will work fine in the end. I feel it will be good to disallow that to prevent spreading confusion in usage. Given that this is Java classpath, it is not unreasonable to ban the use of `:` or `;` although they are valid filename characters on many common filesystems. People just shouldn't use them particularly for Java classpath in an image. What do you think?

briandealwis: I've been mulling this over…  The `jib.container.extraClasspath` property also benefits from this, so users can use the more-natural looking `-Djib.container.extraClasspath=/d1:/d2:/d3` rather than comma-separated values.  Considering that we don't actually perform any interpretation or evaluation of these values, I'm inclined to _laissez faire_.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -162,6 +162,8 @@ public void set(String image) {
 
     @Parameter private Map<String, String> environment = Collections.emptyMap();
 
+    @Nullable @Parameter private List<String> extraClasspath;
```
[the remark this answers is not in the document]
briandealwis: I'm thinking this is more like args (below).

jib-gradle-plugin/src/test/resources/gradle/projects/simple/build.gradle
chanseokoh: Hmm... one concern is that people may do `/d1;/d2;/d3` on Windows (which won't work) after finding a usage of `/d1:/d2:/d3` on the Internet.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -107,6 +108,23 @@ public void setEnvironment(Map<String, String> environment) {
     this.environment = environment;
   }
 
+  @Input
+  @Nullable
+  @Optional
+  public List<String> getExtraClasspath() {
+    if (System.getProperty(PropertyNames.CONTAINER_EXTRA_CLASSPATH) != null) {
+      // TODO: this should probably support Java-style ":" or ";" separators
```
chanseokoh: Should we support `;`? That may become confusing if we allow passing `/d1:/d2:/d2` as-is as a string.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleRawConfiguration.java
```diff
@@ -80,6 +80,11 @@ public AuthProperty getToAuth() {
     return Optional.ofNullable(jibExtension.getContainer().getArgs());
   }
 
+  @Override
+  public Optional<List<String>> getExtraClasspath() {
```
chanseokoh: We could remove `Optional`, since we made this non-`@Nullable`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -151,7 +151,10 @@ public void set(String image) {
     }
   }
 
-  /** Configuration for {@code container} parameter. */
+  /**
+   * Configuration for {@code container} parameter. Note that {@code entrypoint} and {@code args}
+   * are {@code @Nullable} do handle inheriting values from the base image.
```
chanseokoh: "do handle" --> "to handle"

I'd put the new comment as non-Javadoc comment as `// Note that ...` below the Javadoc; I think Javadoc usually for code consumers rather than holding internal implementation details.

jib-gradle-plugin/src/test/resources/gradle/projects/simple/build.gradle
loosebazooka: I'm leaning towards brian's solution here. Yes the user can use `:` or `;`, but hopefully users who go looking for this kind of advanced config already know what they are doing?

chanseokoh: Alright, users will eventually figure out the right usage if it fails to work. I also think the simplicity in implementation is valuable.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
chanseokoh: Maybe we should just remove this comment, since we will take the property as-is as a string as discussed above.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -37,6 +37,7 @@
   private List<String> jvmFlags = Collections.emptyList();
   private Map<String, String> environment = Collections.emptyMap();
   @Nullable private List<String> entrypoint;
+  @Nullable private List<String> extraClasspath;
```
chanseokoh: Sorry, missed this one. Assign `= Collections.emptyList()`; and remove `@Nullable`, like the `environment` above.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -107,6 +108,22 @@ public void setEnvironment(Map<String, String> environment) {
     this.environment = environment;
   }
 
+  @Input
+  @Nullable
```
chanseokoh: Remove `@Nullable`. I hope NullAway got this.

briandealwis: I've been having problems running builds locally that I haven't looked into.

PR #1644, opened 2019-04-18
Propagate labels from base image / propagate container config in case of V2.1 manifest

### LINKED ISSUE #1641: Container config not propagated when registry uses old V2.1 manifest
This normally works, but not when using `quay.io/sdase/openjdk-runtime:8-hotspot` as a base image. `docker inspect` shows that the base image has `Env`, `Labels`, etc.

This may be related to Quay using the old manifest V2.1; when I retag the base image, push it to Docker Hub (which generates manifest V2.2), and use it as a base image, it works.

### LINKED ISSUE #1643: Labels from base image are not propagated
Even if #595 marked fixed, currently labels are not being propagated. No idea if it ever worked or this is a regression. Should be an easy fix. Will fix it as part of #1641.

chanseokoh:
Fixes #1643.
Fixes #1641.

Travis is failing to install OpenJDK 11, so ignore that for the moment.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
@@ -143,10 +143,7 @@ public class JsonToImageTranslator {
       ContainerConfigurationTemplate containerConfigurationTemplate)
       throws BadContainerConfigurationFormatException {
 
-    List<HistoryEntry> historyObjects = containerConfigurationTemplate.getHistory();
-    for (HistoryEntry historyObject : historyObjects) {
-      imageBuilder.addHistory(historyObject);
-    }
+    containerConfigurationTemplate.getHistory().forEach(imageBuilder::addHistory);
 
     if (containerConfigurationTemplate.getCreated() != null) {
       try {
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
```diff
@@ -149,13 +161,8 @@
...
 
-    if (containerConfigurationTemplate.getContainerEntrypoint() != null) {
-      imageBuilder.setEntrypoint(containerConfigurationTemplate.getContainerEntrypoint());
-    }
-
-    if (containerConfigurationTemplate.getContainerCmd() != null) {
-      imageBuilder.setProgramArguments(containerConfigurationTemplate.getContainerCmd());
-    }
+    imageBuilder.setEntrypoint(containerConfigurationTemplate.getContainerEntrypoint());
+    imageBuilder.setProgramArguments(containerConfigurationTemplate.getContainerCmd());
```
briandealwis: So previously we would only overwrite entrypoint/args if set in the container configuration.  The code further above continues doing this for `created`, `architecture` and `os`, and further below.  Is there a reason to deviate from the previous pattern?

jib-core/src/test/resources/core/json/v21manifest.json
```diff
@@ -5,7 +5,7 @@
     {"blobSum":"sha256:5bd451067f9ab05e97cda8476c82f86d9b69c2dffb60a8ad2fe3723942544ab3"}
   ],
   "history": [
-    {"v1Compatibility":"some v1-compatible object"},
+    {"v1Compatibility": "{\"architecture\":\"amd64\",\"config\":{\"Env\":[\"JAVA_HOME=/opt/openjdk\",\"PATH=/opt/openjdk/bin\"],\"Entrypoint\":[\"/opt/openjdk/bin/java\"],\"Cmd\":[\"-version\"]},\"created\":\"2019-04-17T14:02:34.79404123Z\"}"},
```
briandealwis: Would is be possible to create a new test with a this container-configuration object?  The previous test wasn't wrong: the `v1Compatibility` could be some arbitrary object.

chanseokoh: The next element is `"another v1-compatible object"`, so at least we know the method doesn't fail with an arbitrary string. I just didn't like adding the `@VisibleForTesting` method `getV1Compatibility()` used only in testing and nowhere else.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslator.java
chanseokoh: At some point, looks like we changed `setEntrypoint()`, `setProgramArguments()`, etc. to accept `@Nullable`. As a builder, `imageBuilder` starts with null values, and passing null to them is like no-op hence OK. It also felt a bit weird to do the a check when the method can accept null. Sure, if `setEntrypoint()` was called on `imageBuilder` with a concrete value before and if we pass null then the value will be gone, but calling `setEntrypoint()` twice with different values cannot happen here.

PR #1647, opened 2019-04-19
Remove Blob conversion from ImageToTarballTranslator and TarStreamBuilder

chanseokoh:
I am going to do a serious of refactoring to our repo (not about our API visibility but about making our codebase healthy), and this is the first step. I have many more PRs to come.

When I first came to the repo, `Blob` was a very confusing class. As much as the name says, it is a very general class and can wrap up any object, anything that holds any information, or anything that may potentially be written to somewhere. It often disguises the actual object to the point of not being able to get what it really is. To me, often `Blob` looks like the Java `Object` class.

Another problem is the abuse of `Blob` to write something to `OutputStream`. With `Blob` having `toWrite(OutputStream)`, examples of prevalent abuse in our codebase are

1. Convert anything to `Blob` blindly for the sake of writing to `OutputStream`.
1. Convert anything to `Blob` blindly to compute a digest of a content. (`Blob.writeTo()` returns a digest.)

Currently, `ImageToTarballTranslator` (which effectively represents a tarball for an `Image`) is doing a convoluted way of writing it to a file or a Docker daemon:

`ImageToTarballTranslator` --> `Blob` --> write to an output stream

In this PR, I'm making this straightforward and intuitive by
1. just making `ImageToTarballTranslator` write to an output stream; and
1. decoupling `Blob` from `ImageToTarballTranslator`.

Also, `DockerClient.load()` accepting `Blob` seems is too general and obscure, which doesn't look very different from `load(Object anything)`.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
@@ -23,7 +23,7 @@ import com.google.cloud.tools.jib.builder.BuildStepType;
 import com.google.cloud.tools.jib.builder.ProgressEventDispatcher;
 import com.google.cloud.tools.jib.configuration.BuildConfiguration;
 import com.google.cloud.tools.jib.docker.DockerClient;
-import com.google.cloud.tools.jib.docker.ImageToTarballTranslator;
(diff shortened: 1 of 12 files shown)


chanseokoh:
Similarly to `ImageToTarballTranslator`, I also removed `Blob` from `TarStreamBuilder`, which it had `toBlob()` only to do `toBlob().writeTo(out)`.

PR #1649, opened 2019-04-19
Add default timestamp to not-yet-existant parent dirs

### LINKED ISSUE #1648: jib:build is not reproducible due to directory timestamps not being set
**Description of the issue**:
When I build the code in container,  it  will rebuild all the layer every time. 

**Expected behavior**:
 I don’t have to wait for Docker to rebuild your entire Java application - just deploy the layers that changed.
**Steps to reproduce**: 
1. clone https://github.com/GoogleContainerTools/jib.git , cd  to `examples/spring-boot`
2. build with:
```
docker run -v `pwd`:/tmp/build -w /tmp/build -v $HOME/.m2:/root/.m2 maven:3-jdk-8 mvn clean compile jib:build -Djib.from.image=openjdk:8-alpine -Dimage=<myimage>
```
3. `docker pull` the new image.  I have to pull all the 3 layers, include the 14.38MB dependency layer every time .

**Environment**: 
macos 10.14.3, docker Version 2.0.0.3

loosebazooka:
So this defines a strong behavior for parents that don't exist in the layer yet.
1. order matters, a directory can only be configured if it's added to the layer before any of its children
2. for a file with parent(s) that doesn't exist yet, a default modified time is applied.

fixes #1648 

NOTE: This *does not* fix the out of order issue for defining directory permissions. That should be fixed in a followup (#1650)

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
@@ -193,8 +193,7 @@ public class ReproducibleLayerBuilderTest {
     Path ignoredParent = Files.createDirectories(testRoot.resolve("bbb-ignored"));
     Path fileB = Files.createFile(ignoredParent.resolve("fileB"));
     Path fileC =
-        Files.createFile(
-            Files.createDirectories(testRoot.resolve("ccc-absent")).resolve("fileC"));
+        Files.createFile(Files.createDirectories(testRoot.resolve("ccc-absent")).resolve("fileC"));
 
     Blob layer =
         new ReproducibleLayerBuilder(
@@ -260,7 +259,6 @@ public class ReproducibleLayerBuilderTest {
 
       // we don't care about fileC
     }
-
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
```diff
@@ -183,6 +183,92 @@ public void testToBlob_reproducibility() throws IOException {
...
+                        FilePermissions.fromOctalString("444"),
+                        Instant.ofEpochMilli(4444)),
+                    new LayerEntry(
+                        file3,
+                        AbsoluteUnixPath.get("/root/absenteeParent/file3"),
+                        FilePermissions.fromOctalString("555"),
+                        Instant.ofEpochMilli(5555))))
+            .build();
+
+    Blob equivalentLayer =
```
loosebazooka: It's possible I took the wrong approach to the test, this is pretty hard to read, perhaps I should verify each entry by iterating over the list of entries after writing to a tarfile (I didn't do this because it does things in a crazy alphabetical order)

loosebazooka:
probably going to have to do a cherry pick release off `1.1.1`? wdyt?

chanseokoh:
You mean 1.1.2, right? Sounds right.

jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
```diff
@@ -183,6 +183,84 @@ public void testToBlob_reproducibility() throws IOException {
...
+
+      // parentAAA (custom permissions, custom timestamp)
+      TarArchiveEntry rootParentAAA = in.getNextTarEntry();
+      Assert.assertEquals(040111, rootParentAAA.getMode());
+      Assert.assertEquals(Instant.ofEpochSecond(10), rootParentAAA.getModTime().toInstant());
+
+      // skip over fileA
+      in.getNextTarEntry();
+
+      // parentBBB (default permissions - ignored custom permissions, since fileB added first)
```
chanseokoh: Add a TODO that this should really be fixed and return `040444` and `Instant.ofEpochSecond(40)`?

PR #1655, opened 2019-04-22
Upgrade Jib version to 1.1.2

chanseokoh:


briandealwis:
LGTM though `jib-gradle-plugin/gradle.properties` and `jib-maven-plugin/pom.xml` still have 1.1.2-SNAPSHOT, so there is no 1.1.3-SNAPSHOT.

chanseokoh:
Oh, my mistake. Good catch.

PR #1659, opened 2019-04-22
Refactor public API to hide internal methods

chanseokoh:
Follow-up from #1622.

I removed `BuildSteps`, as I thought having that as another thin layer of abstraction doesn't provide much value than adding one more class to keep track of regarding running async build steps.


TadCordle:
>I removed BuildSteps

Did you mean to delete the class itself? I noticed you removed all of (at least a lot of) the places it was used, but the class itself is still there and even has a few changes in it.

chanseokoh:
Oops. Will delete `BuildSteps`.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/StepsRunnerIntegrationTest.java
```diff
@@ -63,8 +64,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-/** Integration tests for {@link BuildSteps}. */
-public class BuildStepsIntegrationTest {
+/** Integration tests for {@link StepsRunnerFactory}. */
+public class StepsRunnerIntegrationTest {
```
TadCordle: Do you think it would be possible to consolidate a bit by turning this into `ContainerizerIntegrationTest` and moving the `StepsRunnerFactory` stuff into `Containerizer`, by just building the `StepsRunner`s within each of the `to` overloads?

chanseokoh: Ah, if we look at this from a different angle and think of this as `ContainerizerIntegrationTest`, then we may be able to build `StepsRunner` in `Conatinerizer.to()` as before. Let me try that.

chanseokoh: I think it worked out nicely.

jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
```diff
@@ -35,7 +32,7 @@
  * }</pre>
  */
 // TODO: Add tests once JibContainerBuilder#containerize() is added.
```
TadCordle: I feel like we can remove some of these TODOs

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
```diff
@@ -44,7 +43,10 @@
 import javax.annotation.Nullable;
 import org.apache.http.conn.HttpHostConnectException;
 
-/** Runs a {@link BuildSteps} and builds helpful error messages. */
+/** Runs a build steps and builds helpful error messages. */
+// TODO: find a better name for the class, since there is no "steps" visible or mentioned at all;
```
TadCordle: Maybe just something like `JibRunner`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
chanseokoh: Agreed, we now have many tests for `containerize()`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/BuildStepsRunner.java
chanseokoh: Nice. The name fits very well.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/builder/ContainerizerIntegrationTest.java
```diff
@@ -341,43 +319,65 @@ public void testSteps_forBuildToDockerDaemon_multipleTags()
...
         "Hello, world. An argument.\n", new Command("docker", "run", "--rm", "testtar").run());
   }
 
-  private BuildConfiguration.Builder getBuildConfigurationBuilder(
-      ImageReference baseImage, ImageReference targetImage) throws IOException {
-    ImageConfiguration baseImageConfiguration = ImageConfiguration.builder(baseImage).build();
-    ImageConfiguration targetImageConfiguration = ImageConfiguration.builder(targetImage).build();
-    ContainerConfiguration containerConfiguration =
-        ContainerConfiguration.builder()
+  private JibContainer buildRegistryImage(
```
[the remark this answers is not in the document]
chanseokoh: It turned out a bit verbose with, e.g., `Containerizer.to(DockerDaemonImage.named(...))`, so I will retain the helpers.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JibRunner.java
```diff
@@ -44,8 +43,8 @@
 import javax.annotation.Nullable;
 import org.apache.http.conn.HttpHostConnectException;
 
-/** Runs a {@link BuildSteps} and builds helpful error messages. */
-public class BuildStepsRunner {
+/** Runs Jib and builds helpful error messages. */
+public class JibRunner {
```
loosebazooka: Nit: Not sure we need to use `Jib` in the name here. Perhaps `BuildRunner`? I'm okay either way though.

chanseokoh: The previous commit had this [now-gone comment](https://github.com/GoogleContainerTools/jib/pull/1659/commits/4af8648bfde1c644ece9ccb669748ddb236766b1#diff-ea953f6b77cf30e8f5d4a0a34fd3a74bL47). Things had changed when we had our initial jib-core public API, and what this class has been doing since then is just "jibConatinerBuilder.containerize(containerizer)". I think the name "JibRunner" [Tad also suggested](https://github.com/GoogleContainerTools/jib/pull/1659#discussion_r278749078) fits well?

loosebazooka: Sure, I was working under the assumption that using `Jib` doesn't add much actual descriptive value given that this is already in the `Jib` project namespace and so `JibRunner` can mean anything related to Jib?. But again, this is minor and I understand we can disagree.

chanseokoh: I get what you're saying. If we had a practice to prefix "Jib" to anything like "JibBuildConfiguration", "JibCredential", I agree that's meaningless as we are under the Jib namespace and we should avoid that. But since we don't execute such practice, I think "JibRunner" can carry the meaning of someone who runs Jib.

That said, maybe I should go with "JibBuildRunner". I had thought about "BuildRunner" but didn't like it because it is not descriptive enough either; it's not clear what kind of build it runs, given that there are too many things that are buildable. That is, it's unclear what the build result is. Here, I'd like to make it explicit that what we run is (a Jib build).

Issue #718, opened 2018-07-26
Jib should respect `—offline`


briandealwis:
**Description of the issue**:

Jib should respect the `—offline` flags for Gradle and Maven: jib should use the cached _from_ image,  erroring if it doesn't exist, and `:build` should fail.  This might help address some of the issues seen by users in high-latency zones (e.g. #648, #713).

coollog:
This would actually be a fairly involved change since Jib doesn't currently cache the base image manifests. To resolve this issue, Jib would need to:

- cache base image manifests, with the image reference they refer to
- cache the container configuration for those base images as well
- fetch cached base image manifests via image references
- only do this when `--offline` is set, but still pull manifests if `--offline` is not set to check for any updates

YangGuang001:
recently, I use the jib, but my base image is very big, it takes me a long time to generate the image.
I think the Jib only need to check the manifests.

coollog:
Hi @YangGuang001 , the initial run of Jib will download the entire base image. However, subsequent runs should use the cached base image and should not require re-downloading the base image layers. Do your subsequent builds run faster?

PierreBesson:
I must stress the importance of this issue. This is currently a pain point for a lot of JHipster users, especially those who don't have access to a private docker registry on their developer machines.
Currently, we have a big warning in our [jib usage docs](https://www.jhipster.tech/docker-compose/#3) to educate our users on this problem.

chanseokoh:
> Currently, we have a big warning in our [jib usage docs](https://www.jhipster.tech/docker-compose/#3)

@PierreBesson I've read the warning. The warning says the build may fail because of network issues or because you are not logged in a Docker registry. Here are a few things I'd like to point out:
- You'll be using public images, so I don't think the build will fail because you are not logged in. I don't really get the warning.
- Even with `--offline`, the build will fail for the first time when Jib has not yet cached the base image. (Also note that Jib's cache has nothing to do with Docker's.) Your warning gives me the (false) impression that the build will not fail once Jib supports `--offline`, but I think this can be misleading.
- `exportDockerContext` has been removed and is no longer available in recent versions.

PierreBesson:
Thanks for this explanation. I will rework this part of the docs to be more technically accurate. 
I see now that being logged in or not is not really the issue for public base images. I think people might still encounter problems with their proxies though.

> Also note that Jib's cache has nothing to do with Docker's.

I didn't know that, it seems important to explain this in our docs.

[...]

PR #1666, opened 2019-04-24
Add cache methods for saving manifest and container config

TadCordle:
Towards #718.

Adds `writeMetadata` methods to `Cache`, `CacheStorage`, and `DefaultCacheStorageWriter`, as well as helpers for getting the image metadata directory in `DefaultCacheStorageFiles`.

Mostly ignored file-locking for now; I did reuse some methods which involve file locking, but I'll probably need to take another look in the future when the locking design is nailed down.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
@@ -18,7 +18,10 @@ package com.google.cloud.tools.jib.cache;
 
 import com.google.cloud.tools.jib.blob.Blob;
 import com.google.cloud.tools.jib.image.DescriptorDigest;
+import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.LayerEntry;
+import com.google.cloud.tools.jib.image.json.ContainerConfigurationTemplate;
+import com.google.cloud.tools.jib.image.json.ManifestTemplate;
 import com.google.common.collect.ImmutableList;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -53,6 +56,22 @@ public class Cache {
 
   private Cache(CacheStorage cacheStorage) {
     this.cacheStorage = cacheStorage;
+  }
+
+  /**
+   * Saves an image manifest and container configuration.
+   *
+   * @param imageReference the image reference the save the manifest and container configuration for
+   * @param manifestTemplate the manifest
+   * @param containerConfigurationTemplate the container configuration
+   * @throws IOException if an I/O exception occurs
+   */
+  public void writeMetadata(
+      ImageReference imageReference,
+      ManifestTemplate manifestTemplate,
+      ContainerConfigurationTemplate containerConfigurationTemplate)
+      throws IOException {
+    cacheStorage.writeMetadata(imageReference, manifestTemplate, containerConfigurationTemplate);
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
@@ -20,8 +20,9 @@ import com.google.cloud.tools.jib.blob.Blob;
 import com.google.cloud.tools.jib.image.DescriptorDigest;
 import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.LayerEntry;
+import com.google.cloud.tools.jib.image.json.BuildableManifestTemplate;
 import com.google.cloud.tools.jib.image.json.ContainerConfigurationTemplate;
-import com.google.cloud.tools.jib.image.json.ManifestTemplate;
+import com.google.cloud.tools.jib.image.json.V21ManifestTemplate;
 import com.google.common.collect.ImmutableList;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -62,16 +63,28 @@ public class Cache {
    * Saves an image manifest and container configuration.
    *
    * @param imageReference the image reference the save the manifest and container configuration for
-   * @param manifestTemplate the manifest
+   * @param manifestTemplate the V2.2 or OCI manifest
    * @param containerConfigurationTemplate the container configuration
    * @throws IOException if an I/O exception occurs
    */
   public void writeMetadata(
       ImageReference imageReference,
-      ManifestTemplate manifestTemplate,
+      BuildableManifestTemplate manifestTemplate,
       ContainerConfigurationTemplate containerConfigurationTemplate)
       throws IOException {
     cacheStorage.writeMetadata(imageReference, manifestTemplate, containerConfigurationTemplate);
+  }
+
+  /**
+   * Saves a V2.1 image manifest.
+   *
+   * @param imageReference the image reference the save the manifest and container configuration for
+   * @param manifestTemplate the V2.1 manifest
+   * @throws IOException if an I/O exception occurs
+   */
+  public void writeMetadata(ImageReference imageReference, V21ManifestTemplate manifestTemplate)
+      throws IOException {
+    cacheStorage.writeMetadata(imageReference, manifestTemplate);
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
@@ -60,9 +60,9 @@ public class Cache {
   }
 
   /**
-   * Saves an image manifest and container configuration.
+   * Saves a manifest and container configuration for a V2.2 or OCI image.
    *
-   * @param imageReference the image reference the save the manifest and container configuration for
+   * @param imageReference the image reference to save the manifest and container configuration for
    * @param manifestTemplate the V2.2 or OCI manifest
    * @param containerConfigurationTemplate the container configuration
    * @throws IOException if an I/O exception occurs
@@ -78,7 +78,7 @@ public class Cache {
   /**
    * Saves a V2.1 image manifest.
    *
-   * @param imageReference the image reference the save the manifest and container configuration for
+   * @param imageReference the image reference to save the manifest and container configuration for
    * @param manifestTemplate the V2.1 manifest
    * @throws IOException if an I/O exception occurs
    */
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
@@ -130,12 +130,14 @@ class DefaultCacheStorageFiles {
    *
    * @param imageReference the image reference
    * @return a path in the form of {@code
-   *     (jib-cache)/images/registry[/port]/repository/(tag|digest-type/digest)}
+   *     (jib-cache)/images/registry[!port]/repository!(tag|digest-type!digest)}
    */
   Path getImageDirectory(ImageReference imageReference) {
-    // Split image reference on '/', ':', and '@' to build directory structure
-    Iterable<String> directories =
-        Splitter.onPattern("\\/|:|@").split(imageReference.toStringWithTag());
+    // Replace ':' and '@' with '!' to avoid directory-naming restrictions
+    String replacedReference = imageReference.toStringWithTag().replace(':', '!').replace('@', '!');
+
+    // Split image reference on '/' to build directory structure
+    Iterable<String> directories = Splitter.on('/').split(replacedReference);
     Path destination = getImagesDirectory();
     for (String dir : directories) {
       destination = destination.resolve(dir);
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
@@ -17,6 +17,8 @@
 package com.google.cloud.tools.jib.cache;
 
 import com.google.cloud.tools.jib.image.DescriptorDigest;
+import com.google.cloud.tools.jib.image.ImageReference;
+import com.google.common.base.Splitter;
 import java.nio.file.Path;
 import java.security.DigestException;
 
@@ -24,6 +26,7 @@ import java.security.DigestException;
 class DefaultCacheStorageFiles {
 
   private static final String LAYERS_DIRECTORY = "layers";
+  private static final String IMAGES_DIRECTORY = "images";
   private static final String SELECTORS_DIRECTORY = "selectors";
   private static final String TEMPORARY_DIRECTORY = "tmp";
   private static final String TEMPORARY_LAYER_FILE_NAME = ".tmp.layer";
@@ -111,6 +114,33 @@ class DefaultCacheStorageFiles {
    */
   Path getLayerDirectory(DescriptorDigest layerDigest) {
     return getLayersDirectory().resolve(layerDigest.getHash());
+  }
+
+  /**
+   * Gets the directory to store the image manifest and configuration.
+   *
+   * @return the directory for the image manifest and configuration
+   */
+  Path getImagesDirectory() {
+    return cacheDirectory.resolve(IMAGES_DIRECTORY);
+  }
+
+  /**
+   * Gets the directory corresponding to the given image reference.
+   *
+   * @param imageReference the image reference
+   * @return a path in the form of {@code
+   *     (jib-cache)/images/registry[/port]/repository/(tag|digest-type/digest)}
+   */
+  Path getImageDirectory(ImageReference imageReference) {
+    // Split image reference on '/', ':', and '@' to build directory structure
+    Iterable<String> directories =
+        Splitter.onPattern("\\/|:|@").split(imageReference.toStringWithTag());
+    Path destination = getImagesDirectory();
+    for (String dir : directories) {
+      destination = destination.resolve(dir);
+    }
+    return destination;
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageWriter.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageWriter.java
@@ -27,7 +27,6 @@ import com.google.cloud.tools.jib.image.ImageReference;
 import com.google.cloud.tools.jib.image.json.BuildableManifestTemplate;
 import com.google.cloud.tools.jib.image.json.ContainerConfigurationTemplate;
 import com.google.cloud.tools.jib.image.json.V21ManifestTemplate;
-import com.google.cloud.tools.jib.json.JsonTemplate;
 import com.google.cloud.tools.jib.json.JsonTemplateMapper;
 import com.google.common.base.Preconditions;
 import com.google.common.io.ByteStreams;
@@ -110,18 +109,16 @@ class DefaultCacheStorageWriter {
   }
 
   /**
-   * Writes a json template to the destination path by writing to a temporary file then moving the
-   * file.
+   * Writes a blob to the destination path by writing to a temporary file then moving the file.
    *
-   * @param jsonTemplate the json template
+   * @param blob the blob
    * @param destination the destination path
    * @throws IOException if an I/O exception occurs
    */
-  private static void writeMetadata(JsonTemplate jsonTemplate, Path destination)
-      throws IOException {
+  private static void writeBlob(Blob blob, Path destination) throws IOException {
     Path temporaryFile = Files.createTempFile(destination.getParent(), null, null);
     temporaryFile.toFile().deleteOnExit();
-    Blobs.writeToFile(JsonTemplateMapper.toBlob(jsonTemplate), temporaryFile);
+    Blobs.writeToFile(blob, temporaryFile);
 
     // Attempts an atomic move first, and falls back to non-atomic if the file system does not
     // support atomic moves.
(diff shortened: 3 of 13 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
```diff
@@ -55,6 +58,22 @@ private Cache(CacheStorage cacheStorage) {
...
+   *
+   * @param imageReference the image reference the save the manifest and container configuration for
+   * @param manifestTemplate the manifest
+   * @param containerConfigurationTemplate the container configuration
+   * @throws IOException if an I/O exception occurs
+   */
+  public void writeMetadata(
+      ImageReference imageReference,
+      ManifestTemplate manifestTemplate,
+      ContainerConfigurationTemplate containerConfigurationTemplate)
```
chanseokoh: Have you thought about the Docker manifest V2.1 case? There's no separate container config file, and the config is embedded in the manifest. Maybe it's better to have different method signatures for V2.1 and V2.2 (+ OCI)? (I haven't looked into OCI manifest format, but I guess that's very similar to Docker V2.2 and doesn't need special handling?)

chanseokoh: See https://github.com/GoogleContainerTools/jib/issues/1641#issuecomment-484187937 as an example of V2.1.

TadCordle: Good point, I think it makes sense to have 2 signatures (`(ImageReference, BuildableManifestTemplate, ContainerConfigurationTemplate)` and `(ImageReference, V21ManifestTemplate)`?)

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
```diff
@@ -113,6 +116,33 @@ Path getLayerDirectory(DescriptorDigest layerDigest) {
...
+
+  /**
+   * Gets the directory corresponding to the given image reference.
+   *
+   * @param imageReference the image reference
+   * @return a path in the form of {@code
+   *     (jib-cache)/images/registry[/port]/repository/(tag|digest-type/digest)}
+   */
+  Path getImageDirectory(ImageReference imageReference) {
+    // Split image reference on '/', ':', and '@' to build directory structure
```
chanseokoh: This means `my-repo:3000` and `my-repo/3000` will point to the same directory. Some random ideas:

Use some kind of mapping function (one-way hashing, encoding, or whatever) from a char to another that can be safely used as a filename on all filesystems. We may apply the mapping function to the entire image reference or only some portions (like some special characters which may or may not include `/`). But still, probably it would be good to have directory structures.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
TadCordle: Actually, maybe it'd make sense to just make `containerConfigurationTemplate` `@Nullable`, and if it's null it fills in for the default os/arch and skips writing the configuration file?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
TadCordle: Maybe we can just use `_` in place of of `:` in the registry name. `_` is an invalid character in the registry component of an image reference, but a valid path character.

chanseokoh: Unfortunately, I could use `_` with `docker tag` on both the registry component part and the tag part.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
TadCordle: Nevermind, it's redundant to write out the config if it exists in the V2.1 manifest. Just going to go with 2 signatures.

chanseokoh: I think `@Nullable` can work. It would be good to think about the corresponding signature for reading the files.

chanseokoh: Ah, didn't see your edit before I put the last comment.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
TadCordle: Ok, I was misremembering then.

TadCordle: Maybe we can use `!` then.

chanseokoh: I like the simple approach. In sum, converting `:` and `@` to `!`, right? @GoogleContainerTools/java-tools-build any objections?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageWriter.java
```diff
@@ -303,4 +364,15 @@ private void writeSelector(DescriptorDigest selector, DescriptorDigest layerDige
       Files.move(temporarySelectorFile, selectorFile, StandardCopyOption.REPLACE_EXISTING);
     }
   }
+
+  private void writeMetadataFile(JsonTemplate template, Path destination) throws IOException {
+    Path temporaryFile = Files.createTempFile(null, null);
+    temporaryFile.toFile().deleteOnExit();
```
chanseokoh: Can you double-check this does not delete the renamed file after JVM terminates?

TadCordle: This is the same logic we use to write the selectors, so it should be fine. In fact, I can probably replace the end of `writeSelector()` with this method (just realized I need to add the try-catch back in though)

chanseokoh: Sounds good.

> In fact, I can probably replace the end of `writeSelector()` with this method

Probably later, because we'll make lock-related changes.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageFiles.java
loosebazooka: Are `@` and `:` only allowed in specific sections?
- `:` only for port
- `@` only for digest
?

are `.` allowed in the image reference?

loosebazooka: I would prefer `.` to `!`, exclamation looks a little crazy in filenames

TadCordle: `:` is used for port, (`registry:port/...`), tag (`registry/repo:tag`), and digest (`registry/repo@sha256:aaaaa...`). And `.` is a valid character to use anywhere in the image reference, so I don't think we should use that.

TadCordle: `!` does look weird, but I don't think it matters whether or not we make the paths pretty. It just needs to not be error-prone and produce a unique conversion from image reference to path.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageWriterTest.java
```diff
@@ -99,6 +108,85 @@ public void testWrite_uncompressed() throws IOException {
...
+        JsonTemplateMapper.readJsonFromFile(manifestJsonFile, V22ManifestTemplate.class);
+    ImageReference imageReference = ImageReference.parse("image.reference/project/thing:tag");
+
+    new DefaultCacheStorageWriter(defaultCacheStorageFiles)
+        .writeMetadata(imageReference, manifestTemplate, containerConfigurationTemplate);
+
+    Assert.assertTrue(
+        Files.exists(
+            defaultCacheStorageFiles
+                .getImageDirectory(imageReference)
```
chanseokoh: For a similar reason, I often prefer using a constant for expected values like `Files.exists(Paths.get(cacheRoot, "sub", "directory", "manifest.wasm.js.json"))`. For example, suppose `getImageDirectory()` became buggy and always returns just the cache root. The test will pass because `writeMedata()` will use the same buggy `getImageDirectory()`.

chanseokoh:
You will do the proper locking in this PR, right?

TadCordle:
Was going to follow-up, but yeah on second thought it seems simple enough that I could do it now.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * @param lockFile the path of the lock file
+   * @return a new {@link LockFile} that can be released later
+   * @throws IOException if creating the lock file fails
+   */
+  public static LockFile lock(Path lockFile) throws IOException {
+    Files.createDirectories(lockFile.getParent());
+    while (true) {
+      try {
+        FileLock fileLock = new FileOutputStream(lockFile.toFile()).getChannel().tryLock();
```
chanseokoh: @dgageot is there a reason for manual polling instead of just using the blocking `FileChannel.lock()`. I can't think of why.

chanseokoh: Also this code catches the general `Exception`, so it doesn't honor thread interruptions. Seems too stringent blocking forever until it gets a lock. It should probably be interruptible?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageWriter.java
```diff
@@ -101,6 +110,36 @@ private static DescriptorDigest getDiffIdByDecompressingFile(Path compressedFile
...
+   * @param jsonTemplate the json template
+   * @param destination the destination path
+   * @throws IOException if an I/O exception occurs
+   */
+  private static void writeMetadata(JsonTemplate jsonTemplate, Path destination)
+      throws IOException {
+    try (LockFile ignored1 = LockFile.lock(Paths.get(destination.toString() + ".lock"))) {
+      Path temporaryFile = Files.createTempFile(destination.getParent(), null, null);
+      temporaryFile.toFile().deleteOnExit();
+      Blobs.writeToFileWithLock(JsonTemplateMapper.toBlob(jsonTemplate), temporaryFile);
```
chanseokoh: Since we don't need a lock for the temp file, I'd like to avoid the "write-with-lock" methods. It may confuse people who read this code later.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageWriter.java
```diff
@@ -101,6 +110,36 @@ private static DescriptorDigest getDiffIdByDecompressingFile(Path compressedFile
...
+   * Writes a json template to the destination path by writing to a temporary file then moving the
+   * file.
+   *
+   * @param jsonTemplate the json template
+   * @param destination the destination path
+   * @throws IOException if an I/O exception occurs
+   */
+  private static void writeMetadata(JsonTemplate jsonTemplate, Path destination)
+      throws IOException {
+    try (LockFile ignored1 = LockFile.lock(Paths.get(destination.toString() + ".lock"))) {
```
chanseokoh: Just curious: does `FileLock` support read/write locks, or is it always mutual exclusion? (Not that it matters even if mutual exclusion for reading.)

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
TadCordle: I tried using `lock()`, and it seems to throw `OverlappingFileLockException` instead of blocking if it's called in multiple threads in the same JVM.

loosebazooka: So maybe it needs to be synchronized on the same jvm?

loosebazooka: I mean using the `synchronized` keyword.

TadCordle: Didn't seem to make a difference.

dgageot: When I used this code, it couldn't be called by multiple threads in the same JVM.
Why use `tryLock()` instead of `lock()` in a loop? I don't remember ;-).

loosebazooka: Ah yeah, no I misunderstood the problem. Still I think we can use some kind of semaphore here.

loosebazooka: Maybe a guava Striped<Semaphore> or something?
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/util/concurrent/Striped.html

loosebazooka: I think `Striped` by be wrong, perhaps you can just use a `concurrentHashMap<Path, Lock>`.

loosebazooka: Sorry I'm just throwing things out here, but something along these lines.

TadCordle: We'll try that for now

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Creates a lock file.
+   *
+   * @param lockFile the path of the lock file
+   * @return a new {@link LockFile} that can be released later
+   * @throws IOException if creating the lock file fails
+   */
+  public static LockFile lock(Path lockFile) throws IOException {
+    LOCK_MAP.putIfAbsent(lockFile, new ReentrantLock());
+    Preconditions.checkNotNull(LOCK_MAP.get(lockFile)).lock();
```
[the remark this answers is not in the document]
TadCordle: `tryLock` never blocks, and `lock` is supposed to, I think (unless it's called from 2 threads in the same JVM?). A timeout might be a good idea, any suggestions for how long it should be?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /** Releases the lock file. */
+  @Override
+  public void close() {
+    try {
+      fileLock.release();
+    } catch (IOException ex) {
+      throw new IllegalStateException("Unable to release lock", ex);
+    }
```
chanseokoh: Perhaps

```java
try {
  fileLock.release();
} catch (...) {
} finally {
  Verify.verify(lockMap.get(lockFile)).unlock();

  try {
    Files.delete(lockFile);
  } catch (...) {
  }
}
```

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,78 @@
...
+   * Creates a lock file.
+   *
+   * @param lockFile the path of the lock file
+   * @return a new {@link LockFile} that can be released later
+   * @throws IOException if creating the lock file fails
+   */
+  public static LockFile lock(Path lockFile) throws IOException {
+    lockMap.putIfAbsent(lockFile, new ReentrantLock());
+    try {
+      Preconditions.checkNotNull(lockMap.get(lockFile)).tryLock(5, TimeUnit.MINUTES);
```
chanseokoh: Did anyone say to use a timed lock? I'd go with `lock()` that blocks indefinitely. `FileChannel.lock()` will anyway block indefinitely.

And knowing that `FileChannel.lock()` throws a checked `FileLockInterruptionException` which extends `IOException`, in this case I think it is OK to do

```java
  try {
    Preconditions.checkNotNull(lockMap.get(lockFile)).lock();
  } catch (InterruptedException ex) {
    throw new IOException("Interrupted while trying to acquire lock", ex);
  }
```

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Creates a lock file.
+   *
+   * @param lockFile the path of the lock file
+   * @return a new {@link LockFile} that can be released later
+   * @throws IOException if creating the lock file fails
+   */
+  public static LockFile lock(Path lockFile) throws IOException {
+    lockMap.putIfAbsent(lockFile, new ReentrantLock());
```
chanseokoh: Maybe `computeIfAbsent()`, i.e.,`Lock lock = lockMap.computeIfAbsent(lockFile, key -> new ReentrantLock())` and merge `putIfAbsent()` and `get()` into one method call. (Haven't tried this.) I see `computeIfAbsent` returns the "current" (existing or computed) value.

Another tiny advantage is that a `ReentrantLock()` object is created only when it is absent with `computeIfAbsent()`. (Currently, the argument `new ReentrantLock()` is always evaluated.)

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * Creates a lock file.
+   *
+   * @param lockFile the path of the lock file
+   * @return a new {@link LockFile} that can be released later
+   * @throws IOException if creating the lock file fails
+   */
+  public static LockFile lock(Path lockFile) throws IOException {
+    lockMap.computeIfAbsent(lockFile, key -> new ReentrantLock()).lock();
```
chanseokoh: Ah, `lock()` ignores thread interruption. I'd go with `lockInterruptibly()` so that it reacts to interruption.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,74 @@
...
+  /** Releases the lock file. */
+  @Override
+  public void close() {
+    try {
+      fileLock.release();
+
+    } catch (IOException ex) {
+      throw new IllegalStateException("Unable to release lock", ex);
+
+    } finally {
```
chanseokoh: Sorry, but for the sake of completeness, I think this should further be

```java
  } finally {
    try {
      ....unlock();
      Files.delete(lockFile);
    } catch (IllegalMonitorStateException | IOException ignored) {
    }
  }
```
because if this instance failed to `lock()` at the beginning due to, e.g., interruption, it will throw `IllegalMonitorStateException` when the caller escapes from the try-resource block: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html#unlock--

jib-core/src/main/java/com/google/cloud/tools/jib/cache/DefaultCacheStorageWriter.java
```diff
@@ -204,6 +243,42 @@ CachedLayer write(UncompressedCacheWrite uncompressedCacheWrite) throws IOExcept
...
+      ContainerConfigurationTemplate containerConfiguration)
+      throws IOException {
+    Preconditions.checkNotNull(manifestTemplate.getContainerConfiguration());
+    Preconditions.checkNotNull(manifestTemplate.getContainerConfiguration().getDigest());
+
+    Path imageDirectory = defaultCacheStorageFiles.getImageDirectory(imageReference);
+    Files.createDirectories(imageDirectory);
+
+    writeMetadata(manifestTemplate, imageDirectory.resolve("manifest.json"));
+    writeMetadata(containerConfiguration, imageDirectory.resolve("config.json"));
```
chanseokoh: Oh! Sorry, missed this important issue. These two writes are sequential, so we need to guard these two with a single lock file so that they are atomic. Directory creation is fine without syncing though, but not sure.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Creates a lock file.
+   *
+   * @param lockFile the path of the lock file
+   * @return a new {@link LockFile} that can be released later
+   * @throws IOException if creating the lock file fails
+   */
+  public static LockFile lock(Path lockFile) throws IOException {
+    try {
+      lockMap.computeIfAbsent(lockFile, key -> new ReentrantLock()).lockInterruptibly();
```
chanseokoh: Let's add a comment like "This first lock is to prevent multiple threads from calling FileChannel.lock(), which would otherwise throw OverlappingFileLockException."

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
chanseokoh: FTR: I think I was wrong: https://github.com/GoogleContainerTools/jib/pull/1696#discussion_r281347986

PR #1669, opened 2019-04-25
Test that images are reproducible

briandealwis:
Adds simple test for Jib Core to ensure that created images have consistent file timestamps, permissions, no implicit parent directories, and consistent ordering.  Verified this catches the directory timestamp issue in #1649, and fails if the sort is removed from `ReproducibleLayerBuilder`.

Aside: this would have been easier if we could create `LayerEntry`s without needing explicit file content.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
@@ -39,8 +39,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.function.BiConsumer;
 import java.util.zip.GZIPInputStream;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
@@ -56,7 +54,6 @@ import org.junit.rules.TemporaryFolder;
  * file orderings.
  */
 public class ReproducibleImageTest {
-  private static String LAYERS_PATH_IN_RESOURCES = "layers/";
 
   @ClassRule public static final TemporaryFolder imageLocation = new TemporaryFolder();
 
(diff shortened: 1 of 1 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
```diff
@@ -0,0 +1,194 @@
...
+  public static void createImage()
+      throws InvalidImageReferenceException, InterruptedException, CacheDirectoryCreationException,
+          IOException, RegistryException, ExecutionException {
+
+    Path root = imageLocation.getRoot().toPath();
+    Path fileA = Files.createFile(root.resolve("fileA.txt"));
+    Path fileB = Files.createFile(root.resolve("fileB.txt"));
+    Path fileC = Files.createFile(root.resolve("fileC.txt"));
+    Path subdir = Files.createDirectory(root.resolve("dir"));
+    Path fileD = Files.createFile(subdir.resolve("fileD.txt"));
```
chanseokoh: nit: `fileD` gives an unused var warning. Just don't assign.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+    imageTar = new File(imageLocation.getRoot(), "image.tar");
+    Containerizer containerizer =
+        Containerizer.to(TarImage.named("jib-core/reproducible").saveTo(imageTar.toPath()));
+
+    ExecutorService executorService = Executors.newCachedThreadPool();
+    containerizer.setExecutorService(executorService);
```
chanseokoh: Do we have to set this? Jib by default will use `Executors.newCachedThreadPool()` and shut it down.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
```diff
@@ -0,0 +1,194 @@
...
+        });
+    Assert.assertEquals(
+        ImmutableSet.of("app/fileA.txt", "app/fileB.txt", "app/fileC.txt", "app/fileD.txt"), paths);
+  }
+
+  @Test
+  public void testTimestampsEpochPlus1s() throws IOException {
+    layerEntriesDo(
+        (layerName, layerEntry) -> {
+          if (layerEntry.isFile() || layerEntry.isDirectory()) {
```
chanseokoh: nit: this test is unnecessary. In our simple example, every entry is either a file or a directory.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
briandealwis: Test copying.  You're right (I hope)

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
briandealwis: That is true, but it worth checking this.  I'll pull the out to a separate test.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
briandealwis: Good point.  And looking at this more, I'll expand this to include a subdirectory to ensure we reproduce the directory structure.

loosebazooka:
My concern is that people might forget to update this and this test stops covering everything? Are these not tests we can include directly in other tests classes like `LayerEntry` ?

briandealwis:
This verifies the actual created image rather than the in-memory representations.

loosebazooka:
I realize maybe this isn't possible but could we associate this test with something concrete in src/main/java? Either `Containerizer` or `ReproducibleLayerBuiler` or anything? If that's really not possible, then LGTM.

briandealwis:
Personally I see that as an argument _for_ this class: it's end-to-end.  If we change / rewrite `ReproducibleLayerBuilder` then there's a chance we miss something when porting its tests to the replacement.  Maybe this test should be more specific and embed the calculated SHAs and file contents.

PR #1670, opened 2019-04-25
 Remove unnecessary cache interfaces

### LINKED ISSUE #1668: Considering removing CacheStorage interface
I believe this was originally in place to allow jib-core users to create their own caching implementations. However, at the moment, `CacheStorage` and `DefaultCacheStorage` are just an extra layer between the frontend `Cache` class and the logic underneath, that doesn't really do anything.

We can keep it if we do decide to expose this in the jib-core API and make the cache customizable, but it also might be nice to simplify and remove this extra layer.

TadCordle:
Review/merge after #1666. Fixes #1668.

* Removes `CacheStorage` and `DefaultCacheStorage` and moves logic from `DefaultCacheStorage` to `Cache`
* Replaces `CachedLayer` interface with its implementation
* Renames `DefaultCacheX` classes to `CacheX`
* Removes `UncompressedCacheWrite`
* Renames each `CacheStorageWriter#write` method to respective `writeCompressed` and `writeUncompressed` methods


jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
```diff
@@ -153,14 +154,14 @@ private static void writeMetadata(JsonTemplate jsonTemplate, Path destination)
    * @return the {@link CachedLayer} representing the written entry
    * @throws IOException if an I/O exception occurs
    */
-  CachedLayer write(Blob compressedLayerBlob) throws IOException {
+  CachedLayer writeCompressed(Blob compressedLayerBlob) throws IOException {
```
loosebazooka: Whats the difference between `writeCompressed` and `writeUncompressed`, given that you're just writing down whatever the method is supplied with?

TadCordle: I removed the `UncompressedCacheWrite` class, so now each `write` method takes a `Blob` as the first parameter which makes it slightly more ambiguous whether it needs to be compressed or uncompressed. The parameter name does indicate which it should be, but I figured it'd be more clear to put it in the method name.

loosebazooka: I guess my question is, are they doing the exact same thing. Is `writeCompressed` actually compressing the blob? Or is it just writing down whatever it is given? Could I potentially send an uncompressed blob to `writeCompressed`?

TadCordle: `writeUncompressed` takes an uncompressed layer and compresses it, and `writeCompressed` uncompresses the layer temporarily at some point to get the diff ID. So it doesn't just write what it's given.

loosebazooka: Ah okay, naming is confusing me. But that's an issue for another day.

chanseokoh: Yeah, the naming is confusing. From my perspective, there are things in our codebase that may naturally be understood as a general object that could have been written without strong ties with Docker concepts. I think one day I will sort this and clean up this.

PR #1671, opened 2019-04-25
Deprecate jib.extraDirectory in Gradle

chanseokoh:
Good for review, but I haven't done a lot of manual testing, so I'll give a second pass myself too.


chanseokoh:
Did a lot of manual testing and fixed a few things. Really ready for review.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
```diff
@@ -67,16 +71,22 @@ public ExtraDirectoryParameters(Project project) {
...
    * Sets paths. {@code paths} can be any suitable object describing file paths convertible by
-   * {@link Project#files} (such as {@code List<File>}).
+   * {@link Project#files} (such as {@link File}, {@code List<File>}, or {@code List<String>}).
    *
    * @param paths paths to set.
    */
-  // non-plural to retain backward-compatibility for the "jib.extraDirectory.path" config parameter
-  public void setPath(Object paths) {
+  public void setPaths(Object paths) {
+    jibExtension.extraDirectoriesConfigured = true;
```
loosebazooka: I'm curious how necessary this is, would prefer we didn't need to keep track of things with flags?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -97,19 +100,34 @@ public void container(Action<? super ContainerParameters> action) {
...
-   * paths convertible by {@link Project#files} (such as {@code List<File>}).
+   * paths convertible by {@link Project#files} (such as {@link File}, {@code List<File>} or {@code
+   * List<String>}).
    *
    * @param extraDirectories paths to set.
    */
-  // non-plural to retain backward-compatibility for the "jib.extraDirectory" config parameter
-  public void setExtraDirectory(Object extraDirectories) {
-    this.extraDirectory.setPath(extraDirectories);
+  public void setExtraDirectories(Object extraDirectories) {
```
loosebazooka: Wouldn't we want to avoid adding getters and setters for extraDirectories? I thought in maven we're pushing everyone to use the complex type configuration for `<extraDirectories><paths.....` I would image we would want to do that here as well... and just not allow `jib.extraDirectories = asdf`

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-complex-properties.xml
```diff
@@ -27,8 +27,8 @@
     <jib.container.labels>key1=value1,key2=value2</jib.container.labels>
     <jib.container.volumes>/var/log,/var/log2</jib.container.volumes>
     <jib.container.format>Docker</jib.container.format>
-    <jib.extraDirectory.path>${project.basedir}/src/main/jib-custom</jib.extraDirectory.path>
-    <jib.extraDirectory.permissions>/foo=755,/bar/cat=777</jib.extraDirectory.permissions>
+    <jib.extraDirectories.paths>${project.basedir}/src/main/jib-custom</jib.extraDirectories.paths>
```
loosebazooka: Is this part of this PR?

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-extra-dirs.gradle
```diff
@@ -15,4 +15,4 @@ dependencies {
 }
 
 jib.to.image = System.getProperty("_TARGET_IMAGE")
-jib.extraDirectory = ['src/main/custom-extra-dir', 'src/main/custom-extra-dir2']
+jib.extraDirectories = ['src/main/custom-extra-dir', 'src/main/custom-extra-dir2']
```
loosebazooka: Do we need to have tests that use jib.extraDirectory directly in the build.gradle?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
chanseokoh: Sounds good with being consistent. We allowed the shortcut to remain backward compatibility after introducing `.permissions`, but we can clean that up since we are making breaking changes with deprecation.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
chanseokoh: Yeah, this is quite unfortunate. The Gradle case wasn't easy for a few reasons.

1. I couldn't find a way (if at all) to log inside `JibExtension` or `ExtraDirectoriesParameters`.
1. The way we can extract `extraDirectories.paths` or `extraDirectory.path` in our codebase for consumption is to first go through `getExtraDirectories()`, which is also used by Gradle to set in config values, so unlike Maven, it's not possible to check if the user is using the old or new syntax (or both) inside `getExtraDirectories()`. On top of that, Gradle always calls both `getExtraDirectory()` and `getExtraDirectories()` regardless of whether the user has ever configured anything.

So, although this is very unfortunate, I couldn't find a good alternative.

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-complex-properties.xml
chanseokoh: Not really, but something I missed to make test projects up to date with the new syntax.

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-extra-dirs.gradle
chanseokoh: I'll add one.

chanseokoh:
Good for another look again.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -232,6 +241,45 @@ public void testDockerDaemon_simpleWithIncompatibleJava11()
     }
   }
 
+  @Test
+  public void testDockerDaemon_simple_deprecatedExtraDirectory()
+      throws DigestException, IOException, InterruptedException {
+    String targetImage = "localhost:6000/simpleimage:gradle" + System.nanoTime();
```
TadCordle: Seems like you could reduce some code duplication between these 3 tests if you move all of this into `assertExtraDirectoryDeprecation` and take the build.gradle name as a parameter?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -67,14 +68,32 @@ static void disableHttpLogging() {
...
+      logger.warn(
+          "'jib.extraDirectory', 'jib.extraDirectory.path', and 'jib.extraDirectory.permissions' "
+              + "are deprecated; use 'jib.extraDirectories', 'jib.extraDirectories.paths' and "
+              + "'jib.extraDirectories.permissions'");
+
+      if (jibExtension.extraDirectoriesConfigured
+          || System.getProperty(PropertyNames.EXTRA_DIRECTORIES_PATHS) != null
+          || System.getProperty(PropertyNames.EXTRA_DIRECTORIES_PERMISSIONS) != null) {
+        throw new IllegalArgumentException(
+            "You cannot configure both 'jib.extraDirectory.path' and 'jib.extraDirectories.paths'");
```
TadCordle: This message about `jib.extraDirectory.path`/`jib.extraDirectories.paths` will show even if you're only using `jib.extraDirectory`/`jib.extraDirectories`, I think. Might be better to just say "You cannot configure both 'jib.extraDirectory' and 'jib.extraDirectories'".

chanseokoh: Good point. I considered it, and there's a reason that I ended up using `.path(s)`.

There were some limitations, and I couldn't find a way whether a user configured `jib.extraDirectory` or `jib.extraDirectory.path` (you can get the sense in https://github.com/GoogleContainerTools/jib/pull/1671#discussion_r280123597). If the user used `jib.extraDirectory.(path)` and saw the message without `.path(s)`, the user may just try change "Directory" to "Directories" and run into another problem, because we won't support the shortcut `jib.extraDirectories` anymore and `jib.extraDirectories.path` won't work either (`path` should be pluralized). So I think for less friction in migration, I thought it'd be better to mention the `path(s)` part than not mentioning it.

TadCordle: Oh, I didn't realize we weren't support the `jib.extraDirectories` shortcut anymore. In that case this makes sense.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -129,6 +129,22 @@ private static void assertDockerInspect(String imageReference)
...
+    BuildResult buildResult =
+        JibRunHelper.buildToDockerDaemon(simpleTestProject, targetImage, pomXml);
+    Assert.assertEquals(
+        "Hello, world. \nrw-r--r--\nrw-r--r--\nfoo\ncat\n",
+        new Command("docker", "run", "--rm", targetImage).run());
+    Assert.assertThat(
+        buildResult.getOutput(),
+        CoreMatchers.containsString(
+            "'jib.extraDirectory', 'jib.extraDirectory.path', and 'jib.extraDirectory.permissions' "
+                + "are deprecated; use 'jib.extraDirectories', 'jib.extraDirectories.paths' and "
```
TadCordle: If we aren't supporting `jib.extraDirectories` by itself, maybe we should remove that from this message (or rephrase to be clear about the configuration block `jib.extraDirectories { ... }`?).

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -67,14 +68,32 @@ static void disableHttpLogging() {
...
   }
 
+  @Deprecated
+  static void checkDeprecatedUsage(JibExtension jibExtension, Logger logger) {
+    if (jibExtension.extraDirectoryConfigured
+        || System.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH) != null
+        || System.getProperty(PropertyNames.EXTRA_DIRECTORY_PERMISSIONS) != null) {
+      logger.warn(
+          "'jib.extraDirectory', 'jib.extraDirectory.path', and 'jib.extraDirectory.permissions' "
+              + "are deprecated; use 'jib.extraDirectories', 'jib.extraDirectories.paths' and "
```
TadCordle: Same here.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
chanseokoh: Ah, good catch!

PR #1676, opened 2019-04-29
 Add cache methods for retrieving manifest and container config

TadCordle:
Towards #718. Review after #1670.

Adds `Cache#retrieveManifest()` and `Cache#retrieveContainerConfiguration` for retrieving the cached manifest/config.


jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
```diff
@@ -62,6 +72,73 @@
...
+  Optional<ManifestTemplate> retrieveManifest(ImageReference imageReference) throws IOException {
+    Path imageDirectory = cacheStorageFiles.getImageDirectory(imageReference);
+    Path manifestPath = imageDirectory.resolve("manifest.json");
+    if (!Files.exists(manifestPath)) {
+      return Optional.empty();
+    }
+
+    try (LockFile ignored = LockFile.lock(imageDirectory.resolve("lock"))) {
+      // TODO: Consolidate with ManifestPuller
+      ObjectNode node =
```
loosebazooka: I feel like we have some code like this somewhere? How easy is it to share that logic? In `PullBaseImageStep` or something?

TadCordle: There's very similar code in `ManifestPuller`, and I wasn't immediately sure how to share it, hence the todo. I wonder if we could stick it in `JsonTemplateMapper`, or even a static method in `ManifestTemplate` or something?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
```diff
@@ -115,6 +116,30 @@ public CachedLayer writeUncompressedLayer(
...
+  }
+
+  /**
+   * Retrieves the cached container configuration for an image reference.
+   *
+   * @param imageReference the image reference
+   * @return the container configuration for the image reference, if found
+   * @throws IOException if an I/O exception occurs
+   */
+  public Optional<ContainerConfigurationTemplate> retrieveContainerConfiguration(
```
chanseokoh: For Manifest V2.2 (and OCI too I believe), it should return both the manifest and the container config at the same time in a single transaction due to concurency.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
```diff
@@ -62,6 +72,73 @@
...
+          new ObjectMapper().readValue(Files.newInputStream(manifestPath), ObjectNode.class);
+      if (!node.has("schemaVersion")) {
```
chanseokoh: I think in these cases, it means cache corruption, right? So I think we should quit. I am fine with crashing with a runtime exception. Maybe with a message to report the bug that Jib corrupted the cache and a way to clear the cache?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
TadCordle: What do you mean by this? Return them both from the same method/using the same lock?

chanseokoh: Yeah, think about this, where Process A wants to get V2.2 manifest and its container config.

1. Process A calls `retrieveManifst()`.
1. Process B calls `writeMetadata(differentManifest, differentContainerConfig)`.
1. Process A calls `retrieveContainerConfig()`.

Then the container config at step 3 is for the new manifest from step 2, which may be out of sync with the manifest 1 in step 1.

PR #1677, opened 2019-04-30
Reorg and refactor code around Blob abuse

chanseokoh:
Continuation of #1647. In #1647, I said the prevalent abuse of `Blob` (which is not much different from an all-consuming `Object` in many contexts) is to wrap anything in a `Blob` only for the sake of writing it to an output stream or compute a digest, which is an unnecessary and convoluted round-trip and making the target invisible underneath `Blob`.

Another complexity that hurt readability (to me at least) was the similarities between `WriterBlob`, `BlobWriter`, and `Blob`. `Blob` and `WriterBlob` look very similar,
```java
public interface Blob {
}

public interface BlobWriter {
  void writeTo(OutputStream outputStream) throws IOException;
}
```
but yet, you create a `Blob` (which is actually `WriterBlob`) out of `BlobWriter`
```java
Blob writerBlob = Blobs.from(blobWriter);
```
and `WriterBlob.writeTo()` makes use of `blobWriter.writeTo()`. Before someone becomes familiarized with the code, all this will sound pretty confusing to them. With the `Blob` abuse I said at the top, this is aggravating.

Some highlighted changes:

- Instead of going "something --> Blob.from(something) --> blob.writeTo()" to compute a digest or write to an output stream, I introduced `DigestUtil` to do it directly without the abuse.
- Decoupled `Blob` from `JsonTemplateMapper`. Now JSON-related classes sit below `Blob` in the dependency graph.
- Added `JsonBlob`. This makes sense because `Blob` sits above `JsonTemplate`.
- Renamed `BlobWriter` to `WritableContents`, and moved it to the `hash` package that sits below `Blob`.
- All `Blob.writeTo()` methods simply and consistently delegate to `DigestUtil.computeDigest()`. This resulted in removing `BlobDescriptor.fromPipe()`, which I think was not named or placed nicely, hose code is also duplicated some `Blob.writeTo()` method.
- `BlobDescriptor` is merely a holder of (size, digest).


chanseokoh:
Another thing to note is that, there's no `toBlob()` functions anymore after #1647 and this. All blobs are constructed through `Blobs.from()`.

chanseokoh:
For some unknown reason, compilation always fails on Kokoro MacOS with lots of `NoClassDefFoundError` and `ClassNotFoundException`. So weird.

loosebazooka:
It also looks like installing the plugin for integration tests is failing too?

chanseokoh:
Let me close this and reopen another one.

PR #1684, opened 2019-05-01
Reorg and refactor repo - continued

chanseokoh:
For Blob changes, following the same spirit in #1647 and #1678.

Other changes involve removing various classes around progress reporting, decoupling functionalities to have small, fine-grained general components, etc.


jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -27,10 +29,37 @@
  */
 public class Authorization {
 
+  /**
```
chanseokoh: note to reviewers: these are just moved from `Authorizations`.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/DelayedConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.function.BinaryOperator;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+/**
```
chanseokoh: note to reviewers: this "delay" functionality (and the code) is factored out from `ListenableCountingOutputStream`.

chanseokoh:
Ready for review.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -246,14 +245,15 @@ public BlobDescriptor checkBlob(DescriptorDigest blobDigest)
...
    *
    * @param blobDigest the digest of the BLOB to download
-   * @param blobSizeConsumer callback to receive the total size of the BLOb to pull
-   * @param blobProgressListener listener for progress of the pull
+   * @param blobSizeListener callback to receive the total size of the BLOb to pull
+   * @param writtenByteCountListener listens on byte count written to an output stream during the
+   *     pull
    * @return a {@link Blob}
    */
   public Blob pullBlob(
```
TadCordle: Maybe it could be simpler if these blob pull/push clsases took a single `ProgressEventDispatcherWrapper` with the set/dispatch methods instead of two consumers? I'm not sure how it fits into the class hierarchy though.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ProgressEventDispatcherWrapper.java
```diff
@@ -65,8 +57,9 @@ public void close() {
     progressEventDispatcher.close();
   }
 
-  void initializeWithBlobSize(long blobSize) {
+  void setProgressTarget(long allocationUnits) {
```
TadCordle: Seems a bit weird to have to call the constructor and this method to fully initialize a `ProgressEventDispatcherWrapper`, maybe we can consolidate this by changing the signature of `RegistryClient#pullBlob`?

chanseokoh: Yeah, I did not like this wrapper either. The problem Q originally faced was the progress reporting design that

1. You have to know and set in advance how much progress units you will allocate when creating a progress event dispatcher.
1. But you cannot know how much bytes (allocation units) you will download until you read `Content-Length`.
1. So the wrapper that will create and hold a progress event dispatcher when this method is called. Other than that, this class is just a thin wrapper that delegates `dispatchProgress()`.

For this reason, obviously it's best to get rid of this unfortunate wrapper if possible. As an attempt to work around the issue, Q probably also felt it doesn't make much sense to have this wrapper in the signature and make it a hard dependency from the API standpoint, also given that there are only just a couple places this wrapper is required. And as I said in the earlier comment, my intention is to actually decouple progress dispatcher stuff from `RegistryClient`.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/DelayedConsumer.java
briandealwis: [Might be better called](https://blog.bitsrc.io/understanding-throttling-and-debouncing-973131c1ba07) "ThrottlingConsumer", but it's really a `ThrottlingDispatcher`? 

The _adder_ might be better called _reducer_, as in map-reduce.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/DelayedConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+
+    previousCallback = getNow.get();
+  }
+
+  @Override
+  public void accept(T value) {
+    valueSoFar = valueSoFar == null ? value : valueAdder.apply(valueSoFar, value);
+
+    Instant now = getNow.get();
+    if (previousCallback.plus(delayBetweenCallbacks).isBefore(now)) {
```
briandealwis: I've looked over this formula several times and it feels hard to read!  Maybe extract `previousCallback.plus(delayBetweenCallbacks)` to `nextFireTime`, so it reads more like
```
if(now.isAfter(nextFireTime)) {
```
?

jib-core/src/main/java/com/google/cloud/tools/jib/hash/CountingDigestOutputStream.java
```diff
@@ -68,31 +65,16 @@ public void computeDigest() {
...
       bytesSoFar = 0;
+      return blobDescriptor;
 
     } catch (DigestException ex) {
       throw new RuntimeException("SHA-256 algorithm produced invalid hash: " + ex.getMessage(), ex);
     }
   }
 
-  /** @return the number of bytes written and used to compute the most recent digest */
-  public long getBytesHahsed() {
```
briandealwis: I wonder how long this misspelled method has been around :-)

jib-core/src/main/java/com/google/cloud/tools/jib/hash/CountingDigestOutputStream.java
```diff
@@ -54,10 +48,13 @@ public CountingDigestOutputStream(OutputStream outputStream) {
   }
 
   /**
-   * Computes the hash and remembers the size of the bytes written to compute the hash. The buffer
-   * resets after this method is called, so this method should only be called once per computation.
+   * Computes the hash and returns it along with the size of the bytes written to compute the hash.
+   * The buffer resets after this method is called, so this method should only be called once per
+   * computation.
```
briandealwis: I wonder how many places do we call this method multiple times?

jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
```diff
@@ -54,14 +58,10 @@ public boolean retrySupported() {
...
-    outputStream =
-        blobProgressListener == null
-            ? outputStream
-            : new ListenableCountingOutputStream(
-                outputStream,
-                blobProgressListener::handleByteCount,
-                blobProgressListener.getDelayBetweenCallbacks());
-    blob.writeTo(outputStream);
-    outputStream.flush();
+    try (DelayedConsumer<Long> delayedCountListener =
```
briandealwis: This use of the DelayedConsumer seems like a decision that should be made by the creator of this class.

chanseokoh: That's a good point.

jib-core/src/main/java/com/google/cloud/tools/jib/hash/CountingDigestOutputStream.java
chanseokoh: I'm guessing none, because the last time I touched this class, it was reporting wrong bytes when called more than once. (Basically, `bytesSoFar` was not reset.)

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/DelayedConsumer.java
chanseokoh: `ThrottlingConsumer` sounds good. As a general class for any `Consumer`, `Dispatcher` doesn't sound naturally to me.

About "reducer", I think it adversely adds confusion, since "reducer" usually means the "foldl" (or `Stream.reduce()`, `std::accumulate()`) that acts on a list while taking this binary operator, to "fold" or "reduce" them into a single element.

chanseokoh:
I've addressed the comments. This should be good now.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/ThrottledConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.function.BinaryOperator;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+/**
+ * Wraps a {@link Consumer} so that multiple consume calls ({@link #accept}) within a short period
+ * of time are merged into a single later call.
+ */
+public class ThrottledConsumer<T> implements Consumer<T>, Closeable {
```
loosebazooka: Can you kind of explain why this is needed? Are we doing something unavoidable somewhere that requires this? Maybe I missed the thread where this was already discussed.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -27,10 +29,37 @@
...
+   * @param token the token
+   * @return an {@link Authorization} with a {@code Bearer} token
+   */
+  public static Authorization withBearerToken(String token) {
```
loosebazooka: I think these can be in the style of `fromX` instead of `withX` if we're moving them from an `Authorizations` class.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/ThrottledConsumer.java
chanseokoh: Currently, this is only used for throttling progress event dispatch. `ListenableCountingOutputStream` had the throttling capability integrated for listening, and in this PR, I factored out the throttling functionality into this independent class. I am pretty sure throttling is not strictly necessary, but perhaps Q wanted to avoid dispatching too many progress update in a very short time span. For example, if debug logging is enabled, we will see too many progress events, as I guess `ListenableCountingOutputStream` may call `write()` insanely many times for a large blob push/pull.

chanseokoh: (I had added a note to this line in the original commit that explained this is factored out from another place, but I see the note got removed when I renamed this class.)

loosebazooka: What do you think about a minimal class (without the generics, injected adders, etc) that just handles the progress event consumption? This class seems a little over-engineered for what we want it to do?

chanseokoh: Yeah, I think we can specialize it. No need to be overly generic.

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/ThrottledLongConsumer.java
```diff
@@ -0,0 +1,82 @@
...
+import java.time.Instant;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import javax.annotation.Nullable;
+
+/**
+ * Wraps a {@code Consumer<Long>} so that multiple consume calls ({@link #accept}) within a short
+ * period of time are merged into a single later call with the value accumulated up to that point.
+ */
+public class ThrottledLongConsumer implements Consumer<Long>, Closeable {
```
briandealwis: ThrottledAccumulator or ThrottledAccumulatingConsumer?

jib-core/src/main/java/com/google/cloud/tools/jib/hash/Digests.java
```diff
@@ -45,6 +45,10 @@ public static DescriptorDigest computeJsonDigest(List<? extends JsonTemplate> te
     return computeDigest(contents, ByteStreams.nullOutputStream()).getDigest();
   }
 
+  public static BlobDescriptor computeDigest(JsonTemplate template) throws IOException {
+    return computeDigest(template, ByteStreams.nullOutputStream());
+  }
+
   public static BlobDescriptor computeDigest(JsonTemplate template, OutputStream optionalOutStream)
```
briandealwis: This `optionalOutStream` seems an odd name choice… since it's not optional.

jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
```diff
@@ -16,112 +16,71 @@
...
-import java.time.Instant;
 import java.util.function.Consumer;
-import java.util.function.Supplier;
 
-/**
- * Counts the number of bytes written and reports the count to a callback. The count is reported
- * with certain time delays to avoid calling the callback too often.
- */
+/** Counts the number of bytes written and reports the count to a callback. */
 public class ListenableCountingOutputStream extends OutputStream {
```
briandealwis: Ugh this name.  `NotifyingOutputStream`?  Or even `CountingOutputStream`?

jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/ThrottledLongConsumer.java
chanseokoh: Going with `ThrottledAccumulatingConsumer`, as I want it to show that it's fundamentally a `Consumer`.

jib-core/src/main/java/com/google/cloud/tools/jib/hash/Digests.java
chanseokoh: No more "optional"s.

jib-core/src/main/java/com/google/cloud/tools/jib/http/ListenableCountingOutputStream.java
chanseokoh: Guava has `CountingOutputStream` (that just does byte counting, no listening), so going with `NotifyingOutputStream`.

jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
```diff
@@ -20,20 +20,23 @@
...
-  @Nullable private final BlobProgressListener blobProgressListener;
+  private final Consumer<Long> writtenByteCountListener;
 
-  public BlobHttpContent(
-      Blob blob, String contentType, @Nullable BlobProgressListener blobProgressListener) {
+  public BlobHttpContent(Blob blob, String contentType) {
+    this(blob, contentType, ignored -> {});
+  }
+
+  public BlobHttpContent(Blob blob, String contentType, Consumer<Long> writtenByteCountListener) {
```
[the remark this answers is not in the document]
chanseokoh: Makes sense. Not immediately clear how I can change this though.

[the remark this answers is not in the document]
loosebazooka: Should it not contain a writeTo? Or should writeTo take the listener?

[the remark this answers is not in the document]
chanseokoh: The class is implementing `HttpContent::writeTo`, so it's not like we can simply change the signature at this level. Perhaps next time.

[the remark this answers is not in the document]
loosebazooka: Ah interesting, didn't know HttpContent was something outside of our codebase.

PR #1686, opened 2019-05-02
Return absolute paths to extra directories

### LINKED ISSUE #1660: FilesMojoV2 should resolve and use absolute paths for extra directories
`FilesMojo` resolves/prints the absolute paths of the extra directories, but `FilesMojoV2` does not. Is this intended?

loosebazooka:
Fixes #1660.

~~I still need to update the tests.~~

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
@@ -169,6 +169,13 @@ public class FilesMojoV2 extends AbstractMojo {
   }
 
   private List<Path> resolveExtraDirectories(MavenProject project) throws MojoExecutionException {
+    return collectExtraDirectories(project)
+        .stream()
+        .map(path -> path.isAbsolute() ? path : project.getBasedir().toPath().resolve(path))
+        .collect(Collectors.toList());
+  }
+
+  private List<Path> collectExtraDirectories(MavenProject project) throws MojoExecutionException {
     // Try getting extra directory from project/session properties
     String deprecatedProperty =
         MavenProjectProperties.getProperty(PropertyNames.EXTRA_DIRECTORY_PATH, project, session);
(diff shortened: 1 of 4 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
```diff
@@ -183,7 +183,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     String property = newProperty != null ? newProperty : deprecatedProperty;
     if (property != null) {
       List<String> paths = ConfigurationPropertyValidator.parseListProperty(property);
-      return paths.stream().map(Paths::get).collect(Collectors.toList());
+      return paths.stream().map(Paths::get).map(Path::toAbsolutePath).collect(Collectors.toList());
```
chanseokoh: I think this is not enough. For example, users can currently set the extra path to a project-relative path (like `src/main/my-custom`), and this will just become `/src/main/my-custom`. I *think* what should be done is to resolve this against the project root in case it is a relative path. @briandealwis is that right?

loosebazooka: Yeah I just realized the problem as I was updating tests. We have to check if they're absolute, and otherwise resolve against the current module root.

loosebazooka:
Ugh, gotta fix filesMojo

jib-maven-plugin/src/test/resources/maven/projects/multi/complex-service/pom.xml
```diff
@@ -18,7 +18,13 @@
...
         <artifactId>jib-maven-plugin</artifactId>
         <configuration>
-          <extraDirectory>${project.basedir}/src/main/other-jib</extraDirectory>
+          <extraDirectories>
+            <paths>
+              <path>src/main/jib1</path> <!-- should resolve to this projectRoot -->
+              <path>${project.basedir}/src/main/jib2</path> <!-- provided as absolute within project -->
+              <path>/some/random/absolute/path/jib3</path> <!-- provided as absolute outside of project -->
+            </paths>
+            </extraDirectories>
```
TadCordle: nit: extra indentation

PR #1687, opened 2019-05-02
Add offline mode to jib-core

### LINKED ISSUE #1683: OCI base image support is likely to be fishy or even not working
We does accept OCI manifest and read it into `OCIManifestTemplate` in memory.

* `ManifestPuller`
```java
    if (schemaVersion == 2) {
      // 'schemaVersion' of 2 can be either Docker V2.2 or OCI.
      String mediaType = node.get("mediaType").asText();
      ...
      if (OCIManifestTemplate.MANIFEST_MEDIA_TYPE.equals(mediaType)) {
        return manifestTemplateClass.cast(
            JsonTemplateMapper.readJson(jsonString, OCIManifestTemplate.class));
      }
```

`readJson()` above returns an `OCIManifestTemplate` object (which is cast into the super `ManifestTemplate` by `.cast()`).

But, when we re-cast the in-memory object, we case it into either `V21ManifestTemplate` or `V22ManifestTemplate`.

* `PullBaseImageStep`
```java
    ManifestTemplate manifestTemplate =
        registryClient.pullManifest(buildConfiguration.getBaseImageConfiguration().getImageTag());

    // TODO: Make schema version be enum.
    switch (manifestTemplate.getSchemaVersion()) {
      case 1:
        V21ManifestTemplate v21ManifestTemplate = (V21ManifestTemplate) manifestTemplate;
        ...

      case 2:
        V22ManifestTemplate v22ManifestTemplate = (V22ManifestTemplate) manifestTemplate;
        ...
```
So although I haven't tested myself, I think Jib will throw class cast exception.

Also, we should have a "default" case in the above switch.

TadCordle:
Towards #718. Also fixes #1683.

Adds a `Containerizer#setOfflineMode` to set whether or not to set offline mode. Errors out if attempting to build to a registry, or if there is missing information from the cache.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
@@ -118,9 +118,9 @@ public class JibIntegrationTest {
   public void testOffline()
       throws IOException, InterruptedException, InvalidImageReferenceException, ExecutionException,
           RegistryException, CacheDirectoryCreationException {
-    LocalRegistry offlineRegistry = new LocalRegistry(5001);
-    offlineRegistry.start();
-    offlineRegistry.pullAndPushToLocal("busybox", "busybox");
+    LocalRegistry tempRegistry = new LocalRegistry(5001);
+    tempRegistry.start();
+    tempRegistry.pullAndPushToLocal("busybox", "busybox");
     Path cacheDirectory = cacheFolder.getRoot().toPath();
 
     ImageReference targetImageReferenceOnline =
@@ -148,7 +148,6 @@ public class JibIntegrationTest {
               .setOfflineMode(true));
       Assert.fail();
     } catch (ExecutionException ex) {
-      Assert.assertTrue(ex.getCause() instanceof IOException);
       Assert.assertEquals(
           "Cannot run Jib in offline mode; localhost:5001/busybox not found in cache",
           ex.getCause().getMessage());
@@ -161,7 +160,7 @@ public class JibIntegrationTest {
             .setAllowInsecureRegistries(true));
 
     // Run again in offline mode, should succeed this time
-    offlineRegistry.stop();
+    tempRegistry.stop();
     jibContainerBuilder.containerize(
         Containerizer.to(DockerDaemonImage.named(targetImageReferenceOffline))
             .setBaseImageLayersCache(cacheDirectory)
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
@@ -119,7 +119,7 @@ public class JibIntegrationTest {
       throws IOException, InterruptedException, InvalidImageReferenceException, ExecutionException,
           RegistryException, CacheDirectoryCreationException {
     LocalRegistry tempRegistry = new LocalRegistry(5001);
-    tempRegistry.start();
+    tempRegistry.startLocalRegistry();
     tempRegistry.pullAndPushToLocal("busybox", "busybox");
     Path cacheDirectory = cacheFolder.getRoot().toPath();
 
@@ -149,7 +149,7 @@ public class JibIntegrationTest {
       Assert.fail();
     } catch (ExecutionException ex) {
       Assert.assertEquals(
-          "Cannot run Jib in offline mode; localhost:5001/busybox not found in cache",
+          "Cannot run Jib in offline mode; localhost:5001/busybox not found in local Jib cache",
           ex.getCause().getMessage());
     }
 
@@ -160,7 +160,7 @@ public class JibIntegrationTest {
             .setAllowInsecureRegistries(true));
 
     // Run again in offline mode, should succeed this time
-    tempRegistry.stop();
+    tempRegistry.stopLocalRegistry();
     jibContainerBuilder.containerize(
         Containerizer.to(DockerDaemonImage.named(targetImageReferenceOffline))
             .setBaseImageLayersCache(cacheDirectory)
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
@@ -24,6 +24,7 @@ import com.google.cloud.tools.jib.image.InvalidImageReferenceException;
 import com.google.cloud.tools.jib.registry.LocalRegistry;
 import com.google.cloud.tools.jib.registry.RegistryException;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -32,13 +33,17 @@ import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.ClassRule;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
 
 /** Integration tests for {@link Jib}. */
 public class JibIntegrationTest {
 
   @ClassRule
   public static final LocalRegistry localRegistry = new LocalRegistry(5000, "username", "password");
+
+  @Rule public final TemporaryFolder cacheFolder = new TemporaryFolder();
 
   /**
    * Pulls a built image and attempts to run it.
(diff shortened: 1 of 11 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -52,6 +52,16 @@ public LocalRegistry(int port, String username, String password) {
   /** Starts the local registry. */
   @Override
   protected void before() throws IOException, InterruptedException {
+    start();
```
chanseokoh: +1 for readability. How about `startLocalRegistry()`?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -52,6 +52,16 @@ public LocalRegistry(int port, String username, String password) {
...
+  }
+
+  @Override
+  protected void after() {
+    stop();
```
chanseokoh: `stopLocalRegistry`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -235,6 +237,18 @@ public Containerizer setAllowInsecureRegistries(boolean allowInsecureRegistries)
...
   }
 
+  /**
+   * Sets whether or not to run the build in offline mode. In offline mode, the base image is
+   * retrieved from the cache instead of pulled from a registry.
+   *
+   * @param offline if {@code true}, the build will run in offline mode
+   * @return this
+   */
+  public Containerizer setOfflineMode(boolean offline) {
```
chanseokoh: The Javadoc says "In offline mode, the base image is retrieved from the cache instead of pulled from a registry". If we were going to fail `jib:build` with the offline mode, we should prominently document that too.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -283,6 +297,10 @@ boolean getAllowInsecureRegistries() {
     return allowInsecureRegistries;
   }
 
+  boolean getOfflineMode() {
```
chanseokoh: Maybe `isOfflineMode` or `isOffline`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
```diff
@@ -82,6 +82,9 @@ public CachedLayer call() throws IOException, CacheCorruptedException {
       Optional<CachedLayer> optionalCachedLayer = cache.retrieve(layerDigest);
       if (optionalCachedLayer.isPresent()) {
         return optionalCachedLayer.get();
+      } else if (buildConfiguration.getOffline()) {
+        throw new IOException(
+            "Cannot run Jib in offline mode; Cache is missing image layer " + layerDigest);
```
chanseokoh: Chances are pretty low, but I can see this can really happen in practice. For example, Jib downloaded and saved the manifest, but while Jib downloading layers, you press Ctrl+C (haven't tried though), and you happen to run it in the offline mode next. This isn't exactly cache corruption (although in some sense yes). Let's be a little bit more specific, like "... local Jib cache for base images is missing ..." Maybe a bit more helpful to suggest running online once more. The error message is relevant to jib-core users too.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
```diff
@@ -395,6 +411,15 @@ public boolean getAllowInsecureRegistries() {
     return allowInsecureRegistries;
   }
 
+  /**
+   * Gets whether or not to run the build in offline mode.
+   *
+   * @return {@code true} if the build will run in offline mode; {@code false} otherwise
+   */
+  public boolean getOffline() {
```
chanseokoh: `isOffline` or `isOfflineMode`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -258,10 +271,48 @@ public BaseImageWithAuthorization call()
...
+   *     format
+   */
+  private Image<Layer> pullBaseImageOffline()
+      throws IOException, CacheCorruptedException, BadContainerConfigurationFormatException,
+          LayerCountMismatchException {
+    ImageReference baseImage = buildConfiguration.getBaseImageConfiguration().getImage();
+    Optional<ManifestAndConfig> metadata =
+        buildConfiguration.getBaseImageLayersCache().retrieveMetadata(baseImage);
+    if (!metadata.isPresent()) {
+      throw new IOException("Cannot run Jib in offline mode; " + baseImage + " not found in cache");
```
chanseokoh: " not found in local Jib cache"

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -239,7 +239,8 @@ public Containerizer setAllowInsecureRegistries(boolean allowInsecureRegistries)
 
   /**
    * Sets whether or not to run the build in offline mode. In offline mode, the base image is
-   * retrieved from the cache instead of pulled from a registry.
+   * retrieved from the cache instead of pulled from a registry, and the build will fail if the base
+   * image is not in the cache.
```
chanseokoh: Javadoc nit: I believe it will always fail when pushing to a registry (when using `RegistryImage`) too, right?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -52,6 +52,16 @@ public LocalRegistry(int port, String username, String password) {
   /** Starts the local registry. */
   @Override
   protected void before() throws IOException, InterruptedException {
+    startLocalRegistry();
```
chanseokoh: Oh, sorry. I thought this was a test class. Since this is `@Rule`, `startLocalRegistry()` sounds redundant.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -258,10 +271,49 @@ public BaseImageWithAuthorization call()
...
+          "Cannot run Jib in offline mode; " + baseImage + " not found in local Jib cache");
+    }
+
+    ManifestTemplate manifestTemplate = metadata.get().getManifest();
+    if (manifestTemplate instanceof V21ManifestTemplate) {
+      return JsonToImageTranslator.toImage((V21ManifestTemplate) manifestTemplate);
+    }
+
+    ContainerConfigurationTemplate configurationTemplate =
+        metadata.get().getConfig().orElseThrow(IllegalStateException::new);
```
chanseokoh: I'm fine with `metadata.get().getConfig().get()`. Not much different from throwing `NoSuchElementException`. This should never happen, unless there is a bug.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -476,6 +476,9 @@ JibContainer containerize(
       Containerizer containerizer, Supplier<ExecutorService> defaultExecutorServiceFactory)
       throws IOException, CacheDirectoryCreationException, InterruptedException, RegistryException,
           ExecutionException {
+    if (containerizer.isOfflineMode() && containerizer.getImageConfiguration().isOnlineImage()) {
```
chanseokoh: I see this is the only place where you want to check `isOnlineImage()` (essentially if it was `Containerizer.to(RegistryImage)`). Since the check is within `Containerizer`, I think we don't have to add `isOnlineImage` to `ImageConfiguration`. Instead, `Containerizer` can have an additional field like `mustBeOnline`, initialized to true only when `RegistryImage`, and you can fail early in `setOfflineMode()`. Do you think this will work?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
TadCordle: Right, I only added this to avoid an IntelliJ warning.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
```diff
@@ -109,6 +114,64 @@ public void testScratch()
         inspectOutput.contains("\"Layers\": ["));
   }
 
+  @Test
+  public void testOffline()
+      throws IOException, InterruptedException, InvalidImageReferenceException, ExecutionException,
+          RegistryException, CacheDirectoryCreationException {
+    LocalRegistry tempRegistry = new LocalRegistry(5001);
```
chanseokoh: `LocalRegistry` is already defined as a `@Rule` in this test class. Can't we use that? Looks like we can. It solves another minor problem that the registry will stop even if the test fails. (It is better to do `tempRegistry.stop()` in `finally` if you don't want to reuse it.)

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -145,16 +146,20 @@ public static Containerizer to(TarImage tarImage) {
   @Nullable private Path applicationLayersCacheDirectory;
   @Nullable private EventHandlers eventHandlers;
   private boolean allowInsecureRegistries = false;
+  private boolean offline = false;
+  private boolean mustBeOnline;
```
chanseokoh: nit: let's be consistent by saying `= false`.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ImageConfiguration.java
```diff
@@ -30,6 +30,7 @@
 
     private ImageReference imageReference;
     private ImmutableList<CredentialRetriever> credentialRetrievers = ImmutableList.of();
+    private boolean isOnlineImage = false;
```
chanseokoh: Stray field.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
TadCordle: The reason is because I want to manually stop it in the middle of the test to make sure everything still works, even if the registry can't be reached.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
TadCordle: Gives a warning for redundant initialization, I guess since it's a parameter in the constructor.

TadCordle: I guess I should make it final then.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
chanseokoh: Ah, okay.

PR #1689, opened 2019-05-02
Add offline mode to plugins

### LINKED ISSUE #718: Jib should respect `—offline`

TadCordle:
Fixes #718. Review after #1687.


loosebazooka:
Can we hold off on this until v1.2.0 is done?

TadCordle:
We can leave it for the next release if you want, but this is the last PR for finishing offline mode (until we do docker cache stuff at least).

loosebazooka:
Eh, I didn't realize the part that is downloading the manifests is already in jib core. Lets just get it in with 1.2.0

TadCordle:
Ah yeah, that was this PR https://github.com/GoogleContainerTools/jib/pull/1687

PR #1690, opened 2019-05-02
No generics for Image<> and ImageLayers<>

chanseokoh:
`Image<T extends Layer>` --> `Image`

`ImageLayers<T extends Layer>` --> `ImageLayers`

Does anyone remember why they were generics, like future extensibility?


loosebazooka:
I think it was initially designed with some advanced extensibility. It's probably okay to remove them if we haven't yet used the templating in any meaningful way yet.

PR #1699, opened 2019-05-07
No maven debug output

chanseokoh:
Too much garbage. I don't remember there was anything useful.


kokoro/presubmit.bat
```diff
@@ -11,7 +11,7 @@ REM FOR /f "tokens=*" %%i IN ('docker ps -aq') DO docker rm -vf %%i
 
 cd jib-core && call gradlew.bat clean build --info --stacktrace && ^
 cd ../jib-plugins-common && call gradlew.bat clean build --info --stacktrace && ^
-cd ../jib-maven-plugin && call mvnw.cmd clean install -B -U -X && ^
```
TadCordle: Will this hide the stack trace if there's an error? Maybe we should replace with `-e`?

Issue #1428, opened 2019-01-17
Finalize initial GA-worthy public API surface for jib-core


briandealwis:
`JibContainerBuilder` has methods that require use of objects outside of `com.google.cloud.tools.jib.api`, which complicates turning Jib Core into a JPMS module as we cannot just export `com.google.cloud.tools.jib.api`.  For example (not exhaustive):

    com.google.cloud.tools.jib.configuration.LayerConfiguration 
      - addLayerConfiguration
    com.google.cloud.tools.jib.configuration.Port
       - addExposedPort
    com.google.cloud.tools.jib.filesystem.AbsoluteUnixPath
       - addLayer
       - addVolume*
       - setWorkingDirectory

There are other helper classes that are very useful too:

    com.google.cloud.tools.jib.image.ImageReference
       - for parsing
    com.google.cloud.tools.jib.frontend.CredentialRetrieverFactory
       - at a minimum for inferCredentialHelper() and dockerConfig()

loosebazooka:
do we still need the `api` package if we're defining what's internal and what's external via JPMS?

briandealwis:
JPMS defines what's internal and external by packages.  So _yes_, unless we create internal packages (e.g., `.internal`).

chanseokoh:
We believe the public API is in a reasonably good shape. Closing.

PR #1708, opened 2019-05-09
Move useful classes to api package

TadCordle:
Moved:
* `AbsoluteUnixPath`
* `CacheDirectoryCreationException`
* `DescriptorDigest`
* `FilePermissions`
* `ImageReference`
* `InvalidImageReferenceException`
* `Port`
* `RegistryException` (and some its children)
* `RelativeUnixPath`

Part of #1428


chanseokoh:
I think we should rename `DescriptorDigest`, especially given that this is becoming public (later). I feel the `Descriptor` part is out of place. The class might have been specifically for [`application/vnd.oci.descriptor.v1+json`](https://github.com/opencontainers/image-spec/blob/master/descriptor.md) (this is not even clear), but in reality, this is just a SHA-256 digest and used for everything, including all sorts of JSON and layers. This really should be `Sha256Digest` or just `Digest` if we anticipate extension in the future.

And I'd like to change `BlobDescriptor` to `SizeAndDigest` or something similar as I said in the past.

But let's not do renaming in this PR.

PR #1712, opened 2019-05-10
Evaluate settings.xml later, and per id

### LINKED ISSUE #1709: [regression in 1.2 - vs 1.0] com.google.cloud.tools.jib.maven.DecryptedMavenSettings requires all servers to be decryptable
**Description of the issue**:

in 1.0 only the used servers by the maven plugin was decrypted.
In 1.2, all are decrypted due to com.google.cloud.tools.jib.maven.DecryptedMavenSettings#from.

**Expected behavior**:

Get back 1.0 behavior.

**Steps to reproduce**: 

Add a ciphered server not used by jib which can't be deciphered.

TIP: use an invalid password, in my case it was ${env.MISSING_VAR}

**Environment**: 

Linux but probably any.

**Log output**: 

Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.2.0:build (build) on project xxxx: Unable to decrypt settings.xml: [ERROR] Failed to decrypt password for server jetbrains: org.sonatype.plexus.components.sec.dispatcher.SecDispatcherException: java.io.FileNotFoundException: /home/travis/.m2/settings-security.xml (No such file or directory) @ server: jetbrains -> [Help 1]

(note: intentionally jib does not use settings.xml and it was working great in 1.0 but with 1.2 forced decryption makes the build failing, workaround is to pass decryption data but it is not desired)

loosebazooka:
- Also adds in SettingsFixture for tests that want to
  use an actual instance of settings.xml and optionally
  can create a SettingsDecrypter from a settings-security.xml
- The tests should be much more concrete now.
- `Function<String, Optional<AuthProvider>>` was turned in an interface `InferredAuth`, so we could handle exceptions better.

fixes #1709


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InferredAuthProvider.java
```diff
@@ -0,0 +1,7 @@
+package com.google.cloud.tools.jib.plugins.common;
```
chanseokoh: copyright

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InferredAuthProvider.java
```diff
@@ -0,0 +1,7 @@
...
+
+import java.util.Optional;
+
+public interface InferredAuthProvider {
```
chanseokoh: Hmm... I remember sometimes build fails if a public class doesn't have a Javadoc.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InferredAuthException.java
```diff
@@ -0,0 +1,11 @@
+package com.google.cloud.tools.jib.plugins.common;
+
+/**
+ * Indicates that the {@link InferredAuthProvider} failed encountered a failure while trying to
```
chanseokoh: "failed encountered" --> "encountered"

jib-maven-plugin/src/test/resources/maven/settings/settings-security.xml
```diff
@@ -0,0 +1,3 @@
+<settingsSecurity>
+  <master>{}X6FMi5KzfGgHeRrZdTiwUdXfVSSqcNdXvVjt3tw7oxs=</master>
```
chanseokoh: The empty content in the two curly braces seems weird. Shouldn't it be

`{X6FMi5KzfGgHeRrZdTiwUdXfVSSqcNdXvVjt3tw7oxs=}`

?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -18,33 +18,58 @@
...
 
-  private static void configureProxy(DecryptedMavenSettings settings, String protocol) {
-    settings
-        .getProxies()
-        .stream()
-        .filter(Proxy::isActive)
-        .filter(proxy -> protocol.equals(proxy.getProtocol()))
-        .findFirst()
-        .ifPresent(ProxyProvider::setProxyProperties);
+    System.out.println(proxies.size());
```
chanseokoh: stray debug output

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -18,33 +18,58 @@
...
+    if (proxies.size() == 0) {
+      return;
+    }
+
+    SettingsDecryptionRequest request = new DefaultSettingsDecryptionRequest().setProxies(proxies);
+    System.out.println("Decrypter" + decrypter);
```
chanseokoh: stray debug output

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -18,33 +18,58 @@
...
-        .filter(Proxy::isActive)
-        .filter(proxy -> protocol.equals(proxy.getProtocol()))
-        .findFirst()
-        .ifPresent(ProxyProvider::setProxyProperties);
+    System.out.println(proxies.size());
+    if (proxies.size() == 0) {
+      return;
+    }
+
+    SettingsDecryptionRequest request = new DefaultSettingsDecryptionRequest().setProxies(proxies);
```
chanseokoh: Is it not possible to decrypt only the proxies we will use?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -18,33 +18,58 @@
...
+
+    for (SettingsProblem problem : result.getProblems()) {
+      if (problem.getSeverity() == SettingsProblem.Severity.ERROR
+          || problem.getSeverity() == SettingsProblem.Severity.FATAL) {
+        throw new MojoExecutionException(
+            "Unable to decrypt proxy info from settings.xml: " + problem);
+      }
+    }
+
+    result.getProxies().forEach(ProxyProvider::setProxyProperties);
```
chanseokoh: This may not be right? What we've been doing is to call this twice on 1) the first active HTTP; and 2) the first active HTTPS. Looks like this code is calling this for all active proxies.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
loosebazooka: Oops

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
loosebazooka: Looks like I can fix this and [this](https://github.com/GoogleContainerTools/jib/pull/1712/files#r283058510) together.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InferredAuthProvider.java
loosebazooka: oh good catch.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
loosebazooka: I was confused why the tests were passing, but we already have some code to prevent this (incase the user has set outside of the jib process), either way I'll reduce the search scope: https://github.com/GoogleContainerTools/jib/blob/a132b7f0a3a104d39875023deda60f125cb7e6e6/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java#L58

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
loosebazooka: this should be fixed now.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -18,33 +18,64 @@
...
    * @param settings Maven settings
    */
-  static void init(DecryptedMavenSettings settings) {
-    configureProxy(settings, "https");
-    configureProxy(settings, "http");
-  }
+  static void populateSystemProxyProperties(Settings settings, SettingsDecrypter decrypter)
+      throws MojoExecutionException {
+    List<Proxy> proxies =
+        PROXY_PROTOCOLS
```
chanseokoh: How about having this as a local var instead of having a field? It gives me a false impression that I can readily add other protocols, but it really must be either `http` or `https` to form `http.proxyHost` or `https.proxyHost`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ProxyProvider.java
```diff
@@ -18,33 +18,64 @@
...
    *
    * @param settings Maven settings
    */
-  static void init(DecryptedMavenSettings settings) {
-    configureProxy(settings, "https");
-    configureProxy(settings, "http");
-  }
+  static void populateSystemProxyProperties(Settings settings, SettingsDecrypter decrypter)
+      throws MojoExecutionException {
+    List<Proxy> proxies =
```
chanseokoh: nit: since this method is long, `activeHttpAndHttpsProxies` for clarity? They must have exactly two proxies, one for HTTP and the other for HTTPS.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentialsTest.java
```diff
@@ -17,45 +17,78 @@
 package com.google.cloud.tools.jib.maven;
 
 import com.google.cloud.tools.jib.plugins.common.AuthProperty;
-import java.util.Arrays;
+import com.google.cloud.tools.jib.plugins.common.InferredAuthException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Optional;
-import org.apache.maven.settings.Server;
+import org.hamcrest.core.StringStartsWith;
```
chanseokoh: Huh, I've been using `org.hamcrest.CoreMatches::startsWith`. What's the preference?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/ProxyProviderTest.java
```diff
@@ -187,82 +227,49 @@ public void testSetProxyProperties_someValuesUndefined() {
...
+      throws MojoExecutionException {
+    ProxyProvider.populateSystemProxyProperties(httpsOnlyProxySettings, settingsDecrypter);
 
+    Assert.assertEquals("proxy2.example.com", System.getProperty("https.proxyHost"));
     Assert.assertNull(System.getProperty("http.proxyHost"));
-    Assert.assertEquals("proxy2 host", System.getProperty("https.proxyHost"));
+  }
+
+  @Test
+  public void testPopulateSystemProxyProperties_EncryptedProxy() throws MojoExecutionException {
```
chanseokoh: "_encryptedProxy" (capitalization)

jib-maven-plugin/src/test/resources/maven/settings/settings-security.xml
chanseokoh: I guess this is correct, given that tests are working.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InferredAuthProvider.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Find auth credentials for a specific registry.
+   *
+   * @param registry we want credential for
+   * @return auth information for the registry (can be empty)
+   * @throws InferredAuthException if the auth discovery process resulted in a fatal error
+   */
+  Optional<AuthProperty> inferredAuth(String registry) throws InferredAuthException;
```
chanseokoh: `inferAuth`?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentialsTest.java
loosebazooka: I have no preference, this was probably my first search result. Looks like `CoreMatchers` just calls into `StringStartsWith`. I changed it anyway to `CoreMatchers` for consistency.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsProxyProvider.java
```diff
@@ -18,33 +18,62 @@
...
-  static void init(DecryptedMavenSettings settings) {
-    configureProxy(settings, "https");
-    configureProxy(settings, "http");
-  }
+  static void activateHttpAndHttpsProxies(Settings settings, SettingsDecrypter decrypter)
+      throws MojoExecutionException {
+    List<Proxy> proxies = new ArrayList<>(2);
+    for (String protocol : ImmutableList.of("http", "https")) {
+      if (areProxyPropertiesSet(protocol)) {
+        return;
```
chanseokoh: `continue;`

loosebazooka: ugh, why are the tests passing? I'm so confused.

loosebazooka: I'm not confused, I need to write a few more tests.

loosebazooka: done

loosebazooka: :man_facepalming:

PR #1714, opened 2019-05-13
Add version to skaffold-files-v2 output

briandealwis:
Add a `version` field to the JSON object output by the `skaffold-files-v2` goal so as to allow the format to evolve.  We populate this field with the Jib plugin version.

Note that the `ProjectInfo.VERSION` field is `@Nullable`, since it will be `null` when run from tests.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/ProjectInfo.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/ProjectInfo.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.jib;
 
+import javax.annotation.Nullable;
+
 /** Constants relating to the Jib project. */
 public class ProjectInfo {
 
@@ -25,7 +27,8 @@ public class ProjectInfo {
   /** Link to file an issue against the GitHub repository. */
   public static final String GITHUB_NEW_ISSUE_URL = GITHUB_URL + "/issues/new";
 
-  /** The project version. */
+  /** The project version. May be {@code null} if the version cannot be determined. */
+  @Nullable
   public static final String VERSION = ProjectInfo.class.getPackage().getImplementationVersion();
 
   private ProjectInfo() {}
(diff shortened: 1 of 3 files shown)


jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutputTest.java
```diff
@@ -26,7 +26,7 @@
 public class SkaffoldFilesOutputTest {
```
loosebazooka: I find it strange that this is in plugins-common, and if it is shouldn't we be able to do an install and get a concrete version?

chanseokoh:
Have you checked if this doesn't break Skaffold?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutputTest.java
TadCordle: About the plugins-common question, I think we decided to move it there in this PR https://github.com/GoogleContainerTools/jib/pull/1485#pullrequestreview-204434440

briandealwis:
Skaffold is fine: Go's JSON unmarshalling only assigns struct fields that are found in the JSON.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/SkaffoldFilesOutputTest.java
loosebazooka: This comment is a mistake, I thought this test was actually using the gradle/maven plugins and we were being lazy about it. I should've looked at the whole file.

jib-core/src/main/java/com/google/cloud/tools/jib/ProjectInfo.java
```diff
@@ -25,7 +27,8 @@
   /** Link to file an issue against the GitHub repository. */
   public static final String GITHUB_NEW_ISSUE_URL = GITHUB_URL + "/issues/new";
 
-  /** The project version. */
+  /** The project version. May be {@code null} if the version cannot be determined. */
+  @Nullable
```
loosebazooka: Should this be injected into some config by maven/gradle? Or maybe it's another value used by skaffold files?

loosebazooka: I think I want to refactor `ProjectInfo.VERSION` out of here. It's kind of an implementation detail that this resolved correctly. I'd rather populate it from the maven/gradle plugins with a reasonable default. But that's not the purpose of this PR.https://github.com/GoogleContainerTools/jib/issues/1716

PR #1717, opened 2019-05-14
Remove RegistryAuthenticator.Initializer

chanseokoh:
@loosebazooka I don't know if you tried something on the `Initializer`, but I think it's textually redundant in that it's duplicated in other methods.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -173,28 +173,27 @@ class PullBaseImageStep
         } catch (RegistryUnauthorizedException registryUnauthorizedException) {
           // The registry requires us to authenticate using the Docker Token Authentication.
           // See https://docs.docker.com/registry/spec/auth/token
-          RegistryAuthenticator registryAuthenticator = null;
           try {
-            registryAuthenticator =
+            RegistryAuthenticator registryAuthenticator =
                 buildConfiguration
                     .newBaseImageRegistryClientFactory()
                     .newRegistryClient()
                     .getRegistryAuthenticator();
+            if (registryAuthenticator != null) {
+              Authorization pullAuthorization =
+                  registryAuthenticator.setCredential(registryCredential).authenticatePull();
+
+              return new BaseImageWithAuthorization(
+                  pullBaseImage(pullAuthorization, progressEventDispatcher), pullAuthorization);
+            }
+
           } catch (InsecureRegistryException insecureRegistryException) {
             // Cannot skip certificate validation or use HTTP; fall through.
           }
-          if (registryAuthenticator == null) {
-            eventDispatcher.dispatch(
-                LogEvent.error(
-                    "Failed to retrieve authentication challenge for registry that required token authentication"));
-            throw registryUnauthorizedException;
-          }
-
-          registryAuthorization =
-              registryAuthenticator.setCredential(registryCredential).authenticatePull();
-
-          return new BaseImageWithAuthorization(
-              pullBaseImage(registryAuthorization, progressEventDispatcher), registryAuthorization);
+          eventDispatcher.dispatch(
+              LogEvent.error(
+                  "Failed to retrieve authentication challenge for registry that required token authentication"));
+          throw registryUnauthorizedException;
         }
       }
     }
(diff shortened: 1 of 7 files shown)


chanseokoh:
@loosebazooka @briandealwis we talked about starting afresh about auth, and I think this PR may actually help that by eliminating the weird `Initializer` out of the picture. This looks easier to understand. WDYT?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -176,27 +173,27 @@ public BaseImageWithAuthorization call()
...
-            throw registryUnauthorizedException;
+          try {
+            RegistryAuthenticator registryAuthenticator =
+                buildConfiguration
+                    .newBaseImageRegistryClientFactory()
+                    .newRegistryClient()
+                    .getRegistryAuthenticator();
+            if (registryAuthenticator != null) {
+              Authorization pullAuthorization =
+                  registryAuthenticator.setCredential(registryCredential).authenticatePull();
```
loosebazooka: Maybe `authenticatePull()` and `authenticatePush()` should take the credential? There doesn't seem to be any point to setting the credential on the object (it forces us to make it mutable). Wdyt? I don't see us reusing `registryAuthenticator` anywhere anyway.

chanseokoh: That's a great idea. I actually think `getRegistryAuthenticator()` can accept `Credential` directly.

loosebazooka: I think there's some code that returns a null `RegistryAuthenticator`, does that matter here?

chanseokoh: Well, it's simpler to pass `Credential` to `authenticatePush/Pull`. Done.

chanseokoh: > I think there's some code that returns a null `RegistryAuthenticator`

The code doing the null check on `RegistryAuthenticator`. Are you talking about the `MalformedURLException` case? Seems like no one is throwing this checked exception, so I removed that.

chanseokoh: Oh, I think I got what you're saying. I decided to pass `Credential` to `authenticatePush/Pull` as you originally suggested.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -245,21 +156,25 @@ private String getToken() {
   /**
    * Authenticates permissions to pull.
    *
+   * @param credential the credential used to authenticate
    * @return an {@code Authorization} authenticating the pull
    * @throws RegistryAuthenticationFailedException if authentication fails
    */
-  public Authorization authenticatePull() throws RegistryAuthenticationFailedException {
-    return authenticate("pull");
+  public Authorization authenticatePull(@Nullable Credential credential)
```
loosebazooka: Sorry, one more question. Should we remove the `@Nullable` on `Credential`? Can we authenticate without credentials?

chanseokoh: Some registries require authorizing with an auth server even if the image is public, like public images on Docker Hub. And sometimes, it's inevitably null if no credential is configured, which may or may not lead to a failure. The auth server will decide if it will be authorized or not.

loosebazooka: For some reason I thought we had logic to decide if we needed to authorize at all?

chanseokoh: Yeah, there is some place something similar is happening. For example, first you just try pulling a public Docker Hub image, and if it fails (usually it does, if not always) and the registry returns with the auth challenge. Then we go to the auth server in the challenge to get the pull authorization. As a public image, you don't need any credentials. (Not setting the `Authorization` header) Passing null in this case will work as the image is public.

PR #1720, opened 2019-05-16
Refactor progress completion reporting

### LINKED ISSUE #1356: Progress feedback cleanup: unfinished allocation management
Originally from #1297

Consider have [`AllocationCompletionTracker` only give back unfinished leaf allocations](https://github.com/GoogleContainerTools/jib/pull/1333#discussion_r240398273) or to try seeing if it's cleaner to have `ProgressLogger` track the unfinished allocations.

chanseokoh:
In the current form, progress reporting, completion tracking, and printing output is a highly complex business. I'm taking care of a few things to mitigate that:

- There's no need to transmit unfinished `Allocation` from the completion tracker to the progress displayer because we only print the task descriptions (i.e., `Allocation` descriptions). So, having `getUnfinishedLeafTasks()` instead of `getUnfinishedAllocations()`.

- Fixes #1356: the progress displayer doesn't have to do the computation to filter leaf tasks out of all tasks transmitted. The completion tracker can just return unfinished leaves from the get-go.


jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/AllocationCompletionTracker.java
```diff
@@ -112,56 +115,67 @@ boolean updateProgress(Allocation allocation, long units) {
...
           .getParent()
           .ifPresent(
               parentAllocation ->
-                  updateInsertionOrderUnits(
+                  updateIndexedRemainingUnits(
                       Preconditions.checkNotNull(completionMap.get(parentAllocation)), 1L));
     }
   }
+
+  ImmutableList<String> getUnfinishedLeafTasks() {
```
chanseokoh: Note to reviewers: this is moved from `ProgressDisplayGenerator.getLeafAllocations()`, fixing the TODO in that method. (#1356). Check the original method first.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/ProgressDisplayGenerator.java
```diff
@@ -81,50 +79,5 @@ private static String generateProgressBar(double progress) {
...
-
-  /**
-   * Gets a list of just the leaf {@link Allocation}s in {@code unfinishedAllocations} in the same
-   * order as they appear in {@code unfinishedAllocations}.
-   *
-   * @param unfinishedAllocations the list of unfinished {@link Allocation}s
-   * @return the list of unfinished {@link Allocation}s
-   */
-  // TODO: Optimization: Change AllocationCompletionTracker#getUnfinishedAllocations to only return
-  // leaf Allocations so that this computation is unnecessary.
```
chanseokoh: Note to reviewers: this TODO is #1356.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -68,7 +68,7 @@
         new ProgressEventHandler(
             update -> {
               lastProgress = update.getProgress();
-              areTasksFinished = update.getUnfinishedAllocations().isEmpty();
+              areTasksFinished = update.getUnfinishedLeafTasks().isEmpty();
```
briandealwis: Would this test be better pushed into `Update`?

PR #1725, opened 2019-05-17
Move EventDispatcher logic to EventHandlers

TadCordle:
Consolidating the event system a bit. Main changes:

* Changed `EventHandlers` to use a builder to make it immutable 
* Moved `DefaultEventDispatcher#dispatch` to `EventHandlers`
* Removed `EventDispatcher` and `DefaultEventDispatcher`
* Renamed a bunch of stuff, of course

Towards https://github.com/GoogleContainerTools/jib/issues/1428


jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -523,30 +522,25 @@ BuildConfiguration toBuildConfiguration(
         .setToolName(containerizer.getToolName())
         .setExecutorService(executorService);
```
chanseokoh: nit: just chain `setEventHandlers` here.

jib-core/src/main/java/com/google/cloud/tools/jib/event/EventHandlers.java
```diff
@@ -24,45 +25,90 @@
...
-   * @return this
+   * @return the builder
    */
-  public <E extends JibEvent> EventHandlers add(
-      JibEventType<E> eventType, Consumer<E> eventConsumer) {
-    Class<E> eventClass = eventType.getEventClass();
-    handlers.put(eventClass, new Handler<>(eventClass, eventConsumer));
-    return this;
+  public static Builder builder() {
+    return new Builder();
```
chanseokoh: Personally I don't see a great value of this method, since I can do `new EventHandlers.Builder()`. Is this considered a good practice? I don't know very well.

TadCordle: I'm not sure either, but it's consistent with the rest of the code base (and is a bit more concise).

chanseokoh: OK, we're good then.

TadCordle:
Oops, merged before I saw that. I can look into it.

PR #1726, opened 2019-05-17
Make Containerizer EventHandlers non optional

TadCordle:
Regarding https://github.com/GoogleContainerTools/jib/commit/7f7c2e607772fdab324226dfcc7432ddb8bdda73#r33584738


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/AuthenticationMethodRetrieverIntegrationTest.java
```diff
@@ -26,7 +26,7 @@
 /** Integration tests for {@link AuthenticationMethodRetriever}. */
 public class AuthenticationMethodRetrieverIntegrationTest {
 
-  private static final EventHandlers EVENT_HANDLERS = EventHandlers.none();
+  private static final EventHandlers EVENT_HANDLERS = EventHandlers.NONE;
```
chanseokoh: Maybe at this point, we don't need the field (`EVENT_HANDLERS`) anymore. We can use `EventHandlers.NONE` directly in code.

PR #1728, opened 2019-05-20
Move useful event classes to api package

TadCordle:
Towards #1428. Moved the following classes to `api`:
* `BuildStepType`
* `EventHandlers`
* `JibEvent`
* `JibEventType`
* `LayerCountEvent`
* `LogEvent`

All of these classes are useful for setting up event handlers, so they should be part of the public API.


briandealwis:
I'm hesitant about exposing TimerEvent, TimerEvent, and ProgressEvent (and Allocation).  They've had no external consumption or feedback.  I wonder if we couldn't fold TimerEvent into ProgressEvent.

TadCordle:
If we aren't exposing them, is there much point having them in the first place? Why have events that aren't exposed and therefore can't be handled by the user?

If we do stick with exposing `TimerEvent`, maybe we should also expose `TimerEventHandler` like `ProgressEventHandler` is exposed.

briandealwis:
If they're not in `.api` then it  means that they still need some maturation and their API may change (_Use at own risk_). 

I think LogEvent is worth keeping separate: it's pretty solid and it easily maps to other loggers.

I assumed that the parent relationship in TimerEvent's Timer instances and ProgressEvent's Allocation instances were the same.  But maybe not.  It's hard to figure out how to use the information returned, which is why I don't think they're ready as API.

TadCordle:
Hmm, I think you're right. I'll move `ProgressEvent` and `TimerEvent` back to their original package for now.

TadCordle:
The problem now is that `ProgressEvent` and `TimerEvent` are still exposed in the api package from `JibEventType`. Hmmm.

TadCordle:
Going off your first suggestion, maybe we don't even need `JibEventType` anymore, and just change how methods are added to, e.g. `eventHandlers.add(LogEvent.class, logEvent -> ...)`. Could try this in a followup if it seems like a good idea.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -17,11 +17,11 @@
 package com.google.cloud.tools.jib.api;
 
 import com.google.cloud.tools.jib.Command;
-import com.google.cloud.tools.jib.builder.BuildStepType;
```
briandealwis: I don't think we need to expose `BuildStepType` either.

TadCordle: `BuildStepType` is used in `LayerCountEvent`.

`.event.events` is mainly where I put the event types themselves, and `.event` was for everything else. I can combine them if you want, but I kinda liked having all the event types isolated in their own package (especially if we get rid of `JibEventType`). But since there aren't that many maybe we don't need separate packages.

briandealwis: `.api` had been flat until this.  If we want to go this direction then might make more sense to invert and use the `.internal`.

TadCordle: Hmm that's true. I'll make it completely flat for now until we figure out what we want in the public api/fix all the unintended exposed stuff, then maybe we can do another pass where we organize into sub-packages/use `.internal` for everything else.

chanseokoh:
So I've looked into `LayerCountEvent`, and I think it should not be moved to the API. See my comment [here](https://github.com/GoogleContainerTools/jib/pull/1744#issue-283386086).

TadCordle:
Makes sense.

TadCordle:
Going to close this and rework given our recent discussion

PR #1729, opened 2019-05-20
Adjust MainClassFinder api

TadCordle:
Towards #1428. Made `MainClassFinder` use a static method instead of requiring instantiation, and moves the class to the `api` package (this is a useful class for building the entrypoint for a Java container).


jib-core/src/main/java/com/google/cloud/tools/jib/api/MainClassFinder.java
```diff
@@ -137,29 +135,39 @@ public MethodVisitor visitMethod(
...
-  public Result find() {
+  public static Result find(List<Path> files, EventHandlers eventHandlers) {
     List<String> mainClasses = new ArrayList<>();
     for (Path file : files) {
-      findMainClass(file).ifPresent(mainClasses::add);
+      // Makes sure classFile is valid.
+      if (!Files.exists(file)
+          || !Files.isRegularFile(file)
+          || !file.toString().endsWith(".class")) {
+        continue;
```
briandealwis: We should log this, maybe at debug?

PR #1731, opened 2019-05-21
Remove JibEventType

TadCordle:
Towards #1428, followup to #1728. Simplifies things by only needing to pass in a `JibEvent` class rather than wrapping it in a `JibEventType`, and also prevents exposing events that may not be API-ready.


chanseokoh:
I thought about this too, but I wasn't sure if accepting `Class` as an argument of a public library method looks pretty. Theoretically it also allows users to pass any user-derived `JibEventType` although it is harmless as no-op. Maybe OK to remove the type class if we go for `addTimerEventHander()`, etc. that does not accept `Class`. There are pros and cons.

TadCordle:
>Maybe OK to remove the type class if we go for `addTimerEventHander()`

I kind of wanted to refrain from doing something like this, as it means we'd have to expose events that aren't API-ready (see https://github.com/GoogleContainerTools/jib/pull/1728#issuecomment-494455060). Maybe that shouldn't matter and we should focus on making all the event types API-ready.

To me it feels cleaner to directly correspond the event type to its class, rather than using a `JibEventType` constant. It does allow users to pass any user-derived `JibEventType`, but that also means users can dispatch their own events, since `dispatch()` is in `EventHandlers` now, which could be useful.

loosebazooka:
I feel like the event handler should be strictly one-way in the api: Users of jib-core can handle events, but they cannot send events. What we do in plugins-common can be an internal/hidden implementation.

loosebazooka:
Also what do you think about splitting moving the events into `api` and the change in logic into 2 PRs?

TadCordle:
We did already merge this https://github.com/GoogleContainerTools/jib/pull/1725 which moved `dispatch()` into `EventHandlers`, so that means some of this will need to be undone or reworked.

TadCordle:
This was based off a closed branch, so it'll be easier to close and rework

PR #1733, opened 2019-05-21
Move ImageFormat to api package

TadCordle:
Towards #1428. Moves `ImageFormat` to api package and removes references to underlying classes.


jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -389,7 +390,8 @@ public JibContainerBuilder addLabel(String key, String value) {
    * @return this
    */
   public JibContainerBuilder setFormat(ImageFormat imageFormat) {
-    buildConfigurationBuilder.setTargetFormat(imageFormat.getManifestTemplateClass());
+    buildConfigurationBuilder.setTargetFormat(
```
briandealwis: Can we move this logic in `BuildConfiguration` instead?  That way we avoid passing classes around, and pushes the class decision to be closer to the location that actually instantiates the class.

PR #1736, opened 2019-05-23
 Move LayerConfiguration and LayerEntry to api package

TadCordle:
Towards #1428. Also adds a new `LayerConfiguration#addEntry` overload that just takes a `LayerEntry`, and moves the default providers from `LayerEntry` to `LayerConfiguration`.


jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerEntry.java
```diff
@@ -76,7 +59,7 @@ public LayerEntry(Path sourceFile, AbsoluteUnixPath extractionPath) {
     this(
         sourceFile,
         extractionPath,
-        DEFAULT_FILE_PERMISSIONS_PROVIDER.apply(sourceFile, extractionPath));
+        LayerConfiguration.DEFAULT_FILE_PERMISSIONS_PROVIDER.apply(sourceFile, extractionPath));
```
briandealwis: I'm not a fan of this backreach into LayerConfiguration.  To my mind, the providers provide a policy, and LayerEntry is the mechanism, and it's good to [keep policy separated from mechanism](https://en.wikipedia.org/wiki/Separation_of_mechanism_and_policy).  I know we use these constructors from our tests, but I think fixing the tests would be best.

My broader thoughts were that it should be possible for someone to configure a different timestamp provider that could look up the path in a git commit log, for example, and use the last commit's timestamp on the file.  When we have the policy baked in down low, it's hard to ensure that an alternative provider will always be consulted.

Anyways, I'm fine with this for now and we'll tackle separation later.

briandealwis: And I meant to push approve.

PR #1744, opened 2019-05-29
Remove informative BuildStepType from events

chanseokoh:
`BuildStepType` is purely informational, and I noticed that it's only used to be passed into `ProgressEvent` and `LayerCountEvent`. And most of the enum values are not ever used.

However, `ProgressEvent` has an `Allocation` which has a description. And we don't make use of the build type anywhere. Even if `ProgressEvent` goes public, I think the description in `Allocation` is enough. The steps listed in `BuildStepType` seem rather internal information, which may change over time.

For `LayerCountEvent`, this is only for testing, and the constructor already says it's for internal use only. The use-case in our testing is too specific in that the event is fired in only 4 steps. (I even wish we could get rid of this, ~~but there's no good alternative to get layer counts in testing~~.) And I don't think this event will ever go public. Even if we assume this is public, the build steps feel like internal information that the user should not depend on, as I said.

 I think removing this from build steps helps understand the code with one less internal state field to reason about.


briandealwis:
I don't think `LayerCountEvent` is even consumed by anything!

TadCordle:
>I don't think LayerCountEvent is even consumed by anything!

If that's true, I wouldn't mind just getting rid of it.

TadCordle:
Although, I feel like it's less that we're using the `LayerCountEvent` in the test, and more that we're testing `LayerCountEvent` itself. I don't think there's much value lost in removing it, if any.

chanseokoh:
What is a bit meaningful is that we check the layer size in some build steps (like `PullAndCacheBaseImageLayerStep`) in the integration test (`ContainerizerIntegrationTest`). Let me see if I can check the layer size in other ways.

chanseokoh:
Removed it. Looks nice. The integration test passes locally.

Issue #905, opened 2018-08-30
Add java application metadata in docker container


ST-DDT:
**Description of the issue**:

Feature-Request: I have a spring boot application that exposes some metadata about the currently running application.
i collect these metadata using the following snippet:

````java
@Configuration
public class MetricsConfig {

    @Bean
    public MeterRegistryCustomizer<MeterRegistry> metricsCommonTags() {
        final Package pack = getClass().getPackage();
        return registry -> registry.config()
                .commonTags(
                // @formatter:off
"implementation-title",	Objects.toString(pack.getImplementationTitle(), "DEVault"),
"implementation-version", Objects.toString(pack.getImplementationVersion(), "DEV")
		// @formatter:on
                );
    }

}
````

If i start the application using the fat jar or as a library of another project this code works fine, however
if i launch it using the generated docker image it only uses the fallback values. I assume this is caused by the missing META-INF/MANIFEST.mf file. Currently there is no way to specify these attributes and the plugin does not pull them by itself. (Similar to #159 )

These metadata are also used by Log4j2 (and other Logging frameworks) to include a version number with the stacktrace entries.

**Expected behavior**:

I would expect that the metadata would be present also in the docker image.
/
It should be possible to specify these metadata in the plugin's configuration.

**Steps to reproduce**:

* Write a main class that prints: `Main.class.getPackage().getImplementationVersion()`
* Add build metadata to the jar plugin config (see below)
* Build the jar
* Run the jar
* It will show the version number of the project
* Build the docker image
* Run the docker image
* It will show null

````xml
        <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-jar-plugin</artifactId>
            <configuration>
                <archive>
                    <manifest>
                        <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                        <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
                    </manifest>
                    <manifestEntries>
                        <Build-Time>${maven.build.timestamp}</Build-Time>
                    </manifestEntries>
                </archive>
            </configuration>
        </plugin>
````

**Environment**:

* Maven 3.5
* Java 8
* Windows 10

(but I assume this is a general issue)

**Possible Workaround**

Move docker packaging to its own project that uses the original project as (jar) dependency.

chanseokoh:
[...]

PR #1746, opened 2019-05-29
Add Maven support for containerizing packaged JAR

chanseokoh:
For #905 on the Maven side.

Things are working. We can discuss if the implementation and the new configuration make sense. No tests are added for now.

-- added from below

As for the new config,

```xml
<configuration>
  <!--
    Optional.
    "exploded" is the default and the current behavior.
    "packaged" puts a JAR artifact (and potentially a WAR in the future).
  -->
  <containerizingMode>exploded</containerizingMode>
</configuration>
```

When the mode is "packaged", the layers built by Jib will be

1. Dependency layer
2. "Others or Classpath" layer (as already provided by `JavaContainerBuilder`): to contain a single JAR under `/classpath`, e.g., `/classpath/awesome-app-2.3.jar`

Therefore whenever the app is modified, only the JAR layer needs to be rebuilt.

And the entrypoint will be set to

```
java -cp /classpath/*:/app/libs/* example.MainClass
```

In order for the JAR containerization to work, the build has to run with `mvn package` instead of `mvn compile`. I should not forget to show a correct error message when the JAR artifact is not built.


chanseokoh:
I'll add some more explanation about what's going on and how the new config will look like shortly.

chanseokoh:
As for the new config,

```xml
<configuration>
  <!--
    Optional.
    "exploded" is the default and the current behavior.
    "packaged" puts a JAR artifact (and potentially a WAR in the future).
  -->
  <containerizingMode>exploded</containerizingMode>
</configuration>
```

When the mode is "packaged", the layers built by Jib will be

1. Dependency layer
2. "Others" layer: to contain a single JAR under `/classpath`, e.g., `/classpath/awesome-app-2.3.jar`

And the entrypoint will be set to

```
java -cp /classpath/*:/app/libs/* example.MainClass
```

In order for the JAR containerization to work, the build has to run with `mvn package` instead of `mvn compile`. I should not forget to show a correct error message when the JAR artifact is not built.

examples/helloworld/src/main/java/example/HelloWorld.java
```diff
@@ -28,9 +30,11 @@
 public class HelloWorld {
 
   public static void main(String[] args) throws URISyntaxException, IOException {
```
chanseokoh: The previous code did not work inside a JAR, as it was complaining that there is no "file system" inside it. The new code works both inside and outside a JAR.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -471,21 +472,15 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
             .collect(Collectors.toList());
     for (Path file : addedDependencies) {
       // Add dependencies to layer configuration
```
chanseokoh: This is simple code cleanup. No change to the logic.

chanseokoh:
I still need to add integration tests and may sort out a couple things, but good enough to start reviewing.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -380,6 +399,17 @@ static AbsoluteUnixPath getAppRootChecked(
     }
   }
 
+  @VisibleForTesting
+  static ContainerizingMode getContainerizingModeChecked(RawConfiguration rawConfiguration)
+      throws InvalidContainerizingModeException {
+    String mode = rawConfiguration.getContainerizingMode();
+    try {
+      return ContainerizingMode.valueOf(mode.toUpperCase(Locale.US));
```
briandealwis: So you want to allow _PaCkAgEd_ and _eXPLODEd_?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -197,36 +199,42 @@ static int getVersionFromString(String versionString) {
...
-                  .collect(Collectors.toList()))
+          .addDependencies(getDependencies())
           .toContainerBuilder();
 
     } catch (IOException ex) {
       throw new IOException(
           "Obtaining project build output files failed; make sure you have compiled your project "
               + "before trying to build the image. (Did you accidentally run \"mvn clean "
-              + "jib:build\" instead of \"mvn clean compile jib:build\"?)",
+              + "jib:build\" instead of \"mvn clean {compile|package} jib:build\"?)",
```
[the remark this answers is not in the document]
briandealwis: Oh, and we use `prepare-package` in Skaffold-Jib.  Maybe we should use `package` regardless.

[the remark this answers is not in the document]
loosebazooka: Maybe? Is creating a jar really all that expensive?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -197,36 +199,42 @@ static int getVersionFromString(String versionString) {
...
         Path explodedWarPath =
-            Paths.get(project.getBuild().getDirectory()).resolve(project.getBuild().getFinalName());
-        return JavaContainerBuilderHelper.fromExplodedWar(baseImage, explodedWarPath, appRoot);
+            Paths.get(project.getBuild().getDirectory(), project.getBuild().getFinalName());
+        return JavaContainerBuilderHelper.fromExplodedWar(javaContainerBuilder, explodedWarPath);
+      }
+
+      if (containerizingMode == ContainerizingMode.PACKAGED) {
+        String jarName = project.getBuild().getFinalName() + "." + project.getPackaging();
+        Path jar = Paths.get(project.getBuild().getDirectory(), jarName);
```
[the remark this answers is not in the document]
chanseokoh: I could not find any other way. I believe it is the Maven JAR plugin (in the usual case) that packages a JAR, so I think the Maven API doesn't really have such a method. Looking at https://maven.apache.org/plugins/maven-jar-plugin/jar-mojo.html, I think the JAR name is fixed unless `<classifier>` is used. That said, this code is incomplete in that if either `<classifier>` or `<outputDirectory>` is defined, this won't work. In that case, I think we should really parse the JAR plugin config, but I'm punting that for the moment. Basically the same goes with the WAR case.

For the fat-JAR case, unfortunately it will end up duplicating dependencies. I think we should improve this, for example, by identifying if it is a fat JAR with various heuristics and skipping creating a dep layer. If you can think of other potential issues or edge cases, we should discuss too. But basically, we should think about how we can improve this going forward.

And I don't really grasp what kind of problems we will see with shaded JARs. Do you something specific as an example?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -151,6 +152,10 @@ public void buildTar()
       throw new GradleException(
           "container.appRoot is not an absolute Unix-style path: " + ex.getInvalidPathValue(), ex);
```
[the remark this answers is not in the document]
chanseokoh: These exceptions for purely for syntactically invalid config values. `/dev/pts/0` is OK, since it is syntactically correct Unix path. No semantics involved.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -154,7 +154,7 @@ private static boolean isProgressFooterEnabled(Project project) {
 
   @Override
   public JibContainerBuilder createContainerBuilder(
-      RegistryImage baseImage, ContainerizingMode containerizingMode) {
+      RegistryImage baseImage, AbsoluteUnixPath appRoot, ContainerizingMode containerizingMode) {
```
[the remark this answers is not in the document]
chanseokoh: The field was a real annoyance and is just gone with #1752!

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -623,6 +621,11 @@ boolean getAllowInsecureRegistries() {
     return allowInsecureRegistries;
   }
 
+  public String getContainerizingMode() {
```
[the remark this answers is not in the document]
chanseokoh: The current design is that all raw values are passed through via `RawConfiguration`. Any interpretation, processing, or validation happens after reading values from `RawConfiguration`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: Haha, yeah. We won't document it. I think it's a good tradeoff to keep the code simple.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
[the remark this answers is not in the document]
chanseokoh: Fat-JAR: actually, I think the only way to run it without us putting a dependency layer is to do `java -jar fat.jar`. That is, we go into the more complicated runnable JAR business.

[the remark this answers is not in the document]
briandealwis: [`MavenProject#getAttachedArtifacts()`](https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/project/MavenProject.java#L927) will return the list of artifacts that have been created for this project.  The shade plugin allows [replacing the artifact via the `shadedArtifactAttached` property](https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#shadedArtifactAttached).  I don't know whether it's worth asking users for the classifier of the artifact to be included.

(Aside: I think Jib should be attaching the result of `buildTar`.)

That said, we still won't know whether the jar is a fat jar or a slim jar (i.e., needs dependencies).  Perhaps we should have a third containerizing-mode that indicates that dependencies are ignored.  `standalone`?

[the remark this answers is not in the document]
chanseokoh: Ah, that's interesting. I'll look into the method.

> (Aside: I think Jib should be attaching the result of `buildTar`.)

Can you elaborate on it? I don't think I get the context. What is the result and where do we attach it?

> That said, we still won't know whether the jar is a fat jar or a slim jar (i.e., needs dependencies). Perhaps we should have a third containerizing-mode that indicates that dependencies are ignored. `standalone`?

Yeah, we need to see if there are reasonable or limited ways to check if it is a fat jar, and depending on the outcome, we may need to add another mode for that. But I think just skipping dependencies won't just work with `java -cp ... main.Class`. In any case, I think this PR is good for the first iteration and reasonable complete in the non-fat case.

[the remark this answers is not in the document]
chanseokoh: > Can you elaborate on it? I don't think I get the context. What is the result and where do we attach it?

Never mind. I get it now.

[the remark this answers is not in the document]
chanseokoh: Unfortunately, `MavenProject#getAttachedArtifacts()` is returning an empty list.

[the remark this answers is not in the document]
briandealwis: Ah, I think you're considering a jar that contains other jars and whose manifest has a `Class-Path` attribute.  Whereas I'm considering a jar that includes *the contents* of other jars.

+1 to addressing this in subsequent PRs.  We might be able to examine the jar and determine if it's a fat/uberjar.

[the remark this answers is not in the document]
briandealwis: Ah, the maven-jar-plugin uses attach-artifact if there's a classifier involved, otherwise it sets the file on the project artifact.  Here's the relevant code:

https://github.com/apache/maven-jar-plugin/blob/80f58a84aacff6e671f5a601d62a3a3800b507dc/src/main/java/org/apache/maven/plugins/jar/AbstractJarMojo.java#L295-L307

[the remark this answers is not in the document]
chanseokoh: Ah, now I understand what is going on. There's still an issue. In my previous testing, `project.getArtifact()` was returning an `Artifact` instance, but `artifact.getFile()` was returning null. However, I realized if you run the "package" goal, `setFile()` is called and `artifact.getFile()` returns the actual path to the JAR. That is, if the JAR plugin is executed, we're good, but we cannot require the user always combine "package" with "jib:build". Still no good alternative to get the path to the generated JAR.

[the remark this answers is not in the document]
briandealwis: For `containerizerMode=packaged` I think requiring a `package` before `jib:build` is acceptable 😄  I suppose we could automatically invoke `package` if the artifact is not present.  But an error seems more appropriate.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -211,30 +211,44 @@ public JibContainerBuilder createContainerBuilder(
...
+      Set<Artifact> artifacts = project.getArtifacts();
+      List<Path> dependencies =
+          artifacts.stream().map(Artifact::getFile).map(File::toPath).collect(Collectors.toList());
+      javaContainerBuilder.addDependencies(dependencies);
+
+      switch (containerizingMode) {
+        case EXPLODED:
+          // Add resources, and classes
+          Path classesOutputDirectory = Paths.get(project.getBuild().getOutputDirectory());
+          Predicate<Path> isClassFile = path -> path.getFileName().toString().endsWith(".class");
```
briandealwis: Just use `path.endsWith(".class")`?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -211,30 +211,44 @@ public JibContainerBuilder createContainerBuilder(
...
+          javaContainerBuilder
+              .addResources(classesOutputDirectory, isClassFile.negate())
+              .addClasses(classesOutputDirectory, isClassFile);
+          break;
+
+        case PACKAGED:
+          // Add a JAR
+          String jarName = project.getBuild().getFinalName() + "." + project.getPackaging();
```
briandealwis: So I think this should be `project.getArtifact().getFile()`, but if it doesn't exist then throw an error (_"project build result is missing, maybe you should run `package`"_)

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: `Path.endsWith()` works on path elements, not on the entire string.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: As we discussed offline, let's check `project.getArtifact().getFile()` first and fall back to this heuristic.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
briandealwis: Ah right.  I think I've asked this before.  Might be worth adding a comment.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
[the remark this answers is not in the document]
chanseokoh: Resolved. FTR, the issue was that `package` should always be combined with a Jib goal at the time of running `mvn` for `getArtifact()` to return non-null `File`. See https://github.com/GoogleContainerTools/jib/pull/1746#discussion_r290355384

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: I've expanded the logic to identify the JAR. PTAL.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: Comments added.

chanseokoh:
Hmm... On second thought, I don't like the fallback for identifying the input JAR. It may act differently depending on whether `package` is combined with a Jib goal. I think we better need consistent and reproducible behavior.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: I think I'll revert using the Maven API (`getArtifact()` and `getAttachedArtifacts()`), as their results are non-deterministic. Rather, I think we should target the explicit `maven-jar-plugin` configuration. (In the future, we could consider checking other popular plugins.)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -161,4 +165,11 @@ boolean getAllowInsecureRegistries() {
     }
     return allowInsecureRegistries.get();
   }
+
+  @Input
+  @Optional
+  public String getContainerizingMode() {
+    String property = System.getProperty(PropertyNames.CONTAINERIZING_MODE);
```
loosebazooka: do we allow system properties for all values?

loosebazooka:
I think this looks okay, I do find that `containerizingMode` is a little long.
Do any of these make sense?
- `mode`
- `packaging` // presumably this conflicts with maven properties?

chanseokoh:
We may go with `mode` then. To me, `packaging` is a bit confusing term in the context of containerization.

TadCordle:
I feel like `mode` is really vague.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
chanseokoh: Yeah, we've been doing this for config parameters we offer.

briandealwis: It allows applying jib to a project without any modification.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -361,7 +362,9 @@ public void testCreateContainerBuilder_noClassesFiles() throws InvalidImageRefer
     Mockito.when(mockMainSourceSetOutput.getClassesDirs())
         .thenReturn(new TestFileCollection(ImmutableSet.of(nonexistentFile)));
     gradleProjectProperties.createContainerBuilder(
-        RegistryImage.named("base"), AbsoluteUnixPath.get("/anything"));
+        RegistryImage.named("base"),
+        AbsoluteUnixPath.get("/anything"),
+        ContainerizingMode.EXPLODED);
```
briandealwis: Should we have a constant for the default mode?  If we change the default, then we need to remember to change this value here (and in setupBuildConfiguration) too

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -197,36 +199,55 @@ static int getVersionFromString(String versionString) {
...
-                  .map(File::toPath)
-                  .collect(Collectors.toList()))
-          .toContainerBuilder();
+      return javaContainerBuilder.toContainerBuilder();
 
     } catch (IOException ex) {
       throw new IOException(
-          "Obtaining project build output files failed; make sure you have compiled your project "
+          "Obtaining project build output files failed; make sure you have "
+              + (containerizingMode == ContainerizingMode.PACKAGED ? "packaged" : "compiled")
```
briandealwis: Is it worth pulling these out as a variable (e.g., `String necessaryGoal = ...`)

jib-maven-plugin/src/test/resources/maven/projects/simple/src/main/java/com/test/HelloWorld.java
```diff
@@ -34,37 +35,48 @@ public static void main(String[] args) throws IOException, URISyntaxException {
     String greeting = Greeting.getGreeting();
 
     // Gets the contents of the resource file 'world'.
-    ClassLoader classLoader = HelloWorld.class.getClassLoader();
-    Path worldFile = Paths.get(classLoader.getResource("world").toURI());
-    String world = new String(Files.readAllBytes(worldFile), StandardCharsets.UTF_8);
+    ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
```
briandealwis: As noted elsewhere, the context classloader is not recommended for use unless absolutely necessary.  It shouldn't be necessary here.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -153,12 +154,15 @@ private static boolean isProgressFooterEnabled(Project project) {
 
   @Override
   public JibContainerBuilder createContainerBuilder(
-      RegistryImage baseImage, AbsoluteUnixPath appRoot) {
+      RegistryImage baseImage, AbsoluteUnixPath appRoot, ContainerizingMode containerizingMode) {
+    JavaContainerBuilder javaContainerBuilder =
+        JavaContainerBuilder.from(baseImage).setAppRoot(appRoot);
+
     try {
       if (isWarProject()) {
```
briandealwis: Should it be an error if they use containerizingMode with a war project?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -197,36 +199,55 @@ static int getVersionFromString(String versionString) {
...
   @Override
   public JibContainerBuilder createContainerBuilder(
-      RegistryImage baseImage, AbsoluteUnixPath appRoot) throws IOException {
+      RegistryImage baseImage, AbsoluteUnixPath appRoot, ContainerizingMode containerizingMode)
+      throws IOException {
+    JavaContainerBuilder javaContainerBuilder =
+        JavaContainerBuilder.from(baseImage).setAppRoot(appRoot);
+
     try {
       if (isWarProject()) {
```
briandealwis: Should it be an error if they specify containerizingMode with a war project?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: Sadly, one is a past-tense verb (ends with "-ed"), so they are slightly different.

examples/helloworld/src/main/java/example/HelloWorld.java
```diff
@@ -28,9 +30,11 @@
...
-    Path worldFile = Paths.get(Resources.getResource("world").toURI());
-    String world = new String(Files.readAllBytes(worldFile), StandardCharsets.UTF_8);
-
-    System.out.println("Hello " + world);
+    ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
```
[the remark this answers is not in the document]
chanseokoh: I had some trouble, as apparently other loades couldn't locate and read a random file (`world` text file at the root) inside a JAR, but let me try hard this time.

chanseokoh:
`HelloWorld.class.getResourceAsStream()` worked when explicitly saying `/world` is at root.

loosebazooka:
Yeah my naming suggestions were a little weak, maybe we can talk about this tomorrow at the design before submitting.

loosebazooka:
I think we decided in the design meeting that this should be good to go. Might need a rebase?

chanseokoh:
@loosebazooka I've tweaked the code around adding dependencies while resolving merging conflicts. Please take a second look at the last merge commit (https://github.com/GoogleContainerTools/jib/pull/1746/commits/be7610be57a2e4fd4e0cb399fabdcd0aae63e50f) to see things look good.

PR #1752, opened 2019-05-30
No need to pass appRoot when instantiate ProjectProperties

chanseokoh:
It has been a real pain that we have to pass `appRoot` to instantiate a `ProjectProperties` instance. This was a chicken-and-egg situation: you need to know a certain project property to process and compute a correct `appRoot` value, but in order for `PluginConfigurationProcessor (PCP)` to process it correctly, it needs an instance of `ProjectProperties`.

So, to circumvent this annoyance, we had to do these unpleasant things:
- Exposed `PCP.getAppRootChecked()` as `public`, which is inconsistent with other `getXXXChecked()` families that are exclusively used by `PCP`.
- Had a different method signature for `PCP.getAppRootChecked()` from other `getXXXChecked()` families.
- Needed `TaskCommon.isWarProject()` and `MojoCommon.isWarProject()`.
- Had to call `getAppRootChecked()` at the beginning of MOJOs to compute `appRoot` before proper config computation is carried out by `PCP`.

But now it seems like we don't have to pass `appRoot` to instantiate `ProjectProperties`.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -183,7 +183,8 @@ public class PluginConfigurationProcessor {
 
     JibContainerBuilder jibContainerBuilder =
         projectProperties
-            .createContainerBuilder(baseImage)
+            .createContainerBuilder(
+                baseImage, getAppRootChecked(rawConfiguration, projectProperties))
             .setEntrypoint(computeEntrypoint(rawConfiguration, projectProperties))
             .setProgramArguments(rawConfiguration.getProgramArguments().orElse(null))
             .setEnvironment(rawConfiguration.getEnvironment())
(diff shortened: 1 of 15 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -183,7 +183,8 @@ static PluginConfigurationProcessor processCommonConfiguration(
 
     JibContainerBuilder jibContainerBuilder =
         projectProperties
-            .createContainerBuilder(baseImage)
+            .createContainerBuilder(
+                baseImage, getAppRootChecked(rawConfiguration, projectProperties))
```
TadCordle: Maybe it doesn't matter much, but we're checking the approot twice now (once here and once in `computeEntrypoint`. I wonder if we should just call `getAppRootChecked()` once and pass the result into both `createContainerBuilder` and `computeEntrypoint`.

chanseokoh: Valid point, but `computeEntrypoint()` should still accept both `rawConfiguration` and `projectProperties`, and it's cheap to call `getAppRootChecked()`. I think it's better than adding another argument to pass to `computeEntrypoint()` from the API perspective. No need to make caller to have `appRoot` computed when the method can get it easily. Also prevents passing an inconsistent `appRoot` value by accident.

PR #1753, opened 2019-05-31
Prevent LayerEntry backreach into LayerConfiguration

### LINKED ISSUE #1743: Prevent LayerEntry backreach into LayerConfiguration
https://github.com/GoogleContainerTools/jib/pull/1736#discussion_r287351547

>I'm not a fan of this backreach into LayerConfiguration. To my mind, the providers provide a policy, and LayerEntry is the mechanism, and it's good to keep policy separated from mechanism. I know we use these constructors from our tests, but I think fixing the tests would be best.

>My broader thoughts were that it should be possible for someone to configure a different timestamp provider that could look up the path in a git commit log, for example, and use the last commit's timestamp on the file. When we have the policy baked in down low, it's hard to ensure that an alternative provider will always be consulted.

TadCordle:
Fixes #1743.


chanseokoh:
I'll let @briandealwis review this.

Issue #1745, opened 2019-05-29
jib doesn't update images when included Gradle projects change


emersonf:
**Description of the issue**:
`gradle jibDockerBuild` isn't creating new Docker images when code in Gradle included builds changes.

For the time being, I'm calling `gradle build jibDockerBuild` to work around this but that's counter-intuitive.

**Expected behavior**:
When code in an included build changes and `gradle jibDockerBuild` is re-run, a new image should be built containing the latest code from both the project and its included builds.  

**Steps to reproduce**:
Assume a composite Gradle build where an app's `build.gradle` declares a dependency on a library using

```
    implementation("foo:some-lib:1.0.0") { changing = true }
```

and its `settings.gradle` contains

```
includeBuild '../some-lib'
```

1. If I intentionally introduce a compile-time error in `some-lib` and run `gradle jibDockerBuild` on the app, the build fails, so I know the right library code is being compiled.
1. If I make a non-breaking change in `some-lib`, `jibDockerBuild` succeeds but the Docker image it constructs is identical to the one before the change, it just has a new `LastTagTime`. 
1. If I make a change to both the app and to the library and run `gradle jibDockerBuild`, a new image is built, but it still executes old `some-lib` code.
1. If I open a shell, I can see `some-lib.jar` in `/apps/libs`, which isn't getting updated.
1. If I run `gradle build` before `gradle jibDockerBuild`, the library is updated.

**Environment**: macOS, JDK8, Gradle

**`jib-gradle-plugin` Configuration**: 
```groovy
jib {
    from {
        image = "gcr.io/distroless/java:debug"
    }
    to {
        image = "foo/bar"
    }
    container {
        jvmFlags = ["-Duser.timezone=UTC"]
        ports = ["80"]
    }
}
```

loosebazooka:
Any chance you give provide a simple project on which to test this out?

loosebazooka:
I suspect that you could not be triggering a new build of `some-lib` somehow in your gradle configuration. Jib is not aware of what's being built underneath and will just work with what it's given from gradle.

Usually for multiproject builds, I expect to see the `compile project('some-lib')`, what's the reason you're using the composite build syntax?

emersonf:
[...]

PR #1754, opened 2019-05-31
Resolve input files more comprehensively on jibBuildTar

loosebazooka:
We haven't been including the all the input to the task correct (this only matter with jibBuildTar which has an `@Output` defined - all other jib tasks run always).

Also we had some weird code (BuildTarTask:83) that is cleaned up here.

Bug: https://github.com/GoogleContainerTools/jib/issues/1745#issuecomment-497333324

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
@@ -73,8 +73,8 @@ public class BuildTarTask extends DefaultTask implements JibTask {
   }
 
   /**
-   * @return the input files to this task are all the output files for all the dependencies of the
-   *     {@code classes} task.
+   * @return a collection of all the files that jib includes in the image. Only used to calculate
+   *     UP-TO-DATE.
    */
   @InputFiles
   public FileCollection getInputFiles() {
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
@@ -80,11 +80,7 @@ public class BuildTarTask extends DefaultTask implements JibTask {
   public FileCollection getInputFiles() {
     List<Path> extraDirectories =
         Preconditions.checkNotNull(jibExtension).getExtraDirectories().getPaths();
-    return extraDirectories
-        .stream()
-        .map(Path::toFile)
-        .map(directory -> GradleProjectProperties.getInputFiles(directory, getProject()))
-        .reduce(getProject().files(), getProject()::files);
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -80,11 +80,7 @@ public void setTargetImage(String targetImage) {
   public FileCollection getInputFiles() {
```
briandealwis: The javadoc makes it sound like only the classes dependencies matter, and there's no mention of the extras.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -270,25 +269,26 @@ public boolean isWarProject() {
   /**
    * Returns the input files for a task.
    *
-   * @param extraDirectory the image's configured extra directory
    * @param project the gradle project
+   * @param extraDirectories the image's configured extra directories
    * @return the input files to the task are all the output files for all the dependencies of the
    *     {@code classes} task
```
briandealwis: No mention of the extras?

loosebazooka:
>Why do we treat `BuildDockerTask` and `BuildImageTask` differently: why no `@Input` nor `@Output`? Because they don't have a formal output?

Yes in gradle, unless you have an `Output` defined, it does not check if things are up to date or not. I feel like `BuildDockerTask` should also have an output so we don't rebuild a whole image when we don't need to (Is that how it works?)?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
loosebazooka: fixed.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
loosebazooka: fixed.

chanseokoh:
> it does not check if things are up to date or not. I feel like BuildDockerTask should also have an output so we don't rebuild a whole image

Isn't it basically that we should always push an image to a Docker daemon regardless of whether there is a change to input? Because often you do `docker rmi` to remove images and you expect `jib:dockerBuild` to push an image again if it's missing. I *think* performance is not a big deal, since we cache base and application layers.

loosebazooka:
> Because often you do docker rmi to remove images and you expect jib:dockerBuild to push an image again if it's missing. 

This is a good point, I didn't think of it.

TadCordle:
Will this fix https://github.com/GoogleContainerTools/jib/issues/1757?

PR #1758, opened 2019-06-03
Add internal Skaffold-init task/goal

TadCordle:
Adds task/goal for skaffold-init usage that prints the `to.image` configuration and project name to be consumed by skaffold.

e.g.

```
BEGIN JIB
{"image":"gcr.io/image/name","project":"project-name"}
```

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
@@ -44,7 +44,7 @@ public class SkaffoldInitTask extends DefaultTask {
     if (!getProject().equals(getProject().getRootProject())) {
       skaffoldInitOutput.setProject(getProject().getName());
     }
-    System.out.println("\nBEGIN JIB");
+    System.out.println("\nBEGIN JIB JSON");
     System.out.println(skaffoldInitOutput.getJsonString());
   }
 }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
@@ -21,10 +21,14 @@ import javax.annotation.Nullable;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.tasks.TaskAction;
 
+/**
+ * Prints out to.image configuration and project name, used for Jib project detection in Skaffold.
+ *
+ * <p>Expected use: {@code ./gradlew _jibSkaffoldInit -q}
+ */
 public class SkaffoldInitTask extends DefaultTask {
 
-  @Nullable
-  private JibExtension jibExtension;
+  @Nullable private JibExtension jibExtension;
 
   public SkaffoldInitTask setJibExtension(JibExtension jibExtension) {
     this.jibExtension = jibExtension;
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
@@ -16,7 +16,9 @@
 
 package com.google.cloud.tools.jib.gradle;
 
+import com.google.cloud.tools.jib.plugins.common.SkaffoldInitOutput;
 import com.google.common.base.Preconditions;
+import java.io.IOException;
 import javax.annotation.Nullable;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.tasks.TaskAction;
@@ -36,10 +38,13 @@ public class SkaffoldInitTask extends DefaultTask {
   }
 
   @TaskAction
-  public void listModulesAndTargets() {
-    String target = Preconditions.checkNotNull(jibExtension).getTo().getImage();
+  public void listModulesAndTargets() throws IOException {
+    SkaffoldInitOutput skaffoldInitOutput = new SkaffoldInitOutput();
+    skaffoldInitOutput.setImage(Preconditions.checkNotNull(jibExtension).getTo().getImage());
+    if (!getProject().equals(getProject().getRootProject())) {
+      skaffoldInitOutput.setProject(getProject().getName());
+    }
     System.out.println("\nBEGIN JIB");
-    System.out.println(target == null ? "?" : target);
-    System.out.println(getProject().getName());
+    System.out.println(skaffoldInitOutput.getJsonString());
   }
 }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2019 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


chanseokoh:
I wonder if it isn't easier to load a JSON in Go than parsing lines?

And do you plan to call this in every submodule? Didn't we say we call this at the project root, working like the dependency list goal/task?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/SkaffoldInitMojo.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Prints out to.image configuration and project name, used for Jib project detection in Skaffold.
+ *
+ * <p>Expected use: {@code ./mvnw jib:_skaffold-init -q}
+ */
+@Mojo(name = SkaffoldInitMojo.GOAL_NAME, requiresDependencyCollection = ResolutionScope.NONE)
+public class SkaffoldInitMojo extends JibPluginConfiguration {
+
+  @VisibleForTesting static final String GOAL_NAME = "_skaffold-init";
```
chanseokoh: No test committed to the PR.

TadCordle:
From my testing, if it's called on the project root, it will automatically run on all sub-projects.

As for loading JSON vs. parsing lines, this is probably simpler. JSON approach would require an additional struct to unmarshal the json into, whereas this is just a couple regex calls.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/SkaffoldInitMojo.java
TadCordle: Whoops, copy-paste.

briandealwis:
I'm not a fan of the name as it doesn't actually do any initialization for Skaffold.  _Analyze_ or _report_?

I think this plugin needs to run as an aggregator and output JSON: parsing the result for multiple projects looks a bit error-prone.
```sh
$ cd examples/multi-module
$ mvn -q com.google.cloud.tools:jib-maven-plugin:1.2.1-SNAPSHOT:_skaffold-init

BEGIN JIB
gcr.io/${env.PROJECT_ID}/jib-multimodule:0.1.0
Jib Multi-module Example

BEGIN JIB
gcr.io/${env.PROJECT_ID}/name-service:0.1.0
name-service

BEGIN JIB
gcr.io/${env.PROJECT_ID}/hello-service:0.1.0
hello-service
```

TadCordle:
Is there a gradle equivalent for "aggregator"?

TadCordle:
I guess JSON would be more useful for cases like where `to.image` isn't configured, or if the module name isn't needed for the root project, so those fields can just be omitted instead of using special values.

TadCordle:
I'd actually prefer this name since it's clear at a glance that it's used for skaffold init. But if you or others feel strongly about it, we can change it.

chanseokoh:
Oh, that being said, why isn't the Mojo in the `.maven` package?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SkaffoldInitMojo.java
```diff
@@ -0,0 +1,51 @@
...
+    skaffoldInitOutput.setImage(getTargetImage());
+    if (getProject().getParent() != null) {
+      skaffoldInitOutput.setProject(getProject().getName());
+    }
+
+    System.out.println("\nBEGIN JIB");
+    try {
+      System.out.println(skaffoldInitOutput.getJsonString());
+    } catch (IOException ex) {
+      throw new MojoExecutionException(ex.getMessage());
```
chanseokoh: Wrap the cause too: `MojoExecutionException(message, ex)`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SkaffoldInitMojo.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Prints out to.image configuration and project name, used for Jib project detection in Skaffold.
+ *
+ * <p>Expected use: {@code ./mvnw jib:_skaffold-init -q}
+ */
+@Mojo(name = SkaffoldInitMojo.GOAL_NAME, requiresDependencyCollection = ResolutionScope.NONE)
+public class SkaffoldInitMojo extends JibPluginConfiguration {
+
+  static final String GOAL_NAME = "_skaffold-init";
```
chanseokoh: nit: `private`

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldInitOutput.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class SkaffoldInitOutput {
+
+  @JsonInclude(Include.NON_NULL)
+  @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+  private static class SkaffoldInitTemplate {
+
+    @Nullable private String image;
+
+    @Nullable private String project;
```
chanseokoh: I know this is the same pattern in `SkaffoldFilesOutput`, but I wonder if `SkaffoldInitOutput` can have these fields directly. Why do we have to declare another class for the additional wrapping? Is jackson not able to take care of it?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public SkaffoldInitTask setJibExtension(JibExtension jibExtension) {
+    this.jibExtension = jibExtension;
+    return this;
+  }
+
+  @TaskAction
+  public void listModulesAndTargets() throws IOException {
+    SkaffoldInitOutput skaffoldInitOutput = new SkaffoldInitOutput();
+    skaffoldInitOutput.setImage(Preconditions.checkNotNull(jibExtension).getTo().getImage());
```
chanseokoh: If `getImage()` returns null, what happens to the JSON? (No idea if "null" is a built-in type in JSON or how it will be represented.) Will that be handled properly on the Skaffold side?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SkaffoldInitMojo.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  public void execute() throws MojoExecutionException {
+    SkaffoldInitOutput skaffoldInitOutput = new SkaffoldInitOutput();
+    skaffoldInitOutput.setImage(getTargetImage());
+    if (getProject().getParent() != null) {
+      skaffoldInitOutput.setProject(getProject().getName());
```
chanseokoh: Likewise, you won't set the `project` field for root projects, meaning the field will be `null`. Will that be handled properly?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
TadCordle: If the field is null, it'll be omitted from the JSON output.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SkaffoldInitMojo.java
TadCordle: I can check again tomorrow, but I'm pretty sure this can't be private, or it can't be referenced in the `@Mojo` annotation. Maybe it can just be hard-coded instead of using a field.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/SkaffoldInitTask.java
loosebazooka: would like to see a test for this to document our assumptions and expectations. Tests in general?

PR #1759, opened 2019-06-03
Avoid exposing EventHandlers in public API

TadCordle:
Towards #1428.

- Moves `JibEvent` and `LogEvent` to api package
- Removes `setEventHandlers()` from `Containerizer` and replaces it with `addEventHandler()`
- Various changes in plugins as a result


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -227,8 +202,27 @@ public void waitForLoggingThread() {
...
 
   @Override
-  public EventHandlers getEventHandlers() {
-    return eventHandlers;
+  public void applyEventHandlers(Containerizer containerizer) {
+    containerizer
+        .addEventHandler(LogEvent.class, this::log)
+        .addEventHandler(
+            TimerEvent.class,
+            new TimerEventHandler(message -> consoleLogger.log(LogEvent.Level.DEBUG, message)))
```
chanseokoh: This is a minor point, but WDYT of making `ConsoleLogger::log` accept `LogEvent`? That is, `consoleLogger.log(LogEvent.debug(message))`.

TadCordle: It doesn't seem much neater to me than `consoleLogger.log(Level.DEBUG, message)`, so I think I'll leave it as is for now so the PR doesn't get too muddy. Maybe in a follow up if we want.

jib-core/CHANGELOG.md
```diff
@@ -5,10 +5,13 @@ All notable changes to this project will be documented in this file.
...
 ### Added
 
+- `Containerizer#addEventHandler` for adding event handlers
+
 ### Changed
 
 - Multiple classes have been moved to the `com.google.cloud.tools.jib.api` package
-- Removed `EventDispatcher` and `DefaultEventDispatcher`; events are now dispatched directly from `EventHandlers`
+- Event handlers are now added directly to the `Containerizer` rather than adding them to an `EventHandlers` object first
+- Removed multiple classes to simplify the event system (`JibEventType`, `BuildStepType`, `EventDispatcher`, `DefaultEventDispatcher`, `LayerCountEvent`)
```
chanseokoh: Thanks for document these!

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -109,9 +109,7 @@
    * {@code project-name:project-version} if target image is not configured
    *
    * @param targetImage the configured target image reference
-   * @param eventHandlers the {@link EventHandlers} used to dispatch log events
-   * @param generatedName the image name to use if {@code targetImage} is {@code null}
-   * @param generatedTag the tag to use if {@code targetImage} is {@code null}
+   * @param projectProperties the {@link ProjectProperties} used to handle the log events
```
chanseokoh: nit: more than log events, since now it's used to populate other info.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -217,13 +218,32 @@ public Containerizer setApplicationLayersCache(Path cacheDirectory) {
...
-   * @param eventHandlers the {@link EventHandlers}
+   * <p><b>Note: Implementations of {@code eventConsumer} must be thread-safe.</b>
+   *
+   * @param eventType the event type that {@code eventConsumer} should handle
+   * @param eventConsumer the event handler
+   * @param <E> the type of {@code eventClass}
+   * @return this
+   */
+  public <E extends JibEvent> Containerizer addEventHandler(
+      Class<E> eventType, Consumer<E> eventConsumer) {
```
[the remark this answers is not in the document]
TadCordle: Good catch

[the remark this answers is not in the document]
chanseokoh: Hold on for this a moment. I remember there was some issue around that. I'll take a look again later.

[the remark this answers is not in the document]
chanseokoh: Never mind.

Unrelated, since `EventHandlers` is internal, how about removing `EventHandler.add(Consumer<JibEvent>)` that is more like a helper method? It's enough to have `add(eventType, eventConsumer)`.

PR #1771, opened 2019-06-10
Overhaul async execution infrastructure in Jib

chanseokoh:
Overhauls the async execution infrastructure as proposed in the design doc.

This is a fairly significant revamp that fundamentally changes the infrastructure for concurrency in Jib, so I've added extensive comments to explain important bits to help reviewers understand the changes and confirm the correctness.

**Note: `StepsRunner` is the class that plays a critical role, but due to the length in changes, GitHub collapses the file by default. Make sure to expand `StepsRunner`.**

This PR almost completes the overhaul, but I shaped it in the way to reduce code diff for easier one-to-one comparison. I do need a follow-up PR for cleanup to move methods and classes.

There may still exist a few minor things that I need to sort out, but this is in good shape for initial review.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
```diff
@@ -18,91 +18,48 @@
...
   }
 
   @Override
-  public BuildResult call() throws ExecutionException, InterruptedException {
-    return AsyncDependencies.using(listeningExecutorService)
-        .addSteps(NonBlockingSteps.get(pullAndCacheBaseImageLayersStep))
-        .addSteps(buildAndCacheApplicationLayerSteps)
-        .addStep(NonBlockingSteps.get(buildImageStep))
-        .whenAllSucceed(this::afterPushBaseImageLayerFuturesFuture)
-        .get();
```
chanseokoh: Confirming: `call()` can run only after `pullAndCacheBaseImageLayersStep` and `buildImageStep` is complete. That is, sub-steps to pull base image layers are scheduled and the target `Image` instance is (being) built.

Actually, `pullAndCacheBaseImageLayersStep` is not necessary because `buildImageStep` can run only after all the base image layers are downloaded and cached.

Then, the core logic (`afterPushBaseImageLayerFuturesFuture`) can run only after all the base + application layers are cached and available and the target `Image` is conclusively built and available.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
```diff
@@ -44,10 +41,7 @@
   private final DescriptorDigest layerDigest;
   private final @Nullable Authorization pullAuthorization;
 
-  private final ListenableFuture<CachedLayer> listenableFuture;
-
   PullAndCacheBaseImageLayerStep(
-      ListeningExecutorService listeningExecutorService,
       BuildConfiguration buildConfiguration,
       ProgressEventDispatcher.Factory progressEventDispatcherFactory,
       DescriptorDigest layerDigest,
```
chanseokoh: Note to reviewers: notably, this step is already in the pattern of the new async redesign. This step downloads and cached the base image layer denoted by the given layer digest. No future or async business. The pull Authorization is given as a concrete value too.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayersStep.java
```diff
@@ -16,81 +16,44 @@
...
-      PullBaseImageStep pullBaseImageStep) {
-    this.listeningExecutorService = listeningExecutorService;
-    this.buildConfiguration = buildConfiguration;
-    this.progressEventDispatcherFactory = progressEventDispatcherFactory;
-    this.pullBaseImageStep = pullBaseImageStep;
-
-    listenableFuture =
-        AsyncDependencies.using(listeningExecutorService)
-            .addStep(pullBaseImageStep)
-            .whenAllSucceed(this);
```
chanseokoh: Confirming: the core logic can start only after a base `Image` instance is created.

Knowing that, I put the core logic of "...LayersStep" (note the plural) into the "makeList()" static method instead of "call()", following the same pattern used in other steps that spawns sub-tasks: `makeList()` dynamically creates sub-tasks to pull each layer, and `call()` has the logic to download a layer. (To reduce code diff, I defer merging this class into `PullAndCacheBaseImageLayerStep` (note the singular) will be done in a follow-up.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -148,23 +132,20 @@ public BaseImageWithAuthorization call()
...
-        ListeningExecutorService directExecutorService = MoreExecutors.newDirectExecutorService();
-        RetrieveRegistryCredentialsStep retrieveBaseRegistryCredentialsStep =
+        // registry.credentials.RegistryCredentialsRetriever.
+        Credential registryCredential =
             RetrieveRegistryCredentialsStep.forBaseImage(
-                directExecutorService,
-                buildConfiguration,
-                progressEventDispatcher.newChildProducer());
+                    buildConfiguration, progressEventDispatcher.newChildProducer())
+                .call();
```
chanseokoh: Note to reviewers: now we can directly run ".call()", without futzing with an executor service and `NonBlockingSteps.get()`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushBlobStep.java
```diff
@@ -18,62 +18,44 @@
...
     this.progressEventDipatcherFactory = progressEventDipatcherFactory;
-    this.authenticatePushStep = authenticatePushStep;
+    this.authorization = authorization;
     this.blobDescriptor = blobDescriptor;
     this.blob = blob;
-
-    listenableFuture =
-        AsyncDependencies.using(listeningExecutorService)
-            .addStep(authenticatePushStep)
-            .whenAllSucceed(this);
```
chanseokoh: Confirming: this step can run only after an authorization is available.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushContainerConfigurationStep.java
```diff
@@ -16,90 +16,58 @@
...
-  }
-
-  @Override
-  public AsyncStep<PushBlobStep> call() throws ExecutionException {
-    ListenableFuture<PushBlobStep> pushBlobStepFuture =
-        AsyncDependencies.using(listeningExecutorService)
-            .addStep(authenticatePushStep)
-            .addStep(NonBlockingSteps.get(buildImageStep))
-            .whenAllSucceed(this::afterBuildConfigurationFutureFuture);
-    return () -> pushBlobStepFuture;
```
chanseokoh: Confirming: this core logic can run only after all of the followings are ready:
1. push Authorization
1. the target `Image` instance built

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushContainerConfigurationStep.java
```diff
@@ -16,90 +16,58 @@
...
-          progressEventDispatcher.newChildProducer(),
-          authenticatePushStep,
-          blobDescriptor,
-          Blobs.from(containerConfiguration));
+              buildConfiguration,
+              progressEventDispatcher.newChildProducer(),
+              pushAuthorization,
+              Digests.computeDigest(containerConfiguration),
+              Blobs.from(containerConfiguration))
+          .call();
```
chanseokoh: Note to reviewers: we can just directly call the core to upload a blob.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
```diff
@@ -40,84 +37,37 @@
...
-    }
-    for (AsyncStep<PushBlobStep> pushApplicationLayerStep :
-        NonBlockingSteps.get(pushApplicationLayersStep)) {
-      dependencies.addStep(NonBlockingSteps.get(pushApplicationLayerStep));
-    }
-    return dependencies
-        .addStep(NonBlockingSteps.get(NonBlockingSteps.get(pushContainerConfigurationStep)))
-        .whenAllSucceed(this::afterAllPushed)
-        .get();
-  }
```
chanseokoh: Confirming: the core logic (`afterAllPushed`) can run only after everything is concretely available:
1. push Authorization
1. base image layers downloaded and cached
1. application layers built and cached
1. target `Image` instance built
1. container configuration pushed into the target registry

Note that building a target `Image` instance already requires that all the layers are available.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayerStep.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.cloud.tools.jib.http.Authorization;
+import com.google.common.collect.ImmutableList;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+
+class PushLayerStep implements Callable<BlobDescriptor> {
```
chanseokoh: Note to reviewers: this is renamed from `PushLayersStep` (note the plural), but GitHub fails to recognize it.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayerStep.java
```diff
@@ -0,0 +1,91 @@
...
+  private final ProgressEventDispatcher.Factory progressEventDispatcherFactory;
+
+  private final Authorization pushAuthorization;
+  private final Future<CachedLayerAndName> cachedLayerAndName;
+
+  static ImmutableList<PushLayerStep> makeList(
+      BuildConfiguration buildConfiguration,
+      ProgressEventDispatcher.Factory progressEventDispatcherFactory,
+      Authorization pushAuthorization,
+      List<Future<CachedLayerAndName>> cachedLayers) {
```
chanseokoh: Note to reviewers: this is the only exception where a step gets a "future" value. This is to allow pushing layers even if Jib is still downloading other layers. In any case, the class is greatly simplified, and no concurrency-related scheduling is being done inside a step. The pattern to create sub-tasks is same as pulling base image layers and building application layers.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayerStep.java
```diff
@@ -0,0 +1,91 @@
...
+    this.buildConfiguration = buildConfiguration;
+    this.progressEventDispatcherFactory = progressEventDispatcherFactory;
+    this.pushAuthorization = pushAuthorization;
+    this.cachedLayerAndName = cachedLayerAndName;
+  }
+
+  @Override
+  public BlobDescriptor call()
+      throws IOException, RegistryException, ExecutionException, InterruptedException {
+    CachedLayer layer = cachedLayerAndName.get().getCachedLayer();
```
chanseokoh: Note to reviewers: this is the blocking point for the step, and the only place for blocking in any step among all.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayerStep.java
```diff
@@ -0,0 +1,91 @@
...
+    return new PushBlobStep(
+            buildConfiguration,
+            progressEventDispatcherFactory,
+            pushAuthorization,
+            new BlobDescriptor(layer.getSize(), layer.getDigest()),
+            layer.getBlob())
+        .call();
```
chanseokoh: Note to reviewers: calling the core push logic directly.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -40,23 +50,24 @@
  */
 public class StepsRunner {
```
chanseokoh: Note to reviewers: this is the central and the only place for async scheduling and dependent input management in the new design.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -40,23 +50,24 @@
...
 
-    @Nullable private AsyncStep<BuildResult> finalStep;
+    private Future<ImageAndAuthorization> baseImageAndAuth = failedFuture();
+    private Future<List<Future<CachedLayerAndName>>> baseImageLayers = failedFuture();
+    private List<Future<CachedLayerAndName>> applicationLayers = new ArrayList<>();
+    private Future<Image> builtImage = failedFuture();
+    private Future<Credential> targetRegistryCredentials = failedFuture();
+    private Future<Authorization> pushAuthorization = failedFuture();
+    private Future<List<Future<BlobDescriptor>>> baseImageLayerPushResults = failedFuture();
+    private Future<List<Future<BlobDescriptor>>> applicationLayerPushResults = failedFuture();
```
chanseokoh: Note to reviewers: layers push results (digest and size of pushed layers) are not actually used in the both old and new designs. They are merely for blocking the final registry push step until we are certain that the layers are all pushed.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
```diff
@@ -36,75 +30,42 @@
...
   }
 
   @Override
-  public BuildResult call() throws ExecutionException, InterruptedException {
-    return AsyncDependencies.using(listeningExecutorService)
-        .addSteps(NonBlockingSteps.get(pullAndCacheBaseImageLayersStep))
-        .addSteps(buildAndCacheApplicationLayerSteps)
-        .addStep(NonBlockingSteps.get(buildImageStep))
-        .whenAllSucceed(this::writeTarFile)
-        .get();
```
chanseokoh: Confirming: the core logic can start only after
1. the base image layers are downloaded and cached;
1. the application layers are built and cached; and
1. the target `Image` instance is created and available.

Note building a target `Image` already requires 1 and 2.

chanseokoh:
Overhauls the async execution infrastructure as proposed in the design doc.

This is a fairly significant revamp that fundamentally changes the infrastructure for concurrency in Jib, so I've added extensive comments to explain important bits to help reviewers understand the changes and confirm the correctness.

**Note: StepsRunner is the class that plays a critical role, but due to the length in changes, GitHub collapses the file by default. There are other files collapses by default as they were removed, so make sure to expand StepsRunner.**

This PR almost completes the overhaul, but I shaped it in the way to reduce code diff for easier one-to-one comparison. I do need a follow-up PR for cleanup to move methods and classes.

There may exist a few minor things that I need to sort out, but this is in good shape for initial review.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
loosebazooka: This call of `RetrieveRegistryCredentialsStep` seems out of place. Should it be an actual step in our `buildSteps` in the same way `retrieveTargetRegistryCredentials` generates a step in our build?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushContainerConfigurationStep.java
loosebazooka: If PushBlob isn't being thrown to the executor, is it still a step?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayerStep.java
loosebazooka: I'm not sure I understand the scope of this statement? Is this the only place we call `get()` and expect it to wait?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -74,197 +85,222 @@ public static StepsRunner begin(BuildConfiguration buildConfiguration) {
...
+  private void enqueueBuildAndCache() {
+    stepsToRun.add(this::pullBaseImage);
+    stepsToRun.add(this::pullAndCacheBaseImageLayers);
+    stepsToRun.add(this::buildAndCacheApplicationLayers);
+    stepsToRun.add(this::buildImage);
+  }
+
+  public StepsRunner dockerLoadSteps(DockerClient dockerClient) {
+    rootProgressDescription = "building image to Docker daemon";
+    enqueueBuildAndCache();
```
TadCordle: Feel free to ignore this since it might just be personal preference, but I feel like inlining this and copy-pasting the `add` steps between each `...Steps()` method may be easier to read at a glance, even though it's more duplicate code.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: This is related to my comment below about renaming these classes (and potentially convert `call()` to a static method). For example, if regard this "Step" as a task and assume it had a method
```
getCredentials(buildConfiguration, ...);
```
instead of
```
forBaseImage(buildConfiguration, ...).call();
```
this wouldn't feel weird. As I said in the other comment, we should definitely think about renaming these classes and methods.

> Should it be an actual step in our

For this, I think this is OK. The "steps" enqueued `StepsRunner` are the unit of works that we think worth parallelization. At least I am not changing the current behavior in that the `PullBaseImageStep` was getting credentials serially (using a direct executor). That looks OK, because getting a credential is fast enough, all happening locally (e.g., reading `~/.docker/config.json`, ...).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/CachedLayerAndName.java
```diff
@@ -0,0 +1,41 @@
...
+package com.google.cloud.tools.jib.builder.steps;
+
+import com.google.cloud.tools.jib.cache.CachedLayer;
+import javax.annotation.Nullable;
+
+/** Simple structure to hold a pair of {#link CachedLayer} and its string name. */
+class CachedLayerAndName {
+
+  private CachedLayer cachedLayer;
+  @Nullable private String name;
```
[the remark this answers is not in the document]
chanseokoh: I also wanted that badly. But when we save or load `CachedLayer`, the name is irrelevant and gets lost, so I avoided putting a field there to make it clear that caching a layer does not involve any naming.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
```diff
@@ -145,7 +145,7 @@ public void testOffline()
     } catch (ExecutionException ex) {
       Assert.assertEquals(
           "Cannot run Jib in offline mode; localhost:5001/busybox not found in local Jib cache",
-          ex.getCause().getMessage());
+          ex.getCause().getCause().getCause().getMessage());
```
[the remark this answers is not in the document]
chanseokoh: We can unroll `ExecutionException`; I just did at the plugin level: https://github.com/GoogleContainerTools/jib/pull/1771/files#diff-c56a8fadef6560c72d23356bd2804d5f

```java
      Throwable throwable = ex;
      while (throwable instanceof ExecutionException && throwable.getCause() != null) {
        throwable = throwable.getCause();
      }
```

Maybe I should do it at the library level.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/CachedLayerAndName.java
```diff
@@ -0,0 +1,41 @@
...
+  CachedLayerAndName(CachedLayer cachedLayer, @Nullable String name) {
+    this.cachedLayer = cachedLayer;
+    this.name = name;
+  }
+
+  CachedLayer getCachedLayer() {
+    return cachedLayer;
+  }
+
+  @Nullable
```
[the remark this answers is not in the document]
loosebazooka: I think if it's not escaping into the API surface, this might be okay? Since error prone's job is to work off these annotations? But have we generally moved off `@Nullable` in our code?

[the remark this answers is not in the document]
chanseokoh: I think I'm tolerant with that for this type of class that is awkwardly required in Java in a verbose way just to return multiple values from a method. I even wish I don't define getters.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: SGTM. Looks better with comments.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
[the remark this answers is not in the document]
TadCordle: Yeah, if the intermediary causes aren't important then I think it'd be nice at the library level.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -40,23 +50,24 @@
...
-    @Nullable private PushContainerConfigurationStep pushContainerConfigurationStep;
+    private static <E> Future<E> failedFuture() {
+      return Futures.immediateFailedFuture(
+          new IllegalStateException("invalid usage; required step not configured"));
+    }
 
-    @Nullable private AsyncStep<BuildResult> finalStep;
+    private Future<ImageAndAuthorization> baseImageAndAuth = failedFuture();
+    private Future<List<Future<CachedLayerAndName>>> baseImageLayers = failedFuture();
+    private List<Future<CachedLayerAndName>> applicationLayers = new ArrayList<>();
```
loosebazooka: why is this not a `failedFuture()` ?

chanseokoh: About why it is not a `Future` at the top level: in both of the original and new designs, you already know what (and how many) application layers you should build at the beginning, so it is different from `baseImageLayers` where you don't know what and how many layers to download.

I think I'll just make it `@Nullable` and do `checkNotNull` to guard against a wrong schedule order as in the original design. (To be frank, I instantiated only to avoid having to use `checkNotNull`, but on second thought, it's better to explicitly guard wrong usage.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
```diff
@@ -40,84 +39,37 @@
 import java.util.concurrent.ExecutionException;
 
 /** Pushes the final image. Outputs the pushed image digest. */
-class PushImageStep implements AsyncStep<BuildResult>, Callable<BuildResult> {
+class PushImageStep implements Callable<BuildResult> {
```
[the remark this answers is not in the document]
chanseokoh: I agree. I have a follow-up PR that actually restructure this class to remove `listeningExecutorService`, so I'll do this simple renaming in the next follow-up.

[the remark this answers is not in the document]
chanseokoh: https://github.com/GoogleContainerTools/jib/blob/async-restructure-followup/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushManifestStep.java#L38-L42 (PR not filed yet)

loosebazooka:
LGTM, can you schedule a bug out/stress test on this? I'd love for us to build some silly complex projects and test this out.

chanseokoh:
> can you schedule a bug out/stress test on this?

Good idea. I have two follow-up PRs (one for clean-up, and the other for optional optimization which might not get landed), so I'll schedule it once they are cleared.

PR #1779, opened 2019-06-17
Update README and FAQ about containerizing mode

chanseokoh:
Should be merged after releasing 1.4.0.


docs/faq.md
```diff
@@ -161,7 +162,17 @@ Normally, the plugin sets a default entrypoint for java applications, or lets yo
...
+Currently Jib does not natively support creating an image that runs a runnable JAR (or WAR) through `java -jar runnable.jar` (although it is not impossible to configure Jib to do so at the expense of more complex project setup.) The intention of Jib is to add individual class files, resources, and dependency JARs into the container instead of putting a runnable JAR. This lets Jib choose an opinionated, optimal layout for the application on the container image, which also allows it to skip the extra JAR-packaging step.
+
+However, you can set `<containerizingMode>packaged` (Maven) or `containerizingMode = 'packaged'` (Gradle) to containerize a JAR, but note that your application will always be run via `java -cp your.MainClass` (even if it is an executable JAR). Some disadvnatags:
```
PascalSchumacher: Should be `disadvantages` I guess.

chanseokoh: Oops, thanks!

docs/faq.md
```diff
@@ -161,7 +162,17 @@ Normally, the plugin sets a default entrypoint for java applications, or lets yo
...
+Currently Jib does not natively support creating an image that runs a runnable JAR (or WAR) through `java -jar runnable.jar` (although it is not impossible to configure Jib to do so at the expense of more complex project setup.) The intention of Jib is to add individual class files, resources, and dependency JARs into the container instead of putting a runnable JAR. This lets Jib choose an opinionated, optimal layout for the application on the container image, which also allows it to skip the extra JAR-packaging step.
+
+However, you can set `<containerizingMode>packaged` (Maven) or `containerizingMode = 'packaged'` (Gradle) to containerize a JAR, but note that your application will always be run via `java -cp your.MainClass` (even if it is an executable JAR). Some disadvantages:
+
+- You need to run the JAR-packaging step (`mvn package` in Maven or the `jar` task in Gradle).
+- Reduced granularity in building and caching: if any of your Java source files or resource files are updated, not only the JAR has to be rebuilt but the layer containing the JAR in the image has to be recreated and pushed to a registry or a Docker daemon as a whole as a new layer.
```
TadCordle: Maybe something like:

```suggestion
- Reduced granularity in building and caching: if any of your Java source files or resource files are updated, not only the JAR has to be rebuilt, but the entire layer containing the JAR in the image has to be recreated and pushed to the destination.
```

docs/faq.md
```diff
@@ -161,7 +162,17 @@ Normally, the plugin sets a default entrypoint for java applications, or lets yo
...
+- If it is a fat or shaded JAR embedding all dependency JARs, you are duplicating the dependency JARs in the image. Worse, it results in far more reduced granularity in building and caching, as dependency JARs can be huge and all of them need to be pushed repeatedly even if they do not change.
+
+### I want to containerize a JAR.
```
TadCordle: Is there a reason for adding this? It seems redundant.

jib-maven-plugin/README.md
```diff
@@ -221,6 +221,7 @@ Field | Type | Default | Description
 `from` | [`from`](#from-object) | See [`from`](#from-object) | Configures the base image to build your application on top of.
 `container` | [`container`](#container-object) | See [`container`](#container-object) | Configures the container that is run from your image.
 `extraDirectories` | [`extraDirectories`](#extradirectories-object) | See [`extraDirectories`](#extradirectories-object) | Configures the directories used to add arbitrary files to the image.
+`containerizingMode` | `String` | `exploded` | If set to `packaged`, puts the JAR artifact built at `${project.build.directory}/${project.build.finalName}.jar` (the default location where many JAR-buidiling plugins put a JAR registered as a main artifact, such as the Maven JAR Plugin) into the final image. If set to `exploded` (default), containerizes individual `.class` files and resources files.
```
TadCordle: nit: for the Maven readme, we've been labeling types like "string" instead of "`String`". Maybe we want to change the whole thing to be more like the gradle readme instead?

jib-gradle-plugin/README.md
```diff
@@ -176,6 +176,7 @@ Field | Type | Default | Description
 `from` | [`from`](#from-closure) | See [`from`](#from-closure) | Configures the base image to build your application on top of.
 `container` | [`container`](#container-closure) | See [`container`](#container-closure) | Configures the container that is run from your built image.
 `extraDirectories` | [`extraDirectories`](#extradirectories-closure) | See [`extraDirectories`](#extradirectories-closure) | Configures the directories used to add arbitrary files to the image.
+`containerizingMode` | `String` | `exploded` | If set to `packaged`, puts the JAR artifact built by the Gradle Java plugin into the final image. If set to `exploded` (default), containerizes individual `.class` files and resources files.
```
TadCordle: Should it be "Gradle Jar plugin" not "Gradle Java plugin"?

chanseokoh: Not totally sure, but I thought the "Java Plugin" just provides the `jar` task. https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_tasks

docs/faq.md
chanseokoh: Is there a way to change the title without changing the current link (`https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#i-want-to-containerize-an-executable-jar`)? I wanted to make the old link continue to work.

jib-maven-plugin/README.md
chanseokoh: Ah, I'll fix this. Things in XML are not Java types, so that's probably why we just say "string".

docs/faq.md
TadCordle: I think you could try using just a normal anchor? We do something like that in the readmes.

`<a name="i-want-to-containerize-an-executable-jar"></a>### I want to containerize a JAR.`

jib-gradle-plugin/README.md
TadCordle: Ah ok, I wasn't sure.

docs/faq.md
chanseokoh: This works. Since I can change the title, I think I'll explain `<containerizingMode>` first and then explain runnable JARs later.

chanseokoh: Updated wordings and shuffled sentences. PTAL.

docs/faq.md
```diff
@@ -160,19 +159,17 @@ See [Extended Usage](../jib-gradle-plugin#extended-usage) for the `container.for
...
+The intention of Jib is to add individual class files, resources, and dependency JARs into the container instead of putting a JAR. This lets Jib choose an opinionated, optimal layout for the application on the container image, which also allows it to skip the extra JAR-packaging step.
 
-However, you can set `<containerizingMode>packaged` (Maven) or `containerizingMode = 'packaged'` (Gradle) to containerize a JAR, but note that your application will always be run via `java -cp your.MainClass` (even if it is an executable JAR). Some disadvantages:
+However, you can set `<containerizingMode>packaged` (Maven) or `containerizingMode = 'packaged'` (Gradle) to containerize a JAR, but note that your application will always be run via `java -cp ... your.MainClass` (even if it is an executable JAR). Some disadvantages:
```
TadCordle: `jib.containerizingMode = 'packaged'` for Gradle?

chanseokoh: Nice. Done.

Issue #1608, opened 2019-04-08
Allow setting file timestamps from plugins


TadCordle:
From #1079

ST-DDT:
Although https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified is not the best way to determine whether a resource has changed it is still a possible way to do so. Setting the file timestamps to zero/a fixed arbitrary value invalidates all results that would be generated by this mechanisms. Since this tool does not create a jar for the projects classes there aren't any other metadata to extract the age of the image from at all. Setting the file timestamp to the current timestamp is often better than setting it to zero. The following article describes the usage of the file timestamps with spring:
https://www.logicbig.com/tutorials/spring-framework/spring-web-mvc/last-modified.html

On a side node, in most bigger projects there are quite a few classes that are generated automatically and might contains comments such as `Generated on XXX` these would invalidate the reproducible builds anyway so we should change your policy from reproducible to predictable.

chanseokoh:
Thanks for the input, @ST-DDT. 

> Setting the file timestamp to the current timestamp is often better than setting it to zero.

Aside from the current timestamp, do you also see a need to set an arbitrary timestamp (for some files or all files)?

And note that, as soon as you assign the current timestamp to files (and not using a fixed timestamp), whenever you build a new image, the layers will always be considered different: https://github.com/GoogleContainerTools/jib/issues/1079#issuecomment-479220028

ST-DDT:
I would prefer the git commit timestamp as the timestamp for all files, but I can imagine that other users will use other SCM-tools with a similar timestamps. Setting each file to their individual last changed date (from scm) might be an alternative (to only rebuild the layers if they are acutally different), but this is probably too hard to implement. AFAICT I don't need the timestamps on `.jar`s (or `.class`es)

chanseokoh:
Sounds like you don't need a feature to explicitly configure file timestamps to a fixed value, for example, as long as Jib supports automatically setting timestamps of all the files in the image to the timestamp of the last git commit. Can you confirm that? We are trying to see if there actually exists a use case for configuring an explicit value to design an optimal way of file timestamp configuration.

ST-DDT:
If jib adds a "use git commit timestamp option" then I dont need the option to set it to parameter x.

 It just feels a little bit redundant to make jib read the git timestamp, if other plugins do that already and provide these values for other plugins. Also it would be less flexible then setting it to a parameter/option.

[...]

PR #1782, opened 2019-06-18
Allow setting file timestamps from plugins (#1608)

warrkan:
Addressed ticket #1608. Implemented possibility to set modification times for container files and extra dir files.
The modification time value can be:

- epoch_plus_second - current behavior, modification time will be trimmed to epoch + 1 second
- keep_original - original modification time will be kept
- date in ISO 8601 format, e.g 2019-06-18T12:00:00Z

Unfortunately I was not able to run all integration tests successfully because of verification errors in BuildImageMojoIntegrationTest.


jib-core/src/main/java/com/google/cloud/tools/jib/api/KeepOriginalModificationTimeProvider.java
```diff
@@ -0,0 +1,26 @@
+package com.google.cloud.tools.jib.api;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Instant;
+
+/** Modification time provider which returns original file modification time */
+public class KeepOriginalModificationTimeProvider implements ModificationTimeProvider {
```
chanseokoh: I'll need to decide whether we want to have this in jib-core. Maybe in jib-plugins-common. Particularly, the method throwing a runtime exception is not an ideal behavior as a library implementation.

warrkan:
Thanks @chanseokoh for quick response and a review itself. Yes, probably having possibility to specify modification times for separate files from extra dirs is over-engineering and should be simplified, there was no actual need in it. Anyway, will be waiting for further comment and updates.

chanseokoh:
Hi @warrkan, we had a discussion, and at least as an initial implementation, we think we should have a single config that uniformly applies to all the files including the extra directories. If we find a use case for the need to individually customizing mod times of extra directory files, we can implemented that later. For now, we don't see the usefulness of that, like you said you don't need it. So, let's take out anything related to the extra directory customization. Then you can start addressing my other comments.

warrkan:
@chanseokoh I addressed your comment about extra configuration. Travis build failed, but at first glance everything seems ok. Can you please take a look?

chanseokoh:
@warrkan thanks a lot!

Travis passed. (Someone restarted the build?) Anyways, I guess you can continue working on other open review comments? If you have any doubt, questions, or comment, feel free to update.

warrkan:
@chanseokoh I addressed remaining comments. The only open topic left (as I can see) is to move provider implementations out of jib-core.

chanseokoh:
Thanks @warrkan! We really appreciate it. This feature was one of the popular requests for Jib.

I ran full integration test internally and noticed that there are integration test failures, which will be difficult for you to debug (the full integration test makes use of pushing to an actual GCP project). And there are a few other things left to refactor and clean up, so we think it's better to merge this to a separate branch first. Then we can work on fixing the failures and more refactoring.

Merging this to the `i1608-warrkan-file-timestamps` branch now.

warrkan:
@chanseokoh great news! thanks for reviews and assisting. Hope to see it in master soon.

PR #1785, opened 2019-06-19
Separate out project dependencies, add integration test to verify

loosebazooka:
Uh so here's the PR with the weirdo test. Basically injecting a task to inspect the layer configuration state and outputting it so the test can pick it up and verify. I don't know how else to do it without mocking the world.


chanseokoh:
```
jib-gradle-plugin/src/test/resources/gradle/projects/multi-service/complex-service/libs/dependency-1.0.0.jar
jib-gradle-plugin/src/test/resources/gradle/projects/multi-service/complex-service/libs/dependencyX-1.0.0-SNAPSHOT.jar
```
There files are added accidentally, aren't they? (Note `multi-service`, not `all-local-multi-service`.)

loosebazooka:
> There files are added accidentally, aren't they? (Note multi-service, not all-local-multi-service.)
oh right yeah. I'll clean that up. I'm mostly curious what people think of this test.

chanseokoh:
How about these?

1. We already have a code that opens an image tar, goes through all the layers inside the tar (the code assumes all `.tar.gz` files in the image tar are layer tars), opens the embedded layer tars, and lists all the files in each layer:
   https://github.com/GoogleContainerTools/jib/blob/v1.3.0-maven/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java#L137
   We won't be able to explicitly verify the layer type this way, but I think it is enough to check the order of dependency layers.

1. Or add some code somewhere to log files in each layer as DEBUG.

loosebazooka:
What about a task `_jibDebugShowBuildConfiguration` or something instead of (2).

chanseokoh:
I actually prefer (1). It doesn't look crazy, we already have the code, and it doesn't add any new production code. To me, `_jibDebugShowBuildConfiguration` seems more code to maintain.

TadCordle:
Also running tasks/goals in tests is a pretty big slow down, which can be annoying.

loosebazooka:
Well I still need to load the project context which is what I'm explicitly testing here. I can trust (from gradle) that JavaContainerBuilder is doing the right thing. What I need to know is that the gradle plugin code is doing the right things splitting the dependencies.

loosebazooka:
Pushed a new less ridiculous test.

jib-gradle-plugin/src/test/resources/gradle/projects/all-local-multi-service/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+// this code is for GradleProjectPropertiesTest (we execute this task to inspect the deep layer configuration)
+import com.google.cloud.tools.jib.gradle.*
+import com.google.cloud.tools.jib.plugins.common.*
+import com.google.cloud.tools.jib.api.*
+
+subprojects {
+  task inspectLayers {
+    dependsOn "classes"
+    doLast {
+      def projectProps = new GradleProjectProperties(project, project.logger)
```
chanseokoh: You don't need this, do you?

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
+package com.google.cloud.tools.jib.gradle;
```
chanseokoh: copyright

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+import org.junit.Test;
+
+public class GradleLayerConfigurationIntegrationTest {
+
+  @ClassRule
+  public static final TestProject multiTestProject = new TestProject("all-local-multi-service");
+
+  @Test
+  public void testGradleLayerConfiguration_multiModule() throws IOException {
+    multiTestProject.build(":complex-service:jibBuildTar");
```
chanseokoh: Is it possible to run Jib on the root project, or it is recommend to call Jib on sub-modules this way?

loosebazooka: It kind of is just doing what I want it to. I guess I could make it one test and check everything, is that you what you were expecting?

jib-gradle-plugin/src/test/resources/gradle/projects/all-local-multi-service/build.gradle
loosebazooka: Correct. I thought I cleared that out.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurationIntegrationTest.java
chanseokoh: Not really expecting it. If this does what you want to do, that's good. I was just wondering if `all-local-multi-module` is in a structure found in practice and what is the correct way to set up and call Jib.

loosebazooka: Yeah, I think it's somewhat realistic. It might be unnecessary to put `simple-service` in there. But to execute things one at a time, this is the way I would do it.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurationIntegrationTest.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  // returns all files in layers (*.tar.gz) in a image tar
+  private List<List<String>> getLayers(Path tar) throws IOException {
+    List<List<String>> layers = new ArrayList<>();
+
+    try (TarArchiveInputStream image = new TarArchiveInputStream(Files.newInputStream(tar))) {
+      TarArchiveEntry entry;
+      while ((entry = image.getNextTarEntry()) != null) {
+        System.out.println(entry.getName());
```
chanseokoh: stray debug code

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -149,21 +152,43 @@ public JibContainerBuilder createContainerBuilder(
...
+                  .getConfigurations()
+                  .getByName(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME)
+                  .getResolvedConfiguration()
+                  .getResolvedArtifacts()
+                  .stream()
+                  .filter(
+                      artifact ->
+                          artifact.getId().getComponentIdentifier()
+                              instanceof ProjectComponentIdentifier)
+                  .map(ResolvedArtifact::getFile)
```
chanseokoh: This is unclear, but should we do `filter(File::exists)`? For what I know, if do the filtering for `allFiles` and `classesOutputDirectories`. (`JavaContainerBuilder` does the check and throws `NoSuchFileException`, which might be a good thing actually. It's unclear to me if this can ever happen.)

jib-gradle-plugin/src/test/resources/gradle/projects/all-local-multi-service/lib/src/test/java/com/lib/LibTest.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.lib;
+
+import org.junit.Assert;
+import org.junit.Test;
```
chanseokoh: Q: how is it possible that this class is compiled? I don't see a junit dep defined in `lib/build.gradle`.

loosebazooka: Hah that's a good question, no actual test is run there in the end. The dependency doesn't trigger running tests. I think I'll just take this class out. But it should have no effect.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
loosebazooka: Yeah I think it's safer to not filter, if we have an issue in the configuration, it should pick it up. `File::exists` will just quietly ignore the error.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/GradleLayerConfigurationIntegrationTest.java
loosebazooka: ugh, oops.

PR #1787, opened 2019-06-20
Use bootWar instead of war for Spring Boot war projects

### LINKED ISSUE #1786: jibs war support doesn't play well with the spring boot gradle plugin
When the `war` and `org.springframework.boot` plugins are both applied to a gradle project, the latter disables the `war` task and registers a `bootWar` task instead.

jib should therefore use the `bootWar` task in this case.

larsgrefer:
fixes #1786

### DIFF
--- a/jib-gradle-plugin/build.gradle
+++ b/jib-gradle-plugin/build.gradle
@@ -70,6 +70,7 @@ dependencies {
   testCompile 'junit:junit:4.12'
   testCompile 'org.mockito:mockito-core:2.23.4'
   testCompile 'org.slf4j:slf4j-api:1.7.25'
+  testCompile 'org.springframework.boot:spring-boot-gradle-plugin:2.1.6.RELEASE'
 
   compile gradleApi()
 
--- a/jib-gradle-plugin/build.gradle
+++ b/jib-gradle-plugin/build.gradle
@@ -70,6 +70,8 @@ dependencies {
   testCompile 'junit:junit:4.12'
   testCompile 'org.mockito:mockito-core:2.23.4'
   testCompile 'org.slf4j:slf4j-api:1.7.25'
+
+  // only for testing a concrete Spring Boot example in a test (not for test infrastructure)
   testCompile 'org.springframework.boot:spring-boot-gradle-plugin:2.1.6.RELEASE'
 
   compile gradleApi()
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
@@ -27,7 +27,7 @@ import javax.annotation.Nullable;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.logging.Logger;
-import org.gradle.api.plugins.WarPluginConvention;
+import org.gradle.api.plugins.WarPlugin;
 import org.gradle.api.tasks.bundling.War;
 import org.gradle.internal.logging.events.LogEvent;
 import org.gradle.internal.logging.events.OutputEventListener;
@@ -39,9 +39,8 @@ class TaskCommon {
 
   @Nullable
   static War getWarTask(Project project) {
-    WarPluginConvention warPluginConvention =
-        project.getConvention().findPlugin(WarPluginConvention.class);
-    if (warPluginConvention == null) {
+
+    if (!project.getPlugins().hasPlugin(WarPlugin.class)) {
       return null;
     }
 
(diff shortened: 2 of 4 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -43,6 +44,14 @@ static War getWarTask(Project project) {
     if (warPluginConvention == null) {
       return null;
     }
+
+    if (project.getPlugins().hasPlugin("org.springframework.boot")) {
+      Task bootWar = project.getTasks().findByName("bootWar");
```
loosebazooka: How does this handle users who are building jars with the springboot plugin?

larsgrefer: It doesn't.

- Jib already works well for "normal" spring boot java applications
- For "normal" spring boot java applications both `warPluginConvention` and `project.getTasks().findByName("bootWar")` would be null, so my proposed change wouldn't do anything in this case.

loosebazooka: Ah I see, I didn't realize that's how `bootwar` works. So yeah we'll take a look at this.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -43,6 +44,14 @@ static War getWarTask(Project project) {
...
+      if (bootWar != null) { // Spring Boot > 2.0
+        return (War) bootWar;
+      }
+    }
+
     return (War) warPluginConvention.getProject().getTasks().findByName("war");
```
chanseokoh: Question as I am not a Gradle expert: can we just to `project.getTasks().findByName("war")` similarly, or is getting `warPluginConvention` still necessary? And should we still check at the beginning `if (warPluginConvention == null)` even when the bootWar case?

larsgrefer: Yes. `project` and `warPluginConvention.getProject()` will evaluate to the very same object instance.

Instead of checking for the presence of the `WarPluginConvention` it would be simpler to just check for the presence of the `WarPlugin` itself. See #1788 and #1789 for more information about this.

chanseokoh: Sounds good. Then, could you update `getWarTask()` here that way? (Check `WarPlugin` at the beginning, and don't use `warPluginConvention.getProject()`, right?)

larsgrefer: Sure. I'll do that.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -38,12 +39,19 @@
...
-    WarPluginConvention warPluginConvention =
-        project.getConvention().findPlugin(WarPluginConvention.class);
-    if (warPluginConvention == null) {
+
+    if (!project.getPlugins().hasPlugin(WarPlugin.class)) {
       return null;
     }
-    return (War) warPluginConvention.getProject().getTasks().findByName("war");
+
+    if (project.getPlugins().hasPlugin("org.springframework.boot")) {
```
loosebazooka: Can you add a test to verify this behavior too?

larsgrefer: Sure

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
larsgrefer: Done

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
larsgrefer: Done

jib-gradle-plugin/build.gradle
```diff
@@ -70,6 +70,7 @@ dependencies {
   testCompile 'junit:junit:4.12'
   testCompile 'org.mockito:mockito-core:2.23.4'
   testCompile 'org.slf4j:slf4j-api:1.7.25'
+  testCompile 'org.springframework.boot:spring-boot-gradle-plugin:2.1.6.RELEASE'
```
chanseokoh: Let's add a comment

`// only for testing a concrete Spring Boot example in a test (not for test infrastructure)`

larsgrefer: You have write access to this branch of my fork of jib. So you can do such small changes directly.

larsgrefer:
I've resolved the merge conflicts and added the comment.

Issue #1654, opened 2019-04-20
while deployment in k8s always failed error as : failed to open log file "/var/log/pods/105c52b6-6396-11e9-b608-00163e10ea14/bs/0.log"


ZhengKexue:
**Description of the issue**:
failed to open log file "/var/log/pods/105c52b6-6396-11e9-b608-00163e10ea14/bs/0.log": open /var/log/pods/105c52b6-6396-11e9-b608-00163e10ea14/bs/0.log: no such file or directory

**Expected behavior**:
Pod start success 
**Steps to reproduce**: 
1.  kubectl run bs --image=xxx --port=8080 --restart=Never --namespace=test
2.  kubectl get pods --namespace=test
     then find the pod can't start success as below:
     bs                                            0/1       ContainerCannotRun   0          12s
3. kubectl logs xxxxx --namespace=test
   check error logs as below: 
  failed to open log file "/var/log/pods/105c52b6-6396-11e9-b608-00163e10ea14/bs/0.log": open /var/log/pods/105c52b6-6396-11e9-b608-00163e10ea14/bs/0.log: no such file or directory

**Environment**: 
gradle 4.6
Mac OS 10.13.3

**`jib-maven-plugin` Configuration**: 
```xml
PASTE YOUR pom.xml CONFIGURATION HERE
```

**`jib-gradle-plugin` Configuration**: 
```groovy
 plugins {
	id 'com.google.cloud.tools.jib' version '1.1.1'
}
jib{
	container{
		mainClass = "com.test.BackendApplication"
	}
}
```

**Log output**: 

**Additional Information**: 
 build command as below ,the application is normal springboot application . build already success and image already push to remoter repository.

```java
gradle jib    -Djib.from.image=openjdk:alpine     -Djib.to.image=xxx     -Djib.to.auth.username=xxx    -Djib.to.auth.password=xxx     -Djib.to.environment=LOCAL_ENV="test"     -Djib.to.ports=8080     -Djib.container.useCurrentTimestamp=true     -Djib.container.workingDirectory=/home/admin     -Djib.container.user=admin
```

ZhengKexue:
> **Description of the issue**:
> failed to open log file "/var/log/pods/105c52b6-6396-11e9-b608-00163e10ea14/bs/0.log": open /var/log/pods/105c52b6-6396-11e9-b608-00163e10ea14/bs/0.log: no such file or directory
> 
> **Expected behavior**:
> Pod start success
[...]

Issue #808, opened 2018-08-07
Could not push image in 0.9.8 but success in 0.9.7 gradle


Demonian:
**Description of the issue**:
I encounter the same problem with gradle plugin. The same configs works for version 0.9.7, but gives error in 0.9.8. It looks like [this issue](https://github.com/GoogleContainerTools/jib/issues/790) but there are not useful solution there.

**Expected behavior**:
Should work with both versions

**Steps to reproduce**:
1. Execute `./gradlew clean jib` with version 0.9.7 (success)
1. Execute `./gradlew clean jib` with version 0.9.8 (error)

**Environment**: 
MacOs version 10.13.6
gradle wrapper 4.8.1

**`jib-gradle-plugin` Configuration**: 
```groovy
jib {
	from {
        image = 'openjdk:8-jdk-alpine'
		auth {
			username = dockerUsername
			password = dockerPassword
		}
	}
    to {
        image = "demonian/${bootJar.baseName}:${bootJar.version}"
		auth {
			username = dockerUsername
			password = dockerPassword
		}

    }
    container {
		useCurrentTimestamp = true
    }
}
```

**Log output**:
**### Success with jib 0.9.7 logs:**
Built and pushed image as demonian/prime:0.0.1-SNAPSHOT
:jib (Thread[Task worker for ':' Thread 3,5,main]) completed. Took 9.24 secs.

BUILD SUCCESSFUL in 10s
4 actionable tasks: 4 executed

**### Error with jib 0.9.8, but according to logs I see that it pushed the image:**
```
Authenticating with push to registry.hub.docker.com : 1755.206 ms
RUNNING Pushing BLOB digest: sha256:196d20239e57e877897f1804f66edb53fa657eefd403191e40b2c451fc66a223, size: 142
RUNNING Pushing BLOB digest: sha256:5b8637e76e31c34f9f794885baae1511bc52127ba052dd0a8011df837122bc5c, size: 749
RUNNING Pushing BLOB digest: sha256:a19961ec08095eb70ca3a070945223dcf51500a255a745272276139bf0a29f4e, size: 15515814
BLOB : digest: sha256:5b8637e76e31c34f9f794885baae1511bc52127ba052dd0a8011df837122bc5c, size: 749 already exists on registry
Pushing BLOB digest: sha256:5b8637e76e31c34f9f794885baae1511bc52127ba052dd0a8011df837122bc5c, size: 749 : 752.68 ms
[...]

PR #1793, opened 2019-06-20
Re-enable cross-repository blob mounts

### LINKED ISSUE #169: Have cross-repo mount automatically put source repository for base image in same registry.

briandealwis:
Re-enables `mount`/`from` for #169.  The speed-up can be immense.  Building and pushing an image with `openjdk:8` as the base image to a new Docker Hub image takes more than 3 *minutes*, whereas with `mount`/`from` it takes about 8 seconds.

This implementation is is similar in spirit to #1024, but we do some additional checking to ensure we have permission to request a blob mount.  Basically the Docker Registry Bearer Token includes the set of repositories and their granted permission.  Private repositories won't be included if we don't have permission.

High-level summary of the changes:
  - `Authorization` now parses the provided Bearer Token to maintain a list of repositories and their granted permissions.  It provides a new method, `canAccess(repository, access-scope)`.  For non-bearer tokens, `canAccess()` returns true.
  - `RegistryClient.pushBlob()` checks that the `Authorization` allows a cross-repository mount and ignores the mount point if not (this avoids the [@chanseokoh's francium25 and francium25paran permission problem](https://github.com/GoogleContainerTools/jib/pull/1024#discussion_r220295824) discovered previously)
  - `BuildConfiguration.newTargetImageRegistryClientFactory()` configures the target `RegistryClient` to request permissions for the base and target repositories when base and target registries are the same, which is recorded in the `RegistryEndpointRequestProperties` object as a _source image name_
  - `RegistryAuthenticator` requests permission for the source repository when set. A key change here is that if authorizing against the additional repository fails, we retry with only the target repository.  NOTE: this doesn't seem to happen in practice with Docker Hub¸, and hence the need to check the resulting of the Docker Registry Bearer Token's access grants instead

I debated whether I should provide a property to disable blob mounts or if I should rework `RegistryClient.pushBlob()` to retry if the blob-mount fails.  The [Registry Spec](https://docs.docker.com/registry/spec/api/#pushing-an-image) says

> If a mount fails due to invalid repository or digest arguments, the registry will fall back to the standard upload behavior and return a `202 Accepted` with the upload URL in the `Location` header:

Fixes #169


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -180,26 +183,30 @@ public Authorization authenticatePush(@Nullable Credential credential)
...
-  String getServiceScopeRequestParameters(String scope) {
-    return "service="
-        + service
-        + "&scope=repository:"
-        + registryEndpointRequestProperties.getImageName()
-        + ":"
-        + scope;
+  String getServiceScopeRequestParameters(Map<String, String> repositoryScopes) {
+    StringBuilder parameters = new StringBuilder("service=").append(service);
+    for (Entry<String, String> pair : repositoryScopes.entrySet()) {
```
loosebazooka: Do we expect this to only return what is necessary? Is it possible this generates a string with some unused scopes?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -138,7 +138,17 @@ private String makeUserAgent() {
...
   public static Factory factory(EventHandlers eventHandlers, String serverUrl, String imageName) {
-    return new Factory(eventHandlers, new RegistryEndpointRequestProperties(serverUrl, imageName));
+    return factory(eventHandlers, serverUrl, imageName, null);
+  }
+
+  public static Factory factory(
+      EventHandlers eventHandlers,
+      String serverUrl,
+      String imageName,
+      @Nullable String sourceImageName) {
```
loosebazooka: Is there a situation where this is null and the [previous `factory(` method](https://github.com/GoogleContainerTools/jib/pull/1793/files#diff-3482bc5337be75b1d8e501d4ee8a9654R140) would not be sufficient?

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
```diff
@@ -0,0 +1,124 @@
...
+package com.google.cloud.tools.jib.http;
+
+import com.google.common.collect.Multimap;
+import org.junit.Assert;
+import org.junit.Test;
+
+/** Tests for {@link Authorization}. */
+public class AuthorizationTest {
+  @Test
+  public void testDecode_dockerToken() {
```
loosebazooka: Instructions to help generate these encoded strings if we need to do it again?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -34,9 +46,66 @@
...
+   * Decode the <a href="https://docs.docker.com/registry/spec/auth/jwt/">Docker Registry v2 Bearer
+   * Token</a> to list the granted repositories with their levels of access.
+   *
+   * @param token a Docker Registry Bearer Token
+   * @return a mapping of repository to granted access scopes, or {@code null} if there is no JWT
+   *     the token is not a Docker Registry Bearer Token
+   */
+  @VisibleForTesting
+  @Nullable
+  static Multimap<String, String> decodeTokenRepositoryGrants(String token) {
```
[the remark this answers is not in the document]
briandealwis: Yeah ok.  We don't actually can about validating the token.  We can just pull out the "middle" element and decode it.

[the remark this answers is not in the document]
loosebazooka: I mean only if it seems somewhat reasonable to do.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
briandealwis: In theory it could contain all sorts of scopes.  But in practice the `repositoryScopes` will contain at most two:

  1. the target repository &rarr; `push,pull`
  2. *iff on the same registry*, then the source repository &rarr; `pull`

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
briandealwis: oicwyga — I initially was trying to get rid of the 2-arg version of `RegistryEndpointRequestProperties`.  Will restore.

briandealwis: and then I can remove the nullable

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
briandealwis: trial and error, of course.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -34,9 +46,66 @@
...
+    return new Authorization("Bearer", token, decodeTokenRepositoryGrants(token));
   }
 
+  /**
+   * Decode the <a href="https://docs.docker.com/registry/spec/auth/jwt/">Docker Registry v2 Bearer
+   * Token</a> to list the granted repositories with their levels of access.
+   *
+   * @param token a Docker Registry Bearer Token
+   * @return a mapping of repository to granted access scopes, or {@code null} if there is no JWT
+   *     the token is not a Docker Registry Bearer Token
```
chanseokoh: ".. no JWT token _or_ the token ... "

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -34,9 +46,66 @@
...
+    try {
+      DecodedJWT jwt = JWT.decode(token);
+      // Make sure they look like valid access claims.
+      JsonNode[] accessClaims = jwt.getClaim("access").asArray(JsonNode.class);
+      if (accessClaims == null
+          || !Stream.of(accessClaims).allMatch(Authorization::isValidAccessClaim)) {
+        return null;
+      }
+      return Stream.of(accessClaims)
+          .filter(n -> "repository".equals(n.get("type").asText()))
```
chanseokoh: `access` or `claim` instead of `n`? I didn't know it meant "Node" until I saw code below.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -34,9 +46,66 @@
...
+      return null;
+    }
+  }
+
+  /**
+   * Check that the provided access object looks genuine: should be a JSON object with a non-empty
+   * string "type" field.
+   */
+  private static boolean isValidAccessClaim(JsonNode n) {
+    if (!n.isObject() || Strings.isNullOrEmpty(n.get("type").asText())) {
```
chanseokoh: Can `n.get("type")` return null? (You are checking null from `asText()`.)

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -227,15 +234,43 @@ boolean isOAuth2Auth(@Nullable Credential credential) {
    */
   private Authorization authenticate(@Nullable Credential credential, String scope)
       throws RegistryAuthenticationFailedException {
+    // try authorizing againat both the main repository and the source repository too
```
chanseokoh: typo: against

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
```diff
@@ -0,0 +1,124 @@
...
+                + ".eyJhY2Nlc3MiOlt7InR5cGUiOiJyZXBvc2l0b3J5IiwibmFtZSI6ImxpYnJhcnkvb3BlbmpkayIsImFjdGlvbnMiOlsicHVsbCJdfV0sImF1ZCI6InJlZ2lzdHJ5LmRvY2tlci5pbyIsImV4cCI6MTU2MTA0MzkwNSwiaWF0IjoxNTYxMDQzNjA1LCJpc3MiOiJhdXRoLmRvY2tlci5pbyIsImp0aSI6Ikc5bWpiOE9GeU5STFlpY3ZUMFZxIiwibmJmIjoxNTYxMDQzMzA1LCJzdWIiOiIifQ"
+                + ".jblwG_taIVf3IRiv200ivsc8q_IUj-M9QePKPAULfXdSZlY6H9n_XWtT6lw43k-J6QHfmnY4Yuh3eZq61KS7AT9yggM1VuolRCvYztSZ-MZHMIlvSE2KCc0wXa5gNQarjmDJloYduZuyLaKaRUUbO4osk1MuruODY_c2g2j16ce0Z8XVJ-7R8_J_Z8g0GdtFAfPO4bqpg9dj31MA8AKl3h-ru8NXcs3y1PkrYHpEGCgpcGcUQwLY7uiIrzjr0trCUbsLsv6iq2XTXnN_tTrfvL1R3yTB6gITvXZdsnU3r_UIDTzexTtlZWdntucJAGKX9HMA_jYEcTZ4ZhyEzETGpw");
+    Assert.assertEquals(1, decoded.size());
+    Assert.assertTrue(decoded.containsEntry("library/openjdk", "pull"));
+    Assert.assertFalse(decoded.containsEntry("library/openjdk", "push"));
+    Assert.assertFalse(decoded.containsEntry("randorepo", "push"));
```
chanseokoh: typo: randomrepo

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
```diff
@@ -0,0 +1,124 @@
...
+  public void testCanAccess_bearer_withToken() {
+    // a synthetic token for accessing docker.io's openjdk with push and pull
+    Authorization authorization =
+        Authorization.fromBearerToken(
+            "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhY2Nlc3MiOlt7InR5cGUiOiJyZXBvc2l0b3J5IiwibmFtZSI6ImxpYnJhcnkvb3BlbmpkayIsImFjdGlvbnMiOlsicHVsbCIsInB1c2giXX1dfQ.VEn96Ug4eseKHX3WwP3PlgR9P7Y6VuYmMm-YRUjngFg");
+    Assert.assertNotNull(authorization);
+    Assert.assertTrue(authorization.canAccess("library/openjdk", "pull"));
+    Assert.assertTrue(authorization.canAccess("library/openjdk", "push"));
+    Assert.assertFalse(authorization.canAccess("library/openjdk", "other"));
+    Assert.assertFalse(authorization.canAccess("randorepo", "push"));
```
chanseokoh: ditto

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
```diff
@@ -0,0 +1,124 @@
...
+  @Test
+  public void testCanAccess_bearer_withNonToken() {
+    // non-Docker Bearer Tokens are assumed to allow access to all
+    Authorization authorization =
+        Authorization.fromBearerToken(
+            "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c");
+    Assert.assertNotNull(authorization);
+    Assert.assertTrue(authorization.canAccess("library/openjdk", "pull"));
+    Assert.assertTrue(authorization.canAccess("library/openjdk", "push"));
+    Assert.assertTrue(authorization.canAccess("randorepo", "push"));
```
chanseokoh: ditto

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
chanseokoh: Is there a way to easily synthesize these encoded values from human-readable JSONs? Can you document that as comments in the file?

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
```diff
@@ -0,0 +1,124 @@
...
...lNrbEVVVEFlRncweE9UQXhNVEl3TURJeU5EVmFGdzB5TURBeE1USXdNREl5TkRWYU1FWXhSREJDQmdOVkJBTVRPMUpMTkZNNlMwRkxVVHBEV0RWRk9rRTJSMVE2VTBwTVR6cFFNbEpMT2tOWlZVUTZTMEpEU0RwWFNVeE1Pa3hUU2xrNldscFFVVHBaVWxsRU1JSUJJakFOQmdrcWhraUc5dzBCQVFFRkFBT0NBUThBTUlJQkNnS0NBUUVBcjY2bXkveXpHN21VUzF3eFQ3dFplS2pqRzcvNnBwZFNMY3JCcko5VytwcndzMGtIUDVwUHRkMUpkcFdEWU1OZWdqQXhpUWtRUUNvd25IUnN2ODVUalBUdE5wUkdKVTRkeHJkeXBvWGc4TVhYUEUzL2lRbHhPS2VNU0prNlRKbG5wNGFtWVBHQlhuQXRoQzJtTlR5ak1zdFh2ZmNWN3VFYWpRcnlOVUcyUVdXQ1k1Ujl0a2k5ZG54Z3dCSEF6bG8wTzJCczFmcm5JbmJxaCtic3ZSZ1FxU3BrMWhxYnhSU3AyRlNrL2tBL1gyeUFxZzJQSUJxWFFMaTVQQ3krWERYZElJczV6VG9ZbWJUK0pmbnZaMzRLcG5mSkpNalpIRW4xUVJtQldOZXJZcVdtNVhkQVhUMUJrQU9aditMNFVwSTk3NFZFZ2ppY1JINVdBeWV4b1BFclRRSURBUUFCbzRHeU1JR3ZNQTRHQTFVZER3RUIvd1FFQXdJSGdEQVBCZ05WSFNVRUNEQUdCZ1JWSFNVQU1FUUdBMVVkRGdROUJEdFNTelJUT2t0QlMxRTZRMWcxUlRwQk5rZFVPbE5LVEU4NlVESlNTenBEV1ZWRU9rdENRMGc2VjBsTVREcE1VMHBaT2xwYVVGRTZXVkpaUkRCR0JnTlZIU01FUHpBOWdEc3lWMDVaT2xWTFMxSTZSRTFFVWpwU1NVOUZPa3hITmtFNlExVllWRHBOUmxWTU9rWXpTRVU2TlZBeVZUcExTak5HT2tOQk5sazZTa2xFVVRBS0JnZ3Foa2pPUFFRREFnTkpBREJHQWlFQXFOSXEwMFdZTmM5Z2tDZGdSUzRSWUhtNTRZcDBTa05Rd2lyMm5hSWtGd3dDSVFEMjlYdUl5TmpTa1cvWmpQaFlWWFB6QW9TNFVkRXNvUUhyUVZHMDd1N3ZsUT09Il19"
```
chanseokoh: I'd like to ensure this token has no other access than pulling public Docker Hub images.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -34,9 +46,66 @@
...
+      return false;
+    }
+    if ("repository".equals(n.get("type").asText())) {
+      // repository should have a name and array of permitted actions
+      return !Strings.isNullOrEmpty(n.get("name").asText())
+          && n.get("actions").isArray()
+          && Iterators.all(n.get("actions").iterator(), JsonNode::isTextual);
+    }
+    return true;
+  }
```
chanseokoh: nit: needs a blank line

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -34,9 +46,66 @@
    * @return an {@link Authorization} with a {@code Bearer} token
    */
   public static Authorization fromBearerToken(String token) {
-    return new Authorization("Bearer", token);
+    return new Authorization("Bearer", token, decodeTokenRepositoryGrants(token));
   }
 
+  /**
+   * Decode the <a href="https://docs.docker.com/registry/spec/auth/jwt/">Docker Registry v2 Bearer
```
chanseokoh: Let's not interleave `decodeTokenRepositoryGrants` and `isValidAccessClaim` between the `fromXXX` static initializer. Put it after the last `fromXXX`. Maybe moving `fromBearerToken` to make it the last will help.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -45,23 +114,26 @@ public static Authorization fromBearerToken(String token) {
...
    * @return an {@link Authorization} with a base64-encoded {@code username:password} string
    */
   public static Authorization fromBasicToken(String token) {
-    return new Authorization("Basic", token);
+    return new Authorization("Basic", token, null);
   }
 
   private final String scheme;
   private final String token;
+  @Nullable private final Multimap<String, String> repositoryGrants;
```
chanseokoh: Since the token and this value is related (i.e., the value is actually part of a token) if not null, I'd like to add a comment like

"If token is a Bearer token, repositoryGrants must be the repository grant information extracted from the token. Otherwise, it must be null."

or something similar you think suitable. Because when someone new sees these fields, they may think these values can be independent.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -94,4 +166,16 @@ public boolean equals(Object other) {
...
+
+  /**
+   * Check if this authorization allows accessing the specified repository.
+   *
+   * @param repository repository in question
+   * @param access the access scope ("push" or "pull")
+   * @return true if the repository was covered
+   */
+  public boolean canAccess(String repository, String access) {
+    // if null then we assume that all repositories are granted
```
chanseokoh: Worth explaining this in the Javadoc.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -34,7 +44,57 @@
...
+    //   "aud":"registry.docker.io",
+    //   "iss":"auth.docker.io",
+    //   "exp":999,
+    //   "iat":999,
+    //   "jti":"zzzz",
+    //   "nbf":999,
+    //   "sub":"e3ae001d-xxx"
+    // }
+    //
+    TokenPayloadTemplate payload;
```
chanseokoh: This method doesn't return `IOException`, so I think the code looks aesthetically better if

```java
try {
  TokenPayloadTemplate payload = ...
  if (payload.access == null) {
    return null;
  }
  return payload.access...
} catch (IOException ex) {
  return null;
}
```

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
```diff
@@ -94,4 +157,40 @@ public boolean equals(Object other) {
...
+  /**
+   * Represents an access claim for a repository in a Docker Registry Bearer Token payload.
+   *
+   * <pre>{"type": "repository","name": "library/openjdk","actions":["push","pull"]}</pre>
+   */
+  @JsonIgnoreProperties(ignoreUnknown = true)
+  private static class AccessClaim implements JsonTemplate {
+    @Nullable private String type;
+    @Nullable private String name;
+    private List<String> actions = new ArrayList<>();
```
chanseokoh: This cannot be nullable? (The `access` field is nullable OTOH.)

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
briandealwis: This was the token returned by Docker Hub for the `examples/helloworld` image built after a `docker logout` with a `mvn package jib:dockerBuild -Djib.from.image=openjdk -Djava.util.logging.config.file=...` with a logging config file to [obtain a network trace as per our FAQ](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-can-i-examine-network-traffic).  

You can copy and paste this JWT into `jwt.io`'s debugger.  Alternatively you can base64 decode the middle component (the payload).

jib-core/src/test/java/com/google/cloud/tools/jib/http/AuthorizationTest.java
briandealwis: Not a typo but not worth confusion

jib-core/src/main/java/com/google/cloud/tools/jib/http/Authorization.java
loosebazooka: Yeah I don't understand why "null" is valid

briandealwis: The other non-Docker Container Registry Token methods don't provide this information.

chanseokoh: I think the method name and the Javadoc can still be confusing. I think we just need to clarify (and justify). For example, I can create an `Authorization` instance with `fromBasicCredentials("my-username", "my-passwd")`. Then `canAccess("any repository", "pull or push or anything")` will always return true. I feel this is not intuitive.

briandealwis: What about making this specific to the blob-mount: `canAttemptBlobMount(repository)`

chanseokoh: That technically works. Man, naming is hard. I see depending on the name, we may want to relocate this class from the `.http` package. (Relocation can be done later.) This class is meant for holding the very string value for the Authorization HTTP header (i.e., for `Authorization: <some schema> <token value>`) as explained in the class Javadoc, hence was holding the two fields `schema` and `token`. I thought `repositoryGrants` was OK, because it is actually information in `token`. But if we incorporate the notions of BLOb or cross-repo mount, it may make sense to get this out of `.http`.

That said, I don't have a strong opinion about what the name should be. Maybe the logic for `canAttemptBlobMount` can be outside of this method, I don't know.

briandealwis: Hmm good point.  I'll move this into `RegistryClient` — it's only called from `pushBlobStep`.

briandealwis:
Moved the _Docker Registry Bearer Token_ related code into RegistryClient, though I kept the test separate as `DockerRegistryBearerTokenTest`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -298,9 +385,7 @@ public boolean pushBlob(
       Consumer<Long> writtenByteCountListener)
       throws IOException, RegistryException {
 
-    if (sourceRepository != null
-        && authorization != null
-        && !authorization.canAccess(sourceRepository, "pull")) {
+    if (sourceRepository != null && canMountBlobs(authorization, sourceRepository)) {
```
chanseokoh: I think the condition is reversed. `... && !canMountBlobs()`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -337,6 +422,29 @@ public boolean pushBlob(
...
+  /**
+   * Check if the authorization allows using the specified repository can be mounted by the remote
+   * registry as a source for blobs. More specifically, we can only check if the repository is not
+   * disallowed.
+   *
+   * @param repository repository in question
+   * @return {@code true} if the repository appears to be mountable
+   */
+  @VisibleForTesting
+  static boolean canMountBlobs(@Nullable Authorization authorization, String repository) {
```
chanseokoh: can{Try|Attempt}BlobMount()?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
briandealwis: :blush: (thank you for proving yet again the value of code review)

briandealwis: We should have tests around the RegistryClient.  Maybe can make `TestWebServer` take a list of expect-response pairs.

loosebazooka:
Do you think we should hide this behind a flag for now?

briandealwis:
We had partially implemented it in 0.9.8 but it broke some people so we backed it out for 0.9.9: #808 #1654 

I will add a flag to disable it.  The wins from having it enabled are immense (8s vs 3:34m).

loosebazooka:
okay, so makes sense to enable by default?

briandealwis:
I somehow fat-fingered the commit button before I could edit that mess of a commit message. Blech.

loosebazooka:
Life goes on

PR #1797, opened 2019-06-20
CHANGELOG.md for Spring Boot bootWar support

chanseokoh:

PR #1798, opened 2019-06-20
Add some documentation for the Spring Boot WAR support

larsgrefer:
see #1797 #1787 #1786

### DIFF
--- a/jib-gradle-plugin/README.md
+++ b/jib-gradle-plugin/README.md
@@ -415,6 +415,8 @@ jib {
 }
 ```
 
+If you are using Spring Boot in your war project, make sure to [create a deployable war file](https://docs.spring.io/spring-boot/docs/2.1.x/reference/html/howto-traditional-deployment.html#howto-create-a-deployable-war-file). (Extending from `SpringBootServletInitializer` and declaring the embedded container as provided)
+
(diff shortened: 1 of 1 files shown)


jib-gradle-plugin/README.md
```diff
@@ -415,6 +415,8 @@ jib {
 }
 ```
 
+If you are using Spring Boot in your war project, make sure to [create a deployable war file](https://docs.spring.io/spring-boot/docs/2.1.x/reference/html/howto-traditional-deployment.html#howto-create-a-deployable-war-file). (Extending from `SpringBootServletInitializer` and declaring the embedded container as provided)
```
chanseokoh: Hmm... do you think this helps? Although Spring Boot is really popular, I think we don't generally want to explain in Jib docs some detailed instructions on how to create a WAR using various web frameworks. Knowing how to create a deployable WAR doesn't seem related to explaining how to use Jib for WAR projects. If there's something necessary to say to use Jib in case of a Spring Boot WAR project, but I don't really see something special in this update.

TadCordle: Agreed, I feel like something like this would be better suited in the FAQ or something than in the README, if anything.

PR #1799, opened 2019-06-20
Fix root progress construction

### LINKED ISSUE #1796: Test failure about root progress dispatch
This looks like a flaky test, but I don't remember having seen this. It means the progress event dispatched more completion than the total expected. It's concerning, so I'll dig into it.

```
    * What went wrong:
    Execution failed for task ':jibDockerBuild'.
    > Progress exceeds max for 'building image to Docker daemon': 1 more beyond 5

    * Try:
    Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

    * Get more help at https://help.gradle.org

    BUILD FAILED in 4s
    3 actionable tasks: 3 executed
        at org.gradle.testkit.runner.internal.DefaultGradleRunner$1.execute(DefaultGradleRunner.java:237)
        at org.gradle.testkit.runner.internal.DefaultGradleRunner$1.execute(DefaultGradleRunner.java:234)
        at org.gradle.testkit.runner.internal.DefaultGradleRunner.run(DefaultGradleRunner.java:298)
        at org.gradle.testkit.runner.internal.DefaultGradleRunner.build(DefaultGradleRunner.java:234)
        at com.google.cloud.tools.jib.gradle.TestProject.build(TestProject.java:82)
        at com.google.cloud.tools.jib.gradle.JibRunHelper.buildToDockerDaemon(JibRunHelper.java:96)
        at com.google.cloud.tools.jib.gradle.EmptyProjectIntegrationTest.testDockerDaemon_userNames(EmptyProjectIntegrationTest.java:112)
```

chanseokoh:
Fixes #1796.

I missed that the root progress tree should be created in the main thread. (`newChildProducer()` should not be called in child threads.)

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
@@ -147,7 +147,7 @@ public class ProgressEventDispatcher implements Closeable {
   private long decrementRemainingAllocationUnits(long units) {
     Preconditions.checkState(!closed);
 
-    if (remainingAllocationUnits > units) {
+    if (remainingAllocationUnits >= units) {
       remainingAllocationUnits -= units;
       return units;
     }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -32,9 +32,7 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 import java.nio.file.Path;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
-import java.util.Deque;
 import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -103,8 +101,7 @@ public class StepsRunner {
   private final List<Runnable> stepsToRun = new ArrayList<>();
 
   @Nullable private String rootProgressDescription;
-  private final Deque<ProgressEventDispatcher.Factory> childProgressDispatcherFactories =
-      new ArrayDeque<>();
+  @Nullable private ProgressEventDispatcher rootProgressDispatcher;
 
   private StepsRunner(
       ListeningExecutorService executorService, BuildConfiguration buildConfiguration) {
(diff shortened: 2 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
```diff
@@ -147,7 +147,7 @@ public void dispatchProgress(long progressUnits) {
   private long decrementRemainingAllocationUnits(long units) {
     Preconditions.checkState(!closed);
 
-    if (remainingAllocationUnits > units) {
+    if (remainingAllocationUnits >= units) {
```
chanseokoh: Note: this is unrelated to #1796 and does not change the behavior. (If they are same, the remaining units become 0.) I just think `>=` looks more intuitive.

TadCordle: To me `>=` immediately followed by the subtraction looks less intuitive, but it's small.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -241,7 +239,7 @@ private void writeTarFile(Path outputPath) {
             () ->
                 new WriteTarFileStep(
                         buildConfiguration,
-                        childProgressDispatcherFactorySupplier.get(),
+                        childProgressDispatcherFactories.pop(),
```
loosebazooka: what am I missing here? Why can't this call just be: progressEventDispatcher.newChildProducer() ?

TadCordle: It sounds like `newChildProducer()` shouldn't be called in separate threads, so we need to create them all in the main thread first and then pass them to each step when they run.

loosebazooka: Either way, I find this mechanism a little strange. Why not just keep a reference to progressEventDispatcher instead of pre-generating children?

chanseokoh: For this scheduling, if we just do `progressEventDispatcher.newChildProducer()`, it will not be called immediately because it is wrapped in a runnable as `() -> ... progressEventDispatcher.newChildProducer() ...`. Here, the main thread should create child producers and construct the tree properly.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/ProgressEventDispatcher.java
chanseokoh: That may be true if you look at this condition and the true branch, but I found it difficult to digest the false branch if the remaining units are equal to the units to subtract.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: You were right @loosebazooka. I can directly call `progressEventDispatcher.newChildProducer()`; just need to do it at the beginning.

PR #1803, opened 2019-06-21
Apply consistent async schedule pattern / decouple async aspect from PushImageStep 

chanseokoh:
Follow-up to #1771.

- Refactored `PushImageStep` to follow the pattern of `static makeList()` of returning a list of sub-steps. No more accepting `listenableExecutorService` or dealing with async scheduling of sub-steps inside.
- Merged `PullAndCacheBaseImageLayersStep` and `PullAndCacheBaseImageLayerStep`, following the same pattern.

All the other changes are minor things: moving code around, renaming, etc.

@TadCordle I had renamed `PushImageStep` to `PushManifestStep` but reverted it at the last moment for easy review. Will do renaming later.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
```diff
@@ -35,6 +40,32 @@
 
   private static final String DESCRIPTION = "Pulling base image layer %s";
 
+  static ImmutableList<PullAndCacheBaseImageLayerStep> makeList(
```
chanseokoh: Note: just moved from `PullAndCacheBaseImageLayersStep`, which is gone now.

chanseokoh:
Not that the pattern of `static makeList()` is ideal here. The pattern has been there (although a bit inconsistently), and I applying it throughout for consistency.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -228,7 +297,7 @@ private void pushApplicationLayers() {
                         Verify.verifyNotNull(results.applicationLayers))));
   }
 
-  private void pushImage() {
+  private void pushImages() {
```
loosebazooka: I think this should remain `pushImage`? Do we really consider multiple tags to be multiple images?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -151,7 +220,7 @@ private void pullAndCacheBaseImageLayers() {
         executorService.submit(
             () ->
                 scheduleCallables(
```
loosebazooka: I know this is old code, but why are submitting something that then submits some more things? Is this some kind of delayed submit? Could we just call scheduleCallables without the wrapper Callable?

chanseokoh: This scheduling depends on getting `results.baseImageAndAuth` (a `Future`), so calling `.get()` on it will block. Retrieving the base image should not be on the main thread of `StepsRunner`.

PR #1804, opened 2019-06-21
Leverage Nullaway checks using Optional for async step results

chanseokoh:
Getting a result from a `Future` invalidates Nullaway that has been extremely useful, as we can no longer enforce if `Future.get()` may return null or not. `Optional` works wonders here.

I see two steps among all may return null:

1. `RetrieveRegistryCredentialsStep` returning a `Credential`
2. `AuthenticatePushStep` returning an `Authorization`

We can leverage Nullaway checks using `Optional`.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
@@ -22,14 +22,14 @@ import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.builder.ProgressEventDispatcher;
 import com.google.cloud.tools.jib.builder.TimerEventDispatcher;
 import com.google.cloud.tools.jib.configuration.BuildConfiguration;
+import com.google.cloud.tools.jib.event.EventHandlers;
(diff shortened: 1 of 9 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -71,28 +71,26 @@ static RetrieveRegistryCredentialsStep forTargetImage(
...
     try (ProgressEventDispatcher ignored =
             progressEventDispatcherFactory.create("retrieving credentials for " + registry, 1);
-        TimerEventDispatcher ignored2 =
-            new TimerEventDispatcher(buildConfiguration.getEventHandlers(), description)) {
+        TimerEventDispatcher ignored2 = new TimerEventDispatcher(eventHandlers, description)) {
       for (CredentialRetriever credentialRetriever : credentialRetrievers) {
         Optional<Credential> optionalCredential = credentialRetriever.retrieve();
         if (optionalCredential.isPresent()) {
-          return optionalCredential.get();
+          return optionalCredential;
```
TadCordle: Why bother returning an `Optional` if it's always present?

TadCordle: Oh I guess I see what's happening here.

Issue #1567, opened 2019-03-19
Support referencing base images from a manifest list


briandealwis:
Follow-on to #1547 (and #1360), users should be able to identify a platform for base image from a manifest list.  [Manifest lists](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list) are a mechanism for aggregating and serving platform-specific images from a registry.  The list specifies manifests with a set of annotations for `os`, `architecture`, `os.version`, `os.features`, `variant` (cpu), and `features` (cpu).

There's [a nice article describing use of manifest lists](https://medium.com/@mauridb/docker-multi-architecture-images-365a44c26be6).  [The Docker CLI support an experimental `--platform` argument](https://docs.docker.com/engine/reference/commandline/build/) for specifying os/arch/variant combinations.

A related question is support for creating and pushing manifest lists — including perhaps being able to reference images built by jib within the reactor.

- [X] when fixing this bug, remember to ~~remove~~/update the [FAQ entry](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image)

chanseokoh:
There's something I am still not getting it. For example, `openjdk:latest` or `openjdk:11` seems like a manifest list, and we know Jib has been working well. But #1360 suggests, perhaps when reference through a digest, using a manifest list doesn't work. Hence the questions to clear:

1. Is it certain that `openjdk:11` is a manifest list? (Probably.) If so, how has Jib been able to cope with a manifest list? Or, is the registry that returns amd/linux by default in this situation?
1. But pulling a manifest list by a digest doesn't work. Is it the use of a digest that makes this difference? That is, unlike using a tag, a registry doesn't assume amd/linux by default?

briandealwis:
Jib doesn't use manifest lists: to do so requires that we request `application/vnd.docker.distribution.manifest.list.v2`, but [we don't](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPuller.java#L73).  A manifest list is just a map of platforms to a sha1, which is why referencing a specific digest works.

`openjdk` does provide a manifest list:
```sh
$ docker manifest inspect openjdk
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
[...]

PR #1805, opened 2019-06-21
add v2.2 manifest list template

loosebazooka:
part of #1567

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
@@ -73,7 +73,7 @@ public class V22ManifestListTemplate implements ManifestTemplate {
     return SCHEMA_VERSION;
   }
 
-  private List<ManifestDescriptorTemplate> manifests;
+  @Nullable private List<ManifestDescriptorTemplate> manifests;
 
   @VisibleForTesting
   List<ManifestDescriptorTemplate> getManifests() {
@@ -82,7 +82,7 @@ public class V22ManifestListTemplate implements ManifestTemplate {
   }
 
   public List<String> getDigestsForPlatform(String architecture, String os) {
-    return manifests
+    return getManifests()
         .stream()
         .filter(
             manifest ->
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
@@ -16,7 +16,7 @@
 
 package com.google.cloud.tools.jib.image.json;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.google.cloud.tools.jib.json.JsonTemplate;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
@@ -77,8 +77,7 @@ public class V22ManifestListTemplate implements ManifestTemplate {
 
   @VisibleForTesting
   List<ManifestDescriptorTemplate> getManifests() {
-    Preconditions.checkNotNull(manifests);
-    return manifests;
+    return Preconditions.checkNotNull(manifests);
   }
 
   public List<String> getDigestsForPlatform(String architecture, String os) {
@@ -102,32 +101,10 @@ public class V22ManifestListTemplate implements ManifestTemplate {
    */
   static class ManifestDescriptorTemplate implements JsonTemplate {
 
+    @JsonIgnoreProperties(ignoreUnknown = true)
     static class Platform implements JsonTemplate {
       @Nullable private String architecture;
       @Nullable private String os;
-
-      // ignored properties
-      @Nullable
-      @JsonProperty("os.version")
-      private String osVersion;
-
-      @Nullable
-      @JsonProperty("os.features")
-      private List<String> osFeatures;
-
-      @Nullable private String variant;
-      @Nullable private List<String> features;
-      // end ignored
-
-      @Nullable
-      String getArchitecture() {
-        return architecture;
-      }
-
-      @Nullable
-      public String getOs() {
-        return os;
-      }
     }
 
     @Nullable private String mediaType;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2019 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  private List<ManifestDescriptorTemplate> manifests = new ArrayList<>();
+
+  @VisibleForTesting
+  List<ManifestDescriptorTemplate> getManifests() {
+    return manifests;
+  }
+
+  public List<String> getDigestForPlatform(String architecture, String os) {
```
TadCordle: `getDigestsForPlatform`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -0,0 +1,171 @@
...
+    return manifests
+        .stream()
+        .filter(
+            manifest -> {
```
TadCordle: would it look better to consolidate this to
```
manifest -> 
    manifest.platform != null 
        && os.equals(manifest.platform.os)
        && architecture.equals(manifest.platform.architecture
```
```

File: jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java [hunk 3]
```diff
@@ -0,0 +1,171 @@
...
+  public static final String MANIFEST_MEDIA_TYPE =
+      "application/vnd.docker.distribution.manifest.list.v2+json";
+  private static final int SCHEMA_VERSION = 2;
+
+  @Override
+  public int getSchemaVersion() {
+    return SCHEMA_VERSION;
+  }
+
+  private List<ManifestDescriptorTemplate> manifests = new ArrayList<>();
```
chanseokoh: Can this be `final`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -0,0 +1,171 @@
...
+      @Nullable
+      @JsonProperty("os.version")
+      private String osVersion;
+
+      @Nullable
+      @JsonProperty("os.features")
+      private List<String> osFeatures;
+
+      @Nullable private String variant;
+      @Nullable private List<String> features;
```
chanseokoh: Any rule for this type of field, given that you did `private List<ManifestDescriptorTemplate> manifests = new ArrayList<>()` above?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+
+    @Nullable private String mediaType;
+    @Nullable private String digest;
+    private long size;
+    @Nullable private Platform platform;
+
+    /** Required by Jackson-JSON */
+    public ManifestDescriptorTemplate() {}
```
chanseokoh: Interesting. Shouldn't this constructor be implicit and exist anyways?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
loosebazooka: I think features is technically nullable in that the json doesn't **need** to include it. However manifest **must** be defined in the json.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
loosebazooka: Yeah, for some reason I was getting an error with jackson, but that might be because I was using non-static inner classes. Let me try again.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
loosebazooka: I think it technically can, but it shouldn't be. Jackson is actually injecting some other list into there. The empty array list is a way to specify that this shouldn't be nullable?

loosebazooka: ended up re-organizing this.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
loosebazooka: I changed all this to be consistent.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -0,0 +1,154 @@
...
+  public int getSchemaVersion() {
+    return SCHEMA_VERSION;
+  }
+
+  @Nullable private List<ManifestDescriptorTemplate> manifests;
+
+  @VisibleForTesting
+  List<ManifestDescriptorTemplate> getManifests() {
+    Preconditions.checkNotNull(manifests);
+    return manifests;
```
chanseokoh: nit: `return Preconditions.checkNotNull(manifests)`??

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -0,0 +1,154 @@
...
+   * @see <a href="https://github.com/opencontainers/image-spec/blob/master/descriptor.md">OCI
+   *     Content Descriptors</a>
+   */
+  static class ManifestDescriptorTemplate implements JsonTemplate {
+
+    static class Platform implements JsonTemplate {
+      @Nullable private String architecture;
+      @Nullable private String os;
+
+      // ignored properties
```
chanseokoh: Are you going to use these optional fields soon in follow-ups? Otherwise, I think we can annotate the class with `@JsonIgnoreProperties(ignoreUnknown = true)` like in #1793?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -0,0 +1,131 @@
...
+        .map(ManifestDescriptorTemplate::getDigest)
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * Template for inner JSON object representing content descriptor for a layer or container
+   * configuration.
+   *
+   * @see <a href="https://github.com/opencontainers/image-spec/blob/master/descriptor.md">OCI
+   *     Content Descriptors</a>
```
briandealwis: Sorry, I'm late to the party, but this link seems out of place since is a docker manifest list?

(If it is equally applicable to an OCI Image Index, maybe it should be broken out as a separate class?)

loosebazooka: Uh this javadoc needs to be cleaned up. Oops

Issue #1808, opened 2019-06-24
Enhance RegistryClient tests


briandealwis:
We should have tests around the RegistryClient network requests.  Maybe can make `TestWebServer` take a list of expect-response pairs.

_Originally posted by @briandealwis in https://github.com/GoogleContainerTools/jib/pull/1793/review_comment/create_

chanseokoh:
Sounds good. Just don't start working on this now; I have a branch modifying `TestWebServer` that I was planning to push at some far point in the future, and I think I should push that now.

chanseokoh:
@louismurerwa some tests we can add:
- `RegistryClient.pullManifest()`
- `RegistryClient.pushManifest()`
- `RegistryClient.checkBlob()`
- `RegistryClient.pullBlob()`
- `RegistryClient.pushBlob()`

Issue #1434, opened 2019-01-18
Update google-http-client and google-http-client-apache-v2 to 1.31


chanseokoh:
#1415 upgraded google-http-client to 1.27.0, but *a lot of our usage have been marked deprecated* with 1.27.0. Unfortunately, due to [this bug](https://github.com/googleapis/google-http-java-client/issues/503) in 1.27.0, it is not possible to get away from the deprecated usage.

To address all these issues, they made significant changes to 1.28.0. They had to break the Apache adapter into its own artifact google-http-client-apache with two different implementation flavors:
   - google-http-client-apache 1.28.0 using the legacy Apache HttpClient; and
   - 2.0.0 for the modern one.

They said the legacy version was necessary for internal usage: https://github.com/googleapis/google-http-java-client/pull/543, https://github.com/googleapis/google-http-java-client/pull/558, https://github.com/googleapis/google-http-java-client/issues/576#issuecomment-454929854

We may choose to stick with the legacy google-http-client-*apache* 1.28.0, but I think [moving to 2.0.0 is the right thing to do](https://github.com/googleapis/google-http-java-client/issues/576#issuecomment-454979169). In any case, I realized there are some complication and annoyance: https://github.com/GoogleContainerTools/jib/pull/1415#issuecomment-454925503, https://github.com/googleapis/google-http-java-client/issues/576#issuecomment-454988761

Another side effect with google-http-client 1.28.0 is that [we need to upgrade Guava to 26.0-jre](https://github.com/googleapis/google-http-java-client/issues/576#issuecomment-454972184) from the current 23.5-jre.

Lastly, I noticed one very useful and important feature in the modern 2.0.0. Looking at the code, there seems *no need to configure proxy auth ourselves* (that we did in #1337), as long as we build the Apache `HttpClient` with `useSystemProperties()`. (The default `HttpClient` built with `ApacheHttpTransport` turns it on.) Then, it will use `SystemDefaultCredentialsProvider`, and looking at the code, it will pick up `http(x).proxy{User|Password}`. It does look like we can remove our `addProxyCredentials()` code.

(FTR, https://github.com/GoogleContainerTools/jib/compare/remove-deprecated?expand=1 has a not-so-great reference implementation, but has [some problem](https://github.com/GoogleContainerTools/jib/pull/1415#issuecomment-454925503).)

chanseokoh:
[...]

PR #1809, opened 2019-06-24
Add test for proxy user/password properties

chanseokoh:
- Enhance `TestWebServer` to return a given (and only given) set of responses in order.
- Emulate a proxy server response to test HTTP proxy user/password properties.

Towards #1434 (which is blocked on a new 1.30.2 release.)

Adding the test early to unblock #1808.

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
@@ -60,7 +60,7 @@ public class TestWebServer implements Closeable {
       throws IOException, InterruptedException, GeneralSecurityException, URISyntaxException {
     this.https = https;
     this.responses = responses;
-    serverSocket = createServerSocket(https);
+    serverSocket = https ? createHttpsServerSocket() : new ServerSocket(0);
     ignoreReturn(executorService.submit(this::serveResponses));
     threadStarted.acquire();
   }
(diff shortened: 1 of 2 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -47,14 +49,26 @@
...
+  }
+
+  public TestWebServer(boolean https, List<String> responses)
+      throws IOException, InterruptedException, GeneralSecurityException, URISyntaxException {
     this.https = https;
-    serverSocket = createServerSocket(https);
-    ignoreReturn(executorService.submit(this::serve200));
+    this.responses = responses;
+    serverSocket = https ? createHttpsServerSocket() : new ServerSocket(0);
+    ignoreReturn(executorService.submit(this::serveResponses));
```
TadCordle: I don't understand what this is for?

chanseokoh: Nothing really changed. I just renamed the method `serve200` to `serveResponses`.

Nothing really changed about assigning `serverSocket` as well.

If you are asking what this PR is for, I'm just adding one more test to verify that our `Connection` class is supporting `http(s).proxyUser/Password` properly. Although I need this test in a later PR that has not been filed, I wanted to merge this test now, since #1808 will most like require modifying `TestWebServer`, which I am modifying here to add the test.

TadCordle: I meant `ignoreReturn` in general. I know it's not new, I just don't know why it's needed in the first place.

chanseokoh: `submit()` returns a `Future`, and it causes a warning that I am not making any use of the return value. I will update the comment to eliminate confusion.

TadCordle: Oh I see

PR #1810, opened 2019-06-25
fix javadoc comment for manifest list json template

loosebazooka:
minor cleanup

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
@@ -92,13 +92,7 @@ public class V22ManifestListTemplate implements ManifestTemplate {
         .collect(Collectors.toList());
   }
 
-  /**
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -92,13 +92,7 @@ public int getSchemaVersion() {
...
   }
 
-  /**
-   * Template for inner JSON object representing content descriptor for a layer or container
-   * configuration.
-   *
-   * @see <a href="https://github.com/opencontainers/image-spec/blob/master/descriptor.md">OCI
-   *     Content Descriptors</a>
-   */
+  /** Template for inner JSON object a single platform specific manifest. */
```
TadCordle: Template for inner JSON object **representing** a single platform specific manifest.

loosebazooka: ugh :+1:

Issue #1360, opened 2018-12-18
Use digest instead of version tag for baseimage


dzwicker:
**Description of the issue**:
As docker tags can be reference different container digest over time we want to use a digest for the base image.

**Expected behavior**:
Downloading the base image and build the new application image.

**Steps to reproduce**: 

1. Use a digest instead of a tag for the from image
2. call `gradle jib`

**Environment**: 
- JVM: 11.0.1 (AdoptOpenJDK 11.0.1+13)
- OS: Mac OS X 10.14.2 x86_64
- Gradle: 5.0

**`jib-gradle-plugin` Configuration**: 
```groovy
jib {
    from {
        // image = 'openjdk:11-jre-slim'
        image = 'openjdk@sha256:8ab7b3078b01ba66b937b7fbe0b9eccf60449cc101c42e99aeefaba0e1781155'
    }
    to {
        image = "repo/project/${rootProject.name}-${project.name}:${project.version.toString().replaceAll('\\+', '_')}"
    }
}
```

**Additional info**:
The problem is that the docker registry returns an unknown mediatype for the mainfest. It is `application/vnd.docker.distribution.manifest.list.v2+json`. The `list` is the problem as it is unknown by the `ManifestPuller`. 

**Log output**:
```

Getting base image openjdk@sha256:8ab7b3078b01ba66b937b7fbe0b9eccf60449cc101c42e99aeefaba0e1781155...
Building dependencies layer...
Building snapshot dependencies layer...
Building resources layer...
Building classes layer...
The base image requires auth. Trying again for openjdk@sha256:8ab7b3078b01ba66b937b7fbe0b9eccf60449cc101c42e99aeefaba0e1781155...
Retrieving registry credentials for registry.hub.docker.com...
No credentials could be retrieved for registry registry.hub.docker.com

> Task :sp-rest:jibDockerBuild
Task ':sp-rest:jibDockerBuild' is not up-to-date because:
  Task has not declared any outputs despite executing actions.
Adding corresponding output directories of source sets to image
        '/Users/dzwicker/authada/backend/eservice-v3/sp/sp-rest/build/classes/java/main' (not found, skipped)
        '/Users/dzwicker/authada/backend/eservice-v3/sp/sp-rest/build/classes/kotlin/main'

> Task :sp-rest:jibDockerBuild FAILED
:sp-rest:jibDockerBuild (Thread[Execution worker for ':' Thread 7,5,main]) completed. Took 3.127 secs.

[...]

PR #1811, opened 2019-06-26
Support manifest lists when referenced by sha256

loosebazooka:
This PR adds support for manifest lists when referenced by sha256 as described in #1360 

This PR:
- does not allow the user to specify any platform information in their configuration. 
- will still delegate to the registry default when using a non-sha256 tag
- also makes no changes to offline mode - that should be handled automatically


jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
```diff
@@ -89,6 +90,24 @@ public void testHandleResponse_v22()
...
+
+    Mockito.when(mockResponse.getBody()).thenReturn(v22ManifestList);
+    ManifestTemplate manifestTemplate =
+        new ManifestPuller<>(
+                fakeRegistryEndpointRequestProperties,
+                "test-image-tag",
+                V22ManifestListTemplate.class)
+            .handleResponse(mockResponse);
+
+    Assert.assertThat(manifestTemplate, CoreMatchers.instanceOf(V22ManifestListTemplate.class));
```
chanseokoh: Maybe it's good to test one or two values, since I think there is a possibility this test may pass even if `v22manifest_list.json` is not a manifest list, knowing unknown fields will be ignored.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
```diff
@@ -89,6 +90,24 @@ public void testHandleResponse_v22()
...
+    Path v22ManifestListFile =
+        Paths.get(Resources.getResource("core/json/v22manifest_list.json").toURI());
+    InputStream v22ManifestList = new ByteArrayInputStream(Files.readAllBytes(v22ManifestListFile));
+
+    Mockito.when(mockResponse.getBody()).thenReturn(v22ManifestList);
+    ManifestTemplate manifestTemplate =
+        new ManifestPuller<>(
+                fakeRegistryEndpointRequestProperties,
+                "test-image-tag",
+                V22ManifestListTemplate.class)
```
chanseokoh: I think it's good to also test when this is `ManifestTemplate.class`, which goes into a bit more interesting branch to autodetect the manifest template type.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPullerIntegrationTest.java
```diff
@@ -58,6 +64,38 @@ public void testPull_v22() throws IOException, RegistryException, InterruptedExc
     Assert.assertTrue(v22ManifestTemplate.getLayers().size() > 0);
   }
 
+  @Test
+  public void testPull_v22ManifestList()
+      throws IOException, RegistryException, InterruptedException {
+    localRegistry.pullAndPushToLocal("busybox", "busybox");
```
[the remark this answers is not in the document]
loosebazooka: Yeah that was some bad copy pasta :\

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -258,6 +272,25 @@ private Image pullBaseImage(
...
 
+  /**
+   * Looks through a manifest list for any amd64/linux manifest and downloads and returns the first
+   * manifest it finds.
+   */
+  private ManifestTemplate obtainPlatformSpecificImageManifest(
+      RegistryClient registryClient, V22ManifestListTemplate manifestListTemplate)
+      throws RegistryException, IOException {
+
+    List<String> digests = manifestListTemplate.getDigestsForPlatform("amd64", "linux");
```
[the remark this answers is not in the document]
loosebazooka: So obtainPlatformSpecificManifestList is only triggered when the user directly references a manifest list by sha256 (and not if they use a regular tag)... so we're basically defining the behavior here to be amd64/linux

jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
loosebazooka: Makes sense.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
loosebazooka: I think the media type should catch that, but I can play around with this test some more.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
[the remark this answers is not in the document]
briandealwis: From what I recall, requesting a manifest media-type for a SHA that is actually a manifest list will cause the registry to return the manifest for Linux/amd64.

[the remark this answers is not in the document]
loosebazooka: So it only returns a manifest list if we give it multiple accept types?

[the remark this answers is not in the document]
loosebazooka: Okay, so I tested this and it looks like it will always return a manifest list no matter what you "accept" if you use a sha256.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
chanseokoh: If a manifest class given is not `ManifestTemplate.class`, the `handleResponse()` code just boils down to

```java
      return JsonTemplateMapper.readJson(jsonString, manifestTemplateClass);
```

so I think it's possible that this test will succeed even if the input is a manifest (not a list), which also has `schemaVerison` (and `mediaType`). I think it's worth checking the existence of the `manifests` field.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPuller.java
```diff
@@ -71,7 +72,12 @@ public BlobHttpContent getContent() {
     if (manifestTemplateClass.equals(OCIManifestTemplate.class)) {
       return Collections.singletonList(OCIManifestTemplate.MANIFEST_MEDIA_TYPE);
     }
+    if (manifestTemplateClass.equals(V22ManifestListTemplate.class)) {
+      return Collections.singletonList(V22ManifestListTemplate.MANIFEST_MEDIA_TYPE);
+    }
 
+    // why is V22ManifestListTemplate missing in the default behavior? We don't explicitly accept
+    // it, we only handle it if referenced by sha256 (see getManifestTemplateFromJson)
```
briandealwis: This comment is confusing.  Is this a dialogue with the reader? Or a todo/wtf?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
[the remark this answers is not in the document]
chanseokoh: I think the discussion digressed. It is not about whether we requested a manifest list by SHA. Here, we already got a manifest list, and we are to pick some manifest out of the list to return. The question is whether we always return a Linux/amd64 manifest (and fail if there isn't) or simply the first one in the list.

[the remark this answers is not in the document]
briandealwis: Here's what I discovered earlier:
> ok you made me dig into it: the relevant code is here: https://github.com/docker/distribution/blob/0dae0957e5fe3156c265d22bef4cba9efbf388e2/registry/handlers/manifests.go#L190 
> - if the client requested a SHA1 that maps to a manifest-list but they don't support manifest-lists, then it returns a manifest corresponding to the defaultOS/defaultArch which is defined at the top as being linux/amd64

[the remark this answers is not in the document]
briandealwis: I've updated that comment: it's not requesting a SHA1 but a tag

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPuller.java
briandealwis: This intention is that we'll only include `V22ManifestListTemplate` in the defaults here when we offer support for building for an alternative platform?

loosebazooka: Right, I can update the comment to be more clear.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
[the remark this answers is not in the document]
loosebazooka: I also I think I just understood Chanseok's initial question. Docker will pull for the platform it is running on in the case of a manifest list.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPuller.java
```diff
@@ -118,10 +125,6 @@ private T getManifestTemplateFromJson(String jsonString)
       throw new UnknownManifestFormatException("Cannot find field 'schemaVersion' in manifest");
     }
 
-    if (!manifestTemplateClass.equals(ManifestTemplate.class)) {
```
loosebazooka: This code would unsafely create template objects that didn't first validate schema and mediatype. So now we always validate the value and then create the objects. Any mismatch will trigger a ClassCastException (indicating programming error or registry error).

chanseokoh: I remember something. This is about whether the registry client API should support the usage of `pullManifest(imageTag, templateClass)`: https://github.com/GoogleContainerTools/jib/blob/v1.3.0-maven/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java#L196-L197

Our plugins don't make use of it, but as an API, it may make sense to keep it.

If we remove this, we can actually clean up a lot more. From what I recall, `ManfiestPuller` won't need to accept the `manifestTemplateClass` argument.

chanseokoh: And it makes sense to update [this Javadoc](https://github.com/GoogleContainerTools/jib/blob/v1.3.0-maven/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java#L190-L191) too.

loosebazooka: You can still verify that you're actually pulling something you wanted... in which case ManifestPuller should probably still accept a template class. Either way, I think we should do it in a new PR.

chanseokoh: Agreed, we can do it later. (It is basically the question of whether we maintain the `RegistryClient` API surface in a sane manner. A library throwing a `ClassCastException` won't look nice. But we aren't even sure if we will ever release this as a library.)

jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
chanseokoh: Can we check the `manifests` field?

jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
chanseokoh: Still pending?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPuller.java
loosebazooka: Ah yeah, but a ClassCastException should only happen if something went horribly wrong.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/ManifestPullerTest.java
loosebazooka: done:https://github.com/GoogleContainerTools/jib/pull/1811/files#diff-fabb196f84239d7e62118d3d9e66aa89R112

PR #1814, opened 2019-06-27
Remove ImageLayers in Image

chanseokoh:
`ImageLayer` had a useful feature to remove duplicate layers, but I realized we don't use it anywhere. Looks like we can use a plain list and remove the class.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
```diff
@@ -43,31 +44,25 @@
...
-    int layerCount = buildConfiguration.getLayerConfigurations().size();
+    List<LayerConfiguration> layerConfigurations = buildConfiguration.getLayerConfigurations();
 
     try (ProgressEventDispatcher progressEventDispatcher =
             progressEventDispatcherFactory.create(
-                "preparing application layer builders", layerCount);
+                "preparing application layer builders", layerConfigurations.size());
         TimerEventDispatcher ignored =
             new TimerEventDispatcher(buildConfiguration.getEventHandlers(), DESCRIPTION)) {
-      ImmutableList.Builder<BuildAndCacheApplicationLayerStep> buildAndCacheApplicationLayerSteps =
```
chanseokoh: Note: this is a simple cleanup, unrelated to this issue.

PR #1817, opened 2019-06-28
Update changelog for manifest lists

loosebazooka:
wondering if I should also update the plugin changelogs?

### DIFF
--- a/jib-core/CHANGELOG.md
+++ b/jib-core/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


chanseokoh:
Yeah, it's a new feature for plugins too.

jib-core/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- Manifest lists referenced directly by sha256 are automatically parsed and the first `linux/amd64` manifest is used. ([#1811](https://github.com/GoogleContainerTools/jib/issues/1811))
```
chanseokoh: nit: we don't use a period.

PR #1818, opened 2019-06-28
Allow setting file timestamps from plugins

### LINKED ISSUE #1608: Allow setting file timestamps from plugins

chanseokoh:
Migrated from #1782. Contains commits from an external contributor, so CLA bot keeps complaining. Manaully marking CLA OK.

Fixes #1608.

```gradle
// ISO 8601 date time format
container.filesModificationTime = '2011-12-03T10:15:30+09:00'
```
```gradle
container.filesModificationTime = 'EPOCH_PLUS_SECOND` // default
```

```xml
<container>
  <filesModificationTime>2011-12-03T10:15:30Z</filesModificationTime>
</container>
```

I noticed the target file modification time should be factored in when determining if an application layer for the files were cached, so I modified `LayerEntriesSelector` accordingly.


chanseokoh:
Graduated from the Draft status. Ready for review.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
```diff
@@ -83,10 +87,15 @@ public int compareTo(LayerEntryTemplate otherLayerEntryTemplate) {
       if (extractionPathComparison != 0) {
         return extractionPathComparison;
       }
-      int lastModifiedTimeComparison =
-          lastModifiedTime.compareTo(otherLayerEntryTemplate.lastModifiedTime);
-      if (lastModifiedTimeComparison != 0) {
-        return lastModifiedTimeComparison;
+      int sourceLastModifiedTimeComparison =
```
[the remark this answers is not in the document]
chanseokoh: The order can be arbitrary, as the purpose is to be able to determine if two layers (list of `LayerEntries`) are "same" or not (in the sense whether they would lead to the same layer tarball.)

*However*, unrelated, now I think more about this, I think adding source file timestamps here as another criterion for checking equality may have a significant negative performance impact. That is, whenever a file is touched but not actually modified (e.g., copied the same resource file into the build directory), the cache will regard it as a different file. But the source file timestamps are only meaningful for `KEEP_ORIGINAL`, so for the default behavior and for setting a fixed timestamp over all files, I think this will result in a huge overhead as the cache will create a new tarball only to find out later that the layer already exists.

I will have to think about this a bit more. This could be tricky to optimize.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -175,6 +177,8 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
   private RelativeUnixPath dependenciesDestination = RelativeUnixPath.get("libs");
   private RelativeUnixPath othersDestination = RelativeUnixPath.get("classpath");
   @Nullable private String mainClass;
+  private BiFunction<Path, AbsoluteUnixPath, Instant> lastModifiedTimeProvider =
```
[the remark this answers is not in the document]
chanseokoh: I think the interface works too and could be better. I just wanted to introduce minimal changes to retain the current public API, as our other existing code is using `BiFunction`. For example, the current `LayerConfiguration` code has

```java
  /** Default file modification time (EPOCH + 1 second). */
  public static final Instant DEFAULT_MODIFIED_TIME = Instant.ofEpochSecond(1);

  public static final BiFunction<Path, AbsoluteUnixPath, Instant> DEFAULT_MODIFIED_TIME_PROVIDER =
      (sourcePath, destinationPath) -> DEFAULT_MODIFIED_TIME;

    ...

    public Builder addEntryRecursive(
        Path sourceFile,
        AbsoluteUnixPath pathInContainer,
        BiFunction<Path, AbsoluteUnixPath, FilePermissions> filePermissionProvider,
        BiFunction<Path, AbsoluteUnixPath, Instant> lastModifiedTimeProvider)
        throws IOException {
```

As least this PR doesn't change the public API. But we can switch to an interface anytime.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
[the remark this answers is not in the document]
loosebazooka: So `target` is the only one that matters?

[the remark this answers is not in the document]
chanseokoh: No, source matters too, because for `KEEP_ORIGINAL`, the files of the cached layer (a tarball) are set with the timestamps of the source files in that case. And if the files in the layer should have different timestamps (e.g., when the user changes the configuration to `EPOCH_PLUS_SECOND` or sets a fixed timestamp, or the source timestamp changed while using `KEEP_ORIGINAL`), Jib does need to create a new cached layer. So, for these scenarios, the source file timestamps should be factored into determining whether the layers are same or different, and this is correct.

However, suppose the user never uses `KEEP_ORIGINAL`. Then we never look into source file timestamps, hence no need to factor them into the difference computation at all. However, this PR always take source timestamps into account. Therefore, for example, whenever the timestamp of a `.class` file is touched but the content remains exactly same, Jib will always create a new tarball first only to find out it did not have to. In this scenario where `KEEP_ORIGINAL` is never used, source timestamps do not play any role in creating a layer tarball. Did I make it clear?

chanseokoh:
I think for now, I'll get rid of anything related to `KEEP_ORIGINAL` and revert changes to the selector. I honestly think `KEEP_ORIGINAL` is not really useful in many cases. For example, it will create new images every time you do a clean build even when not changing an app. We can reconsider this option later if there are use-cases, but I think even such cases they should set a fixed timestamp dynamically and in a reproducible manner. Will work on gutting out `KEEP_ORIGINAL`.

chanseokoh:
Removed anything related to `KEEP_ORIGINAL`. Good for review.

Will update CHANGELOG in a follow-up.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -366,6 +376,11 @@ public void testCreateContainerBuilder_correctFiles()
             applicationDirectory.resolve("classes/HelloWorld.class"),
             applicationDirectory.resolve("classes/some.class")),
         layers.classesLayerEntries.get(0).getLayerEntries());
+
+    assertModificationTime(Instant.ofEpochSecond(32), layers.snapshotsLayerEntries);
+    assertModificationTime(Instant.ofEpochSecond(32), layers.dependenciesLayerEntries);
+    assertModificationTime(Instant.ofEpochSecond(32), layers.resourcesLayerEntries);
+    assertModificationTime(Instant.ofEpochSecond(32), layers.classesLayerEntries);
```
briandealwis: I'd prefer this magic value be a constant — I was a bit surprised by this and had to go hunting to find the source.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -544,33 +557,40 @@ public void testCreateContainerBuilder_noErrorIfWebInfClassesDoesNotExist()
...
       throws IOException, InvalidImageReferenceException, CacheDirectoryCreationException {
     temporaryFolder.newFolder("jib-exploded-war");
     setUpWarProject(temporaryFolder.getRoot().toPath());
-    setupBuildConfiguration(JavaContainerBuilder.DEFAULT_WEB_APP_ROOT); // should pass
+    setupBuildConfiguration("/anything"); // should pass
   }
 
   private BuildConfiguration setupBuildConfiguration(String appRoot)
       throws InvalidImageReferenceException, IOException, CacheDirectoryCreationException {
+    return setupBuildConfiguration(appRoot, DEFAULT_CONTAINERIZING_MODE);
```
briandealwis: Given that this is a test class where we're explicitly testing both _exploded_ and _packaged_ modes, then this  choice should be explicit within each test method.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleRawConfigurationTest.java
```diff
@@ -68,6 +68,7 @@ public void testGetters() {
     Mockito.when(containerParameters.getPorts()).thenReturn(Arrays.asList("80/tcp", "0"));
     Mockito.when(containerParameters.getUseCurrentTimestamp()).thenReturn(true);
     Mockito.when(containerParameters.getUser()).thenReturn("admin:wheel");
+    Mockito.when(containerParameters.getFilesModificationTime()).thenReturn("modification time");
```
briandealwis: `"modification time"` seems an odd choice given that the other values returned here are legitimate values.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -484,6 +488,19 @@ public JavaContainerBuilder setMainClass(String mainClass) {
...
   }
 
+  /**
+   * Sets the last modified time provider for container files.
+   *
+   * @param lastModifiedTimeProvider a provider that takes a source path and destination path on the
+   *     container and returns the file modification time that should be set for that path
+   * @return this
+   */
+  public JavaContainerBuilder setLastModifiedTimeProvider(
```
briandealwis: `setModificationTimeProvider()`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
briandealwis: `modificationTimeProvider`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -175,6 +177,8 @@ public static JavaContainerBuilder from(RegistryImage registryImage) {
   private RelativeUnixPath dependenciesDestination = RelativeUnixPath.get("libs");
   private RelativeUnixPath othersDestination = RelativeUnixPath.get("classpath");
   @Nullable private String mainClass;
+  private BiFunction<Path, AbsoluteUnixPath, Instant> lastModifiedTimeProvider =
+      LayerConfiguration.DEFAULT_MODIFIED_TIME_PROVIDER;
```
briandealwis: `DEFAULT_MODIFICATION_TIME_PROVIDER`

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -643,7 +660,8 @@ private void addFileToLayer(
     if (!layerBuilders.containsKey(layerType)) {
       layerBuilders.put(layerType, LayerConfiguration.builder());
     }
-    layerBuilders.get(layerType).addEntry(sourceFile, pathInContainer);
+    Instant modifiedTime = lastModifiedTimeProvider.apply(sourceFile, pathInContainer);
```
briandealwis: either `modificationTime` or `lastModifiedTime`

jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerConfiguration.java
```diff
@@ -99,17 +99,34 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
      */
     public Builder addEntry(
         Path sourceFile, AbsoluteUnixPath pathInContainer, FilePermissions permissions) {
+      return addEntry(sourceFile, pathInContainer, permissions, DEFAULT_MODIFIED_TIME);
```
briandealwis: `DEFAULT_MODIFICATION_TIME`

jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerConfiguration.java
```diff
@@ -99,17 +99,34 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
...
+
+    /**
+     * Adds an entry to the layer with the given file modification time. Only adds the single source
+     * file to the exact path in the container file system. See {@link Builder#addEntry(Path,
+     * AbsoluteUnixPath)} for more information.
+     *
+     * @param sourceFile the source file to add to the layer
+     * @param pathInContainer the path in the container file system corresponding to the {@code
+     *     sourceFile}
+     * @param lastModifiedTime the file modification timestamp
```
briandealwis: Given the description, maybe `modificationTime` would be better?

jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerEntry.java
```diff
@@ -55,8 +55,7 @@
    * @param extractionPath the path in the container file system corresponding to the {@code
    *     sourceFile}
    * @param permissions the file permissions on the container
-   * @param lastModifiedTime the file modification time, default to 1 second since the epoch
-   *     (https://github.com/GoogleContainerTools/jib/issues/1079)
+   * @param lastModifiedTime the file modification time
```
briandealwis: Given the description, maybe `modificationTime` would be better?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -122,7 +125,8 @@ public void testPluginConfigurationProcessor_defaults()
       throws InvalidImageReferenceException, IOException, CacheDirectoryCreationException,
           MainClassInferenceException, InvalidAppRootException, InvalidWorkingDirectoryException,
           InvalidContainerVolumeException, IncompatibleBaseImageJavaVersionException,
-          NumberFormatException, InvalidContainerizingModeException {
+          NumberFormatException, InvalidContainerizingModeException,
+          InvalidFilesModificationTimeException {
```
briandealwis: I wonder if we should have a `JibConfigurationException` base class.

chanseokoh: Yeah, it is about time to do something about handling invalid config values. I'll create an issue as a tech debt. Hopefully we get to it soon.

loosebazooka:
Just so I'm clear, we only allow two options now?
`epoch+1` and `some fixed time`? What happens when someone implements a custom timestamp provider, does our selector break?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -134,10 +132,7 @@ private static boolean isProgressFooterEnabled(Project project) {
 
   @Override
   public JibContainerBuilder createContainerBuilder(
```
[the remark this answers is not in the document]
chanseokoh: Hmm... yeah, it feels odd. It does make sense to pass necessary initialize arguments instead of passing `JavaContainerBuilder` as before. But at the same time, I don't want to pass too many arguments. At least I'll rename it to `createJibContainerBuilder`. If we end up disliking it, we can go back.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -186,6 +184,8 @@ public void set(String image) {
     @Nullable @Parameter private String user;
 
     @Nullable @Parameter private String workingDirectory;
+
+    private String filesModificationTime = "EPOCH_PLUS_SECOND";
```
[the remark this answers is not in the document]
chanseokoh: Good catch! It worked without it somehow, but we should add it.

chanseokoh:
> What happens when someone implements a custom timestamp provider, does our selector break?

Jeepers... this made me ponder again how the cache works, and I realized it was right to modify `LayerEntriesSelector`. A few days ago, I was confused and mistakenly thought `LayerEntriesSelector` was originally factoring in the target mod time and not the source file mod time, but it was the opposite. We've always been checking the mod tile of source files. Now that the target mod tile is configurable, the cache should factor that in too. In sum, I'll restore the `LayerEntriesSelector`. (However, I won't revive `KEEP_ORIGINAL`, which I think hurts reproducibility a lot.)

jib-core/src/main/java/com/google/cloud/tools/jib/cache/LayerEntriesSelector.java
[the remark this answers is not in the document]
chanseokoh: > I think adding source file timestamps here as another criterion for checking equality may have a significant negative performance impact.

The cache was already doing this. I was briefly confused. I'm restoring the `LayerEntriesSelector` changes (https://github.com/GoogleContainerTools/jib/pull/1818#issuecomment-515184225).

chanseokoh:
Restored the changes to `LayerEntriesSelector` to factor in target modification time when computing a hash. Sorry for going back and forth. This should be the final stable state.

jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerConfiguration.java
```diff
@@ -167,7 +184,10 @@ public Builder addEntryRecursive(
         BiFunction<Path, AbsoluteUnixPath, FilePermissions> filePermissionProvider)
         throws IOException {
       return addEntryRecursive(
-          sourceFile, pathInContainer, filePermissionProvider, DEFAULT_MODIFIED_TIME_PROVIDER);
+          sourceFile,
+          pathInContainer,
+          DEFAULT_FILE_PERMISSIONS_PROVIDER,
```
TadCordle: Shouldn't this be `filePermissionProvider`?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -484,6 +488,19 @@ public JavaContainerBuilder setMainClass(String mainClass) {
...
   }
 
+  /**
+   * Sets the modification time provider for container files.
+   *
+   * @param modificationTimeProvider a provider that takes a source path and destination path on the
+   *     container and returns the file modification time that should be set for that path
+   * @return this
+   */
+  public JavaContainerBuilder setModificationTimeProvider(
```
TadCordle: I'm not sure how I feel about adding methods like these to `JavaContainerBuilder`. I feel like it's best to keep this class as simple as possible and delegate the non-java-specific stuff to `JibContainerBuilder` or something else, but I guess this is necessary since the timestamp provider needs to be applied as each file is added?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -59,7 +59,7 @@
     /**
      * Adds a {@link TarArchiveEntry} if its extraction path does not exist yet. Also adds all of
      * the parent directories on the extraction path, if the parent does not exist. Parent will have
-     * modified time to set to {@link LayerConfiguration#DEFAULT_MODIFIED_TIME}.
+     * modification time to set to {@link LayerConfiguration#DEFAULT_MODIFICATION_TIME}.
```
TadCordle: modification time ~~to~~ set to

jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerConfiguration.java
chanseokoh: Gosh, nice catch. Glad you got this.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -135,7 +135,8 @@ private static void assertExtraDirectoryDeprecationWarning(String pomXml)
     BuildResult buildResult =
         JibRunHelper.buildToDockerDaemon(simpleTestProject, targetImage, pomXml);
     Assert.assertEquals(
-        "Hello, world. \nrw-r--r--\nrw-r--r--\nfoo\ncat\n",
+        "Hello, world. \n1970-01-01T00:00:01Z\nrw-r--r--\nrw-r--r--\nfoo\ncat\n"
```
TadCordle: The output for this test program is starting to get more and more complicated, I wonder if we should think about making it more readable (not this PR)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -164,6 +165,13 @@ public void buildDocker()
       throw new GradleException(
           "container.volumes is not an absolute Unix-style path: " + ex.getInvalidVolume(), ex);
 
+    } catch (InvalidFilesModificationTimeException ex) {
+      throw new GradleException(
+          "container.filesModificationTime should be an ISO 8601 date-time (see "
```
TadCordle: `jib.container.filesModificationTime`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -144,6 +145,13 @@ public void buildImage()
       throw new GradleException(
           "container.volumes is not an absolute Unix-style path: " + ex.getInvalidVolume(), ex);
 
+    } catch (InvalidFilesModificationTimeException ex) {
+      throw new GradleException(
+          "container.filesModificationTime should be an ISO 8601 date-time (see "
```
TadCordle: `jib.container.filesModificationTime`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -162,6 +163,13 @@ public void buildTar()
       throw new GradleException(
           "container.volumes is not an absolute Unix-style path: " + ex.getInvalidVolume(), ex);
 
+    } catch (InvalidFilesModificationTimeException ex) {
+      throw new GradleException(
+          "container.filesModificationTime should be an ISO 8601 date-time (see "
```
TadCordle: `jib.container.filesModificationTime`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
chanseokoh: That's better, but all the other error messages around this code do not start with `jib.container`, so I think I should go without it for consistency here.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
chanseokoh: Yeah, files are added recursively inside `JavaContainerBuilder`, so I also find passing a provider necessary in its current state.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
chanseokoh: I felt the same. It's growing duplicates unnecessarily.

PR #1813, opened 2019-06-27
Introduce support for enforcing minimum Jib plugin versions

briandealwis:
Jib provides some internal goals and properties to support Skaffold’s Jib binding for Maven and Gradle. Some of the improvements to this binding cannot be made backwards compatible, and may require the user to update their version of Jib.  We need some way to detect and enforce these version requirements.

This PR:
  1. Adds support to Jib's goals and tasks to define an expected or required version range as specified through a new property `jib.requiredVersion`.  This property can take two forms: a minimum version number (e.g., `1.4.0`) or a Maven-like version-range (`[1.4,2.0)`).
  2. Changes the Gradle `JibPlugin#apply()` to perform a version check if `jib.requiredVersion` is set.
  3. Changes the various Maven goals to perform a version check if `jib.requiredVersion` is set.
  4. Adds a new Maven goal `_skaffold-ensure-jib-up-to-date` and corresponding Gradle task `_skaffoldEnsureJibUpToDate` that fail if `jib.requiredVersion` is not set.  These goals will be called by Skaffold to fail and thus prompt users to update to a newer version of Jib.  Ideallyt these goals could be removed once we're confident there are no users of older Jib or Skaffold.

PR #1821, opened 2019-07-03
Remove useless version field from skaffold-files-v2

briandealwis:
Removes the `version` field from the `_skaffold-files-v2` / `_skaffoldFilesV2` output added in #1714.  This field was added in support of GoogleContainerTools/skaffold#2122 but unfortunately was proved inadequate.  We're instead introducing explicit version checks within Jib with #1813.


briandealwis:
There are no compatibility concerns from its removal: the field was introduced with Jib 1.3.0, was intended for internal use only, and was never actually used.

PR #1831, opened 2019-07-08
Leverage NullAway checks with Optional for registry client framework

chanseokoh:
Currently, the registry client framework has `callRegistryEndpoint()`, `call???()`, `handleResponse()`, etc. always return `@Nullable` generic results. Not only did this lead to code like below to manually suppress null warnings, but this also renders the NullAway checks largely useless.

```java
    T manifestTemplate = callRegistryEndpoint(manifestPuller);
    if (manifestTemplate == null) {
      throw new IllegalStateException("ManifestPuller#handleResponse does not return null");
    }
``` 
```java
    return Verify.verifyNotNull(callRegistryEndpoint(...));
```
```java
        URL putLocation = callRegistryEndpoint(...);
        Preconditions.checkNotNull(putLocation);
```

`Optional` comes to the rescue, as in #1804.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/AuthenticationMethodRetriever.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/AuthenticationMethodRetriever.java
@@ -57,7 +57,7 @@ class AuthenticationMethodRetriever
    * The request did not error, meaning that the registry does not require authentication.
    *
    * @param response ignored
-   * @return the authenticator
+   * @return {@link Optional#empty()}
    */
   @Override
   public Optional<RegistryAuthenticator> handleResponse(Response response) {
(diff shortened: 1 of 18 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/AuthenticationMethodRetriever.java
```diff
@@ -55,12 +57,11 @@ public BlobHttpContent getContent() {
    * The request did not error, meaning that the registry does not require authentication.
    *
    * @param response ignored
-   * @return {@code null}
+   * @return the authenticator
```
TadCordle: Should we be more specific here since this isn't really returning anything?

chanseokoh: Oh, oops. Fixed the Javadoc.

PR #1832, opened 2019-07-09
Simplify DockerClient

TadCordle:
Re-ordered the methods and removed the builder pattern since it seems unnecessary.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
@@ -35,102 +35,6 @@ import java.util.function.Function;
 /** Calls out to the {@code docker} CLI. */
 public class DockerClient {
 
-  /** Builds a {@link DockerClient}. */
-  public static class Builder {
-
-    private Path dockerExecutable = DEFAULT_DOCKER_CLIENT;
-    private ImmutableMap<String, String> dockerEnvironment = ImmutableMap.of();
-
-    private Builder() {}
-
-    /**
-     * Sets a path for a {@code docker} executable.
-     *
-     * @param dockerExecutable path to {@code docker}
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -111,51 +85,26 @@ public static DockerClient newDefaultClient() {
     };
   }
 
-  private static final Path DEFAULT_DOCKER_CLIENT = Paths.get("docker");
+  public static final Path DEFAULT_DOCKER_CLIENT = Paths.get("docker");
```
chanseokoh: Should be moved to the top of the class.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -111,51 +85,26 @@ public static DockerClient newDefaultClient() {
...
-
-  /**
-   * Checks if Docker is installed on the user's system and accessible by running the default {@code
-   * docker} command.
-   *
-   * @return {@code true} if Docker is installed on the user's system and accessible
-   */
-  public static boolean isDefaultDockerInstalled() {
-    return isDockerInstalled(DEFAULT_DOCKER_CLIENT);
+  public DockerClient(Map<String, String> dockerEnvironment, Path dockerExecutable) {
```
chanseokoh: Perhaps switch the order of the arguments. `defaultProcessBuilderFactory` takes them in the opposite order. I think `dockerExecutable` path is more important to be first.

TadCordle: Oh yeah, I'd switched them because I made one of them nullable, but then changed my mind and forgot to switch them back.

PR #1835, opened 2019-07-10
Move to unified gradle build (WIP

loosebazooka:
So I've spent a lot of time solving stupid issues and messing around with this build to get it working on the command line and in the IDEs.

Things that I have not tested yet/don't current work:
- maven pom generation
- maven integration tests
- no change to kokoro or release scripts

What I want to start off with is ensuring this works in everyones' environments. So if you can download this branch and test it on your machine (eclipse especially) that would be nice.


build.gradle
```diff
@@ -0,0 +1,232 @@
...
+      if (dependencyProject.configurations.hasProperty('api')) {
+        implementation dependencyProject.configurations.api.dependencies
+      }
+      // if we find any project dependencies that are brought in transitively, go remove them
+      project.sourceProjects.each { projectToRemove ->
+        project.configurations.implementation.dependencies.remove projectToRemove
+      }
+    }
+    // keep track of all dependencyProjects for later removal
+    sourceProjects += dependencyProject
```
chanseokoh: Doesn't seem like this is used later. No point of adding `dependencyProject`?

loosebazooka: Line 216 is using it. `Project.sourceProjects`

chanseokoh:
One more: I occasionally hit the following "plugin metadata file absent" error with the old setup, and refreshing Gradle projects used to clear them. (I don't know when exactly this happens.)

```
org.gradle.testkit.runner.InvalidPluginMetadataException: Test runtime classpath does not contain plugin metadata file 'plugin-under-test-metadata.properties'
```

I am seeing the same error with the new setup, and it does not go away. The file is generated at `jib-gradle-plugin/build/pluginUnderTestMetadata/plugin-under-test-metadata.properties` in the old setup. I cannot find the file in the new setup and don't know how I (or Gradle or the IDE) can generate it. I don't really know much about this file, but [this doc](https://docs.gradle.org/current/javadoc/org/gradle/testkit/runner/GradleRunner.html#withPluginClasspath--) says that the `java-gradle-plugin` generates the file. But I see `jib-gradle-plugin/build.gradle` applies the plugin.

loosebazooka:
Plugin under test metadata is a gradle task that must be run so that tests that apply the plugin work. I think in eclipse you'll have to have some pre-test task that executes that goal.

chanseokoh:
Okay, I manually ran `./gradlew pluginUnderTestMetadata` at the root (the `jib/` directory), which created `jib-gradle-plugin/build/pluginUnderTestMetadata/plugin-under-test-metadata.properties`. Then I manually added `jib-gradle-plugin/build/pluginUnderTestMetadata` as an extra classpath folder for my test run configuration in Eclipse, and it works. Something unnecessary in the old setup. I'll be a bit of a pain if I have to do this manual addition for every test run configuration. Not sure if this is a bug in the Eclipse Gradle plugin.

loosebazooka:
Plugin under test metadata is always required. I'm a little confused why it worked before for you in eclipse. I'll take a look again

loosebazooka:
@chanseokoh @briandealwis any idea why the JIB_INTEGRATION_TESTING_PROJECT is not being picked up?

loosebazooka:
never mind I'm a fool, we don't do integration tests for presubmit.

jib-core/kokoro/release_build.sh
```diff
@@ -5,5 +5,5 @@ set -e
 # Display commands to stderr.
 set -x
 
-cd github/jib/jib-core
-./gradlew prepareRelease
+cd github/jib
+./gradlew :jib-core:prepareRelease
```
loosebazooka: I'm not actually convinced this matches up with our release scripts, I'll have to see.

build.gradle
```diff
@@ -0,0 +1,314 @@
...
+
+  sourceCompatibility = JavaVersion.VERSION_1_8
+  targetCompatibility = JavaVersion.VERSION_1_8
+  compileJava.options.encoding = 'UTF-8'
+
+  /* PROJECT DEPENDENCY CONTRAINTS */
+  // define all versioned dependencies here and apply in subprojects as necessary without version
+  dependencies {
+    constraints {
+      implementation 'com.google.http-client:google-http-client:1.27.0'
```
chanseokoh: #1882 will upgrade this to 1.31.0 and add another related dependency.

loosebazooka:
some of these changes can probably be broken out into unrelated fixes, I'll do that to make this isolated to the build stuff.

loosebazooka:
need to factor in #1886 dependency.

jib-gradle-plugin/build.gradle
```diff
@@ -185,5 +67,5 @@ pluginBundle {
     }
   }
 }
-tasks.publishPlugins.dependsOn integrationTest
+//tasks.publishPlugins.dependsOn integrationTest
```
loosebazooka: I need to fix this

TadCordle:
We'll probably need to update CONTRIBUTING.md as well.

loosebazooka:
This should be ready now to look at.

CONTRIBUTING.md
```diff
@@ -63,34 +63,24 @@ integration tests will create local registries on ports 5000 and 6000.
...
-appear both in the originating project _and_ in the subprojects,
-and can interfere with Eclipse's incremental compilation.
+Jib is a mix of Gradle and Maven projects, however we build everything using one
+unifed gradle build. There is special code to include some projects directly as
+source, but importing your project should be pretty straight forward.
 
-The following instructions first import the projects, and then replaces
-the source-set-style references to use Eclipse's project references
-instead.
+// TODO: Check with chanseok/brian on how importing into eclipse will work
```
loosebazooka: @chanseokoh @briandealwis despite my best efforts, eclipse refuses to install on my machine. Need your help verifying how to load the project into eclipse.

jib-core/build.gradle
```diff
@@ -1,69 +1,20 @@
...
-  implementation 'com.fasterxml.jackson.core:jackson-databind:2.9.9.2'
-  implementation 'org.ow2.asm:asm:7.0'
-
-  testImplementation 'junit:junit:4.12'
-  testImplementation 'org.mockito:mockito-core:2.23.4'
-  testImplementation 'org.slf4j:slf4j-api:1.7.25'
-  testImplementation 'com.github.stefanbirkner:system-rules:1.19.0'
+  implementation "com.google.http-client:google-http-client:${dependencyVersions.GOOGLE_HTTP_CLIENT}"
+  implementation "com.google.http-client:google-http-client-apache-v2:${dependencyVersions.GOOGLE_HTTP_CLIENT_APACHE_V2}"
+  implementation "org.apache.commons:commons-compress:${dependencyVersions.COMMONS_COMPRESS}"
```
chanseokoh: [#1902](https://github.com/GoogleContainerTools/jib/pull/1902/files#diff-2c8a5b94f9299ef70d5acb27fae2bfd8R46-R47) adds `google-auth-library-oauth2-http` and upgrades Guava version. Can you update them here?

chanseokoh: Ah, wait. I think it isn't required in `jib-core` but in `jib-plugins-common`. Let me check this again.

loosebazooka: will update when it merges

chanseokoh: Wrong. It is required in `jib-core`. Can you add it and upgrade Guava?

CONTRIBUTING.md
```diff
@@ -63,34 +63,24 @@ integration tests will create local registries on ports 5000 and 6000.
...
 
   1. Ensure you have installed the Gradle tooling for Eclipse, called
      _Buildship_ (available from [the Eclipse
      Marketplace](https://marketplace.eclipse.org/content/buildship-gradle-integration)).
-  2. **Import the Maven project:** Use _File &rarr; Open Projects from File System..._
```
TadCordle: I know it renders correctly, but since you deleted step 2 the numbering in the plain text is off.

CONTRIBUTING.md
```diff
@@ -17,16 +17,19 @@ again.
...
-To build, use the provided `build.sh` which builds and tests each of the
-components into your local `~/.m2/repository`.  Note that this script does
-not run integration tests.
+And 1 internal component:
+
+  - `jib-plugins-common`: a library with helpers for plugin builders
+
+The project is configured as a single gradle build, to build the whole
+project run `./gradlew build`, to install into the local maven repository
+run `./gradlew install`.
```
TadCordle: This might be punctuated a bit more clearly: "The project is configured as a single gradle build. Run `./gradlew build` to build the whole project, or run `./gradlew install` to install into the local maven repository."

CONTRIBUTING.md
```diff
@@ -63,34 +63,24 @@ integration tests will create local registries on ports 5000 and 6000.
...
 
-Jib is a mix of Gradle and Maven projects.  But since we do not
-publish all artifacts, we instead have the projects reference
-source from their dependencies directly using `sourceSets` and
-`build-helper-maven-plugin`.  This complicates importing these projects
-as the Gradle and Maven tooling for Eclipse faithfully replicate the
-direct source references.  This has a side-effect that classes
-appear both in the originating project _and_ in the subprojects,
-and can interfere with Eclipse's incremental compilation.
+Jib is a mix of Gradle and Maven projects, however we build everything using one
```
TadCordle:
```suggestion
Although Jib is a mix of Gradle and Maven projects, we build everything using one
```

chanseokoh:
Yeah, definitely this is a lot nicer. No more copying source directories, and I really like it.

For the record, the outcome of the most recent try on Eclipse:

> Now, I had to additionally configure extra build path through the IDE mechanism. jib-gradle-plugin and jib-maven-plugin could not see classes in jib-core, so I had to add jib-core as an additional dependent project for them. Interestingly, I did not have to add jib-plugins-common as an explicit project dependency. I wonder if this is also the case in IntelliJ.

This is still true, unfortunately. Not a big deal for me, as I only have to do this once and it's done.

> Okay, I manually ran `./gradlew pluginUnderTestMetadata` at the root (the `jib/` directory), which created `jib-gradle-plugin/build/pluginUnderTestMetadata/plugin-under-test-metadata.properties`.

This was probably due to my lack of understanding of this thing. The old setup had the same problem; nothing new. I learned running test on the command line just generates this file. Once generated, I'm good.

> Then I manually added `jib-gradle-plugin/build/pluginUnderTestMetadata` as an extra classpath folder for my test run configuration in Eclipse, and it works. Something unnecessary in the old setup.

No longer the case. No need to add an extra classpath. It works now. Or maybe I just misunderstood this part back then.

chanseokoh:
> test if Javadoc errors (correctly) fail the build

This may not be working. I added a wrong `@param` to a Javadoc of a class in `jib-maven-plugin`, but `./gradlew javadoc` doesn't fail.

loosebazooka:
@chanseokoh I'll look into the javadoc issue

loosebazooka:
@chanseokoh about the javadoc, are you sure the bad `@param` is on a public method? The gradle javadoc (maybe different than maven) only generates javadoc for public apis

chanseokoh:
Yeah, it does fail. Didn't realized it only works for public things.

TadCordle:
Is there more work to be done on this or is this good to go?

loosebazooka:
I gotta rebase on the 1.5.1 release, but other than that, yeah

PR #1837, opened 2019-07-11
Remove unused code / renaming methods

chanseokoh:
The code to validate some system properties are unused. It doesn't hurt to check them somewhere, but I think it doesn't provide much value.


chanseokoh:
Oops. They are being used.

PR #1840, opened 2019-07-11
Skip downloading base image layers if they exists in target repository

### LINKED ISSUE #1673: Skip downloading and caching base image layers (BLOb) if target registry already has them
For registry push (`jib:build`), I wonder if we can entirely skip downloading the base image (layer tarballs) by checking ahead if the target registry has all the layers. I think in most cases, the target registry will have the base image. Even if it doesn't, it will have them available as soon as Jib or anyone pushes the base image or any images based on it. (I assume a layer in a registry is shared registry-wide across all repositories/users.) Java base images are often very large, and I think this will particularly helpful on CI/CD environment.

Similarly, we may also be able to skip downloading the base image for Docker daemon push, if we could check if the daemon has some layer, but I think this is less valuable.

chanseokoh:
Fixes #1673.

This will give substantial speedup for CI/CD environments where people usually don't persist the Jib base image cache.

## Design

I tried to do intelligent, fine-grained BLOb checking on each layer. For example, if only a subset of the layers are missing in the target repo, it downloads and caches only those layers.

I also made it avoid checking the layer twice in the situation where it should push missing layers, because currently `PushBlobStep` always does a Blob check before pushing.

I spent a lot of time tinkering with several different designs and implementations in order to maintain generality and loose coupling between build steps and components, and I think I can settle with this design. So, other advantages from this are
- generality: any layer can be checked and skipped using the same framework with no change; whether or not the layer is from a base image doesn't matter (hence future proof)
- Therefore, checking/downloading/pushing layers can maintain the same level of parallelization as before. For example, it is possible that Jib is downloading one missing base image layer while pushing another missing base image layer as well as building/pushing application layers all at the same time. Potentially and theoretically in the future, the same story can be applied for application layers too. (I am saying theoretically here, because you have to cache applications layers once anyway.)

## Speedup

For a large image like `adoptopenjdk/openjdk9-openj9`,
```
[INFO] BUILD SUCCESS
[INFO] Total time:  19.681 s
[INFO] Finished at: 2019-07-11T17:53:30-04:00
```

```
[INFO] BUILD SUCCESS
[INFO] Total time:  3.420 s
[INFO] Finished at: 2019-07-11T17:53:36-04:00
```
(However, I admit `adoptopenjdk/openjdk9-openj9` is 466MB and way larger than `openjdk:8-slim` or Distroless.)

## Interaction with `--offline`

As pointed out before, not caching the base image layers may cause a problem for `--offline` when layers are missing. However, in reality, this would be extremely rare, because `--offline` can be used only with `jib:dockerBuild` and `jib:buildTar`; for those goals, Jib always downloads base image layers. And people do understand they have to do an online build at least once before they can do `--offline`.

But in an extremely rare case where the user
1. never did `jib:dockerBuild` or `jib:buildTar` on their local dev machine
1. and the base image exists in the target repo
1. but did `jib:build` which skipped downloading layers (i.e., manifest JSON is cached but layers are not)
1. now does `--offline jib:dockerBuild` without ever trying online `jib:dockerBuild`

then this case is still covered, as Jib will show the following error message.

`[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.3.1-SNAPSHOT:dockerBuild (default-cli) on project helloworld: Cannot run Jib in offline mode; local Jib cache for base image is missing image layer sha256:.... Rerun Jib in online mode with "-Djib.cacheBaseImage=always" to re-download the base image layers. -> [Help 1]`

(`-Djib.cacheBaseImage=always` is actually unnecessary because `jib:dockerBuild` and `jib:buildTar` always download layers and `--offline` is effective only for those goals, but I added it just for the unlikely case someone tries `jib:build` instead of `jib:dockerBuild` to do one-time online caching in this situation.)


chanseokoh:
Jeepers, I should have create a "draft" PR. Doesn't seem possible to change it to a draft now. I still need to add tests.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -120,9 +120,10 @@ private StepsRunner(
 
   public StepsRunner dockerLoadSteps(DockerClient dockerClient) {
     rootProgressDescription = "building image to Docker daemon";
+
     // build and cache
     stepsToRun.add(this::pullBaseImage);
-    stepsToRun.add(this::pullAndCacheBaseImageLayers);
+    stepsToRun.add(() -> pullAndCacheBaseImageLayers(true));
```
briandealwis: Worth adding `// always pull layers for docker builds`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -132,9 +133,10 @@ public StepsRunner dockerLoadSteps(DockerClient dockerClient) {
 
   public StepsRunner tarBuildSteps(Path outputPath) {
     rootProgressDescription = "building image to tar file";
+
     // build and cache
     stepsToRun.add(this::pullBaseImage);
-    stepsToRun.add(this::pullAndCacheBaseImageLayers);
+    stepsToRun.add(() -> pullAndCacheBaseImageLayers(true));
```
briandealwis: Worth adding `// always pull layers for build-tar`?

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -32,6 +32,8 @@
 
   private static final String SERIALIZE = "jibSerialize";
 
+  private static final String FORCE_LAYER_DOWNLOAD = "jib.forceDownload";
```
briandealwis: `jib.cacheBase` with `always` or `as-needed`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
```diff
@@ -36,14 +38,48 @@
 import javax.annotation.Nullable;
 
 /** Pulls and caches a single base image layer. */
-class PullAndCacheBaseImageLayerStep implements Callable<CachedLayerAndName> {
+class PullAndCacheBaseImageLayerStep implements Callable<PreparedLayer> {
```
briandealwis: Can we come up with a different name for this step, since it may not actually pull-and-cache the given image layer?  `ObtainBaseImageLayerStep`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
```diff
@@ -36,14 +38,48 @@
...
+
+    return makeList(
+        buildConfiguration, progressEventDispatcherFactory, baseImageAndAuth, blobChecker);
+  }
+
+  private static ImmutableList<PullAndCacheBaseImageLayerStep> makeList(
+      BuildConfiguration buildConfiguration,
+      ProgressEventDispatcher.Factory progressEventDispatcherFactory,
+      ImageAndAuthorization baseImageAndAuth,
+      BlobChecker blobChecker) {
```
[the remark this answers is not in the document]
chanseokoh: Hmm... for best optimization, I need to know three states:

1. Unknown: never checked the target registry and `PushBlobStep` should call `registryClient.checkBlob()` before pushing.
2. Doesn't exist: `PushBlobStep` should skip calling `registryClient.checkBlob()` and just push.
3. Exists in target: no need to push; `PushBlobStep` isn't created at all.

That is, if we have only two states (true and false), I can't decide if I can skip checking the existence later.

And another reason is that the code in other places is general in that this `PreparedLayer` can be either a base image layer or an application layer. It doesn't matter, as long as `PreparedLayer` returns correct state out of these three. For now, the "unknown" applies to application layers, and theoretically it is possible to skip building application layers in this fashion in the future. And this code is general and works for tar/Docker builds; currently we always say "unknown" for Docker to signal that we did not do existence check, but potentially in the future we may be able to determine if layers already exist in the Docker engine and skip downloading layers as well. Does that make sense?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushLayerStep.java
```diff
@@ -47,45 +45,53 @@
...
-    this.cachedLayerAndName = cachedLayerAndName;
+    this.preparedLayer = preparedLayer;
   }
 
   @Override
   public BlobDescriptor call()
       throws IOException, RegistryException, ExecutionException, InterruptedException {
-    CachedLayer layer = cachedLayerAndName.get().getCachedLayer();
+    PreparedLayer layer = preparedLayer.get();
+    if (layer.existsInTarget().orElse(false)) { // skip pushing if known to exist in registry
```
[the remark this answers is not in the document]
chanseokoh: > Moreover, I actually need the `Future` result of pushing a layer to get the digest (`BlobDescriptor`);

Sorry, this wasn't true. But it still is the case I find it difficult to take the logic out of here as I explained later.

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
chanseokoh: Will work on this tomorrow. Need to update error messages that mention this property too. I think I'll throw `IllegalArgumentException` for wrong values.

chanseokoh: Will go for `jib.cacheBaseImage` (instead of `jib.cacheBase`) with acceptable values of `always` and `as-needed`.

chanseokoh:
I have addressed all the comments, AFAIKT. Good for another review pass.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -212,18 +217,24 @@ private void pullBaseImage() {
...
-                    PullAndCacheBaseImageLayerStep.makeList(
-                        buildConfiguration,
-                        childProgressDispatcherFactory,
-                        results.baseImageAndAuth.get())));
+                    forcePull
+                        ? ObtainBaseImageLayerStep.makeListForForcedDownload(
+                            buildConfiguration,
+                            childProgressDispatcherFactory,
+                            results.baseImageAndAuth.get())
+                        : ObtainBaseImageLayerStep.makeListForSelectiveDownload(
```
briandealwis: I don't much like these method names: there's a lot of implicit knowledge required to connect `forcePull` to _selective downloading_.  Can we rename `forcePull` to `layersRequiredLocally`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -212,18 +217,24 @@ private void pullBaseImage() {
             new PullBaseImageStep(buildConfiguration, childProgressDispatcherFactory));
   }
 
-  private void pullAndCacheBaseImageLayers() {
+  private void pullAndCacheBaseImageLayers(boolean forcePull) {
```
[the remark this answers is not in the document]
chanseokoh: Fixed, thanks. No more `pullAndCache...` in the repo.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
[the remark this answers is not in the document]
briandealwis: If we're representing a tri-state, maybe an enum would be better rather than an `Optional<Boolean>`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
```diff
@@ -69,38 +106,50 @@
...
     try (ProgressEventDispatcher progressEventDispatcher =
             progressEventDispatcherFactory.create("checking base image layer " + layerDigest, 1);
         TimerEventDispatcher ignored =
             new TimerEventDispatcher(
                 buildConfiguration.getEventHandlers(), String.format(DESCRIPTION, layerDigest))) {
+
+      Optional<Boolean> layerExists = blobChecker.exists(layerDigest);
+      if (layerExists.orElse(false)) {
+        return new PreparedLayer.Builder(layer).setStateInTarget(layerExists).build();
+      }
```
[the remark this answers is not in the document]
briandealwis: Good point.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullAndCacheBaseImageLayerStep.java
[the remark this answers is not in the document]
chanseokoh: I agree it's better. If you think the tri-state is OK, how about my migrating to enum in a follow-up?

(Just telling you I tried enum initially and I later switched to `Optional<Boolean>` as it was bloating some code.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ObtainBaseImageLayerStep.java
```diff
@@ -69,38 +107,50 @@
...
-  public CachedLayerAndName call() throws IOException, CacheCorruptedException {
+  public PreparedLayer call() throws IOException, CacheCorruptedException, RegistryException {
+    DescriptorDigest layerDigest = layer.getBlobDescriptor().getDigest();
     try (ProgressEventDispatcher progressEventDispatcher =
             progressEventDispatcherFactory.create("checking base image layer " + layerDigest, 1);
         TimerEventDispatcher ignored =
             new TimerEventDispatcher(
                 buildConfiguration.getEventHandlers(), String.format(DESCRIPTION, layerDigest))) {
+
+      Optional<Boolean> layerExists = blobExistenceChecker.exists(layerDigest);
```
briandealwis: What if we inverted this checker to be _blobRequiredChecker_?  Then we're removing any knowledge here about existence — which is frankly a bit odd since it may not exist locally, but the code here is for downloading locally.  And then I think it can just be a `Predicate` as the force-download will always return `true`.  `PreparedLayer` would have a `existsLocally` instead of `existsInTarget`, and `PushBlob*` would skip pushing blobs if they didn't exist locally (which makes sense).

jib-core/CHANGELOG.md
```diff
@@ -3,8 +3,6 @@ All notable changes to this project will be documented in this file.
 
 ## [unreleased]
 
-### Added
-
```
briandealwis: Is this necessary to be removed?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ObtainBaseImageLayerStep.java
```diff
@@ -36,14 +38,49 @@
...
+      Authorization pushAuthorization) {
+    Verify.verify(!buildConfiguration.isOffline());
+
+    RegistryClient targetRegistryClient =
+        buildConfiguration
+            .newTargetImageRegistryClientFactory()
+            .setAuthorization(pushAuthorization)
+            .newRegistryClient();
+    BlobExistenceChecker blobExistenceChecker =
+        digest -> Optional.of(targetRegistryClient.checkBlob(digest).isPresent());
```
briandealwis: Hmm, this doesn't account for the cross-repo blob-mount case.  Maybe add a TODO for now.

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -91,6 +93,24 @@ public static boolean isUserAgentEnabled() {
     return Strings.isNullOrEmpty(System.getProperty(DISABLE_USER_AGENT));
   }
 
+  /**
+   * Gets whether to always cache base image layers. Determined from the {@code jib.cacheBaseImage}
+   * system property.
+   *
+   * @return true if the property is set to {@code always}; false otherwise
+   */
+  public static boolean alwaysCacheBaseImage() {
```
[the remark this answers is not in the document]
chanseokoh: Good idea. I'll file an issue for adding an API. It should be independent of this class, which is all about system properties.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ObtainBaseImageLayerStep.java
chanseokoh: *sigh* I swear I did post a long comment, but it got lost and nowhere to find. I will update later.

chanseokoh: > `PushBlob*` would skip pushing blobs if they didn't exist locally (which makes sense).

The catch is this: say we have

1. B_LOCAL_REMOTE: base layer that exists locally and remotely
2. A_LOCAL_REMOTE: app layer that exists locally and remotely
3. B_LOCAL: base layer that exists locally but not remotely
4. A_LOCAL: app layer that exists locally but not remotely
5. B_REMOTE: base layer that exists remotely but not locally

then the table should be

| layer | eventually pushed | `BlobPusher` should call |
| --- | --- | --- |
| B_LOCAL_REMOTE | no | nothing |
| A_LOCAL_REMOTE | no | `checkBlob` |
| B_LOCAL | yes | `pushBlob` |
| A_LOCAL | yes | `checkBlob` and `pushBlob` |
| B_REMOTE | no | nothing |

Note B_* (base layers) and A_* (app layers) work differently (i.e., calling `checkBlob` or not). `BlobPusher` should be able to differentiate them do proper actions.

chanseokoh: @briandealwis does this make sense?

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -91,6 +93,24 @@ public static boolean isUserAgentEnabled() {
...
+    String property = System.getProperty(CACHE_BASE_IMAGE, "as-needed");
+    if ("always".equals(property)) {
```
TadCordle: I'd rather we just make this a boolean property instead of a string needing two special keywords. I don't think this property in particular needs to be too future proof.

chanseokoh: Initially it was Boolean, but @briandealwis suggested these values. I thought this isn't a big deal to avoid being stuck with Boolean.

briandealwis: I don't think `true`/`false` capture the situation, unfortunately.  `jib.cacheBaseImage=false` may still result in caching the base image with `dockerBuild` or `buildTar`.

TadCordle: Then why not something like `jib.alwaysCacheBaseImage`?

briandealwis: Sure that would work.

chanseokoh: Okay, I tried to think other potential modes but couldn't. I'll go back to Boolean.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ObtainBaseImageLayerStep.java
briandealwis: Sorry, missed this.  Yes, that makes sense.  We need to test whether B is REMOTE to know if we need to download it. so as to be able to push it later.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ObtainBaseImageLayerStep.java
```diff
@@ -36,14 +38,50 @@
...
+      ImageAndAuthorization baseImageAndAuth,
+      Authorization pushAuthorization) {
+    Verify.verify(!buildConfiguration.isOffline());
+
+    RegistryClient targetRegistryClient =
+        buildConfiguration
+            .newTargetImageRegistryClientFactory()
+            .setAuthorization(pushAuthorization)
+            .newRegistryClient();
+    // TODO: also check if cross-repo blob mount is possible.
```
briandealwis: I wonder if this is as easy as adding `buildConfiguration.getBaseImageRegistry().equals(buildConfiguration.getTargetImageRegistry())` on in the existing-checker below?

chanseokoh: I thought it involves more than that, like checking
```
JibSystemProperties.useCrossRepositoryBlobMounts()
    && canAttemptBlobMount(authorization, sourceRepository))
```
to ensure that we will attempt the blob mount. And I think it is that we can *attempt* blob mount, and it might be possible that a server may not allow/support the mount by returning `202 Accepted` instead of `201 Created`, in which case `RegistryClient.pushBlob()` will fall back to do the usual pushing. So I think this is actually complicated.

But I think we are already pretty good without blob mount. Most of the time, the base image will be in the target repository. The blob mount can be a further optimization, but I think it does not give a huge value.

raizoor:
Hi team,

Have any time for release it? Sounds good to my CI/CD method.

Thanks !!!

PR #1849, opened 2019-07-16
Increase executor shutdown timeout in tests

### LINKED ISSUE #1839: Flaky test about progress footer display
From Kokoro Mac presubmit.

I've never seen this. I think this is a rare failure. Worth looking into it though.

```
com.google.cloud.tools.jib.plugins.common.logging.AnsiLoggerWithFooterTest > testLog_sameFooter FAILED
    java.lang.AssertionError: expected:<[footer, , message, footer, , another message, footer]> but was:<[footer, , message, footer]>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:118)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at com.google.cloud.tools.jib.plugins.common.logging.AnsiLoggerWithFooterTest.testLog_sameFooter(AnsiLoggerWithFooterTest.java:130)
```

chanseokoh:
Fixes #1839. Unlike #1846, `SingleThreadedExecutor` is not code only for testing, so I'm retaining the same 1-sec timeout in prod.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/PlainConsoleLogger.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/PlainConsoleLogger.java
@@ -44,8 +44,11 @@ class PlainConsoleLogger implements ConsoleLogger {
   @Override
   public void log(Level logLevel, String message) {
     if (messageConsumers.containsKey(logLevel)) {
-      singleThreadedExecutor.execute(() -> messageConsumers.get(logLevel).accept(message));
+      return;
     }
+    Consumer<String> messageConsumer = messageConsumers.get(logLevel);
+
+    singleThreadedExecutor.execute(() -> messageConsumer.accept(message));
   }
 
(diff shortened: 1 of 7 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/PlainConsoleLogger.java
```diff
@@ -43,7 +43,7 @@
 
   @Override
   public void log(Level logLevel, String message) {
-    if (!messageConsumers.containsKey(logLevel)) {
+    if (messageConsumers.containsKey(logLevel)) {
```
TadCordle: Is this intentional?

chanseokoh: Jeepers! Thanks.

PR #1869, opened 2019-07-24
Update faq for expired credentials in config.json

### LINKED ISSUE #1827: Update FAQ for credhelper in config.json being lower pri than "auths"
related to #1819 we should update: https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#what-should-i-do-when-the-registry-responds-with-unauthorized to indicate you could have auth issues from old `gcloud docker` injecting expiring `"auths"` for `gcr.io` addresses.

loosebazooka:
fixes #1827


docs/faq.md
```diff
@@ -471,6 +471,7 @@ If the registry returns `401 Unauthorized` or `"code":"UNAUTHORIZED"`, it is oft
...
       - Specific credentials (not recommend): [`<from/to><auth><username>/<password>`](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#using-specific-credentials) or in [`settings.xml`](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#using-maven-settings) for Maven / [`from/to.auth.username/password`](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#using-specific-credentials) for Gradle
       - These parameters can also be set through properties: [Maven](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#system-properties) / [Gradle](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#system-properties)
+* You previous used the deprecated `gcloud docker` to configure the `auths` block in `config.json`. Those credentials expire and need to be removed. Edit your `config.json` and remove `auths` associated with `gcr.io` (for example: `"https://asia.gcr.io"`). You can run `gcloud auth configure-docker` to correctly configure the `credHelpers` block for interactions with Google Container Registry.
```
chanseokoh: "You previously ..."

I'd like to put "short-lived" into "Those credentials expire ..." to emphasize it expires very soon. (An hour or two, isn't it?)

briandealwis:
This`auths` problem may happen for other registries and reasons too though. Can we instead say something like

> Your `auths` block in `config.json` may encode  time-sensitive credentials that have now expired.  For example `gcloud docker` used this approach.

loosebazooka:
okay one more try, PTAL

docs/faq.md
```diff
@@ -466,6 +466,7 @@ If the registry returns `401 Unauthorized` or `"code":"UNAUTHORIZED"`, it is oft
...
+   - `$HOME/.docker/config.json` may also contain short lived authorizations in the `auths` block that may have expired. In the case of Google Container Registry, if you had previously used `gcloud docker` to configure these authorizations, you should remove them by editing your `config.json` and deleting lines from `auths` associated with `gcr.io` (for example: `"https://asia.gcr.io"`). You can then run `gcloud auth configure-docker` to correctly configure the `credHelpers` block for more robust interactions with gcr.
```
briandealwis: A hyphen makes this read better, and "them" is a bit ambiguous:
```suggestion
   - `$HOME/.docker/config.json` may also contain short-lived authorizations in the `auths` block that may have expired. In the case of Google Container Registry, if you had previously used `gcloud docker` to configure these authorizations, you should remove these stale authorizations by editing your `config.json` and deleting lines from `auths` associated with `gcr.io` (for example: `"https://asia.gcr.io"`). You can then run `gcloud auth configure-docker` to correctly configure the `credHelpers` block for more robust interactions with gcr.
```

Issue #1468, opened 2019-02-08
Support for using a local Docker cache for base images


menya84:
**Description of the issue**:

**Expected behavior**:

**Steps to reproduce**: 

**Environment**: 

**`jib-maven-plugin` Configuration**: 
```xml
PASTE YOUR pom.xml CONFIGURATION HERE
```

**`jib-gradle-plugin` Configuration**: 
```groovy
PASTE YOUR build.gradle CONFIGURATION HERE
```

**Log output**: 

**Additional Information**:

briandealwis:
Hi @menya84.  Jib does support using images from private registries as a base image.  Could you provide some details?  There are some suggestions for [fixing connection problems with registries in the Jib FAQ](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-can-i-diagnose-problems-pulling-or-pushing-from-remote-registries).

menya84:
I didn’t mean private registry, I am talking about local image been just built and not pushed anywhere.
But even if I push it to some remote registry and then build new image with some changes locally then the remote one will be used as a base anyway.

coollog:
Hi @menya84 , do you mean a local image built with Docker and accessible via your local Docker daemon?

menya84:
Yes, exactly.

chanseokoh:
Being able to pull images from the cache of a local Docker daemon is an interesting idea. #718 is related, but the difference is that #718 is about not fetching remote metadata but using Jib's local cache. But https://github.com/GoogleContainerTools/jib/issues/718#issuecomment-440394884 actually mentions the idea of using the local Docker daemon. One thing to consider in both cases is the different behavior of Docker and Jib: Docker does not check remote images, while Jib always uses the most up-to-date images. I think it would make sense to not check remote images when using local images, whether Jib's cache or Docker's.

menya84:
I think we can use metadata of an image to make sure Jib always uses really latest image.
Firstly we fetch remote metadata and compare it with locally cached image metadata, and pull from remote only if remote image has newer timestamp otherwise do not pull and use locally cached imaged of the local Docker daemon.
This would be great solution because often it's much more chipper to quickly build base image locally and not push/pull every time to/from remote while we are playing in development mode.

chanseokoh:
[...]

PR #1873, opened 2019-07-30
Add save to DockerClient

TadCordle:
Towards #1468. Adds a `DockerClient` method to `docker save` a specified image to the specified path.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
@@ -159,6 +159,38 @@ public class DockerClient {
   }
 
   /**
+   * Saves an image tarball from the Docker daemon.
+   *
+   * @see <a
+   *     href="https://docs.docker.com/engine/reference/commandline/save/">https://docs.docker.com/engine/reference/commandline/save</a>
+   * @param imageReference the image to save
+   * @param outputPath the destination path to save the output tarball
+   * @return stdout from {@code docker}
+   * @throws InterruptedException if the 'docker save' process is interrupted
+   * @throws IOException if saving the blob from 'docker save' fails
+   */
+  public String save(ImageReference imageReference, Path outputPath)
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -158,6 +158,38 @@ public String load(ImageTarball imageTarball) throws InterruptedException, IOExc
...
   }
 
+  /**
+   * Saves an image tarball from the Docker daemon.
+   *
+   * @see <a
+   *     href="https://docs.docker.com/engine/reference/commandline/save/">https://docs.docker.com/engine/reference/commandline/save</a>
+   * @param imageReference the image to save
+   * @param outputPath the destination path to save the output tarball
+   * @return stdout from {@code docker}
```
chanseokoh: Do we need this? For a successful save, `docker save` doesn't output anything.

TadCordle: Yeah I guess not.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -158,6 +158,35 @@ public String load(ImageTarball imageTarball) throws InterruptedException, IOExc
...
+   * @param outputPath the destination path to save the output tarball
+   * @throws InterruptedException if the 'docker save' process is interrupted
+   * @throws IOException if creating the tarball fails
+   */
+  public void save(ImageReference imageReference, Path outputPath)
+      throws InterruptedException, IOException {
+    // Runs 'docker save'.
+    Process dockerProcess = docker("save", imageReference.toString(), "-o", outputPath.toString());
+
+    try (InputStreamReader stdout =
```
chanseokoh: Now we don't need this (and mocking), like the `tag` method below.

PR #1874, opened 2019-07-30
Stay on Ubuntu Trusty on Travis for oraclejdk8

chanseokoh:
We are hitting build failures as Travis cannot install Oracle JDK 8. It is not available on Xenial, so one way to resolve this is to stay on Trusty. Switching to openjdk8 is another option, but it will hit the Javadoc-related bug. There does exist a workaround (modifying our Javadoc to have fully qualified class names for nested classes) for the Javadoc bug, but I don't really like that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,10 @@ git:
 language: java
 matrix:
   include:
-    - name: OpenJDK 8
-      jdk: openjdk8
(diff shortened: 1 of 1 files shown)


chanseokoh:
"Oracle JDK 8 on Trusty" Travis build succeeded, so this should be good: https://travis-ci.org/GoogleContainerTools/jib/jobs/565566417

PR #1882, opened 2019-08-01
Upgrade google-http-client to eliminate deprecated usage

### LINKED ISSUE #1434: Update google-http-client and google-http-client-apache-v2 to 1.31

chanseokoh:
Fixes #1434.

Finally, google-http-client 1.31 is release, and we can move to the new world. jib-core will be once again free from any deprecated usage.

As noted in #1434, we no longer need to take care of `http(s).username|passworld` properties on our own. Note that I had previously committed a [test for this](https://github.com/GoogleContainerTools/jib/pull/1809/files#diff-7ca2549d686973d167e5c5a32b626108R108-R119) in #1809, so we can rest assured that this continues to work.


jib-core/src/test/java/com/google/cloud/tools/jib/http/ConnectionWithProxyCredentialsTest.java
```diff
@@ -1,167 +0,0 @@
...
-import org.apache.http.auth.AuthScope;
-import org.apache.http.auth.Credentials;
-import org.apache.http.impl.client.DefaultHttpClient;
-import org.junit.After;
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Test;
-
-/** Tests for {@link Connection} with setting proxy credentials. */
-public class ConnectionWithProxyCredentialsTest {
```
TadCordle: Do we test anywhere else to make sure things still work with a proxy? (Do we even need to?)

chanseokoh: Yeah, we are covered. The test below shows that `ApacheHttpTransport` does read the properties and send it through the `Authorization` header.

> Note that I had previously committed a [test for this](https://github.com/GoogleContainerTools/jib/pull/1809/files#diff-7ca2549d686973d167e5c5a32b626108R108-R119) in #1809, so we can rest assured that this continues to work.

TadCordle: Ah, I should read the description more, sorry.

PR #1883, opened 2019-08-01
Add tarball extractor

TadCordle:
Towards #1468. Adds a utility class for extracting tarballs for processing `docker save`/`jibBuildTar` output.


jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+public class TarExtractor {
+
+  /**
+   * Extracts a tarball to the specified destination.
+   *
+   * @param source the tarball to extract
+   * @param destination the output directory
+   * @throws IOException if extraction fails
+   */
+  public static void extract(Path source, Path destination) throws IOException {
```
chanseokoh: Now I think what I missed. Ideally, we will have better performance by accepting an input stream, because `docker save` can output to stdout. I see the plan was to docker-save to a tarball and read it again to extract, but we can avoid reading twice and having to do disk write, saving space and time. (Of course, the assumption is that we don't need the tarball.) I think it's worth avoiding extra read-write on disk.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+    try (InputStream in = Files.newInputStream(source)) {
+      try (TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(in)) {
```
chanseokoh: You can use a single try-resource.

```java
try (InputStream in = ...;
   TarArchiveInputStream ...) {
```

And in this case, I think using `BufferedInputStream` is meaningful. `BufferedInputStream` may loop to do multiple `read()` before returning, so it could reduce the number of `out.write()` if this happens.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
TadCordle: We don't really "need" the tarball, but it lets us reuse the code path for using a tarball base image. I'm imagining adding 2 new build steps (a docker save build step that returns a path to a tarball, and an extraction build step that converts the tar into the `Image` + `ContainerConfigurationTemplate` + layers). A docker daemon base image would use both new build steps, and a tarball base image would only need to use the second one.

You are right about performance, although it might be tricky navigating the tar stream in a non-linear fashion, wouldn't it? We need to parse the manifest first to get the names of the layers + configuration, which means we'd need to skip over entries if the manifest isn't the first entry in the tar. Unless we could determine which files are which based on extension, or if we can traverse the tar contents out of order? I'm not too familiar with the tar archive api.

chanseokoh: Okay, let's just move forward with this. I will have a chance to revisit this once your design is implemented. I don't think I can suggest something concrete until I have a firm understanding of the implementation.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
TadCordle: Should we use a `BufferedOutputStream` as well?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+   * @throws IOException if extraction fails
+   */
+  public static void extract(Path source, Path destination) throws IOException {
+    try (InputStream in = new BufferedInputStream(Files.newInputStream(source));
+        TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(in)) {
+      byte[] buffer = new byte[1024];
+      TarArchiveEntry entry = tarArchiveInputStream.getNextTarEntry();
+      while (entry != null) {
+        if (entry.isDirectory()) {
+          Files.createDirectories(destination.resolve(entry.getName()));
```
chanseokoh: Let's [guard the Zip-Slip](https://github.com/GoogleContainerTools/jib/blob/8d82956f38c43d6c23f37fcb39ea1ed895372283/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java#L47-L53) vulnerability.

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.stream.Stream;
+import org.junit.Assert;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+/** Tests for {@link TarExtractor}. */
+public class TarExtractorTest {
+
+  @ClassRule public static TemporaryFolder temporaryFolder = new TemporaryFolder();
```
chanseokoh: `final`

`@Rule` should be enough as long as `TemporaryFolder` is cheap. Probably it's cheap enough. If we use `@ClassRule`, we should be careful because files will be persisted across tests.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
chanseokoh: I think it doesn't hurt, although we may be not so bad without it because we are not calling `write()` frequently with small bits.

PR #1886, opened 2019-08-01
Restore proxy system properties after test

chanseokoh:
This is worth restoring (or clearing).

### DIFF
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
@@ -25,56 +25,51 @@ import java.util.Arrays;
 import java.util.Collections;
 import org.gradle.api.Project;
 import org.gradle.testfixtures.ProjectBuilder;
-import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.contrib.java.lang.system.RestoreSystemProperties;
 
 /** Tests for {@link JibExtension}. */
 public class JibExtensionTest {
 
+  @Rule public final RestoreSystemProperties systemPropertyRestorer = new RestoreSystemProperties();
+
   private JibExtension testJibExtension;
   private Project fakeProject;
 
-  private static void clearProperties() {
-    System.clearProperty("jib.from.image");
-    System.clearProperty("jib.from.credHelper");
-    System.clearProperty("jib.to.image");
-    System.clearProperty("jib.to.tags");
-    System.clearProperty("jib.to.credHelper");
-    System.clearProperty("jib.container.appRoot");
-    System.clearProperty("jib.container.args");
-    System.clearProperty("jib.container.entrypoint");
-    System.clearProperty("jib.container.environment");
-    System.clearProperty("jib.container.extraClasspath");
-    System.clearProperty("jib.container.format");
-    System.clearProperty("jib.container.jvmFlags");
-    System.clearProperty("jib.container.labels");
-    System.clearProperty("jib.container.mainClass");
-    System.clearProperty("jib.container.ports");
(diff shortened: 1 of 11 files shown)


chanseokoh:
Now pulling in `com.github.stefanbirkner:system-rules:1.19.0` to use `RestoreSystemProperties`. The rule saved all current properties before tests and restore them after done.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
```diff
@@ -25,58 +25,53 @@
...
-    System.clearProperty("jib.extraDirectory.path");
-    System.clearProperty("jib.extraDirectory.permissions");
-    System.clearProperty("jib.extraDirectories.paths");
-    System.clearProperty("jib.extraDirectories.permissions");
-  }
-
   @Before
   public void setUp() {
-    clearProperties();
+    Assert.assertNull(System.getProperty("jib.from.image"));
```
loosebazooka: shouldn't we still be clearing them as part of test setup?

loosebazooka: presumably changing to an assert will just fail on systems with already defined system properties (which shouldn't affect our tests given our restoreSystemProperties usage)

chanseokoh: Yeah, if these assertions fail, we will know there exist some places where we fail to clean up. I am opting for fixing them if that happens, rather than just making the tests here work with a workaround.

loosebazooka: Oh right they're system properties. I was thinking env.

loosebazooka: Although I don't think it's an individual test's responsibility to cleanup other tests. It should ensure its own settings and restore whatever it changed.

chanseokoh: Yeah, that is what I am saying here. The previous code was rather clearing these properties for the tests in this class to cleanup leftover from other tests. And I am now making individual tests restore whatever they changed. These Jib-specific properties must always be null whenever any test completes. This invariant should not be violated.

loosebazooka: Hrmm.. maybe we're saying the same thing differently, but I think the assertions should be actions rather than checks in the `@Before` step.

Clear these system properties so this test guarantee to pass irrespective of the state of the system when the tests start?

TadCordle: Does the `RestoreSystemProperties` clear the values automatically? If not I think I agree with what @loosebazooka is saying, I don't think the tests should fail if a user does something like `./gradlew build -Djib.from.image=...`

loosebazooka: no it just saves and reloads whatever state existed before.

loosebazooka: I think if we wanted to have assertions we could have a final safety test that ran last?

chanseokoh: Okay, I will revert back to clearing them.

@TadCordle BTW, my intention was actually to catch such potential cases and fix them.

chanseokoh: There are other places where we've been doing the null check instead of clearing, so I'll fix them too.

Issue #1763, opened 2019-06-06
AccessDeniedException happen when building the image


HuangAo:
**Description of the issue**:
when i build a springboot project on local win10 enviroment, encounter AcceessDeniedException, and i checked there's no folder named "805c6914704cef9e79074e36853298287ad1262f368d8648d52ff307dbebcf28" under target\jib-cache\layers mentioned in the ERROR log
```
[DEBUG] TIMING  Building and pushing image
[INFO]  TIMING  Retrieving registry credentials for registry-1.docker.io
[INFO]  TIMING  Pulling base image manifest
[INFO]  TIMING  Building application layers
[INFO]  TIMED   Building application layers : 1.0 ms
[INFO]  TIMING  Building classes layer
[INFO]  TIMING  Building dependencies layer
[INFO]  TIMING  Building extra files layer
[INFO]  TIMING  Building snapshot dependencies layer
[INFO]  TIMING  Building resources layer
[INFO]  TIMING  Setting up to push layers
[INFO]  TIMED   Setting up to push layers : 2.0 ms
[INFO]  TIMED   Building snapshot dependencies layer : 647.0 ms
[INFO]  TIMED   Building extra files layer : 656.0 ms
[INFO]  Building resources layer built sha256:24785422b8265b79eb17fce1274dda5622fd8f97e5baf233aef635367a122547
[INFO]  TIMED   Building resources layer : 1744.0 ms
[INFO]  Building classes layer built sha256:2c4829a880a4958ac7272e9101e967e764c7fd76aaefb9cae2ce4436481f94f4
[INFO]  TIMED   Building classes layer : 1853.0 ms
[INFO]  Cannot verify server at https://127.0.0.1:5000/v2/openjdk11/manifests/alpine-jre. Attempting again with no TLS verification.
[INFO]  Failed to connect to https://127.0.0.1:5000/v2/openjdk11/manifests/alpine-jre over HTTPS. Attempting again with HTTP: http://127.0.0.1:5000/v2/openjdk11/manifests/alpine-jre
[INFO]  Cannot verify server at https://127.0.0.1:5000/v2/openjdk11/blobs/sha256:37888a81f3758f5726758ffa035b649656c56899effdca7fe941209041620b41. Attempting again with no TLS verification.
[...]

Issue #1635, opened 2019-04-16
Randomly failing maven build with WindowsException


arifC:
**Description of the issue**:
From time to time I get the following exception failing the build. It seems Windows related with some cache access error. The root exception I get is **AccessDeniedException** based on some **WindowsExceptions**. happens quite often when the build runs through very fast. When I'm on battery with my laptop (slower CPU freq) the error doesn't occur.

`Caused by: java.nio.file.AccessDeniedException: C:\Users\I506803\examples\personservice\target\jib-cache\tmp\4235202945488612732 -> C:\Users\I506803\examples\personservice\target\jib-cache\layers\b4617547f616e55fb347a9d4cf7db25dc27bf11608a4e6ef749c13df3334538b
    at sun.nio.fs.WindowsException.translateToIOException (WindowsException.java:83)
    at sun.nio.fs.WindowsException.rethrowAsIOException (WindowsException.java:97)
    at sun.nio.fs.WindowsFileCopy.move (WindowsFileCopy.java:387)
    at sun.nio.fs.WindowsFileSystemProvider.move (WindowsFileSystemProvider.java:287)
    at java.nio.file.Files.move (Files.java:1395)
    at com.google.cloud.tools.jib.cache.DefaultCacheStorageWriter.moveIfDoesNotExist (DefaultCacheStorageWriter.java:74)
    at com.google.cloud.tools.jib.cache.DefaultCacheStorageWriter.write (DefaultCacheStorageWriter.java:184)
    at com.google.cloud.tools.jib.cache.DefaultCacheStorage.write (DefaultCacheStorage.java:69)
    at com.google.cloud.tools.jib.cache.Cache.writeUncompressedLayer (Cache.java:82)
    at com.google.cloud.tools.jib.builder.steps.BuildAndCacheApplicationLayerStep.call (BuildAndCacheApplicationLayerStep.java:138)
    at com.google.cloud.tools.jib.builder.steps.BuildAndCacheApplicationLayerStep.call (BuildAndCacheApplicationLayerStep.java:40)
    at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly (TrustedListenableFutureTask.java:125)
    at com.google.common.util.concurrent.InterruptibleTask.run (InterruptibleTask.java:57)
    at com.google.common.util.concurrent.TrustedListenableFutureTask.run (TrustedListenableFutureTask.java:78)
[...]

Issue #1438, opened 2019-01-21
DirectoryNotEmptyException on Windows


madegroot:
**Description of the issue**:
Unable to use the jib Gradle plug-in on Windows due to DirectoryNotEmptyException.

**Expected behavior**:
Usage of jib and jibDockerBuild on Windows

**Steps to reproduce**: 
```
D:\repos\personal-repos\gradle-init-project>gradle init --type java-library --project-name jvm-library
Select build script DSL:
  1: groovy
  2: kotlin
Enter selection (default: groovy) [1..2] 1
Select test framework:
  1: junit
  2: testng
  3: spock
Enter selection (default: junit) [1..3] 1
Source package (default: jvm.library):
```
```D:\repos\personal-repos\gradle-init-project>notepad build.gradle```
- add to the plug-in section: `id 'com.google.cloud.tools.jib' version '1.0.0'` 
- add below the plug-in section `jib.to.image = 'madegroot/helloworld'`

`D:\repos\personal-repos\gradle-init-project>notepad src\main\java\jvm\library\Library.java`
- add to class `public static void main(String[] args) { System.out.println("Hello World"); }`
```
D:\repos\personal-repos\gradle-init-project>gradlew build
D:\repos\personal-repos\gradle-init-project>gradlew jib -Djib.to.auth.username=madegroot -Djib.to.auth.password=QQQ --stacktrace

Containerizing application to madegroot/helloworld...
Base image 'gcr.io/distroless/java' does not use a specific image digest - build may not be reproducible
Executing tasks:
[====================          ] 66,7% complete
> setting up to push layers

> Task :jib FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':jib'.
> com.google.cloud.tools.jib.plugins.common.BuildStepsExecutionException: Build image failed

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':jib'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:95)
        at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:91)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:57)
[...]

PR #1887, opened 2019-08-02
Retry cache installation attempts on Windows

### LINKED ISSUE #1855: Retry atomic move on windows when access denied exception
Related to #1763 

Windows anti virus software might have file locks on newly created files which results in this exception.

There are two parts:
- [ ] Backoff and retry if we can
- [ ] Provide better error messaging ("perhaps your anti virus software is intefering")

briandealwis:
Introduce a simple `Retry` class that retries an operation until it succeeds to a fixed number of attempts.  It was easier to pull this out as a library class for testing purposes.  The `Retry` class is used in `CacheStorageWriter` to retry the move of a downloaded blob into place in the cache.  This operation may fail on Windows because (we suspect) of anti-virus software and/or Windows Defender.

Fixes #1855 and #1438, #1635, #1763


chanseokoh:
One high-level feedback: I think we should print out a warning prominently whenever Jib does a retry, and especially for Windows, we include a message about possible anti-virus interference. As I said a while ago, this anti-virus may happen pretty often, if not always to some users, and they may just get the false impression that Jib is slower than other tools. For this kind of repeated failures, I think we should try to make the user correct the situation or at least let them know the potential performance hit.

That said, now I think it would be ideal if Jib would not have to be worried about potential inference of anti-virus, which may require re-designing the cache. For example, I think Maven will most likely be free of this concern, as it must have been battle-tested and it would be sad if people have to turn off anti-virus whenever they have to use Maven.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
```diff
@@ -81,15 +82,27 @@ private static void moveIfDoesNotExist(Path source, Path destination) throws IOE
       return;
     }
 
+    // Some Windows users report java.nio.file.AccessDeniedException that we suspect is caused
+    // by anti-virus programs, like Windows Defender, that open new files for scanning.
+    // Retry the rename up to 5 times.
```
chanseokoh: ... 5 times at the interval of 15 ms.

briandealwis:
Agreed — I'm just not sure how to log this. I guess using j.u.logging should be enough in this case: it can be redirected if necessary.

Maven downloads directly into the repository location, and people do encounter repository corruption (we did with CT4E, if you remember, because we weren't obtaining the right lock). search for `.lastUpdated`.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -0,0 +1,159 @@
...
+        action.run();
+
+        // no exception: check the stop condition, if provided
+        if (successCondition == null || successCondition.getAsBoolean()) {
+          return true;
+        }
+      } catch (Exception caughtException) {
+        @SuppressWarnings("unchecked")
+        E ex = (E) caughtException;
+        // if this is the last iteraion, no more retries
```
chanseokoh: typo: iteration

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -0,0 +1,159 @@
...
+   * retryable}.
+   *
+   * @return true if the action was run successfully, or {@code false} if the action was unable to
+   *     complete
+   * @throws E exception thrown during the action
+   */
+  public boolean run() throws E {
+    for (int i = 0; i < maximumRetries; i++) {
+      // sleep between attempts, but not on the first attempt
+      if (i > 0 && sleepMilliseconds >= 0) {
```
chanseokoh: `sleepMilliseconds > 0` instead of `>=`?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -0,0 +1,159 @@
...
+
+      // Do we need to continue?
+      if (successCondition != null && successCondition.getAsBoolean()) {
+        return true;
+      }
+      try {
+        action.run();
+
+        // no exception: check the stop condition, if provided
+        if (successCondition == null || successCondition.getAsBoolean()) {
```
chanseokoh: I think the configurable `successCondition` is a bit confusing. For example, this check assumes that if a `successCondition` is not explicitly specified, then the operation is deemed successful. But the check in 137, which will be evaluation right after this, assumes that until the given `successCondition` is satisfied, the operation is not deemed successful.

To put this in another way, say I call

- `until(false)` --> executes the operation 5 times and returns false. This feels right.
- `until(true)` --> never executes the operation. Doesn't feel intuitive due to the prevalent familiarity to the structure `do { } until <condition>`.
- not calling `until()` --> executes the operation once. Perhaps `until(true)` should work in this fashion, i.e., the default stop condition when `until()` is not called should be equal to `until(true)`.

Or should the `Action.run()` interface return Boolean, and have the client of `Retry` take the responsibility to evaluate a terminal condition as part of the action?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
```diff
@@ -81,15 +82,27 @@ private static void moveIfDoesNotExist(Path source, Path destination) throws IOE
...
+              .until(() -> Files.exists(destination))
+              .maximumRetries(5)
+              .retryOnException(ex -> ex instanceof FileSystemException)
+              .sleep(15, TimeUnit.MILLISECONDS)
+              .run();
+      if (success) {
+        return;
       }
+    } catch (IOException ex) {
+      exception = ex;
```
chanseokoh: I think you can throw the exception directly here. There's no looping. Calling `run()` will either throw an exception or not.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/RetryTest.java
```diff
@@ -0,0 +1,127 @@
...
+import java.util.concurrent.TimeUnit;
+import org.junit.Assert;
+import org.junit.Test;
+
+/** Tests for {@link Retry}. */
+public class RetryTest {
+  private volatile int actionCount = 0;
+  private final Retry.Action<Exception> countingAction =
+      () -> {
+        synchronized (this) {
```
chanseokoh: I don't think we really need synchronization. There is no additional thread and every code is executed in the main thread. Even if a thread was involved, we can use `LongAdder` or `AtomicInteger.incrementAndGet()` (both under `java.util.concurrent.atomic`?

jib-core/src/test/java/com/google/cloud/tools/jib/cache/RetryTest.java
```diff
@@ -0,0 +1,127 @@
...
+      Assert.assertEquals("whee", ex.getMessage());
+      Assert.assertEquals(5, actionCount);
+    }
+  }
+
+  @Test
+  public void testNonRetryableException() {
+    // the exception is not ok and so should only try 1 time
+    try {
+      Retry.<Exception>action(countingExceptionAction)
```
chanseokoh: I think `<Exception>` can be omitted.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/RetryTest.java
```diff
@@ -0,0 +1,127 @@
...
+    boolean result = Retry.action(countingAction).maximumRetries(2).until(() -> false).run();
+    Assert.assertFalse(result);
+    Assert.assertEquals(2, actionCount);
+  }
+
+  @Test
+  public void testRetryableException() {
+    // all exceptions are retryable by default, so retry 5 times
+    try {
+      Retry.<Exception>action(countingExceptionAction).until(() -> false).run();
```
chanseokoh: ditto

chanseokoh:
> Maven downloads directly into the repository location, and people do encounter repository corruption (we did with CT4E, if you remember, because we weren't obtaining the right lock). search for .lastUpdated.

This is not really related to this issue, but I remember the issue. I was conjecturing that it is perhaps due to explicitly executing Maven (for example, our CT4E code runs `mvn package` programmatically on a sample test project actually open in a test Eclipse instance that runs as Junis tests) inside Eclipse and this somehow interferes with m2e inside Eclipse during tests. But I may be wrong, of course.

But my point is not really about corruption. Corruption sounds like an independent issue that should not happen and be fixed if it can ever happen. Rather, I am saying that essential and popular tools like Maven, Gradle, etc., does not require that you disable anti-virus to use it (at all or efficiently). So, in the same general sense, if it is technically feasible to make Jib free from being affected by anti-virus, I think that would be ideal from the user perspective.

chanseokoh:
And not that we should make Jib ideal right now. We will go for this retry approach now. I just wish in the future we could eventually make anti-virus a non-issue.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/RetryTest.java
briandealwis: You do... If it's volatile. But it's a good point that since it's on the same thread then volatile isn't required.

briandealwis:
I see what your saying. Excellent point. I wonder how many people run maven in parallel against the same local repository. Maybe we should be placing the Jib cache inside the `maven.local.repo`...

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
briandealwis: Oops, I meant for 0 as it (almost) acts like a yield. Will fix `sleep()` instead.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
briandealwis: Maybe `until()` isn't the right name.

The idea is that `until()` provides a test as to whether we're in the desired end-state.  If we're already in that state, then running the action is unnecessary.

We were doing this in `moveIfDoesNotExist()` by [an initial `Files.exists(destination)` check](https://github.com/GoogleContainerTools/jib/blob/85c930fccd4ea95f060a6c308d3c52a56dda0341/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java#L81).

Although I could merge this into the action, it would end up looking like:
```
Retry.action(() -> {
    if(Files.exists(destination)) {
      return true;
    }
    Files.move(source, destination);
    return Files.exists(destination);
  })

File: jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java (hunk 5 above)
Comment (briandealwis) [reply jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java]: +1

File: jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java (hunk 4 above)
Comment (briandealwis) [reply jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java]: I can't think of a better wording, and that's not so bad.

File: jib-core/src/test/java/com/google/cloud/tools/jib/cache/RetryTest.java [hunk 9]
```diff
@@ -0,0 +1,119 @@
...
+  @Test
+  public void testInterruptSleep() throws Exception {
+    // interrupt the current thread so as to cause the retry's sleep() to throw
+    // an InterruptedException
+    Thread.currentThread().interrupt();
+    boolean result = Retry.action(unsuccessfulAction).sleep(10, TimeUnit.SECONDS).run();
+    Assert.assertFalse(result);
+    Assert.assertEquals(1, actionCount);
+    // This thread should be marked as interrupted (plus clear the flag for the test)
+    Assert.assertTrue(Thread.interrupted());
```
chanseokoh: I feel clearing the interrupted state is important enough to warrant the structure

```java
try {
  ...
} finally {
  Assert.assertTrue(Thread.interruped());
}
```

jib-core/src/test/java/com/google/cloud/tools/jib/cache/RetryTest.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  @Test
+  public void testInvalid_maximumRetries() {
+    try {
+      Retry.action(successfulAction).maximumRetries(0);
+      Assert.fail();
+    } catch (IllegalArgumentException ex) {
+      /* expected */
```
chanseokoh: nit: I think it's meaningful to add a helpful message like the value should be non-negative. It also makes it clear that the exception should come from `.maximumRetries()` and not `.action()`.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/RetryTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  @Test
+  public void testInvalid_sleep() {
+    try {
+      Retry.action(successfulAction).sleep(-1, TimeUnit.DAYS);
+      Assert.fail();
+    } catch (IllegalArgumentException ex) {
+      /* expected */
```
chanseokoh: nit: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
```diff
@@ -75,21 +77,36 @@ private WrittenLayer(
...
-        // TODO to log that the destination exists
-        throw ex;
-      }
-    }
+    String message =
+        String.format(
+            "unable to move: %s to %s; this is often caused by interference from antivirus",
+            source, destination);
+    // TODO figure out a better logging scheme
+    Logger.getLogger(CacheStorageWriter.class.getName()).warning(message);
```
chanseokoh: Ugh... I didn't know this class does not accept `EventHandlers`. Does this make Maven/Gradle print out a warning? Otherwise, I think this is not so meaningful. And I feel a bit awkward because we don't use `Logger.getLogger` anywhere else in our codebase.

In any case, I was hoping that we could print out the warning whenever there is a retry (not that when all the retries fail). Maybe we can discuss this in more detail later.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+package com.google.cloud.tools.jib.cache;
+
+import com.google.common.base.Preconditions;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+
+/**
+ * A simple class for retrying an action until it succeeds, or has retried too often and failed. By
```
TadCordle: nit: Remove "A simple class for" and just start with "Retries an action..."

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -0,0 +1,142 @@
...
+ *   <li>The retry instance can be configured to {@link #sleep(long, TimeUnit) sleep between
+ *       retries}.
+ *   <li>The maximum retry count {@link #maximumRetries(int) is configurable} (5 times by default).
+ * </ul>
+ *
+ * @param <E> the class of exceptions that may be thrown
+ */
+public class Retry<E extends Exception> {
+
+  /** An Action may throw an exception of type {@code E}. */
```
TadCordle: `{@link Action}`?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -0,0 +1,142 @@
...
+
+      try {
+        // Do we need to continue?
+        if (action.run()) {
+          return true;
+        }
+
+      } catch (Exception caughtException) {
+        @SuppressWarnings("unchecked")
+        E ex = (E) caughtException;
```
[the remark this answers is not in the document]
briandealwis: Good point on the exception.  Compiler seems happy to rethrow the caught exception.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
briandealwis: I'll remove this.  The `IOException` should be propagated upwards with the warning.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
briandealwis: This is the `Action` interface 😄  I'll change `Action` to a phrase.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
```diff
@@ -75,21 +77,36 @@ private WrittenLayer(
    * @throws IOException if an I/O exception occurs
    */
   private static void moveIfDoesNotExist(Path source, Path destination) throws IOException {
-    // If the file already exists, we skip renaming and use the existing file. This happens if a
-    // new layer happens to have the same content as a previously-cached layer.
-    if (Files.exists(destination)) {
+    // Some Windows users report java.nio.file.AccessDeniedException that we suspect is caused
+    // by anti-virus programs, like Windows Defender, that open new files for scanning.
+    // Retry the rename up to 5 times, with 15ms pause between each retry.
+    boolean success =
```
loosebazooka: Should this run exclusively on windows machines?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
briandealwis: If I force the move action to fail, Jib reports the right thing:

> [ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.4.1-SNAPSHOT:buildTar (default-cli) on project first: unable to move: /Users/bsd/Manumitting/Projects/GooglePluginEclipse/jibplayground/maven/app/target/jib-cache/tmp/4013610408160898804/.tmp.layer to /Users/bsd/Manumitting/Projects/GooglePluginEclipse/jibplayground/maven/app/target/jib-cache/tmp/4013610408160898804/932da51564135c98a49a34a193d6cd363d8fa4184d957fde16c9d8527b3f3b02; this is often caused by interference from antivirus

briandealwis: Tweaked the message

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -0,0 +1,142 @@
...
+   * @return the instance
+   * @param <E> the class of exceptions that may be thrown
+   */
+  public static <E extends Exception> Retry<E> action(Action<E> action) {
+    return new Retry<E>(action);
+  }
+
+  private final Action<E> action;
+  private int maximumRetries = 5;
+  private Predicate<? super E> retryOnException = ignored -> true; // continue to retry
```
loosebazooka: nit:  I would expect something named `retryOnException` to be true/false. But I have no naming alternative :(

also I'm confused how this is `Predicate<? super E>`, how can the exception thrown be super of E?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
loosebazooka: does this still need to be removed?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
briandealwis: I already changed it to just be `Predicate<Exception>`.  It's not that the exception will be a super of `E`, but that you can reuse a predicate that takes a `Throwable` (for example).

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java
```diff
@@ -110,29 +110,26 @@ private Retry(Action<E> action) {
...
         if (action.run()) {
           return true;
         }
 
+      } catch (InterruptedException ex) {
+        // Restore the interrupted status
+        Thread.currentThread().interrupt();
+        return false;
+
       } catch (Exception caughtException) {
```
chanseokoh: nit: `catch (Exception ex)` should work (short variable name).

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
```diff
@@ -75,21 +76,34 @@ private WrittenLayer(
...
+                    return true;
+                  }
+                  Files.move(source, destination);
+                  return Files.exists(destination);
+                })
+            .maximumRetries(5)
+            .retryOnException(ex -> ex instanceof FileSystemException)
+            .sleep(15, TimeUnit.MILLISECONDS)
+            .run();
+    if (success) {
```
chanseokoh: nit: I think I prefer

```java
  if (!success) {
    ...
    throw new IOException(...);
  }
}
```

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
briandealwis: It shouldn't matter.  There is a possibility this applies to other platforms.  I don't have a strong opinion on this.

loosebazooka: I can live with that.

loosebazooka:
Looks good. Any idea how we can simulate this? Use a debugger or something to pause and go lock the file?

briandealwis:
I tried reproducing these build errors in my Windows VM (which had no anti-virus, but did have Windows Defender, I believe), but was unsuccessful.

briandealwis:
I'll merge this and ping the associated issues to ask if the affected users can try building from HEAD.

briandealwis:
NB: #1438 involves errors removing the temp directory and isn't fixed by this PR.

PR #1888, opened 2019-08-02
Add container creation timestamp configuration

### LINKED ISSUE #1609: Allow configuring container creation timestamp in plugins
instead of just whether or not to use current timestamp. We could change the `useCurrentTimestamp` parameter to take either a boolean, or an integer number of milliseconds since epoch, or rename the parameter to just `timestamp` and take a string to determine whether to use current timestamp or a specific timestamp.

TadCordle:
Fixes #1609. Adds `container.creationTime` config to maven and gradle. Accepted values are:
- `EPOCH` - epoch + 1 second, same as files modification option
- `USE_CURRENT_TIMESTAMP` - uses the current timestamp
- an ISO 8601 date time

Also deprecates `container.useCurrentTimestamp` in favor of the `USE_CURRENT_TIMESTAMP` value.


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -89,6 +89,12 @@ static void checkDeprecatedUsage(JibExtension jibExtension, Logger logger) {
             "You cannot configure both 'jib.extraDirectory.path' and 'jib.extraDirectories.paths'");
       }
     }
+
+    if (jibExtension.getContainer().getUseCurrentTimestamp()) {
+      logger.warn(
+          "'jib.container.useCurrentTimestamp' is deprecated; use 'jib.container.creationTime' to "
```
chanseokoh: People will use `useCurrentTimestamp` to set the current time. The message should also include a way to achieve this, like using some constant like `USE_CURRENT_TIMESTAMP`.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/TaskCommonTest.java
```diff
@@ -131,6 +133,16 @@ public void testCheckDeprecatedUsage_extraDirectoryAndExtraDirectoriesConfigured
     }
   }
 
+  @Test
+  public void testCheckDeprecatedUsage_useCurrentTimestampConfigured() {
+    Mockito.when(containerParameters.getUseCurrentTimestamp()).thenReturn(true);
+    TaskCommon.checkDeprecatedUsage(jibExtension, logger);
+    Mockito.verify(logger, Mockito.times(1))
```
chanseokoh: The default behavior is `Mockito.times(1)`, so no need to specify that. The reason it is present in the extra directory test is that there are multiple depreated configurations and I wanted to ensure the user sees the message only once.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -366,8 +368,18 @@ AuthConfiguration getTargetImageAuth() {
   boolean getUseCurrentTimestamp() {
```
chanseokoh: Let's add `@Deprecated` for tracking purposes. I tried to add `@Deprecated` as much as possible to the code related to `extraDirectory` that should be removed later, except for the code where you can't really miss removing them like tests that call a method that will certainly be removed or where the notion `@Deprecated` doesn't apply well.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/JibPluginConfigurationTest.java
```diff
@@ -228,7 +233,7 @@ public void testEmptyOrNullTags() {
   }
 
   @Test
-  public void testDeprecatedSystemProperties() {
+  public void testDeprecatedSystemProperties_extraDirectories() {
```
chanseokoh: _extraDirectories --> _extraDirectory (the deprecated one)

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/JibPluginConfigurationTest.java
```diff
@@ -247,7 +252,17 @@ public void testDeprecatedSystemProperties() {
...
+    sessionProperties.put("jib.container.useCurrentTimestamp", "true");
+    Assert.assertTrue(testPluginConfiguration.getUseCurrentTimestamp());
+    Mockito.verify(log, Mockito.times(1))
+        .warn(
+            "jib.container.useCurrentTimestamp is deprecated; use jib.container.creationTime to "
+                + "specify an ISO 8601 timestamp instead");
+  }
+
+  @Test
+  public void testDeprecatedProperties_extraDirectories() {
```
chanseokoh: ditto

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidFilesModificationTimeException.java
```diff
@@ -18,6 +18,7 @@
 
 import java.time.format.DateTimeParseException;
 
+/** Exception when an invalid file timestamp configuration is encountered. */
```
chanseokoh: :+1:

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -212,6 +212,8 @@ static PluginConfigurationProcessor processCommonConfiguration(
           LogEvent.warn(
               "Setting image creation time to current time; your image may not be reproducible."));
       jibContainerBuilder.setCreationTime(Instant.now());
+    } else {
+      jibContainerBuilder.setCreationTime(getConfiguredTime(rawConfiguration.getCreationTime()));
```
chanseokoh: For extraDirectories, I think I throw an error if both extraDirectory and extraDirectories are given to eliminate any possible pitfall while the user is migrating. I think it's worth the effort to prevent such an unfortunate peril that may otherwise not be easily detected.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -458,22 +454,37 @@ static ContainerizingMode getContainerizingModeChecked(
...
-        case "EPOCH_PLUS_SECOND":
-          Instant epochPlusSecond = Instant.ofEpochSecond(1);
-          return (ignored1, ignored2) -> epochPlusSecond;
-
-        default:
-          Instant timestamp =
-              DateTimeFormatter.ISO_DATE_TIME.parse(modificationTime, Instant::from);
-          return (ignored1, ignored2) -> timestamp;
-      }
+      Instant instant = getConfiguredTime(modificationTime);
```
chanseokoh: This should be the right, but just letting you know that originally we had `KEEP_ORIGINAL` for files which would return different timestamps determined in a later stage when files are actually added to a layer as `LayerEntry`. If we revive `KEEP_ORIGINAL`, then we will probably need to revert this back, but for now, I agree we should go for this code reuse. No action needed.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
TadCordle: Yeah I'm thinking about going back over this logic, although at the same time I feel like the deprecation warning might be enough to clue people into the problem if they're getting unexpected results with both configuration options present.

chanseokoh: The difference is that the user will get to know something is wrong far later at runtime when a wrong timestamp matters and it may actually take some time until they notice there is a problem. And people may not really pay much attention to check all warnings. It feels bad to make the code complex for a tiny thing, but they will be gone once we get to 2.0, and considering the benefit, I think it's worth. I think it could really preventing people from opening new issues from this.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
TadCordle: Should the same constant be available for `filesModificationTime`?

chanseokoh: That's viable. I'll file a new issue.

TadCordle: Ok, I can still take care of it in this PR since they share this code path.

chanseokoh: I'd like to hold on on that. Created #1889 and explained the reason.

chanseokoh: I think the message should include the keyword so that people know what to do to preserve the behavior of `<useCurrentTimestamp>`

"... to specify an ISO 8601 date-time or a special keyword ..."

TadCordle: Yeah, doing that now, forgot to do that on the gradle side.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MojoCommon.java
```diff
@@ -38,6 +38,22 @@
...
+      if (!jibPluginConfiguration.getCreationTime().equals("EPOCH_PLUS_SECOND")) {
+        throw new IllegalArgumentException(
+            "You cannot configure both <container><useCurrentTimestamp> and "
+                + "<container><creationTime>");
+      }
+      jibPluginConfiguration
+          .getLog()
+          .warn(
+              "<container><useCurrentTimestamp> is deprecated; use <container><creationTime> with "
+                  + "the value CURRENT_TIME instead");
```
chanseokoh: `USE_CURRENT_TIMESTAMP`

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -282,13 +277,21 @@ private static float getBuildTimeFromVerifierLog(Verifier verifier) throws IOExc
...
+        "1970-01-01T00:00:01Z",
         new Command("docker", "inspect", "-f", "{{.Created}}", imageReference).run().trim());
   }
 
+  private static void assertCreationTimeIsAfter(Instant before, String imageReference)
+      throws IOException, InterruptedException {
+    String inspect =
+        new Command("docker", "inspect", "-f", "{{.Created}}", imageReference).run().trim();
+    Instant after = Instant.parse(inspect);
+    Assert.assertTrue(after.isAfter(before) || after.equals(before));
```
chanseokoh: Probably it's safe to remove `equals(before)`? It is desirable because otherwise the assertion will also succeed for a fixed timestamp. This should be for explicitly testing USE_CURRENT_TIMESTAMP is working.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/EmptyProjectIntegrationTest.java
```diff
@@ -67,7 +67,7 @@ public void testBuild_empty() throws IOException, InterruptedException, DigestEx
             + System.nanoTime();
     Assert.assertEquals("", JibRunHelper.buildAndRun(emptyTestProject, targetImage));
     assertDockerInspect(targetImage);
-    JibRunHelper.assertCreationTimeEpoch(targetImage);
+    JibRunHelper.assertCreationTimeEpochPlusOne(targetImage);
```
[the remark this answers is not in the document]
TadCordle: Container creation time has always been epoch. This PR changes it to epoch+1 so the `EPOCH_PLUS_ONE` value is consistent with the file modification timestamp configuration.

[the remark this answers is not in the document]
chanseokoh: Aha. I think this change is worth documenting in CHANGELOG.

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/complex-build.gradle
```diff
@@ -30,7 +30,7 @@ jib {
     }
   }
   container {
-    useCurrentTimestamp = true
+    creationTime = 'USE_CURRENT_TIMESTAMP'
```
chanseokoh: Oops, I think some of my comments got lost. Is this being tested explicitly (using the keyword in the new config)? For Gradle, I couldn't find in this PR.

TadCordle: Yes https://github.com/GoogleContainerTools/jib/blob/5d3d0d673b25c5533a39f1780267d8bf4730df5d/jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build.gradle#L23

TadCordle: I think that's handled here https://github.com/GoogleContainerTools/jib/pull/1888/files#diff-4e0bf50db814ac7da352939e82b1240eR310

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/TaskCommonTest.java
```diff
@@ -41,14 +41,16 @@
...
   @Before
   public void setUp() {
-    System.clearProperty("jib.extraDirectory.path");
-    System.clearProperty("jib.extraDirectory.permissions");
-    System.clearProperty("jib.extraDirectories.paths");
-    System.clearProperty("jib.extraDirectories.permissions");
+    Assert.assertNull(System.getProperty("jib.extraDirectory.path"));
+    Assert.assertNull(System.getProperty("jib.extraDirectory.permissions"));
+    Assert.assertNull(System.getProperty("jib.extraDirectories.paths"));
+    Assert.assertNull(System.getProperty("jib.extraDirectories.permissions"));
```
chanseokoh: I guess things got awry when merging with master. These should retain `clearProperty()`.

PR #1890, opened 2019-08-05
Minor tests cleanup

loosebazooka:
separated out from #1835


jib-maven-plugin/src/test/resources/maven/application/source/HelloWorld.java
```diff
@@ -0,0 +1 @@
+
```
loosebazooka: google java format adds new lines to these empty files, I dunno why

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenRawConfigurationTest.java
```diff
@@ -101,7 +102,8 @@ public void testGetters() {
     Assert.assertEquals(
         Arrays.asList("--log", "info"), rawConfiguration.getProgramArguments().get());
     Assert.assertEquals(
-        new HashSet<>(Arrays.asList("additional", "tags")), rawConfiguration.getToTags());
+        new HashSet<>(Arrays.asList("additional", "tags")),
+        Sets.newHashSet(rawConfiguration.getToTags()));
```
TadCordle: What's this change for?

chanseokoh:
@loosebazooka I'm taking care of this in #1886.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenRawConfigurationTest.java
loosebazooka: It's from error prone: "java.lang.Iterable does not have well-defined equals behavior"

loosebazooka: `rawConfiguration.getToTags()` returns iterable.

loosebazooka:
@chanseokoh ah okay, I guess more than a few things from #1835 have been/are being fixed. I'll remove it from this change.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenRawConfigurationTest.java
chanseokoh: I don't remember why `getToTags()` is returning `Iterable<String>` instead of `Set<String>`. Maybe we can try making it return `Set<String>`.

PR #1899, opened 2019-08-09
Simplify unauth handling code

chanseokoh:
Initial batch for #1898.

I think I will be able to simplify code further after this.

Issue #1358, opened 2018-12-17
Progress feedback cleanup: Refactor out ProjectProperties#waitForLoggingThread and ProjectProperties#getEventHandlers


coollog:
Originally from #1297 

Refactor out `ProjectProperties#waitForLoggingThread` and `ProjectProperties#getEventHandlers`

diego92sigma6:
Closing due to cleanup. Please reopen if needed.

PR #1901, opened 2019-08-12
Refactor PluginConfigurationProcessor and Maven/Gradle plugin code

### LINKED ISSUE #1898: Simplify HelpfulSuggestions.forNoCredentialsDefined
The method employs some non-trivial logic to re-construct the image reference, but I don't get. The image reference should just be `registry + "/" + repository`. There may have been a reason that we did this, but whatever it was, I think now it is unnecessary. Moreover, after I checked the code involving this, `noCredentialsDefined...` is not accurate anyway and misleading actually because it doesn't account for the credentials defined in `~/.docker/config.json` anyway.

```java
  public String forNoCredentialsDefined(String registry, String repository) {
    Preconditions.checkNotNull(baseImageReference);
    Preconditions.checkNotNull(targetImageReference);

    String unauthorizedEntity = registry;
    if (registry.equals(baseImageReference.getRegistry())
        && repository.equals(baseImageReference.getRepository())
        && noCredentialsDefinedForBaseImage) {
      unauthorizedEntity = baseImageReference.toString();
    } else if (registry.equals(targetImageReference.getRegistry())
        && repository.equals(targetImageReference.getRepository())
        && noCredentialsDefinedForTargetImage) {
      unauthorizedEntity = targetImageReference.toString();
    }

     return suggest(
         "make sure your credentials for '"
            + unauthorizedEntity
             + "' are set up correctly. See "
             + "https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#what-should-i-do-when-the-registry-responds-with-unauthorized for help");
   }
```

chanseokoh:
Closes #1898.

As mentioned in #1899, I can greatly simplify the code further. Before #1899, the final "helpful suggestion" instance could be created only after we processed the configuration. Now, there is no need, so we can directly execute `JibBuildRunner` inside `PluginConfigurationProcessor`.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -100,42 +97,19 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-                pluginConfigurationProcessor.getContainerizer(),
-                projectProperties::log,
-                new MavenHelpfulSuggestions(HELPFUL_SUGGESTIONS_PREFIX));
-
-      } finally {
-        // TODO: This should not be called on projectProperties.
-        projectProperties.waitForLoggingThread();
-        getLog().info("");
-      }
+      PluginConfigurationProcessor.runJibForDockerDaemonImage(
```
loosebazooka: I think this is generally nicer and cleaner, althought it seems weir'd to run the actual build from a static method defined in `PluginConfigurationProcessor` ?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -179,6 +155,10 @@ public void buildDocker()
     } catch (InvalidImageReferenceException ex) {
       throw new GradleException(
           HelpfulSuggestions.forInvalidImageReference(ex.getInvalidReference()), ex);
+
+    } finally {
+      // TODO: This should not be called on projectProperties.
```
TadCordle: I think these TODOs are a bit redundant since there's already one in `ProjectProperties`.

chanseokoh: This is about #1358. I can make the comment clearer.

TadCordle: Right, I meant there's already a todo about moving these methods out of `ProjectProperties`, inside the interface itself. I don't think we need extra todos where they're called as well.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
chanseokoh: I restored the original method name (`process...ForDockerDaemonImage`) and made it return a configured `JibBuildRunner` instance, on which this plugin code will call `runBuild()` to kick the tire.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -105,37 +102,17 @@ public void buildDocker()
...
+        GradleProjectProperties.getForProject(getProject(), getLogger());
     try {
-      RawConfiguration gradleRawConfiguration = new GradleRawConfiguration(jibExtension);
-      GradleProjectProperties projectProperties =
-          GradleProjectProperties.getForProject(getProject(), getLogger());
-
-      PluginConfigurationProcessor pluginConfigurationProcessor =
-          PluginConfigurationProcessor.processCommonConfigurationForDockerDaemonImage(
-              gradleRawConfiguration,
+      PluginConfigurationProcessor.processCommonConfigurationForDockerDaemonImage(
```
loosebazooka: now that I see it, I think a name like `create....` or `get....` might make sense.

chanseokoh: Renamed the public methods to start with `createJibBuildRunner...` and also updated an out-of-date comment.

Issue #1905, opened 2019-08-15
allow using saved images (tar balls) as base images


aespinosa:
**Environment**:

- *Jib version:* 1.4.0
- *Build tool:* maven 3.6.1
- *OS:* macOS

**Description of the issue**:

The use case might have similarities to #1807 in a way.  But in this case, I'm using a custom base image only for my java app (like adding kerberos packages to upstream java images).

And since i'm only using it for my project, it would make sense to not publish this intermediary base image and just publish the final image at the end of the build:

**Expected behavior**:

```
docker build -t my-custom-image .
docker save my-custom-image -o base.tar
# maybe it will be something like this??
mvn jib:loadBaseCache -Dfrom=base.tar
mvn jib:build
```

** pom.xml **

```
<!---  ... --->
<from>
   <image>./base.tar</image> 
</from>
<!--- ... --->
```

raizoor:
@aespinosa 

Do you want to use a local image instead of a base image of a registry?
If yes, see the changelog to new release:
```
When building to a registry, Jib now skips downloading and caching base image layers that already exist in the target registry. This feature will be particularly useful in CI/CD environments. However, if you want to force caching base image layers locally, set the system property -Djib.alwaysCacheBaseImage=true (#1840)
```

TadCordle:
Hi @aespinosa,

This is actually being worked on, alongside support for using the local Docker cache for base images (#1468). The tentative plan is to allow configuring different types of base images using a URL scheme like the following:

registry:
```
<from>
  <image>gcr.io/project/image</image>
</from>

<from>
  <image>registry://gcr.io/project/image</image>
</from>
```

docker daemon:
```
<from>
  <image>docker://docker-daemon/image</image>
</from>
```

tar:
```
<from>
  <image>tar:///path/to/tarball.tar</image>
</from>
```

I'll leave this open to track the tar half, since the other issue is focused mostly on docker daemon base images.

aespinosa:
thanks @TadCordle .  that URL scheme looks neat!

TadCordle:
@aespinosa @raizoor v1.6.0 has been released with this feature. You can configure Jib's source image to tar on the filesystem the `tar://` prefix:

Gradle: `jib.from.image = 'tar:///path/to/saved.tar'`
Maven: `<from><image>tar:///path/to/saved.tar</image></from>`

PR #1906, opened 2019-08-15
Add ExtractTarStep and SaveDockerStep

TadCordle:
Towards #1468 and #1905.

Will probably need to figure out ways to optimize this in a follow-up. For local images, we unfortunately need to compress uncompressed layers and recalculate the compressed layer digests, and this is a bit slow.

For Jib-built images, we name all the layers with their digest, so we could maybe leverage that, but there's no guarantee that this is true for all tarballs (e.g. a `docker save`d tarball contains uncompressed layers named "layer.tar" in a folder that is named with the uncompressed layer's digest).

Edit: We discussed in the design meeting that it might make sense to save the compressed layer + digest in a folder named with the layer's diff id in Jib's cache; this way we can retrieve the diff ID for a layer in a local base image, use that to check if the layer is in the cache, and only do all the work of compressing/calculating the digest if it isn't.

Also todo: check blob existence on target registry.


jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
```diff
@@ -77,5 +80,31 @@ public static String writeToString(Blob blob) throws IOException {
     return byteArrayOutputStream.toByteArray();
   }
 
+  /**
+   * Gets a {@link Blob} that is {@code blob} compressed.
+   *
+   * @param blob the {@link Blob} to compress
+   * @return the compressed {@link Blob}
+   */
+  public static Blob compress(Blob blob) {
```
TadCordle: Note: these two methods were moved from `CacheTest`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,140 @@
...
+
+    // Process layer blobs
+    // TODO: Parallelize/other optimizations; calculating layer digests is slow for large layers
+    List<PreparedLayer> layers = new ArrayList<>();
+    V22ManifestTemplate newManifest = new V22ManifestTemplate();
+    for (int index = 0; index < layerFiles.size(); index++) {
+      Path file = destination.resolve(layerFiles.get(index));
+
+      // Compress layers if necessary and calculate the digest/size
+      Blob blob = layersAreCompressed ? Blobs.from(file) : Blobs.compress(Blobs.from(file));
```
TadCordle: Note: not sure how much we need to worry about reproducibility at this point? Do we need to wipe layer file blob timestamps?

loosebazooka: Nah, it doesn't fee like our job to mess with and mutate the base layers, at least at this point.

jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
```diff
@@ -77,5 +80,31 @@ public static String writeToString(Blob blob) throws IOException {
...
+
+  /**
+   * Gets a {@link Blob} that is {@code blob} decompressed.
+   *
+   * @param blob the {@link Blob} to decompress
+   * @return the decompressed {@link Blob}
+   * @throws IOException if an I/O exception occurs
+   */
+  public static Blob decompress(Blob blob) throws IOException {
+    return Blobs.from(new GZIPInputStream(new ByteArrayInputStream(Blobs.writeToByteArray(blob))));
```
chanseokoh: This was reasonable in test code, but I expect there must be a more efficient way for production. It is very efficient to first write all the blob content to an array in memory and then have `GZIPInputStream` to read the content out of the array. We should leverage stream and definitely avoid writing holding the content in memory.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.zip.GZIPInputStream;
+
+/** Extracts a tar file base image. */
+public class ExtractTarStep implements Callable<LocalImage> {
+
+  static class LocalImage {
+    Image baseImage;
```
chanseokoh: Make the fields `final`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,140 @@
...
+
+    // Check the first layer to see if the layers are compressed already. 'docker save' output is
+    // uncompressed, but a jib-built tar has compressed layers.
+    boolean layersAreCompressed =
+        layerFiles.size() > 0 && isGzipped(destination.resolve(layerFiles.get(0)));
+
+    // Process layer blobs
+    // TODO: Optimize; compressing/calculating layer digests is slow
+    List<PreparedLayer> layers = new ArrayList<>();
+    V22ManifestTemplate newManifest = new V22ManifestTemplate();
```
chanseokoh: Can we have a more descriptive name than "new"? Being new is a relative notion and it may not be straightforward without a context.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Override
+  public LocalImage call()
+      throws IOException, LayerCountMismatchException, BadContainerConfigurationFormatException {
+    TarExtractor.extract(tarPath, destination);
+    DockerManifestEntryTemplate loadManifest =
+        new ObjectMapper()
+            .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true)
+            .readValue(
+                Files.newInputStream(destination.resolve("manifest.json")),
```
chanseokoh: Let's close the stream as soon as we are done with it.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,140 @@
...
+  static boolean isGzipped(Path path) throws IOException {
+    try (InputStream inputStream = new BufferedInputStream(Files.newInputStream(path))) {
+      inputStream.mark(2);
+      int magic = (inputStream.read() & 0xff) | ((inputStream.read() << 8) & 0xff00);
+      return magic == GZIPInputStream.GZIP_MAGIC;
+    }
+  }
+
+  private final Path tarPath;
+  private final Path destination;
```
chanseokoh: I think these are unused.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,140 @@
...
+    LocalImage(Image baseImage, List<PreparedLayer> layers) {
+      this.baseImage = baseImage;
+      this.layers = layers;
+    }
+  }
+
+  @VisibleForTesting
+  static boolean isGzipped(Path path) throws IOException {
+    try (InputStream inputStream = new BufferedInputStream(Files.newInputStream(path))) {
+      inputStream.mark(2);
```
chanseokoh: Is it possible to put some reference for this test logic as a comment?

And just a nit: I don't think `BufferedInputStream` is needed, since we only read 3 bytes.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,140 @@
...
+      // https://containers.gitbook.io/build-containers-the-hard-way/#docker-load-format
+      CachedLayer layer =
+          CachedLayer.builder()
+              .setLayerBlob(blob)
+              .setLayerDigest(blobDescriptor.getDigest())
+              .setLayerSize(blobDescriptor.getSize())
+              .setLayerDiffId(configuration.getLayerDiffId(index))
+              .build();
+
+      // TODO: Check blob existence on target registry (online mode only)
```
chanseokoh: To me, the comment is not clear why we need/want to do this for what. (I'm wondering it too.)

jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
TadCordle: This isn't being used anywhere so I might just move this back to the test class.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
TadCordle: They're used in the first line of `call()`: `TarExtractor.extract(tarPath, destination)`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
TadCordle: I'll just remove this, I mixed up what the purpose of the blob existence check was.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,147 @@
...
+      throws IOException, LayerCountMismatchException, BadContainerConfigurationFormatException {
+    TarExtractor.extract(tarPath, destination);
+
+    InputStream manifestStream = Files.newInputStream(destination.resolve("manifest.json"));
+    DockerManifestEntryTemplate loadManifest =
+        new ObjectMapper()
+            .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true)
+            .readValue(manifestStream, DockerManifestEntryTemplate[].class)[0];
+    manifestStream.close();
+    ContainerConfigurationTemplate configuration =
```
chanseokoh: `containerConfiguration`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -0,0 +1,147 @@
...
+    // Process layer blobs
+    // TODO: Optimize; compressing/calculating layer digests is slow
+    List<PreparedLayer> layers = new ArrayList<>();
+    V22ManifestTemplate v22Manifest = new V22ManifestTemplate();
+    for (int index = 0; index < layerFiles.size(); index++) {
+      Path file = destination.resolve(layerFiles.get(index));
+
+      // Compress layers if necessary and calculate the digest/size
+      Blob blob = layersAreCompressed ? Blobs.from(file) : Blobs.compress(Blobs.from(file));
+      BlobDescriptor blobDescriptor = blob.writeTo(ByteStreams.nullOutputStream());
```
chanseokoh: Potential optimization idea:

So, we end up writing each (compressed) blob content (layer) two times: first to compute the layer digest (which is required to assemble the manifest) and secondly to actually push to the registry. (I may be wrong about this though.) We can push (we currently do) the manifest as the last step, so I think theoretically it's possible that you start pushing base image layers early while compressing on-the-fly and this will get us their digests. This way, we may be able to write compressed contents once.

But I may be missing something in this process. Can't remember all the details.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/SaveDockerStep.java
```diff
@@ -0,0 +1,45 @@
...
+  private final DockerClient dockerClient;
+
+  SaveDockerStep(BuildConfiguration buildConfiguration, DockerClient dockerClient) {
+    this.buildConfiguration = buildConfiguration;
+    this.dockerClient = dockerClient;
+  }
+
+  @Override
+  public Path call() throws IOException, InterruptedException {
+    Path outputPath = Files.createTempDirectory("jib-docker-save").resolve("out.tar");
```
chanseokoh: Since the file will usually be huge, and we will create a new tar every time the user uses a local Docker image, I think it is good to try a cleanup later. (But don't crash if we fail to clean up.) Maybe we could explicitly delete (`MoreFiles.deleteRecursively()`) the directly once we know we'll no longer use it, or just use some built-in cleanup facilities. I remember something like `deleteOnExit()`. Not sure if it works for deleting files recursively. Or, we could use a [shutdown hook](https://github.com/GoogleCloudPlatform/appengine-plugins-core/blob/8316f066213315badc9a7353a428edc576ccb544/src/main/java/com/google/cloud/tools/managedcloudsdk/components/WindowsBundledPythonCopier.java#L58).

Maybe we should do this for the untarred directory? What do you think in general?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
TadCordle: This sounds like it may be possible, but would probably require more build step refactoring, so I'll hold off on it for now.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/SaveDockerStep.java
TadCordle: I added shutdown hooks to delete both the saved tar and the extracted contents, although I think in the future we should probably cache the extracted layers instead of deleting them.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
chanseokoh: Yeah, let's just keep this in mind. This, and also the idea to untar the output of `docker save` on-the-fly and skip writing the tar onto the disk. Oh, and the possibility to reuse the tar (untarred or not), if we can correctly notice changes. Maybe file an issue to track regardless of how difficult they are?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/FileOperations.java
```diff
@@ -85,5 +87,24 @@ public static OutputStream newLockingOutputStream(Path file) throws IOException
     return Channels.newOutputStream(channel);
   }
 
+  /**
+   * Sets up a shutdown hook that tries to delete a directory.
```
chanseokoh: Seems like `deleteRecursively` will work just fine for a file too.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
TadCordle: I'll open one for the caching ideas and another for the other optimizations (on-the-fly untar, compression while pushing) after this gets merged.

PR #1919, opened 2019-08-20
Add local base image support to jib core 

TadCordle:
Just starting on this and still not even sure it works yet, lots of tests to add/fix. Just posting progress.

Towards #1468 and #1905.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -19,8 +19,10 @@ package com.google.cloud.tools.jib.builder.steps;
 import com.google.cloud.tools.jib.api.Credential;
 import com.google.cloud.tools.jib.blob.BlobDescriptor;
 import com.google.cloud.tools.jib.builder.ProgressEventDispatcher;
+import com.google.cloud.tools.jib.builder.steps.ExtractTarStep.LocalImage;
 import com.google.cloud.tools.jib.builder.steps.PullBaseImageStep.ImageAndAuthorization;
 import com.google.cloud.tools.jib.configuration.BuildConfiguration;
+import com.google.cloud.tools.jib.configuration.ImageConfiguration;
 import com.google.cloud.tools.jib.docker.DockerClient;
 import com.google.cloud.tools.jib.global.JibSystemProperties;
 import com.google.cloud.tools.jib.http.Authorization;
@@ -31,6 +33,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
@@ -59,6 +62,7 @@ public class StepsRunner {
           new IllegalStateException("invalid usage; required step not configured"));
     }
 
+    private Future<Path> tarPath = failedFuture();
     private Future<ImageAndAuthorization> baseImageAndAuth = failedFuture();
     private Future<List<Future<PreparedLayer>>> baseImageLayers = failedFuture();
     @Nullable private List<Future<PreparedLayer>> applicationLayers;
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -208,6 +229,37 @@ private void authenticatePush() {
...
+        executorService.submit(
+            () -> new ImageAndAuthorization(localImageFuture.get().baseImage, null));
+    results.baseImageLayers =
+        executorService.submit(
+            () ->
+                localImageFuture
+                    .get()
+                    .layers
+                    .stream()
+                    .map(layer -> executorService.submit(() -> layer))
```
chanseokoh: Does `.map(Futures::immediateFuture)` work?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -208,6 +229,37 @@ private void authenticatePush() {
...
   }
 
+  private void saveDocker() {
+    Optional<DockerClient> dockerClient =
+        buildConfiguration.getBaseImageConfiguration().getDockerClient();
+    Verify.verify(dockerClient.isPresent());
+
+    results.tarPath =
+        executorService.submit(
+            () -> new SaveDockerStep(buildConfiguration, dockerClient.get()).call());
```
chanseokoh: In this case where `SaveDockerStep` doesn't depend on any future value, no need to wrap it in an additional lambda.

`.submit(new SaveDockerStep(buildConfiguration, dockerClient.get()))`

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ImageConfiguration.java
```diff
@@ -30,6 +34,8 @@
 
     private ImageReference imageReference;
     private ImmutableList<CredentialRetriever> credentialRetrievers = ImmutableList.of();
+    @Nullable private DockerClient dockerClient;
+    @Nullable private Path tarPath;
```
chanseokoh: I have had some reservation, as this feels a bit like a quick ad-hoc fix and doesn't feel like OO-style. `dockerClient` and `tarPath` are incompatible in that they shouldn't be both set, and as such, we need some code to guard it. And relying on on the assumption that a non-null `dockerClient` implies the semantic that this `ImageConfiguation` represents a local Docker image (and likewise that a non-null `tarPath` implies a tarball image) whereas a null value implies a remote registry image seems requiring such external domain knowledge in understanding the code. I wish given an image configuration it clearly presented what the image is. Though I can't think of a good and concrete suggestion right now. @GoogleContainerTools/java-tools-build thoughts?

TadCordle: I guess we could make `RegistryImage`, `DockerDaemonImage`, and `TarImage` implement a common `BaseImage` interface, store a `BaseImage` in `ImageConfiguration`, and check the type/cast to get the required fields in `StepsRunner`?

TadCordle: I'm not sure how this affects the jib-core public API though.

TadCordle: In fact if we do this, I think we can remove `ImageConfiguration` completely. In which case I think I'd rather refactor this in a follow-up.

loosebazooka: Does setting `tarPath` mean that `imageReference` can be optional too?

TadCordle: Yes for base images, no for target images. It's handled in a somewhat hacky way [here](https://github.com/GoogleContainerTools/jib/pull/1919/files#diff-d390f64e06290a875d4e6a149d0f18ccR98) that I'd like to improve once this part is figured out.

loosebazooka: > In fact if we do this, I think we can remove ImageConfiguration completely, but that can be handled in a follow-up.

I think I would prefer we designed removing `ImageConfiguration` first and then submitted this PR.

TadCordle: What I had in mind for removing `ImageConfiguration` follows my common interface idea; if each image type implemented a common `ContainerImage` interface, then `ImageConfiguration` could just contain that instead of the 4 fields we have now. And at that point, it's unnecessary, so we could remove it and just stick the `ContainerImage` straight into `BuildConfiguration`, since those will always have the fields we need (after casting to the correct type first, at least).

I didn't want to remove it in this PR because it's used in a lot of places and would probably make this PR a lot bigger, and also I'm not sure if the check/cast way of going about this is better than what we have now.

chanseokoh: I'm fine with taking care of this after merging this. And I still have no good suggestion as how exactly.

PR #1925, opened 2019-08-21
Fix a JavaContainerBuilder error message

TadCordle:


jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -514,8 +514,8 @@ public JibContainerBuilder toContainerBuilder() throws IOException {
       throw new IllegalStateException(
           "Failed to construct entrypoint on JavaContainerBuilder; "
               + "jvmFlags were set, but mainClass is null. Specify the main class using "
-              + "JavaContainerBuilder#setMainClass(String), or consider using a "
-              + "jib.frontend.MainClassFinder to infer the main class.");
+              + "JavaContainerBuilder#setMainClass(String), or consider using MainClassFinder to "
```
loosebazooka: So this is the underlying API right? The gradle/maven plugin user will never see this?

TadCordle: Correct.

PR #1926, opened 2019-08-22
Adjust TarImage API in preparation for local base image support

### LINKED ISSUE #1918: Adjust TarImage api
With local base image support coming up, the `TarImage` api doesn't make a lot of sense at the moment and should probably be changed. To create a `TarImage` you need to do `TarImage.named("gcr.io/...").saveTo(".../image.tar")`. This doesn't make sense for tar base images since you 
1. are using an existing file, not saving to a file with `saveTo`, and
2. only require the image name when you are building a tar, not retrieving one

I think we need a way of skipping setting the image name and should just throw an `IllegalArgumentException` or something if we try to build a tar without an image reference.

One possible idea is to switch the order of arguments and rename `saveTo`, like `TarImage.at(".../image.tar").named("gcr.io/...");` where the last named call is optional. Here, the image name isn't required for base images, and the method names are flexible enough to be used for both base images and target images.

TadCordle:
Fixes #1918. Changes the usage from `TarImage.named(...).saveTo(...)` to `TarImage.at(...).named(...)`. Throws an exception if the target image name is missing.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
@@ -42,12 +42,12 @@ public class TarImage {
     return new TarImage(path);
   }
 
-  private final Path outputFile;
+  private final Path path;
   @Nullable private ImageReference imageReference;
 
   /** Instantiate with {@link #at}. */
-  private TarImage(Path outputFile) {
-    this.outputFile = outputFile;
+  private TarImage(Path path) {
+    this.path = path;
   }
 
   /**
@@ -74,13 +74,8 @@ public class TarImage {
     return named(ImageReference.parse(imageReference));
   }
 
-  /**
-   * Gets the output file to save the tarball archive to.
-   *
-   * @return the output file
-   */
   Path getPath() {
-    return outputFile;
+    return path;
   }
 
   Optional<ImageReference> getImageReference() {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
@@ -33,7 +33,7 @@ import javax.annotation.Nullable;
 public class TarImage {
 
   /**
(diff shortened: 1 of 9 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
```diff
@@ -17,80 +17,73 @@
...
-  private final Path outputFile;
-
-  /** Instantiate with {@link #named}. */
-  private TarImage(ImageReference imageReference, Path outputFile) {
-    this.imageReference = imageReference;
-    this.outputFile = outputFile;
-  }
-
   /**
    * Gets the output file to save the tarball archive to.
```
chanseokoh: The description and the `@return` don't hold anymore.

jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
```diff
@@ -17,80 +17,73 @@
...
+   * @return a new {@link TarImage}
+   */
+  public static TarImage at(Path path) {
+    return new TarImage(path);
+  }
 
-    private Builder(ImageReference imageReference) {
-      this.imageReference = imageReference;
-    }
+  private final Path outputFile;
```
chanseokoh: Why "output", given that it can point to a base image?

jib-core/src/main/java/com/google/cloud/tools/jib/api/TarImage.java
```diff
@@ -17,80 +17,73 @@
 package com.google.cloud.tools.jib.api;
 
 import java.nio.file.Path;
+import java.util.Optional;
+import javax.annotation.Nullable;
 
 /**
  * Builds to a tarball archive.
```
chanseokoh: "Builds" doesn't apply anymore. (Or, does it still from the end-user perspective?)

TadCordle: I think I'll make these comment changes in the other PR since this one doesn't cover base image support.

PR #1930, opened 2019-08-26
Add plugin configuration for local base images

### LINKED ISSUE #1468: Support for using a local Docker cache for base images

### LINKED ISSUE #1905: allow using saved images (tar balls) as base images

TadCordle:
Fixes #1468, fixes #1905.

Adds plugin configuration for different base image types. New configurations:
- `registry://ima.ge/name` to use a registry image at `ima.ge/name`
- `docker://imagename` to use a Docker daemon image named `imagename`
- `tar:///path/to/tar.tar` to use an image tarball in the filesystem at path `/path/to/tar.tar`


jib-core/src/main/java/com/google/cloud/tools/jib/api/Jib.java
```diff
@@ -30,6 +42,15 @@
    */
   public static JibContainerBuilder from(String baseImageReference)
       throws InvalidImageReferenceException {
+    if (baseImageReference.startsWith(DOCKER_DAEMON_IMAGE_PREFIX)) {
+      return from(
+          DockerDaemonImage.named(
+              ImageReference.parse(
```
chanseokoh: Can't you use `named(String)` which calls `ImageReference.parse()` internally?

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Support for local base images using a URL scheme in `jib.from.image`. Use the format `docker://name-of-image` to use a base image stored in the local Docker daemon, or `tar:///path/to/file.tar` to use a base image tarball stored on the filesystem ([#1468](https://github.com/GoogleContainerTools/jib/issues/1468), [#1905](https://github.com/GoogleContainerTools/jib/issues/1905))
```
chanseokoh: If `tar://` supports a relative path (if so, I think it should be relative to the project root), worth mentioning?

jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -154,6 +158,22 @@ private static String buildAndRun(
     return output;
   }
 
+  private static String buildAndRunFromLocalBase(
+      Path projectRoot, String targetImage, String baseImage)
+      throws VerificationException, IOException, InterruptedException {
+    Verifier verifier = new Verifier(projectRoot.toString());
```
chanseokoh: This is a minor point, but how about reusing `build(projectRoot, imageReference, pomXml, buildTwice)`? `pom-localbase.xml` can embed the base image reference. Will need two POM files though. We are explicitly testing local image references for the config, so to me, having concrete config values as dedicated POM samples makes sense.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/JibRunHelper.java
```diff
@@ -65,6 +65,24 @@ static String buildAndRun(
...
+  static String buildAndRunFromLocalBase(String target, String base)
+      throws IOException, InterruptedException, DigestException {
+    BuildResult buildResult =
+        SingleProjectIntegrationTest.simpleTestProject.build(
+            "clean",
+            "jib",
+            "-Djib.useOnlyProjectCache=true",
+            "-Djib.console=plain",
+            "-D_TARGET_IMAGE=" + target,
+            "-D_BASE_IMAGE=" + base,
```
chanseokoh: Similarly, how about reusing `buildAndRun()` or `buildToDockerDaemon()`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ConfigurationPropertyValidator.java
```diff
@@ -66,11 +66,11 @@
     // Warn if a system property is missing
     String missingProperty =
         "%s system property is set, but %s is not; attempting other authentication methods.";
-    if (!commandlinePassword.isEmpty() && commandlineUsername.isEmpty()) {
+    if (!commandlinePassword.isEmpty()) {
       logger.accept(
           LogEvent.warn(String.format(missingProperty, passwordProperty, usernameProperty)));
     }
-    if (!commandlineUsername.isEmpty() && commandlinePassword.isEmpty()) {
+    if (!commandlineUsername.isEmpty()) {
```
chanseokoh: This makes sense. Then, can we do the same for `auth` checking below? The logic has the same structure.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -197,24 +195,12 @@ static JibContainerBuilder processCommonConfiguration(
...
-        projectProperties,
-        baseImage,
-        baseImageReference,
-        PropertyNames.FROM_AUTH_USERNAME,
-        PropertyNames.FROM_AUTH_PASSWORD,
-        rawConfiguration.getFromAuth(),
-        inferredAuthProvider,
-        rawConfiguration.getFromCredHelper().orElse(null));
-
     // 3. create and configure JibContainerBuilder
```
chanseokoh: nit: the number doesn't match now. Just throw the numbers away?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -254,6 +240,56 @@ static JibContainerBuilder processCommonConfiguration(
...
+  static JavaContainerBuilder getJavaContainerBuilderWithBaseImage(
+      RawConfiguration rawConfiguration,
+      ProjectProperties projectProperties,
+      InferredAuthProvider inferredAuthProvider)
+      throws IncompatibleBaseImageJavaVersionException, InvalidImageReferenceException,
+          FileNotFoundException {
+    // Use image configuration as-is if it's a local base image
+    String baseImageConfig =
+        rawConfiguration.getFromImage().orElse(getDefaultBaseImage(projectProperties));
+    if (baseImageConfig.startsWith(Jib.DOCKER_DAEMON_IMAGE_PREFIX)
```
chanseokoh: I wonder if it may be OK to check Java versions for Docker images. What's the rationale to exclude here?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -704,13 +721,34 @@ public void testGetBaseImage_projectHigherThanJava11() {
   }
 
   @Test
-  public void testGetBaseImage_incompatibleJava8BaseImage() {
+  public void testGetJavaContainerBuilderWithBaseImage_nonRegistry()
+      throws IncompatibleBaseImageJavaVersionException, IOException, InvalidImageReferenceException,
+          CacheDirectoryCreationException {
+    Mockito.when(rawConfiguration.getFromImage()).thenReturn(Optional.of("tar:///path/to.tar"));
```
chanseokoh: Could add a Docker image test quickly?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -704,13 +721,34 @@ public void testGetBaseImage_projectHigherThanJava11() {
...
+  }
+
+  @Test
+  public void testGetJavaContainerBuilderWithBaseImage_registry()
+      throws IncompatibleBaseImageJavaVersionException, InvalidImageReferenceException, IOException,
+          CacheDirectoryCreationException {
+    Mockito.when(rawConfiguration.getFromImage()).thenReturn(Optional.of("ima.ge/name"));
+    ImageConfiguration result = getCommonImageConfiguration();
+    Assert.assertEquals("ima.ge/name", result.getImage().toString());
+    Assert.assertFalse(result.getDockerClient().isPresent());
```
chanseokoh: I think we also need `Assert.assertFalse(result.getTarPath().isPresent())`, since a tar image configuration returns a non-null `ImageReference`. That said, I became curious as to what exactly `getImage()` will return for a tar base image?

TadCordle: Right now it's just `scratch` as a placeholder; there's a TODO for this when we figure out what to do about the whole exclusive parameter thing in `ImageConfiguration`.

jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
TadCordle: At least for the tar case, the base image path needs to be passed in at runtime since it uses a temporary directory. So I decided to use that for the docker daemon base image test as well.

chanseokoh: Ah, I see. That's fair. Minor point anyway, so I'm fine.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
TadCordle: Yeah, I guess no reason not to.

TadCordle:
Going to add `registry://` back in given recent chat.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
```diff
@@ -71,8 +71,12 @@
    */
   private static final String REFERENCE_REGEX =
       String.format(
-          "^(?:(%s)/)?(%s)(?:(?::(%s))|(?:@(%s)))?$",
-          REGISTRY_REGEX, REPOSITORY_REGEX, TAG_REGEX, DescriptorDigest.DIGEST_REGEX);
+          "^(?:(%s))?(?:(%s)/)?(%s)(?:(?::(%s))|(?:@(%s)))?$",
```
chanseokoh: Do we need this prefix extension that is only specific to Jib? That could remain external and this class may remain general. Also feels more consistent as it doesn't parse `docker://` or `tar://`. Why only `registry://`?

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Support for local base images by prefixing `jib.from.image>` with `docker://` to build from a docker daemon image, or `tar://` to build from a tarball image ([#1468](https://github.com/GoogleContainerTools/jib/issues/1468), [#1905](https://github.com/GoogleContainerTools/jib/issues/1905))
```
chanseokoh: Remove a stray `>` in `in jib.from.image>`

jib-maven-plugin/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Support for local base images by prefixing `jib.from.image>` with `docker://` to build from a docker daemon image, or `tar://` to build from a tarball image ([#1468](https://github.com/GoogleContainerTools/jib/issues/1468), [#1905](https://github.com/GoogleContainerTools/jib/issues/1905))
```
chanseokoh: `<from><image>` since Maven.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
TadCordle: It's in `ImageReference` because of the suggestion to require the registry when using the `registry://` prefix. Although, I'm not really sure if that's something we should do, considering the conclusion we came to about following the URI spec.

If we want to be consistent we can do one of the following:
- parse all of the prefixes in `ImageReference`, but that doesn't work for `tar://` since it's a path, not an image reference.
- remove the prefix before calling `ImageReference.parse`, but then we can't tell if the registry was explicit or not without doing some extra parsing outside of `ImageReference`.
- not require an explicit registry with the `registry://` prefix.

PR #1933, opened 2019-08-27
Add `upstreamClient` property to jib which tools using jib to build and push images can use.

tejal29:
skaffold uses `jib` to build and push java images. In order to track,  skaffold-jib adoption we want to add `UpstreamClient(skaffold-version)` to images pushed to GCR by skaffold via jib.
See https://github.com/GoogleContainerTools/skaffold/pull/2723 for kaniko implementation.

If upstreamClient is present, it will added to the user-agent when pushing to registry

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
@@ -23,7 +23,7 @@ import com.google.common.collect.Range;
 /** Names of system properties defined/used by Jib. */
 public class JibSystemProperties {
 
-  public static final String _JIB_UPSTREAM_CLIENT = "jib.upstreamClient";
+  public static final String UPSTREAM_CLIENT = "_JIB_UPSTREAM_CLIENT";
 
   @VisibleForTesting public static final String HTTP_TIMEOUT = "jib.httpTimeout";
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
@@ -23,6 +23,8 @@ import com.google.common.collect.Range;
 /** Names of system properties defined/used by Jib. */
 public class JibSystemProperties {
 
+  public static final String UPSTREAM_CLIENT = "jib.upstreamClient";
+
   @VisibleForTesting public static final String HTTP_TIMEOUT = "jib.httpTimeout";
 
   @VisibleForTesting static final String CROSS_REPOSITORY_BLOB_MOUNTS = "jib.blobMounts";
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -23,6 +23,8 @@
 /** Names of system properties defined/used by Jib. */
 public class JibSystemProperties {
 
+  public static final String UPSTREAM_CLIENT = "jib.upstreamClient";
```
chanseokoh: Sorry for coming late (after vacation). I don't know what "upstream client" means, but because this is an internal property, I prefer `_JIB_UPSTREAM_CLIENT`, like the one `_JIB_DISABLE_USER_AGENT` we have below. Otherwise, LGTM.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryClientTest.java
```diff
@@ -72,6 +77,19 @@ public void testGetUserAgent() {
     Assert.assertTrue(registryClient.getUserAgent().endsWith(" some user agent suffix"));
   }
 
+  @Test
+  public void testGetUserAgentWithUpstreamClient() {
+    System.setProperty(JibSystemProperties.UPSTREAM_CLIENT, "skaffold-0.34.0");
```
chanseokoh: Oh, BTW, I believe the User-Agent string doesn't really conform to the spec or customs. Some references:

- https://tools.ietf.org/html/rfc7231#section-5.5.3
- https://en.wikipedia.org/wiki/User_agent#Format_for_human-operated_web_browsers

I know people don't follow the spec very well for User-Agent, so I'm fine with the current string. But I was just wondering if you can change it to `skaffold/0.34.0`. I don't really care though.

tejal29: yeah, we should change this in skaffold. +1. I will create a PR over there and change this test.

jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -23,6 +23,8 @@
 /** Names of system properties defined/used by Jib. */
 public class JibSystemProperties {
 
+  public static final String _JIB_UPSTREAM_CLIENT = "jib.upstreamClient";
```
TadCordle: Ah, I think he meant the variable name should stay as `UPSTREAM_CLIENT`, but the value should be `"_JIB_UPSTREAM_CLIENT"`

chanseokoh: Ah, yeah. Good catch.

tejal29: oh wov, i wasnt aware you can have "_" in property names.

chanseokoh:
@tejal29 we just released 1.6.0 that has this `upstreamClient` property. I think Skaffold should put `-D_JIB_UPSTREAM_CLIENT=...`? Is Skaffold already passing this when invoking Jib?

PR #1943, opened 2019-08-29
Fix IOException when both source and target are Docker daemon

### LINKED ISSUE #1942: IOException when building from local base image to docker damon
When using a docker daemon base image, and building to the docker daemon, the following is thrown:

```
...
Caused by: java.io.IOException: This archive contains unclosed entries.
        at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.finish(TarArchiveOutputStream.java:289)
        at org.apache.commons.compress.archivers.tar.TarArchiveOutputStream.close(TarArchiveOutputStream.java:307)
        at com.google.cloud.tools.jib.blob.Blobs.lambda$compress$0(Blobs.java:92)
...
```

TadCordle:
Fixes #1942


jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
```diff
@@ -116,6 +116,21 @@ public void testBasic_dockerDaemonBaseImage()
             targetImageReference.withTag(jibContainer.getDigest().toString()).toString()));
   }
 
+  @Test
+  public void testBasic_dockerDaemonBaseImageToDockerDaemon()
```
loosebazooka: are both the compressed and uncompressed codepaths now tested?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -123,7 +124,15 @@ public LocalImage call()
       Path file = destination.resolve(layerFiles.get(index));
 
       // Compress layers if necessary and calculate the digest/size
-      Blob blob = layersAreCompressed ? Blobs.from(file) : Blobs.compress(Blobs.from(file));
+      Blob blob = Blobs.from(file);
```
loosebazooka: should we solve the underlying issue or calling `Blobs.writeTo` should *not* close the output stream according to the javadoc on the interface https://github.com/GoogleContainerTools/jib/blob/1dd28e1224158bf9cb9d01e6a5edb6423c2d8bbf/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java#L26

but our FileBlob implementation appears to be doing exactly that?

Ofcourse this can be in a followup PR.

TadCordle: I'll just file an issue for this for now. I don't *think* there's a problem with closing the output stream, but we should be clear about whether or not the stream is closed in each implementation.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
TadCordle: I'll add one for compressed -> docker daemon

PR #1950, opened 2019-09-04
Rework I/O error handling

### LINKED ISSUE #1895: improve error message when http timeout
**Environment**:

- *Jib version: 1.4.0*
- *Build tool:* Maven 3.5.4
- *OS: Windows

**Description of the issue**:
While running mvn compile jib:build, mvn timesout it pushes 50% of the built image.
Below is the screenshot of the build failure.

```
      <configuration>
          <to>
            <image>images.test.net/container/kube/tmp/python/
          </to>
        </configuration>
```
timeout in 28 second
only image is placeholder is modified, user and password are supplied in m2 settins file

**Expected behavior**:
It should push the build image

**Steps to reproduce**:
  1. Ran the mvn compile jib:build in example/hello-world

chanseokoh:
Fixes #1895.

I've been mulling over how to handle most of the IO errors, and I settled with logging errors to better show the registry being involved.

Also removed two exceptions that don't seem really necessary. They have not been caught in the inner `callWithAllowInsecureRegistryHandling`, so it was incomplete anyway.


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -160,7 +159,22 @@ static boolean isBrokenPipe(IOException original) {
...
+    try {
+      return callWithAllowInsecureRegistryHandling(initialRequestUrl);
+
+    } catch (IOException ex) {
+      String registry = registryEndpointRequestProperties.getServerUrl();
+      String repository = registryEndpointRequestProperties.getImageName();
+      logError("I/O error for image [" + registry + "/" + repository + "]:");
+      logError("    " + ex.getMessage());
+      if (isBrokenPipe(ex)) {
+        logError(
```
TadCordle: What's the reasoning behind logging the message separately from the exception instead of throwing an exception with a custom message? Just to log the error message even if the exception is caught elsewhere?

chanseokoh: Not just to log. The main reason is to tell the user for which image (registry/repository) the error was for. That's the original issue in #1895, where the user had to spend a lot of time figuring out that it was the connection to `gcr.io` that was timing out. The user didn't know anything about what the default base image is and where it is hosted.

Now, we can get all sorts of `IOException`s (extended by `SocketTimeoutException`, `SSLException`, `ConnectException`, `NoHttpResponseException`, ..), and we can't mutate their fixed message to include the image information. Creating another `IOException` with our custom message and wrapping the cause just for the sake of customizing the message doesn't like a good idea, as it hides the cause such that, e.g., a `catch (SSLException ex)` block no longer works at a higher stack.

TadCordle: Oh I see now.

PR #1953, opened 2019-09-04
Fix footer display

### LINKED ISSUE #1952: Progress bar (AnsiLoggerWithFooter) display issue
Basically, `AnsiLoggerWithFooter` constantly erases the progress bars (footer) and prints them again.

```
[INFO] Containerizing application to Docker daemon as francium25/test...
[INFO] Executing tasks:
[INFO] [============                  ] 40.0% complete
[INFO] > building image to Docker daemon
```

To do so, it prints out "cursor up" escape sequences to move up the cursor, erases everything to the end of display, and prints an updated footer. This process is repeated. The footer is printed with the `Level.LIFECYCLE` level, for which Maven prefixes `[INFO]`.

While working on #1950, I noticed `eventHandler.dispatch(LogEvent.error(...))` ends up printing `[INFO]` instead of `[ERROR]` when the progress bar is enabled. The reason is related to how `AnsiLoggerWithFooter` erases and re-draws the footer.

Right after `eraseFooter()` in `AnsiLoggerWithFooter.log()`, the display becomes like below momentarily:

```
[INFO] Containerizing application to Docker daemon as francium25/test...
[INFO]
```

Note the last line has the lone `[INFO]` with an empty message. Therefore, the current code moves up the cursor up once to print a log message on that last line.

```java
          // If a previous footer was erased, the message needs to go up a line.
          String messagePrefix = didErase ? CURSOR_UP_SEQUENCE : "";
          messageConsumer.accept(messagePrefix + message);
```

Then, when the log level is not `LIFECYCLE` (e.g., `ERROR`), the display will momentarily look like below:

```
[INFO] Containerizing application to Docker daemon as francium25/test...
[INFO]  THIS LINE SHOULD BE PREFIXED WITH "[ERROR]"
[ERROR] 
```

Now the cursor will be on the last line that has the lone `[ERROR]` prefix. After printing the footer, it immediately becomes

```
[INFO] Containerizing application to Docker daemon as francium25/test...
[INFO]  THIS LINE SHOULD BE PREFIXED WITH "[ERROR]"
[INFO] Executing tasks:
[INFO] [============                  ] 40.0% complete
[INFO] > building image to Docker daemon
```

chanseokoh:
Fixes #1952.

Instead of moving the cursor up one line, this moves it up two lines and then down one line to overwrite the previous Maven prefix left by the footer that got erased.

### DIFF
--- a/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooterTest.java
+++ b/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooterTest.java
@@ -132,21 +132,29 @@ public class AnsiLoggerWithFooterTest {
 
     Assert.assertEquals(
         Arrays.asList(
-            "\033[1mfooter\033[0m",
-            "\033[1A\033[0J",
-            "\033[1Amessage",
-            "\033[1mfooter\033[0m",
-            "\033[1A\033[0J",
-            "\033[1Aanother message",
-            "\033[1mfooter\033[0m"),
+            "\033[1mfooter\033[0m", // single-line footer in bold
+
+            // now triggered by logging a message
+            "\033[1A\033[0J", // cursor up and erase to the end
+            "\033[2A", // up two lines
+            "message",
+            "\033[1mfooter\033[0m", // footer
+
+            // by logging another message
+            "\033[1A\033[0J", // cursor up and erase
+            "\033[2A", // up two
+            "another message",
+            "\033[1mfooter\033[0m"), // footer
         messages);
     Assert.assertEquals(
         Arrays.asList(
             Level.LIFECYCLE,
             Level.LIFECYCLE,
             Level.INFO,
+            Level.INFO,
             Level.LIFECYCLE,
             Level.LIFECYCLE,
+            Level.INFO,
             Level.INFO,
             Level.LIFECYCLE),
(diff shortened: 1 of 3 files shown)


jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooterTest.java
```diff
@@ -163,28 +171,40 @@ public void testLog_changingFooter() {
...
-            "\033[1Amessage",
-            "\033[1mfooter\033[0m",
-            "\033[1A\033[0J",
-            "\033[1A\033[1mtwo line\033[0m",
-            "\033[1mfooter\033[0m",
-            "\033[1A\033[1A\033[0J",
-            "\033[1Aanother message",
-            "\033[1mtwo line\033[0m",
-            "\033[1mfooter\033[0m"),
+            "\033[1mfooter\033[0m", // single-line footer in bold
```
TadCordle: Does this test make sure the original problem is fixed?

chanseokoh: In a way, yes for both tests. The old test was testing moving the cursor up one line. Now it tests moving up twice and down once. And tests `accept()` is called twice with right log level (`Level.WARN` in this test).

Issue #1912, opened 2019-08-16
Optimize local base image support: cache compressed layers


TadCordle:
Following up #1906. With the initial iteration of local base image support, none of the base image layers are cached. This means every build requires the costly operation of compressing the layers (if not already compressed) and re-calculating the compressed layer digests. Some ideas we can try to get around this:
- [x] Try caching the compressed layers using their diff-ID, or some other easily accessible layer identifier. (#1957)
- [ ] Cache an entire image to avoid performing `docker save` every time, and retrieve using the image id reported by `docker inspect`. (#2049)

PR #1957, opened 2019-09-05
Cache local base image layers

TadCordle:
Part of #1912. Need to add tests.

I did some manual tests and here is how long each build took, running `jibBuildDocker` on a small test project with the following base images.

`docker://gcr.io/distroless/java`
- First build: 8s
- Second build: 2s

`tar://distroless.tar`
- First build: 6s
- Second build: 1s

`docker://openjdk:8`
- First build: 29s
- Second build: 9s

`tar://openjdk.tar`
- First build: 23s
- Second build: 2s

This is a pretty big improvement; it seems the next bottleneck to overcome is the `docker save` that runs with a `docker://` base image.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -146,42 +151,39 @@ public LocalImage call()
...
+                        new GZIPOutputStream(Files.newOutputStream(compressedFile));
+                    NotifyingOutputStream notifyingOutputStream =
+                        new NotifyingOutputStream(compressorStream, throttledProgressReporter)) {
+                  blob.writeTo(notifyingOutputStream);
+                }
+                blob = Blobs.from(compressedFile);
               }
-              blob = Blobs.from(compressedFile);
+
+              CachedLayer layer = cache.writeTarLayer(diffId, blob);
```
chanseokoh: Remember that this introduces another layer of unnecessary inefficiency that it basically copies an already existing file into the cache directory.

- If the layer is already compressed, hypothetically we could just move it. However, the layer file from `docker save` is (for now) under a temp directory, which may be on a filesystem different than the one for the cache directory, and moving a file may degrade into copying in many cases. When things happen concurrently this may be tricky to handle. Just keep this in mind for later optimization.

- If it is not compressed, I think this is an easy case to avoid copying. Instead of compressing the file into another file first and then let `writeTarLayer` to copy it, the blob can be constructed into the way that `writeTarLayer` will save the content directly while compressing. For example,

```java
Blob blobToBeCompressedWhenWritten =
    Blobs.from(outputStream -> // This will be the "cache file output stream" that writeTarLayer will open for writing.
        try (... = new GZIPOutputStream(outputStream); ...) {
            uncompressedLayerBlob.writeTo(notifyingOutputStream); })
```
Untested though.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -146,42 +151,39 @@ public LocalImage call()
...
-                blob.writeTo(notifyingOutputStream);
+            Optional<CachedLayer> optionalLayer = cache.retrieveTarLayer(diffId);
+            if (optionalLayer.isPresent()) {
+              // Retrieve pre-compressed layer from cache
+              CachedLayer layer = optionalLayer.get();
+              layers.add(new PreparedLayer.Builder(layer).build());
+              v22Manifest.addLayer(layer.getSize(), layer.getDigest());
+            } else {
+              // Compress layers and calculate the digest/size
+              Blob blob = Blobs.from(file);
```
chanseokoh: Now that the loop has become pretty long, I think it warrants to rename `file` to `layerFile`. And the comment "calculate the digest/size" doesn't seem to hold anymore.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
chanseokoh: Actually, wouldn't concurrency matter in any case? Is this handled at a lower level inside `Cache`?

TadCordle: Worth noting for the first point that we do need to write out the pre-compressed layers anyway, because we need to calculate the layer digest.

TadCordle: Also, we should be fine as far as concurrency goes. As with the registry case, we write each local base image layer to a unique temporary directory first before moving them atomically to their final location in the cache.

TadCordle:
I think this should be mostly good for review.

TadCordle:
>Travis is failing. Did you rename the branch at some point?

Hmm, I did accidentally start the skaffold fix on this branch so I needed to revert some stuff, I must have messed up and renamed this branch instead of making a new one. I'll fix it tomorrow.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -145,43 +151,40 @@ public LocalImage call()
...
+          Path layerFile = destination.resolve(layerFiles.get(index));
+          DescriptorDigest diffId = configurationTemplate.getLayerDiffId(index);
 
-          // Compress layers if necessary and calculate the digest/size
-          Blob blob = Blobs.from(file);
           try (ProgressEventDispatcher childDispatcher =
                   childProgressFactories
                       .get(index)
-                      .create("compressing " + file, Files.size(file));
+                      .create("compressing layer " + diffId, Files.size(layerFile));
```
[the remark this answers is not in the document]
TadCordle: I had actually tried the `progressEventDispatcher.newChildProducer().create()` idea before, and it always crashed. I can't remember the reason, I can try again later to see if it was an unrelated problem that got fixed.

Oh yeah, the `progressEventDispatcher.dispatchProgress(1);` I think is leftover from experiments I was doing. Didn't mean for that to make it in.

As for moving things out into its own method, I was going to wait until we work on parallelizing it, but it's probably fine to do it now,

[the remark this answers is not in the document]
TadCordle: Hmm, not creating the child producers in advance seems to be working now, I guess it was a different problem.

PR #1958, opened 2019-09-05
Reorder priority for retrieving auth from Docker config (~/.docker/config.json)

### LINKED ISSUE #1819: Prioritize credhelpers block in docker config.json
UPDATE by @chanseokoh: the background and the symptom are in #1460 (closed as a dup of this). In sum, running `gcloud docker` (deprecated and shouldn't be used) puts short-lived credentials in the `auths` block whose email is `not@val.id` and Jib picks up the credentials in the `auths` block before `credHelpers`.

---

This might hold true for all registries.

But at least for gcr.io this is the preferred option. Since old expired credentials generated by `gcloud docker` can result in unexpected auth failures.

chanseokoh:
Fixes #1819.

This is actually more than that. The priority is now `credHelper` --> `credsStore` --> `auths`. According to https://github.com/moby/moby/pull/25851 (and assuming it wasn't reverted), looks like at least `credHelper` should take precedence over `credsStore`:

> This change adds a "credentialHelpers" member to the config file where users can specify credential helpers on a per-registry basis. This does not over-load the "credsStore" field, which now acts as the 'default' credential store in the event that one is not explicitly specified for the registry in question.

Also, I've lifted the restriction that there should be a matching `auths` registry entry for `credsStore` to be returned. The practice to define an empty `auths` entry is documented in [here](https://github.com/GoogleCloudPlatform/docker-credential-gcr#using-the-credsstore-for-docker-clients-since-v1110), but my guess is that this is attributed to a quick workaround to prevent a failure in certain circumstances: https://github.com/awslabs/amazon-ecr-credential-helper/issues/9#issuecomment-241184224 Nonetheless, it seems to me the workaround is no longer relevant anymore with [this change](https://github.com/moby/moby/pull/26354) in recent Docker versions where Docker queries the credential helper to populate a list of supported registries instead of using the registries in the `auths` section. This is also explained in https://github.com/awslabs/amazon-ecr-credential-helper/issues/9#issuecomment-310515232

Will update CHANGELOG soon.

### DIFF
--- a/jib-core/CHANGELOG.md
+++ b/jib-core/CHANGELOG.md
@@ -11,7 +11,8 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - `TarImage` is constructed using `TarImage.at(...).named(...)` instead of `TarImage.named(...).saveTo(...)` ([#1918](https://github.com/GoogleContainerTools/jib/issues/1918))
-- For retrieving credentials from Docker config (`~/.docker/config.json`), `credHelpers` now takes precedence over `credsStore`, followed by `auths`. Moreover, the legacy `credsStore` no longer requires defining empty registry entries in `auths` to be used. This means that if `credsStore` is defined, `auths` will be completely ignored. ([#1958](https://github.com/GoogleContainerTools/jib/pull/1958))
+- For retrieving credentials from Docker config (`~/.docker/config.json`), `credHelpers` now takes precedence over `credsStore`, followed by `auths`. ([#1958](https://github.com/GoogleContainerTools/jib/pull/1958))
+- The legacy `credsStore` no longer requires defining empty registry entries in `auths` to be used. This now means that if `credsStore` is defined, `auths` will be completely ignored. ([#1958](https://github.com/GoogleContainerTools/jib/pull/1958))
 
 ### Fixed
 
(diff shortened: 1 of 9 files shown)


chanseokoh:
I don't have high hopes. It's a breaking change. Although I think we can justify this reordering and people will be able to make it right and better: `auths` is discouraged as you put passwords as plain text, and `credsStore` is old enough to ditch and no real reason to use.

jib-core/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - `TarImage` is constructed using `TarImage.at(...).named(...)` instead of `TarImage.named(...).saveTo(...)` ([#1918](https://github.com/GoogleContainerTools/jib/issues/1918))
+- For retrieving credentials from Docker config (`~/.docker/config.json`), `credHelpers` now takes precedence over `credsStore`, followed by `auths`. Moreover, the legacy `credsStore` no longer requires defining empty registry entries in `auths` to be used. This means that if `credsStore` is defined, `auths` will be completely ignored. ([#1958](https://github.com/GoogleContainerTools/jib/pull/1958))
```
TadCordle: I feel like our changelog entries are getting longer and longer, haha. Maybe we can split this into two (`credStore` change different from `credHelpers` precedence)

Issue #1522, opened 2019-02-28
Incorrect progress report usage in BlobPuller


chanseokoh:
From https://github.com/GoogleContainerTools/jib/issues/1512#issuecomment-468078006.

```java
  public Void handleResponse(Response response) ... {
    // Notify the progress listener that the bytes we will read is the "Content-Length" value.
    blobSizeConsumer.accept(response.getContentLength());

    try (OutputStream outputStream =
        new ListenableCountingOutputStream(
            destinationOutputStream,
            blobProgressListener::handleByteCount,
            blobProgressListener.getDelayBetweenCallbacks())) {
      BlobDescriptor receivedBlobDescriptor = response.getBody().writeTo(outputStream);
      ...
    }
```
Now, in case of `Content-Encoding: gzip` (or something similar), `HttpResponse::getContent()` creates and returns `GZIPInputStream`, which `response.getBody()` calls to create an `InputStreamBlob`.

However, `GZIPInputStream.read(buffer)` returns the counts of the *uncompressed* bytes copied on the `buffer` (understandably), rather than returning the actual network bytes it read. Therefore, we end up reporting more bytes written to `outputStream`.

Unfortunately, it is tricky to make this right, since we can't know the uncompressed bytes we will read in advance.

Not sure if there can be other places with the same issue.

PR #1960, opened 2019-09-06
Add progress events to LoadDockerStep

### LINKED ISSUE #1959: Add progress indicator to LoadDockerStep
We can probably track progress in `LoadDockerStep` similar to how we track progress in `SaveDockerStep` (as in https://github.com/GoogleContainerTools/jib/pull/1951).

TadCordle:
Fixes #1959.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
@@ -23,6 +23,7 @@ import com.google.cloud.tools.jib.configuration.BuildConfiguration;
 import com.google.cloud.tools.jib.docker.DockerClient;
 import com.google.cloud.tools.jib.docker.ImageTarball;
 import com.google.cloud.tools.jib.event.EventHandlers;
+import com.google.cloud.tools.jib.event.progress.ThrottledAccumulatingConsumer;
 import com.google.cloud.tools.jib.image.Image;
 import java.io.IOException;
 import java.util.concurrent.Callable;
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -120,16 +119,19 @@ public DockerClient(Path dockerExecutable, Map<String, String> dockerEnvironment
...
    */
-  public String load(ImageTarball imageTarball) throws InterruptedException, IOException {
+  public String load(ImageTarball imageTarball, Consumer<Long> writtenByteCountListener)
+      throws InterruptedException, IOException {
     // Runs 'docker load'.
     Process dockerProcess = docker("load");
 
-    try (OutputStream stdin = dockerProcess.getOutputStream()) {
+    try (NotifyingOutputStream stdin =
+        new NotifyingOutputStream(dockerProcess.getOutputStream(), writtenByteCountListener)) {
```
chanseokoh: I think we can merge the two outermost try blocks?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
```diff
@@ -50,27 +52,34 @@
...
+      eventHandlers.dispatch(LogEvent.progress("Loading to Docker daemon..."));
 
-    try (ProgressEventDispatcher ignored =
-        progressEventDispatcherFactory.create("loading to Docker daemon", 1)) {
       ImageReference targetImageReference =
           buildConfiguration.getTargetImageConfiguration().getImage();
+      ImageTarball imageTarball = new ImageTarball(builtImage, targetImageReference);
+      try (ProgressEventDispatcher progressEventDispatcher =
+              progressEventDispatcherFactory.create(
+                  "loading to Docker daemon", imageTarball.getSize());
```
chanseokoh: The only problem is that we will go over the total progress allocation, because it will also write two more files (`config.json` and `manifest.json`), whose size we cannot know ahead, unfortunately. We won't crash, as we lifted the assertion that the progress completion cannot go over the given allocation at some point due to some corner case (#1522), but still, the computation here will always be incorrect. BTW, `BlobPuller` uses `blobSizeListener` coming from `ThrottledProgressEventDispatcherWrapper::setProgressTarget()` to lazily set the total progress, so we can do the same, but this require more changes to `ImageTarball`, so I can't say I like it.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
```diff
@@ -83,4 +82,17 @@ public void writeTo(OutputStream out) throws IOException {
 
     tarStreamBuilder.writeAsTarArchiveTo(out);
   }
+
+  /**
+   * Returns the total size of the image's layers in bytes.
+   *
+   * @return the total size of the image's layers in bytes
+   */
+  public long getSize() {
```
chanseokoh: `getTotalLayerSize()` to make it clear this is not the tarball size?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LoadDockerStep.java
TadCordle: I don't really mind that much that the calculation is incorrect as long as it's close. The layers are almost always going to be much bigger than the manifest and configuration, so the progress being reported will at least be close to accurate. It certainly doesn't matter with the plugins, since the progress bar moves too fast to be able to even tell that there's any inaccuracy, but I suppose it might matter a little bit for jib-core users if they need totally accurate progress reporting. But even then, the progress event API still isn't finalized.

chanseokoh: Agreed. Let's just leave a code comment that the progress will always go over the limit.

PR #1961, opened 2019-09-06
Update Blobs#compress javadoc

### LINKED ISSUE #1944: Blob javadoc says it does not close the output stream, but several of its implementations do
For example, `FileBlob#writeTo` looks like this:

```
try (InputStream fileIn = new BufferedInputStream(Files.newInputStream(file))) {
  return Digests.computeDigest(fileIn, outputStream);
}
```

We should either remove the part of the javadoc that says `"Does not close the {@code outputStream}."`, or we should make sure all implementations follow this rule.

TadCordle:
Fixes #1944. 

The issue assumed several blobs close their output streams, contrary to the javadoc in `Blob`, but this wasn't true (several implementations like `FileBlob` close the input stream, not the output stream). `Blobs#compress` does contradict the javadoc, but since this is only used in tests, I just decided to update the javadoc for `compress` and add `@VisibleForTesting` instead of changing the original `Blob` javadoc.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
@@ -18,6 +18,7 @@ package com.google.cloud.tools.jib.blob;
 
 import com.google.cloud.tools.jib.hash.WritableContents;
 import com.google.cloud.tools.jib.json.JsonTemplate;
+import com.google.common.annotations.VisibleForTesting;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blobs.java
```diff
@@ -79,11 +80,13 @@ public static String writeToString(Blob blob) throws IOException {
   }
 
   /**
-   * Gets a {@link Blob} that is {@code blob} compressed.
+   * Gets a {@link Blob} that is {@code blob} compressed. Note that the output stream is closed when
+   * the blob is written.
```
chanseokoh: Hmm... I don't quite like having an exception. It makes sense to not close an externally-owned stream. For example, some class may establish a network connection and write multiple blobs (in addition to other things) over the wire during the course. They may find it surprising if writing some opaque blob abruptly shuts down their output channel and renders the connection broken.

I think we can leave the output stream open by not using the try-resource block, simple as that. The owner of the output stream should eventually close it anyway.

Or, we can just move this function back to the test code and forget about any concern. The code is used only in tests, as you noticed.

TadCordle: That last point was mostly my intent, hence the added `@VisibleForTesting`. Do you think we should just duplicate the method in the two test classes that use it like we had before?

chanseokoh: Yeah, I think that's fine. We can be sure the method is not used in prod.

PR #1965, opened 2019-09-09
Turn off progress bar when capturing network trace

chanseokoh:


docs/faq.md
```diff
@@ -533,12 +533,13 @@ com.google.api.client.http.level=CONFIG
...
 ```sh
-mvn -Djava.util.logging.config.file=path/to/log.properties -DjibSerialize=true ...
+mvn -Djava.util.logging.config.file=path/to/log.properties -DjibSerialize=true -Djib.console=plain ...
 ```
 or
 ```sh
-gradle -Djava.util.logging.config.file=path/to/log.properties -DjibSerialize=true ...
+gradle -Djava.util.logging.config.file=path/to/log.properties -DjibSerialize=true -Djib.console=plain ...
 ```
+Note there is no dot(.) in `jibSerialize`.
```
TadCordle: I feel like I've brought this up before and can't remember where, but I wonder if we should file an issue about making these system property names more consistent. Maybe for 2.0?

loosebazooka: yeah we can make the change now and deprecate jibSerialize (marked for removal in 2.0)

loosebazooka: https://github.com/GoogleContainerTools/jib/issues/1967

PR #1971, opened 2019-09-10
Fix progress bar logger for Gradle

### LINKED ISSUE #1963: Log output gets overwritten
Jib output used to say:
```
Containerizing application to Docker daemon as test...

Container entrypoint set to [java, -cp, /app/resources:/app/classes:/app/libs/*, Hello]

Built image to Docker daemon as test
Executing tasks:
```

but now the first two messages get overwritten shortly after printing them, and the final output becomes:

```




Built image to Docker daemon as test
Executing tasks:
```

chanseokoh:
Fixes #1963.

#1963 was a regression introduced in #1953 that fixed a logging issue in Maven. But it broke Gradle.

Unfortunately, I can't think of a better and safer way than to maintain two separate paths for Maven and Gradle. I restored the original trick to fix the Gradle issue.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooter.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooter.java
@@ -40,6 +40,9 @@ class AnsiLoggerWithFooter implements ConsoleLogger {
   /** ANSI escape sequence template for moving the cursor up multiple lines. */
   private static final String CURSOR_UP_SEQUENCE_TEMPLATE = "\033[%dA";
 
+  /** ANSI escape sequence template for moving the cursor up multiple lines. */
+  private static final String CURSOR_UP_SEQUENCE = String.format(CURSOR_UP_SEQUENCE_TEMPLATE, 1);
+
   /** ANSI escape sequence for erasing to end of display. */
   private static final String ERASE_DISPLAY_BELOW = "\033[0J";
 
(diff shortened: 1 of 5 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooter.java
```diff
@@ -103,13 +116,16 @@ public void log(Level logLevel, String message) {
 
     singleThreadedExecutor.execute(
         () -> {
-          if (eraseFooter()) {
-            // If a previous footer was erased, the message needs to go up a line. However, we go up
-            // two lines (and then down one line by calling "accept()" once) to correct an issue in
-            // Maven: https://github.com/GoogleContainerTools/jib/issues/1952
+          boolean didErase = eraseFooter();
```
TadCordle: L119-126 and L152-159 are the same, I wonder if we can pull those lines into `eraseFooter()`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/ConsoleLoggerBuilder.java
```diff
@@ -39,11 +39,15 @@
    *
    * @param singleThreadedExecutor a {@link SingleThreadedExecutor} to ensure that all messages are
    *     logged in a sequential, deterministic order
+   * @param mavenFix fixes logging issue in Maven
+   *     (https://github.com/GoogleContainerTools/jib/issues/1952)
    * @return a new {@link ConsoleLoggerBuilder}
    */
-  public static ConsoleLoggerBuilder rich(SingleThreadedExecutor singleThreadedExecutor) {
+  public static ConsoleLoggerBuilder rich(
+      SingleThreadedExecutor singleThreadedExecutor, boolean mavenFix) {
```
TadCordle: I don't know about this parameter name, maybe we should just have two different `richGradle` and `richMaven` methods instead of passing in a `boolean`. Although I'm not sure if I really like that more. It at least sounds less hacky (then again I guess this is a bit of a hack, and maybe we shouldn't try to hide that).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooter.java
chanseokoh: Yeah, the code is duplicated, but I find the logic of the duped code is a bit weird to shove into `eraseFooter()` for the sake of removing redundancy. The caller is expected to follow up with doing the correct thing after calling `eraseFooter()`(i.e., must print the next message with the returned prefix), so I don't quite like it from the API perspective. I think this amount of dup is not worse than that.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/ConsoleLoggerBuilder.java
chanseokoh: Yeah, agreed. Kinda expected this reaction. I wanted to avoid putting a specific builder name, since theoretically we may have more builders in the future. I think I just came up with a much better name: `enableTwoCursorUpJump`. The rich console is all about the hacky cursor trick, so this kind of console-controlling flag names for the constructor method sound pretty natural to me.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/AnsiLoggerWithFooter.java
```diff
@@ -103,13 +116,16 @@ public void log(Level logLevel, String message) {
...
+
+          // If a previous footer was erased, the message needs to go up a line. But only do so when
+          // not doing the two-cursor-up fix.
+          String prefix = didErase && !enableTwoCursorUpJump ? CURSOR_UP_SEQUENCE : "";
```
TadCordle: Maybe personal preference, but I think the shorter code here is harder to follow than just separating each condition fully.
```
if (didErase) {
  if (enableTwoCursorUpJump) {
    messageConsumer.accept(String.format(CURSOR_UP_SEQUENCE_TEMPLATE, 2));
    messageConsumer.accept(message);
  } else {
    messageConsumer.accept(CURSOR_UP_SEQUENCE + message);
  }
} else {
  messageConsumer.accept(message);
}
```

Unfortunately that leads to even more duplication.

chanseokoh: Yeah, that's true. Expanded the code.

PR #1973, opened 2019-09-10
Update 1.6.0 docs for jib.serialize property name

loosebazooka:

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -519,7 +519,7 @@ See more at [Using Google Container Registry (GCR) with Minikube](https://ryanes
 
 ### How can I examine network traffic?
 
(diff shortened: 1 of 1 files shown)


jib-core/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- To disable parallel execution, the property `jib.serialize` should be used instead of `jibSerialize`. ([#1968](https://github.com/GoogleContainerTools/jib/issues/1968))
```
TadCordle: FYI I think we can leave the changelog updates in the original PR, since those are explicit about what's still unreleased, just not the documentation.

loosebazooka: Haha, yeah you're right.

PR #1974, opened 2019-09-10
Fix skaffold init for multimodule maven projects

TadCordle:
Quick idea to fix https://github.com/GoogleContainerTools/skaffold/issues/2844.

Instead of running `jib:_skaffold-init` separately on every module including the root project, we use an aggregator to run the goal once across all modules. This lets us more easily detect the root project, which helps prevent `skaffold init` from setting the `module` field when we don't want it to.

The problem with this approach is that using an aggregator makes getting the `<to><image>` configuration on sub-projects harder; in this case we can only parse it from the configuration block and ignore system properties.

wdyt?

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SkaffoldInitMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SkaffoldInitMojo.java
@@ -19,35 +19,75 @@ package com.google.cloud.tools.jib.maven.skaffold;
 import com.google.cloud.tools.jib.maven.JibPluginConfiguration;
 import com.google.cloud.tools.jib.plugins.common.SkaffoldInitOutput;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import java.io.IOException;
+import java.util.List;
+import javax.annotation.Nullable;
+import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.apache.maven.project.MavenProject;
+import org.codehaus.plexus.util.xml.Xpp3Dom;
 
 /**
  * Prints out to.image configuration and project name, used for Jib project detection in Skaffold.
  *
  * <p>Expected use: {@code ./mvnw jib:_skaffold-init -q}
  */
-@Mojo(name = SkaffoldInitMojo.GOAL_NAME, requiresDependencyCollection = ResolutionScope.NONE)
+@Mojo(
+    name = SkaffoldInitMojo.GOAL_NAME,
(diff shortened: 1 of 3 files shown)


loosebazooka:
> The problem with this approach is that using an aggregator makes getting the <to><image> configuration on sub-projects harder; in this case we can only parse it from the configuration block and ignore system properties.

Does the person running skaffold init even have a chance to set the image name via property?

TadCordle:
Maybe not system properties, but pom properties?

chanseokoh:
Ah, so it will be sort of broken when the user uses a pom property? I think using pom properties is common, so that's concerning. Hmm...

TadCordle:
Maybe we can parse the properties block the same way.

TadCordle:
Oh wait we can just use `project.getProperties()`, probably.

chanseokoh:
It's still incomplete in that when the person does `<image>example.com/${image.name}</image>`. Sad!

TadCordle:
Ah right. Maybe we can do the replacement ourselves?

chanseokoh:
That sounds like a tough thing to do. Can there be other alternatives? For example, I noticed `<relativePath/>` is empty in the Spring Boot project. Not exactly sure what the Maven API return for this though. From the [doc](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Example_1), it is suggested that the value could by default `../pom.xml` if not specified, so if `<relativePath/>` returns an empty string, we could figure that the parent POM doesn't exist locally. And I think a Spring Boot project is a somewhat special case to use a parent POM this way, making the project setup unable to have a custom parent.

```xml
<parent>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-parent</artifactId>
	<version>2.1.8.RELEASE</version>
	<relativePath/> <!-- lookup parent from repository -->
</parent>
```

TadCordle:
I think that can work (barring anymore unusual edge cases), using `getProject().getParent().getFile()`. It at least works with the 4 manual tests cases I've been using (spring initalizr project, a parent + 2 modules, a parent + 1 module, and a single pom project).

chanseokoh:
What does it return for each four cases? And I guess `getProject().getParent().getFile()` is basically what `<relativePath>` and there's no other API for that?

TadCordle:
single pom project: `project.getParent()` is null
Initializr: `project.getParent().getFile()` is null
2 modules: each module returns the path to the parent pom
1 module: same thing

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SkaffoldInitMojo.java
```diff
@@ -37,12 +38,17 @@
   @Override
   public void execute() throws MojoExecutionException {
     checkJibVersion();
+    MavenProject project = getProject();
+    // Ignore parent projects
+    if (project.getModules().size() > 0) {
```
chanseokoh: So this was another bug, right?

TadCordle: I guess you could call it that. It was something I was aware of but didn't think was necessary to change until now.

PR #1975, opened 2019-09-10
Fix progress allocation when given non-positive size

### LINKED ISSUE #1972: Progress allocation created 0 should better be 1 allocation unit
~~I thought it was safe to pass 0 allocation units, but in fact, it isn't. It will crash when 0 or negative. The `allocatinoUnits < 0` is pointless, so not sure how it got there.~~

```java
  private Allocation(String description, long allocationUnits, @Nullable Allocation parent) {
    ...
    this.allocationUnits = allocationUnits < 0 ? 0 : allocationUnits;
    ....
    this.fractionOfRoot = (parent == null ? 1.0 : parent.fractionOfRoot) / allocationUnits;
  }
```

~~Ideally we should make it safe to accept 0.~~

chanseokoh:
Fixes #1972.

I think setting the minimum unit to 1 still makes sense; the progress dispatcher in this case would represent the task as a work to be completed. Also, dividing by 0 makes `fractionOfRoot` infinity, so it's better to avoid it. 

And it should definitely divide the parent fraction by `this.allocation` instead of the argument `allocation`. This has to be fixed.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/Allocation.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/event/progress/Allocation.java
@@ -65,10 +65,10 @@ public class Allocation {
 
   private Allocation(String description, long allocationUnits, @Nullable Allocation parent) {
     this.description = description;
-    this.allocationUnits = allocationUnits < 0 ? 0 : allocationUnits;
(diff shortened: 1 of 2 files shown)


TadCordle:
LGTM, but it looks like it was based off the wrong branch, so there are a bunch of extra changes in here.

chanseokoh:
Oops, should be good now.

PR #1981, opened 2019-09-12
Move docker client configuration to all tasks and goals

### LINKED ISSUE #1932: Add global docker client configuration
The docker client environment and executable path are configurable on jib's `jibDockerBuild` task and `jib:dockerBuild` goal, but not configurable on any other tasks. With local base image support, the docker parameters should be moved to `JibExtension`/`JibPluginConfiguration` so they can be configured on any task/goal.

TadCordle:
Fixes #1932.

This PR moves the `jib.dockerClient` configuration from `BuildDockerTask` and `BuildDockerMojo` to `JibExtension` and `JibPluginConfiguration` to be accessible from all tasks and goals, allowing users to configure the docker client for docker daemon base images.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -91,10 +91,10 @@ public class BuildDockerTask extends DefaultTask implements JibTask {
       throws IOException, BuildStepsExecutionException, CacheDirectoryCreationException,
           MainClassInferenceException {
     Preconditions.checkNotNull(jibExtension);
-    Path dockerExecutable = jibExtension.getDockerClient().getExecutablePath();
-    Map<String, String> dockerEnvironment = jibExtension.getDockerClient().getEnvironment();
 
     // Check deprecated parameters
+    Path dockerExecutable = jibExtension.getDockerClient().getExecutablePath();
+    Map<String, String> dockerEnvironment = jibExtension.getDockerClient().getEnvironment();
     if (getDockerClient().getExecutable() != null) {
       jibExtension.getDockerClient().setExecutable(getDockerClient().getExecutable());
       getProject()
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -34,11 +34,9 @@ import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.nio.file.Path;
 import javax.annotation.Nullable;
-import org.gradle.api.Action;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.GradleException;
 import org.gradle.api.tasks.Nested;
-import org.gradle.api.tasks.Optional;
 import org.gradle.api.tasks.TaskAction;
 import org.gradle.api.tasks.options.Option;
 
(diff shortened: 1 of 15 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -87,7 +90,34 @@ public void dockerClient(Action<? super DockerClientParameters> action) {
...
+    if (getDockerClient().getExecutable() != null) {
+      if (dockerExecutable != null) {
+        throw new GradleException(
+            "Cannot configure 'jibDockerBuild.dockerClient.executable' and 'jib.dockerClient.executable' simultaneously");
+      }
+      getProject()
+          .getLogger()
+          .warn(
+              "'jibDockerBuild.dockerClient.executable' is deprecated; use 'jib.dockerClient.executable' instead.");
+      dockerExecutable = getDockerClient().getExecutablePath();
```
chanseokoh: Why overwrite when `dockerExecutable` is already configured?

Maybe we can simplify the code a bit:
```java
if (getDockerClient().getExecutable() != null || !getDockerClient().getEnvironment().isEmpty()) {
  ...warn(...);
}
if ((dockerExecutable != null && getDockerClient().getExecutable() != null) || (!dockerEnvironment.isEmpty() && !getDockerClient().getEnvironment().isEmpty()) {
    throw new GradleException("Cannot configure 'jib.dockerClient' and 'jibDockerBuild.dockerClient' ...");
}

File: jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java [hunk 2]
```diff
@@ -48,6 +49,9 @@ public String getExecutable() {
   @Internal
   @Nullable
   Path getExecutablePath() {
+    if (System.getProperty(PropertyNames.DOCKER_CLIENT_EXECUTABLE) != null) {
```
chanseokoh: I think we should double-check if this is needed. I feel like `getExecutable()` *might* set `executable` when the property is defined.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -63,7 +63,7 @@ public static JibBuildRunner createJibBuildRunnerForDockerDaemonImage(
       InferredAuthProvider inferredAuthProvider,
       ProjectProperties projectProperties,
       @Nullable Path dockerExecutable,
-      @Nullable Map<String, String> dockerEnvironment,
+      Map<String, String> dockerEnvironment,
```
chanseokoh: Since `rawConfiguration` now holds `dockerExecutable` and `dockerEnvironment`, we don't need to pass them? I think it's good for overall consistency.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/DockerClientParameters.java
TadCordle: It doesn't look like it does.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
TadCordle: Ah I jumped on this too soon. The reason for the overwrite is because `PluginConfigurationProcessor` needs to use either `getDockerClient().getExecutable()`, or `jibExtension.getDockerClient().getExecutable()`. If we don't do the overwrite, then this will break users since their original configuration won't be passed forward.

PR #1987, opened 2019-09-13
Fix release by ignoring SNAPSHOT deps on core/plugins-common

loosebazooka:
the SNAPSHOT dependencies are a consequence of the monolith, the
release plugin just needs to ingore those projects, ideally the
plugin wouldn't care that compileOnly dependencies are SNAPSHOTS
but whatever, they might have a reason for that.

oh also this fixes the tagging on the maven plugin

This should fix the problem for now.... also filed: https://github.com/researchgate/gradle-release/issues/310


loosebazooka:
both maven/gradle release scripts seem to be okay to go.

PR #1998, opened 2019-09-16
Fix base image java version check

### LINKED ISSUE #1995: Custom base image ignored with Jib Gradle 1.6.0
**Environment**:

- *Jib version:* 1.6.0
- *Build tool:* Gradle 5.6.1
- *OS:* Fedora 30

**Description of the issue**:

I have a Gradle app that is configured to use Java 12:

```
sourceCompatibility = 12
targetCompatibility = 12
```

It is packaged with Jib. Since Jib includes support for Java 11 only by default, I specify a Java 12 base image as follows. Unfortunately, when I run `gradle jibDockerBuild`, the task fails with the following message:

> Your project is using Java 12 but the base image is for Java 11, perhaps you should configure a Java 12-compatible base image using the 'jib.from.image' parameter, or set targetCompatibility = 11 or below in your build configuration

**Expected behavior**:

Gradle `jibDockerBuild` task completes with success.

**Steps to reproduce**:

  1. Create a basic Java 12 app using Gradle.
  2. Include Gradle Jib definition as follows.
  3. Run `gradle jibDockerBuild`.

```groovy
jib {
    container {
        jvmFlags = [
                '-server',
                '-Xms6g',
                '-Xmx6g',
                '-XX:+UnlockExperimentalVMOptions',
                '-XX:+UseShenandoahGC',
                '-Dauto-update.temp.dir=/root',
                '-Dmanagement.endpoints.jmx.exposure.exclude=*',
                '-Dmanagement.endpoints.web.exposure.include=health',
                '-Dspring.application.admin.enabled=false',
                '-Dspring.jmx.enabled=false'
        ]
        mainClass 'com.acme.Application'
    }
    from {
        image 'adoptopenjdk/openjdk12:slim'
    }
    to {
        auth {
            image = 'registry.hub.docker.com/acme/api:latest'
            username = 'acme'
            password = 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
        }
        tags = ['latest']
    }
}
```

**Additional Information**: Downgrading to Jib Gradle version 1.5.1 solves the issue. It seems a change was introduced with version 1.6.0. I tried adding the prefix 'registry://' to the base image name but the task still fails.

TadCordle:
Fixes #1995

### DIFF
--- a/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
+++ b/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
@@ -842,6 +842,18 @@ public class PluginConfigurationProcessorTest {
     }
   }
 
+  // https://github.com/GoogleContainerTools/jib/issues/1995
+  @Test
+  public void testGetJavaContainerBuilderWithBaseImage_java12BaseImage()
+      throws InvalidImageReferenceException, IOException, IncompatibleBaseImageJavaVersionException,
+          CacheDirectoryCreationException {
+    Mockito.when(projectProperties.getMajorJavaVersion()).thenReturn(12);
+    Mockito.when(rawConfiguration.getFromImage()).thenReturn(Optional.of("regis.try/java12image"));
+    ImageConfiguration imageConfiguration = getCommonImageConfiguration();
+    Assert.assertEquals("regis.try", imageConfiguration.getImageRegistry());
+    Assert.assertEquals("java12image", imageConfiguration.getImageRepository());
+  }
+
   @Test
   public void testGetValidVolumesList() throws InvalidContainerVolumeException {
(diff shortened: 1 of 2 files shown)


jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -842,6 +842,18 @@ public void testGetJavaContainerBuilderWithBaseImage_incompatibleJava11JettyBase
     }
   }
 
+  // https://github.com/GoogleContainerTools/jib/issues/1995
+  @Test
+  public void testGetJavaContainerBuilderWithBaseImage_java12BaseImage()
```
loosebazooka: Should we add a test that fails if java version is `>11` and no image is set?

TadCordle: Sure

PR #2004, opened 2019-09-16
Fix base image reproducibility warning

### LINKED ISSUE #1969: Base image reproducibility warning does not match with the check
The message talks about not using a digest
```
Base image example.com/repo does not use a specific image digest - build may not be reproducible
```
but the code only checks whether the base is using `:latest` (implicit or explicit).
```java
          if (baseImageConfiguration.getImage().usesDefaultTag()
              && !baseImageConfiguration.getImage().isScratch()) {
            eventHandlers.dispatch(
                LogEvent.warn(
                    "Base image '"
                        + baseImageConfiguration.getImage()
                        + "' does not use a specific image digest - build may not be reproducible"));
```
Perhaps we should warn whenever a digest is not used.

Note currently the warning doesn't get printed for distroless when no base image is given, because we auto-select the tag (either `:8` or `:11`).

TadCordle:
Fixes #1969.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildConfiguration.java
@@ -245,7 +245,7 @@ public class BuildConfiguration {
       switch (missingFields.size()) {
         case 0: // No errors
           Preconditions.checkNotNull(baseImageConfiguration);
-          if (baseImageConfiguration.getImage().usesDefaultTag()
+          if (!baseImageConfiguration.getImage().isTagDigest()
               && !baseImageConfiguration.getImage().isScratch()) {
             eventHandlers.dispatch(
                 LogEvent.warn(


jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildConfigurationTest.java
```diff
@@ -246,4 +249,38 @@ public void testBuilder_missingValues() throws IOException {
...
+        .setBaseImageConfiguration(
+            ImageConfiguration.builder(
+                    ImageReference.parse(
+                        "image@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"))
+                .build())
+        .build();
+    Mockito.verify(mockEventHandlers, Mockito.never())
+        .dispatch(
+            LogEvent.warn(
+                "Base image 'image:tag' does not use a specific image digest - build may not be reproducible"));
```
chanseokoh: This won't match anyway because it explicitly states `image:tag`. Is it possible to do `LogEvent.warn(anyString())`?

PR #2005, opened 2019-09-16
Mark maven goals as threadSafe

### LINKED ISSUE #1904: jib-maven-plugin should be marked as `threadsafe`
Maven complains when using the `jib-maven-plugin` with multiple threads (`mvn -T2`).
```
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that have goals not marked   *
[WARNING] * as @threadSafe to support parallel building.                  *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
```

Our Maven goals should be marked as `threadsafe=true`.

TadCordle:
Fixes #1904.

The issue makes a good point about Jib having the potential to be "smarter" with its concurrency, but maybe we should open a new issue about that.

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -42,7 +42,8 @@ import org.apache.maven.plugins.annotations.ResolutionScope;
 /** Builds a container image and exports to the default Docker daemon. */
 @Mojo(
     name = BuildDockerMojo.GOAL_NAME,
-    requiresDependencyResolution = ResolutionScope.RUNTIME_PLUS_SYSTEM)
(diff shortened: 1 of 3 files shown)


loosebazooka:
Yeah there's probably a few things we could do if we did maven/gradle specific things.

briandealwis:
I don't think `@threadSafe` buys us anything until we're smarter, and in fact can make us look worse.

TadCordle:
I don't feel strongly about it. It *is* thread-safe, so it's nice to remove the warning even if it may be sub-optimal (which would only be for the first run, since caching takes over afterwards). But I can close this if you think it's important to make it optimal before removing the warning.

briandealwis:
Good point: my bad results were for a pathological case; it's unlikely that anybody would run 50 simultaneous builds.

PR #2009, opened 2019-09-17
Minor cleanups

chanseokoh:


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -245,10 +243,7 @@ private Authorization authenticate(@Nullable Credential credential, String scope
                 scope,
                 registryEndpointRequestProperties.getSourceImageName(),
                 "pull");
-        Authorization auth = authenticate(credential, scopes);
-        if (auth != null) {
-          return auth;
-        }
+        return authenticate(credential, scopes);
```
loosebazooka: Seems like this may not be right? Authenticate needs to fail-over to the next one if this fails?

loosebazooka: oh is it not nullable?

chanseokoh: Yeah, I found it cannot be null.

PR #2012, opened 2019-09-18
Check exit code for 'docker inspect'

### LINKED ISSUE #2011: Cryptic error when Docker base image doesn't exist
The command to get the image size (`docker inspect -f {{.Size}}`) fails if the image doesn't exist, hence the cryptic error.

```
Caused by: java.lang.NumberFormatException: For input string: ""
    at java.lang.NumberFormatException.forInputString (NumberFormatException.java:65)
    at java.lang.Long.parseLong (Long.java:702)
    at java.lang.Long.parseLong (Long.java:817)
    at com.google.cloud.tools.jib.docker.DockerClient.sizeOf (DockerClient.java:235)
    at com.google.cloud.tools.jib.builder.steps.SaveDockerStep.call (SaveDockerStep.java:57)
    at com.google.cloud.tools.jib.builder.steps.SaveDockerStep.call (SaveDockerStep.java:32)
    at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly (TrustedListenableFutureTask.java:125)
    at com.google.common.util.concurrent.InterruptibleTask.run (InterruptibleTask.java:69)
    at com.google.common.util.concurrent.TrustedListenableFutureTask.run (TrustedListenableFutureTask.java:78)
    at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1128)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:628)
    at java.lang.Thread.run (Thread.java:834)
```

chanseokoh:
Fixes #2011.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
@@ -136,7 +136,8 @@ public class DockerClient {
       imageTarball.writeTo(stdin);
 
     } catch (IOException ex) {
-      // Tries to read from stderr.
+      // Tries to read from stderr. Not using getStderrOutput(), as we want to show the error
+      // message from the tarball I/O write failure when reading from stderr fails.
       String error;
       try (InputStreamReader stderr =
           new InputStreamReader(dockerProcess.getErrorStream(), StandardCharsets.UTF_8)) {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
@@ -27,6 +27,7 @@ import java.io.BufferedOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -140,11 +141,10 @@ public String load(ImageTarball imageTarball, Consumer<Long> writtenByteCountLis
       try (InputStreamReader stderr =
           new InputStreamReader(dockerProcess.getErrorStream(), StandardCharsets.UTF_8)) {
         error = CharStreams.toString(stderr);
-
       } catch (IOException ignored) {
-        // This ignores exceptions from reading stderr and throws the original exception from
+        // This ignores exceptions from reading stderr and uses the original exception from
         // writing to stdin.
-        throw ex;
+        error = ex.getMessage();
```
briandealwis: Couldn't this code use the same `getStderrOutput(dockerProcess))` as below?

chanseokoh: I actually tried that and contemplated the code. Went back and forth. `getStderrOutput()` returns the canned `unknown` message when an IO error occurs while reading the stderr message. However, the original code wants that if such an IO error occurs, it shows the error message of the original IO exception that occurred while writing into the process, which reveals more useful information. I could change `getStderrOutput()` to accept a "default" error message to print if it fails to read stderr for any reason, but since here is the only usage for that, so I decided to leave it read stderr manually here.

briandealwis: That makes sense; would be worth a comment to that effect here

PR #2018, opened 2019-09-23
Fix local base image temp directory cleanup

### LINKED ISSUE #2016: FileOperations.deleteRecursiveOnExit() doesn't seem to work
I always get the following exception (triggered when specifying a Docker base image, as Jib tries to delete a temp directory from `docker save`) with Maven. Gradle doesn't complain, but it doesn't delete the temp directory either. I think JVM unloads classes by the time the shutdown hook code executes, unfortunately. But deleting a directory recursively may already be too expensive to be executed as a shutdown hook. I wish we could delete these directories much earlier before JVM termination.

```
[INFO] BUILD SUCCESS
[INFO] Total time:  5.810 s
[INFO] Finished at: 2019-09-23T12:47:38-04:00
Exception in thread "Thread-3" Exception in thread "Thread-2" java.lang.NoClassDefFoundError: com/google/common/io/RecursiveDeleteOption
        at com.google.cloud.tools.jib.filesystem.FileOperations.lambda$deleteRecursiveOnExit$1(FileOperations.java:102)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ClassNotFoundException: com.google.common.io.RecursiveDeleteOption
        at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:50)
        at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass(ClassRealm.java:271)
        at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:247)
        at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:239)
        ... 2 more
java.lang.NoClassDefFoundError: com/google/common/io/RecursiveDeleteOption
        at com.google.cloud.tools.jib.filesystem.FileOperations.lambda$deleteRecursiveOnExit$1(FileOperations.java:102)
        at java.lang.Thread.run(Thread.java:748)
```

TadCordle:
Fixes #2016. Rather than using a shutdown hook to clean up temporary directories, we keep track of them in a thread-safe set that gets passed into the build steps and iterate through them when the build steps finish. I'm not sure how I feel about passing around a list that gets modified inside the build steps, but I think it's overall simpler than doing something like returning multiple values (the temp directories + whatever the steps were originally returning) and collecting them in `StepsRunner`.

Also, right now the only two steps that use temporary directories are `SaveDockerStep` and `ExtractTarStep`, which happen one after the other, so the thread-safe data structure may be unnecessary at the moment.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -88,14 +88,17 @@ static boolean isGzipped(Path path) throws IOException {
...
   private final Path tarPath;
   private final ProgressEventDispatcher.Factory progressEventDispatcherFactory;
+  private final Queue<Path> directoriesToDelete;
 
   ExtractTarStep(
       BuildConfiguration buildConfiguration,
       Path tarPath,
-      ProgressEventDispatcher.Factory progressEventDispatcherFactory) {
+      ProgressEventDispatcher.Factory progressEventDispatcherFactory,
+      Queue<Path> directoriesToDelete) {
```
briandealwis: Why a `Queue` specifically instead of a `Collection` or `Set`?  (Queue's API can throw an exception on `add()`, which is unhandled below.)

Would it make sense to maintain this cleanup logic in the `BuildConfiguration`?  Post-Build Clean Up Hooks?

TadCordle: No reason really, I guess we can use a `Set`.

Not sure if we should keep track of the directories to cleanup in `BuildConfiguration`, since I think `BuildConfiguration` is supposed to be immutable?

I suppose another option would be to create the temporary directories beforehand in `StepsRunner` and pass the paths into the build steps that need them so they're easier to cleanup later.

chanseokoh: This is simpler, but just throwing an idea that I think looks better from an API standpoint:

In a similar vain, we could turn the set to a class like `TempDirectoryProvider` that has `newDirectory()` and keeps track of created directories to delete them later? We already have `TemporaryDirectory`, but looks like it doesn't fit into this in its current form. (And perhaps `StepsRunner` will have to call `close()` in the `finally` block.)

TadCordle: That does seem cleaner from an API standpoint, although I don't like adding new classes for such small use cases. I'll give it a shot regardless.

TadCordle:
I added `TempDirectoryProvider`, which had a lot in common with `TemporaryDirectory`, so I ended up just removing the latter completely. I think this is nicer.

Issue #1913, opened 2019-08-16
Optimize local base image support: more efficient layer processing


TadCordle:
Following up #1906.

- [x] Multi-threaded layer compression
  - [x] First pass (#2022)
  - [x] Separate build steps to allow pushing layers immediately after compressing (#2144)

Old ideas that we don't want to do anymore:
- When extracting the tar, process each entry without writing them to disk (conflicts with caching mechanism)
- Instead of computing the layer digest and then pushing to a registry, push base image layers while compressing on-the-fly (https://github.com/GoogleContainerTools/jib/pull/1906#discussion_r314796323) (conflicts with caching mechanism)
- Untar `docker save` output directly instead of writing to disk first (i.e. `docker save | tar -xf -`) (Doesn't seem to save time, and hurts code readability)
- Use `BEST_SPEED` while gzipping instead of the default compression method (faster compression, but slows down push steps since blobs are larger)

TadCordle:
For the record, I tried "Untar docker save output directly instead of writing to disk first", and it doesn't seem to save much time, if any. It makes the code harder to read as well, since `DockerClient`, `TarExtractor`, and `LocalBaseImageSteps` need to be more tightly connected, so I think we can skip this one.

TadCordle:
I also just tried using `BEST_SPEED` for layer compression. The results are sort of what I expected; the layer compression executes faster, so stuff like `jibDockerBuild` and `jibBuildTar` seems to be slightly faster overall. But the faster compression also results in larger blobs, which means pushing to a registry is slower. In that case this seems like it's not worth it since it will actually cause a slowdown for pushing to registries, especially for users with weaker connections.

loosebazooka:
Should we make it configurable?

TadCordle:
I don't think it's worth adding an extra configuration for it. The speed difference is pretty small anyway, and it's only noticeable when the image hasn't been cached yet.

PR #2022, opened 2019-09-25
Parallelize local base image compression

TadCordle:
Towards #1913. This only really makes a difference for un-cached local base image layers, but it provides a decent speedup in that case. Here were the build times using maven + a clean cache on the master branch vs. these changes:

`docker://gcr.io/distroless/java` 
* master: **7.6 seconds**
* this PR: **6.8 seconds**

`docker://openjdk:8`
* master: **29.2 seconds**
* this PR: **16.4 seconds**

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
@@ -51,6 +51,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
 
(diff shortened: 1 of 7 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/ExtractTarStep.java
```diff
@@ -139,21 +145,31 @@ public LocalImage call()
...
-                  configurationTemplate.getLayerDiffId(index),
-                  layerFile,
-                  layersAreCompressed,
-                  progressEventDispatcher.newChildProducer());
+          DescriptorDigest diffId = configurationTemplate.getLayerDiffId(index);
+          ProgressEventDispatcher.Factory progressEventDispatcherFactory =
+              progressEventDispatcher.newChildProducer();
+
+          cachedLayers.add(
+              executorService.submit(
```
chanseokoh: Ideally, the code here won't deal with concurrency and `StepsRunner` should be able to maximize concurrency by individually compressing and caching each layer. For example, right now it is possible that some registry base image layers are being pulled in while some base layers can be pushed simultaneously. (I am sure constructing a base image manifest can be delayed, since I believe the reason we need them at an early stage is to be able to download the layers from a registry.) Also this would be nice for the consistent async framework design where each step doesn't get an `ExecutorService`.

But I think this is OK for short-term simple improvement for a low-hanging fruit. We should continue to track for the ideal async handling.

TadCordle: Yeah, I agree this should be iterated on to be more consistent/less self-contained. But I figured it'd be nice to at least get the small performance gain in without adding too much complexity at the moment.

TadCordle: Also if we change the build steps more it'd be nice to keep the second bullet in #1913 in mind.

>Instead of computing the layer digest and then pushing to a registry, push base image layers while compressing on-the-fly (#1906 (comment))

chanseokoh: Yeah, I'm already getting a headache over all these. Sounds like a fun challenge!

PR #2023, opened 2019-09-26
Update examples/java-agent plugin versions

briandealwis:
Update plugin versions for `examples/java-agent`, and tweak the example to listen on `$PORT` if provided, as provided by Cloud Run.

### DIFF
--- a/examples/java-agent/pom.xml
+++ b/examples/java-agent/pom.xml
@@ -11,8 +11,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <jib-maven-plugin.version>1.6.1</jib-maven-plugin.version>
     <maven-compiler-plugin.version>3.8.0</maven-compiler-plugin.version>
-    <download-maven-plugin.version>1.4.1</download-maven-plugin.version>
-    <git-commit-id-plugin.version>2.2.5</git-commit-id-plugin.version>
+    <download-maven-plugin.version>1.4.2</download-maven-plugin.version>
+    <git-commit-id-plugin.version>3.0.1</git-commit-id-plugin.version>
 
     <!-- agents will be extracted relative to this build location -->
     <agent-extraction-root>${project.build.directory}/jib-agents</agent-extraction-root>
@@ -28,12 +28,12 @@
     <dependency>
       <groupId>com.sparkjava</groupId>
       <artifactId>spark-core</artifactId>
-      <version>2.7.2</version>
+      <version>2.9.1</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
-      <version>1.7.21</version>
+      <version>1.7.28</version>
     </dependency>
   </dependencies>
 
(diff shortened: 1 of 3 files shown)


examples/java-agent/src/main/java/example/HelloWorld.java
```diff
@@ -17,9 +17,14 @@
 package example;
 
 import static spark.Spark.get;
+import static spark.Spark.port;
```
loosebazooka: The static import seems rather non-google style.
Should we be using `Spark.port(...)`, `Spark.get(...)`

examples/java-agent/pom.xml
```diff
@@ -77,7 +77,9 @@
           <to>
             <image>gcr.io/REPLACE-WITH-YOUR-GCP-PROJECT/image-built-with-jib</image>
           </to>
-          <extraDirectory>${agent-extraction-root}</extraDirectory>
+          <extraDirectories>
+            <paths>${agent-extraction-root}</paths>
+          </extraDirectories>
           <container>
             <ports>4567</ports>
```
loosebazooka: does this need to be modified for cloud run $PORT?

examples/java-agent/src/main/java/example/HelloWorld.java
briandealwis: It is SparkJava style though.  For example: http://sparkjava.com and http://sparkjava.com/documentation#port

examples/java-agent/pom.xml
briandealwis: We can't actually change this on the fly though.  The default for SparkJava is to listen on port 4567.  Cloud Run passes in a PORT and the container is required to listen on that port; Cloud Run doesn't care about the container image's exposed ports.

Let me add a note to the source and here to this effect.  I also forgot to update the build.gradle to use the `extraDirectories`.

examples/java-agent/src/main/java/example/HelloWorld.java
```diff
@@ -17,9 +17,16 @@
...
 
 public class HelloWorld {
   public static void main(String[] args) {
-    get("/hello", (req, res) -> "Hello World");
+    // Allow use with Cloud Run which requires listening on the value in PORT
+    String portEnv = System.getenv("PORT");
```
chanseokoh: Is it defined as 4567 by default somewhere? It should continue to work locally.

briandealwis: SparkJava listens on 4567 by default (mentioned in the README.md)

examples/java-agent/src/main/java/example/HelloWorld.java
```diff
@@ -17,9 +17,16 @@
...
+    if(portEnv != null) {
```
chanseokoh: `if(portEnv` --> `if (portEnv`, or it's just not in Google style?

briandealwis: Huh, I thought we had a validator to break the build.

TadCordle: I don't think that runs on the example projects.

PR #1921, opened 2019-08-21
Add configuration option for filename used by buildTar goal of maven plugin (#1561)

hoppermann:
Hello,

in our projects, we also have the need to specify the file name of the tar file. For this we have created this pull request.

Fix #1561

Issue #1561, opened 2019-03-16
Add configuration option for filename used by buildTar goal of maven plugin


jrivard:
The current (as of 1.0.2) maven plugin has the very useful goal of buildTar which creates a docker tar image named "jib-maven.tar".  It would be very convenient for their to be a configuration option to control the filename for situations when the version or other data needs to be placed into the filename.

Issue #1282 suggests using a separate plugin to accomplish this, but this is awkward at best.  So in this issue I'm requesting specifically that an option be added to allow setting the filename directly by the jib-maven-plugin.

blagerweij:
I would like the same feature for the Gradle plugin!

chanseokoh:
@hoppermann @codeseedr FYI, we are actively working on this based on #2024.

TadCordle:
@jrivard @tishun @oers @codeseedr @blagerweij We've released 1.7.0, which adds a `jib.outputPaths` (gradle) / `<outputPaths>` (maven) configuration block, which lets you specify the `tar`, `digest`, and `imageId` file paths.

PR #2024, opened 2019-09-26
Proposal RFC: custom output file locations

TadCordle:
Formal proposal for #1561, following some discussion in #1921.

doc: https://github.com/GoogleContainerTools/jib/blob/i1561-proposal/proposals/output_files.md


TadCordle:
@jrivard @blagerweij @hoppermann @codeseedr

Feel free to let us know if this meets what you're looking for with this feature, or request changes!

proposals/outputFiles.md
```diff
@@ -0,0 +1,46 @@
...
+</configuration>
+```
+
+#### Gradle (`build.gradle`)
+```groovy
+jib {
+  outputFiles {
+    tar = file('/some/location.tar')
+    digest = file('/some/other/location.digest')
+    id = file('$buildDir/id')
```
loosebazooka: think `'` doesn't expand variables, this example might need to be `"$buildDir/id"`

TadCordle: Ah ok

proposals/output_files.md
```diff
@@ -0,0 +1,46 @@
...
+
+- Image tar -> `${buildDir}/jib-image.tar`
+- Image ID -> `${buildDir}/jib-image.id`
+- Image digest -> `${buildDir}/jib-image.digest`
+
+## Proposed Configuration
+
+The proposal is to allow users to configure their build with the following rules:
+1. Extensions to the filename (like id, digest, tar) will not be automatically appended
+1. Existing files at the specified locations will be overwritten
```
chanseokoh: These are a good point I haven't thought about. For example, we mark `@OutputFile` for the tar in Gradle, and I believe this establishes the usual Gradle input-output dependency relation so that if nothing is changed, Gradle won't rebuild the tar. So does this propose that we take out `@OutputFile` and always regenerate it?

And now I don't remember what was good about allowing path outside the build directory. Should we reconsider this? For example, you can only specify a relative path anchored at the build output directory. But then, people can do `../../`. But it is easy to guard against going out of the build directory.

proposals/output_files.md
```diff
@@ -0,0 +1,46 @@
...
+1. Running `clean` will not delete output files created outside of the project's build directory
+
+#### Maven (`pom.xml`)
+```xml
+<configuration>
+  <outputFiles>
+    <tar>/some/location.tar</tar>
+    <digest>/some/other/location.digest</digest>
+    <id>${project.build.directory}/id</id>
```
chanseokoh: We should clarify if a relative path can be used (I think it should) and what would be the final path in that case. Probably relative to the build output directory.

briandealwis: `id` is a bit confusing given that it's used quite widely in maven to identify `<execution>` blocks and others.  Maybe `imageId`?

proposals/output_files.md
```diff
@@ -0,0 +1,46 @@
...
+#### Gradle (`build.gradle`)
+```groovy
+jib {
+  outputFiles {
+    tar = file("/some/location.tar")
+    digest = file("/some/other/location.digest")
+    id = file("$buildDir/id")
+  }
+}
+```
```
briandealwis: Would we then support properties like `jib.outputFiles.tar=xxx`?

(Would `outputs` be less restrictive — in case we need to support a directory in the future?)

proposals/output_files.md
TadCordle: >So does this propose that we take out `@OutputFile` and always regenerate it?

That shouldn't be necessary, the only time we'd want to force an overwrite is if any of the inputs change. If we remove `@OutputFile` then I think there's a chance we overwrite the tar with an identical tar, which is useless.

>And now I don't remember what was good about allowing path outside the build directory. Should we reconsider this?

Unless there's some sort of explicit vulnerability here then I don't think there's a problem with allowing it. We're introducing this change is to make the builds more customizable, and limiting the paths they can choose seems like it could just add another possible roadblock.

proposals/output_files.md
TadCordle: I do too, but I'm thinking relative to the build file/root directory makes more sense. At least that's how relative paths work with `tar://` base images.

proposals/output_files.md
TadCordle: >Would we then support properties like `jib.outputFiles.tar=xxx`?

Yeah, I can add that in here.

>(Would `outputs` be less restrictive — in case we need to support a directory in the future?)

Maybe `outputPaths`? If we change to `outputs` then I think we'd need to change the fields to something like `digestPath`, `imageIdPath`, `tarPath`, since something like `jib.outputs.digest` sounds like you're setting the digest or something weird.

proposals/output_files.md
loosebazooka: Relative paths should just be resolved in the build tool default manner?

TadCordle: Yeah, that's probably a better way of putting it.

proposals/output_files.md
chanseokoh: Oh, I think I mistakenly thought the statement was saying "... will always be overwritten". "... overwritten if necessary" could make this clear.

Additionally, I was thinking if `@OutputFile` is given, Gradle will delete the file when running `clean`, as the input-output relation is automatically established and maintained. Is it not the case if the file is outside the build directory? Can you confirm if Gradle `clean` simply deletes its usual build directory (`build/`)?

proposals/output_files.md
```diff
@@ -0,0 +1,52 @@
...
+    <imageId>${project.build.directory}/id</imageId>
+  </outputPaths>
+</configuration>
+```
+
+#### Gradle (`build.gradle`)
+```groovy
+jib {
+  outputPaths {
+    tar = file("/absolute/location.tar")
```
loosebazooka: I think we should allow the user to use strings here. Gradle lets you accept `Object` and then resolve that yourself using `project.file()`. Like how we do in `ExtraDirectoryParameters.setPaths(Object paths)`

TadCordle: I thought gradle auto-converts strings to paths by default? I thought the reason we used `Object` with `extraDirectories` was so we could accept both a single path or a list of paths.

proposals/output_files.md
loosebazooka: I thought `@OutputFile` cleans if you specify that you want to clean that task specifically. For example `./gradlew cleanMyCustomTask`. But yeah worth confirming what it is.

proposals/output_files.md
loosebazooka: Yeah, I think we're saying the same thing. This example of gradle config uses `file()` everywhere though. Think one example with just `tar = "/absolute/location.tar"` might be helpful.

proposals/output_files.md
TadCordle: Looks like if I build a tar outside the build directory, `./gradlew clean` does not delete it even though the path is marked with `@OutputFile`.

proposals/output_files.md
TadCordle: Oh, sure.

proposals/output_files.md
TadCordle: @loosebazooka looks like that works, `./gradlew cleanJibBuildTar` deletes the tar.

TadCordle: Maybe we should mark the digest and id files as `@OutputFile` as well?

loosebazooka:
looks good to me, @chanseokoh, @briandealwis we good?

chanseokoh:
Looks good to me.

PR #2025, opened 2019-09-26
Add configuration for custom output file locations

### LINKED ISSUE #1561: Add configuration option for filename used by buildTar goal of maven plugin

TadCordle:
Fixes #1561.

Adds `outputPaths` configuration block with `tar`, `digest`, and `imageId` fields for setting output file locations.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
@@ -58,7 +58,9 @@ public class WriteTarFileStep implements Callable<BuildResult> {
     try (ProgressEventDispatcher ignored =
         progressEventDispatcherFactory.create("writing to tar file", 1)) {
       // Builds the image to a tarball.
-      Files.createDirectories(outputPath.getParent());
+      if (outputPath.getParent() != null) {
+        Files.createDirectories(outputPath.getParent());
+      }
       try (OutputStream outputStream =
           new BufferedOutputStream(FileOperations.newLockingOutputStream(outputPath))) {
         new ImageTarball(builtImage, buildConfiguration.getTargetImageConfiguration().getImage())
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -5,10 +5,10 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- `jib.outputFiles` object for configuration output file locations ([#1561](https://github.com/GoogleContainerTools/jib/issues/1561))
-  - `jib.outputFiles.tar` configures output path of `jibBuildTar` (`build/jib-image.tar` by default)
-  - `jib.outputFiles.digest` configures the output path of the image digest (`build/jib-image.digest` by default)
-  - `jib.outputFiles.id` configures output path of the image id  (`build/jib-image.id` by default)
+- `jib.outputPaths` object for configuration output file locations ([#1561](https://github.com/GoogleContainerTools/jib/issues/1561))
+  - `jib.outputPaths.tar` configures output path of `jibBuildTar` (`build/jib-image.tar` by default)
+  - `jib.outputPaths.digest` configures the output path of the image digest (`build/jib-image.digest` by default)
+  - `jib.outputPaths.imageId` configures output path of the image id  (`build/jib-image.id` by default)
 
 ### Changed
 
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- `jib.outputFiles` object for configuration output file locations ([#1561](https://github.com/GoogleContainerTools/jib/issues/1561))
+  - `jib.outputFiles.tar` configures output path of `jibBuildTar` (`build/jib-image.tar` by default)
+  - `jib.outputFiles.digest` configures the output path of the image digest (`build/jib-image.digest` by default)
+  - `jib.outputFiles.id` configures output path of the image id  (`build/jib-image.id` by default)
+
 ### Changed
 
 - Local base image layers are now processed in parallel, speeding up builds using large local base images. ([#1913](https://github.com/GoogleContainerTools/jib/issues/1913))
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -376,7 +376,7 @@ public class SingleProjectIntegrationTest {
         output, new Command("docker", "run", "--rm", imageReferenceWithDigest).run());
 
     String id =
-        readDigestFile(simpleTestProject.getProjectRoot().resolve("build/different-jib-image.id"));
+        readDigestFile(simpleTestProject.getProjectRoot().resolve("different-jib-image.id"));
     Assert.assertNotEquals(digest, id);
     Assert.assertEquals(output, new Command("docker", "run", "--rm", id).run());
 
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -121,7 +121,7 @@ public class SingleProjectIntegrationTest {
                 + "            }"));
   }
 
-  static String assertDigestFile(Path digestPath) throws IOException, DigestException {
+  static String readDigestFile(Path digestPath) throws IOException, DigestException {
     Assert.assertTrue(Files.exists(digestPath));
     String digest = new String(Files.readAllBytes(digestPath), StandardCharsets.UTF_8);
     return DescriptorDigest.fromDigest(digest).toString();
@@ -217,12 +217,12 @@ public class SingleProjectIntegrationTest {
         output);
 
     String digest =
-        assertDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.digest"));
+        readDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.digest"));
     String imageReferenceWithDigest = ImageReference.parse(targetImage).withTag(digest).toString();
     Assert.assertEquals(
         output, JibRunHelper.buildAndRun(simpleTestProject, imageReferenceWithDigest));
 
-    String id = assertDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.id"));
+    String id = readDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.id"));
     Assert.assertNotEquals(digest, id);
     Assert.assertEquals(output, new Command("docker", "run", "--rm", id).run());
 
@@ -369,7 +369,7 @@ public class SingleProjectIntegrationTest {
     String output = buildAndRunComplex(targetImage, "testuser2", "testpassword2", localRegistry2);
 
     String digest =
-        assertDigestFile(
+        readDigestFile(
             simpleTestProject.getProjectRoot().resolve("build/different-jib-image.digest"));
     String imageReferenceWithDigest = ImageReference.parse(targetImage).withTag(digest).toString();
     localRegistry2.pull(imageReferenceWithDigest);
@@ -377,8 +377,7 @@ public class SingleProjectIntegrationTest {
         output, new Command("docker", "run", "--rm", imageReferenceWithDigest).run());
 
     String id =
-        assertDigestFile(
-            simpleTestProject.getProjectRoot().resolve("build/different-jib-image.id"));
+        readDigestFile(simpleTestProject.getProjectRoot().resolve("build/different-jib-image.id"));
     Assert.assertNotEquals(digest, id);
     Assert.assertEquals(output, new Command("docker", "run", "--rm", id).run());
 
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -18,9 +18,13 @@ package com.google.cloud.tools.jib.gradle;
 
 import com.google.cloud.tools.jib.Command;
 import com.google.cloud.tools.jib.IntegrationTestingConfiguration;
+import com.google.cloud.tools.jib.api.DescriptorDigest;
+import com.google.cloud.tools.jib.api.ImageReference;
+import com.google.cloud.tools.jib.api.InvalidImageReferenceException;
 import com.google.cloud.tools.jib.registry.LocalRegistry;
 import com.google.common.base.Splitter;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.DigestException;
@@ -117,6 +121,12 @@ public class SingleProjectIntegrationTest {
                 + "            }"));
   }
 
+  static String assertDigestFile(Path digestPath) throws IOException, DigestException {
+    Assert.assertTrue(Files.exists(digestPath));
+    String digest = new String(Files.readAllBytes(digestPath), StandardCharsets.UTF_8);
+    return DescriptorDigest.fromDigest(digest).toString();
+  }
+
   private static void assertExtraDirectoryDeprecationWarning(String buildFile)
       throws DigestException, IOException, InterruptedException {
     String targetImage = "localhost:6000/simpleimage:gradle" + System.nanoTime();
@@ -134,7 +144,7 @@ public class SingleProjectIntegrationTest {
                 + "'jib.extraDirectories.permissions'"));
   }
 
-  private static void buildAndRunComplex(
+  private static String buildAndRunComplex(
       String imageReference, String username, String password, LocalRegistry targetRegistry)
       throws IOException, InterruptedException {
     Path baseCache = simpleTestProject.getProjectRoot().resolve("build/jib-base-cache");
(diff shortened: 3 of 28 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -684,6 +695,39 @@ Path getDockerClientExecutable() {
     return dockerClient.environment;
   }
 
+  Path getTarOutputPath() {
+    String property = getProperty(PropertyNames.OUTPUT_PATHS_TAR);
+    if (property != null) {
+      return Paths.get(property);
+    }
+    return outputPaths.tar == null
+        ? Paths.get(Preconditions.checkNotNull(project).getBuild().getDirectory())
```
chanseokoh: nit: `getProject().getBuild()`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -684,6 +695,39 @@ Path getDockerClientExecutable() {
...
+        : outputPaths.tar.toPath();
+  }
+
+  Path getDigestOutputPath() {
+    String property = getProperty(PropertyNames.OUTPUT_PATHS_DIGEST);
+    if (property != null) {
+      return Paths.get(property);
+    }
+    return outputPaths.digest == null
+        ? Paths.get(Preconditions.checkNotNull(project).getBuild().getDirectory())
```
chanseokoh: nit: `getProject().getBuild()`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -684,6 +695,39 @@ Path getDockerClientExecutable() {
...
+        : outputPaths.digest.toPath();
+  }
+
+  Path getIdOutputPath() {
+    String property = getProperty(PropertyNames.OUTPUT_PATHS_IMAGE_ID);
+    if (property != null) {
+      return Paths.get(property);
+    }
+    return outputPaths.imageId == null
+        ? Paths.get(Preconditions.checkNotNull(project).getBuild().getDirectory())
```
chanseokoh: nit: `getProject().getBuild()`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -684,6 +695,39 @@ Path getDockerClientExecutable() {
...
+    if (property != null) {
+      return Paths.get(property);
+    }
+    return outputPaths.digest == null
+        ? Paths.get(Preconditions.checkNotNull(project).getBuild().getDirectory())
+            .resolve("jib-image.digest")
+        : outputPaths.digest.toPath();
+  }
+
+  Path getIdOutputPath() {
```
chanseokoh: getImageIdOutputPath

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
```diff
@@ -190,4 +190,19 @@ public String getCreationTime() {
...
+    return jibPluginConfiguration.getTarOutputPath();
+  }
+
+  @Override
+  public Path getDigestOutputPath() {
+    return jibPluginConfiguration.getDigestOutputPath();
+  }
+
+  @Override
+  public Path getIdOutputPath() {
```
chanseokoh: getImageIdOutputPath

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
```diff
@@ -91,4 +91,10 @@
   Map<String, String> getDockerEnvironment();
 
   String getContainerizingMode();
+
+  Path getTarOutputPath();
+
+  Path getDigestOutputPath();
+
+  Path getIdOutputPath();
```
chanseokoh: getImageIdOutputPath

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
```diff
@@ -58,7 +58,9 @@ public BuildResult call() throws ExecutionException, InterruptedException, IOExc
     try (ProgressEventDispatcher ignored =
         progressEventDispatcherFactory.create("writing to tar file", 1)) {
       // Builds the image to a tarball.
-      Files.createDirectories(outputPath.getParent());
+      if (outputPath.getParent() != null) {
+        Files.createDirectories(outputPath.getParent());
```
chanseokoh: Does this work correctly when the working directory is not the project root? For example, `mvn -f examples/helloworld/pom.xml compile ...`.

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- `jib.outputFiles` object for configuration output file locations ([#1561](https://github.com/GoogleContainerTools/jib/issues/1561))
```
chanseokoh: outputPaths here and below

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -117,6 +121,12 @@ private static void assertDockerInspect(String imageReference)
                 + "            }"));
   }
 
+  static String assertDigestFile(Path digestPath) throws IOException, DigestException {
+    Assert.assertTrue(Files.exists(digestPath));
+    String digest = new String(Files.readAllBytes(digestPath), StandardCharsets.UTF_8);
```
chanseokoh: nit: `Files.readAllLines` might be easier?

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -157,11 +167,14 @@ private static void buildAndRunComplex(
...
     Assert.assertThat(history, CoreMatchers.containsString("jib-gradle-plugin"));
+
+    String output = new Command("docker", "run", "--rm", imageReference).run();
     Assert.assertEquals(
         "Hello, world. An argument.\n1970-01-01T00:00:01Z\nrwxr-xr-x\nrwxrwxrwx\nfoo\ncat\n"
             + "1970-01-01T00:00:01Z\n1970-01-01T00:00:01Z\n"
             + "-Xms512m\n-Xdebug\nenvvalue1\nenvvalue2\n",
-        new Command("docker", "run", "--rm", imageReference).run());
+        output);
+    return output;
```
chanseokoh: Is returning `output` ever useful? Because the above assertion already checks that the output is always some fixed string. There cannot be any variation.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -195,10 +209,23 @@ public void testBuild_simple() throws IOException, InterruptedException, DigestE
...
         "Hello, world. An argument.\n1970-01-01T00:00:01Z\nrw-r--r--\nrw-r--r--\nfoo\ncat\n"
             + "1970-01-01T00:00:01Z\n1970-01-01T00:00:01Z\n",
-        JibRunHelper.buildAndRun(simpleTestProject, targetImage));
+        output);
+
+    String digest =
+        assertDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.digest"));
+    String imageReferenceWithDigest = ImageReference.parse(targetImage).withTag(digest).toString();
+    Assert.assertEquals(
+        output, JibRunHelper.buildAndRun(simpleTestProject, imageReferenceWithDigest));
```
chanseokoh: Do we have to run this again? `jib-image.id` is also generated when pushing to a registry. I'd like to avoid adding another Jib run if it's not so meaningful.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -91,7 +91,7 @@ public FileCollection getInputFiles() {
    */
   @OutputFile
   public String getOutputFile() {
-    return getTargetPath().toString();
+    return Preconditions.checkNotNull(jibExtension).getOutputPaths().getTarPath().toString();
```
chanseokoh: Likewise, will a relative path always work correctly?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleRawConfiguration.java
```diff
@@ -185,4 +185,19 @@ public String getCreationTime() {
...
+    return jibExtension.getOutputPaths().getTarPath();
+  }
+
+  @Override
+  public Path getDigestOutputPath() {
+    return jibExtension.getOutputPaths().getDigestPath();
+  }
+
+  @Override
+  public Path getIdOutputPath() {
```
chanseokoh: getImageIdOutputPath

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -56,6 +56,11 @@
  *       '/path/on/container/file2': 123
  *     ]
  *   }
+ *   outputPaths {
+ *     tar = file('jib-image.tar')
+ *     digest = file('jib-image.digest')
+ *     imageId = file('jib-image.id')
```
chanseokoh: I don't like this example much, since it will put these files at the project root, different from the current default behavior. It can create confusion to those new to read our code.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/OutputPathsParameters.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.cloud.tools.jib.plugins.common.PropertyNames;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import javax.inject.Inject;
+import org.gradle.api.Project;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Internal;
+
+/** Object that configures where Jib should create its build output files. */
+public class OutputPathsParameters {
```
chanseokoh: @loosebazooka can you take a second look at this Gradle part?

jib-maven-plugin/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- `<outputFiles>` object for configuration output file locations ([#1561](https://github.com/GoogleContainerTools/jib/issues/1561))
```
chanseokoh: outputPaths

jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -76,20 +76,12 @@ private static String getTestImageReference(String label) {
     return nameBase + label + System.nanoTime();
   }
 
-  static String assertImageDigest(Path projectRoot) throws IOException, DigestException {
-    Path digestPath = projectRoot.resolve("target/jib-image.digest");
+  static String assertDigestFile(Path digestPath) throws IOException, DigestException {
```
chanseokoh: Looking at this again, I think `readDigestFile()` sounds better than `assertDigestFile`.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -117,6 +121,12 @@ private static void assertDockerInspect(String imageReference)
                 + "            }"));
   }
 
+  static String assertDigestFile(Path digestPath) throws IOException, DigestException {
```
chanseokoh: Looking at this again, I think `readDigestFile()` sounds better than `assertDigestFile`.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
TadCordle: That returns a list so I think this is cleaner.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
TadCordle: It's used in `testBuild_complex` to make sure the output is the same when it's run using the digest and image id. I wanted to avoid copy-pasting the long string.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
TadCordle: Not sure what you mean; this is run twice to make sure the digest generated by Jib is valid. The same logic happens in the maven integration tests, it's just been missing from gradle for a while it seems.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/WriteTarFileStep.java
TadCordle: Yup

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
TadCordle: This one seems to act a bit weird. It looks like for maven, paths are relative to the project root, but for gradle, paths are relative to the directory the command is run from. If I do something like build the tar outside the project directory then try to rebuild inside the project directory, it skips generating the tar the second time. Even stranger is that when I delete the original tar it still refuses to regenerate it. Might be safer to just remove `@OutputFile`.

chanseokoh: Hmm... I like `@OutputFile` because a tarball can be large and Gradle will avoid generating it. (Now you said it, I wonder if Maven works in the same manner.)

Some options:
- We just say this is a bug to be fixed in Gradle. Sounds like difficult to fix though.
- Don't support relative path. I think this is not a big deal.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
chanseokoh: Ah... let me look at this again.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
chanseokoh: Yeah, I saw it's being used, but what I saying is that whenever `buildAndRunComplex()` is called, the output is some fixed string which is verified inside `buildAndRunComplex()`. The returned output will always be same no matter how many you call `buildAndRunComplex()` differently, so comparing the two output strings seems redundant.

TadCordle: Maybe I'm not following, but the reason for returning the output (even though yes, it should never change) is because building with the digest/running with the id should return identical output to the original build; it doesn't matter exactly what that output is, just that it's the same. Would it be better style to assert the output outside of `buildAndRunComplex()`?

chanseokoh: Ah, sorry. I misunderstood.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
chanseokoh: Actually, I think we can just prepend the project root if this is a relative path. The behavior will be no different from specifying an absolute path.

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -195,10 +209,22 @@ public void testBuild_simple() throws IOException, InterruptedException, DigestE
...
     Assert.assertEquals(
         "Hello, world. An argument.\n1970-01-01T00:00:01Z\nrw-r--r--\nrw-r--r--\nfoo\ncat\n"
             + "1970-01-01T00:00:01Z\n1970-01-01T00:00:01Z\n",
-        JibRunHelper.buildAndRun(simpleTestProject, targetImage));
+        output);
+
+    String digest =
+        readDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.digest"));
+    String imageReferenceWithDigest = ImageReference.parse(targetImage).withTag(digest).toString();
+    Assert.assertEquals(output, JibRunHelper.pullAndRunBuiltImage(imageReferenceWithDigest));
```
chanseokoh: Now that you run `pullAndRunBuiltImage()`, I think you don't need `new Command("docker", "run", ...)` below anymore. And the order would be less confusing if we do

```java
    String digest = readDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.digest"));
    String id = readDigestFile(simpleTestProject.getProjectRoot().resolve("build/jib-image.id"));
    Assert.assertNotEquals(digest, id);

    String imageReferenceWithDigest = ImageReference.parse(targetImage).withTag(digest).toString();
    Assert.assertEquals(output, JibRunHelper.pullAndRunBuiltImage(imageReferenceWithDigest))
```
```

TadCordle: The second `new Command("docker", "run", ...)` is to test running with the generated image id; this one tests the digest.

chanseokoh: Ah, now I get it. Cool!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
TadCordle: Done.

TadCordle:
Ugh windows.

PR #2033, opened 2019-10-02
mockito 3.1.0

sullis:


loosebazooka:
Looks like this is generally safe: https://github.com/mockito/mockito/releases/tag/v3.0.0

> Mockito 3.0.0 is equivalent to Mockito 2, but now requires Java 8. No API changes were made.

PR #2035, opened 2019-10-02
Don't pass redundant parameter

chanseokoh:
The API route base can be computed and is always "server URL + /v2/".

I also noted every `getApiRoute()` implementation does not put an extra leading slash, so I updated the method in tests reflect the reality more closely. An example implementation:
```java
  public URL getApiRoute(String apiRouteBase) throws MalformedURLException {
    return new URL(
        apiRouteBase + registryEndpointRequestProperties.getImageName() + "/manifests/" + imageTag);
  }
```

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
@@ -56,8 +56,6 @@ class RegistryEndpointCaller<T> {
    */
   @VisibleForTesting static final int STATUS_CODE_PERMANENT_REDIRECT = 308;
 
-  private static final String DEFAULT_PROTOCOL = "https";
-
   private static boolean isHttpsProtocol(URL url) {
     return "https".equals(url.getProtocol());
   }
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -140,8 +135,6 @@ static boolean isBrokenPipe(IOException original) {
       @Nullable Function<URL, Connection> insecureConnectionFactory)
       throws MalformedURLException {
     this.eventHandlers = eventHandlers;
-    this.initialRequestUrl =
-        registryEndpointProvider.getApiRoute(DEFAULT_PROTOCOL + "://" + apiRouteBase);
```
TadCordle: Can we remove the `DEFAULT_PROTOCOL` constant, or is it used elsewhere? (even if it is, I think just hardcoding "https" is fine)

chanseokoh: Good catch.

PR #2038, opened 2019-10-03
Add cache methods for writing/reading local base image container config

TadCordle:
Part of `docker save` portion of #1912. Adds methods for writing/retrieving container configuration to/from the cache using an image ID.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
@@ -131,7 +131,13 @@ public class Cache {
   }
 
   /**
-   * Writes a container configuration to {@code (cache directory)/local/config/(image id)}.
+   * Writes a container configuration to {@code (cache directory)/local/config/(image id)}. An image
+   * ID is a SHA hash of a container configuration JSON. The value is also shown as IMAGE ID in
+   * {@code docker images}.
+   *
+   * <p>Note: the {@code imageID} to the {@code containerConfiguration} is a one-way relationship;
+   * there is no guarantee that {@code containerConfiguration}'s SHA will be {@code imageID}, since
+   * the original container configuration is being rewritten here rather than being moved.
    *
    * @param imageId the ID of the image to store the container configuration for
    * @param containerConfiguration the container configuration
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
```diff
@@ -130,6 +130,19 @@ public CachedLayer writeTarLayer(DescriptorDigest diffId, Blob compressedBlob)
     return cacheStorageWriter.writeTarLayer(diffId, compressedBlob);
   }
 
+  /**
+   * Writes a container configuration to {@code (cache directory)/local/config/(image id)}.
```
chanseokoh: I'd like to add something like

"An image ID is a SHA hash of a container configuration JSON. The value is also shown as IMAGE ID in `docker images`."

But as I commented in the design doc, this can still be confusing without a correct context. Basically, you can put anything for `imageId` to associate it with `containerConfiguration`, and almost 100% time, the actual SHA of the saved `containerConfiguration` will be different from `imageId`.

TadCordle: To be clear, I don't think it really matters that the SHA of the saved configuration will be different from the image ID; what matters is that there's a consistent relationship from the SHA reported by the docker daemon image to the cached configuration. In other words, it's less "the configuration saved here is identical to the docker daemon image configuration with the same SHA" and more "we've seen a configuration with this SHA before, and we saved the configuration here after processing it a bit". But yeah, I can update some of the comments to make sure the misconception is less likely to happen.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
```diff
@@ -223,6 +223,26 @@
...
+   * Retrieves the {@link ContainerConfigurationTemplate} for the image with the given image ID.
+   *
+   * @param imageId the image ID
+   * @return the {@link ContainerConfigurationTemplate} referenced by the image ID, if found
+   * @throws IOException if an I/O exception occurs
+   */
+  Optional<ContainerConfigurationTemplate> retrieveLocalConfig(DescriptorDigest imageId)
+      throws IOException {
+    Path configPath =
+        cacheStorageFiles.getLocalDirectory().resolve("config").resolve(imageId.getHash());
```
chanseokoh: Now I remember what was the risk of moving a file instead of moving a directory. Atomic move can fail (e.g., on NFS) and fall back to copying. Then it is possible that a reader opens a file that is being filled in by a writer concurrently. Does that make sense?

TadCordle: Ah, yeah that makes sense.

chanseokoh: Hmm... on second thought this isn't any different from copying a directory recursively. A file can still be being copied. So I guess moving a file is not worse.

TadCordle: Oh I just made the changes. I guess I'll revert since the former is simpler. Do you think we should go with a lockfile then, or is it rare enough to not bother?

chanseokoh: Yeah, I guess we are good. Not using a lockfile is currently a known flaw. :( Our assumption requires that the atmoic move should work. When it doesn't, we can't guarantee correctness.

PR #2039, opened 2019-10-03
Check base image cache if referenced by digest

### LINKED ISSUE #1881: Base image metadata (manifest and container configuration) is pulled for every build
**Environment**:

- *Jib version:* 1.4.0
- *Build tool:*  maven 3.6.1
- *OS:* Windows 10

**Description of the issue**:

We would like for JIB to only consult the registry containing the from-image, if that image is not already cached.    Currently it appears that this happens for each and every build (probably to see if the image has been updated).  For our purposes we are fine with just populating the cache once, and then use that entry for the lifetime of the instance (pod).  If we need updating, we just restart the pod.

**Expected behavior**:

Registry containing from-image is only contacted at all if image is not cached.  If cached, no network traffic should happen.  This can emulated with a VERY long timeout before trying to refresh a cache entry.

This should be settable as a pom-entry, and overridable from the command line.

**Steps to reproduce**:

Any build.  For our proof-of-concept build, we use "adoptopenjdk/openjdk11:slim" as the from image.

```xml
          <plugin>
                <!-- https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin -->
                <version>1.4.0</version>
                <configuration>
                    <container>
                        <ports>
                            <port>8080</port>
                        </ports>
                    </container>
                    <from>
                        <!-- https://stackoverflow.com/a/52431765/53897 -->
                        <image>adoptopenjdk/openjdk11:slim</image>
                    </from>
                    <to>
                        <image>${docker.image.local.name}</image>
                    </to>
                    <extraDirectories>
                        <paths>
                            <path>extra-directory-for-jib</path>
                        </paths>
                    </extraDirectories>
                    <dockerClient>
                        <environment>
                            <key3>value3</key3>
                            <key4>value4</key4>
                        </environment>
                    </dockerClient>

                </configuration>
            </plugin>
```

**Additional Information**:

chanseokoh:
Closes #1881.

If the base image is referenced by a digest, pull the manifest only when it was never pulled. This is what we decided to do regarding the issue: https://github.com/GoogleContainerTools/jib/issues/1881#issuecomment-517401175 and https://github.com/GoogleContainerTools/jib/issues/1881#issuecomment-517683340.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -29,7 +29,6 @@ import com.google.cloud.tools.jib.builder.TimerEventDispatcher;
 import com.google.cloud.tools.jib.builder.steps.PullBaseImageStep.ImageAndAuthorization;
 import com.google.cloud.tools.jib.cache.CacheCorruptedException;
 import com.google.cloud.tools.jib.configuration.BuildConfiguration;
-import com.google.cloud.tools.jib.configuration.ImageConfiguration;
 import com.google.cloud.tools.jib.event.EventHandlers;
 import com.google.cloud.tools.jib.event.events.ProgressEvent;
 import com.google.cloud.tools.jib.http.Authorization;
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -101,36 +101,37 @@ public ImageAndAuthorization call()
     EventHandlers eventHandlers = buildConfiguration.getEventHandlers();
     // Skip this step if this is a scratch image
     ImageConfiguration baseImageConfiguration = buildConfiguration.getBaseImageConfiguration();
-    if (baseImageConfiguration.getImage().isScratch()) {
+    ImageReference imageReference = baseImageConfiguration.getImage();
```
TadCordle: I think we can combine these two lines? Doesn't look like `baseImageConfiguration` is used anywhere else.

PR #2041, opened 2019-10-04
Get size, image ID, and diff IDs from "docker inspect"

TadCordle:
Part of `docker save` portion of #1912.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
@@ -47,6 +47,9 @@ import java.util.function.Function;
 /** Calls out to the {@code docker} CLI. */
 public class DockerClient {
 
+  /**
+   * Contains the size, image ID, and diff IDs of an image inspected with {@code docker inspect}.
+   */
   public static class InspectResults {
     private long size;
     private String imageId;
@@ -124,6 +127,14 @@ public class DockerClient {
     };
   }
 
+  /**
+   * Parses the results of {@code docker inspect} into an {@link InspectResults}.
+   *
+   * @param output the output of the {@code docker inspect} command containing the size, image ID,
+   *     and diff IDs
+   * @return the {@link InspectResults}
+   * @throws DigestException if parsing the digests fails
+   */
   @VisibleForTesting
   static InspectResults parseInspectResults(String output) throws DigestException {
     List<String> items = Splitter.on(',').splitToList(output);
@@ -135,7 +146,7 @@ public class DockerClient {
         Splitter.on(' ').splitToList(items.get(2).replace("[", "").replace("]", ""));
     List<String> processedDiffIds = new ArrayList<>(diffIds.size());
     for (String diffId : diffIds) {
-      processedDiffIds.add(DescriptorDigest.fromDigest(diffId).getHash());
+      processedDiffIds.add(DescriptorDigest.fromDigest(diffId.trim()).getHash());
     }
 
     return new InspectResults(size, imageId, processedDiffIds);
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -42,6 +47,33 @@
 /** Calls out to the {@code docker} CLI. */
 public class DockerClient {
 
+  /**
+   * Contains the size, image ID, and diff IDs of an image inspected with {@code docker inspect}.
+   */
+  public static class InspectResults {
```
briandealwis: Can we use a more specific name, like `CachedImageDetails`?  This is capturing the results of our `inspect -f ...`, not the general `docker inspect` case.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
```diff
@@ -215,24 +272,25 @@ public void tag(ImageReference originalImageReference, ImageReference newImageRe
...
+   * @throws DigestException if parsing the image ID or diff IDs failed
    */
-  public long sizeOf(ImageReference imageReference) throws IOException, InterruptedException {
-    Process sizeProcess = docker("inspect", "-f", "{{.Size}}", imageReference.toString());
-
-    if (sizeProcess.waitFor() != 0) {
+  public InspectResults inspect(ImageReference imageReference)
+      throws IOException, InterruptedException, DigestException {
+    Process inspectProcess =
+        docker("inspect", "-f", "{{.Size}},{{.Id}},{{.RootFS.Layers}}", imageReference.toString());
```
briandealwis: You could use `{{range .RootFS.Layers}}{{.}},{{end}}` in the format to simplify the parsing and treat items 3..N as the layer diffsIds?

Is it worth adding [`--type image`](https://github.com/docker/cli/blob/master/docs/reference/commandline/inspect.md#specify-target-type---type) to avoid any possibility of the user providing a name to something that is not an image?

TadCordle: Looking at this again, I'm thinking it might be simpler to just format the docker output as a JSON string and parse that instead. e.g. `docker inspect -f '{"size":{{.Size}},"imageId":"{{.Id}}","diffIds":{{json .RootFS.Layers}}}'`. That way `parseInspectResults` just becomes `return JsonTemplateMapper.readJson(output, InspectResults.class);`

The method already takes an `ImageReference`, so I think it'd be strange for the user to pass in the name of a non-image here. Not sure the extra validation is necessary.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
TadCordle: I'll name it `DockerImageDetails` so we don't confuse it with Jib's cache.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/DockerClient.java
briandealwis: I don't think the use of `--type image` hurts.  But the possibility of clashes does seem small: volumes and containers can't have `/` in their names, and the people are unlikely to use official docker images like `openjdk` without a tag (I hope).

briandealwis: Without `--type image` we get a confusing error if we have a non-image object by the same name:
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.6.2-SNAPSHOT:dockerBuild (default-cli) on project first: 'docker inspect' command failed with error: Template parsing error: template: :1:10: executing "" at <.Size>: map has no entry for key "Size"
```
With --type image` we get a more obvious error:
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.6.2-SNAPSHOT:dockerBuild (default-cli) on project first: 'docker inspect' command failed with error: Error: No such image: openjdk
```

`--type image` has an added bonus as it uses `No such *image*` in the error.  If we use non-existent object:
```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.6.2-SNAPSHOT:dockerBuild (default-cli) on project first: 'docker inspect' command failed with error: Error: No such object: non-existant
```

I think the more meaningful error warrants the extra two arguments.

PR #2044, opened 2019-10-07
Combine local base image steps

TadCordle:
Doesn't change any behavior, just refactors some things in preparation for the rest of #1912.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -90,25 +90,57 @@ static boolean isGzipped(Path path) throws IOException {
...
+                new ThrottledAccumulatingConsumer(dockerProgress::dispatchProgress)) {
+          dockerClient.save(imageReference, tarPath, throttledProgressReporter);
+        }
+      }
+
+      return extractTar(tarPath, tempDirectoryProvider, progressEventDispatcher.newChildProducer());
+    }
+  }
+
+  LocalImage tarImageStep(Path tarPath)
```
briandealwis: With this name, I expected the method to tar *up* an image, not extract a tar.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -234,35 +232,42 @@ private void authenticatePush() {
...
-                new ExtractTarStep(
-                        executorService,
-                        buildConfiguration,
-                        results.tarPath.get(),
-                        childProgressDispatcherFactory,
-                        tempDirectoryProvider)
-                    .call());
+                new LocalBaseImageSteps(
+                        executorService, buildConfiguration, childProgressDispatcherFactory)
+                    .tarImageStep(tarPath.get()));
```
briandealwis: This name is also odd as the other `Step`s are classes that are `Callable`, whereas this produces a result.

TadCordle:
I'll fix the naming and change the methods to return `Callable`s. I had it like that at one point but removed it because it seemed like an unnecessary complication, but I agree that the consistency is better.

>the explicit chaining between steps in StepsRunner is not necessarily a bad thing.

I agree with this too, but this isn't really what the change is for - it's to eventually let us skip `docker save` if `docker inspect` indicates that the cache has everything we need. Having these two build steps separate and chained together made sense earlier, but keeping them separate would make the `docker save` caching optimization very messy. This is the best I could come up with that lets us not tie the tar/docker daemon image functionality together so strictly while still letting us share a decent amount of code.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -90,25 +91,58 @@ static boolean isGzipped(Path path) throws IOException {
...
+
+  Callable<LocalImage> processTarBaseImageStep(Path tarPath) {
+    return () -> {
+      try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
+        return extractTar(tarPath, tempDirectoryProvider, progressEventDispatcherFactory);
+      }
+    };
+  }
+
+  private LocalImage extractTar(
```
briandealwis: Suggest rename to `handleDockerImageTar()`?  Since this does more than extract a tar file — and it's `docker-save` specific (not OCI images, for example).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -90,25 +91,58 @@ static boolean isGzipped(Path path) throws IOException {
...
     this.executorService = executorService;
     this.buildConfiguration = buildConfiguration;
-    this.tarPath = tarPath;
     this.progressEventDispatcherFactory = progressEventDispatcherFactory;
-    this.tempDirectoryProvider = tempDirectoryProvider;
   }
 
-  @Override
-  public LocalImage call()
+  Callable<LocalImage> processDockerDaemonBaseImageStep(DockerClient dockerClient) {
```
briandealwis: `loadDockerDaemonBaseImage`?  Or `loadImageFromDockerDaemonStep`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -234,35 +232,40 @@ private void authenticatePush() {
...
     Preconditions.checkArgument(dockerClient.isPresent());
-
-    results.tarPath =
+    ProgressEventDispatcher.Factory childProgressDispatcherFactory =
+        Verify.verifyNotNull(rootProgressDispatcher).newChildProducer();
+    Future<LocalImage> localImageFuture =
+        executorService.submit(
+            new LocalBaseImageSteps(
+                    executorService, buildConfiguration, childProgressDispatcherFactory)
+                .processDockerDaemonBaseImageStep(dockerClient.get()));
```
briandealwis: I like the centralization into a single class, but this code looks a bit unwieldy.  Make it a static?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
TadCordle: I think we should avoid "`load`" so we don't confuse it with `docker load`, would `saveDockerDaemonBaseImageStep` be better?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
TadCordle: Ah, the original reason it was non-static was to avoid copying the `executorService`, `buildConfiguration`, `childProgressDispatcherFactory` parameter list in every method signature. I'll see if I can do something about it though, this pattern is a bit strange.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -184,7 +225,8 @@ public LocalImage call()
     }
   }
 
-  private CachedLayer getCachedTarLayer(
+  private static CachedLayer getCachedTarLayer(
```
briandealwis: This method does something, so `get` doesn't seem right?  `cacheTarLayer()`?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -88,27 +89,66 @@ static boolean isGzipped(Path path) throws IOException {
...
+            tarPath,
+            tempDirectoryProvider,
+            progressEventDispatcherFactory);
+      }
+    };
   }
 
-  @Override
-  public LocalImage call()
+  private static LocalImage handleDockerImageTar(
```
briandealwis: `cacheDockerImageTar`? (`handle` could mean many things)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -88,27 +89,67 @@ static boolean isGzipped(Path path) throws IOException {
...
       ProgressEventDispatcher.Factory progressEventDispatcherFactory,
-      TempDirectoryProvider tempDirectoryProvider) {
-    this.executorService = executorService;
-    this.buildConfiguration = buildConfiguration;
-    this.tarPath = tarPath;
-    this.progressEventDispatcherFactory = progressEventDispatcherFactory;
-    this.tempDirectoryProvider = tempDirectoryProvider;
+      Path tarPath) {
+    return () -> {
+      try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
```
chanseokoh: Now it seems like you don't need to pass `tempDirectoryProvider`, because there's no more chaining between `Step`s. I know `retrieveDockerDaemonImageStep` has to use a temp dir provider, but I still think `cacheDockerImageTar` doesn't need to require passing one. It can create one inside it.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -157,14 +196,17 @@ public LocalImage call()
         for (int index = 0; index < layerFiles.size(); index++) {
           Path layerFile = destination.resolve(layerFiles.get(index));
           DescriptorDigest diffId = configurationTemplate.getLayerDiffId(index);
-          ProgressEventDispatcher.Factory progressEventDispatcherFactory =
+          ProgressEventDispatcher.Factory layerProgressFactory =
```
chanseokoh: Nit: How about `layerProgressDispatcherFactory`? "Progress factory" sounds weird.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -234,35 +232,35 @@ private void authenticatePush() {
...
-                        tempDirectoryProvider)
-                    .call());
+            LocalBaseImageSteps.retrieveTarImageStep(
+                executorService,
+                buildConfiguration,
+                childProgressDispatcherFactory,
+                tarPath.get())));
+  }
+
+  private void getImageAndLayers(Future<LocalImage> localImageFuture) {
```
chanseokoh: I'm trying to think of a better name for this helper method, since this sounds too general while is only for the base image. And what it does is simple retrieval. `assignLocalImageResult`?

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageStepsTest.java
```diff
@@ -67,22 +67,24 @@ public void setup() throws IOException {
     Mockito.when(progressEventDispatcher.newChildProducer()).thenReturn(childFactory);
     Mockito.when(childFactory.create(Mockito.anyString(), Mockito.anyLong()))
         .thenReturn(childDispatcher);
+    tempDirectoryProvider = new TempDirectoryProvider();
```
chanseokoh: I don't see any problem with

`private *final* TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider();`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -88,27 +89,67 @@ static boolean isGzipped(Path path) throws IOException {
...
+  static Callable<LocalImage> retrieveDockerDaemonImageStep(
+      ExecutorService executorService,
+      BuildConfiguration buildConfiguration,
+      ProgressEventDispatcher.Factory progressEventDispatcherFactory,
+      DockerClient dockerClient) {
+    return () -> {
+      ImageReference imageReference = buildConfiguration.getBaseImageConfiguration().getImage();
+      try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider();
+          ProgressEventDispatcher progressEventDispatcher =
+              progressEventDispatcherFactory.create("processing base image " + imageReference, 2)) {
```
chanseokoh: I think the outermost `try` can be merged with the next `try`.

And unrelated, but why the progress unit is 2? Doesn't seem like we dispatch an extra event manually to make it 2.

TadCordle: We create two child producers (one for docker save, one for `cacheDockerImageTar()`)

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageStepsTest.java
TadCordle: Since we're not passing a `TempDirectoryProvider` anymore, looks like we can just remove this.

PR #2048, opened 2019-10-08
Request source image access only if distinct from the base image

### LINKED ISSUE #2047: Allow to pull, modify and push using the same RegistryImage
**Environment**:

- *Jib version:* jib-core 0.11.0
- *Build tool:* maven 3.6.2
- *OS:* macos

**Description of the issue**:
At the moment it is not possible to use the same RegistryImage to pull, modify and push the modified image to a registry.  If I try that I get this error:
```Multiple entries with same key: <image/imagename>=pull and <image/imagename>=pull,push```

**Expected behavior**:
It would be very useful to be able to specify the same RegistryImage in `Jib.from` and in `Containerizer.to`

**Steps to reproduce**:
Write a method which does something like:
```
JibContainerBuilder jibContainerBuilder = Jib.from("<myRegistryImage>");
jibContainerBuilder.addLayer(Arrays.asList(Paths.get("random.txt")),  AbsoluteUnixPath.get("/")); 
jibContainerBuilder.containerize(Containerizer.to("<myRegistryImage>"));

```
**Log output**: 
```
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Multiple entries with same key: <myRegistryImage>=pull and <myRegistryImage>=pull,push                
```

### PR DESCRIPTION
**Pull request by briandealwis**
Fixes #2047

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointRequestProperties.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointRequestProperties.java
@@ -57,4 +57,11 @@ class RegistryEndpointRequestProperties {
   String getSourceImageName() {
     return sourceImageName;
   }
+
+  /**
+   * Return {@code true} if we have a source image and it is difference from the base image name.
+   */
+  public boolean hasDistinctSourceImageName() {
+    return sourceImageName != null && !imageName.equals(sourceImageName);
+  }
 }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointRequestProperties.java [hunk 1]
```diff
@@ -57,4 +57,11 @@ String getImageName() {
   String getSourceImageName() {
     return sourceImageName;
   }
+
+  /**
+   * Return {@code true} if we have a source image and it is difference from the base image name.
```
TadCordle:
```suggestion
   * Return {@code true} if we have a source image and it is different from the base image name.
```

briandealwis:
I hadn't, but just did and everything worked.

Issue #2034, opened 2019-10-02
query the image digest of a tag from the registry


aespinosa:
**Environment**:

- *Jib version:* jib-core 0.11.0
- *Build tool:* maven 3.6
- *OS:* macos

**Description of the issue**:

Looking at https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestPuller.java#L89-L94 the ManifestPuller doesn't read the response header of `Docker-Content-Digest` to get the sha256 digest you can use and pass to tools like `docker pull <image-name>@sha256:....` or `kubectl set image deployment/name container=<image-name>@sha256...` for CI/ CD purposes. 

**Expected behavior**:

A new method `getDigest()` from the puller

**Steps to reproduce**:
Run the following test:

```
public class RegistryTest {
    @Test
    public void registryDigest() throws IOException, RegistryException {
        RegistryClient.Factory factory = RegistryClient.factory(EventHandlers.NONE, "registry-1.docker.io", "library/centos");
        Authorization auth = factory.newRegistryClient().getRegistryAuthenticator().get().authenticatePull(null);
        factory.setAuthorization(auth);
        RegistryClient client = factory.newRegistryClient();

        V22ManifestListTemplate manifest = client.pullManifest("7", V22ManifestListTemplate.class);
        // If this gets stale, update by getting the output of
        // skopeo  inspect docker://centos:7 | jq '.Digest' -r
        Assertions.assertEquals("sha256:307835c385f656ec2e2fec602cf093224173c51119bbebd602c53c3653a3d6eb", manifest.getDigest());

        // These refer to other digests found in the manifest.json
//        Assertions.assertEquals("foo", manifest.getDigestsForPlatform("amd64", "linux"));
//        V22ManifestTemplate manifest = client.pullManifest("7", V22ManifestTemplate.class);
//        Assertions.assertEquals("foo", manifest.getContainerConfiguration().getDigest());
    }
}
```

**Log output**: 

none, the code doesn't compile

chanseokoh:
Hi @aespinosa,

Allowing to get the manifest digest makes sense. Actually, we were recently discussing some ways to get the hash, and we will probably re-compute the digest on our end instead of using `Docker-Content-Digest` (https://github.com/GoogleContainerTools/jib/issues/1884#issuecomment-537160441).

[...]

Issue #1884, opened 2019-08-01
Jib should report the base image digest that is used


briandealwis:
Jib should report the image digest used for the base image to simplify configuring a fixed digest, and also for traceability.

tejaskasetty:
Hi! Can I take this up?

chanseokoh:
Sure, thanks!

chanseokoh:
Presumably we could just read the [`Docker-Content-Digest` HTTP header](https://docs.docker.com/registry/spec/api/#pulling-an-image) in `ManifestPuller` and let the puller return the SHA along with the JSON manifest.

Alternatively we could compute the SHA on the fly inside `ManifestPuller` (with SHA256 for now).

However, @jonjohnsonjr sent us some links and proceed with caution with the header approach, as registries have implemented this so differently. Generally it's questionable to trust the header value.
- https://github.com/docker/distribution/issues/2395
- https://github.com/GoogleContainerTools/kaniko/issues/298 (Huawie registry doesn't prepend `sha256:` in the header value)
- https://github.com/knative/serving/issues/3997 / https://github.com/knative/serving/pull/3998

tejaskasetty:
So I shall work on getting the manifest from the registry first. This should resolve #2034 as well.

chanseokoh:
Thanks. Check my comments I made on your repo, although I am sure you already did. To fix #2034, maybe you have to change the return type of `ManifestPuller`. If that involves too many changes, I am OK with just printing the digest inside `ManifestPuller` as the first step.

tejaskasetty:
@chanseokoh There is one problem with computing the digest form the response directly. Manfiest has a signatures. And we compute the digest from the manifest without the signatures. And you had suggested to compute the digest directly from the inputStream.

jonjohnsonjr:
Only schema 1 images have signatures, which jib doesn't support (I believe).

chanseokoh:
@jonjohnsonjr Jib doesn't support pushing schema 1 images, but it supports pulling schema 1 base images.

@tejaskasetty I think we can ignore schema 1 manifests. They will be gone sooner or later. We are checking if the schema is 1 or 2 in `ManifestPuller` (I believe), so perhaps we should report the digest only when it is 2. And the digest of the schema 2 manifests will exactly be the SHA256 checksum of the blob content.

chanseokoh:
@tejaskasetty I merged #2052. Do you plan to continue on this?

[...]

PR #2052, opened 2019-10-10
Return both manifest and digest for pull manifest

tejaskasetty:
Pull manifest now will return the digest along with the manifest of the base image. This will be helpful for logging the base image digest #1884.
Fixes - #2034

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPullerIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPullerIntegrationTest.java
@@ -51,7 +51,7 @@ public class ManifestPullerIntegrationTest {
             .setAllowInsecureRegistries(true)
             .newRegistryClient();
     V21ManifestTemplate manifestTemplate =
-        registryClient.pullManifest("latest", V21ManifestTemplate.class);
+        registryClient.pullManifest("latest", V21ManifestTemplate.class).getManifest();
 
     Assert.assertEquals(1, manifestTemplate.getSchemaVersion());
     Assert.assertTrue(manifestTemplate.getFsLayers().size() > 0);
@@ -61,7 +61,7 @@ public class ManifestPullerIntegrationTest {
   public void testPull_v22() throws IOException, RegistryException {
     RegistryClient registryClient =
         RegistryClient.factory(EventHandlers.NONE, "gcr.io", "distroless/java").newRegistryClient();
-    ManifestTemplate manifestTemplate = registryClient.pullManifest("latest");
+    ManifestTemplate manifestTemplate = registryClient.pullManifest("latest").getManifest();
 
     Assert.assertEquals(2, manifestTemplate.getSchemaVersion());
     V22ManifestTemplate v22ManifestTemplate = (V22ManifestTemplate) manifestTemplate;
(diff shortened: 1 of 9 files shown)


tejaskasetty:
@googlebot I fixed it

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -199,28 +200,28 @@ private Image pullBaseImage(
       ProgressEventDispatcher progressEventDispatcher)
       throws IOException, RegistryException, LayerPropertyNotFoundException,
           LayerCountMismatchException, BadContainerConfigurationFormatException {
+    EventHandlers eventHandlers = buildConfiguration.getEventHandlers();
```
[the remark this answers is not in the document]
tejaskasetty: Actually I was working on #1884 as well where I will be logging the digest so I created it. That change has creeped into this commit.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPullerIntegrationTest.java
```diff
@@ -50,8 +51,9 @@ public void testPull_v21() throws IOException, RegistryException {
         RegistryClient.factory(EventHandlers.NONE, "localhost:5000", "busybox")
             .setAllowInsecureRegistries(true)
             .newRegistryClient();
-    V21ManifestTemplate manifestTemplate =
+    ManifestAndDigest<V21ManifestTemplate> manifestDigest =
         registryClient.pullManifest("latest", V21ManifestTemplate.class);
+    V21ManifestTemplate manifestTemplate = manifestDigest.getManifest();
```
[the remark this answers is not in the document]
tejaskasetty: Yes, I had left it that way cuz I had doubt. Should I cover any assertions on the digest. If not I will change it back to just get the manifest template.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ManifestAndDigest.java
```diff
@@ -0,0 +1,49 @@
+/*
```
chanseokoh: I think this class should be under `com/google/cloud/tools/jib/registry`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -216,11 +218,11 @@ private Image pullBaseImage(
           .dispatch(
               LogEvent.lifecycle(
                   "The base image reference is manifest list, searching for linux/amd64"));
-      manifestTemplate =
+      manifestAndDigest =
           obtainPlatformSpecificImageManifest(
               registryClient, (V22ManifestListTemplate) manifestTemplate);
+      manifestTemplate = manifestAndDigest.getManifest();
```
chanseokoh: Nit: this can just be `obtainPlatformSpecificImageManifest(...).getManifest()`. I think we will report the digest of the manifest list in this case. This also avoids reusing the `manifestAndDigest` variable.

chanseokoh:
BTW, you said the computed digest is incorrect for schema 1 manifests. I didn't really get why and what exactly is causing an issue in that case. Could you elaborate on this? Is this because of "signatures"? Never heard of this "signature" thing. And, if so, probably we should ignore and not report the computed digest for schema 1 manifests? Did I get it right?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
tejaskasetty: Yes, we will actually report the digest the switch case following this block under case 2. Since, I was working on both of them together. I initially had the digest log code as well. But later removed it to keep it different in separate PR. Hence, the `manifestAndDigest` variable

tejaskasetty:
Yes, exactly.

PR #2054, opened 2019-10-10
Add extra tags to tar image

### LINKED ISSUE #2043: Add the tags to the manifest.json
**Environment**:

- *Jib version:* 1.6.0
- *Build tool:* maven 3.6.0
- *OS:* Windows 10

**Description of the issue**:
The `<tags>` feature (specify multiple tags for the image) does not work for `buildTar`, as the `manifest.json` only contains `"repoTags": [ "image:latest"]`. It would be very useful to push the tags to this metadata. 

I think I could add it to the library, but first, is this desirable? To have all the tags in the tar? Currently it is done as an upload of the image and then tagging it multiple times (or uploading it multiple times with different tags for remote repositories). I think it does, so that the tar is self-contained.

**Expected behavior**:

When you define multiple tags, they end up in the tar, not as extra calls to the docker client/docker registry.

**Steps to reproduce**:

  1. on a random project `mvn clean compile jib:buildTar -Djib.to.tags=extr`
  2. open the `jib-image.tar/manifest.json` and see it only contains the  ":latest" tag

TadCordle:
Fixes #2043.

Adds `jib.to.tags` to tars built with jib so `docker load`ing the tar will tag the image appropriately. This also removes the need for `DockerClient#tag`, since it's done automatically now.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
@@ -23,7 +23,6 @@ import com.google.common.collect.Multimap;
 import com.google.common.io.CharStreams;
 import java.io.File;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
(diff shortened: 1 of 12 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
```diff
@@ -41,17 +42,17 @@
...
 
   /**
    * Instantiate with an {@link Image}.
    *
    * @param image the image to convert into a tarball
-   * @param imageReference image reference to set in the manifest
+   * @param buildConfiguration used for setting target image information in manifest
    */
-  public ImageTarball(Image image, ImageReference imageReference) {
+  public ImageTarball(Image image, BuildConfiguration buildConfiguration) {
```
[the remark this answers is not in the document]
TadCordle: Yeah, I was thinking about this too today. I think we can just pass in the target image and tags.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
```diff
@@ -111,7 +111,7 @@ public void testTarballStructure() throws IOException {
 
   @Test
   public void testManifest() throws IOException {
-    try (InputStream input = Files.newInputStream(imageTar.toPath())) {
+    try (InputStream ignored = Files.newInputStream(imageTar.toPath())) {
```
loosebazooka: Can this just be removed? Or can we at least explain why we're doing this?

TadCordle: Good point

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
```diff
@@ -42,16 +43,20 @@
 
   private final Image image;
   private final ImageReference imageReference;
+  private final ImmutableSet<String> allTargetImageTags;
 
   /**
    * Instantiate with an {@link Image}.
    *
    * @param image the image to convert into a tarball
    * @param imageReference image reference to set in the manifest
```
chanseokoh: I'd like to make it clear that the tag portion of this is ignored, and you'll get nothing if you pass an empty set for `allTargetImageTags`.

Or, we could accept `additionalTargetImageTags` instead, and do

```java
this.allTargetImageTags = new Set of imageReference.getTag() + additionalTargetImageTags
```

jib-core/src/main/java/com/google/cloud/tools/jib/docker/json/DockerManifestEntryTemplate.java
```diff
@@ -48,15 +48,15 @@
...
+  private final List<String> repoTags = new ArrayList<>();
   private final List<String> layers = new ArrayList<>();
 
   public void setConfig(String config) {
     this.config = config;
   }
 
-  public void setRepoTags(String repoTags) {
-    this.repoTags = Collections.singletonList(repoTags);
+  public void addRepoTag(String repoTag) {
```
chanseokoh: Oh, sorry I missed this part. Isn't `setRepoTags` required for deserialization? Do we deserialize a Docker manifest somewhere, possibly for the Docker base image feature? We should check if tags are properly populated in that case?

chanseokoh: Hopefully jackson can smartly make use of `addRepoTag`, but I think we need to confirm (if we are deserializing this field to make use of it).

TadCordle: No, I don't think getters and setters affect de/serialization if we use `JsonTemplate`. Only fields.

```
@JsonInclude(JsonInclude.Include.NON_NULL)
@JsonAutoDetect(
    fieldVisibility = JsonAutoDetect.Visibility.ANY,
    getterVisibility = JsonAutoDetect.Visibility.NONE,
    setterVisibility = JsonAutoDetect.Visibility.NONE,
    creatorVisibility = JsonAutoDetect.Visibility.NONE)
public interface JsonTemplate {}
```

TadCordle: Note that we do the same thing with layer files (i.e. `addLayerFile`, not `setLayerFiles`)

PR #2051, opened 2019-10-09
Add in syncmap json templates

loosebazooka:
Adds in the syncMap type that we will use to provide sync information to skaffold.

PR #2055, opened 2019-10-10
Add in maven sync map generation

loosebazooka:
Looking for a sanity check here as I mess with things. Not sure where I should be adding methods and whatever.

Part of #2097 

~~includes code from #2051 (so review that first)~~

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
@@ -1,0 +1,63 @@
+package com.google.cloud.tools.jib.maven.skaffold;
+
+import com.google.cloud.tools.jib.filesystem.TempDirectoryProvider;
+import com.google.cloud.tools.jib.maven.JibPluginConfiguration;
+import com.google.cloud.tools.jib.maven.MavenProjectProperties;
+import com.google.cloud.tools.jib.maven.MavenRawConfiguration;
+import com.google.cloud.tools.jib.maven.MojoCommon;
+import com.google.cloud.tools.jib.plugins.common.ContainerizingMode;
+import com.google.cloud.tools.jib.plugins.common.InvalidContainerizingModeException;
+import com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor;
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.ResolutionScope;
+
+@Mojo(
+    name = SyncMapMojo.GOAL_NAME,
+    requiresDependencyCollection = ResolutionScope.COMPILE_PLUS_RUNTIME)
+public class SyncMapMojo extends JibPluginConfiguration {
+
+  @VisibleForTesting static final String GOAL_NAME = "_skaffold-sync-map";
+
+  @Override
+  public void execute() throws MojoExecutionException {
+    checkJibVersion();
+    if (MojoCommon.shouldSkipJibExecution(this)) {
+      return;
+    }
+
+    // add check that means this is only for jars
+    if (!"jar".equals(getProject().getPackaging())) {
+      throw new MojoExecutionException("Sync is currently only available for jar style projects");
+    }
+    // add check for exploded containerization
+    try {
+      if (!ContainerizingMode.EXPLODED.equals(ContainerizingMode.from(getContainerizingMode()))) {
+        throw new MojoExecutionException("Sync is only available in exploded mode");
+      }
+    } catch (InvalidContainerizingModeException ex) {
+      throw new MojoExecutionException("Invalid containerizing mode", ex);
+    }
+
+    try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
+      MavenProjectProperties projectProperties =
+          MavenProjectProperties.getForProject(
+              getProject(), getSession(), getLog(), tempDirectoryProvider);
+
+      MavenRawConfiguration configuration = new MavenRawConfiguration(this);
+
+      try {
+        // should I print something out here to ID it? sync map might require that "package" is
+        // also run next to in the multimodule case to identify submodule dependencies
+        // print BEGIN JIB SYNC MAP
+        String syncMapJson =
+            PluginConfigurationProcessor.getSkaffoldSyncMap(configuration, projectProperties);
+        System.out.println(syncMapJson);
+        // print END JIB SYNC MAP
+      } catch (Exception e) {
+        throw new MojoExecutionException("Failed to generate Jib file map", e);
+      }
+    }
+  }
+}
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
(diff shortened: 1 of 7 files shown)


loosebazooka:
Needed to make some things public, and add in some other noOp style things. <-- this feels less than acceptable but would otherwise require decoupling some pretty major components.

loosebazooka:
I think this is ready for review again.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
```diff
@@ -0,0 +1,62 @@
+package com.google.cloud.tools.jib.maven.skaffold;
```
chanseokoh: copyright

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public void execute() throws MojoExecutionException {
+    checkJibVersion();
+    if (MojoCommon.shouldSkipJibExecution(this)) {
+      return;
+    }
+
+    // add check that means this is only for jars
+    if (!"jar".equals(getProject().getPackaging())) {
+      throw new MojoExecutionException("Sync is currently only available for jar style projects");
```
chanseokoh: If an error ever happens with Skaffold (and someone will hit some problem), I'd like to make the exception/log message clearly convey who was at fault and where exactly.

`"smart(?) Skaffold sync is currently only available for jar style Jib projects, but the packaging of " + project.getArtifactId() (or path or whatever) + " is " + project.getPackaging()`

And should we fall back to just not doing advanced sync when these kinds of checks fail? For example, they could have configured manual sync for some extra directory files, and they would continue to work regardless of whether or not it is JAR project or the exploded mode after we log a warning? (But then, I guess this smart sync is incompatible with other sync modes and this MOJO is called only when Skaffold determined it would to the smart sync? But even in that case, we may just give up syncing after printing a warning instead of just blocking using Skaffold? I don't know really.)

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+    // add check for exploded containerization
+    try {
+      if (!ContainerizingMode.EXPLODED.equals(ContainerizingMode.from(getContainerizingMode()))) {
+        throw new MojoExecutionException("Sync is only available in exploded mode");
```
chanseokoh: Likewise, it'd be good to make it clear the error is at the Jib level and output the project name (or path). And make the reader easily realize what the "exploded mode" here means.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojoTest.java
```diff
@@ -0,0 +1,108 @@
+package com.google.cloud.tools.jib.maven.skaffold;
```
chanseokoh: copyright

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
```diff
@@ -0,0 +1,62 @@
...
+
+      try {
+        String syncMapJson =
+            PluginConfigurationProcessor.getSkaffoldSyncMap(configuration, projectProperties);
+
+        System.out.println("\nBEGIN JIB JSON");
+        System.out.println(syncMapJson);
+
+      } catch (Exception e) {
+        throw new MojoExecutionException("Failed to generate Jib file map", e);
```
chanseokoh: Likewise, something more informative that will increase the likelihood the reader can understand the error without spending too much time and potentially fix their project (if possible).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldSyncMapTemplate.java
```diff
@@ -6,6 +6,7 @@
 import com.google.cloud.tools.jib.api.LayerEntry;
```
chanseokoh: This one doesn't have copyright either.

Personally I would avoid coupling `LayerEntry` for `JsonTemplate` classes, as a low-level class for a JSON template and just for that. The template thing is at a very low level in the hierarchy. But I don't feel strongly about this, and this certainly isn't to be address in this PR.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
```diff
@@ -0,0 +1,63 @@
...
+      MavenRawConfiguration configuration = new MavenRawConfiguration(this);
+
+      try {
+        // should I print something out here to ID it? sync map might require that "package" is
+        // also run next to in the multimodule case to identify submodule dependencies
+        // print BEGIN JIB SYNC MAP
+        String syncMapJson =
+            PluginConfigurationProcessor.getSkaffoldSyncMap(configuration, projectProperties);
+        System.out.println(syncMapJson);
+        // print END JIB SYNC MAP
```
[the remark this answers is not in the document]
TadCordle: :man_shrugging:

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojo.java
loosebazooka: I think our base case is that we only allow one type of sync for now. But I'll update the message.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/SkaffoldSyncMapTemplate.java
loosebazooka: It's only used to populate it, it's not actually part of the json template. Merely a convenience.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/SyncMapMojoTest.java
```diff
@@ -0,0 +1,160 @@
...
+    assertFilePaths(
+        projectRoot.resolve("complex-service/target/classes/com/test/HelloWorld.class"),
+        AbsoluteUnixPath.get("/app/classes/com/test/HelloWorld.class"),
+        generated.get(1));
+
+    List<FileTemplate> direct = parsed.getDirect();
+    Assert.assertEquals(1, direct.size());
+    assertFilePaths(
+        m2.resolve(
+            "com/google/cloud/tools/tiny-test-lib/0.0.1-SNAPSHOT/tiny-test-lib-0.0.1-SNAPSHOT.jar"),
```
chanseokoh: Interesting. `runBuild()` does `mvn package` only, and I know this custom-built JAR is under `jib-maven-plugin/build/resources/test/maven/projects/multi/complex-service/local-m2-repo/com/google/cloud/tools/tiny-test-lib`. Who installs this into `<user.home>/.m2/repository`? I just can't understand how this ended up in the user's m2 repo.

And does this mean that we will sync a SNAPSHOT dependency (which is not a project dependency) too? When the SNAPSHOT dependency in a local Maven repo is updated?

loosebazooka: Honestly, I think I was on autopilot when I was testing this, I have no idea how this works. Lemme go look at it again.

loosebazooka: So it turns out that maven treats the the local-m2-repo as a remote repository because we define it as:
```
  <repositories>
    <repository>
      <id>local-m2-repo</id>
      <url>file:${project.basedir}/local-m2-repo</url>
    </repository>
  </repositories>
```

and so it's copied into the .m2 when the build happens from the remote "local-m2-repo", I'll clean this up in a followup.

PR #2057, opened 2019-10-11
Cleanup code

TadCordle:
Cleaning up some stuff.

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
@@ -144,7 +144,7 @@ public class CredentialRetrieverFactoryTest {
     Assert.assertFalse(
         credentialRetrieverFactory.googleApplicationDefaultCredentials().retrieve().isPresent());
 
-    Mockito.verifyNoInteractions(mockLogger);
+    Mockito.verifyNoMoreInteractions(mockLogger);
   }
 
   @Test
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
@@ -144,7 +144,7 @@ public class CredentialRetrieverFactoryTest {
     Assert.assertFalse(
         credentialRetrieverFactory.googleApplicationDefaultCredentials().retrieve().isPresent());
 
-    Mockito.verifyZeroInteractions(mockLogger);
+    Mockito.verifyNoMoreInteractions(mockLogger);
   }
 
   @Test
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
@@ -144,7 +144,7 @@ public class CredentialRetrieverFactoryTest {
     Assert.assertFalse(
         credentialRetrieverFactory.googleApplicationDefaultCredentials().retrieve().isPresent());
 
-    Mockito.verifyNoMoreInteractions(mockLogger);
+    Mockito.verifyNoInteractions(mockLogger);
   }
 
   @Test
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryErrorExceptionBuilderTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryErrorExceptionBuilderTest.java
@@ -33,19 +33,19 @@ public class RegistryErrorExceptionBuilderTest {
   @Test
   public void testAddErrorEntry() {
     RegistryErrorExceptionBuilder builder =
-        new RegistryErrorExceptionBuilder("do something", mockHttpResponseException);
-
-    builder.addReason(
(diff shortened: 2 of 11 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
```diff
@@ -144,7 +144,7 @@ public void testGoogleApplicationDefaultCredentials_notGoogleContainerRegistry()
     Assert.assertFalse(
         credentialRetrieverFactory.googleApplicationDefaultCredentials().retrieve().isPresent());
 
-    Mockito.verifyZeroInteractions(mockLogger);
+    Mockito.verifyNoMoreInteractions(mockLogger);
```
chanseokoh: Is this necessary? I think `ZeroInteractions` are stricter to ensure there was no interaction with `mockLogger`, while `NoMoreInteractions` allows interactions up to this call.

TadCordle:
@briandealwis I changed it because apparently `verifyZeroInteractions` is deprecated.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryErrorExceptionBuilderTest.java
```diff
@@ -33,19 +33,19 @@
...
-        new ErrorEntryTemplate(ErrorCodes.MANIFEST_UNKNOWN.name(), "manifest unknown"));
-    builder.addReason(new ErrorEntryTemplate(ErrorCodes.TAG_INVALID.name(), "tag invalid"));
-    builder.addReason(
-        new ErrorEntryTemplate(ErrorCodes.MANIFEST_UNVERIFIED.name(), "manifest unverified"));
-    builder.addReason(
-        new ErrorEntryTemplate(ErrorCodes.UNSUPPORTED.name(), "some other error happened"));
-    builder.addReason(new ErrorEntryTemplate("unknown", "some unknown error happened"));
+        new RegistryErrorExceptionBuilder("do something", mockHttpResponseException)
+            .addReason(
+                new ErrorEntryTemplate(ErrorCodes.MANIFEST_INVALID.name(), "manifest invalid"))
```
chanseokoh: Yeah, and this is what the TODO was saying. `ErrorEntryTemplate` is meant for representing a JSON. Here, the code is just re-purposing an existing JSON template only to be able to pass a code and a string conveniently. No JSON needs to be constructed really.

But I think for this cleanup, we are OK leave these as-is, as long as we leave the TODO I just mentioned.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -230,7 +230,7 @@ public void testEntrypoint()
     Assert.assertEquals(
         Arrays.asList("custom", "entrypoint"),
         buildConfiguration.getContainerConfiguration().getEntrypoint());
-    Mockito.verifyZeroInteractions(logger);
+    Mockito.verifyNoMoreInteractions(logger);
```
chanseokoh: ditto

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -288,7 +288,7 @@ public void testEntrypoint_defaultWarPackaging()
 
     Assert.assertNotNull(buildConfiguration.getContainerConfiguration());
     Assert.assertNull(buildConfiguration.getContainerConfiguration().getEntrypoint());
-    Mockito.verifyZeroInteractions(logger);
+    Mockito.verifyNoMoreInteractions(logger);
```
chanseokoh: ditto

jib-core/src/test/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactoryTest.java
chanseokoh: Ah, just saw the comment about the method being deprecated.

briandealwis:
Oh!  Though [the deprecation notice](https://static.javadoc.io/org.mockito/mockito-core/3.1.0/org/mockito/Mockito.html#verifyZeroInteractions-java.lang.Object...-) says to use `verifyNoInteractions()` not `verifyNo*More*Interactions()`.  That said, it also says that `verifyZeroInteractions()` is equivalent to `verifyNoMoreInteractions()` (!!).

chanseokoh:
Reading the deprecation notice, I think we should use `verifyNoInteractions` instead of `verifyNoMoreInteractions`. I can imagine there's a possibility that `verifyZeroInteractions` == `verifyNoMoreInteractions` (misleading and that might be the reason it's deprecated) and `verifyNoMoreInteractions` != `verifyNoInteractions`.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryErrorExceptionBuilderTest.java
TadCordle: Makes sense.

chanseokoh:
Oh, but I see that the tests could really be meaning `verifyNoMoreInteractions` instead of "zero interactions". In that case, the change makes sense.

TadCordle:
Haha I'll change it back.

briandealwis:
The tests passed on Travis with `verifyNoInteractions()`.  I'd be in favour of keeping them.

PR #2063, opened 2019-10-14
Report base image digest

tejaskasetty:
Base image digest is reported/logged only when the manifest schema version is 2.
Fixes - #1884

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -201,30 +201,29 @@ class PullBaseImageStep implements Callable<ImageAndAuthorization> {
       ProgressEventDispatcher progressEventDispatcher)
       throws IOException, RegistryException, LayerPropertyNotFoundException,
           LayerCountMismatchException, BadContainerConfigurationFormatException {
+    EventHandlers eventHandlers = buildConfiguration.getEventHandlers();
     RegistryClient registryClient =
         buildConfiguration
             .newBaseImageRegistryClientFactory()
             .setAuthorization(registryAuthorization)
             .newRegistryClient();
 
-    ManifestTemplate manifestTemplate =
-        registryClient
-            .pullManifest(buildConfiguration.getBaseImageConfiguration().getImageTag())
-            .getManifest();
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -201,30 +201,29 @@ private Image pullBaseImage(
...
-      manifestTemplate =
+      eventHandlers.dispatch(
+          LogEvent.lifecycle(
+              "The base image reference is manifest list, searching for linux/amd64"));
+      manifestAndDigest =
           obtainPlatformSpecificImageManifest(
-                  registryClient, (V22ManifestListTemplate) manifestTemplate)
-              .getManifest();
+              registryClient, (V22ManifestListTemplate) manifestTemplate);
+      manifestTemplate = manifestAndDigest.getManifest();
```
chanseokoh: In the manifest list case, I have been thinking it makes sense to report the SHA of the manifest list. But now I think the digest of the actual image will be more helpful. Or, we could log both of them. @GoogleContainerTools/java-tools-build thoughts?

tejaskasetty: The digest of the actual image will be more helpful. Logging both would be fine but i don't see any advantage in knowing the SHA of manifest list.

chanseokoh: Yeah, I think the image digest would be helpful.

That said, let's do this. We log the image digest as LIFECYCLE to display it to the user, and in addition to this, if the base image reference was a manifest list, log the manifest list digest as INFO. That will help the user and us easily determine if it was a manifest list by looking at a log when there's a problem. Makes sense?

tejaskasetty: There is already a log as LIFECYCLE informing the user that the reference is manifest list. Should I follow it up with INFO log having the manifest list digest?

chanseokoh: Ah, we already inform that. Then I think this is already good enough.

PR #2068, opened 2019-10-16
Fix tar base image failing with missing explicit directory entries

TadCordle:
Fixes #2067

PR #2069, opened 2019-10-16
Maven: explode final WAR instead of using intermediate exploded WAR directory

### LINKED ISSUE #1091: Explode WAR instead of working on exploded WAR directory for Maven
Explode the WAR file rather than using this directory. The contents of the final WAR may be different from this directory (it's possible to include or exclude files when packaging a WAR. Also the exploded WAR directory is configurable with `<webappDirectory>` and may not be at `build.getFinalName()`.

https://github.com/GoogleContainerTools/jib/pull/1068/files#r221696396

chanseokoh:
Fixes #1091.

I will have to see if packaging a WAR file in tests works on Windows.


chanseokoh:
Yeah, as I feared, test code has problems on Windows.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -187,8 +201,10 @@ public JibContainerBuilder createJibContainerBuilder(
       throws IOException {
     try {
       if (isWarProject()) {
-        Path explodedWarPath =
-            Paths.get(project.getBuild().getDirectory(), project.getBuild().getFinalName());
+        Build build = project.getBuild();
+        Path war = Paths.get(build.getDirectory(), build.getFinalName() + ".war");
```
TadCordle: Do we need to check if this file exists?

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenProjectPropertiesTest.java
```diff
@@ -168,6 +175,32 @@ private static Path getResource(String path) throws URISyntaxException {
...
   }
 
+  private static Path zipUpDirectory(Path sourceRoot, Path targetZip) throws IOException {
+    try (ZipOutputStream zipOut = new ZipOutputStream(Files.newOutputStream(targetZip));
+        Stream<Path> stream = Files.walk(sourceRoot)) {
+      for (Iterator<Path> iterator = stream.iterator(); iterator.hasNext(); ) {
+        Path source = iterator.next();
+        if (source.equals(sourceRoot)) {
+          continue;
+        }
```
TadCordle: `DirectoryWalker` is an option to reduce the code here (we use it elsewhere in the plugins, although it is technically internal to jib-core). It'd be something like

```
for (Path source : new DirectoryWalker(sourceRoot).filterRoot().walk()) {
  ...
}
```

chanseokoh: Good catch. Very nice!

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: Somehow I couldn't make a comment here. Copying it over from below:

For checking the file, I think we're good. The same happens for the intermediate exploded WAR directory. We catch IOException and suggest to do mvn package.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -187,8 +201,10 @@ public JibContainerBuilder createJibContainerBuilder(
...
+        Path explodedWarPath = tempDirectoryProvider.newDirectory();
+        ZipUtil.unzip(war, explodedWarPath);
```
TadCordle: Sort of unrelated to this PR: I haven't tested, but I noticed that the unzip logic is very similar to the tar extraction logic. I don't really know anything about the zip file format, but I wonder if the fix [here](https://github.com/GoogleContainerTools/jib/pull/2068) should also be applied to `ZipUtil.unzip()`.

chanseokoh: I think that is a very good idea.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenProjectPropertiesTest.java
```diff
@@ -168,6 +176,29 @@ private static Path getResource(String path) throws URISyntaxException {
     return Paths.get(Resources.getResource(path).toURI());
   }
 
+  private static Path zipUpDirectory(Path sourceRoot, Path targetZip) throws IOException {
+    try (ZipOutputStream zipOut = new ZipOutputStream(Files.newOutputStream(targetZip));
+        Stream<Path> stream = Files.walk(sourceRoot)) {
```
TadCordle: Oh sorry almost forgot, I don't think we need this `Stream<Path> stream = Files.walk(sourceRoot)` anymore.

chanseokoh: Nice catch. Not feeling well today. Should probably go home early.

chanseokoh:
Yeah, I tend to forget that recently. :laughing:

PR #2073, opened 2019-10-17
Advanced configuration avoidance (jib-gradle-plugin)

### LINKED ISSUE #2072: Jib Gradle plugin creating tasks even when not run
**Environment**:

- *Jib version:* v1.6.2-SNAPSHOT >
- *Build tool:* Gradle 5.2.1
- *OS:* MacOS

**Description of the issue**:
When running `gradle help` and you use the `jib-gradle-plugin` all the jib related tasks are created during configuration instead during execution.

image

**Expected behavior**:
Running `gradle help` should not create any jib related task.

**Steps to reproduce**:

  1. gradle help

slowr:
Avoid creating jib tasks (also dependents like jar, assemble, war, etc) when not needed.

Before,
image

After,
image

Resolves #2072

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
@@ -40,19 +40,18 @@ class TaskCommon {
 
   @Nullable
   static TaskProvider<Task> getWarProviderTask(Project project) {
-
     if (!project.getPlugins().hasPlugin(WarPlugin.class)) {
       return null;
     }
 
-    try {
-      if (project.getPlugins().hasPlugin("org.springframework.boot")) {
+    if (project.getPlugins().hasPlugin("org.springframework.boot")) {
+      try {
         return project.getTasks().named("bootWar");
+      } catch (UnknownTaskException ignored) {
       }
-      return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
-    } catch (UnknownTaskException ignored) {
-      return null;
     }
+
+    return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
   }
 
   /** Disables annoying Apache HTTP client logging. */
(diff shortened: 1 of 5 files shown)


slowr:
It seems that current `GRADLE_MIN_VERSION` is set to `4.6` which does not support `TaskProvider`.

I bumped the minimum version to `4.9` which introduced `TaskProvider` but this seems to be a major change.

loosebazooka:
4.9 is quite old, I don't think we have to worry too much about it. We can bump that number

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -38,20 +39,20 @@
...
-    if (project.getPlugins().hasPlugin("org.springframework.boot")) {
-      Task bootWar = project.getTasks().findByName("bootWar");
-      if (bootWar != null) { // Spring Boot > 2.0
-        return (War) bootWar;
+    try {
+      if (project.getPlugins().hasPlugin("org.springframework.boot")) {
+        return project.getTasks().named("bootWar");
       }
+      return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
+    } catch (UnknownTaskException ignored) {
```
loosebazooka: I think the order is wrong here? catch should happen somewhere else? Or if we never intended to come here we should throw an illegal state exception or something.

Since we already checked if the `war-plugin` exists on the project, we should by default return `war` task.

chanseokoh: Ah, nice catch.

slowr: I just put the `named('war')` inside the try..catch block to make sure I always handle the exception it throws when the task is not there. 

In this case it should never be triggered _unless_ the war task name is changed so I wanted to be safe.

Should I move it outside of the try..catch block?

chanseokoh: If `named("bootWar")` can throws the exception, which I think is a real possibility, this method should return `named("war")` instead of null, shouldn't it?

slowr: oh I see what you mean.. in case that the project has `org.springframework.boot` plugin but not `bootWar` task then it will return null while there is a possibility that there is a `war` task as well (because of the warplugin check above). I will move the return outside the try..catch and remove the null return.

Issue #1868, opened 2019-07-24
Better way to handle invalid configuration values to show error messages


chanseokoh:
`PluginConfigurationProcessor` throws many exceptions.

```
      throws InvalidImageReferenceException, MainClassInferenceException, InvalidAppRootException,
          IOException, InvalidWorkingDirectoryException, InvalidContainerVolumeException,
          IncompatibleBaseImageJavaVersionException, NumberFormatException,
          InvalidContainerizingModeException, InvalidFilesModificationTimeException {
```

Many exceptions are only for syntactically invalid config values, but not all of them. Plugin code handles these exceptions in long catch blocks. Sometimes we use `HelpfulSuggestions` for error messages, but sometimes not. This long blocks are being copied over three tasks (build, dockerBuild, buildTar). I honestly don't know how we can best handle this.

<details>
<summary>details (click to unfold)</summary>

```java
    } catch (InvalidContainerizingModeException ex) {
      throw new MojoExecutionException(
          "invalid value for <containerizingMode>: " + ex.getInvalidContainerizingMode(), ex);

    } catch (InvalidWorkingDirectoryException ex) {
      throw new MojoExecutionException(
          "<container><workingDirectory> is not an absolute Unix-style path: "
              + ex.getInvalidPathValue(),
          ex);

    } catch (InvalidContainerVolumeException ex) {
      throw new MojoExecutionException(
          "<container><volumes> is not an absolute Unix-style path: " + ex.getInvalidVolume(), ex);

    } catch (InvalidFilesModificationTimeException ex) {
      throw new MojoExecutionException(
          "<container><filesModificationTime> should be an ISO 8601 date-time (see "
              + "DateTimeFormatter.ISO_DATE_TIME) or special keyword \"EPOCH_PLUS_SECOND\": "
              + ex.getInvalidFilesModificationTime(),
          ex);

    } catch (IncompatibleBaseImageJavaVersionException ex) {
      throw new MojoExecutionException(
          HelpfulSuggestions.forIncompatibleBaseImageJavaVesionForMaven(
              ex.getBaseImageMajorJavaVersion(), ex.getProjectMajorJavaVersion()),
          ex);

    } catch (InvalidImageReferenceException ex) {
      throw new MojoExecutionException(
          HelpfulSuggestions.forInvalidImageReference(ex.getInvalidReference()), ex);

    } catch (IOException | CacheDirectoryCreationException | MainClassInferenceException ex) {
[...]

PR #2083, opened 2019-10-18
Consolidate error handling in plugins

TadCordle:
An idea that makes #1868 less of a big issue. I don't think it fixes it completely, because `PluginConfigurationProcessor` still throws a ton of different exceptions, and the liberal use of `instanceof` feels hacky, but the code reduction here is pretty nice. Wondering if we should use this until we find a more solid solution for `PluginConfigurationProcessor` throwing so many different exceptions.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -16,22 +16,10 @@
 
 package com.google.cloud.tools.jib.gradle;
 
-import com.google.cloud.tools.jib.api.CacheDirectoryCreationException;
-import com.google.cloud.tools.jib.api.InvalidImageReferenceException;
 import com.google.cloud.tools.jib.docker.DockerClient;
-import com.google.cloud.tools.jib.plugins.common.BuildStepsExecutionException;
 import com.google.cloud.tools.jib.plugins.common.HelpfulSuggestions;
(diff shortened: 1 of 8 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -87,9 +75,7 @@ public void dockerClient(Action<? super DockerClientParameters> action) {
   }
 
   @TaskAction
-  public void buildDocker()
-      throws IOException, BuildStepsExecutionException, CacheDirectoryCreationException,
-          MainClassInferenceException {
+  public void buildDocker() {
```
chanseokoh: There are pros and cons, but to me, the downside of hiding each exception under the uber `Exception` seems to weigh more than the advantages. I would go with `buildDocker() throws IOException, BuildStepsExecutionException, ...` and have `handleJibException(<some function type> buildTask)` like

```java
  try {
    buildTask.run();
  } catch (InvalidAppRootException ex) {
    ...
```

so that whenever we introduce a new type of exception, we don't forget to handle it properly. Implementation-wise, passing a callable function is one way like above, but having an abstract base class could be another. But either of them will introduce some more code and that's the downside.

TadCordle: Yeah, good point. I think I'll close this and we can figure out a better long-term solution in the future.

PR #2089, opened 2019-10-21
Reproducible jars proposal

loosebazooka:
A proposal for handing reproducibility in mutli module/project builds.

Staged: https://github.com/GoogleContainerTools/jib/blob/proposal-reproducible-jar/proposals/archives/reproducible_jars.md


loosebazooka:
It looks like we're going to mostly skip this in favor of documentation. But I'll add in all the updates and we can leave this in the docs archive.

loosebazooka:
@TadCordle can you just take a final look over this before I merge?

PR #2090, opened 2019-10-23
Use artifact ID instead of project name in SkaffoldInitMojo

### LINKED ISSUE #2088: skaffold init sets `project:` to project name instead of module name
@ivanporty said Jib generated something like this with 1.5.1 for a Spring Boot project.

```yaml
    jib:
      project: Cloud Code Hello World
      args:
      - -Dimage=java-hello-world
```

After #1974 (>= 1.6.0), Jib no longer generates `project:` for Spring Boot projects that is effectively a single-module project, but I think it still generates it when needed, and it seems like it can contain a module/project name with whitespace.

We should also verify the Gradle side too.

TadCordle:
Towards #2088.

I *think* `project.name` is still the correct field to use with gradle, but I'm not 100% sure.


TadCordle:
It looks like gradle does allow spaces in the module name. I can have a module in a folder named "Complex Service", put `include ':Complex Service'` in the root `settings.gradle`, and run stuff like `gradle :Complex\ Service:build`.

chanseokoh:
Looks good then. The important thing alongside whitespace is that `project.name` is the right value to use on the command-line, and looks like it is.

PR #2092, opened 2019-10-23
Enhance TestWebServer and some cleanups

chanseokoh:
No production code is affected.

Some preparatory work ahead of the `allowInsecureRegistry` work.


jib-core/src/test/java/com/google/cloud/tools/jib/http/RequestWrapper.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.http;
+
+/** Helper to expose package-private methods. */
```
TadCordle: Any reason we can't just make the method public and mark with `@VisibleForTesting`?

chanseokoh: I often don't like adding a new method to production code that will be used only by testing, though it depends.

chanseokoh: WDYT? Although this case doesn't add a new method, making a method `public` has a significant implication likewise.

TadCordle: Hmm, I guess it's fine. Can always clean up later if we decide differently.

chanseokoh: Okay, at least this class replaces the more complex `MockConnection`, so it doesn't make the situation worse.

Issue #1316, opened 2018-12-07
Improve error logging in case of errors


smil2k:
**Description of the issue**:

This is the highest level of debugging verbosity I could achieve:
mvn -Dhttps.proxyHost=proxy -Dhttps.proxyPort=3128 compile jib:build -DjibSerialize=true -Djava.util.logging.config.file=logger.properties -X

logging.properties:
```
handlers = java.util.logging.ConsoleHandler
java.util.logging.ConsoleHandler.level=ALL

# CONFIG hides authentication data
# ALL includes authentication data
com.google.api.client.http.level=CONFIG
```

Even at this level I was not able to figure out the reason why the push into the repository did not succeeded. (See log.)

The actual problem was that in front of nexus3 there is was an nginx proxy which did not allow the request through. (Docker did return the cause of the problem:) 
```
5e93ace71c0b: Pushed 
ff3a3ba9afd4: Pushed 
8eefbd08e143: Pushing [==================================================>] 40.84 MB/40.84 MB
5b0bbc8b30cc: Pushing [==================================================>] 99.88 MB
6189abe095d5: Pushing [==================================================>] 1.966 MB
c5204564c844: Pushing [==================================================>] 16.82 MB/16.82 MB
error parsing HTTP 413 response body: invalid character '<' looking for beginning of value: "<html>\r\n<head><title>413 Request Entity Too Large</title></head>\r\n<body bgcolor=\"white\">\r\n<center><h1>413 Request Entity Too Large</h1></center>\r\n<hr><center>nginx/1.15.3</center>\r\n</body>\r\n</html>\r\n"
```

**Expected behavior**:

At least the http status or preferable the full HTTP request should be visible on the highest level of logging set. 

**Steps to reproduce**: 

Push anything greater than 50mb into a nexus3 using ingress in front.

**Environment**: 
Linux, Maven 3.6

**Log output**: 
```
[DEBUG] TIMED   Building container configuration : 2.0 ms
[DEBUG] TIMING  Pushing container configuration
[DEBUG] TIMING  Pushing BLOB digest: sha256:ebd874fd0c121bdab6f200b0c346c3d340a7664e23d8431ed04b261f06f08e5b, size: 1443
[DEBUG] BLOB : digest: sha256:ebd874fd0c121bdab6f200b0c346c3d340a7664e23d8431ed04b261f06f08e5b, size: 1443 already exists on registry
[...]

PR #2095, opened 2019-10-24
Refactoring and renaming

chanseokoh:
Trying to reduce the diff of the upcoming PR for changing Jib HTTP framework.

A lot of changes are renaming `httpResponseException`: I'll most likely introduce `ResponseException` that wraps `HttpResponseException` (somewhat similar like currently `Response` wrapping `HttpResponse` and `Request` wrapping `HttpRequest`), so changing variable names in advance.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/Request.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/Request.java
@@ -50,9 +50,7 @@ public class Request {
      * @return this
      */
     public Builder setAuthorization(@Nullable Authorization authorization) {
-      if (authorization != null) {
-        headers.setAuthorization(authorization.toString());
-      }
+      headers.setAuthorization(authorization == null ? null : authorization.toString());
       return this;
     }
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
@@ -146,24 +146,9 @@ class RegistryEndpointCaller<T> {
    * @throws RegistryException for known exceptions when interacting with the registry
    */
   T call() throws IOException, RegistryException {
-    try {
-      String apiRouteBase = "https://" + registryEndpointRequestProperties.getServerUrl() + "/v2/";
-      URL initialRequestUrl = registryEndpointProvider.getApiRoute(apiRouteBase);
-      return callWithAllowInsecureRegistryHandling(initialRequestUrl);
-
-    } catch (IOException ex) {
-      String registry = registryEndpointRequestProperties.getServerUrl();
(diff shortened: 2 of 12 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/http/Request.java
```diff
@@ -50,9 +50,7 @@ public Request build() {
      * @return this
      */
     public Builder setAuthorization(@Nullable Authorization authorization) {
-      if (authorization != null) {
-        headers.setAuthorization(authorization.toString());
-      }
+      headers.setAuthorization(authorization == null ? null : authorization.toString());
```
TadCordle: Probably intended, but this changes the behavior (if header's authorization is not null, setting `null` in this case sets header's authorization to null, but with the old behavior setting `null` would leave it unchanged). Is this fine?

chanseokoh: Good question. It changes the behavior, but there's no code where we call this more than once, so we are safe. And I think the new behavior makes more sense.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointCaller.java
```diff
@@ -265,70 +252,73 @@ private T call(URL url, Function<URL, Connection> connectionFactory)
...
 
         } else {
           // Unknown
-          throw httpResponseException;
+          throw responseException;
         }
       }
+
+    } catch (SSLException ex) {
+      throw ex;
```
loosebazooka: just wondering what this is about?

chanseokoh: So originally this `IOException` catch block was in this method. Then I moved it into another method in [this PR](https://github.com/GoogleContainerTools/jib/pull/1950/files#diff-1aa0b60da6c25b066e47abc6ebe10509) while adding logging for every `IOException`. The main reason I moved it back then was that almost all the time I shouldn't be logging errors for `SSLException` (which extends `IOException`), because `SSLException` as of now usually is expected for TLS failover. And since we are going to take out the TLS failover from this class, I now want to move this block back to its original place mostly for syntactically better-looking code. But to do so, I need to add this weird catch block to not log errors for `SSLException`.

But I just realized I should still need to log broken pipe in the `SSLException` case too after checking the [history](https://github.com/GoogleContainerTools/jib/issues/1316#issuecomment-478100424) behind it. I'll push up a small change.

chanseokoh: But once the TLS failover is decoupled from this class, I should be able to log errors for all `IOException` including `SSLException`, since `SSLException` in the new world will conclusively mean that the `SSLException` is final and you can no longer proceed.

Issue #2097, opened 2019-10-24
SyncMap output for Skaffold


loosebazooka:
Just a tracking bug

- [x] Add json templates (#2051)
- [x] Expose containerizing mode parser (#2098)
- [x] Share skip logic (#2102)
- [x] Fix test project naming (#2103)
- [x] Add intermediate build config api (#2115)
- [x] Fix travis (#2122)
- [x] Add maven goal (#2055)
- [x] Add gradle task (#2149)

PR #2098, opened 2019-10-24
Moving parsing code for containerizing mode

loosebazooka:
part of #2097

required for #2055

### DIFF
--- a/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ContainerizingModeTest.java
+++ b/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ContainerizingModeTest.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.jib.plugins.common;
 import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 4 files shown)


jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ContainerizingModeTest.java
```diff
@@ -0,0 +1,51 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.common;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ContainerizingModeTest {
```
chanseokoh: Nit: we always have a silly Javadoc for test classes.

loosebazooka: oh oops. Will add in.

loosebazooka: oh I see why you said "silly"

Issue #1923, opened 2019-08-21
Convert allowInsecureRegistries to list of registries instead of binary true/false 


chanseokoh:
I think this is a nice enhancement and useful for multi-module projects. It ensures only those registries are insecure, preventing an issue like https://github.com/GoogleContainerTools/skaffold/pull/2674#issuecomment-523497039.

Potentially it can accept a glob pattern. Or, maybe just a suffix for simplicity.

We also need to think about aliases.

briandealwis:
And since registries would be explicitly identified then we could automatically apply `sendCredentialsOverHttp` for those registries.

chanseokoh:
Or perhaps automatically set `sendCredentialsOverHttp` only for loopback addresses (localhost). AFAIK, OS by-passes a network stack entirely for loopback as optimization, so it's pretty safe.

JoeWang1127:
Closing as part of cleanup. Please re-open if it's still relevant

Issue #946, opened 2018-09-06
Optimize insecure registry failover


chanseokoh:
https://github.com/GoogleContainerTools/jib/pull/945#issuecomment-419225961

For example, some class could cache a list of hosts during the entire lifecycle of a build to remember which hosts required ignoring cert validation error or HTTP.

chanseokoh:
#1923 should be independent of this, but leaving a note to consider the effect anyway.

PR #2100, opened 2019-10-24
New HTTP framework to incorporate automatic insecure connection handling

### LINKED ISSUE #2074: -Djib.allowInsecureRegistries=true does not seems to work
**Environment**: Docker image:  gcr.io/cloud-builders/mvn

- *Jib version:* 1.6.1

- *Build tool: Maven

- *OS:* Docker Container

**Description of the issue**:

I am trying to use the `-Djib.allowInsecureRegistries=true` option to my tool and I am getting an error of this type when using it :

```
building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
```

I am using it while building a sample helloworld application https://github.com/che-samples/console-java-simple

**Expected behavior**:

Working

**Steps to reproduce**:

  1.  Checkout https://github.com/che-samples/console-java-simple
  2.  docker run -v it to /builder/home
  3. Run
```
    - mvn
    - compile
    - com.google.cloud.tools:jib-maven-plugin:build
    - -Duser.home=/builder/home
    - -Dimage=image-registry.openshift-image-registry.svc:5000/test/image
    - -Djib.allowInsecureRegistries=true
```
  
**Log output**: 

<details>
  <summary>Full log output</summary>
<pre>
Downloading from central: https://repo.maven.apache.org/maven2/com/google/cloud/tools/jib-maven-plugin/maven-metadata.xml
Progress (1): 1.5 kB
                    
Downloaded from central: https://repo.maven.apache.org/maven2/com/google/cloud/tools/jib-maven-plugin/maven-metadata.xml (1.5 kB at 2.5 kB/s)
Downloading from central: https://repo.maven.apache.org/maven2/com/google/cloud/tools/jib-maven-plugin/1.6.1/jib-maven-plugin-1.6.1.pom
Progress (1): 2.2/3.5 kB
Progress (1): 3.5 kB    

[...]

[INFO] Containerizing application to image-registry.openshift-image-registry.svc:5000/tekton-pipelines/console-java-simple...
[INFO] Retrieving registry credentials for image-registry.openshift-image-registry.svc:5000...
[INFO] Getting base image gcr.io/distroless/java:8...
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  16.607 s
[INFO] Finished at: 2019-10-17T14:02:14Z
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.6.1:build (default-cli) on project console-java-simple: Build image failed: Failed to authenticate with registry image-registry.openshift-image-registry.svc:5000/tekton-pipelines/console-java-simple because: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
</pre>
</details>

chanseokoh:
Introducing thread-safe `TlsFailoverHttpClient` (still named `Connection` for now) designed to be instantiated as a singleton stateful instance throughout the Jib build life cycle.

The TLS failover feature is decoupled from `RegistryEndpointCaller` and integrated into `TlsFailoverHttpClient`. The new client provides a complete abstraction for taking care of all possible TLS failover scenarios and controlling whether to send credentials over plain-HTTP.

`RegistryAuthenticator` will make use of the new client, so this will fix #2074.

In the future, this setup will allow

- Caching TLS failover results to take care of the performance issue (#946).
- Unified handling to implement the insecure server list (#1923).
- Reusing HttpTransport or HttpRequestFactory. Currently, `Connection` always creates a new HttpTransport for each HTTP call, creating a new connection pool each time. But this might be intended, I don't know. Not changing this behavior now.


jib-core/src/test/java/com/google/cloud/tools/jib/http/ConnectionTest.java
```diff
@@ -22,140 +22,375 @@
...
   }
 
-  private void testSend(String httpMethod, SendFunction sendFunction) throws IOException {
-    try (Connection connection = testConnection) {
-      sendFunction.send(connection, fakeRequest);
-    }
-
-    Mockito.verify(mockHttpRequest).setHeaders(httpHeadersArgumentCaptor.capture());
-    Mockito.verify(mockHttpResponse).disconnect();
+  private void testCall(String httpMethod, CallFunction callFunction) throws IOException {
```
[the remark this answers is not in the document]
chanseokoh: Changed the name to `verifyCall`. I will take care of the order in a followup. I'm just not changing the original order for easy review.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -17,68 +17,63 @@
...
-    addProxyCredentials(transport);
-    return url -> new Connection(url, transport);
+  private static HttpTransport getInsecureHttpTransport() {
+    try {
+      ApacheHttpTransport insecureTransport =
+          new ApacheHttpTransport.Builder().doNotValidateCertificate().build();
+      addProxyCredentials(insecureTransport);
+      return insecureTransport;
+    } catch (GeneralSecurityException ex) {
+      throw new RuntimeException("platform does not support TSL protocol", ex);
```
briandealwis: TSL &rarr; TLS

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -17,68 +17,63 @@
...
+    return "https".equals(url.getProtocol());
+  }
+
+  private static URL toHttp(URL url) {
+    GenericUrl httpUrl = new GenericUrl(url);
+    httpUrl.setScheme("http");
+    return httpUrl.toURL();
+  }
+
+  private static HttpTransport getHttpTransport() {
```
briandealwis: Given we have `getInsecureHttpTransport()`, would this make sense as `getSecureHttpTransport()`?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -114,87 +109,157 @@ private static void addProxyCredentials(ApacheHttpTransport transport, String pr
...
+    this(
+        enableHttpAndInsecureFailover,
+        sendAuthorizationOverHttp,
+        logger,
+        Connection::getHttpTransport,
+        Connection::getInsecureHttpTransport);
   }
 
-  @Override
-  public void close() throws IOException {
```
briandealwis: We should still have a `close()` or `shutDown()` to actually release resources (imagine the jib-core case).

chanseokoh: So that means this class should maintain a list of `Response` objects that are generated. It certainly is a nice feature as a safety net, but I think we should add that later considering the value and the new code to maintain the list. I'll file a new issue.

(Honestly, I feel comfortable making the user responsible for closing `Response`, which is the way Apache HttpClient works. I've seen IDEs warn if `AutoCloseable` is left open. However, it is true a method like `shutDown()` will make all `Response`s disconnect immediately.)

briandealwis: Hmm, well `HttpTransport` does have a `shutdown()` method, and it's overridden by `ApacheHttpTransport` which seems to do actually something.

chanseokoh: Ah, it's probably good to call `shutdown()` on `HttpTransport`. I'll make a note in #2111. At least this PR retains the current state of not calling `shutdown()` and doesn't make anything worse about this matter.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -114,87 +109,157 @@ private static void addProxyCredentials(ApacheHttpTransport transport, String pr
...
    * @param httpMethod the HTTP request method
+   * @param url endpoint URL
    * @param request the request to send
    * @return the response to the sent request
    * @throws IOException if building the HTTP request fails.
    */
-  public Response send(String httpMethod, Request request) throws IOException {
-    Preconditions.checkState(httpResponse == null, "Connection can send only one request");
+  public Response call(String httpMethod, URL url, Request request) throws IOException {
+    if (!isHttpsProtocol(url) && !enableHttpAndInsecureFailover) {
```
briandealwis: I think it's worth explicitly calling out that we're following the approach taken by Docker and always contacting HTTPS and then retrying as HTTP (if allowed).

Could we make the failure-over check be a method that takes the URL (`canFailOverToHttp(url)`?) so that we can swap out a different implementation (e.g., a table lookup populated from `.docker/daemon.json`) without changing this code?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -114,87 +109,157 @@ private static void addProxyCredentials(ApacheHttpTransport transport, String pr
...
+      }
+
+      try {
+        logInsecureHttpsFailover(url);
+        return call(httpMethod, url, request, insecureHttpTransportFactory.get());
+
+      } catch (SSLException ignored) { // This is usually when the server is plain-HTTP.
+        logHttpFailover(url);
+        return call(httpMethod, toHttp(url), request, secureHttpTransportFactory.get());
+      }
```
briandealwis: So the problem here (and it was there before) is that we lose the SSL exception which is useful for tracking problems.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
briandealwis: sgtm

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
briandealwis: It has made diagnosis of misconfigured registries issues difficult though.  I remember wondering why there was no PKIX exception being reported.  Maybe the right thing here is to have specific exceptions that records the cascade:
  - unable to reach via SSL: exception
  - unable to reach by plain HTTP: exception

And you know, perhaps if `enableHttpAndInsecureFailover == true`, then we should use the `insecureHttpTransportFactory` in the first place.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
```diff
@@ -156,9 +164,11 @@ private RegistryAuthenticator(
    * @param credential the credential used to authenticate
    * @return an {@code Authorization} authenticating the pull
    * @throws RegistryAuthenticationFailedException if authentication fails
+   * @throws RegistryCredentialsNotSentException if authentication is failed and credentials were
```
briandealwis: This seems an odd name.  I think something about _insecure_ matters here.  Maybe it should be a code on `RegistryAuthenticationFailedException`?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
chanseokoh: The second exception can never be a PKIX exception (that can be thrown when the server and the client can at least talk on TLS). Perhaps you are misremembering something. I don't recall any particular issue from this. And this `SSLException` almost all the time means the endpoint is not TLS except for serious I/O errors, I honestly think surfacing `SSLException` will cause more confusion most of the time. And this failover will happen a lot with a local HTTP Docker registry. The path is also similar with `ConnectException` when the port is not given. The first `ConnectException` is not recorded for the port 443.

And note we are recording logs when we failover, so I think we are covered. And I am going to increase it to warning when I cache an insecure registry list, and it will become very obvious.

For these reasons, I'd like to retain the current behavior. It was actually my reasoning when I first implemented this, but then the log level was decreased to INFO.

And for now, we should always try the secure transport first (if I understood what you said correctly); some registries may be secure and others may not.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -17,68 +17,80 @@
...
+import java.util.function.Supplier;
+import javax.net.ssl.SSLException;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.impl.client.DefaultHttpClient;
 
 /**
- * Sends an HTTP {@link Request} and stores the {@link Response}. Clients should not send more than
- * one request.
+ * Thread-safe HTTP client with the automatic insecure connection failover feature. Intended to be
```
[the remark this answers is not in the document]
chanseokoh: I expanded it further: "from secure HTTPS to insecure HTTPS or HTTP".

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryAuthenticator.java
chanseokoh: Yeah, I don't really like it's throwing an additional exception, but `RegistryCredentialsNotSentException` is an existing exception that we introduced specifically to identify the case where Jib got 401 (unauthorized) over plain-HTTP when `Authorization:` was not sent and then catch this to show a nice message that the user may want to use `-DsendCredentialsOverHttp`. There is code at an upper level that catches this exception. There may be better ways to restructure things for this, but I don't have a concrete idea now that will make it look better. :/

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryEndpointCallerTest.java
```diff
@@ -321,20 +166,13 @@ public void testCall_unauthorized() throws IOException, RegistryException {
...
-
-    Mockito.when(mockConnection.send(Mockito.eq("httpMethod"), Mockito.any()))
-        .thenThrow(Mockito.mock(SSLPeerUnverifiedException.class)) // server is not HTTPS
-        .thenThrow(new HttpResponseException(redirectResponse)) // redirect to HTTP
-        .thenThrow(new HttpResponseException(unauthroizedResponse)); // final response
-    Mockito.when(mockInsecureConnection.send(Mockito.eq("httpMethod"), Mockito.any()))
-        .thenThrow(Mockito.mock(SSLPeerUnverifiedException.class)); // server is not HTTPS
-
-    RegistryEndpointCaller<String> insecureEndpointCaller = createRegistryEndpointCaller(true, -1);
+    ResponseException unauthroizedException =
```
briandealwis:
```suggestion
    ResponseException unauthorizedException =
```

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryEndpointCallerTest.java
```diff
@@ -346,39 +184,20 @@ public void testCall_credentialsNotSentOverHttp() throws IOException, RegistryEx
...
-    HttpResponse redirectResponse = mockRedirectHttpResponse("http://newlocation");
-
-    Mockito.when(mockConnection.send(Mockito.eq("httpMethod"), Mockito.any()))
-        .thenThrow(Mockito.mock(SSLPeerUnverifiedException.class)) // server is not HTTPS
-        .thenThrow(new HttpResponseException(redirectResponse)) // redirect to HTTP
-        .thenReturn(mockResponse); // final response
-    Mockito.when(mockInsecureConnection.send(Mockito.eq("httpMethod"), Mockito.any()))
-        .thenThrow(Mockito.mock(SSLPeerUnverifiedException.class)); // server is not HTTPS
-
+    ResponseException unauthroizedException =
```
briandealwis:
```suggestion
    ResponseException unauthorizedException =
```

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
briandealwis: I was thinking some more and I agree with you: with the current allowInsecureRegistries=true/false, we need to do full verification and then failover to no-validate and then plain HTTP.  When we allow insecure-registries on a per-host basis, we can skip to just first disable cert validation.

(BTW: I was referring to the original `SSLException ex` from line 223.  But unless we have allowInsecureRegistries==true then it should be thrown…. Hmm.)

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -38,22 +38,32 @@
...
  *
  * <ul>
- *   <li>When port was given:
+ *   <li>When a port is provided (for example {@code my-registry:5000/my-repo}):
  *       <ol>
  *         <li>Attempts secure HTTPS.
- *         <li>If it fails due to {@link SSLException}, attempts insecure HTTPS.
- *         <li>If it fails again due to {@link SSLException}, attempts plain-HTTP.
+ *         <li>If (1) fails due to {@link SSLException}, attempts insecure HTTPS (disables
+ *             certificate validation).
```
briandealwis: Maybe _…re-attempts secure HTTPS on the specified port but disabling certificate validation_?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -38,22 +38,32 @@
...
+ *         <li>If (2) fails again due to {@link SSLException}, attempts plain-HTTP.
  *       </ol>
- *   <li>When port was not given: follows same execution path above, but additionally, if the very
- *       first secure HTTPS (at port 443) fails due to non-timeout {@link ConnectException},
- *       attempts plain-HTTP at port 80.
+ *   <li>When a port is not provided (for example {@code my-registry/my-repo}):
+ *       <ol>
+ *         <li>Attempts secure HTTPS.
+ *         <li>If (1) fails due to {@link SSLException}, attempts insecure HTTPS (disables
+ *             certificate validation).
```
briandealwis: _…re-attempts secure HTTPS on port 443, but disabling certificate validation_?

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -38,22 +38,32 @@
...
+ *         <li>If (2) fails again due to {@link SSLException}, attempts plain-HTTP.
  *       </ol>
- *   <li>When port was not given: follows same execution path above, but additionally, if the very
- *       first secure HTTPS (at port 443) fails due to non-timeout {@link ConnectException},
- *       attempts plain-HTTP at port 80.
+ *   <li>When a port is not provided (for example {@code my-registry/my-repo}):
+ *       <ol>
+ *         <li>Attempts secure HTTPS.
```
briandealwis: _…on port 443 (`https`)_

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -38,22 +38,32 @@
...
+ *         <li>If (1) fails due to {@link SSLException}, attempts insecure HTTPS (disables
+ *             certificate validation).
+ *         <li>If (2) fails again due to {@link SSLException}, attempts plain-HTTP.
+ *         <li>Or, if (1) fails due to non-timeout {@link ConnectException}, attempts plain-HTTP at
+ *             port 80.
+ *       </ol>
+ *       In sum, this follows the same execution path as when a port is provided, but additionally,
+ *       it can fall back to HTTP at port 80 when the very first secure HTTPS (at port 443) fails
+ *       when it cannot connect at all.
```
briandealwis: I found this text (and the previous text) left me wondering if we attempt plain HTTP on port 443.  I think being explicit about the ports in the actual steps is better, and then we can drop this text.

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -38,22 +38,32 @@
...
  * <ul>
- *   <li>When port was given:
+ *   <li>When a port is provided (for example {@code my-registry:5000/my-repo}):
  *       <ol>
  *         <li>Attempts secure HTTPS.
- *         <li>If it fails due to {@link SSLException}, attempts insecure HTTPS.
- *         <li>If it fails again due to {@link SSLException}, attempts plain-HTTP.
+ *         <li>If (1) fails due to {@link SSLException}, attempts insecure HTTPS (disables
+ *             certificate validation).
+ *         <li>If (2) fails again due to {@link SSLException}, attempts plain-HTTP.
```
briandealwis: _attempts plain-HTTP **on the specified port**_

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
```diff
@@ -38,22 +38,32 @@
...
+ * HTTP. Intended to be created once and shared to be called at multiple places. Callers should
+ * close the returned {@link Response}.
  *
  * <p>The failover (if enabled) in the following way:
  *
  * <ul>
- *   <li>When port was given:
+ *   <li>When a port is provided (for example {@code my-registry:5000/my-repo}):
  *       <ol>
  *         <li>Attempts secure HTTPS.
```
briandealwis: _HTTPS **on the specified port**_

jib-core/src/main/java/com/google/cloud/tools/jib/http/Connection.java
chanseokoh: Ah, yeah, you're right. That's the difference. We don't attempt plain HTTP on port 443 in this case. I've dropped the summary text.

jib-core/src/main/java/com/google/cloud/tools/jib/http/ResponseException.java
```diff
@@ -0,0 +1,51 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.http;
+
+import com.google.api.client.http.HttpHeaders;
+import com.google.api.client.http.HttpResponseException;
+import java.io.IOException;
+
+/** Holds an HTTP response exception. */
```
loosebazooka: Could help to explain what requestAuthorizationCleared is in here.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryEndpointProvider.java
```diff
@@ -53,15 +54,15 @@
...
    *
-   * @param responseException the {@link HttpResponseException} to handle
+   * @param responseException the {@link ResponseException} to handle
    * @throws HttpResponseException {@code responseException} if {@code responseException} could not
    *     be handled
    * @throws RegistryErrorException if there is an error with a remote registry
    */
-  default T handleHttpResponseException(HttpResponseException responseException)
-      throws HttpResponseException, RegistryErrorException {
+  default T handleHttpResponseException(ResponseException responseException)
```
loosebazooka: Not sure if we should have `default` if we fully control the workflow here... RegistryEndpointProvider is internal. Can we file a followup to cleanup this part?

chanseokoh: Filed #2117.

jib-core/src/test/java/com/google/cloud/tools/jib/http/ConnectionTest.java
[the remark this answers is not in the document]
chanseokoh: Method reordering taken care of in #2118.

PR #2107, opened 2019-10-28
Don't use bootWar if disabled

### LINKED ISSUE #2096: Check if bootWar/war tasks are enabled
**Environment**:

- *Jib version:* 1.7.0
- *Build tool:* gradle 5.2.1
- *OS:* linux

**Description of the issue**:
When a project enables the war task but disables the bootWar task, Jib incorrectly chooses the bootWar task to depend on, instead of the war task.

**Expected behavior**:
Jib should depend on enabled tasks only.

**Steps to reproduce**:

  1. apply plugins 'org.springframework.boot' and jib
  2. set `bootWar { enabled = false }` and `war { enabled = true }` in build.gradle
  3. gradle jib

**Log output**: 

**Additional Information**: 
https://github.com/GoogleContainerTools/jib/blob/4d8efc902d40945fe5ed296d3bf55f0d42119fef/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java#L167

https://github.com/GoogleContainerTools/jib/blob/4d8efc902d40945fe5ed296d3bf55f0d42119fef/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java#L49

chanseokoh:
Fixes #2096.

~~The only thing that may matter is that now `task.dependsOn(Object...)` takes a `List`, so I'll do some actual tests to see this works as desired.~~ Works fine.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
@@ -40,10 +40,10 @@ class TaskCommon {
 
   @Nullable
   static TaskProvider<Task> getWarTaskProvider(Project project) {
-    if (!project.getPlugins().hasPlugin(WarPlugin.class)) {
-      return null;
+    if (project.getPlugins().hasPlugin(WarPlugin.class)) {
+      return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
     }
-    return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
+    return null;
   }
 
   @Nullable
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
@@ -43,15 +43,18 @@ class TaskCommon {
     if (!project.getPlugins().hasPlugin(WarPlugin.class)) {
       return null;
     }
+    return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
+  }
 
+  @Nullable
+  static TaskProvider<Task> getBootWarTaskProvider(Project project) {
     if (project.getPlugins().hasPlugin("org.springframework.boot")) {
       try {
         return project.getTasks().named("bootWar");
       } catch (UnknownTaskException ignored) { // fall through
       }
     }
-
-    return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
+    return null;
   }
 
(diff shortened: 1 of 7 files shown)


jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -191,21 +187,22 @@ private static Path getResource(String path) throws URISyntaxException {
...
   @Mock private SourceSetContainer mockSourceSetContainer;
   @Mock private SourceSet mockMainSourceSet;
   @Mock private SourceSetOutput mockMainSourceSetOutput;
 
   @Mock(answer = Answers.RETURNS_DEEP_STUBS)
-  private TaskProvider<Task> mockWarTaskProvider;
+  private Project mockProject;
+
+  @Mock(answer = Answers.RETURNS_DEEP_STUBS)
+  private TaskProvider<Task> mockTaskProvider1;
```
TadCordle: Maybe instead of `mockTaskProvider1` and `mockTaskProvider2`, they should be `mockWarTaskProvider` and `mockBootWarTaskProvider`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -43,15 +43,18 @@
     if (!project.getPlugins().hasPlugin(WarPlugin.class)) {
       return null;
     }
+    return project.getTasks().named(WarPlugin.WAR_TASK_NAME);
+  }
 
+  @Nullable
+  static TaskProvider<Task> getBootWarTaskProvider(Project project) {
```
TadCordle: Can we make the logic between `getWarTaskProvider()` and `getBootWarTaskProvider()` consistent? e.g. `getWarTaskProvider()` starts with `if (!project.getPlugins...)` and `getBootWarTaskProvider()` starts with `if(project.getPlugins...)`.

chanseokoh: I feel comfortable with this. This is at least the original behavior. I think the WAR Plugin is an integral part of Gradle (applied with `id 'war'`), so it is unthinkable that it doesn't have the `war` task or will remove the task in the future. I think that simply won't happen. OTOH, the Spring Boot plugin is a third-party and we can exercise caution.

And nice for catching CHANGELOG. This keeps occurring to me :/

chanseokoh: Oh, sorry. You were talking about something else.

PR #2109, opened 2019-10-29
Remove deprecated jibDockerBuild.dockerClient config

### LINKED ISSUE #1983: Remove deprecated jibDockerBuild.dockerClient parameters
#1981 moves `dockerClient` configuration from the `jibDockerBuild` task to `JibExtension` to account for local base images, deprecating the original configuration.

TadCordle:
Fixes #1983.


chanseokoh:
Looks like this was deprecated in v1.6.0 released on Sep 13, so I think this is not old enough to take it out now.

TadCordle:
True, but I think even though it's only been a few months, 2.0 is a good time to do it.

chanseokoh:
That is true. If we don't do it for 2.0, it may have to wait for 3.0. @GoogleContainerTools/java-tools-build thoughts?

briandealwis:
My only reservation is that the deprecation warning is overwritten with our progress console.  It's only shown if we run with `-Djib.console=plain`.

But there's nothing stopping a user from staying on Jib 1.x, and we've done two releases since.  So I'm ok with the removal.

chanseokoh:
> deprecation warning is overwritten with our progress console. It's only shown if we run with `-Djib.console=plain`.

Really? Didn't know. We should have filed a bug for this.

briandealwis:
I only discovered it just today out of curiosity to see the effect of the deprecation message.  It does show up for a bit though, and it will show on CI/CD builds though.

PR #2113, opened 2019-10-31
remove deprecation warnings

### LINKED ISSUE #2086: Issues with Gradle 6.0-rc-1
**Environment**:

- *Jib version:* 1.7.0
- *Build tool:* Gradle 6.0-rc-1
- *OS:* Windows 10 Enterprise

**Description of the issue**:
When I invoke `gradle jib` I get these deprecation warnings:

```
> Task :jib
Property 'jib.allowInsecureRegistries' @Input properties with primitive type 'boolean' cannot be @Optional. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
Property 'jib.container.useCurrentTimestamp' @Input properties with primitive type 'boolean' cannot be @Optional. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0.
```

**Expected behavior**:
No deprecation warning

**Steps to reproduce**:
see above

**Log output**: 
see abouve

glefloch:
close #2086 by removing deprecated `@org.gradle.api.tasks.Optional` annotation

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
@@ -192,7 +192,6 @@ public class JibExtension {
   }
 
   @Input
-  @Optional
   boolean getAllowInsecureRegistries() {
     if (System.getProperty(PropertyNames.ALLOW_INSECURE_REGISTRIES) != null) {
       return Boolean.getBoolean(PropertyNames.ALLOW_INSECURE_REGISTRIES);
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -192,7 +192,6 @@ public OutputPathsParameters getOutputPaths() {
   }
 
   @Input
-  @Optional
```
loosebazooka: I don't know if we have defaults for this parameter, like we do for `useCurrentTimestamp`. Its instantiated with
```	
allowInsecureRegistries = objectFactory.property(Boolean.class);
```

chanseokoh:
But we set the default.

```java
    allowInsecureRegistries = objectFactory.property(Boolean.class);
    containerizingMode = objectFactory.property(String.class);

    // Sets defaults.
    allowInsecureRegistries.set(DEFAULT_ALLOW_INSECURE_REGISTIRIES);
    containerizingMode.set(DEFAULT_CONTAINERIZING_MODE);
```

I'll do an actual testing to make sure these are by default true.

PR #2114, opened 2019-10-31
Add "dev" and "devFull" to build

loosebazooka:
`gradlew dev`: formats the code + runs the build + runs tests
`gradle devFull`: does all of `dev` + run integration tests

`devFull` should fail fast if JIB_INTEGRATION_TEST_PROJECT is not set.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -351,3 +351,20 @@ subprojects {
   }
   /* INCLUDED PROJECT DEPENDENCY HELPER */
 }
+
+tasks.register("dev") {
+  subprojects.each { subproject ->
+    subproject.tasks.check.dependsOn subproject.tasks.googleJavaFormat
+    dependsOn subproject.tasks.check
+  }
+}
+
+tasks.register("devFull") {
+  if (!System.getenv('JIB_INTEGRATION_TESTING_PROJECT')) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -351,3 +351,20 @@ subprojects {
...
+
+tasks.register("dev") {
+  subprojects.each { subproject ->
+    subproject.tasks.check.dependsOn subproject.tasks.googleJavaFormat
+    dependsOn subproject.tasks.check
+  }
+}
+
+tasks.register("devFull") {
+  if (!System.getenv('JIB_INTEGRATION_TESTING_PROJECT')) {
```
chanseokoh: `... && !System.getenv('JIB_INTEGRATION_TESTING_LOCATION')) {`

This is documented in `CONTRIBUTING.md`.

loosebazooka: That's optional though isn't it?

chanseokoh: Defining only one of them is enough to run `./gradlew integrationTest`. https://github.com/GoogleContainerTools/jib/blob/9d84f949a5419fb89cb809cee67ebdb6223e8648/jib-core/src/integration-test/java/com/google/cloud/tools/jib/IntegrationTestingConfiguration.java#L33-L35

loosebazooka: Oh yeah you're right. I'll fix this

loosebazooka: Used same message.

PR #2116, opened 2019-11-01
add javadoc check to dev, cleanup spaces

loosebazooka:
We run `javadoc` on all projects in `dev` mode.

The branch name is deceptive because I was thinking differently when I initially wrote this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -353,23 +353,26 @@ subprojects {
 }
 
 
+/* LOCAL DEVELOPMENT HELPER TASKS */
 tasks.register("dev") {
   subprojects.each { subproject ->
     subproject.tasks.check.dependsOn subproject.tasks.googleJavaFormat
     dependsOn subproject.tasks.check
+    dependsOn subproject.tasks.javadoc
   }
 }
 
 tasks.register("devFull") {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -353,23 +353,26 @@ subprojects {
 }
 
 
+/* LOCAL DEVELOPMENT HELPER TASKS */
```
chanseokoh: How about `/* BEGIN: LOCAL ...`?

build.gradle
```diff
@@ -353,23 +353,26 @@ subprojects {
...
+        "Must set environment variable JIB_INTEGRATION_TESTING_PROJECT to the "
+        + "GCP project to use for integration testing or "
+        + "JIB_INTEGRATION_TESTING_LOCATION to a suitable registry/repository location.");
   }
   dependsOn tasks.dev
   subprojects.each { subproject ->
     dependsOn subproject.tasks.integrationTest
   }
 }
+/* LOCAL DEVELOPMENT HELPER TASKS */
```
chanseokoh: `END: LOCAL ...`?

build.gradle
loosebazooka: I've been kinda using this syntax throughout the file. Can change later, but this is self consistent for now.

PR #2118, opened 2019-11-01
Finalize new HTTP framework changes

chanseokoh:
Follow-up after #2100.

Also fixes #2117.

PR #2119, opened 2019-11-01
Add shutdown() to FailoverHttpClient

### LINKED ISSUE #2111: Implement close() or shutDown() on TlsFailoverHttpClient
`TlsFailoverHttpClient` generates closeable `Response` instances. It can implement `close()` or `shutDown()` to close all `Response`s it generated.

chanseokoh:
Fixes #2111.

This is on top of #2118, so review #2118 first.


jib-core/src/test/java/com/google/cloud/tools/jib/http/FailoverHttpClientTest.java
```diff
@@ -333,6 +333,17 @@ public void testGet_originalRequestHeaderUntouchedWhenClearingHeader() throws IO
         "Basic ZmFrZS11c2VybmFtZTpmYWtlLXNlY3JldA==", request.getHeaders().getAuthorization());
   }
 
+  @Test
+  public void testShutDown() throws IOException {
```
loosebazooka: Can we add a test to verify that previously shutdown connections do not cause this code to break?

chanseokoh: Added some tests.

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerFailoverHttpClientTest.java
```diff
@@ -151,6 +151,21 @@ public void testProxyCredentialProperties()
...
 
+  @Test
+  public void testClosingResourcesMultipleTimes_noErrors()
+      throws IOException, InterruptedException, GeneralSecurityException, URISyntaxException {
+    FailoverHttpClient httpClient = new FailoverHttpClient(true /*insecure*/, false, logger);
+    try (TestWebServer server = new TestWebServer(false, 2);
+        Response ignored1 = httpClient.get(new URL(server.getEndpoint()), request);
+        Response ignored2 = httpClient.get(new URL(server.getEndpoint()), request)) {
+      ignored1.close();
+      ignored2.close();
```
loosebazooka: Do we need to also write a test for transportsCreated?

chanseokoh: Actual transports are automatically created behind the scene (two transports in this test), and `httpClient.shutDown()` here shows there is no problem calling `HttpTransport.shutdown()` multiple times. The other test in `FailoverHttpClient` actually verifies if `httpClient.showDown()` calls `HttpTransport.shutdown()`, so I think I've covered everything?

chanseokoh: I'm going to merge this, but if you want to see something specific fixed, feel free to update, and I'll fix it.

PR #2120, opened 2019-11-01
Upgrade Google HTTP Client to 1.33.0 (Apache v2)

### LINKED ISSUE #1993: Upgrade Google HTTP Client
#1980 downgraded it to 1.27.0.

I filed a [bug](https://github.com/googleapis/google-http-java-client/issues/820) against google-http-client about the slow network throughput.

Note when upgrading, we no longer have to exclude `org.apache.httpcomponents:httpclient` as we did prior to #1980.

chanseokoh:
This is on top of #2118, so review that first.

Fixes #1993. Basically reverting #1980 (which in turn reverted #1882, so this is double-reverting).

- And as in #1980, we no longer need to add proxy credentials ourselves, as the Apache v2 automatically takes care of it.

- Also removed some unnecessary URL direction code in `RegistryEndpointCaller`, as I realized google-http-client by default takes care of them (although it has a [bug](https://github.com/googleapis/google-http-java-client/issues/873)).

- Using the Apache v2, this will clear all the deprecated usage warnings around Google HTTP Client.


loosebazooka:
Is there a concern that we could run into the performance problems from `1.5.x`?

chanseokoh:
https://github.com/googleapis/google-http-java-client/pull/855#issuecomment-546061133 fixed the performance issue in my small experiment ([code here](https://github.com/googleapis/google-http-java-client/issues/820#issuecomment-531378795)) where I was pushing 40MB of data to Docker Hub, back to normal.

But I'll do some more end-to-end tests before merging.

chanseokoh:
I find no performance issue with the new google-http-client.

loosebazooka:
Looks good, I was tracking through the bugs on this performance thing, and I'm confused that the socket buffer is ignored AND that it affected performance?

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerFailoverHttpClientTest.java
```diff
@@ -149,4 +150,43 @@ public void testProxyCredentialProperties()
       }
     }
   }
+
+  @Test
+  public void testVerbatimRedirectionUrls()
+      throws IOException, InterruptedException, GeneralSecurityException, URISyntaxException {
+    String url1 = "?id=301&_auth_=exp=1572285389~hmac=f0a387f0";
```
loosebazooka: how/where do we generate these values? Would be helpful for someone modifying these tests.

jib-core/build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
 
 dependencies {
   implementation "com.google.http-client:google-http-client:${dependencyVersions.GOOGLE_HTTP_CLIENT}"
+  implementation "com.google.http-client:google-http-client-apache-v2:${dependencyVersions.GOOGLE_HTTP_CLIENT_APACHE_V2}"
   implementation("com.google.auth:google-auth-library-oauth2-http:${dependencyVersions.GOOGLE_AUTH_LIBRARY_OAUTH2_HTTP}") {
```
loosebazooka: is this still overriding the value with some broken http client version?
Can we check via enforcement in the root build.gradle if this `exclude` is still required?

loosebazooka: actually I just saw we explicitly enforce this override anyway, perhaps we can take out the `excludes` and see what happens?

chanseokoh: I have been thinking we can remove this. Especially that we don't need to pin down the Apache components. With `excludes`, the dependency tree excludes `google-http-client` (of course),

```
+--- com.google.auth:google-auth-library-oauth2-http:0.18.0
|    +--- com.google.auto.value:auto-value-annotations:1.6.6
|    +--- com.google.code.findbugs:jsr305:3.0.2
|    +--- com.google.auth:google-auth-library-credentials:0.18.0
|    +--- com.google.http-client:google-http-client-jackson2:1.32.1
|    |    \--- com.fasterxml.jackson.core:jackson-core:2.9.9 -> 2.9.10
```
and without `excludes`, no problem as well since Gradle will upgrade its `google-http-client` to 1.33.0. (BTW, `google-auth-library-oauth2-http:0.18.0` is the latest we can get as of now.
```
+--- com.google.auth:google-auth-library-oauth2-http:0.18.0
|    +--- com.google.auto.value:auto-value-annotations:1.6.6
|    +--- com.google.code.findbugs:jsr305:3.0.2
|    +--- com.google.auth:google-auth-library-credentials:0.18.0
|    +--- com.google.http-client:google-http-client:1.32.1 -> 1.33.0 (*)
|    +--- com.google.http-client:google-http-client-jackson2:1.32.1
|    |    +--- com.google.http-client:google-http-client:1.32.1 -> 1.33.0 (*)
|    |    \--- com.fasterxml.jackson.core:jackson-core:2.9.9 -> 2.9.10
```

So I agree we should be good. And we have the check in the root `build.gradle` to ensure the version is correct.

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerFailoverHttpClientTest.java
chanseokoh: For now this can be anything. Added the source anyway.

PR #2125, opened 2019-11-06
Instantiate a singleton stateful HttpClient instance

chanseokoh:
I am propagating the singleton instance downstream. However, technically stuffing the stateful HttpClient instance into `BuildConfiguration` was another option (somewhat similar to our having `ExecutorService` inside it), but I don't think I really like it. In fact, now I am not sure I like having `ExecutorService` inside `BuildConfiguration`. It is already a stateful object with `executorService` and `eventHandlers`, but it may not be that big of a deal. It may merely be the matter of in which way we want to perceive `BuildConfiguration`, and I have no strong opinion.


briandealwis:
Would it make sense to rename BuildConfiguration &rarr; BuildContext?

chanseokoh:
I like the naming change. I think it works reasonably well in the current setup. (We can then stuff everything into this single all-mighty object and pass it around, until we get tired of it. :D)

I'll check if putting HttpClient to `BuildContext` results in natural code flow.

chanseokoh:
So I've tried putting `FailoverHttpClient` into `BuildConfiguration` (to be renamed to `BuildContext`), and I think it worked out reasonably well. I made `BuildConfiguration` `Closeable` and realized that it also makes sense to instantiate `ExecutorService` at the time of building a `BuildConfiguration` instance together with `FailoverHttpClient` in a consistent way now that it is a "build context."

chanseokoh:
Hmm... I've been regarding it as a general purpose secure HTTP client, although it has this special failover feature. You can throw any `https://` URL (and `http://` if you enable insecure connection). I certainly don't think the name `registry` fits in there. There's nothing inherently related to a container registry. It's pretty general.

PR #2129, opened 2019-11-06
Add copyright checkstyle check

loosebazooka:
This does a few more modifications to the provided google_checks.xml
using xml apis from groovy instead of simple string replacement.

Required because checkstyle doesn't let us run multiple passes of
the checkstyle tool. (presumably you could write a custom set of tasks
to do this?)

Also changing file permissions at the "Checker" module means no java
checks would run anywhere, so we have to move where we limit the checks
to just xml, properties to the submodules that use them.

---

I tend to forget to add copyright, this standardizes them. I think @briandealwis might have to update their template?

The only significant change happened in the `build.gradle` file.


loosebazooka:
oh yeah those haven't been used since we moved to the new build.

config/checkstyle/copyright-java.header
```diff
@@ -0,0 +1,15 @@
+^/\*$
+^ \* Copyright 20(17|18|19) Google LLC\.$
```
loosebazooka: FYI: We'll have to update this to include 2020 when we add new files next year.

PR #2132, opened 2019-11-07
Optimize insecure registry failover by caching failover history

### LINKED ISSUE #946: Optimize insecure registry failover

chanseokoh:
Fixes #946.

Significantly cuts the number of HTTP registry connections, by more than half in my testing from 528 to 248. Once a failover occurs for a registry host, it is not repeated (down to 4 from 127 as shown below). 

In practice, however, due to Jib's concurrent connections, it may repeat the failover for the same host but mostly only once or twice.
```
$ grep 'User-Agent: jib' network_traces.sh.log | wc -l
528
$ grep '\(again with no TLS verification\.\)\|\(again with HTTP\.\)' network_traces.sh.log | wc -l
127
```

```
$ grep 'User-Agent: jib' network_traces.sh.log | wc -l
248
$ grep '\(again with no TLS verification\.\)\|\(again with HTTP\.\)' network_traces.sh.log | wc -l
4
```

Increases the failover log level from INFO to WARNING.


chanseokoh:
This is now ready for review.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -71,6 +75,12 @@
  */
 public class FailoverHttpClient {
 
+  private static enum Failover {
```
loosebazooka: maybe a little comment explaining what this enum is for? Is this exclusively for the failover history?

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -209,10 +221,18 @@ public Response put(URL url, Request request) throws IOException {
...
   public Response call(String httpMethod, URL url, Request request) throws IOException {
-    if (!isHttpsProtocol(url) && !enableHttpAndInsecureFailover) {
+    if (!isHttpsProtocol(url)) {
+      if (enableHttpAndInsecureFailover) {
+        return call(httpMethod, url, request, getHttpTransport(true));
+      }
       throw new SSLException("insecure HTTP connection not allowed: " + url);
     }
 
+    Optional<Response> optionalResponse = followFailoverHistory(httpMethod, url, request);
```
loosebazooka: I think the use of `optional` in the name `optionalReposonse` is unnecessary given the type information is already available.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -223,32 +243,49 @@ public Response call(String httpMethod, URL url, Request request) throws IOExcep
 
       try {
         logInsecureHttpsFailover(url);
-        return call(httpMethod, url, request, getHttpTransport(false));
+        Response response = call(httpMethod, url, request, getHttpTransport(false));
+        failoverHistory.put(url.getHost() + ":" + url.getPort(), Failover.INSECURE_HTTPS);
```
loosebazooka: Is this safe? Could we potenially get something like `"hello.com:8000" + "8000"` vs `"hello.com" + "8000:8000"`?

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -223,32 +243,49 @@ public Response call(String httpMethod, URL url, Request request) throws IOExcep
...
       }
       throw ex;
     }
   }
 
+  private Optional<Response> followFailoverHistory(String httpMethod, URL url, Request request)
+      throws IOException {
+    Preconditions.checkArgument(isHttpsProtocol(url));
+    switch (failoverHistory.getOrDefault(url.getHost() + ":" + url.getPort(), Failover.NONE)) {
+      case INSECURE_HTTPS:
```
loosebazooka: should we add a comment that this order is the reverse of how we failover? Incase someone the in the future tries to add something here, but has no context?

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -209,10 +221,18 @@ public Response put(URL url, Request request) throws IOException {
    * @throws IOException if building the HTTP request fails.
    */
   public Response call(String httpMethod, URL url, Request request) throws IOException {
-    if (!isHttpsProtocol(url) && !enableHttpAndInsecureFailover) {
+    if (!isHttpsProtocol(url)) {
+      if (enableHttpAndInsecureFailover) {
```
loosebazooka: I find this part a little confusing: It seems like we're checking for: "http and insecure https are not allowed", but really we're only checking for "http not allowed"?

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
loosebazooka: Is there an issue here that we are following failover history even if enableHttpAndInsecureFailover is false?

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
TadCordle: It looks like `getHost()` excludes the port number.

loosebazooka: I think what I'm really asking is there anyway in which this addition can result in the same string from two different values (yes technically, but no in this case?)?

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
chanseokoh: The `enableHttpAndInsecureFailover` really indicates two things shoved into one flag (ANDed): `enableHttp` AND `enableInsecureFailover`. We can actually split this into two flags, but I thought this isn't really worth it, especially in our context.

And this was actually my attempt to make the code easier to understand by putting a short-circuit (kind of an optimization); I don't really have to change this line in terms of net effect, and things would work as intended without this change actually. This short-circuit (optimization) is like this:

- If the request was plain-HTTP, then
   - if HTTP is allowed, just call the endpoint right away and exit. No need to think about failover.
   - if not allowed, fail right away with `SSLException`.

Does this make sense? Does the code need some more tweaks to make it less confusing?

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
chanseokoh: Are you suggesting we can just return `null` (instead of `Optional.empty()`) if there was previous recorded history?

chanseokoh: `enableHttpAndInsecureFailover` false, then no failover happens and no history will ever be recorded, so `followFailoverHistory` will always return `Optional.empty()` immediately. I think it has no overhead, and this keeps the code simple. But one could surely argue if we add a branch for the no-failover-ever setup, it might increase readability.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
chanseokoh: I don't believe it can ever happen. The host portion from `URL` cannot have `:`. And the port is an integer value.

Although there can be a case where `https://example.com` and `https://example.com:443` will end up two different keys (the former goes into `https://example.com:-1`, but I intended to leave them separate. In practice the number of actual hosts involved in Jib would be merely 2 to 3 at most especially if a registry is insecure, and this keeps the code simple.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
chanseokoh: There are two orders actually.

1. secure HTTPS --> insecure HTTPS --> plain HTTP
1. secure HTTPS --> plain HTTP: no HTTPS attempt in this case.

But the order of the `case`s in the `switch` doesn't matter. How exactly do you want to comment this?

chanseokoh: Re-ordered the cases anyways. Added a comment too.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
loosebazooka: oh no, this is just naming.
```
Optional<Response> response = ....
```

chanseokoh: Sadly, the code below already took the name `response`, and I need to pick a different name for this.

loosebazooka: oh I see, no this is fine. I just needed to look at it again.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
loosebazooka: Right, I think this is fine.
A comment next to it (`// we only care if http is enabled`) or something might be helpful but isn't necessary.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
loosebazooka: oh you know what, I wasn't even thinking that the map would only hold **one** failover value, I'm dumb. You should order this in whatever way makes logical sense to you.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
loosebazooka: how about `responseFromFailoverPath`, or something that indicates the source?

chanseokoh:
Huh, this branch isn't really the one that I originally intended to put up. Somehow I see some commit is lost. The `failoverHistory` map should be a `ConcurrentHashMap`.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -110,6 +120,8 @@ private static HttpTransport getInsecureHttpTransport() {
   private final Supplier<HttpTransport> secureHttpTransportFactory;
   private final Supplier<HttpTransport> insecureHttpTransportFactory;
 
+  private final ConcurrentHashMap<String, Failover> failoverHistory = new ConcurrentHashMap<>();
```
TadCordle: This makes me wonder if we should add any tests for concurrency. Maybe not worth the trouble.

chanseokoh: Good point. But I can't think of a practical way to testing concurrency in a deterministic way (or what really to verify thereof), particularly in this case.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -271,7 +308,7 @@ private Response call(String httpMethod, URL url, Request request, HttpTransport
     try {
       Response response = new Response(httpRequest.execute());
       synchronized (responsesCreated) {
-        responsesCreated.addLast(response);
+        responsesCreated.add(response);
```
TadCordle: For consistency, I think we can make the same change with `trasnsportsCreated` on L323.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
TadCordle: Yeah, I think using a thread-safe data structure is enough for the changes in this PR. Although maybe it's something to think about in the future.

TadCordle:
Oh and changelog, haha

loosebazooka:
We can probably put this out with `1.8`...

PR #2133, opened 2019-11-07
Make TestWebServer thread-safe

chanseokoh:
Not really necessary, but I have to do this for completeness sake after I made the server accept concurrent connections recently.

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
@@ -48,7 +48,7 @@ public class TestWebServer implements Closeable {
   private final ServerSocket serverSocket;
   private final ExecutorService executorService;
   private final Semaphore threadsStarted;
-  private final StringBuilder inputRead = new StringBuilder();
+  private final StringBuffer inputRead = new StringBuffer(); // StringBuilder not thread-safe
 
   private final List<String> responses;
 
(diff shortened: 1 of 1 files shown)


chanseokoh:
Note this class is testing code. No prod involved.

jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -119,14 +119,18 @@ private Void serveResponses() throws IOException {
...
         for (String line = reader.readLine();
             line != null && !line.isEmpty(); // An empty line marks the end of an HTTP request.
             line = reader.readLine()) {
-          inputRead.append(line + "\n");
+          if (firstByte == -1) {
+            inputRead.append(line + "\n");
+          } else {
+            inputRead.append(((char) firstByte) + line + "\n");
+            firstByte = -1;
+          }
```
TadCordle: Could you explain this change? I'm confused about why it's necessary.

chanseokoh: I wanted to ensure lines are never cut in the middle. For example,

Thread 1: does `inputRead.append('G')
Thread 2: does `inputRead.append('G')
Thread 1 & 2: read the rest of their respective line and do `input.append(line + "\n")`

Then `inputRead` will have the content:

 ```
GGET /path HTTP...
ET /path HTTP...
```

I just wanted to prevent that.

PR #2136, opened 2019-11-08
clarify repository behavior in tests

loosebazooka:


loosebazooka:
Yeah gradle treats local file repositories as available locally... and doesn't copy them into the gradle cache, usually at `$HOME/.gradle/caches/modules-2/files-2.1`

PR #2137, opened 2019-11-08
Implement standard multi-threaded socket listening

chanseokoh:
Sorry, this feels like a spam. I realized Error Prone complains with `StringBuffer` and suggests to do explicit synchronization.

Also as part of this change, I decided to do the standard way of implementing a multi-threaded server: the main server thread loops through `serverSocket.accept()` and it should spawn new threads for each socket. The dirty hack was to quickly fix some tests while I was revamping the new HTTP framework.

This only involves test code.

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
@@ -45,12 +45,13 @@ import javax.net.ssl.SSLContext;
 public class TestWebServer implements Closeable {
 
   private final boolean https;
+  private final int numThreads;
+  private final List<String> responses;
+
   private final ServerSocket serverSocket;
   private final ExecutorService executorService;
-  private final Semaphore threadsStarted;
+  private final Semaphore serverStarted = new Semaphore(1);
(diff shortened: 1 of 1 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -66,13 +67,11 @@ public TestWebServer(boolean https, List<String> responses, int numThreads)
...
     this.responses = responses;
+    this.numThreads = numThreads;
     serverSocket = https ? createHttpsServerSocket() : new ServerSocket(0);
-    threadsStarted = new Semaphore(1 - numThreads);
-    executorService = Executors.newFixedThreadPool(numThreads);
-    for (int i = 0; i < numThreads; i++) {
-      ignoreReturn(executorService.submit(this::serveResponses));
-    }
-    threadsStarted.acquire();
+    executorService = Executors.newFixedThreadPool(numThreads + 1);
```
loosebazooka: why +1?

loosebazooka: oh never mind, figured it out. Listen is actually using 1 thread.

PR #2139, opened 2019-11-11
Set Gradle version to run in integration tests

chanseokoh:
`GRADLE_MIN_VERSION` is 4.9.

Verified one of the tests fails if Gradle 5 API is used.

### DIFF
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/TestProject.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/TestProject.java
@@ -75,7 +75,11 @@ class TestProject extends TemporaryFolder implements Closeable {
     projectRoot = newFolder().toPath();
     copyProject(testProjectName, projectRoot);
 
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/TestProject.java
```diff
@@ -75,7 +75,11 @@ protected void before() throws Throwable {
     projectRoot = newFolder().toPath();
     copyProject(testProjectName, projectRoot);
 
-    gradleRunner = GradleRunner.create().withProjectDir(projectRoot.toFile()).withPluginClasspath();
+    gradleRunner =
+        GradleRunner.create()
+            .withGradleVersion(JibPlugin.GRADLE_MIN_VERSION.getVersion())
```
loosebazooka: Should we have parameterized this to run on gradle_min_version and maybe 5,6?

chanseokoh: That's a good idea. One caveat is that it will double (or triple) the time to run the tests.

PR #2144, opened 2019-11-12
Allow pushing local base image layers immediately after compressing

### LINKED ISSUE #1913: Optimize local base image support: more efficient layer processing

TadCordle:
Fixes #1913.

This PR splits up the local base image steps so that the base image layer futures are returned before the layers are actually processed. This means later steps with dependencies on the base image layers can start as soon as each layer finishes being compressed, rather than waiting for all layers to be compressed before continuing.

Amount of time save varies depending on the base image, but it seems to save 4 seconds for uncached distroless (15s -> 11s).

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
@@ -188,66 +188,69 @@ public class LocalBaseImageSteps {
       ProgressEventDispatcher.Factory progressEventDispatcherFactory)
       throws IOException, LayerCountMismatchException {
     ExecutorService executorService = buildConfiguration.getExecutorService();
-    try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider();
-        TimerEventDispatcher ignored =
-            new TimerEventDispatcher(
-                buildConfiguration.getEventHandlers(), "Extracting tar " + tarPath)) {
+    try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
       Path destination = tempDirectoryProvider.newDirectory();
-      TarExtractor.extract(tarPath, destination);
 
-      InputStream manifestStream = Files.newInputStream(destination.resolve("manifest.json"));
-      DockerManifestEntryTemplate loadManifest =
-          new ObjectMapper()
-              .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true)
-              .readValue(manifestStream, DockerManifestEntryTemplate[].class)[0];
-      manifestStream.close();
+      try (TimerEventDispatcher ignored =
+          new TimerEventDispatcher(
+              buildConfiguration.getEventHandlers(),
+              "Extracting tar " + tarPath + " into " + destination)) {
+        TarExtractor.extract(tarPath, destination);
 
-      Path configPath = destination.resolve(loadManifest.getConfig());
-      ContainerConfigurationTemplate configurationTemplate =
-          JsonTemplateMapper.readJsonFromFile(configPath, ContainerConfigurationTemplate.class);
-      BlobDescriptor originalConfigDescriptor =
-          Blobs.from(configPath).writeTo(ByteStreams.nullOutputStream());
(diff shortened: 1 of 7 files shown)


TadCordle:
Hmm something's broken.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -236,36 +236,33 @@ private void saveDocker() {
...
-            LocalBaseImageSteps.retrieveDockerDaemonImageStep(
-                buildConfiguration, childProgressDispatcherFactory, dockerClient.get())));
+            LocalBaseImageSteps.retrieveDockerDaemonLayersStep(
+                buildConfiguration,
+                childProgressDispatcherFactory,
+                dockerClient.get(),
+                tempDirectoryProvider));
+    results.baseImageLayers = executorService.submit(() -> localImage.get().layers);
+    results.baseImageAndAuth =
+        executorService.submit(LocalBaseImageSteps.retrieveImageAndAuthorizationStep(localImage));
```
chanseokoh: I expect this to be at least

```java
executorService.submit(
    () -> LocalBaseImageSteps.retrieveImageAndAuthorizationStep(localImage.get()));
```

to avoid putting `Future` in the method signature of `retrieveImageAndAuthorizationStep()`. But ultimately,

```java
executorService.submit(
    () -> {
      List<PreparedLayer> layers = realizeFutures(localImage.get().layers);
      return LocalBaseImageSteps.retrieveImageAndAuthorizationStep(layers, localImage.get().containerConfiguration);
    }
```

and `retrieveImageAndAuthorizationStep()` shouldn't block or deal with anything related to concurrency like `Future` or `ExecutorService`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -124,26 +124,47 @@ static boolean isGzipped(Path path) throws IOException {
...
+        cacheDockerImageTar(
+            buildContext, tarPath, progressEventDispatcherFactory, tempDirectoryProvider);
+  }
+
+  static Callable<ImageAndAuthorization> retrieveImageAndAuthorizationStep(
+      List<PreparedLayer> layers, ContainerConfigurationTemplate configurationTemplate) {
+    return () -> {
+      // Collect compressed layers and add to manifest
+      V22ManifestTemplate v22Manifest = new V22ManifestTemplate();
+      for (PreparedLayer layerFuture : layers) {
```
chanseokoh: `layerFuture` --> `layer`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java
```diff
@@ -124,26 +124,47 @@ static boolean isGzipped(Path path) throws IOException {
...
-      Path tarPath) {
-    return () -> cacheDockerImageTar(buildContext, tarPath, progressEventDispatcherFactory);
+      Path tarPath,
+      TempDirectoryProvider tempDirectoryProvider) {
+    return () ->
+        cacheDockerImageTar(
+            buildContext, tarPath, progressEventDispatcherFactory, tempDirectoryProvider);
+  }
+
+  static Callable<ImageAndAuthorization> retrieveImageAndAuthorizationStep(
```
chanseokoh: I think the method name is misleading, since it doesn't try retrieving something but just building an `Image` instance from existing scalar values. `returnImageAndAuthorizationStep`?

PR #2146, opened 2019-11-12
Rename BuildConfiguration to BuildContext

chanseokoh:


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildContext.java
```diff
@@ -241,12 +241,12 @@ public Builder setExecutorService(@Nullable ExecutorService executorService) {
     }
 
     /**
-     * Builds a new {@link BuildConfiguration} using the parameters passed into the builder.
+     * Builds a new {@link BuildContext} using the parameters passed into the builder.
      *
      * @return the corresponding build configuration
```
TadCordle: the corresponding build context

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildContextTest.java
```diff
@@ -234,7 +230,7 @@ public void testBuilder_missingValues() throws IOException {
 
     // All required fields missing
     try {
-      BuildConfiguration.builder().build();
+      BuildContext.builder().build();
       Assert.fail("Build configuration should not be built with missing values");
```
TadCordle: Build context should not be built with missing values?

TadCordle: Or maybe just change it to "BuildContext"

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildContext.java
chanseokoh: good catch

PR #2151, opened 2019-11-14
Minor fixes

chanseokoh:
Does this make sense? I'm still not entirely certain if "\n" will work just right or not on different platforms. I've always thought it does when Java prints it out. But at least this makes it consistent with others.


loosebazooka:
I wonder if we should test this with skaffold before merging? Not sure what kind of newlines it is handling?

chanseokoh:
Good idea. Let's just not merge this before the release. I think @TadCordle could test these things out quickly later.

TadCordle:
This will work with skaffold, since it just does a simple regex for `BEGIN JIB JSON\r?\n({.*})`. Any `\r`s or `\n`s beforehand should be irrelevant.

Issue #1945, opened 2019-08-30
Submodule dependencies may not be built reproducibly without user intervention


loosebazooka:
for a project A that has a dependecy on project Lib, if lib.jar is not built reproducibly, then the layer lib.jar is built into will also not be.

This currently can only be solved with user intervention. Perhaps in gradle using the techniques described here: https://purpledevcat.com/2018/02/14/reproducible-builds-in-java.html
and in maven using some sort of custom packager.

loosebazooka:
Perhaps we should expand and rewrite these jars and erase the required data. And we should expect user intervention when we make guarantees. I think we can, for project dependencies, do the following transformations (as described [here](https://zlika.github.io/reproducible-build-maven-plugin))

- sorts ZIP entries by name
- replaces file timestamps in ZIP entries with a fixed value
- removes timestamps, user names and tool versions in MANIFEST.MF

chanseokoh:
I have realized that, even for a single module project, we may have the same problem when using the "packaged" containering mode.

Also, since we are exploding WAR and `MANIFEST.MF` goes into the resources directory, this may also be the case for WAR.

loosebazooka:
Done in documentation... still hard for users to discover.

PR #2155, opened 2019-11-15
Update multimodule example with module dependency

### LINKED ISSUE #2042: Module-to-module dependencies in a multi-module maven file are not copied into the container
**Environment**:

- *Jib version:* 1.6.1
- *Build tool:* 3.6.0 (jdk 1.8.0_222)
- *OS:* Windows 10

**Description of the issue**:
In a multi module pom, the intermodule dependencies are not copied into the container when using the exposed mode and not running package. 

Based on the [multi-module example](https://github.com/GoogleContainerTools/jib/tree/master/examples/multi-module) you run `mvn compile jib:build` but that doesn't work, as the dependencies are not packaged, and jib wants to have all dependencies as jar.

As I see it, this is either this is an issue with the example, or the exposed mode should also copy the class files of the intermodule dependencies.

**Expected behavior**:

Either:
1. warning that intermodule dependencies are not included when running compile, so run with `mvn package`
2. copy class files of intermodule depedencies just like the class files of the module. That would fit better with the exposed mode of operation.

**Steps to reproduce**:

  1. clone: https://github.com/SWAT-engineering/jib-test
  2. run: `mvn clean compile jib:dockerBuild`
  3. `docker run --rm test-app/app`:
```
Exception in thread "main" java.lang.NoClassDefFoundError: common/A
        at app.B.main(B.java:7)
Caused by: java.lang.ClassNotFoundException: common.A
        at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
        ... 1 more
```
  4. run `mvn clean package jib:dockerBuild`
  3. `docker run --rm test-app/app`:
```
1764
```

Anyway, thanks for the great tool, I have a workaround for now, but just thought to post it here, it might be a nice extension to the documentation or else a new feature.

loosebazooka:
fixes #2042 
adds some info on #1945 

Rendered: https://github.com/GoogleContainerTools/jib/tree/multi-module-docs/examples/multi-module


examples/multi-module/shared-library/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <plugins>
+      <plugin>
+        <groupId>com.google.cloud.tools</groupId>
+        <artifactId>jib-maven-plugin</artifactId>
+        <configuration>
+          <!-- we don't want jib to execute on this module -->
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+      <!-- we want this library to built reproducibly -->
```
chanseokoh: to be built

loosebazooka:
Will wait for #2156 so I can rebase against changes made to this example.

PR #2164, opened 2019-11-19
Update ISSUE_TEMPLATE.md

chanseokoh:
When there's no space, emphasis doesn't get rendered.


TadCordle:
Really? The rich diff isn't showing any difference.

chanseokoh:
Ah, it doesn't get rendered if it has a colone like *test:*test

Issue #1852, opened 2019-07-16
Way to exclude dependencies in the image or have right support for various forms of fat/shaded JARs


maxbrunet:
Hi again,

When building with `containerizingMode = 'packaged'`, our dependencies are built into the jar archive. Is it possible to not add the dependencies to `<appRoot>/libs` for a smaller image?

chanseokoh:
I replied to #1851 and just saw your [comment](https://github.com/GoogleContainerTools/jib/issues/1851#issuecomment-512006402). The [FAQ](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#i-want-to-containerize-an-executable-jar) explains the current limitation of `containerizingMode = 'packaged'`, and currently fat JARs or alternatively excluding dependencies are not supported yet. Supporting fat JARs created in various ways including by Spring Boot is something on our radar, but we are not sure when we will get to it.

I think I'll open this issue, which asks for either having
- proper support for fat JARs created in various ways; or
- just some means to exclude dependency JARs

chanseokoh:
It is best to avoid putting a fat JAR if possible, because it nullifies all the benefits of fine-grained layering by Jib that brings in fast iterations and caching. It'd be ideal if Jib could just use the original (non-fat) JAR. And because Spring Boot leaves the original JAR in the build directory, it seems not difficult for Jib to automatically detect Spring Boot and use the original JAR.

We may still introduce an option to not copy dependency JARs if we should really copy a fat JAR, but for now, it's worth implementing this auto-detection for Spring Boot.

chanseokoh:
Just for the record here, in the next Jib release, Spring Boot projects with `containerizingMode='package'` should work seamlessly for both Gradle and Maven out of the box.

Anyways, I don't think we'll need to provide a general way to exclude adding dependencies into the image for the sake of being able to put a fat JAR (with `<containerizingMode>packaged`). Now I think almost all the time, we can just put a normal, non-fat JAR (as long as a project generates the normal JAR and we can locate it) regardless of whether maven-shade-plugin, maven-assembly-plugin, or Gradle shadowJar plugin is applied to create a fat JAR.

chanseokoh:
> However, for a potential feature for generalized fined-grained dependency control, see #1962 (comment)

The [Jib Extension Framework](https://github.com/GoogleContainerTools/jib-extensions) is now available with the latest Jib versions. You can easily extend and tailor the Jib plugins behavior to your liking.

[...]

PR #2170, opened 2019-11-21
Detect Spring Boot fat-JAR repackaging and use original thin JAR

chanseokoh:
Implements https://github.com/GoogleContainerTools/jib/issues/1852#issuecomment-557275771 for Maven. That is, auto-detect Spring Boot fat-JAR repackaging and use the original thin JAR.

Additionally takes care of the corner cases with `<classifier>` and `<outputDirectory>` from `maven-jar-plugin`. This is basically irrelevant of the Spring Boot fat JAR stuff, but I decided to fix them for correctness.

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
@@ -40,6 +40,7 @@ import com.google.cloud.tools.jib.plugins.common.logging.SingleThreadedExecutor;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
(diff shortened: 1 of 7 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -427,9 +430,54 @@ public boolean isOffline() {
...
+      consoleLogger.log(
+          Level.LIFECYCLE, "Spring Boot repackaging (fat JAR) detected; using the original JAR");
+      if (outputDirectory.equals(buildDirectory)) { // Spring renames original only when needed
+        suffix += ".original";
+      }
+    }
+
+    String jarName =
+        project.getBuild().getFinalName() + (classifier == null ? "" : '-' + classifier) + suffix;
+    consoleLogger.log(Level.LIFECYCLE, "Using JAR: " + outputDirectory.resolve(jarName));
```
TadCordle: Does this need to be logged at `lifecycle`? Would be nice to keep log output minimal.

chanseokoh:
Oh, wait. Something's not right. Looking into the issue.

chanseokoh:
Added a Spring Boot integration test that verifies the image is built with a thin JAR and prints out "Hello world".

While doing that, I realized that JVM does not interpret the wildcard (`*`) in `java -cp /some/path/*`. Unfortunately, it has to end with `.jar` for JAR files. However, Spring Boot renames the original JAR with the `.jar.original` suffix, so we need to make the JAR filename end with `.jar` inside the image. The simplest solution I think is to copy the original JAR to a temp location for the sake of having the `.jar` suffix.

chanseokoh:
Okay, let's give this a spin. This won't make things worse even if this goes wrong, because the packaged mode has not been working for Spring Boot projects.

PR #2176, opened 2019-11-25
Use ConsoleLogger instead of native Gradle logger

### LINKED ISSUE #2175: Do not make use of native Gradle and Maven loggers
Using the native Gradle logger was probably why [our deprecation messages was overwritten](https://github.com/GoogleContainerTools/jib/pull/2109#issuecomment-555174722).

We should be careful not to make use of the native Maven or Gradle loggers. We can prevent misuse in `ProjectProperties` by just using `consoleLogger`. But it is very easy to accidentally use the native loggers in Maven mojos and Gradle tasks. Unfortunately, now I realized we had used the native loggers when printing deprecation notices.

chanseokoh:
See #2175.

However, I can never figure out why the test doesn't pass on the Travis Java 11 build. Doesn't make sense to me. Still looking into it.

### DIFF
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
@@ -206,7 +206,6 @@ public class GradleProjectPropertiesTest {
 
   @Before
   public void setup() throws URISyntaxException, IOException {
-    Mockito.when(mockLogger.isLifecycleEnabled()).thenReturn(true);
     Mockito.when(mockLogger.isDebugEnabled()).thenReturn(true);
     Mockito.when(mockLogger.isInfoEnabled()).thenReturn(true);
     Mockito.when(mockLogger.isWarnEnabled()).thenReturn(true);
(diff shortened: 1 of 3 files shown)


chanseokoh:
Ah, probably due to concurrency with `ConsoleLogger`.

jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenProjectPropertiesTest.java
```diff
@@ -222,6 +222,10 @@ private static Path zipUpDirectory(Path sourceRoot, Path targetZip) throws IOExc
 
   @Before
   public void setUp() throws IOException, URISyntaxException {
+    Mockito.when(mockLog.isDebugEnabled()).thenReturn(true);
+    Mockito.when(mockLog.isWarnEnabled()).thenReturn(true);
+    Mockito.when(mockLog.isErrorEnabled()).thenReturn(true);
```
chanseokoh: `isInfoEnabled()` is an unnecessary stubbing in Maven.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -374,6 +379,7 @@ public void testCreateContainerBuilder_noClassesFiles() throws InvalidImageRefer
         .thenReturn(new TestFileCollection(ImmutableSet.of(nonexistentFile)));
     gradleProjectProperties.createJibContainerBuilder(
         JavaContainerBuilder.from(RegistryImage.named("base")), ContainerizingMode.EXPLODED);
+    gradleProjectProperties.waitForLoggingThread();
```
TadCordle: I noticed we never call this in `MavenProjectPropertiesTest`, should we be?

chanseokoh: `MavenProjectPropertiesTest` doesn't have any `verify(mockLog)`, so need there (yet).

Issue #1851, opened 2019-07-16
Allow custom JAR task/path with containerizingMode = 'packaged' (also applies to WAR and Maven)


maxbrunet:
Hi there!

We build our SpringBoot application with a [`BootJar`](https://docs.spring.io/spring-boot/docs/2.0.2.RELEASE/gradle-plugin/api/) task (which inherits from `Jar`) in Graddle, and the new `containerizingMode = 'packaged'` feature hard-codes the task name to `jar`:

https://github.com/GoogleContainerTools/jib/blob/e3e3638987c2780dc55f2b88ac100f47390d384c/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java#L213-L218

Would it possible to have this name configurable?

Like:
```groovy
jib {
	jarTaskName = 'bootJar'
}
```
(or pass the task object directly, or even explicitly the path to the jar file)

I have currently worked around that by pointing the attributes of `jar` to `bootjar`:

```groovy
jar {
    enabled = false
    destinationDir = tasks.bootJar.getDestinationDir()
    archiveName = tasks.bootJar.getArchiveName()
}
```

I guess another alternative could be (not tested and I'd like not to have to rename my `bootJar` task):

```groovy
task jar(type: BootJar) {
	...
}
```

loosebazooka:
The one issue, and I think @chanseokoh can confirm, is that this isn't support for fatJar types. It's just Jars with dependencies still added into the container separately.

chanseokoh:
Hmm... this sounds familiar. I wonder if we should do the same for `bootJar` as in #1786. But I'm not entirely sure the interaction between `jar` and `bootJar` is same with `war` and `bootWar`.

But as @loosebazooka mentioned, you should avoid containerizing a fat JAR, as it results in [duplicating dependencies](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#i-want-to-containerize-an-executable-jar).

So, this issue is basically a feature request to support containerizing a fat JAR.

maxbrunet:
Hi @loosebazooka, thank you for quick reply, I actually just opened another issue about that #1852, though it would be separate problems (If it's by design, both will be quickly closed I guess). I'm pretty new to building Java apps/Gradle, I'm probably missing some pieces.

chanseokoh:
[...]

PR #2178, opened 2019-11-26
Detect bootJar task (Spring Boot fat JAR) and enable jar task to use the thin JAR

chanseokoh:
Implements https://github.com/GoogleContainerTools/jib/issues/1852#issuecomment-557275771 for Gradle. That is, this automatically enables https://github.com/GoogleContainerTools/jib/issues/1851#issuecomment-558293165 and makes use of the thin JAR by the `jar` task.

### DIFF
--- a/examples/multi-module/README.md
+++ b/examples/multi-module/README.md
@@ -1,4 +1,4 @@
-# Multi-module example (DRAFT)
+# Multi-module example
 
 This example shows how to build multiple containers for a multi-module project in both **Maven** and **Gradle**.
 
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
@@ -158,6 +158,8 @@ public class TestWebServer implements Closeable {
    * connections can be intermixed. However, no lines will ever be broken in the middle.
    */
   public String getInputRead() {
-    return inputRead.toString();
+    synchronized (inputRead) {
+      return inputRead.toString();
+    }
   }
 }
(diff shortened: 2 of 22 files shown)


chanseokoh:
> Would it make sense to start a document describing how jib handles different frameworks? <-- not necessary but could be nice to point specific framework users to?

I need to keep researching on it. Work in progress.

ewirch:
The commit message says: "Detect bootJar task and enable jar task". But it does more. It also forces the classifier to 'original'.

https://github.com/GoogleContainerTools/jib/commit/a692b8b0cde1f982f6e28309dd08ca80460950e2#diff-aef8bf567c91df6622aa0ea29ed0c682R192

Why would you do so? This broke our build. I can't set the classifier manually any more. Except: not applying jib.

chanseokoh:
Hi @ewirch,

When you apply the Spring Boot plugin, it will automatically disable the `jar` task. But sounds like you manually enabled it and set the `classifier` for either the `jar` task or the `bootJar` task. Could you share your `build.gradle` (or just the part related to configuring both the `jar` and `bootJar` tasks)?

ewirch:
Sure thing. The important part is here:
```
jar {
    enabled = true
}

bootJar {
    archiveClassifier = 'boot'
}
```
This part is reused in multiple sub-projects of a multi project build. This will give use two artifacts per sub-project:
- `<name>.jar`
- `<name>-boot.jar`

We use both in different deployment scenarios. After updating jib to 2.0.0 we end up with:
- `<name>-original.jar`
- `<name>-boot.jar`

without a option to get the original naming back. Sure, we could adjust our build scripts to use `<name>-original.jar` instead. But it feels incorrect for a Docker image building plugin to interfere in such a way with the build configuration.

chanseokoh:
@ewirch that makes sense. I think we can fix this easily. How about this?

Basically, if the classifier (or I guess `archiveClassifier`) of `bootJar` is set, then make Jib not set the `classifier` of the `jar` task. The only reason we set the `classifier` was to avoid `bootJar` and `jar` writing each JAR to the same location (overwriting each other).

PR #2181, opened 2019-12-02
Build correct OCI tars

### LINKED ISSUE #2124: jibBuildTar create wrong OCI format tarbal
When run the jibBuildTar gradle target and the container.format='OCI' set the resulted jib-image.tar still not contains index.json. 

Per the OCI image spec, there must be an index.json file present in OCI format images: https://github.com/opencontainers/image-spec/blob/master/image-layout.md

TadCordle:
Fixes #2124. Tested manually using `podman pull oci-archive:build/jib-image.tar` and it seems to work.

This PR adds OCI output with `jibBuildTar` when `jib.container.format = "OCI"`. The major differences in the OCI archive are as follows:
* Layers, config, and manifest are in a `sha/256` directory, named with their sha256 digests.
* `oci-layout` file in root directory defining the image layout version
* `index.json` that points to the image manifest


jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
```diff
@@ -61,6 +68,55 @@ public ImageTarball(
...
+        Blobs.from(manifest).writeTo(ByteStreams.nullOutputStream());
+    tarStreamBuilder.addByteEntry(
+        JsonTemplateMapper.toByteArray(manifest),
+        "blobs/sha256/" + manifestDescriptor.getDigest().getHash());
+
+    // Adds the oci-layout and index.json
+    tarStreamBuilder.addByteEntry(
+        "{\"imageLayoutVersion\": \"1.0.0\"}".getBytes(StandardCharsets.UTF_8), "oci-layout");
+    OCIIndexTemplate index = new OCIIndexTemplate();
+    // TODO: figure out how to tag with allTargetImageTags
```
TadCordle: Note: need to look into a way of doing this, if possible. Docker tars have `RepoTags` in the manifest, but I haven't found an OCI equivalent yet.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
```diff
@@ -61,6 +68,55 @@ public ImageTarball(
...
+    // Adds the manifest to the tarball
+    BlobDescriptor manifestDescriptor =
+        Blobs.from(manifest).writeTo(ByteStreams.nullOutputStream());
+    tarStreamBuilder.addByteEntry(
+        JsonTemplateMapper.toByteArray(manifest),
+        "blobs/sha256/" + manifestDescriptor.getDigest().getHash());
+
+    // Adds the oci-layout and index.json
+    tarStreamBuilder.addByteEntry(
+        "{\"imageLayoutVersion\": \"1.0.0\"}".getBytes(StandardCharsets.UTF_8), "oci-layout");
```
TadCordle: This looks a little hacky, but I'd rather just write this string directly than implement another `JsonTemplate`.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
```diff
@@ -16,17 +16,24 @@
 
 package com.google.cloud.tools.jib.docker;
```
loosebazooka: Also is `docker` the right package for this stuff? Should we have an oci package with an ImageTarball implementation?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -174,7 +177,7 @@ public static JibBuildRunner createJibBuildRunnerForRegistryImage(
         processCommonConfiguration(
             rawConfiguration, inferredAuthProvider, projectProperties, containerizer);
 
-    // Note Docker and tar builds don't set the configured format.
+    // Note Docker build doesn't set the configured format.
```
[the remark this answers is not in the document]
TadCordle: Last I tried, Docker still doesn't work with OCI.

jib-core/src/test/resources/core/fileA
```diff
@@ -1 +1 @@
-Crepe cakes are good.
+Crepe cakes are good.
```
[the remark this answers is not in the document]
TadCordle: Yeah, the file length on windows ended up being different from the extra `\r`, which made the manifest digest calculate to something different and the tests failed.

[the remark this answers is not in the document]
TadCordle: Now that I typed that out, maybe that's a problem.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
TadCordle: oh right, I was going to ask someone about this.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
TadCordle: Hmm yeah I should probably move this to `image` or `tar`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
```diff
@@ -71,6 +73,16 @@ public DescriptorDigest getDigest() {
...
     }
+
+    @VisibleForTesting
+    @Nullable
+    public Map<String, String> getAnnotations() {
+      return annotations;
+    }
+
+    void setAnnotations(Map<String, String> annotations) {
+      this.annotations = annotations;
```
chanseokoh: I think it's safer to copy the map.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/OCIIndexTemplate.java
```diff
@@ -0,0 +1,75 @@
...
+ *       }
+ *     }
+ *   ]
+ * }
+ * }</pre>
+ *
+ * @see <a href="https://github.com/opencontainers/image-spec/blob/master/image-index.md">OCI Image
+ *     Index Specification</a>
+ */
+public class OCIIndexTemplate implements JsonTemplate {
```
chanseokoh: This should probably be `OciIndexTemplate` per Google style? But I see we already have `OCIManifestTemplate`. I hope it's not late to rename it to `OciManifestTemplate`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/OCIIndexTemplate.java
```diff
@@ -0,0 +1,75 @@
...
+   * @param descriptor the manifest blob descriptor
+   * @param imageReferenceName the image reference name
+   */
+  public void addManifest(BlobDescriptor descriptor, String imageReferenceName) {
+    BuildableManifestTemplate.ContentDescriptorTemplate contentDescriptorTemplate =
+        new BuildableManifestTemplate.ContentDescriptorTemplate(
+            OCIManifestTemplate.MANIFEST_MEDIA_TYPE, descriptor.getSize(), descriptor.getDigest());
+    contentDescriptorTemplate.setAnnotations(
+        ImmutableMap.of("org.opencontainers.image.ref.name", imageReferenceName));
+    manifests.add(contentDescriptorTemplate);
```
chanseokoh: This makes this class mutable. Do other JSON templates work in this fashion as well?

TadCordle: Yes, `BuildableManifestTemplate` has an `addLayer` method.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/OCIIndexTemplate.java
TadCordle: Sure, I'll rename them both.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
TadCordle: https://github.com/opencontainers/image-spec/issues/796

jib-core/src/test/java/com/google/cloud/tools/jib/image/ImageTarballTest.java
```diff
@@ -132,4 +139,81 @@ public void testWriteTo()
...
+            ImageReference.parse("my/image:tag"),
+            ImmutableSet.of("tag", "another-tag", "tag3"));
+
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    imageToTarball.writeTo(out);
+    ByteArrayInputStream in = new ByteArrayInputStream(out.toByteArray());
+    try (TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(in)) {
+
+      // Verifies layer with fileA was added.
+      TarArchiveEntry headerFileALayer = tarArchiveInputStream.getNextTarEntry();
```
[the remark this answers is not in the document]
TadCordle: Afaik, a `TarArchiveEntry` only represents the metadata for an entry and not the contents themselves. So when we get the next `TarArchiveEntry`, we're essentially just getting the header (and then `CharStreams.toString(new InputStreamReader(tarArchiveInputStream, ...))` gets the rest).

chanseokoh:
Looks good. But Kokoro Mac is failing for an unrelated reason after our Kokoro change. It's a known issue for which Q implemented some workaround.

```
# Workaround for issue with calling 'docker login'. It defaults to using docker-credential-osxkeychain and errors with:
# Error saving credentials: error storing credentials - err: exit status 1, out: `User interaction is not allowed.`
# TODO: Follow-up with Kokoro about why this is happening.
rm /usr/local/bin/docker-credential-osxkeychain || true
```

TadCordle:
Regarding multiple tags, [this comment](https://github.com/opencontainers/image-spec/issues/796#issuecomment-561861980) says that it should already be possible. I opened an issue with podman [here](https://github.com/containers/libpod/issues/4646) since what I tested seems to be a problem with podman, not the OCI spec. So theoretically we can add this to Jib now, but I don't have a way of testing it yet.

chanseokoh:
I also find [this comment](https://github.com/opencontainers/image-spec/issues/796#issuecomment-561849236) interesting.

> The [containerd export code](https://github.com/containerd/containerd/blob/master/images/archive/exporter.go) now generates an archive which can both be loaded back into existing Docker versions (1.10+) as well as future version which will support loading the OCI archives directly.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/ImageTarball.java
chanseokoh: Let's file an issue on our repo to track this TODO, along with https://github.com/opencontainers/image-spec/issues/796 and https://github.com/containers/libpod/issues/4646.

TadCordle: Was planning on doing it after this got merged, but I can do it now.

TadCordle: https://github.com/GoogleContainerTools/jib/issues/2190

PR #2186, opened 2019-12-05
Fix Kokoro MacOS

### LINKED ISSUE #2189: Kokoro MacOS builds broken
On the new macOS Mojave, Kokoro builds are failing with

```
Command 'docker login localhost:5000 -u username --password-stdin' failed: Error saving credentials: error storing credentials - err: exit status 1, out: `error storing credentials - err: exec: "docker-credential-osxkeychain": executable file not found in $PATH, out: ```
```

This is because we delete `docker-credential-osxkeychain` as a workaround to side-step an issue that I'm going to explain later.
```
# Workaround for issue with calling 'docker login'. It defaults to using docker-credential-osxkeychain and errors with:
# Error saving credentials: error storing credentials - err: exit status 1, out: `User interaction is not allowed.`
# TODO: Follow-up with Kokoro about why this is happening.                                         
rm /usr/local/bin/docker-credential-osxkeychain || true
```
Not having `docker-credential-osxkeychain` hasn't cause an issue with the old macOS High Sierra, but for some reason, it looks like we must have `docker-credential-osxkeychain`. But of course, not deleting the file causes the error described in the code comment above on both High Sierra and Mojave.
```
Error saving credentials: error storing credentials - err: exit status 1, out: `User interaction is not allowed.`
```
Now, the reason for the error is that `docker login` attempts to access `docker-credential-osxkeychain` that is protected (probably) at the OS level. Only pre-approved application may modify the store, and our doing `docker login` is clearly not allowed.

On the Mojave image, I see the content of `~/.docker/config.json` is

```json
{
  "auths" : {

  },
  "credsStore" : "desktop",
  "stackOrchestrator" : "swarm"
}
```

Somehow, I think "desktop" and "osxkeychain" are interrelated regarding the "user interaction is not allowed" error, as hinted in https://github.com/docker/compose/issues/6517. I think there is a GUI interface in "Docker for Mac" (or "Docker Desktop for Mac"; I don't know if they are same or which one is being used on Kokoro if they are different) to make it not default to the protected credential store, but I cannot find a way to do it on the command line. And there does seem like a command to completely open up the protected store, but it requires a password to do so. I don't know the password.

Anyways, one working workaround is to physically replace `docker-credential-osxkeychain` and `docker-credential-desktop` (may not need to replace both) with something else.

```
cp $( which docker-credential-gcr ) $( which docker-credential-osxkeychain )
cp $( which docker-credential-gcr ) $( which docker-credential-desktop )
```

Another workaround is to clear "credsStore".

```
cat <<< '{"credsStore":"gcr"}' > "${HOME}/.docker/config.json"
```

Deleting the file `~/.docker/config.json` doesn't work, because it gets regenerated at random points. Omitting the `credsStore` field (i.e., `cat <<< '{}' > "${HOME}/.docker/config.json"`) didn't work either.

chanseokoh:
Fixes #2189.

### DIFF
--- a/kokoro/continuous.sh
+++ b/kokoro/continuous.sh
@@ -3,14 +3,17 @@
 set -e
 set -x
 
-gcloud components install docker-credential-gcr
+# On Mac, the default "credsStore" is set to "desktop". However, "desktop" is a
+# protected credential store, so "docker login" fails to modify it.
+# https://github.com/GoogleContainerTools/jib/issues/2189
+if [ "${KOKORO_JOB_CLUSTER}" = "MACOS_EXTERNAL" ]; then
+  cat <<< '{"credsStore":"gcr"}' > "${HOME}/.docker/config.json"
+fi
 
-# For macOS to find docker-credential-gcr
-export PATH=$PATH:/usr/local/Caskroom/google-cloud-sdk/latest/google-cloud-sdk/bin/
+gcloud components install docker-credential-gcr
 
 # docker-credential-gcr uses GOOGLE_APPLICATION_CREDENTIALS as the credentials key file
 export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_KEYSTORE_DIR}/72743_jib_integration_testing_key
-docker-credential-gcr configure-docker
 
 # Stops any left-over containers.
 docker stop $(docker ps --all --quiet) || true
--- a/kokoro/presubmit.sh
+++ b/kokoro/presubmit.sh
@@ -3,8 +3,14 @@
 set -e
 set -x
 
+# On Mac, the default "credsStore" is set to "desktop". However, "desktop" is a
+# protected credential store, so "docker login" fails to modify it.
+# https://github.com/GoogleContainerTools/jib/issues/2189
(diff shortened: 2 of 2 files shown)


chanseokoh:
Fixes #2189. This now works, for both presubmit and continuous Kokoro MacOS builds.

kokoro/continuous.sh
```diff
@@ -3,14 +3,17 @@
 set -e
 set -x
 
-gcloud components install docker-credential-gcr
+# On Mac, the default "credsStore" is set to "desktop". However, "desktop" is a
+# protected credential store, so "docker login" fails to modify it.
+# https://github.com/GoogleContainerTools/jib/issues/2189
+if [ "${KOKORO_JOB_CLUSTER}" = "MACOS_EXTERNAL" ]; then
```
loosebazooka: should we file a bug against kokoro? or is that working as intended?

chanseokoh: I don't actually know, but perhaps it's a sane request to do something about this, so that `docker login` just works. I'll file a bug.

Issue #2140, opened 2019-11-11
Minimum gradle version -> 5.0 (for jib 2.0)


loosebazooka:
Would probably help keep us compatible with 6.0+, and the last 4.X release was basically a year ago.

PR #2188, opened 2019-12-05
Use correct configuration for Skaffold dependencies

lightoze:
Because configurations were changed in Gradle 5, existing implementation fails to detect dependencies between project modules.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2.java
@@ -36,6 +36,7 @@ import org.gradle.api.artifacts.Dependency;
 import org.gradle.api.artifacts.ProjectDependency;
 import org.gradle.api.artifacts.PublishArtifact;
 import org.gradle.api.initialization.Settings;
+import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskAction;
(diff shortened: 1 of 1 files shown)


lightoze:
@googlebot I signed it!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2.java
```diff
@@ -176,7 +177,10 @@ private void addProjectFiles(Project project) {
 
       // Search through all dependencies
       for (Configuration configuration :
-          currentProject.getConfigurations().getByName("runtime").getHierarchy()) {
+          currentProject
+              .getConfigurations()
+              .getByName(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME)
```
chanseokoh: @lightoze in this class, there's another use of `getByName("runtime")` for SNAPSHOT, non-project dependencies. Does it need to be changed as well?

```java
    for (File file : project.getConfigurations().getByName("runtime")) {
      if (!projectDependencyJars.contains(file) && file.toString().contains("SNAPSHOT")) {
```

And you said this is broken with Gradle 5. After changing this, will it continue to work with Gradle 4.9? (Gradle 4.9 is the minimum version Jib supports/requires.)

@loosebazooka looks like the value is now changed to "runtimeClasspath" from "runtime." Does this make sense?

lightoze: @chanseokoh probably that should be changed too, but it does not affect the project I have on hands so I have nothing to test on. As such "runtime" is deprecated and should be used if only for compatibility.
Gradle 4 should continue to work but I haven't tested.

loosebazooka: Yeah I think with 2.0 we plan on going to minimum of gradle 5 anyway: https://github.com/GoogleContainerTools/jib/issues/2140

loosebazooka: Also, yeah I agree we should use `runtimeClasspath` everywhere...

lightoze: @chanseokoh @loosebazooka OK changed that too and reworked patch into a single commit.

chanseokoh:
I'll temporarily override the kokoro-macos failure (due to #2189) and merge this.

PR #2191, opened 2019-12-05
Fix Kokoro Ubuntu

chanseokoh:
I removed this in #2186, since it worked on macOS without it. Turns out, this is still required on Ubuntu. I confirm this works.


chanseokoh:
Should work. We will see...

PR #2192, opened 2019-12-05
Use a different marker so we don't clash with other jib json outputs

loosebazooka:
This is to potentially protect us from the case where we run sync and files tasks together.


loosebazooka:
> Nevermind, I guess that'd break backwards compatibility.

Oh I didn't realize init had one too :\. But, yeah it's probably worth coming up with a strategy for this eventually.

briandealwis:
Could we make this header a bit more easy to identify with a common prefix, and future-proofed by embedding a version?
```
BEGIN JIB JSON: SYNCMAP v1
```

So we can scan looking for `BEGIN JIB JSON` and if we're missing a colon then it indicates an older Jib.

We can now force a minimum Jib version, but that's quite disruptive.  It might be worth looking at HTTP and how it handles version negotiation.

loosebazooka:
Updating this with new style, PTAL

Issue #2050, opened 2019-10-09
Cannot pull nginx. Unauthorized for registry-1.docker.io/library/nginx: 401 Unauthorized


nikhileshva:
**Environment**:

- *Jib version: 1.6.1
- *Build tool: Maven
- *OS: CentOS (Building maven on Jenkins)

**Description of the issue**:
```
Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.6.1:build (default-cli) on project remoteportal.client: Build image failed, perhaps you should make sure your credentials for 'registry-1.docker.io/library/nginx' are set up correctly. See https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#what-should-i-do-when-the-registry-responds-with-unauthorized for help: Unauthorized for registry-1.docker.io/library/nginx: 401 Unauthorized
```

**Expected behavior**:
It should pull the public image without needing authorisation. 

**Steps to reproduce**:

  1. I'm running this on a centOS instance with Jenkins pipeline. 
  2. It works on local (Docker for Mac)
  3. It worked for me a week back, because I have an image of this app in ECR. 

**`jib-maven-plugin` Configuration**: 
```xml
<plugin>
                <groupId>com.google.cloud.tools</groupId>
                <artifactId>jib-maven-plugin</artifactId>
                <configuration>
                    <from>
                        <image>nginx</image>
                    </from>
                    <to>
                        <image>${ecrEndpoint}/frontend-${docker-artifact-id}</image>
                        <tags>
                            <tag>${project.version}</tag>
                        </tags>
                    </to>
                    <container>
                        <ports>
                            <port>80</port>
                            <port>443</port>
                        </ports>
                        <entrypoint>
                            <command>nginx</command>
                            <arg>-g</arg>
                            <arg>daemon off;</arg>
                        </entrypoint>
                    </container>
                    <extraDirectories>
                        <paths>${project.build.directory}/it-temp/frontend-dist/</paths>
                    </extraDirectories>
                    <skip>false</skip>
                </configuration>
            </plugin>
```

**Log output**: 

**Additional Information**: 
I'm running the `mvn jib:build` in the Jenkins pipeline.

chanseokoh:
[...]

PR #2200, opened 2019-12-17
Display where Jib got credentials for base and target images

### LINKED ISSUE #2056: Display if credentials were successfully retrieved and from which source
From #2046.

Auth has the most friction, and it will help a lot to report if Jib picked up some credentials and where if so. It will help cases like #2050 too.

chanseokoh:
Fixes #2056.

```
[INFO] Using credentials from Docker config (/usr/local/home/guest/.docker/config.json) for francium25/test
```
```
[INFO] Using credential helper docker-credential-gcr for gcr.io/project/repo
```
```
[INFO] Using credentials from Maven settings file for gcr.io/project/repo
```
```
[INFO] Using credentials from <from><auth> for gcr.io/project/repo
```
```
[INFO] Using credentials from from.auth for gcr.io/project/repo
```

Overall, it will look like
```
[INFO] Containerizing application to francium25/test...
[WARNING] Base image 'gcr.io/chanseok-playground-new/test' does not use a specific image digest - build may not be reproducible
[INFO] Using credentials from Docker config (/usr/local/home/guest/.docker/config.json) for francium25/test
[INFO] The base image requires auth. Trying again for gcr.io/<redacted>...
[INFO] Using credentials from <from><auth> for gcr.io/<redacted>
[INFO] Executing tasks:
[INFO] [===============               ] 50.0% complete
[INFO] > building image to registry
```

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
@@ -84,10 +84,8 @@ class RetrieveRegistryCredentialsStep implements Callable<Optional<Credential>> 
         }
       }
 
-      // If no credentials found, give an info (not warning because in most cases, the base image is
-      // public and does not need extra credentials) and return empty.
       eventHandlers.dispatch(
-          LogEvent.info("No credentials could be retrieved for registry " + registry));
+          LogEvent.lifecycle("No credentials could be retrieved for registry " + registry));
       return Optional.empty();
     }
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
@@ -85,7 +85,7 @@ class RetrieveRegistryCredentialsStep implements Callable<Optional<Credential>> 
       }
 
       eventHandlers.dispatch(
-          LogEvent.lifecycle("No credentials could be retrieved for registry " + registry));
+          LogEvent.info("No credentials could be retrieved for registry " + registry));
       return Optional.empty();
     }
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
@@ -255,15 +255,16 @@ public class CredentialRetrieverFactory {
   CredentialRetriever dockerConfig(
       DockerConfigCredentialRetriever dockerConfigCredentialRetriever) {
     return () -> {
+      Path configFile = dockerConfigCredentialRetriever.getDockerConfigFile();
       try {
         Optional<Credential> credentials = dockerConfigCredentialRetriever.retrieve(logger);
         if (credentials.isPresent()) {
-          logGotCredentialsFrom("credentials from Docker config file");
+          logGotCredentialsFrom("credentials from Docker config (" + configFile + ")");
           return credentials;
         }
 
       } catch (IOException ex) {
-        logger.accept(LogEvent.info("Unable to parse Docker config file"));
+        logger.accept(LogEvent.info("Unable to parse Docker config file: " + configFile));
       }
       return Optional.empty();
     };
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
@@ -110,7 +110,7 @@ public class CredentialRetrieverFactory {
    */
   public CredentialRetriever known(Credential credential, String credentialSource) {
     return () -> {
-      logGotCredentialsFrom(credentialSource);
+      logGotCredentialsFrom("credentials from " + credentialSource);
       return Optional.of(credential);
     };
   }
@@ -258,12 +258,12 @@ public class CredentialRetrieverFactory {
       try {
         Optional<Credential> credentials = dockerConfigCredentialRetriever.retrieve(logger);
         if (credentials.isPresent()) {
-          logGotCredentialsFrom("credentials from Docker config");
+          logGotCredentialsFrom("credentials from Docker config file");
           return credentials;
         }
 
       } catch (IOException ex) {
-        logger.accept(LogEvent.info("Unable to parse Docker config"));
+        logger.accept(LogEvent.info("Unable to parse Docker config file"));
       }
       return Optional.empty();
     };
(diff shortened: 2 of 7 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -84,10 +84,8 @@ static RetrieveRegistryCredentialsStep forTargetImage(
         }
       }
 
-      // If no credentials found, give an info (not warning because in most cases, the base image is
-      // public and does not need extra credentials) and return empty.
       eventHandlers.dispatch(
-          LogEvent.info("No credentials could be retrieved for registry " + registry));
+          LogEvent.lifecycle("No credentials could be retrieved for registry " + registry));
```
TadCordle: Does this mean this message will be printed every time for public repositories, or is the removed comment outdated?

chanseokoh: We could leave this as `info`. I just thought calling this out explicitly outweighs adding one more line to the output. It may help in the case where the user hits unauthorized and mistakenly thinks it's failing because of a wrong username or password while in reality Jib couldn't retrieve any credentials. For example, the user typed `-Djib.auth.password` instead of `-Djib.from.auth.password` or put `<auth><username>` outside `<from>`, and it will be easier and less frustrating to identify this mistake themselves. But if you think this change does more harm than good, I'm fine going back.

But to really answer your question, the line will get printed for public images on many repositories (e.g., Docker Hub for sure, and maybe either Quay, OpenShift, or both―I can't remember). But interestingly, GCR doesn't get to this line. This is because we first attempt pulling without credentials (i.e., pulling before calling this method) and GCR succeeds at the first attempt for public images.

As a side note, this first attempt without credentials may not necessarily be a good thing, considering that it is OK to send unrelated or wrong credentials when pulling public GCR images. It's then actually a bad thing for registries like Docker Hub for the reason I'll explain right below.

However, as I said, for registries like Docker Hub, unfortunately this line is always printed, as it requires going through the registry token auth flow for any image. (That is, the first attempt without credentials will always fail.) And when we do the token auth flow, our current code always retrieves and passes credentials, even for public images that don't require credentials. (Another side note: this can cause a problem like https://github.com/GoogleContainerTools/jib/issues/2050#issuecomment-553347560 for public images.)

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
```diff
@@ -258,12 +258,12 @@ CredentialRetriever dockerConfig(
...
         Optional<Credential> credentials = dockerConfigCredentialRetriever.retrieve(logger);
         if (credentials.isPresent()) {
-          logGotCredentialsFrom("credentials from Docker config");
+          logGotCredentialsFrom("credentials from Docker config file");
           return credentials;
         }
 
       } catch (IOException ex) {
-        logger.accept(LogEvent.info("Unable to parse Docker config"));
+        logger.accept(LogEvent.info("Unable to parse Docker config file"));
```
loosebazooka: perhaps the actual config file, in these cases?

"credentials from Docker config file: /path/to/config/file" <-- and maybe the type too? cred helper? auths?
"Unable to parse docker config file: /path/to/config/file"

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
loosebazooka: I vaguely remember somewhere that people were confused by this message for public repos. I could be wrong though.

chanseokoh: That's a fair point. I will leave the current log level. The issue I described can be handle by adding some line to our FAQ.

jib-core/src/main/java/com/google/cloud/tools/jib/frontend/CredentialRetrieverFactory.java
chanseokoh: Sounds good. Made them print the path to a Docker config file.

As for the type, I also thought that would be super useful to show here, but unfortunately it's a bit involved to have that information here. That's why I took the compromise to log that type information in `DockerConfigCredentialRetriever` as INFO. Increasing the log level will reveal the type.

loosebazooka: ah yeah I didn't see that.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
```diff
@@ -122,6 +124,8 @@ public DockerConfigCredentialRetriever(String registry, Path dockerConfigFile) {
             new String(Base64.decodeBase64(auth), StandardCharsets.UTF_8);
         String username = usernameColonPassword.substring(0, usernameColonPassword.indexOf(":"));
         String password = usernameColonPassword.substring(usernameColonPassword.indexOf(":") + 1);
+        logger.accept(
+            LogEvent.debug("Docker config auths section defines credentials for " + registryAlias));
```
loosebazooka: I think we need to add a test for these two debug statements?

chanseokoh: Added tests. I increased the level to INFO from DEBUG, as I think you usually pass `--info` for Gradle.

loosebazooka: Oh does that make our tests too noisy?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/RetrieveRegistryCredentialsStep.java
```diff
@@ -84,8 +84,6 @@ static RetrieveRegistryCredentialsStep forTargetImage(
         }
       }
 
-      // If no credentials found, give an info (not warning because in most cases, the base image is
-      // public and does not need extra credentials) and return empty.
```
loosebazooka: should we leave this comment in?

chanseokoh: I intend to remove this. Doesn't seem to help much. For example, when looking at this line, you will not question why this is `info()` instead of `warning()`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
chanseokoh: I don't think it will. Maybe just a little bit more lines.

PR #2202, opened 2019-12-18
Improve authentication when pulling base image

### LINKED ISSUE #2134: Optimize registry authentication
We are aware of this issue, but we've never filed an issue.

An example where we do sub-optimal registry interactions:

```
# requesting a manifest
> GET https://registry.redhat.io/v2/redhat-openjdk-18/openjdk18-openshift/manifests/latest

# Server responses 401. It already includes WWW-Authenticate.
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Bearer realm="https://registry.redhat.io/auth/realms/rhcc/protocol/redhat-docker-v2/auth",service="docker-registry"

# repeating the same manifest request for no good reason
> GET https://registry.redhat.io/v2/redhat-openjdk-18/openjdk18-openshift/manifests/latest

# same 401 response
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Bearer realm="https://registry.redhat.io/auth/realms/rhcc/protocol/redhat-docker-v2/auth",service="docker-registry"

# starting registry token auth flow, starting with <registry url>/v2/
> GET https://registry.redhat.io/v2/

# server responding with 401 with the same WWW-Authenticate
< HTTP/1.1 401 Unauthorized
< WWW-Authenticate: Bearer realm="https://registry.redhat.io/auth/realms/rhcc/protocol/redhat-docker-v2/auth",service="docker-registry"
```

In this example, the server returns `WWW-Authenticate` on the first manifest request. Not every server may return it at this point, but in this case, we could have avoided having to send two more GET requests.

chanseokoh:
Closes #2134.

I think this is still not an ideal sequence for auth and we can actually simplify this a lot. But as mentioned in the issue, I don't want to change the basic flow of the current behavior even if it is inefficient. This PR will basically add a new shortcut that in practice I expect Jib will take for most registries, so I think this helps a lot. The shortcut is to take and use `WWW-Authenticate` from the first failure. It will cut down 2 HTTP round trips.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -175,7 +175,7 @@ class PullBaseImageStep implements Callable<ImageAndAuthorization> {
         }
 
         // Last resort. Access "<registry URL>/v2/" and hope the registry will return
-        // "WWW-Authenticate: Bearer ...".
+        // "WWW-Authenticate: Bearer ...". Perhaps it's rare to reach here.
         RegistryAuthenticator registryAuthenticator =
             registryClient.getRegistryAuthenticator().orElseThrow(() -> ex);
         Authorization authorization = registryAuthenticator.authenticatePull(credentials);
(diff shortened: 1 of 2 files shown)


chanseokoh:
This is in conflict with #2233, so I'll mark `PR: Not Ready`.

chanseokoh:
Closing in favor of #2789.

PR #2203, opened 2019-12-18
Improve FAQ on auth failure

chanseokoh:
Explain log output coming from #2200.

Should be merged after a new release.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -504,19 +504,19 @@ If the registry returns `401 Unauthorized` or `"code":"UNAUTHORIZED"`, it is oft
 Note, if Jib was able to retrieve credentials, you should see a log message like these:
 
 ```
-[INFO] Using credentials from Docker config (/home/user/.docker/config.json) for localhost:5000/java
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -501,6 +501,24 @@ If the registry returns `401 Unauthorized` or `"code":"UNAUTHORIZED"`, it is oft
 * Typos in username, password, image names, or registry names.
 * You are using a private registry without HTTPS. See [How can I diagnose problems pulling or pushing from remote registries?](#how-can-i-diagnose-problems-pulling-or-pushing-from-remote-registries).
 
+Note, if Jib was able to retrieve credentials, you should see a log message like these:
```
TadCordle: Should we clarify that this is at the `info` level? The `[INFO]` tag is there but maven uses it even for lifecycle messages.

chanseokoh: I just moved the `[INFO]` prefix. This should create no confusion now. These messages will always be printed (if Jib retrieved credentials) regardless of Maven or Gradle.

Issue #2193, opened 2019-12-06
Automatic update checks


TadCordle:
Every so often (once a week?) when Jib runs, it should check if there's a later version available and show a message to update, similar to how skaffold does update checks.

PR #2211, opened 2020-01-03
Add method for retrieving XDG config directory

TadCordle:
Part of #2193. Adds methods for retrieving the XDG_CONFIG_HOME directory, which will hold update-check-related files.

The full paths are:
* Linux: `$HOME/.config/google-cloud-tools-java/jib/`
* Mac: `$HOME/Library/Preferences/Google/Jib/`
* Windows: `%LOCALAPPDATA%\Google\Jib\Config`

One potential issue is that this may end up overlapping with the base image cache directory on windows, so if someone clears their cache without being careful, they may also delete their update check config accidentally.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
@@ -26,7 +26,7 @@ import java.util.Properties;
 import java.util.logging.Logger;
 
 /**
- * Obtains an OS-specific directories based on the XDG Base Directory Specification.
+ * Obtains OS-specific directories based on the XDG Base Directory Specification.
  *
  * <p>Specifically, from the specification:
  *
@@ -43,7 +43,10 @@ import java.util.logging.Logger;
  */
 public class XdgDirectories {
 
-  private static final Logger logger = Logger.getLogger(XdgDirectories.class.getName());
+  private static final Logger LOGGER = Logger.getLogger(XdgDirectories.class.getName());
+  private static final Path JIB_SUBDIRECTORY_LINUX =
+      Paths.get("google-cloud-tools-java").resolve("jib");
+  private static final Path JIB_SUBDIRECTORY_OTHER = Paths.get("Google").resolve("Jib");
 
   public static Path getCacheHome() {
     return getCacheHome(System.getProperties(), System.getenv());
@@ -84,12 +87,12 @@ public class XdgDirectories {
       // Use %LOCALAPPDATA% for Windows.
       String localAppDataEnv = environment.get("LOCALAPPDATA");
       if (localAppDataEnv == null || localAppDataEnv.trim().isEmpty()) {
-        logger.warning("LOCALAPPDATA environment is invalid or missing");
+        LOGGER.warning("LOCALAPPDATA environment is invalid or missing");
         return xdgPath;
       }
       Path localAppData = Paths.get(localAppDataEnv);
       if (!Files.exists(localAppData)) {
-        logger.warning(localAppData + " does not exist");
+        LOGGER.warning(localAppData + " does not exist");
         return xdgPath;
       }
       return localAppData;
@@ -98,7 +101,7 @@ public class XdgDirectories {
       // Use '~/Library/Application Support/' for macOS.
       Path applicationSupport = Paths.get(userHome, "Library", "Application Support");
       if (!Files.exists(applicationSupport)) {
-        logger.warning(applicationSupport + " does not exist");
+        LOGGER.warning(applicationSupport + " does not exist");
         return xdgPath;
       }
       return applicationSupport;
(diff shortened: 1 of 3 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -26,27 +26,32 @@
...
  * <p>Specifically, from the specification:
  *
  * <ul>
- *   <li>This directory is defined by the environment variable {@code $XDG_CACHE_HOME}.
- *   <li>If {@code $XDG_CACHE_HOME} is either not set or empty, a default equal to {@code
- *       $HOME/.cache} should be used.
+ *   <li>These directories are defined by the environment variables {@code $XDG_CACHE_HOME} and
+ *       {@code $XDG_CONFIG_HOME}.
+ *   <li>If {@code $XDG_CACHE_HOME} / {@code $XDG_CONFIG_HOME} is either not set or empty, a default
+ *       equal to {@code $HOME/.cache} / {@code $HOME/.config} should be used.
```
briandealwis:
```diff
- a default equal to 
+ a platform-specific equivalent of 
```

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -101,5 +106,59 @@ static Path getCacheHome(Properties properties, Map<String, String> environment)
...
+      return Paths.get(xdgConfigHome);
+    }
+
+    String userHome = properties.getProperty("user.home");
+    Path xdgPath = Paths.get(userHome, ".config");
+
+    String rawOsName = properties.getProperty("os.name");
+    String osName = rawOsName.toLowerCase(Locale.ENGLISH);
+
+    if (osName.contains("linux")) {
```
briandealwis: We're now duplicating some of this platform-testing logic.  Is it worth extracting into separate `getOs(properties)` method that returns the platform type?  (I wish the JRE already provided this.)

TadCordle:
>And the macOS cache directory really should be Library/Caches

Yeah, I wonder if we should think about a way of moving the cache without breaking users.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
TadCordle: Might be a good idea if we ever need it more. But I think with how little logic it actually is, only duplicating in 2 places isn't so bad.

loosebazooka:
I don't think we'd break anyone except those that have some code to auto clear cache. We could do something like: Identify if an old cache exist -- tell the user to delete it

TadCordle:
Filed #2216

TadCordle:
Since we're thinking about changing the cache directory on different platforms, I'll update this PR so the `getConfigHome` returns the full path to the config for each platform and not just the home (e.g. `$XDG_CONFIG_HOME/google-cloud-tools-java/jib` rather than just `$XDG_CONFIG_HOME`). Then maybe make the cache changes in another PR.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,174 @@
...
+   *
+   * <p>For Linux, this is {@code $HOME/.config/google-cloud-tools-java/jib/}.
+   *
+   * <p>For Windows, this is {@code %LOCALAPPDATA%\Google\Jib/}.
+   *
+   * <p>For macOS, this is {@code $HOME/Library/Preferences/Google/Jib/}.
+   */
+  @VisibleForTesting
+  static Path getConfigHome(Properties properties, Map<String, String> environment) {
+    Path windowsSubDirectory = JIB_SUBDIRECTORY_OTHER.resolve("Config");
```
chanseokoh: Why not having `windowsSubDirectory` as a constant?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,174 @@
...
+      return localAppData.resolve(windowsSubDirectory);
+
+    } else if (osName.contains("mac") || osName.contains("darwin")) {
+      // Use '~/Library/Preferences/' for macOS.
+      Path applicationSupport = Paths.get(userHome, "Library", "Preferences");
+      if (!Files.exists(applicationSupport)) {
+        LOGGER.warning(applicationSupport + " does not exist");
+        return xdgPath.resolve(JIB_SUBDIRECTORY_OTHER);
+      }
+      return applicationSupport.resolve(JIB_SUBDIRECTORY_OTHER);
```
chanseokoh: Looks like this structure is being repeated.
```
if (linux) {
   return path.resolve(JIB_SUBDIRECTORY_LINUX);
} else if (windows) {
   return path.resolve(windowsSubDirectory);
} else /* if (mac) {
   return path.resolve(JIB_SUBDIRECTORY_OTHER);
}
```
Perhaps this logic can be a separate method, like `getConfigHome(getBaseConfigHome(...))` (whatever suitable method names)? If this is indeed repeated, I think factoring this out will make the code more readable.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
TadCordle: It's different for cache and config.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
TadCordle: Sure, I'll give it a try.

TadCordle:
Closing in favor of #2218

PR #2212, opened 2020-01-03
Ignore registry port when inferring auth from maven settings

### LINKED ISSUE #2135: Maven settings.xml auth matching
**Environment**:

- *Jib version:* 1.7.0
- *Build tool:* maven 3.6.2
- *OS:* macOS and gitlab CI

**Description of the issue**:
Docker image tagging: `registry:port/image:version`
Maven expects `<server><id>` and `<repositories><repository><id>` to match.
https://maven.apache.org/settings.html
Problem is that jib expects to match exactly these with `repository:port` part of the docker image tagging.
Maven doesn't like the `:port` - `:` part - at all.

**Expected behavior**:
Matching the previous mentioned settings.xml file entry/key without `:port` part.
Not produce maven to output warning regarding special character... (see log output below)

**Steps to reproduce**:

  1. edit your ~/.m2/settings.xml file with desired and preferred registry `host:port` key. For server.id and registries.registry.id keys.
  2. run `mvn jib:build` or `mvn {any_goal}`
  3. 

**Log output**: 

```
[WARNING] Some problems were encountered while building the effective model for common:jar:HEAD-SNAPSHOT
[WARNING] 'repositories.repository.id' must not contain any of these characters \/:"<>|?* but found :
```

**Additional Information**: 

I suggest to relax the auth retrieval with maven to not use the `:port` part of the docker image tagging.
This can let the user chose whatever port the registry he/she wants to open.
Otherwise you have to put a double entry for server.id part only to please jib... meaning with the `:port` included.

padyx:
This PR is a suggestion to fix #2135 

It changes the lookup of the server information from the Maven settings to ignore the port information from the registry.

It splits the registry information at the (last) colon and extracts the substring before it:
`docker.example.com:12345` would match with the server id `docker.example.com`

I've tried to run as many integration tests as possible (but those working with the docker daemon didn't run on my Fedora 31 with podman).

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
@@ -114,12 +114,12 @@ class MavenSettingsServerCredentials implements InferredAuthProvider {
   @Nullable
   @VisibleForTesting
   Server getServerFromMavenSettings(String registry) {
-    // Find and remove port (if present)
     Server server = settings.getServer(registry);
     if (server != null) {
       return server;
     }
 
+    // try without port
     int index = registry.lastIndexOf(':');
     if (index != -1) {
       return settings.getServer(registry.substring(0, index));
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
@@ -19,7 +19,9 @@ package com.google.cloud.tools.jib.maven;
 import com.google.cloud.tools.jib.plugins.common.AuthProperty;
 import com.google.cloud.tools.jib.plugins.common.InferredAuthException;
 import com.google.cloud.tools.jib.plugins.common.InferredAuthProvider;
+import com.google.common.annotations.VisibleForTesting;
 import java.util.Optional;
+import javax.annotation.Nullable;
 import org.apache.maven.settings.Server;
 import org.apache.maven.settings.Settings;
 import org.apache.maven.settings.building.SettingsProblem;
@@ -58,7 +60,7 @@ class MavenSettingsServerCredentials implements InferredAuthProvider {
   @Override
   public Optional<AuthProperty> inferAuth(String registry) throws InferredAuthException {
 
-    Server server = settings.getServer(registry);
+    Server server = getServerFromMavenSettings(registry);
     if (server == null) {
       return Optional.empty();
     }
@@ -107,5 +109,21 @@ class MavenSettingsServerCredentials implements InferredAuthProvider {
             return CREDENTIAL_SOURCE;
           }
         });
+  }
+
+  @Nullable
+  @VisibleForTesting
+  Server getServerFromMavenSettings(String registry) {
+    // Find and remove port (if present)
+    Server server = settings.getServer(registry);
+    if (server != null) {
+      return server;
+    }
+
+    int index = registry.lastIndexOf(':');
+    if (index != -1) {
+      return settings.getServer(registry.substring(0, index));
+    }
+    return null;
   }
 }
--- a/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentialsTest.java
+++ b/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentialsTest.java
@@ -87,6 +87,34 @@ public class MavenSettingsServerCredentialsTest {
   }
 
   @Test
+  public void testInferredAuth_registryWithHostAndPort() throws InferredAuthException {
+    Optional<AuthProperty> auth =
+        mavenSettingsServerCredentialsNoMasterPassword.inferAuth("docker.example.com:8080");
+    Assert.assertTrue(auth.isPresent());
+    Assert.assertEquals("registryUser", auth.get().getUsername());
+    Assert.assertEquals("registryPassword", auth.get().getPassword());
+  }
+
+  @Test
+  public void testInferredAuth_registryWithHostWithoutPort() throws InferredAuthException {
+    Optional<AuthProperty> auth =
+        mavenSettingsServerCredentialsNoMasterPassword.inferAuth("docker.example.com");
+    Assert.assertTrue(auth.isPresent());
+    Assert.assertEquals("registryUser", auth.get().getUsername());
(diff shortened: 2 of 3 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -58,7 +59,7 @@
   @Override
   public Optional<AuthProperty> inferAuth(String registry) throws InferredAuthException {
 
-    Server server = settings.getServer(registry);
+    Server server = getServerFromMavenSettings(registry);
     if (server == null) {
```
loosebazooka: It'd be nice if this `server == null` code could go into getServerFromMavenSettings

chanseokoh:
Hmm... perhaps we should ignore ports only after there was no exact match with registry port. The motivation to ignore ports was for the corner case where the user uses both `<server>` and `<repository>`, where you cannot have the `:` characters in `<repository><id>`.

padyx:
> Hmm... perhaps we should ignore ports only after there was no exact match with registry port. The motivation to ignore ports was for the corner case where the user uses both `<server>` and `<repository>`, where you cannot have the `:` characters in `<repository><id>`.

Sounds good to me: The first attempt will be an exact match, and if nothing is found, the second one will be attempted without the port.

I'll rework this and push an updated version shortly.

padyx:
> Also curious, is it standard practice to put port in the maven settings.xml?

As far as I know, it's not a very common case: The `id` usually is an arbitrary identifier - it only needs to match between server and repository (or mirror). Apart from that it usually can be chosen at will and maven will just attempt to contact all repositories.
Jib is a bit special here, since the imageName in the `from` and `to` section already define which repository we contact. The maven-deploy-plugin which also publishes to a remote repository, [uses the usual combination of `id` and `repository`](https://maven.apache.org/plugins/maven-deploy-plugin/usage.html) (though in a special section).

Another implementation variant could be to configure the id of the authentication we want to use explicitly in the maven POM. Something like this (identical option for `from` section, but omitted)

```
<configuration>
  ...
  <to>
    <image>gcr.io/my-gcp-project/my-app</image>
    <auth>
      <id>myrepository</id>
    </auth>
  </to>
  ...
</configuration>
```

What do you think about something like this? (Exact configuration TBD)

chanseokoh:
I understand it's a bit weird that `<server><id>` should be a host string for Jib rather than being an arbitrary ID, but I don't like adding a new configuration option to take care of this. I think this is still a minor issue, and there's already a simple workaround for #2135: just define an additional (albeit duplicate) entry for `<server><id>` with the exact match in that case. I think the current Jib behavior works more or less. Ignored ports when there was no match seems enough.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
padyx: Could you elaborate? I can't see an option right now to implement that: There could, or could not be, a server returned.
If I switch this to use `Optional`, I would have to swap this out with an `isPresent` check instead to continue with the decryption below...

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -108,4 +109,23 @@ public String getPasswordDescriptor() {
           }
         });
   }
+
+  @Nullable
+  Server getServerFromMavenSettings(String registry) {
+    // Find and remove port (if present)
+    if (registry == null) {
+      return null;
+    }
```
chanseokoh: You don't need this null check. We check nullability with NullAway, and unless the argument is marked with `@Nullable`, it is never null.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -108,4 +109,23 @@ public String getPasswordDescriptor() {
           }
         });
   }
+
+  @Nullable
```
chanseokoh: Please add `@VisibleForTesting`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -108,4 +109,23 @@ public String getPasswordDescriptor() {
...
+    Server server = settings.getServer(registry);
+    if (server != null) {
+      return server;
+    }
+
+    int index = registry.lastIndexOf(':');
+    if (index != -1) {
+      registry = registry.substring(0, index);
+    }
+    return settings.getServer(registry);
```
chanseokoh: I suggest

```java
    int index = registry.lastIndexOf(':');
    if (index != -1) {
      return settings.getServer(registry.substring(0, index));
    }
    return null;
```
No need to call `settings.getServer()` twice if nothing is configured.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
chanseokoh: Makes sense. I think we can leave this null check here.

padyx:
>  but I don't like adding a new configuration option to take care of this.

No problem, it was only an option in case that a new configuration would better align with the project goals/architecture 😃

I've push to fix the UnitTests and it should be ready for review again.

chanseokoh:
@padyx please see my most recent reviews!

padyx:
@chanseokoh Thanks for your review! I've amended the commit with the suggested changes.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenSettingsServerCredentials.java
```diff
@@ -108,4 +110,20 @@ public String getPasswordDescriptor() {
           }
         });
   }
+
+  @Nullable
+  @VisibleForTesting
+  Server getServerFromMavenSettings(String registry) {
+    // Find and remove port (if present)
```
chanseokoh: Oh, this comment is now confusing. Let's update and move this to line 123:

```java
// try without port
int index = registry.lastIndexOf(':');
```

(Sorry, I should have caught this earily.)

chanseokoh: Never mind. I just did it myself.

PR #2214, opened 2020-01-06
Implement update checks

### LINKED ISSUE #2193: Automatic update checks

TadCordle:
Fixes #2193.

A couple changes from the initial design:

- Instead of getting/setting the last modified time of a file to check the last update time, the time of the previous update check is stored as contents of the file. It seems `Files.setLastModifiedTime()` isn't supported on all platforms.
- The file pulled from GCS contains JSON instead of just the version string, for extensibility.
- The check is daily instead of weekly.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildTarMojo.java
```diff
@@ -65,6 +68,12 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     MavenProjectProperties projectProperties =
         MavenProjectProperties.getForProject(
             getProject(), getSession(), getLog(), tempDirectoryProvider);
+    UpdateChecker updateChecker =
+        UpdateChecker.checkForUpdate(
+            projectProperties.isOffline() || !getLog().isInfoEnabled(),
+            projectProperties.getVersion(),
```
chanseokoh: I believe this returns the project version, like `<version>0.0.1-SNAPSHOT</version>` in `pom.xml`. We have `ProjectInfo.VERSION` (which works only when the code is packaged as a JAR).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,175 @@
...
+      boolean skip, String currentVersion, String versionUrl, Path configDir) {
+    // Abort if offline or update checks are disabled
+    if (skip || Boolean.getBoolean(PropertyNames.DISABLE_UPDATE_CHECKS)) {
+      return Optional.empty();
+    }
+
+    try {
+      Path configFile = configDir.resolve("config.json");
+      if (Files.exists(configFile)) {
+        // Abort of update checks are disabled
```
chanseokoh: Abort *if*

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,175 @@
...
+          if (modifiedTime.plus(Duration.ofDays(1)).isAfter(Instant.now())) {
+            return Optional.empty();
+          }
+        } catch (DateTimeParseException ex) {
+          // Ignore parse failure; assume update check hasn't been performed
+        }
+      }
+
+      // Check for update
+      URLConnection connection = new URL(versionUrl).openConnection();
```
chanseokoh: This is not really necessary given the circumstances, but since it's not a big deal to do it, let's have `HttpURLConnection connection = (HttpURLConnection) ...` and call `connection.disconnect()` at the end.

And there should be some way to set some timeout values with `HttpURLConnection`. It's worth setting them a small value.

And can you verify Maven and Gradle both terminate right away even if this thread is alive, e.g., by putting `Thread.sleep(10000)` here?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,175 @@
...
+      }
+      return Optional.of(
+          "A new version of Jib ("
+              + latestVersion
+              + ") is available (currently using "
+              + currentVersion
+              + "). Update your build configuration to use the latest features and fixes!");
+
+    } catch (IOException ignored) {
+      // Fail silently
```
chanseokoh: How about attempting to delete both the config file and `lastUpdateCheck`? That can fix the situation if any of the files is corrupt.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,175 @@
...
+    return new UpdateChecker(
+        executorService.submit(
+            () -> performUpdateCheck(skip, version, versionUrl, getConfigDir())));
+  }
+
+  @VisibleForTesting
+  static Optional<String> performUpdateCheck(
+      boolean skip, String currentVersion, String versionUrl, Path configDir) {
+    // Abort if offline or update checks are disabled
+    if (skip || Boolean.getBoolean(PropertyNames.DISABLE_UPDATE_CHECKS)) {
```
chanseokoh: I think reading and writing files is fast enough, so they don't have to be in a thread? I think it's OK to run only the network connection code in a thread.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,201 @@
...
+  private TestWebServer testWebServer;
+  private Path configDir;
+
+  @Before
+  public void setUp()
+      throws InterruptedException, GeneralSecurityException, URISyntaxException, IOException {
+    testWebServer =
+        new TestWebServer(
+            false, Collections.singletonList("HTTP/1.1 200 OK\nContent-Length:5\n\n2.0.0"), 1);
+    configDir = temporaryFolder.newFolder().toPath();
```
chanseokoh: Just `temporaryFolder.getRoot()`?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,201 @@
...
+    Optional<String> message =
+        UpdateChecker.performUpdateCheck(true, "1.0.2", testWebServer.getEndpoint(), configDir);
+    Assert.assertFalse(message.isPresent());
+  }
+
+  @Test
+  public void testPerformUpdateCheck_configDisabled() throws IOException {
+    UpdateChecker.ConfigJsonTemplate config = new UpdateChecker.ConfigJsonTemplate();
+    config.setDisableUpdateCheck(true);
+    JsonTemplateMapper.writeTo(config, Files.newOutputStream(configDir.resolve("config.json")));
```
chanseokoh: I think it's nice to write an actual JSON, like `Files.write(..., "{\"disableUpdateCheck\":true}".getBytes(...))`.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,201 @@
...
+      Assert.assertFalse(message.isPresent());
+    }
+  }
+
+  @Test
+  public void testFinishUpdateCheck_success() throws InterruptedException {
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    UpdateChecker updateChecker =
+        new UpdateChecker(executorService.submit(() -> Optional.of("Hello")));
+    Thread.sleep(100);
```
chanseokoh: Does `MoreExecutors.newDirectExecutorService()` work? You may be able to remove `Thread.sleep()`. Otherwise, I think this test can be flaky.

Or better yet, I think `Futures.immediateFuture(Optional.of("Hello"))`.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,201 @@
...
+    Optional<String> result = updateChecker.finishUpdateCheck();
+    Assert.assertTrue(result.isPresent());
+    Assert.assertEquals("Hello", result.get());
+  }
+
+  @Test
+  public void testFinishUpdateCheck_notDone() {
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    UpdateChecker updateChecker =
```
chanseokoh: `Future` is a simple interface, so I think it's better to implement one whose `isDone()` always returns false.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,201 @@
...
+                  }
+                  return Optional.of("Hello");
+                }));
+    Optional<String> result = updateChecker.finishUpdateCheck();
+    Assert.assertFalse(result.isPresent());
+  }
+
+  private void setupConfigAndLastUpdateCheck() throws IOException {
+    UpdateChecker.ConfigJsonTemplate config = new UpdateChecker.ConfigJsonTemplate();
+    JsonTemplateMapper.writeTo(config, Files.newOutputStream(configDir.resolve("config.json")));
```
chanseokoh: Q: this just writes `{}`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,175 @@
...
+   * Begins checking for an update in a separate thread.
+   *
+   * @param skip if {@code true}, the update check itself will be skipped
+   * @param version the current version of Jib being used
+   * @param versionUrl the location to check for the latest version
+   * @param executorService the {@link ExecutorService}
+   * @return a new {@link UpdateChecker}
+   */
+  public static UpdateChecker checkForUpdate(
+      boolean skip, String version, String versionUrl, ExecutorService executorService) {
```
loosebazooka: It feels weird to put "skip" in here?

Could the maven/gradle calls to checkForUpdate be implemented in a way like

```
Updater.for(ProjectProperties).checkForUpdates()
```
where "for" decides whether to skip or not? Or something of that sort.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,175 @@
...
+    return new UpdateChecker(
```
loosebazooka: do you have do some work here to shutdown the executor?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
TadCordle: Nope, writes `{"disableUpdateCheck":false}`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Is there a disadvantage to just running the whole thing in a different thread?

chanseokoh: Not really. I think anyone is fine.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
loosebazooka: I should probably provide more context:

```
ExecutorService x = Executors.newSingleThreadExecutor();
x.submit(() -> System.out.println("goose"));
// x.shutdown();
```

In this example, without x.shutdown, running this will pause until the thread is cleaned up (or something like that)

More context here: https://stackoverflow.com/questions/20057497/program-does-not-terminate-immediately-when-all-executorservice-tasks-are-done

TadCordle: Yeah, I think we can save the executor service and call `shutdown()` in `finishUpdateCheck()`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Looks like it does terminate right away.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Oh, looks like we can shut it down immediately after submitting.

loosebazooka: Should the executor service be isolated to the update checker? I may have missed why we are passing it in?

loosebazooka: Yeah you can use shutdown if you know you're not submitting any more jobs.
But if the executor is coming from outside, then there's no way to ensure that it is exclusively being used for this purpose?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -145,6 +151,10 @@ public void buildDocker()
 
     } finally {
       tempDirectoryProvider.close();
+      updateChecker
+          .finishUpdateCheck()
+          .ifPresent(
+              s -> projectProperties.log(LogEvent.lifecycle("\n\u001B[33m" + s + "\u001B[0m\n")));
```
chanseokoh: I think `s` is too short. s --> updateMessage

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -143,6 +149,10 @@ public void buildImage()
 
     } finally {
       tempDirectoryProvider.close();
+      updateChecker
+          .finishUpdateCheck()
+          .ifPresent(
+              s -> projectProperties.log(LogEvent.lifecycle("\n\u001B[33m" + s + "\u001B[0m\n")));
```
chanseokoh: ditto

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildTarTask.java
```diff
@@ -159,6 +165,10 @@ public void buildTar()
 
     } finally {
       tempDirectoryProvider.close();
+      updateChecker
+          .finishUpdateCheck()
+          .ifPresent(
+              s -> projectProperties.log(LogEvent.lifecycle("\n\u001B[33m" + s + "\u001B[0m\n")));
```
chanseokoh: ditto

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -134,6 +139,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
     } finally {
       tempDirectoryProvider.close();
+      updateChecker
+          .finishUpdateCheck()
+          .ifPresent(
+              s -> projectProperties.log(LogEvent.lifecycle("\n\u001B[33m" + s + "\u001B[0m\n")));
```
chanseokoh: ditto

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildImageMojo.java
```diff
@@ -148,6 +153,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
     } finally {
       tempDirectoryProvider.close();
+      updateChecker
+          .finishUpdateCheck()
+          .ifPresent(
+              s -> projectProperties.log(LogEvent.lifecycle("\n\u001B[33m" + s + "\u001B[0m\n")));
```
chanseokoh: ditto

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildTarMojo.java
```diff
@@ -126,6 +131,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
     } finally {
       tempDirectoryProvider.close();
+      updateChecker
+          .finishUpdateCheck()
+          .ifPresent(
+              s -> projectProperties.log(LogEvent.lifecycle("\n\u001B[33m" + s + "\u001B[0m\n")));
```
chanseokoh: ditto

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,191 @@
...
+   * @return a new {@link UpdateChecker}
+   */
+  public static UpdateChecker checkForUpdate(boolean skip, String versionUrl) {
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    Future<Optional<String>> messageFuture =
+        executorService.submit(
+            () ->
+                performUpdateCheck(
+                    skip,
+                    Preconditions.checkNotNull(ProjectInfo.VERSION),
```
chanseokoh: I can't explain which one is more suitable, but I use `Verify.verifyNotNull()` in this situation.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,191 @@
...
+        ConfigJsonTemplate config =
+            JsonTemplateMapper.readJsonFromFile(configFile, ConfigJsonTemplate.class);
+        if (config.disableUpdateCheck) {
+          return Optional.empty();
+        }
+      } else {
+        // Generate config file if it doesn't exist
+        ConfigJsonTemplate config = new ConfigJsonTemplate();
+        Files.createDirectories(configDir);
+        JsonTemplateMapper.writeTo(config, Files.newOutputStream(configFile));
```
chanseokoh: I think it's meaningful to close `newOutputStream()` in this case.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,191 @@
...
+      }
+
+      // Check for update
+      HttpURLConnection connection = (HttpURLConnection) new URL(versionUrl).openConnection();
+      connection.setConnectTimeout(3000);
+      BufferedReader bufferedReader =
+          new BufferedReader(
+              new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+      String latestVersion = bufferedReader.readLine().trim();
+      connection.disconnect();
```
chanseokoh: I think it's worth putting it in a `finally { }` block to disconnect early even in an error case.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,191 @@
...
+          "A new version of Jib ("
+              + latestVersion
+              + ") is available (currently using "
+              + currentVersion
+              + "). Update your build configuration to use the latest features and fixes!");
+
+    } catch (IOException ex) {
+      try {
+        Files.deleteIfExists(configDir.resolve("config.json"));
+        Files.deleteIfExists(configDir.resolve("lastUpdateCheck"));
```
chanseokoh: nit: `Files.deleteIfExists(configFile)` and `Files.deleteIfExists(lastUpdateCheck)` after putting these various outside the `try` block. As this method is very long, I think it helps avoid the mistake of using different path.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  @Test
+  public void testPerformUpdateCheck_lastUpdateCheckTooSoon() throws IOException {
+    FileTime modifiedTime = FileTime.from(Instant.now().minusSeconds(12));
+    setupConfigAndLastUpdateCheck();
+    Files.write(
+        configDir.resolve("lastUpdateCheck"),
+        modifiedTime.toString().getBytes(StandardCharsets.UTF_8));
+    Files.setLastModifiedTime(configDir.resolve("lastUpdateCheck"), modifiedTime);
```
chanseokoh: Is this necessary? If not, remove it, because it's confusing?

TadCordle: Oops, old stuff I didn't clean up.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -73,6 +75,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
     MavenProjectProperties projectProperties =
         MavenProjectProperties.getForProject(
             getProject(), getSession(), getLog(), tempDirectoryProvider);
+    UpdateChecker updateChecker =
```
loosebazooka: I still kinda think this would be cleaner if we skip wasn't part of the update checker mechanic and we only created an update checker as needed.

```
Optional<UpdateChecker> updaterChecker = MavenUpdateChecker.for(projectProperties, getLog())

...
...
updateChecker.ifPresent().finish....
```

loosebazooka: the  MojoCommon.VERSION_URL could be part of MavenUpdateChecker too

loosebazooka: or something like this (not necessarily this exactly)

TadCordle: I'd rather not make a new class, but I agree the `Optional` approach could be better, so I'll try something like that.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -145,6 +150,12 @@ public void buildDocker()
 
     } finally {
       tempDirectoryProvider.close();
+      updateChecker
+          .flatMap(UpdateChecker::finishUpdateCheck)
```
loosebazooka: oh this is cool, didn't know you could do that with optionals.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,193 @@
...
+                + currentVersion
+                + "). Update your build configuration to use the latest features and fixes!");
+
+      } finally {
+        connection.disconnect();
+      }
+
+    } catch (IOException ex) {
+      try {
+        Files.deleteIfExists(configFile);
```
loosebazooka: I'm not sure I understand what's going on here? why do we delete these files?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,193 @@
...
+   */
+  public Optional<String> finishUpdateCheck() {
+    if (updateMessageFuture.isDone()) {
+      try {
+        return updateMessageFuture.get();
+      } catch (InterruptedException | ExecutionException ignored) {
+        // Fail silently;
+      }
+    }
+    updateMessageFuture.cancel(true);
```
loosebazooka: I think that this just tries to interrupt the process running the thread if it's still running? Do we need to check in `performUpdateCheck` that the thread hasn't been interrupted?

chanseokoh: I regarded this as an optional thing which may have virtually no effect in many cases. @TadCordle tested and observed that the build terminates right away when "the thread is still alive" (https://github.com/GoogleContainerTools/jib/pull/2214#discussion_r366007061). However, I just realized that the test using a single `Thread.sleep()` to simulate a  long-running thread is flawed, since I believe `Thread.sleep()` will unblock when a future is cancelled. @TadCordle the code to test this should look like this:

```java
while (true) {
  try {
    Thread.sleep(999999);
  } catch (Exception e) {}
}
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Was suggested by @chanseokoh to avoid corruption in case the thread is interrupted while writing to a file.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Still seems to terminate the thread immediately with the new test code.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
chanseokoh: About corruption: theoretically the files are shared resources among multiple concurrent Jib runs, although the likelihood is very low that the files are concurrently modified. That's why I am OK with not doing any synchronization for a shared object here.

I'm not saying that this would be the only source of possible corruption. But if the files were ever corrupted somehow, Jib would not be able to recover (it will just throw an `IOExcetion` again and again) without deleting the corrupted files. I just want it to be able to recover.

chanseokoh: And yeah, I think thread interrupted might be another source of possible corruption.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
chanseokoh: That's good. For both Maven and Gradle?

TadCordle: Yup :+1:

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
loosebazooka: deleting the config file could potentially result in the user's non-update-check config to be deleted no?

loosebazooka: I don't think we should ever delete the user's config file.

loosebazooka: We can output a message saying your config is corrupt? And to go clean it up?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,193 @@
...
+    Path configFile = configDir.resolve("config.json");
+    Path lastUpdateCheck = configDir.resolve("lastUpdateCheck");
+    try {
+      if (Files.exists(configFile)) {
+        // Abort if update checks are disabled
+        ConfigJsonTemplate config =
+            JsonTemplateMapper.readJsonFromFile(configFile, ConfigJsonTemplate.class);
+        if (config.disableUpdateCheck) {
+          return Optional.empty();
+        }
```
loosebazooka: I feel like a bad configfile should be caught here -- and the user should be informed that it is indeed corrupt. We then should NOT do an update because there's no way to know if they had previously configured it or not.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,193 @@
...
+      } else {
+        // Generate config file if it doesn't exist
+        ConfigJsonTemplate config = new ConfigJsonTemplate();
+        Files.createDirectories(configDir);
+        try (OutputStream outputStream = Files.newOutputStream(configFile)) {
```
loosebazooka: perhaps another catch here to cleanup this config file (if we're writing it)

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,193 @@
...
+      if (Files.exists(lastUpdateCheck)) {
+        String fileContents =
+            new String(Files.readAllBytes(lastUpdateCheck), StandardCharsets.UTF_8);
+        try {
+          Instant modifiedTime = Instant.parse(fileContents);
+          if (modifiedTime.plus(Duration.ofDays(1)).isAfter(Instant.now())) {
+            return Optional.empty();
+          }
+        } catch (DateTimeParseException ex) {
+          // Ignore parse failure; assume update check hasn't been performed
```
chanseokoh: Following @loosebazooka's suggestion, getting here most likely means the file is corrupted.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
chanseokoh: Yeah, I wanted a simple solution, but in principle I agree it's better to let the user delete the files.

chanseokoh: Also, if it is not corruption but a temporary IO error, I think it sill makes sense to report the error saying that Jib couldn't read its config file (but not a fatal error).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: I think we can auto-delete this one. The user probably shouldn't be touching `lastUpdateCheck`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -137,13 +153,8 @@ public static UpdateChecker checkForUpdate(String versionUrl) {
...
 
-    } catch (IOException ex) {
-      try {
-        Files.deleteIfExists(configFile);
-        Files.deleteIfExists(lastUpdateCheck);
-      } catch (IOException ignored) {
-        // Fail silently
-      }
+    } catch (IOException ignored) {
+      // Fail other exceptions silently
```
loosebazooka: perhaps we should log debug?

loosebazooka: or maybe that's not useful...

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,205 @@
...
+        // Abort if update checks are disabled
+        try {
+          ConfigJsonTemplate config =
+              JsonTemplateMapper.readJsonFromFile(configFile, ConfigJsonTemplate.class);
+          if (config.disableUpdateCheck) {
+            return Optional.empty();
+          }
+        } catch (IOException ex) {
+          warningLogger.accept(
+              "Failed to read global Jib config; you may need to fix or delete " + configFile);
```
chanseokoh: Add the exception message too? I guess better than nothing.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -63,7 +87,8 @@ static void disableHttpLogging() {
     OutputEventListener defaultOutputEventListener = context.getOutputEventListener();
     context.setOutputEventListener(
         event -> {
-          LogEvent logEvent = (LogEvent) event;
+          org.gradle.internal.logging.events.LogEvent logEvent =
```
loosebazooka: I don't think I know what this is about?

TadCordle: Gradle and Jib both have classes named `LogEvent`, so I opted to use the full name for this one.

loosebazooka: oh I figured it out.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.function.Consumer;
+
+/** Checks if Jib is up-to-date. */
+public class UpdateChecker {
+
+  /** JSON template for the configuration file used to enable/disable update checks. */
+  @VisibleForTesting
```
loosebazooka: I think this should be in it's own class outside of UpdateChecker (if we're going to be eventually using a common config file for jib), but we can do this in another PR.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+
+  @VisibleForTesting
+  static Optional<String> performUpdateCheck(
+      Consumer<LogEvent> log, String currentVersion, String versionUrl, Path configDir) {
+    // Abort if offline or update checks are disabled
+    if (Boolean.getBoolean(PropertyNames.DISABLE_UPDATE_CHECKS)) {
+      return Optional.empty();
+    }
+
+    Path configFile = configDir.resolve("config.json");
```
loosebazooka: Similarly we should think about how we want to read configFile earlier (maybe as part of project properties?)?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+   */
+  public Optional<String> finishUpdateCheck() {
+    if (updateMessageFuture.isDone()) {
+      try {
+        return updateMessageFuture.get();
+      } catch (InterruptedException | ExecutionException ignored) {
+        // Fail silently;
+      }
+    }
+    updateMessageFuture.cancel(true);
```
loosebazooka: can we close the connection here to interrupt a read?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Will do later if we end up using the config file for anything other than update checks.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Is it necessary? `updateMessageFuture.cancel(true)` should interrupt the read regardless, shouldn't it? Or will the connection be left open if we interrupt it and don't disconnect here?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+          String fileContents =
+              new String(Files.readAllBytes(lastUpdateCheck), StandardCharsets.UTF_8);
+          Instant modifiedTime = Instant.parse(fileContents);
+          if (modifiedTime.plus(Duration.ofDays(1)).isAfter(Instant.now())) {
+            return Optional.empty();
+          }
+        } catch (DateTimeParseException | IOException ex) {
+          // If reading update time failed, file might be corrupt, so delete it
+          Files.delete(lastUpdateCheck);
+          log.accept(LogEvent.debug("Failed to read lastUpdateCheck; " + ex.getMessage()));
```
chanseokoh: I'd put this before `Files.delete()`. If we got here due to `IOException`, there's good chance `Files.delete()` will fail too. `deleteIfExists()` will be safer?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
chanseokoh: I'm not really concerned about this. Looks like the Maven/Gradle JVM always terminates no matter what I do in the update checker thread. And particularly it doesn't matter much to close the connection if the build is about to exit.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+        // Generate config file if it doesn't exist
+        ConfigJsonTemplate config = new ConfigJsonTemplate();
+        Files.createDirectories(configDir);
+        try (OutputStream outputStream = Files.newOutputStream(configFile)) {
+          JsonTemplateMapper.writeTo(config, outputStream);
+        } catch (IOException ex) {
+          // If attempt to generate new config file failed, delete so we can try again next time
+          Files.deleteIfExists(configFile);
+          log.accept(LogEvent.debug("Failed to generate global Jib config; " + ex.getMessage()));
+        }
```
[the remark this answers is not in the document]
TadCordle: I want to generate the file automatically so it's easier to opt-out when the user goes looking for it.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Will `deleteIfExists()` make a difference when this is inside a block that checks if the file exists?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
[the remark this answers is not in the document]
chanseokoh: Ah, that's a good point.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
chanseokoh: Ah, you're right.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
loosebazooka: I think the 3 second timeout is generally enough.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+      }
+
+      // Check time of last update check
+      if (Files.exists(lastUpdateCheck)) {
+        try {
+          String fileContents =
+              new String(Files.readAllBytes(lastUpdateCheck), StandardCharsets.UTF_8);
+          Instant modifiedTime = Instant.parse(fileContents);
+          if (modifiedTime.plus(Duration.ofDays(1)).isAfter(Instant.now())) {
```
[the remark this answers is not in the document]
loosebazooka: It's easier for us to make the decision to shut it down this way.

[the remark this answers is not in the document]
briandealwis: I just meant pass in the `Instant` as an argument?  @TadCordle had been asking about testing.  Or alternatively pass in the `Clock`.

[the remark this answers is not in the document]
loosebazooka: I commented on the wrong comment... Oops. I meant for the executor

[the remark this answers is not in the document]
loosebazooka: I commented on the wrong comment... Oops. I meant for the executor

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+
+  /**
+   * Begins checking for an update in a separate thread.
+   *
+   * @param log {@link Consumer} used to log messages
+   * @param versionUrl the location to check for the latest version
+   * @return a new {@link UpdateChecker}
+   */
+  public static UpdateChecker checkForUpdate(Consumer<LogEvent> log, String versionUrl) {
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
```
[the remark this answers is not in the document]
loosebazooka: It made some sense to me to manage the lifecycle within the updater (so we could shut it down). But I guess taskcommon could do that.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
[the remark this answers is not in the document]
TadCordle: @briandealwis I think it might help for unit testing, but I think that's in good enough shape right now. The harder problem is integration testing.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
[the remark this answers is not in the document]
TadCordle: It seems a bit unnecessary to me at the moment; like you said, this is plugins common, so we're the only consumer. Wouldn't it just lead to things being more spread out/more duplicate code creating creating an `ExecutorService` for the updater?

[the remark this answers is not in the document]
briandealwis: As a framework, this code shouldn't be making that decision (IMHO).

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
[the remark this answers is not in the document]
briandealwis: This class is hard to test because all of the logic is hard-coded into this single method.  Extract the last-check time determination, the current-time generation (Supplier), the modification time comparison (BiPredicate), and this class should be testable through unit tests.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
+/*
+ * Copyright 2019 Google LLC.
```
loosebazooka: 2020

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,198 @@
+/*
+ * Copyright 2019 Google LLC.
```
loosebazooka: 2020

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  @Test
+  public void testPerformUpdateCheck_failSilently()
+      throws InterruptedException, GeneralSecurityException, URISyntaxException, IOException {
+    try (TestWebServer badServer =
+        new TestWebServer(false, Collections.singletonList("HTTP/1.1 400 Bad Request\n\n"), 1)) {
+      Optional<String> message =
+          UpdateChecker.performUpdateCheck(
+              ignored -> {}, "1.0.2", badServer.getEndpoint(), configDir);
```
briandealwis: Should the logger also be checked that nothing is logged?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/UpdateCheckerTest.java
```diff
@@ -0,0 +1,198 @@
...
+        UpdateChecker.performUpdateCheck(
+            ignored -> {}, "1.0.2", testWebServer.getEndpoint(), configDir);
+    Assert.assertFalse(message.isPresent());
+  }
+
+  @Test
+  public void testPerformUpdateCheck_configDisabled() throws IOException {
+    Files.write(
+        configDir.resolve("config.json"),
+        "{\"disableUpdateCheck\":true}".getBytes(StandardCharsets.UTF_8));
```
briandealwis: Can we also check for invalid content in the config.json, and that the lastUpdateCheck file is removed on invalid content too.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+        executorService.submit(
+            () ->
+                performUpdateCheck(
+                    log, Verify.verifyNotNull(ProjectInfo.VERSION), versionUrl, getConfigDir()));
+    executorService.shutdown();
+    return new UpdateChecker(messageFuture);
+  }
+
+  @VisibleForTesting
+  static Optional<String> performUpdateCheck(
```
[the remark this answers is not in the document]
TadCordle: I'd rather keep it like this; it is a bit long, but it's straightforward to read through. Would rather avoid adding mental context switches to the code. Thoughts @loosebazooka @chanseokoh ?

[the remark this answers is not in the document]
chanseokoh: Without saying much, I think a pseudo-code-like high-level branching description as a method like below helps:

```java
void performUpdateCheck() {
    try {
      if (!checkGlobalConfig(...)) { // or isUpdateCheckDisabled() or something else
        return Optional.empty();
      }
      if (!checkLastUpdateTime(...)) { // or isTimeToCheckBack() or something else, some name that itself gives you the confidence by just reading it that you don't even feel the necessity to look into the method code
        return Optional.empty();
      }
      ... = fetchUrl(versionUrl); // or something else, the method may have a larger scope than just contacting the server, or it may just be okay to not have a method at this point
      
    } catch ( ... ) {
    }
  }
```

A long method may be comfortable to the repo maintainers, but it may not be so to people new to the repo. Like, what I feel when I try to read [this](https://github.com/googleapis/google-http-java-client/blob/f9942bfcbb9f4e9773404bdf16865518393fc9db/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java#L849-L1119).

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -17,25 +17,58 @@
...
 
+  public static final String VERSION_URL = "https://storage.googleapis.com/jib-versions/jib-gradle";
+
+  static Future<Optional<String>> newUpdateChecker(
+      ProjectProperties projectProperties, Logger logger) {
+    if (projectProperties.isOffline() || !logger.isLifecycleEnabled()) {
+      return Futures.immediateFuture(Optional.empty());
+    }
+    return UpdateChecker.checkForUpdate(
+        Executors.newSingleThreadExecutor(), projectProperties::log, VERSION_URL);
```
chanseokoh: If `newUpdateChecker` creates `Executors.newSingleThreadExecutor()`, let's make it "close" it, i.e., calling `shutdown()` in `newUpdateChecker`. And since I know our plugins just terminate without even calling `shutdown()`, but it doesn't harm to guarantee it in `finally`.
```java
} finally {
   executorService.shutdown();
}
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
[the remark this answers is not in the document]
chanseokoh: But since this PR has been dragged too long and having like 100+ comments, let's not do it here. It may trigger another cascade of changes.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -166,10 +166,8 @@ public static JibBuildRunner createJibBuildRunnerForRegistryImage(
         rawConfiguration.getToCredHelper().orElse(null));
 
     boolean alwaysCacheBaseImage =
-        Boolean.valueOf(
-            rawConfiguration
-                .getProperty(PropertyNames.ALWAYS_CACHE_BASE_IMAGE)
-                .orElse(Boolean.FALSE.toString()));
+        Boolean.parseBoolean(
+            rawConfiguration.getProperty(PropertyNames.ALWAYS_CACHE_BASE_IMAGE).orElse("false"));
```
loosebazooka: any reason for this?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+   * @param log {@link Consumer} used to log messages
+   * @param versionUrl the location to check for the latest version
+   * @return a new {@link UpdateChecker}
+   */
+  public static Future<Optional<String>> checkForUpdate(
+      ExecutorService executorService, Consumer<LogEvent> log, String versionUrl) {
+    return executorService.submit(
+        () ->
+            performUpdateCheck(
+                log, Verify.verifyNotNull(ProjectInfo.VERSION), versionUrl, getConfigDir()));
```
loosebazooka: what happens when the verify fails?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  /**
+   * Returns the config directory set by {@link PropertyNames#CONFIG_DIRECTORY} if not null,
+   * otherwise returns the default config directory.
+   *
+   * @return the config directory set by {@link PropertyNames#CONFIG_DIRECTORY} if not null,
+   *     otherwise returns the default config directory.
+   */
+  private static Path getConfigDir() {
```
loosebazooka: I thought we had some sort of static class do this?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
TadCordle: Just a cleanup thing I noticed.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,212 @@
...
+      // Check for update
+      HttpURLConnection connection = (HttpURLConnection) new URL(versionUrl).openConnection();
+      try {
+        connection.setConnectTimeout(3000);
+        BufferedReader bufferedReader =
+            new BufferedReader(
+                new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+        VersionJsonTemplate version =
+            JsonTemplateMapper.readJson(
+                bufferedReader.readLine().trim(), VersionJsonTemplate.class);
```
chanseokoh: `readLine()` seems unsafe. It will break if our JSON file is not a single-line JSON. I think you used `BufferedReader` only to use `readLine()`, so perhaps it's no longer necessary. Maybe use `InputStream` directly?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: Oh wait sorry, this uses system property as part of the plugin frontend, so it doesn't belong in jib core.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
TadCordle: I added it to make nullaway happy. Will `ProjectInfo.VERSION` every actually be null in practice?

chanseokoh: No. As long as the code runs inside a JAR (which is the case, our plugins are downloaded as a JAR), I believe this is never null. However, I think this is null in our tests. We should be good.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -0,0 +1,206 @@
...
+          }
+        } catch (DateTimeParseException | IOException ex) {
+          // If reading update time failed, file might be corrupt, so delete it
+          log.accept(LogEvent.debug("Failed to read lastUpdateCheck; " + ex.getMessage()));
+          Files.delete(lastUpdateCheck);
+        }
+      }
+
+      // Check for update
+      HttpURLConnection connection = (HttpURLConnection) new URL(versionUrl).openConnection();
```
briandealwis: I realize that this doesn't handle proxies.

TadCordle: In what way?

briandealwis: Sorry should have phrased that as a question — does this handle proxies?  But it is being called after activating the `MavenSettingsProxyProvider`, so all good.

PR #2218, opened 2020-01-07
Change base image cache location on mac/windows, add methods for getting XDG config location

### LINKED ISSUE #2216: Change cache location on Mac/Windows
Currently, the base image cache is located at:
* Mac: `$HOME/Library/Application Support/google-cloud-tools-java/jib/`
* Win: `%LOCALAPPDATA/google-cloud-tools-java/jib/`

It may be a good idea to change these to:
* Mac: `$HOME/Library/Caches/Google/Jib/`
* Win: `%LOCALAPPDATA/Google/Jib/Cache/`

We may need to come up with a migration strategy so we don't break users who rely heavily on the cache (e.g. offline users).

Checklist:
- [x] Code (#2218)
- [x] Docs (#2231)

TadCordle:
Fixes #2216.

Adds methods for retrieving the config directory:
* Linux: `$HOME/.config/google-cloud-tools-java/jib/`
* Mac: `$HOME/Library/Preferences/Google/Jib/`
* Windows: `%LOCALAPPDATA%\Google\Jib\Config`

Also changes the cache directory on Mac/Windows to the following:

Before:
* Mac: `$HOME/Library/Application Support/google-cloud-tools-java/jib/`
* Win: `%LOCALAPPDATA/google-cloud-tools-java/jib/`

After:
* Mac: `$HOME/Library/Caches/Google/Jib/`
* Win: `%LOCALAPPDATA\Google\Jib\Cache\`

Linux is still at `$HOME/.cache/google-cloud-tools-java/jib/`. Will need to update documentation to help users migrate the old cache over to the new location.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
@@ -26,7 +26,7 @@ import java.util.Properties;
 import java.util.logging.Logger;
 
 /**
- * Obtains an OS-specific directories based on the XDG Base Directory Specification.
+ * Obtains OS-specific directories based on the XDG Base Directory Specification.
  *
  * <p>Specifically, from the specification:
  *
@@ -43,7 +43,10 @@ import java.util.logging.Logger;
  */
 public class XdgDirectories {
 
-  private static final Logger logger = Logger.getLogger(XdgDirectories.class.getName());
+  private static final Logger LOGGER = Logger.getLogger(XdgDirectories.class.getName());
+  private static final Path JIB_SUBDIRECTORY_LINUX =
+      Paths.get("google-cloud-tools-java").resolve("jib");
+  private static final Path JIB_SUBDIRECTORY_OTHER = Paths.get("Google").resolve("Jib");
 
   public static Path getCacheHome() {
     return getCacheHome(System.getProperties(), System.getenv());
@@ -84,12 +87,12 @@ public class XdgDirectories {
       // Use %LOCALAPPDATA% for Windows.
       String localAppDataEnv = environment.get("LOCALAPPDATA");
       if (localAppDataEnv == null || localAppDataEnv.trim().isEmpty()) {
-        logger.warning("LOCALAPPDATA environment is invalid or missing");
+        LOGGER.warning("LOCALAPPDATA environment is invalid or missing");
         return xdgPath;
       }
       Path localAppData = Paths.get(localAppDataEnv);
       if (!Files.exists(localAppData)) {
-        logger.warning(localAppData + " does not exist");
+        LOGGER.warning(localAppData + " does not exist");
         return xdgPath;
       }
       return localAppData;
@@ -98,7 +101,7 @@ public class XdgDirectories {
       // Use '~/Library/Application Support/' for macOS.
       Path applicationSupport = Paths.get(userHome, "Library", "Application Support");
       if (!Files.exists(applicationSupport)) {
-        logger.warning(applicationSupport + " does not exist");
+        LOGGER.warning(applicationSupport + " does not exist");
         return xdgPath;
       }
       return applicationSupport;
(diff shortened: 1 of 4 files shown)


chanseokoh:
If we have to do this, should we also change the Linux directory to be consistent, from `.../google-cloud-tools-java/jib` to `.../google/jib`? The majority of Java developers are on Windows (in generally, may not be true for developers building images), so may not be a big deal. And if so, maybe we should just go with `../google/jib` in all lower cases for both Windows and Mac? But I'm fine in either way.

TadCordle:
Consistency would certainly make this easier. I don't really care where it ends up either way personally.

chanseokoh:
Ah, okay. Forget changing `google-cloud-tools-java/jib` then.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,161 @@
...
+      Map<String, String> environment,
+      String xdgEnvVariable,
+      String linuxFolder,
+      String windowsFolder,
+      String macFolder) {
+
+    Path windowsSubDirectory = JIB_SUBDIRECTORY_OTHER.resolve(windowsFolder);
+    String rawOsName = properties.getProperty("os.name");
+    String osName = rawOsName.toLowerCase(Locale.ENGLISH);
+    String xdgConfigHome = environment.get(xdgEnvVariable);
```
chanseokoh: xdgConfigHome --> xdgHome

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,161 @@
...
+    String userHome = properties.getProperty("user.home");
+    Path xdgPath = Paths.get(userHome, linuxFolder);
+
+    if (osName.contains("linux")) {
+      // Use environment variable $XDG_CONFIG_HOME if set and not empty.
```
chanseokoh: "Use XDG environment variable if set and not empty."

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,161 @@
...
+      }
+      Path localAppData = Paths.get(localAppDataEnv);
+      if (!Files.exists(localAppData)) {
+        LOGGER.warning(localAppData + " does not exist");
+        return xdgPath.resolve(windowsSubDirectory);
+      }
+      return localAppData.resolve(windowsSubDirectory);
+
+    } else if (osName.contains("mac") || osName.contains("darwin")) {
+      // Use environment variable $XDG_CONFIG_HOME if set and not empty.
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,161 @@
...
+
+    if (osName.contains("linux")) {
+      // Use environment variable $XDG_CONFIG_HOME if set and not empty.
+      if (xdgConfigHome != null && !xdgConfigHome.trim().isEmpty()) {
+        return Paths.get(xdgConfigHome).resolve(JIB_SUBDIRECTORY_LINUX);
+      }
+      return xdgPath.resolve(JIB_SUBDIRECTORY_LINUX);
+
+    } else if (osName.contains("windows")) {
+      // Use environment variable $XDG_CONFIG_HOME if set and not empty.
```
chanseokoh: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,161 @@
...
+      return localAppData.resolve(windowsSubDirectory);
+
+    } else if (osName.contains("mac") || osName.contains("darwin")) {
+      // Use environment variable $XDG_CONFIG_HOME if set and not empty.
+      if (xdgConfigHome != null && !xdgConfigHome.trim().isEmpty()) {
+        return Paths.get(xdgConfigHome).resolve(JIB_SUBDIRECTORY_OTHER);
+      }
+
+      // Use '~/Library/Preferences/' for macOS.
+      Path applicationSupport = Paths.get(userHome, "Library", macFolder);
```
chanseokoh: The name `applicationSupport` is no longer used, right?

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,161 @@
...
+      }
+      return localAppData.resolve(windowsSubDirectory);
+
+    } else if (osName.contains("mac") || osName.contains("darwin")) {
+      // Use environment variable $XDG_CONFIG_HOME if set and not empty.
+      if (xdgConfigHome != null && !xdgConfigHome.trim().isEmpty()) {
+        return Paths.get(xdgConfigHome).resolve(JIB_SUBDIRECTORY_OTHER);
+      }
+
+      // Use '~/Library/Preferences/' for macOS.
```
chanseokoh: "Preferences" is only for the config directory.

jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -0,0 +1,161 @@
...
+   * <p>For Linux, this is {@code $HOME/.config/google-cloud-tools-java/jib/}.
+   *
+   * <p>For Windows, this is {@code %LOCALAPPDATA%\Google\Jib\Config\}.
+   *
+   * <p>For macOS, this is {@code $HOME/Library/Preferences/Google/Jib/}.
+   */
+  @VisibleForTesting
+  static Path getConfigHome(Properties properties, Map<String, String> environment) {
+    return getOsSpecificDirectory(
+        properties, environment, "XDG_CONFIG_HOME", ".config", "Config", "Preferences");
```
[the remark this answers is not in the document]
TadCordle: This might end up being a pain, since the function operates on portions of the os-specific subdirectories, not the whole thing (e.g. mac needs to check if `~/Library/Preferences` exists ignoring the `JIB_SUBDIRECTORY_OTHER` portion, if `$XDG_X_HOME` isn't configured then each OS defaults to `$HOME/.config/[OS-specific subdirectory], etc.)

Issue #1875, opened 2019-07-30
Give access to image repository or full image references built or pushed


chanseokoh:
It was a popular demand to be able to know the image digest/ID of the built image (#933, #1204), and we decided to expose this metadata in the file `jib-image.digest` and `jib-image.id` as clear-text.

Recently there was a request to expose the image repository name similarly, and this seems like a sane feature request. Some tools may just do `jib:build` on user-supplied source and want to know the image name. Also `<to><image>` is optional for `jib:dockerBuild` and Jib will decide the name.

However, the particular request was to get the image repository name that does not include tags. (Note that Jib can push the same image with multiple tags.)

In that light, we may expose both the repository name (without the tag part) and the full image references built or pushed (will be a list).

Knowing that we would expose multiple metadata, we should decide how. For the digest, we create `jib-image.digest`/`jib-image.id` containing the digest/ID, but while we are at it, we may very well create a new structured JSON file containing all data (and potentially more in the future).

briandealwis:
We also expose the `jib-image.digest` too.

TadCordle:
I think we can call this fixed by #2227

chanseokoh:
@Alos we now write a metadata JSON file containing the image repository name, if this still helps.

PR #2227, opened 2020-01-11
Write out the image metadata to a file after build.

mbruggmann:
The jib-maven-plugin will now create a json file with image 
metadata during the build, like so:
```
$ mvn jib:build
...
$ cat target/jib-image.json | jq .
{
  "image": "gcr.io/project/image:tag",
  "imageId": "sha256:61bb3ec31a47cb730eb58a38bbfa813761a51dca69d10e39c24c3d00a7b2c7a9",
  "imageDigest": "sha256:3f1be7e19129edb202c071a659a4db35280ab2bb1a16f223bfd5d1948657b6fc",
  "tags": ["latest", "tag"]
}
```

This is handy if you generate the image name or tag in a dynamic
fashion (eg, including a timstamp) and you have other tools in a
CD pipeline that operate on the generated image. In our case, we
would like to use the reference to the generated image to deploy to
K8S, and our deployment tooling picks it up from a file.

The path to the output file is configurable in 
`outputPaths.imageJson`.

Related issue: #1875


mbruggmann:
@loosebazooka Agree that there are workarounds using other maven/gradle plugins. Two things I'm wary about with that approach: 
* We need to force the plugins to run on the same lifecycle for it to make sense. Also, issuing `mvn jib:build` manually doesn't produce the image metadata file.
* The image name is duplicated in the pom, once for the jib plugin and once for producing the file. This means that if a consumer of the parent pom overwrites the jib `to.image`, passes it in through the CLI, by setting a property etc the output won't match.

How about combining the repository, digests, tags etc into a single `jib-image.json` output file, as per linked issue? Would that make more sense? Otherwise I'm happy to close this and explore other solutions form our side of course.

loosebazooka:
@mbruggmann these are good points. I think we're still working out how we want jib-image.json to work, but this might be a good time to do it.
We will probably leave the other files for backwards compatibility.

loosebazooka:
@mbruggmann the changes look nice, I'll take a deeper look, can you also output the extra tags in the json file?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/JibBuildRunnerTest.java
```diff
@@ -215,4 +222,27 @@ public void testBuildImage_other()
...
+    final Path outputPath = temporaryFolder.newFile("jib-image.json").toPath();
+
+    Mockito.when(mockJibContainer.getTargetImage()).thenReturn(targetImageReference);
+    Mockito.when(mockJibContainer.getImageId()).thenReturn(DescriptorDigest.fromDigest(imageId));
+    Mockito.when(mockJibContainer.getDigest()).thenReturn(DescriptorDigest.fromDigest(digest));
+    Mockito.when(mockJibContainerBuilder.containerize(mockContainerizer))
+        .thenReturn(mockJibContainer);
+    testJibBuildRunner.writeImageJson(outputPath).runBuild();
+
+    final String outputJson = new String(Files.readAllBytes(outputPath), StandardCharsets.UTF_8);
```
briandealwis: I think it would be better to compare this output explicitly as a string to ensure we're producing expected output.  (For example, to detect a situation where we change JSON libraries that produces slightly-invalid JSON.)

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ImageMetadataOutput.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+
+/**
+ * Builds a JSON string containing metadata about a {@link JibContainer} from a build.
+ *
+ * <p>Example:
+ *
+ * <pre>{@code
+ * {
+ *   "image":"gcr.io/project/image:tag",
```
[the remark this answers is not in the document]
mbruggmann: Fixed

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/JibBuildRunnerTest.java
mbruggmann: Had a look at this. Beyond putting the full expected json into a string in the testcase (which seems highly specific) I didn't find a good way of doing that. 

As of now, the next line of `ImageMetadataOutput.fromJson(outputJson)` will at least make sure it properly parses into the expected type so it should catch any invalid JSON, serialization/deserialization differences and similar.

mbruggmann:
Added support for tags now if you'd like to take another look @loosebazooka @briandealwis

jib-core/src/test/java/com/google/cloud/tools/jib/api/JibContainerTest.java
```diff
@@ -44,34 +47,62 @@ public void setUp() throws DigestException, InvalidImageReferenceException {
...
+  }
+
+  @Test
+  public void testEquality_differentImageId() {
+    JibContainer container1 = new JibContainer(targetImage1, digest1, digest1, tags1);
+    JibContainer container2 = new JibContainer(targetImage1, digest1, digest2, tags1);
+
+    Assert.assertNotEquals(container1, container2);
+    Assert.assertNotEquals(container1.hashCode(), container2.hashCode());
+  }
```
briandealwis: Thanks for improving the tests 🥇

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ImageMetadataOutput.java
```diff
@@ -62,7 +68,11 @@ public static ImageMetadataOutput fromJibContainer(JibContainer jibContainer) {
     String image = jibContainer.getTargetImage().toString();
     String imageId = jibContainer.getImageId().toString();
     String imageDigest = jibContainer.getDigest().toString();
-    return new ImageMetadataOutput(image, imageId, imageDigest);
+
+    // Make sure tags always appear in a predictable way, by sorting them into a list
+    List<String> tags = jibContainer.getTags().stream().sorted().collect(Collectors.toList());
```
briandealwis: I think Guava's [`ImmutableList.sortedCopyOf()`](https://guava.dev/releases/28.2-jre/api/docs/com/google/common/collect/ImmutableList.html#sortedCopyOf-java.lang.Iterable-) is a little more compact, and immutable?

loosebazooka:
Probably kind of annyoing, given our integration tests only run on internal machines (and you can't see the output). But can you run the integration tests yourself and fix the issues?
Instructions here: https://github.com/GoogleContainerTools/jib/blob/master/CONTRIBUTING.md#code-reviews

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ImageMetadataOutput.java
mbruggmann: Good idea, updated!

mbruggmann:
The integration-tests caught an actual bug: the code was relying on `ImageReference` equality that wasn't implemented (added in the latests commit). Now  it seems like Travis had what looks like an intermittent error though.  Could someone with write access to the repo retrigger the Travis and Kokoro builds? Thanks

PR #2231, opened 2020-01-14
Update README for new cache location

### LINKED ISSUE #2216: Change cache location on Mac/Windows

TadCordle:
Fixes #2216.

### DIFF
--- a/jib-gradle-plugin/README.md
+++ b/jib-gradle-plugin/README.md
@@ -271,7 +271,7 @@ Property | Type | Default | Description
 --- | --- | --- | ---
 `jib.httpTimeout` | `int` | `20000` | HTTP connection/read timeout for registry interactions, in milliseconds. Use a value of `0` for an infinite timeout.
 `jib.useOnlyProjectCache` | `boolean` | `false` | If set to true, Jib does not share a cache between different Maven projects.
-`jib.baseImageCache` | string | *Platform-dependent*\*\*\* | Sets the directory to use for caching base image layers. This cache can (and should) be shared between multiple images.
+`jib.baseImageCache` | `File` | *Platform-dependent*\*\*\* | Sets the directory to use for caching base image layers. This cache can (and should) be shared between multiple images.
 `jib.applicationCache` | `File` | `[project dir]/target/jib-cache` | Sets the directory to use for caching application layers. This cache can be shared between multiple images.
 `jib.console` | `String` | *None* | If set to `plain`, Jib will print plaintext log messages rather than display a progress bar during the build.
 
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/README.md
```diff
@@ -271,14 +271,16 @@ Property | Type | Default | Description
 --- | --- | --- | ---
 `jib.httpTimeout` | `int` | `20000` | HTTP connection/read timeout for registry interactions, in milliseconds. Use a value of `0` for an infinite timeout.
 `jib.useOnlyProjectCache` | `boolean` | `false` | If set to true, Jib does not share a cache between different Maven projects.
-`jib.baseImageCache` | `File` | `[user cache home]/google-cloud-tools-java/jib` | Sets the directory to use for caching base image layers. This cache can (and should) be shared between multiple images.
+`jib.baseImageCache` | `File` | `[user cache home]/google-cloud-tools-java/jib`\*\*\* | Sets the directory to use for caching base image layers. This cache can (and should) be shared between multiple images.
```
chanseokoh: Hmm... it is not clear what `[user cache home]` is. I guess, for example, it is either `$HOME/.cache/google-cloud-tools-java/jib` or `$XDG_CACHE_HOME/jib`? I actually forgot the exact path.

TadCordle: Here `[user cache home]` means either `$XDG_CACHE_HOME` or `$HOME/.cache/`. Writing out of the full directory for everything gets verbose pretty quickly so I kept the shorthand. But if you think it's necessary to be more explicit I can change it.

chanseokoh: I agree. How about `[platform-dependent cache root]` instead of `[user cache home]`?

TadCordle: This is specifically linux; the footnote goes into other platforms. I'll change it to be clear.

jib-gradle-plugin/README.md
```diff
@@ -271,14 +271,19 @@ Property | Type | Default | Description
 --- | --- | --- | ---
 `jib.httpTimeout` | `int` | `20000` | HTTP connection/read timeout for registry interactions, in milliseconds. Use a value of `0` for an infinite timeout.
 `jib.useOnlyProjectCache` | `boolean` | `false` | If set to true, Jib does not share a cache between different Maven projects.
-`jib.baseImageCache` | `File` | `[user cache home]/google-cloud-tools-java/jib` | Sets the directory to use for caching base image layers. This cache can (and should) be shared between multiple images.
+`jib.baseImageCache` | string | *Platform-dependent*\*\*\* | Sets the directory to use for caching base image layers. This cache can (and should) be shared between multiple images.
```
chanseokoh: nit: probably we can leave `File` (although I guess `string` or anything suitable will work with Gradle)? `jib.applicationCache` just below says `File`.

PR #2232, opened 2020-01-14
Add privacy page for disabling update check

TadCordle:


docs/privacy.md
```diff
@@ -0,0 +1,15 @@
...
+## Update check
+
+To keep Jib up to date, update checks are made to Google servers to see if a new version of
+Jib is available. By default, this behavior is enabled. As a side effect this request is logged.
+
+To disable the update check you have two options:
+
+1. set the `jib.disableUpdateChecks` system property to `true`
+2. set `disableUpdateChecks` to `true` in Jib's global config. The global config is in the following locations by default:
+    * Linux: `$HOME/.config/google-cloud-tools-java/jib/config.json`
```
loosebazooka: presumably this is the wrong time and place to ask this, but why is this one `google-cloud-tools-java` and the others are `Google`?

TadCordle: It's just due to the observation that different platforms seem to have different conventions; Mac and windows seem to like to group things by vendor/app, whereas linux seems to just have a more flat structure. We could change it to be more consistent, but we should do it before 2.0 if we ever want to do it.

chanseokoh: I believe the directory can be `$XDG_CONFIG_HOME/.config/...` if the variable is defined. I think it's good to document things correctly.

loosebazooka:
Looks good, I say submit this for now, as 2.0.0 is out in the wild, but if anyone has any comments after submit, make sure we go back and update this doc.

PR #2234, opened 2020-01-15
CHANGELOG for optionally ignoring registry port in <server><id> in settings.xml

chanseokoh:
For #2212.

### DIFF
--- a/jib-maven-plugin/CHANGELOG.md
+++ b/jib-maven-plugin/CHANGELOG.md
@@ -4,7 +4,6 @@ All notable changes to this project will be documented in this file.
 ## [unreleased]
 
 ### Added
-- When giving registry credentials in `settings.xml`, specifying port in `<server><id>` is now optional. ([#2135](https://github.com/GoogleContainerTools/jib/issues/2135))
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


jib-maven-plugin/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [unreleased]
 
 ### Added
+- When giving registry credentials in `settings.xml`, specifying port in `<server><id>` is now optional. ([#2135](https://github.com/GoogleContainerTools/jib/issues/2135))
```
TadCordle: This sounds like something that was required and is now optional? Maybe it should be something like `Support for optional port in <server><id> in settings.xml.`

chanseokoh: Actually, specifying port was required if an image reference had it. For example, if `<to><image>my-server:1234/awe/some`, it was required to have `<server><id>my-server:1234</id>`.  Now you can say just `<server><id>my-server</id>`. (`<id>my-server:1234` will continue to work.)

chanseokoh: Perhaps I should say `... is no longer required`?

TadCordle: In that case maybe this should go under "Changed" instead of "Added".

PR #2235, opened 2020-01-16
Import Cram as Jib CLI

briandealwis:
Uses the `application` plugin to create a runnable distribution.

I'll leave Graal support for a separate PR as the [Palantír Graal plugin](https://github.com/palantir/gradle-graal) doesn't support GraalVM 19.3.x.


briandealwis:
What was I thinking 🤦‍♂ Windows uses colons in drive specs

jib-cli/README.md
```diff
@@ -0,0 +1,56 @@
...
+a Java library for building containers without Docker.
+
+This CLI tool is _experimental_ and its options and structure
+are almost certain to change.
+
+## Building
+
+Use the `application` plugin's `installDist` task to create a runnable installation.
+```sh
+$ ../gradlew installDist
```
chanseokoh: We can address this later, but because this requires installing JRE and Gradle, I think we should provide pre-compiled binaries, especially when considering Jib CLI is very general and isn't bound to Java developers.

jib-cli/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+dependencies {
+  implementation project(':jib-core')
+  // since jib core doesn't export dependencies to a compile scope
+  // (they are "runtime"), just grab them manually. This means we don't have
+  // to synchronize dependencies between the two projects -- we don't
+  // want to use the sourceProject helper because it does things we don't want
+  // for an unpublished library.
+  implementation project(':jib-core').configurations.implementation.dependencies
+  implementation "com.google.guava:guava:${dependencyVersions.GUAVA}"
+  implementation "info.picocli:picocli:${dependencyVersions.PICOCLI}"
```
chanseokoh: I'm curious how all these work out as a CLI tool. Does it not require a JRE? Is it a single executable binary that embeds all transitive dependency JARs? Or, does it cache dependencies under `~/.m2`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ActualTimestampProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+import com.google.cloud.tools.jib.api.AbsoluteUnixPath;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.function.BiFunction;
+
+/** A timetamp provider that uses the actual file last-modification time from the file-system. */
+class ActualTimestampProvider implements BiFunction<Path, AbsoluteUnixPath, Instant> {
```
chanseokoh: Is this used somewhere? I can't find it. I'm wary of people using this provider if it's going to be used for creating an image, because it greatly undermines reproducibility of a layer.

jib-cli/README.md
briandealwis: The distribution doesn't require Gradle, but it does require a JRE (as does Gradle and Maven).  But otherwise 100% agreed: the real end-goal is to use Graal's native-image to produce an executable, but that's not easy to do with the most recent release of Graal (yet).

jib-cli/build.gradle
briandealwis: Picocli is a nice library for processing command-line arguments.  It also supports more sophisticated command-groups (e.g., `jib build ...` vs `jib jar ...` or `jib inspect ...`), though this code isn't taking advantage of these command-groups yet.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ActualTimestampProvider.java
briandealwis: It's pulled in from `LayerParserDefinition#parseTimestampsDirective()`.

briandealwis:
Marked as not-ready as I have a few more tweaks to pull in.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ActualTimestampProvider.java
briandealwis: It's pulled in from `LayerParserDefinition#parseTimestampsDirective()`.

jib-cli/README.md
chanseokoh: I meant by Gradle that to obtain Jib CLI, the doc says to run `./gradlew installDist`. I wonder if we can "pre-compile" `gradlew installDist` on Linux, Mac, and Windows, and distribute them (whatever it is)?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ActualTimestampProvider.java
briandealwis: It's pulled in from `LayerParserDefinition#parseTimestampsDirective()`.

jib-cli/build.gradle
chanseokoh: Hmm... I don't think I got the answers for my questions, but I think I'll just try this branch myself and see what `./gradlew installDist` generates and how exactly it works on my machine.

briandealwis:
PTAL. Refactored and moved the build functionality into a new `jib build` subcommand.  This should facilitate creating `jib jar` or other subcommands.

chanseokoh:
Builds are failing.

briandealwis:
Oops, and I hadn't updated to the latest jib-build-plan changes.

jib-cli/build.gradle
```diff
@@ -0,0 +1,36 @@
+plugins {
+  id "application"
+}
+
+ext {
+    cliMainClass = "com.google.cloud.tools.jib.cli.JibCli"
```
chanseokoh: nit: indentation is off, and no double quotes

jib-cli/build.gradle
```diff
@@ -0,0 +1,36 @@
+plugins {
+  id "application"
```
chanseokoh: nit: 'application' (not double quotes)

jib-cli/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+}
+
+ext {
+    cliMainClass = "com.google.cloud.tools.jib.cli.JibCli"
+}
+
+// use `run` to build and run the app
+// use `installDist` or `distZip` to create an installable application
+application {
+  applicationName = "jib"
```
chanseokoh: nit: 'jib` (not double quotes)

jib-cli/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+
+  testImplementation "junit:junit:${dependencyVersions.JUNIT}"
+  testImplementation "org.mockito:mockito-core:${dependencyVersions.MOCKITO_CORE}"
+  testImplementation "org.slf4j:slf4j-api:${dependencyVersions.SLF4J_API}"
+  testImplementation "com.github.stefanbirkner:system-rules:${dependencyVersions.SYSTEM_RULES}"
+  testImplementation project(path:':jib-core', configuration:'tests')
+}
+
+sourceSets.test.resources.srcDirs project(':jib-core').sourceSets.test.resources
```
chanseokoh: Haven't looked into the test, so let me confirm. Do you need to copy test resources from `jib-core`?

jib-cli/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+  implementation project(':jib-core').configurations.implementation.dependencies
+  implementation "com.google.guava:guava:${dependencyVersions.GUAVA}"
+  implementation "info.picocli:picocli:${dependencyVersions.PICOCLI}"
+
+
+  testImplementation "junit:junit:${dependencyVersions.JUNIT}"
+  testImplementation "org.mockito:mockito-core:${dependencyVersions.MOCKITO_CORE}"
+  testImplementation "org.slf4j:slf4j-api:${dependencyVersions.SLF4J_API}"
+  testImplementation "com.github.stefanbirkner:system-rules:${dependencyVersions.SYSTEM_RULES}"
+  testImplementation project(path:':jib-core', configuration:'tests')
```
chanseokoh: This is probably unnecessary?

jib-cli/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+
+// use `run` to build and run the app
+// use `installDist` or `distZip` to create an installable application
+application {
+  applicationName = "jib"
+  mainClassName = cliMainClass
+}
+
+dependencies {
+  implementation project(':jib-build-plan')
```
chanseokoh: Can you see if you can remove this (as we are copying `jib-build-plan` source into `jib-core`)? `jib-plugins-common` does not add this, for example. @loosebazooka may have a better idea.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ActualTimestampProvider.java
chanseokoh: So the question is, should we drop support for `ts=actual`? Jib plugins don't support it (and will never after we build an extension framework). I find think this is very useful, especially when considering the negatives. If someone really wanted to set it to the last modification time of some file, I think they could get easily it in a shell script.

```
              + "  ts=timestamp  set last-modified timestamps:\n"
              + "    actual      use actual values in file-system\n"`
```

However, this doesn't need to to addressed in this PR (in case you agree).

jib-cli/README.md
loosebazooka: Ah so we talked about. We can't offer distributed binaries unless we go through an official launch process -- so that will as you mention have to be addressed later.

jib-cli/build.gradle
loosebazooka: Yeah we can remove this I think? But I'm not sure how this build works -- does it just package up the application without depending on anything being in the local maven repo?

jib-cli/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+// use `installDist` or `distZip` to create an installable application
+application {
+  applicationName = 'jib'
+  mainClassName = cliMainClass
+}
+
+dependencies {
+  implementation project(':jib-build-plan')
+  implementation project(':jib-core')
+  implementation project(':jib-core').configurations.implementation.dependencies
```
loosebazooka: We can also remove this line, you would only need this if you're trying to access classes in dependencies of jib-core. We do this in appengine-plugins-common kind of as a hack to not have to re specify the implementation dependencies (that is a little hacky, and maybe we should be more explicit about it -- but that's another issue).

jib-cli/README.md
briandealwis: 👍

jib-cli/build.gradle
briandealwis: I've tried removing this and the build breaks with unresolved symbols 🤷‍♂

loosebazooka: oh yeah, because it's not re-exported. so I think even in jib-core we need to make it an `api` dependency rather than `implementation`?

chanseokoh: So what should be done in this PR?

loosebazooka: I think it's correct for this PR.

Issue #1067, opened 2018-09-28
RegistryClient User-Agent string doesn't conform to spec


TadCordle:
https://github.com/GoogleContainerTools/jib/pull/1066#discussion_r221373768

loosebazooka:
We have too much machinery around this now to change unfortunately, I'll just close it. If it really does break something sometime, we'll fix it then

PR #2242, opened 2020-01-23
Add user agent to update check

TadCordle:
Adds the jib user agent string to the update check request.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
@@ -20,6 +20,9 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.google.cloud.tools.jib.ProjectInfo;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.filesystem.XdgDirectories;
+import com.google.cloud.tools.jib.http.FailoverHttpClient;
+import com.google.cloud.tools.jib.http.Request;
+import com.google.cloud.tools.jib.http.Response;
 import com.google.cloud.tools.jib.json.JsonTemplate;
 import com.google.cloud.tools.jib.json.JsonTemplateMapper;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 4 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -139,11 +150,17 @@ void setDisableUpdateCheck(boolean disableUpdateCheck) {
...
 
       // Check for update
-      HttpURLConnection connection = (HttpURLConnection) new URL(versionUrl).openConnection();
+      FailoverHttpClient httpClient = new FailoverHttpClient(true, false, log);
+      Response response =
+          httpClient.get(
+              new URL(versionUrl),
+              Request.builder()
+                  .setHttpTimeout(3000)
+                  .setUserAgent("jib " + ProjectInfo.VERSION + " " + toolName)
```
chanseokoh: I actually don't like our current string format (#1067), but I guess we'll just go with this for consistency.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -139,11 +150,17 @@ void setDisableUpdateCheck(boolean disableUpdateCheck) {
       }
 
       // Check for update
-      HttpURLConnection connection = (HttpURLConnection) new URL(versionUrl).openConnection();
+      FailoverHttpClient httpClient = new FailoverHttpClient(true, false, log);
+      Response response =
+          httpClient.get(
```
chanseokoh: You can put this line into `try`, can't you? Otherwise, if `httpClient.get()` throws an exception, `httpClient.shutDown()` is not called.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -139,11 +150,17 @@ void setDisableUpdateCheck(boolean disableUpdateCheck) {
       }
 
       // Check for update
-      HttpURLConnection connection = (HttpURLConnection) new URL(versionUrl).openConnection();
+      FailoverHttpClient httpClient = new FailoverHttpClient(true, false, log);
```
chanseokoh: Actually, maybe it's better to ignore logs?

`new FailoverHttpClient(true, false, ignored -> {})`

TadCordle: True, I'll submit another PR.

Issue #2184, opened 2019-12-04
Too many tasks executed on jib build


bs-git:
**Description of the issue**:
While using the jib gradle task (jib) the Plugin configures the jib tasks to depend on ALL assemble tasks on all upstream projects in a multi project build. This causes lots of unecessary artifacts being build even if only the jar (default dependency) of any upstream project is used.

**Expected behavior**:
If the plugin is applied only the default dependency should be used, which in fact is already evaluated for the build of the project (jar/classes). So there should be no need at all to depend on anything on the parent projects. If specific artifacts are needed by a specific project the project should configure this dependency itself, it should not be hard coded within the Jib Plugin itself.

**Steps to reproduce**:

Its in the original code:
`
 COPIED FROM JibPlugin.java:

            // Find project dependencies and add a dependency to their assemble task. We make sure
            // to only add the dependency after BasePlugin is evaluated as otherwise the assemble
            // task may not be available yet.
            List<Project> computedDependencies = getProjectDependencies(projectAfterEvaluation);
            for (Project dependencyProject : computedDependencies) {
              dependencyProject
                  .getPlugins()
                  .withType(
                      BasePlugin.class,
                      unused -> {
                        TaskProvider<Task> assembleTask =
                            dependencyProject.getTasks().named(BasePlugin.ASSEMBLE_TASK_NAME);
                        buildImageTask.configure(task -> task.dependsOn(assembleTask));
                        buildDockerTask.configure(task -> task.dependsOn(assembleTask));
                        buildTarTask.configure(task -> task.dependsOn(assembleTask));
                      });
            }
`
If some kind of dependency  is really required, there should be a minimal dependency setup from any required project. Normally this should only be the "default" gradle artifact - including the classpath.

loosebazooka:
This was added in for: https://github.com/GoogleContainerTools/jib/issues/815. It basically is working as intended:
- The jib module only invokes `classes` since that's all that is required for the build.
- Because of this, the dependency modules also only invoke `classes` - and they will not generate a jar that we can include as a dependency.

^^^ yes this is a consequence of how we do builds. I don't anticipate us trading the performance of skipping the jar creation for better multimodule task selection.

There are a few potential solutions to explore here:
- Perhaps we can find the exact task on the subproject that builds the artifact. We might be able to query which task builds the artifact that we are importing. From the `project dependency` -> `configuration` -> `artifact` -> `builtBy`
[...]

PR #2246, opened 2020-01-24
only depend on assemble on java-library projects

loosebazooka:
limits the scope of: https://github.com/GoogleContainerTools/jib/issues/2184

I tried everything and accessing the builtBy is actually pretty hard. This limits the scope of depending on "assemble" to only subprojects that use the `java-library` plugin.

It is *a* solution, but there are probably others, this should solve the problem for most people right now, except those using `java-library` and want the reduced task dependencies.


loosebazooka:
closed in favor of #2247

PR #2247, opened 2020-01-25
Replace assemble task dependency with runtimeClasspath

### LINKED ISSUE #2184: Too many tasks executed on jib build

seanabraham:
This introduces a new lifecycle task called `jibDependencies` (the name can really be anything, that was just what I came up with) that all the jib tasks depend on and it in turn depends on any requisite tasks / configurations. Being a lifecycle task, it doesn't have any task action itself and essentially exists as a convenient hook for other dependencies. In particular, `jibDependencies` will depend on the `war`/`bootWar` tasks  when applicable, the `jar` task when applicable, and always depends on the projects `runtimeClasspath` configuration.

Gradle allows tasks to depend on `FileCollection`s (rather than just tasks). When doing so Gradle expands the task graph expands to include all the necessary tasks to produce the files listed in the `FileCollection`. Doing so with the `runtimeClasspath` configuration means Gradle will run all the tasks to get all the artifacts necessary to run this project, making it more efficient than using the `assemble` task which will often run slow, and unnecessary tasks. In particular, the main source set's `runtimeClasspath`'s property includes the source set output appended to the project's `runtimeClasspath` configuration so depending on that obviates the need to run even manually run the `classes` task. (See https://docs.gradle.org/current/userguide/java_plugin.html#sec:source_set_properties)

 If we only want to conditionally depend on the runtime classpath I'm happy to make that change though I assume it's never really going to be a bad thing even in the case of the `war` based projects.

Also, for what it's worth — jib itself uses the `runtimeClasspath` to get the list of files to add to the container so it makes sense to use that same collection to set up the task dependencies.
https://github.com/GoogleContainerTools/jib/blob/d3615a6df6f0f7e2654466128e685ea7ee73958a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java#L174 

There are a few different potential approaches here so I'm happy to make changes to go with a different approach. For example, we don't _have_ to introduce the `jibDependencies` task but it made the implementation a little cleaner and also made it so that the changes to tests were minimal (some of them make assumptions about the contents of `task.getDependsOn()` which are no longer true when tasks depend on configurations for example). It might also be useful in the future if there are new dependencies since they can all hook into `jibDependencies`.

This should resolve #2184

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
@@ -26,6 +26,7 @@ import com.google.cloud.tools.jib.gradle.skaffold.SyncMapTask;
 import com.google.cloud.tools.jib.plugins.common.VersionChecker;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -51,7 +52,6 @@ public class JibPlugin implements Plugin<Project> {
   public static final String BUILD_IMAGE_TASK_NAME = "jib";
   public static final String BUILD_TAR_TASK_NAME = "jibBuildTar";
   public static final String BUILD_DOCKER_TASK_NAME = "jibDockerBuild";
-  public static final String DEPENDENCIES_TASK_NAME = "jibDependencies";
   public static final String SKAFFOLD_FILES_TASK_V2_NAME = "_jibSkaffoldFilesV2";
   public static final String SKAFFOLD_INIT_TASK_NAME = "_jibSkaffoldInit";
   public static final String SKAFFOLD_SYNC_MAP_TASK_NAME = "_jibSkaffoldSyncMap";
@@ -178,19 +178,19 @@ public class JibPlugin implements Plugin<Project> {
           TaskProvider<Task> warTask = TaskCommon.getWarTaskProvider(projectAfterEvaluation);
           TaskProvider<Task> bootWarTask =
               TaskCommon.getBootWarTaskProvider(projectAfterEvaluation);
-          TaskProvider<Task> jibDependenciesTask = tasks.register(DEPENDENCIES_TASK_NAME);
+          List<Object> jibDependencies = new ArrayList<>();
           if (warTask != null || bootWarTask != null) {
             // Have all tasks depend on the 'war' and/or 'bootWar' task.
             if (warTask != null) {
-              jibDependenciesTask.configure(it -> it.dependsOn(warTask));
+              jibDependencies.add(warTask);
             }
             if (bootWarTask != null) {
-              jibDependenciesTask.configure(it -> it.dependsOn(bootWarTask));
+              jibDependencies.add(bootWarTask);
             }
           } else if ("packaged".equals(jibExtension.getContainerizingMode())) {
             // Have all tasks depend on the 'jar' task.
             TaskProvider<Task> jarTask = projectAfterEvaluation.getTasks().named("jar");
-            jibDependenciesTask.configure(it -> it.dependsOn(jarTask));
+            jibDependencies.add(jarTask);
 
             if (projectAfterEvaluation.getPlugins().hasPlugin("org.springframework.boot")) {
               Jar jar = (Jar) jarTask.get();
@@ -205,10 +205,10 @@ public class JibPlugin implements Plugin<Project> {
                   .getPlugin(JavaPluginConvention.class)
                   .getSourceSets()
                   .getByName(MAIN_SOURCE_SET_NAME);
-          jibDependenciesTask.configure(it -> it.dependsOn(mainSourceSet.getRuntimeClasspath()));
+          jibDependencies.add(mainSourceSet.getRuntimeClasspath());
 
           jibTaskProviders.forEach(
-              provider -> provider.configure(task -> task.dependsOn(jibDependenciesTask)));
+              provider -> provider.configure(task -> task.setDependsOn(jibDependencies)));
         });
   }
 }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
@@ -51,6 +51,7 @@ public class JibPlugin implements Plugin<Project> {
   public static final String BUILD_IMAGE_TASK_NAME = "jib";
   public static final String BUILD_TAR_TASK_NAME = "jibBuildTar";
   public static final String BUILD_DOCKER_TASK_NAME = "jibDockerBuild";
+  public static final String DEPENDENCIES_TASK_NAME = "jibDependencies";
   public static final String SKAFFOLD_FILES_TASK_V2_NAME = "_jibSkaffoldFilesV2";
   public static final String SKAFFOLD_INIT_TASK_NAME = "_jibSkaffoldInit";
   public static final String SKAFFOLD_SYNC_MAP_TASK_NAME = "_jibSkaffoldSyncMap";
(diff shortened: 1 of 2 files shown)


seanabraham:
@googlebot I signed it!

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -212,11 +143,19 @@ public void testWebAppProject() {
...
+    Assert.assertTrue(taskDependencies.contains(warTask));
+
     for (String taskName : KNOWN_JIB_TASKS) {
-      List<TaskProvider<?>> taskProviders =
-          (List<TaskProvider<?>>) tasks.getByPath(taskName).getDependsOn().iterator().next();
-      Assert.assertEquals(1, taskProviders.size());
-      Assert.assertEquals(warTask, taskProviders.get(0).get());
+      TaskProvider<?> jibDependenciesTaskProvider =
+          (TaskProvider<?>) tasks.getByPath(taskName).getDependsOn().iterator().next();
+      Assert.assertEquals(jibDependenciesTaskProvider.get(), jibDependenciesTask);
```
seanabraham: These test changes (and the ones below) are now checking for whether the `jibDependencies` task depends on the requisite tasks (like the `warTask` or `bootWarTask` below) and whether all the jib tasks depend on `jibDependencies` which is the only task they need to depend on now.

loosebazooka:
Nice, this is the better solution that #2246 . I'll take a deeper look in the next few days.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
```diff
@@ -162,68 +165,49 @@ public void apply(Project project) {
...
-                jar.setEnabled(true);
-                jar.getArchiveClassifier().set("original");
-              }
-            } else {
-              // Have all tasks depend on the 'classes' task.
-              dependsOnTask.add(projectAfterEvaluation.getTasks().named("classes"));
+          TaskProvider<Task> warTask = TaskCommon.getWarTaskProvider(projectAfterEvaluation);
+          TaskProvider<Task> bootWarTask =
+              TaskCommon.getBootWarTaskProvider(projectAfterEvaluation);
+          TaskProvider<Task> jibDependenciesTask = tasks.register("jibDependencies");
```
loosebazooka: I'll check when I can, but this doesn't show up in `gradlew tasks` does it?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -230,15 +171,24 @@ public void testWebAppProject_bootWar() {
...
     Task bootWarTask = tasks.getByPath(":bootWar");
+    Task jibDependenciesTask = tasks.getByPath(":jibDependencies");
     Assert.assertNotNull(warTask);
     Assert.assertNotNull(bootWarTask);
 
+    Set<Task> taskDependencies =
+        jibDependenciesTask
+            .getDependsOn()
+            .stream()
+            .filter(it -> it instanceof TaskProvider)
```
loosebazooka: we've been filtering like `.filter(TaskProvider.class::isInstance)`

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -252,16 +202,25 @@ public void testWebAppProject_bootWarDisabled() {
...
+    Task jibDependenciesTask = tasks.getByPath(":jibDependencies");
     Assert.assertNotNull(warTask);
     Assert.assertNotNull(bootWarTask);
     bootWarTask.setEnabled(false); // should depend on bootWar even if disabled
 
+    Set<Task> taskDependencies =
+        jibDependenciesTask
+            .getDependsOn()
+            .stream()
+            .filter(it -> it instanceof TaskProvider)
```
loosebazooka: `filter(...` like above

loosebazooka:
integ tests look good too.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
seanabraham: Though I can't remember what determines whether or not tasks show up in the `gradlew tasks` output, it doesn't look like it does in this case.

image

seanabraham:
@loosebazooka thanks for the review! I addressed the comments, and confirmed that the `jibDependencies` task doesn't show up in the `gradlew tasks` output. Let me know if there are any other changes you'd like to see.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
loosebazooka: Can you use a const as above https://github.com/GoogleContainerTools/jib/pull/2247/files#diff-aef8bf567c91df6622aa0ea29ed0c682R53

```
JIB_DEPENDECIES_TASK_NAME="jibDependencies"
```

and then reuse it in `JibPluginTest.java`.

loosebazooka:
@seanabraham thanks again. I think adding `jibDependencies` as an explicit task is not something we want to commit to at the moment. Could you just use the previous mechanism. I understand that the plugin logic and tests become more complicated, but the surface of jib exposed to the user is reduced.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
```diff
@@ -162,68 +166,49 @@ public void apply(Project project) {
...
-                jar.setEnabled(true);
-                jar.getArchiveClassifier().set("original");
-              }
-            } else {
-              // Have all tasks depend on the 'classes' task.
-              dependsOnTask.add(projectAfterEvaluation.getTasks().named("classes"));
+          TaskProvider<Task> warTask = TaskCommon.getWarTaskProvider(projectAfterEvaluation);
+          TaskProvider<Task> bootWarTask =
+              TaskCommon.getBootWarTaskProvider(projectAfterEvaluation);
+          TaskProvider<Task> jibDependenciesTask = tasks.register(DEPENDENCIES_TASK_NAME);
```
loosebazooka: Sorry, after thinking about this while we were doing the release, I realized I'm not quite convinced this is how we want to do this. Maybe in the future, the problem is going backwards/deprecating something has a high cost.

Can we just use the mechanism used previously: https://github.com/GoogleContainerTools/jib/pull/2247/files#diff-aef8bf567c91df6622aa0ea29ed0c682L175

seanabraham: Sure, I can make that change

seanabraham:
@loosebazooka refactored to remove the `jibDependencies` task and instead have each of the jib tasks depend directly on relevant items (`war`, `bootWar`, etc. tasks and the runtimeClasspath configuration from the main source set)

loosebazooka:
Awesome, thanks @seanabraham

PR #2251, opened 2020-01-27
Add scripts to update GCS with latest version string

### LINKED ISSUE #2250: Automatically update latest version on GCS in release process
Either add new scripts or update existing scripts to update GCS with new version strings when we make a new release.

TadCordle:
Fixes #2250.

### DIFF
--- a/jib-gradle-plugin/scripts/update_gcs_latest.sh
+++ b/jib-gradle-plugin/scripts/update_gcs_latest.sh
@@ -1,0 +1,41 @@
+#!/bin/bash -
+# Usage: ./scripts/update_gcs_latest.sh <release version>
+
+set -e
+
+EchoRed() {
+	echo "$(tput setaf 1; tput bold)$1$(tput sgr0)"
+}
+EchoGreen() {
+	echo "$(tput setaf 2; tput bold)$1$(tput sgr0)"
+}
+
+Die() {
+	EchoRed "$1"
+	exit 1
+}
+
+# Usage: CheckVersion <version>
+CheckVersion() {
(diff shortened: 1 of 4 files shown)


jib-maven-plugin/scripts/prepare_release.sh
```diff
@@ -58,3 +58,4 @@ EchoGreen 'File a PR for the new release branch:'
 echo https://github.com/GoogleContainerTools/jib/pull/new/${BRANCH}
 
 EchoGreen "Merge the PR after the plugin is released."
+EchoGreen "Run './scripts/update_gcs_latest ${VERSION}' when the release is complete to update the latest version string on GCS."
```
TadCordle: Q: I put the GCS update stuff in a separate script since the maven release process takes some time. Should we just combine the scripts anyway and not worry about GCS reporting an unrelease version for a few minutes/hours?

chanseokoh: I think we should update the version only after we verify a new release is available for consumption for both Gradle and Maven.

Add `.sh`.

jib-gradle-plugin/scripts/update_gcs_latest.sh
```diff
@@ -0,0 +1,42 @@
...
+
+[ $# -ne 1 ] && Die "Usage: ./scripts/update_gcs_latest.sh <release version>"
+
+CheckVersion $1
+
+versionString="{\"latest\":\"$1\"}"
+destination="gs://jib-versions/jib-gradle"
+
+echo $versionString > jib-gradle
+gsutil rm $destination
```
chanseokoh: I believe this is unnecessary and the file will be overwritten. This is safer in the case `gsutil cp` fails.

PR #2256, opened 2020-01-28
Fix integration test

chanseokoh:


chanseokoh:
Ah, wait. Seems not enough.

jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -694,7 +694,10 @@ public void testExecute_springBootPackaged()
                 "-c",
                 "/app/classpath/spring-boot-0.1.0.original.jar")
             .run();
-    Assert.assertEquals("2749 /app/classpath/spring-boot-0.1.0.original.jar\n", sizeOutput);
+    Assert.assertThat(
+        sizeOutput, CoreMatchers.containsString(" /app/classpath/spring-boot-0.1.0.original.jar"));
+    int fileSize = Integer.parseInt(sizeOutput.substring(0, sizeOutput.indexOf(' ')));
+    Assert.assertTrue(fileSize < 3000);
```
loosebazooka: I'm not sure I understand the change here

loosebazooka: was the generated file a different size?

chanseokoh: Yeah, somehow the size was 2748 instead of 2749 when running locally. The point of the test is to see that the JAR is not fat, so it is sufficient to see it is small.

loosebazooka: I see, okay. Can we add a bug to make this test more robust? This seems fine for the purposes of fixes integ-tests.

chanseokoh: I actually think this is a final form to be robust. Thinking how a jar is created by Maven, now I recall that it is almost impossible to reproducibly create the exact same jar. (`MANIFEST.MF` will have build information for example.)

loosebazooka: What if future springboot examples exceed the size? I guess we don't update our springboot examples all that often. test can be update then.

chanseokoh: Yeah, we pin down the Spring Boot version, so I don't think the size will ever noticeably increase.

PR #2261, opened 2020-01-31
Fall back to initial authenticator if bad response

### LINKED ISSUE #2258: Tekton on OpenShift: authentication issue with internal OpenShift registry
From https://github.com/GoogleContainerTools/jib/issues/2074#issuecomment-579921534

> I did a new test using the following parameters top of ocp3 and that fails using jib-maven 2.0.0
>
> ```
>     - name: build-and-push
>       image: gcr.io/cloud-builders/mvn
>       command:
>         - mvn
>         - compile
>         - com.google.cloud.tools:jib-maven-plugin:2.0.0:build
>         - -Djib.from.image=registry.redhat.io/redhat-openjdk-18/openjdk18-openshift
>         - -Djib.allowInsecureRegistries=true
>         - -Djib.from.auth.username=my-user
>         - -Djib.from.auth.password=my-pwd
>         - -Dimage=$(outputs.resources.image.url)
> ```

> If I add to the config too `-DsendCredentialsOverHttp=true`, then we got this error
> ```
> [ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:2.0.0:build (default-cli) on project rest-http: 
> Build image failed: Failed to authenticate with registry docker-registry.default.svc:5000/test/sb-image 
> because: server did not return 'WWW-Authenticate: Bearer' header: Basic 
> realm=openshift,error="access denied" -> [Help 1]
> ```

chanseokoh:
Fixes #2258.

If the server returns unexpected `WWW-Authenticate` responses  when attempting to refresh a bearer token, fall back to use the initial authenticator that succeeded to get the first token. `RegistryAuthenticator` is immutable and safe to reuse.

And fixing this will naturally make Jib to surface 401 eventually after 4 refresh attempts fail when the initial token did not have enough access scope.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
@@ -262,6 +262,8 @@ public class RegistryClient {
   // mutable
   private final AtomicReference<Authorization> authorization = new AtomicReference<>();
   private boolean readOnlyBearerAuth;
+  private final AtomicReference<RegistryAuthenticator> initialBearerAuthenticator =
+      new AtomicReference<>();
 
   /**
    * Instantiate with {@link #factory}.
@@ -344,6 +346,7 @@ public class RegistryClient {
       return false; // server returned "WWW-Authenticate: Basic ..."
     }
 
+    initialBearerAuthenticator.set(authenticator.get());
     if (readOnlyBearerAuth) {
       authorization.set(authenticator.get().authenticatePull(credential));
     } else {
@@ -375,10 +378,13 @@ public class RegistryClient {
       }
     }
 
-    throw new RegistryAuthenticationFailedException(
-        registry,
-        repository,
-        "server did not return 'WWW-Authenticate: Bearer' header: " + wwwAuthenticate);
+    eventHandlers.dispatch(
+        LogEvent.debug(
+            "server did not return 'WWW-Authenticate: Bearer' header. Actual: " + wwwAuthenticate));
+    if (readOnlyBearerAuth) {
+      return Verify.verifyNotNull(initialBearerAuthenticator.get()).authenticatePull(credential);
+    }
+    return Verify.verifyNotNull(initialBearerAuthenticator.get()).authenticatePush(credential);
   }
 
   /**
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryClientTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryClientTest.java
@@ -20,6 +20,7 @@ import com.google.cloud.tools.jib.api.Credential;
 import com.google.cloud.tools.jib.api.DescriptorDigest;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.api.RegistryException;
+import com.google.cloud.tools.jib.api.RegistryUnauthorizedException;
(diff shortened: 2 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -375,10 +378,13 @@ private Authorization refreshBearerAuth(@Nullable String wwwAuthenticate)
...
-        registry,
-        repository,
-        "server did not return 'WWW-Authenticate: Bearer' header: " + wwwAuthenticate);
+    eventHandlers.dispatch(
+        LogEvent.debug(
+            "server did not return 'WWW-Authenticate: Bearer' header. Actual: " + wwwAuthenticate));
+    if (readOnlyBearerAuth) {
+      return Verify.verifyNotNull(initialBearerAuthenticator.get()).authenticatePull(credential);
+    }
+    return Verify.verifyNotNull(initialBearerAuthenticator.get()).authenticatePush(credential);
```
TadCordle: Would it be worth it to pull this out into a function to reduce duplication between here and L374-377?

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryClientTest.java
```diff
@@ -168,6 +169,71 @@ public void testAutomaticTokenRefresh()
...
+    RegistryClient registryClient = createRegistryClient(null);
+    Assert.assertTrue(registryClient.doPushBearerAuth());
+
+    try {
+      registryClient.checkBlob(digest);
+      Assert.fail("Should have given up refreshing after 4 attempts");
+    } catch (RegistryUnauthorizedException ex) {
+      Assert.assertEquals(401, ex.getHttpResponseException().getStatusCode());
+      Assert.assertEquals(5, authServer.getTotalResponsesServed());
+      Assert.assertEquals(6, registry.getTotalResponsesServed());
```
briandealwis: Should this be 5? (1 successful, and then 4 attempts)?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -344,6 +346,7 @@ private boolean doBearerAuth(boolean readOnlyBearerAuth) throws IOException, Reg
       return false; // server returned "WWW-Authenticate: Basic ..."
     }
 
+    initialBearerAuthenticator.set(authenticator.get());
```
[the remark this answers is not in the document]
chanseokoh: Should be okay, since `refreshBearerAuth` is designed in the way that it's called only when `authorization` is a bearer auth token.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
chanseokoh: I think this is fine, considering that there's a difference between `authenticator` and `initialBearerAuthenticator` that the latter is nullable.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/RegistryClientTest.java
chanseokoh: It's 6: 1 response asking to auth + 4 unauth responses for 4 refresh attempts + 1 final unauth response that will be thrown as `RegistryUnauthorizedException`. I'll put a comment in the code.

chanseokoh:
Oops, will add a CHANGELOG entry.

Issue #2150, opened 2019-11-14
Unify test resources


loosebazooka:
We should share as many test resources as possible. This means
1. putting all test resources in a common location
2. rewriting our `TestProject` fixtures to access them
3. fix tests to address any deviation in test projects that have happened over time.
4. and so much more!!!

TadCordle:
Would be nice to streamline the plugin integration test code a bit more as well in general. The tests themselves work fine, but they operate using a lot of awkward halfway-generic helper methods and classes. Some of the methods cover multiple cases, but often require further modification/rewrites when we introduce new tests (e.g. I feel like every time I have to add a new integration test, I have to reread a lot of code to re-learn how it works and end up adding new flags to some helper methods or something to fit the new use case).

Almost feels like we should either just be copy-pasting the code in the tests so they're easier to read/follow, or make a more general/flexible/future-proof testing framework.

PR #2263, opened 2020-02-03
Fix incorrect skaffold init output in multi-module projects

### LINKED ISSUE #2262: _skaffold-init task multimodule logic issues
The issues with the current logic (at least in the maven plugin, gradle task might have the same issues): 

- when a submodule has a different parent specified (e.g. spring boot starter pom) then our current logic fails to set the project id as the parent comes back as null in the InitMojo - this creates issues in multimodule projects (e.g. the Cloud Code java guestbook sample) because the listed projects don't have anything in them
- I'm not sure how frequent nested modules are but currently we don't handle those well either
- we are using the `artifactId` instead of the module name to return a `project` which is a problem as that is interpreted on the skaffold side as a directory

I won't have time to fully implement the thing with tests and everything, but as I played around with the code this seemed to do the job altogether: 

```java
@Mojo(name = InitMojo.GOAL_NAME, requiresDependencyCollection = ResolutionScope.NONE)
public class InitMojo extends JibPluginConfiguration {

  @VisibleForTesting static final String GOAL_NAME = "_skaffold-init";

  @Override
  public void execute() throws MojoExecutionException {
    checkJibVersion();
    MavenProject project = getProject();
    MavenProject topLevelProject = getSession().getTopLevelProject();
    if (!project.equals(topLevelProject)) {
      return;
    }

    if (isParentProject(project)) {
      Set<MavenProject> projects = collectLeafProjects(project);
      for (MavenProject p : projects) {
        print(p, true);
      }
    } else {
      print(project, false);
    }
  }

  private Set<MavenProject> collectLeafProjects(MavenProject project) {
    Set<MavenProject> projects = new HashSet<>();
    for (MavenProject p : project.getCollectedProjects()) {
      if (isParentProject(p)) {
        projects.addAll(collectLeafProjects(p));
      } else {
        projects.add(p);
      }
    }
    return projects;
  }

  private boolean isParentProject(MavenProject project) {
    return project.getPackaging().equals("pom") && project.getModules().size() > 0;
  }

  private void print(MavenProject project, boolean setProject) throws MojoExecutionException {
    SkaffoldInitOutput skaffoldInitOutput = new SkaffoldInitOutput();
    skaffoldInitOutput.setImage(getTargetImage());
    if (setProject) {
      skaffoldInitOutput.setProject(project.getGroupId() + ":" + project.getArtifactId());
    }
    System.out.println();
    System.out.println("BEGIN JIB JSON");
    try {
      System.out.println(skaffoldInitOutput.getJsonString());
    } catch (IOException ex) {
      throw new MojoExecutionException(ex.getMessage(), ex);
    }
  }
}

```

TadCordle:
Fixes #2262.

### DIFF
--- a/jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/pom.xml
+++ b/jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/pom.xml
@@ -34,17 +34,4 @@
       </plugins>
     </pluginManagement>
   </build>
-
-  <repositories>
-     <repository>
-       <id>snapshots-repo</id>
-       <url>https://oss.sonatype.org/content/repositories/snapshots</url>
-       <releases>
-         <enabled>false</enabled>
-       </releases>
-       <snapshots>
-         <enabled>true</enabled>
-       </snapshots>
-     </repository>
-   </repositories>
 </project>
--- a/jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/pom.xml
+++ b/jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/pom.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                             http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <!-- Defines artifact information for this parent POM. -->
+  <groupId>com.jib.test</groupId>
+  <artifactId>spring-boot-multi</artifactId>
+  <packaging>pom</packaging>
+  <version>1.0.0.TEST-SNAPSHOT</version>
+  <name>multimodule</name>
+
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+  </properties>
+
+  <!-- The modules that are part of this project. -->
+  <modules>
+    <module>service-1</module>
+    <module>service-2</module>
+  </modules>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>com.google.cloud.tools</groupId>
+          <artifactId>jib-maven-plugin</artifactId>
+          <version>@@PluginVersion@@</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+
+  <repositories>
+     <repository>
+       <id>snapshots-repo</id>
+       <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+       <releases>
+         <enabled>false</enabled>
+       </releases>
+       <snapshots>
+         <enabled>true</enabled>
+       </snapshots>
+     </repository>
+   </repositories>
+</project>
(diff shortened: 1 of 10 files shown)


TadCordle:
Actually, did some manual tests with this after putting the PR up, and realized the target image will no longer print out for submodules.

loosebazooka:
how are we invoking init? What happens if the project doesn't have jib on it?

jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+       <id>snapshots-repo</id>
+       <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+       <releases>
+         <enabled>false</enabled>
+       </releases>
+       <snapshots>
+         <enabled>true</enabled>
+       </snapshots>
+     </repository>
+   </repositories>
```
chanseokoh: This is probably not needed for this test.

jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/service-1/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>com.google.cloud.tools</groupId>
+          <artifactId>jib-maven-plugin</artifactId>
+          <version>@@PluginVersion@@</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
```
chanseokoh: I bet you've copied this whole setup from another directory, but isn't this redundant?

jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/service-2/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+  <artifactId>service-2</artifactId>
+  <version>1.0.0.TEST-SNAPSHOT</version>
+
+  <parent>
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-parent</artifactId>
+    <version>2.1.6.RELEASE</version>
+  </parent>
+
+  <build>
```
chanseokoh: ditto

jib-maven-plugin/src/test/resources/maven/projects/spring-boot-multi/service-1/pom.xml
loosebazooka: is this maybe an indication that we might need to do some test cleanup?

loosebazooka: oops, just found https://github.com/GoogleContainerTools/jib/issues/2150 again.

chanseokoh: Ah, then I think the `pluginManagement` in the root `pom.xml` is unnecessary?

TadCordle: I think I want to keep it to ensure that the output isn't printed on the root project, even though the Jib plugin is applied?

TadCordle:
@balopat would you mind checking out this branch and seeing if it fixes the issues you were seeing? I took a different approach than you suggested, so it's possible I may have missed an edge case, but from my manual testing I haven't noticed any issues yet.

balopat:
Looks good to me! Thank you for fixing it @TadCordle!!

PR #2268, opened 2020-02-06
Add support for packaging a native-image-produced executable

briandealwis:
This is a prototype for integrating GraalVM's `native-image` support with Jib.  It's currently Maven-only.  This implementation leverages the GraalVM's [`native-image-maven-plugin`](https://medium.com/graalvm/simplifying-native-image-generation-with-maven-plugin-and-embeddable-configuration-d5b283b92f57) ([source](https://github.com/oracle/graal/tree/master/substratevm/src/native-image-maven-plugin)) to do the heavy lifting — and future-proofs Jib.   This PR attempts to find the resulting executable by examining the `native-image-maven-plugin` blocks.

There's a small example in `examples/native-image`.  Note that you'll need to need to use a JVM that supports the JVMCI (JVM Compiler Interface); I haven't had much luck with anything other than the GraalVM JVMs.

Note that you need to build on Linux since `native-image` doesn't support cross-compilation.  If you do anything else, you'll get a puzzling error:
```sh
$ docker run -it --rm native-hello
standard_init_linux.go:211: exec user process caused "exec format error"
```

I had to add a new generator method on JavaContainerBuilder to generate a JibContainerBuilder instance as all the logic for layering doesn't actually matter.

Things to do:
- Check that the current platform matches the image
- Figure out what to do with Gradle since there doesn't seem to be a definitive Gradle `native-image` plugin.  Perhaps we can just make the executable path explicitly configurable.


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -385,8 +386,15 @@ static JavaContainerBuilder getJavaContainerBuilderWithBaseImage(
       RawConfiguration rawConfiguration, ProjectProperties projectProperties)
       throws MainClassInferenceException, InvalidAppRootException, IOException,
           InvalidContainerizingModeException {
+    ContainerizingMode mode = getContainerizingModeChecked(rawConfiguration, projectProperties);
     AbsoluteUnixPath appRoot = getAppRootChecked(rawConfiguration, projectProperties);
 
+    if (mode == ContainerizingMode.NATIVE_IMAGE) {
```
loosebazooka: Can we output a warning here (EXPERIMENTAL), I don't want anyone to hold us to this implementation.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ContainerizingModeTest.java
```diff
@@ -26,6 +26,7 @@
   public void testFrom_validValues() throws InvalidContainerizingModeException {
     Assert.assertEquals(ContainerizingMode.EXPLODED, ContainerizingMode.from("exploded"));
     Assert.assertEquals(ContainerizingMode.PACKAGED, ContainerizingMode.from("packaged"));
+    Assert.assertEquals(ContainerizingMode.NATIVE_IMAGE, ContainerizingMode.from("native_image"));
```
loosebazooka: should `native_image` -> `nativeImage`?

briandealwis: ContainerizingMode currently has a strong requirement that it all be in lower-case:
https://github.com/GoogleContainerTools/jib/blob/4f438d4c66775a34f433eacd5e3758d69ac00d03/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ContainerizingMode.java#L42-L47

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
briandealwis: Ok

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ContainerizingModeTest.java
loosebazooka: Oh so this is an issue directly tied to enums, which means `native-image` would also be invalid? It just feels weird to use `_` when we don't use it anywhere else.

loosebazooka: but I guess this will be very explicitly experimental so we can change it whenever.

briandealwis:
I realized I was wrong in how `native-image-maven-plugin` determines the executable name in the absence of an `<imageName>`.

briandealwis:
Hmm, now that you mention it, the native-image packaging needs to still include the extra-files and resources.

loosebazooka:
I thought resources are already included, no (in the binary)?

Passarinho4:
Hi @briandealwis @loosebazooka !

What are you going to do about this PR? I think this concept is really interesting and I want to test it with my projects. Could I help you or you are going to merge it? 

Cheers

briandealwis:
Hi @Passarinho4 — my hope is that this can be re-implemented using @chanseokoh's plugin extension mechanism.

briandealwis:
Closing this PR — this functionality should be implemented using the new extension mechanism.

chanseokoh:
We will work on creating a new Jib plugin extension to support this soon. But I think anyone can easily [create an extension](https://github.com/GoogleContainerTools/jib-extensions) to make this work.

chanseokoh:
> I thought resources are already included, no (in the binary)?

Although I failed to verify myself with a sample code, I think the answer is yes: https://www.graalvm.org/docs/Native-Image/user/RESOURCES

chanseokoh:
@Passarinho4 we've published a [Jib Maven extension](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-native-image-extension-maven) that enables containerizing a "native-image" project set up with [`native-image-maven-plugin`](https://www.graalvm.org/docs/reference-manual/native-image/#integration-with-maven). You can apply the extension to recent Jib versions.

We are willing to add another extension for Gradle when they have a Gradle counterpart of `native-image-gradle-plugin`.

Sineaggi:
@chanseokoh I think https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html is Gradle's counterpart. Looks like the both of them are [hosted in the same repo](https://github.com/graalvm/native-build-tools/tree/master/native-gradle-plugin).

PR #2277, opened 2020-02-10
Add "Launch Problems" section to FAQ and highlight symlink issue

briandealwis:
- Refactors the FAQ to separate build-time problems from launch-time problems.
- Adds a section with image launch problems, and include the symlink-overwriting problem.
- Highlights symlink-overwriting in the jib-maven-plugin and jib-gradle-plugin docs.


docs/faq.md
```diff
@@ -576,31 +612,25 @@ plugins {
...
 ```
 
-### How do I view debug logs for Jib?
+On examining the container structure with [Dive](https://github.com/wagoodman/dive), the user discovered that the contents of the `/lib` directory had disappeared.
 
-Maven: use `mvn -X -Djib.serialize=true` to enable more detailed logging and serialize Jib's actions.
+The user had used Jib's ability to install extra files into the image ([Maven](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#adding-arbitrary-files-to-the-image), [Gradle](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#adding-arbitrary-files-to-the-image)) to install a DLL by placing it in `src/main/jib/lib/file.dll`. This would normally cause the `file.dll` to be installed in the image as `/lib/file.dll`. But the user's base image used symbolic links for `/lib` and `/lib64`. Jib does not follow such symbolic links when creating the image. And at container initialization time, Docker treats these symlinks as a file, and thus the symbolic link was replaced with `/lib`. As a result, none of the system shared libraries were resolved and dynamically-linked programs failed.
```
chanseokoh:
```suggestion
The user had used Jib's ability to install extra files into the image ([Maven](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#adding-arbitrary-files-to-the-image), [Gradle](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#adding-arbitrary-files-to-the-image)) to install a library file by placing it in `src/main/jib/lib/libfoo.so`. This would normally cause the `libfoo.so` to be installed in the image as `/lib/libfoo.so`. But `/lib` and `/lib64` in the user's base image were symbolic links. Jib does not follow such symbolic links when creating the image. And at container initialization time, Docker treats these symlinks as a file, and thus the symbolic link was replaced with `/lib` as a new directory. As a result, none of the system shared libraries were resolved and dynamically-linked programs failed.
```

docs/faq.md
```diff
@@ -526,6 +529,39 @@ If you encounter issues interacting with a registry other than `UNAUTHORIZED`, c
...
+java.util.logging.ConsoleHandler.level=ALL
+
+# CONFIG hides authentication data
+# ALL includes authentication data
+com.google.api.client.http.level=CONFIG
+```
+
+And then launch your build tool as follows:
+```sh
+mvn -Djava.util.logging.config.file=path/to/log.properties -Djib.serialize=true -Djib.console=plain ...
```
chanseokoh:
```suggestion
mvn -X -Djava.util.logging.config.file=path/to/log.properties -Djib.serialize=true -Djib.console=plain ...
```

docs/faq.md
```diff
@@ -526,6 +529,39 @@ If you encounter issues interacting with a registry other than `UNAUTHORIZED`, c
...
+```
+or
+```sh
+gradle -Djava.util.logging.config.file=path/to/log.properties -Djib.serialize=true -Djib.console=plain ...
+```
+
```
chanseokoh:
```suggestion
If you don't need DEBUG-level logs, omit `-X` (Maven) or `--debug` (Gradle).

```

docs/faq.md
```diff
@@ -526,6 +529,41 @@ If you encounter issues interacting with a registry other than `UNAUTHORIZED`, c
...
+And then launch your build tool as follows:
+```sh
+mvn -X -Djava.util.logging.config.file=path/to/log.properties -Djib.serialize=true -Djib.console=plain ...
+```
+or
+```sh
+gradle --debug -Djava.util.logging.config.file=path/to/log.properties -Djib.serialize=true -Djib.console=plain ...
+```
+If you don't need DEBUG-level logs, omit `-X` (Maven) or `--debug` (Gradle).
+If you don't need DEBUG-level logs, omit `-X` (Maven) or `--debug` (Gradle).
```
chanseokoh: I guess the button was pressed twice.

docs/faq.md
briandealwis: I actually find it useful to just see the network traffic, independent of the Maven verbose logs.  I changed the note below to note that the `-X`/`--gradle` args can be useful.

docs/faq.md
briandealwis: The commit-change function is pretty slow!

PR #2287, opened 2020-02-11
Remove Kokoro Mac workaround that is no longer needed

chanseokoh:
They said this is fixed.


chanseokoh:
Indeed, the MacOS presubmit is passing.

PR #2289, opened 2020-02-12
Do not overwrite Gradle task dependencies

lightoze:
This fixes a bug where user-configured task dependencies (`tasks.jib.dependsOn '...'` immediately after `jib {}` section) are overwritten by the plugin.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
@@ -208,7 +208,7 @@ public class JibPlugin implements Plugin<Project> {
           jibDependencies.add(mainSourceSet.getRuntimeClasspath());
 
           jibTaskProviders.forEach(
-              provider -> provider.configure(task -> task.setDependsOn(jibDependencies)));
+              provider -> provider.configure(task -> jibDependencies.forEach(task::dependsOn)));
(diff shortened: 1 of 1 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
```diff
@@ -208,7 +208,7 @@ public void apply(Project project) {
           jibDependencies.add(mainSourceSet.getRuntimeClasspath());
 
           jibTaskProviders.forEach(
-              provider -> provider.configure(task -> task.setDependsOn(jibDependencies)));
+              provider -> provider.configure(task -> jibDependencies.forEach(task::dependsOn)));
```
chanseokoh: Question: does `task.dependsOn(jibDependencies)` not work? (`dependsOn()` accepts `Object...`, but I am not actually sure if passing `List<TaskProvider<Task>>` will work as intended.)

loosebazooka: @chanseokoh I was curious and tried it out, but it looks like the alternative is to do:
```
task.setDependency(jibDependencies.toArray())
```
For whatever reason, the object translation doesn't work well for collections?

loosebazooka:
Hrmm mac os invocation hung, I retriggered it.

Issue #2294, opened 2020-02-12
Warning from empty global Jib config


chanseokoh:
I first observed this in this [log](https://github.com/GoogleContainerTools/jib/pull/2264#issuecomment-583288030).

I also hit this on Windows:
```
[WARNING] Failed to read global Jib config: No content to map due to end-of-input
 at [Source: (sun.nio.ch.ChannelInputStream); line: 1, column: 0]; you may need to fix or delete C:\Users\chanseok\AppData\Local\Google\Jib\Config\config.json;
```
The code for this is in [`UpdateChcker.java`](https://github.com/GoogleContainerTools/jib/blob/master/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java#L116).

In my case, the config file (`config.json`) is empty (0-byte). I think it's easy to generate an empty file.

First, we need to understand when and why we may create an empty file. ~~And if it seems not possible to prevent that, we should find a way to suppress this warning.~~ Turns out, we never intended to create an empty file and doing so is a bug that wipes out the values set by the user. We should never let this happen.

loosebazooka:
consider swapping the error message around to make the location of the config file more prominent

chanseokoh:
I realized just opening an `OutputStream` creates an empty file. The following code creates an empty `/tmp/foo` by the time the control reaches line 2:
```java
    1: OutputStream os = Files.newOutputStream(Paths.get("/tmp/foo"));
    2: Thread.sleep(30000);
    3: os.toString();
```

So we can actually simulate this issue with
```java
        try (OutputStream outputStream = Files.newOutputStream(configFile)) {
          if (true) while (true) try { Thread.sleep(30000); } catch (InterruptedException e) {}
          JsonTemplateMapper.writeTo(config, outputStream);
```

briandealwis:
Oh!  I wonder if something odd could happen if you ran two jib instances simultaneously?

chanseokoh:
We've seen this at least two times. Maybe we should create a `config.json.tmp` file in the same directory and rename to avoid accidental truncating.

alkoclick:
We have recently started observing this exception and I think it's because we use parallel gradle building

> Oh! I wonder if something odd could happen if you ran two jib instances simultaneously?

Well... :laughing:

loosebazooka:
@alkoclick Is this happening on a CI?

loosebazooka:
Also, are you using `--parallel`

[...]

PR #2295, opened 2020-02-13
Treat empty update check config as non-existent

TadCordle:
I may keep looking into #2294 to see if I can somehow reproduce the error without intentionally creating an empty config file, but this should avoid the warning message in the meantime. Empty config files are now treated as empty, meaning they will be regenerated with defaults.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
@@ -102,7 +102,7 @@ public class UpdateChecker {
 
     try {
       // Check global config
-      if (Files.exists(configFile)) {
+      if (Files.exists(configFile) && Files.size(configFile) > 0) {
         // Abort if update checks are disabled
         try {
           ConfigJsonTemplate config =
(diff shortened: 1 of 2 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -134,7 +133,7 @@ void setDisableUpdateCheck(boolean disableUpdateCheck) {
       }
 
       // Check time of last update check
-      if (Files.exists(lastUpdateCheck)) {
+      if (Files.exists(lastUpdateCheck) && Files.size(lastUpdateCheck) > 0) {
```
chanseokoh: But this means if we ever create an empty file, we can never escape the situation and always call the endpoint. I think for this file, we don't need to check the size, because we don't show a warning. And this will auto-fix the empty file problem by deleting it.

TadCordle: If we create an empty file won't it be fixed when it's written to at L163? I guess it doesn't matter.

chanseokoh: Oh! I didn't realize that.

PR #2300, opened 2020-02-18
Handle and ignore explicitly defined "/" tar entry

### LINKED ISSUE #2195: NPE when adding LayerEntry relative to the root '/'
**Environment**:

- *Jib version:* jib-core 0.12.0
- *Build tool:* cram
- *OS:* macOS

**Description of the issue**:

Seeing NPEs when using jib-core to create a layer with content placed at "/":
```
Caused by: java.lang.NullPointerException
	at com.google.cloud.tools.jib.image.ReproducibleLayerBuilder$UniqueTarArchiveEntries.add(ReproducibleLayerBuilder.java:75)
	at com.google.cloud.tools.jib.image.ReproducibleLayerBuilder$UniqueTarArchiveEntries.access$100(ReproducibleLayerBuilder.java:47)
	at com.google.cloud.tools.jib.image.ReproducibleLayerBuilder.build(ReproducibleLayerBuilder.java:118)
	at com.google.cloud.tools.jib.builder.steps.BuildAndCacheApplicationLayerStep.call(BuildAndCacheApplicationLayerStep.java:107)
	at com.google.cloud.tools.jib.builder.steps.BuildAndCacheApplicationLayerStep.call(BuildAndCacheApplicationLayerStep.java:37)
	... 6 more
```

The code in question:
https://github.com/GoogleContainerTools/jib/blob/a692b8b0cde1f982f6e28309dd08ca80460950e2/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java#L73-L76

The layer configurations had been built with `LayerConfiguration.Builder.addEntryRecursive()`

In this case, `tarEntryPath.getName()` is `/` (the root directory).  `namePath` is a `sun.nio.fs.UnixPath` on "/" with filesystem of type `sun.nio.fs.MacOSXFileSystem`.  But `namePath.getParent()` returns `null`, which adheres to the contract of `Path#getParent()`: _Returns the parent path, or `null` if this path does not have a parent._

So an obvious fix is to change this guard to
```java
 if (namePath.getParent() == null || namePath.getParent() != namePath.getRoot()) {
```

But should we be adding a tar entry for the root directory?

loosebazooka:
fixes #2195

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
@@ -68,10 +68,16 @@ public class ReproducibleLayerBuilder {
         return;
       }
 
+      Path namePath = Paths.get(tarArchiveEntry.getName());
+
+      // Don't add entries for "/"
+      if (namePath.equals(Paths.get("/"))) {
+        return;
+      }
+
       // Adds all directories along extraction paths to explicitly set permissions for those
       // directories.
-      Path namePath = Paths.get(tarArchiveEntry.getName());
-      if (namePath.getParent() != namePath.getRoot()) {
+      if (namePath.getParent() != null) {
         TarArchiveEntry dir = new TarArchiveEntry(DIRECTORY_FILE, namePath.getParent().toString());
         dir.setModTime(LayerConfiguration.DEFAULT_MODIFICATION_TIME.toEpochMilli());
         add(dir);
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
@@ -92,6 +92,10 @@ public class ReproducibleLayerBuilderTest {
     Assert.assertEquals(TarArchiveEntry.DEFAULT_DIR_MODE, extractionPathEntry.getMode());
   }
 
+  private static void verifyNoMoreEntries(TarArchiveInputStream tarArchiveInputStream) throws IOException {
+    Assert.assertNull(tarArchiveInputStream.getNextTarEntry());
+  }
+
   private static LayerEntry defaultLayerEntry(Path source, AbsoluteUnixPath destination) {
     return new LayerEntry(
         source,
@@ -150,6 +154,7 @@ public class ReproducibleLayerBuilderTest {
           Paths.get(Resources.getResource("core/layer/foo").toURI()));
       verifyNextTarArchiveEntryIsDirectory(tarArchiveInputStream, "extract/here/banana/");
       verifyNextTarArchiveEntry(tarArchiveInputStream, "extract/here/banana/blobA", blobA);
+      verifyNoMoreEntries(tarArchiveInputStream);
     }
   }
 
@@ -359,6 +364,38 @@ public class ReproducibleLayerBuilderTest {
     }
   }
 
+  @Test
+  public void testBuild_rootEntryNotAddedDoesntThrowNpe() throws IOException {
+    Path testRoot = temporaryFolder.getRoot().toPath();
+    Path file = Files.createFile(testRoot.resolve("test.file"));
+
+    Blob blob =
+            new ReproducibleLayerBuilder(
+                    ImmutableList.of(
+                            new LayerEntry(
+                                    testRoot,
+                                    AbsoluteUnixPath.fromPath(Paths.get("/")),
+                                    FilePermissions.DEFAULT_FOLDER_PERMISSIONS,
+                                    LayerConfiguration.DEFAULT_MODIFICATION_TIME),
+                            new LayerEntry(
+                                    file,
+                                    AbsoluteUnixPath.fromPath(Paths.get("/test.file")),
+                                    FilePermissions.DEFAULT_FILE_PERMISSIONS,
+                                    LayerConfiguration.DEFAULT_MODIFICATION_TIME)))
+            .build();
+
+
+    Path tarFile = temporaryFolder.newFile().toPath();
+    try (OutputStream out = new BufferedOutputStream(Files.newOutputStream(tarFile))) {
+      blob.writeTo(out);
+    }
+
+    try (TarArchiveInputStream in = new TarArchiveInputStream(Files.newInputStream(tarFile))) {
+      verifyNextTarArchiveEntry(in, "test.file", file);
+      verifyNoMoreEntries(in);
+    }
+  }
+
   private Path createFile(Path root, String filename, String content, long modificationTime)
       throws IOException {
 
@@ -370,4 +407,5 @@ public class ReproducibleLayerBuilderTest {
     Files.setLastModifiedTime(newFile, FileTime.fromMillis(modificationTime));
     return newFile;
   }
+
 }


jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -68,10 +68,16 @@ private void add(TarArchiveEntry tarArchiveEntry) {
...
+      // Don't add entries for "/"
+      if (namePath.equals(Paths.get("/"))) {
+        return;
+      }
+
       // Adds all directories along extraction paths to explicitly set permissions for those
       // directories.
-      Path namePath = Paths.get(tarArchiveEntry.getName());
-      if (namePath.getParent() != namePath.getRoot()) {
+      if (namePath.getParent() != null) {
```
loosebazooka: It turns out this was only ever true if `getParent` and `getRoot` both returned null.
1. This was doing an object based equality anyway
2. `getRoot` is ALWAYS null, because these paths have their leading slash stripped by the TarArchiveEntry
3. we handle the only possible non-null `getRoot` entry above at: https://github.com/GoogleContainerTools/jib/pull/2300/files#diff-9f3ce3d4c2254dbe95c91624e7b434d9R74

loosebazooka: This is contingent on us never creating TarArchiveEntries that are `preserveAbsolutePath == true`.

I think our previous code would work the same way anyway.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -68,10 +68,16 @@ private void add(TarArchiveEntry tarArchiveEntry) {
         return;
       }
 
+      Path namePath = Paths.get(tarArchiveEntry.getName());
+
+      // Don't add entries for "/"
+      if (namePath.equals(Paths.get("/"))) {
```
chanseokoh: You said below that `TarArachiveEntry` strips a leading slash. For a root, does `tarArchiveEntry.getName()` returns `"/"` instead of `""` (empty string)? If it returns `""`, the check will never get matched, because `Paths.get("")` does not equal to `Paths.get("/")`.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
chanseokoh: > This was doing an object based equality anyway

This is bad. However, I got a deja-vu having been suspicious about this check and thought perhaps this code actually works. Indeed, the following code did return true on my machine.
```java
    Path namePath = Paths.get("/foo");
    System.out.println(namePath.getRoot() == namePath.getParent());
```
I can guess why, but we should have really used `equals()`.

> * `getRoot` is ALWAYS null, because these paths have their leading slash stripped by the TarArchiveEntry
> * we handle the only possible non-null `getRoot` entry above at:

Isn't this claim in conflict with the check above (`namePath.equals(Paths.get("/")`)? When the check is true, `namePath.getRoot()` will return `/`.

Anyways, I guess comparing with `getRoot()` is general and future proof regardless of `preserveAbsolutePath`?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
loosebazooka: except for "/", there's special handling for "/", weirdly.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
loosebazooka: Well, can't call equals because of null pointer exception. This whole section is super suspecious, and I'm actually not too confident about this part of code. I think I'd rather push to 2.2.0 milestone than submit here.

loosebazooka: The claim is that when TarArchiveEntry returns name, it's stripped of any "/" prefixed to the path, so the path is never absolute, so root cannot be determined.

So if you do
```
TarArchiveEntry tae = new TarArchiveEntry(new File("some/file"), "/foo");
System.out.println(tae.getName())
```

you will get `foo` and not `/foo`

unless you have explicitly done
```
TarArchiveEntry tae = new TarArchiveEntry("/foo", true)
System.out.println(tae.getName())
```

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
chanseokoh: Ah, interesting. Worth making a code comment.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
chanseokoh: > The claim is that when TarArchiveEntry returns name, it's stripped of any "/" prefixed to the path

This is true. What I'm saying is that this is in conflict with the following check you added:
```java
      // Don't add entries for "/"
      if (namePath.equals(Paths.get("/"))) {
```
Because `new TarArchiveEntry("/").getName()` is an empty string (i.e, `namePath` is `Paths.get("")` which is differetn from `Paths.get("/")`), it seems to be that the check above is meaningless.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
chanseokoh: > except for "/", there's special handling for "/", weirdly.

Actually, I looked into the `TarArchiveEntry` code (`normalizeFileName`), and I think it strips the leading slash for the root too. I don't see any special handling for the root entry. And in my testing, it does return an empty string.

```java
        // No absolute pathnames
        // Windows (and Posix?) paths can start with "\\NetworkDrive\",
        // so we loop on starting /'s.
        while (!preserveAbsolutePath && fileName.startsWith("/")) {
            fileName = fileName.substring(1);
        }
        return fileName;
```

loosebazooka: I was looking more at: TarArchiveEntry.java:369 which occurs after normalization happens TarArchiveEntry.java:361.

For me, following the debugger puts it `/` when `tarArchiveEntry.getName()` is called.

loosebazooka: Also I think I see why we're getting different results, it looks like 
```
TarAchiveEntry("/").getName() != TarArchiveEntry(new File("some/file"), "/").getName()
```

chanseokoh: Ah, I see the difference. If you use `TarArchiveEntry(File)`, `getName()` returns `"/"`. However, when you use `TarArchiveEntry(String)`, it returns `""`.

loosebazooka: Anyway, there's obviously a lot to unpack here, lets just hold off on this PR completely till after we do a `2.1.0`.

tnaroska:
I also ran into https://github.com/GoogleContainerTools/jib/issues/2195
Any chance this work can be picked up again?

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
tnaroska: It seems like the only unclear part is whether to check for "/" or "". Arguably both would represent entries that shouldn't go into the archive.
Why not check for both and get this forward?
```
      // Don't add entries for "/"
      if (tarArchiveEntry.getName().equals("")
       || tarArchiveEntry.getName().equals("/")) {
        return;
      }
```
Also, no need to convert Strings to Path for this part. Only adds another indirection when we already know exactly which strings we're looking for.

mpeddada1: I thought about this for a bit. I think `if (tarArchiveEntry.getName().equals("/")` could work. I tried this condition out with the test we have in the PR and it seems to give us the desired outcome. However, the `tarArchiveEntry.getName().equals("")` check may not be needed. Testing this out through the debugger shows that the tarArchiveEntry is created at this [line](https://github.com/GoogleContainerTools/jib/blob/71e3977d5b1e81b860b29c3c62c8ea807a1d2d28/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java#L152) which results in "/" when .getName() is called for the extraction path of "/".

mpeddada1:
@tnaroska thanks for your interest! It looks like there were still some risks to making the change when we looking into the implementation. We want to be careful about inadvertently breaking current behavior. We still need to gather some information before this can be picked up again.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
chanseokoh: I don't remember all the details about the context of this problem, but what I said back then may be relevant: https://github.com/GoogleContainerTools/jib/pull/2300/files#r381532140

> Ah, I see the difference. If you use `TarArchiveEntry(File)`, `getName()` returns `"/"`. However, when you use `TarArchiveEntry(String)`, it returns `""`.

mpeddada1: Thanks @chanseokoh! It's interesting that they behave differently. That's definitely helpful here.

PR #2310, opened 2020-02-25
Update version strings for 2.1.0

TadCordle:


chanseokoh:
While I was testing some Kotlin code, I realized we missed this: https://github.com/GoogleContainerTools/jib/blob/master/examples/ktor/build.gradle.kts#L4

PR #2320, opened 2020-03-09
Allow more timezone formats

### LINKED ISSUE #2319: time parser should handle all time zone designators
https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators

https://stackoverflow.com/questions/46487403/java-8-date-and-time-parse-iso-8601-string-without-colon-in-offset

loosebazooka:
fixes #2319

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -6,13 +6,15 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 ### Changed
-- `jib.container.creationTime` now accepts more timezone formats:`Z`,`+HHmm`. ([#2320](https://github.com/GoogleContainerTools/jib/issues/2260))
+
+- `jib.container.creationTime` now accepts more timezone formats:`+HHmm`. This allows for easier configuration of creationTime by external systems. ([#2320](https://github.com/GoogleContainerTools/jib/issues/2320))
 
 ### Fixed
 
 ## 2.1.0
 
 ### Added
+
 - Additionally reads credentials from `~/.docker/.dockerconfigjson` and legacy Docker config (`~/.docker/.dockercfg`). Also searches for `$HOME/.docker/*` (in addition to current `System.get("user.home")/.docker/*`). This may help retrieve credentials, for example, on Kubernetes. ([#2260](https://github.com/GoogleContainerTools/jib/issues/2260))
 - New skaffold configuration options that modify how jib's build config is presented to skaffold ([#2292](https://github.com/GoogleContainerTools/jib/pull/2292)):
     - `jib.skaffold.watch.buildIncludes`: a list of build files to watch
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 ### Changed
+- `jib.container.creationTime` now accepts more timezone formats:`Z`,`+HHmm`. ([#2320](https://github.com/GoogleContainerTools/jib/issues/2260))
 
 ### Fixed
 
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -733,13 +733,23 @@ public class PluginConfigurationProcessor {
         default:
           DateTimeFormatter formatter =
               new DateTimeFormatterBuilder()
+                  .parseCaseInsensitive()
                   .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+                  // iso 8601 strict
                   .optionalStart()
                   .appendOffset("+HH:MM", "+00:00")
+                  .optionalStart()
+                  .appendLiteral('[')
+                  .parseCaseSensitive()
+                  .appendZoneRegionId()
+                  .appendLiteral(']')
                   .optionalEnd()
+                  .optionalEnd()
+                  // with no ":" in tz
                   .optionalStart()
                   .appendOffset("+HHMM", "+0000")
                   .optionalEnd()
+                  // 2 digit or Z (zero)
                   .optionalStart()
                   .appendOffset("+HH", "Z")
                   .optionalEnd()
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -733,25 +733,10 @@ public class PluginConfigurationProcessor {
         default:
           DateTimeFormatter formatter =
               new DateTimeFormatterBuilder()
-                  .parseCaseInsensitive()
-                  .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
-                  // iso 8601 strict
+                  .append(DateTimeFormatter.ISO_DATE_TIME) // parses isoStrict
+                  // add ability to parse with no ":" in tz
                   .optionalStart()
-                  .appendOffset("+HH:MM", "+00:00")
-                  .optionalStart()
-                  .appendLiteral('[')
-                  .parseCaseSensitive()
-                  .appendZoneRegionId()
-                  .appendLiteral(']')
-                  .optionalEnd()
-                  .optionalEnd()
-                  // with no ":" in tz
-                  .optionalStart()
-                  .appendOffset("+HHMM", "+0000")
-                  .optionalEnd()
-                  // 2 digit or Z (zero)
-                  .optionalStart()
-                  .appendOffset("+HH", "Z")
+                  .appendOffset("+HHmm", "+0000")
                   .optionalEnd()
                   .toFormatter();
           return formatter.parse(configuredCreationTime, Instant::from);
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -44,6 +44,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
 import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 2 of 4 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -730,9 +731,21 @@ static Instant getCreationTime(String configuredCreationTime, ProjectProperties
...
+                  .appendOffset("+HH:MM", "+00:00")
+                  .optionalEnd()
+                  .optionalStart()
+                  .appendOffset("+HHMM", "+0000")
+                  .optionalEnd()
+                  .optionalStart()
+                  .appendOffset("+HH", "Z")
+                  .optionalEnd()
+                  .toFormatter();
+          return formatter.parse(configuredCreationTime, Instant::from);
```
chanseokoh: Should we support `-HHMM` and `-HH:MM` as well? And I am not sure if `+` in `+HH` is needed. Or perhaps I misunderstood the meaning of `+`.

And should we support [+hh:mm:ss and -hh:mm:ss](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#getId--) as well? I don't expect anyone to set a second for the offset, but a machine may generate it for completeness.

It's also worth making a CHANGELOG entry that Jib no longer accepts the [non-ISO8601 zone format](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_DATE_TIME) (e.g., `[Europe/Paris]`)? Or, should we just use this new formatter only when the built-in formatter fails?

chanseokoh: That said, it's worth adding tests using `-`.

loosebazooka: Yeah I tested them to see it work, but I'll add them to the tests.

loosebazooka: hrmm... maybe I can accept the zone format too. I'll try not to break anything.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -730,9 +731,31 @@ static Instant getCreationTime(String configuredCreationTime, ProjectProperties
           return Instant.now();
 
         default:
-          return DateTimeFormatter.ISO_DATE_TIME.parse(configuredCreationTime, Instant::from);
+          DateTimeFormatter formatter =
+              new DateTimeFormatterBuilder()
+                  .parseCaseInsensitive()
+                  .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+                  // iso 8601 strict
```
loosebazooka: https://github.com/frohoff/jdk8u-dev-jdk/blob/master/src/share/classes/java/time/format/DateTimeFormatter.java#L992

chanseokoh: Aha. Then can't we start with `ISO_DATE_TIME`? If I'm not mistaken, we only need to add `.appendOffset("+HHMM", "+0000")` for the case without a colon. So, can't we do something like this?

```java
new DateTimeFormatterBuilder()
                .append(ISO_DATE_TIME)
                .optionalStart()
                .appendOffset("+HHMM", "+0000")
                .optionalStart()
                .toFormatter(...)
```

If that doesn't work, at least, I think we can base on `.appendOffsetId()` and add only `.appendOffset("+HHMM", "+0000")`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -730,9 +731,31 @@ static Instant getCreationTime(String configuredCreationTime, ProjectProperties
           return Instant.now();
 
         default:
-          return DateTimeFormatter.ISO_DATE_TIME.parse(configuredCreationTime, Instant::from);
+          DateTimeFormatter formatter =
+              new DateTimeFormatterBuilder()
+                  .parseCaseInsensitive()
```
chanseokoh: Is this intended? The JDK code you linked doesn't start with this.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -730,9 +731,31 @@ static Instant getCreationTime(String configuredCreationTime, ProjectProperties
...
+                  .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+                  // iso 8601 strict
+                  .optionalStart()
+                  .appendOffset("+HH:MM", "+00:00")
+                  .optionalStart()
+                  .appendLiteral('[')
+                  .parseCaseSensitive()
+                  .appendZoneRegionId()
+                  .appendLiteral(']')
+                  .optionalEnd()
```
chanseokoh: Judging from this [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_DATE_TIME), shouldn't this be after `.appendOffset("+HH", "Z")`?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -908,11 +909,48 @@ public void testGetCreationTime_useCurrentTimestamp() throws InvalidCreationTime
...
+  @Test
+  public void testGetCreationTime_isoDateTimeValueTimeZoneRegionOnlyAllowedForMostStrict8601Mode() {
+    List<String> invalidTimeStamps =
+        ImmutableList.of(
+            "2011-12-03T01:15:30+0900[Asia/Tokyo]", "2011-12-03T01:15:30+09[Asia/Tokyo]");
+    for (String timeString : invalidTimeStamps) {
+      try {
+        PluginConfigurationProcessor.getCreationTime(timeString, projectProperties);
+        Assert.fail(
+            "creationTime should fail if region specified when zone not in HH:MM mode - "
```
chanseokoh: Hmm... is this true with the built-in `ISO_DATE_TIME`?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -908,11 +909,48 @@ public void testGetCreationTime_useCurrentTimestamp() throws InvalidCreationTime
...
+        // pass
+      }
+    }
+  }
+
+  @Test
+  public void testGetCreationTime_isoDateTimeValueRequiresTimeZone() {
+    try {
+      PluginConfigurationProcessor.getCreationTime("2011-12-03T01:15:30", projectProperties);
+      Assert.fail("creationTime should fail if timezone not specified");
```
chanseokoh: Looks like the built-in [`ISO_DATE_TIME`](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_DATE_TIME) allows this? We go against it?

> The ISO-like date-time formatter that formats or parses a date-time with the offset and zone if available, such as `'2011-12-03T10:15:30'`,

loosebazooka: Yeah I think I misunderstood what was going on. The real issue is that
`Instance.from(DateTimeFormatter.ISO_DATE_TIME("2011-12-03T01:15:30"))` throws an exception

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
loosebazooka: So ISO_DATE_TIME requires you use `HH:MM`, I don't think we have a reason to add in the code to parse the regionId, and this test just verifies that behavior

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
loosebazooka: Yeah, I was just playing around with things, mustve left it in there.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
loosebazooka: I was trying to recreate ISO_DATE_TIME here, but this whole section seems unnecessary.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
loosebazooka: It appears to generate a valid TemporalAccessor intermediary, but that value cannot be converted to Instant.

chanseokoh: Aha, so `Instant.from` fails because the parsed time has no timezone info?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
loosebazooka: Yeah I think you're right,  I think we want to use `+HHmm` which covers `+09`, and `+0900`

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
loosebazooka: right

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
loosebazooka: So it seems like the formatter will accept seconds, but just ignore them?

chanseokoh: It's not completely ignored. The following code prints `1322842530` and `1322842529` respecively.

```java
    System.out.println(Instant.from(DateTimeFormatter.ISO_DATE_TIME.parse("2011-12-03T01:15:30+09:00:00")).getEpochSecond());
    System.out.println(Instant.from(DateTimeFormatter.ISO_DATE_TIME.parse("2011-12-03T01:15:30+09:00:01")).getEpochSecond());
```

However, I don't really think anyone will really use seconds (other than `00`) as part of timezone.

loosebazooka: right because of the ISO_DATE_TIME formatter using `HH:MM:ss`, but for `HHmm` seconds is not allow (I thought it was ignored, but that is untrue). In any case, I can't think of a timezone that is offset by seconds.

loosebazooka:
I will update the changelog too

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 ### Changed
+- `jib.container.creationTime` now accepts more timezone formats:`Z`,`+HHmm`. ([#2320](https://github.com/GoogleContainerTools/jib/issues/2260))
```
chanseokoh: I believe `Z` was already accepted. So, "accepts a timezone format without colon: `+HHmm`"? And if this improves usability with git, worth nothing that? Like, "This helps use git commit timestamp from `git ...` or some git timestamp plugin"?

chanseokoh: Oh, the issue number in `issues/2260` is wrong.

chanseokoh: nit: also add a space after `### Changed`. I saw we have the same problem in line 15, so can you add a space after `### Added` there too?

PR #2323, opened 2020-03-11
Minor fixes and cleanups

chanseokoh:


jib-maven-plugin/.gitignore
```diff
@@ -9,4 +9,3 @@ out
 /.settings
 /.classpath
 /.project
-log.txt
```
TadCordle: Why remove this?

chanseokoh: Is this file generated by something? Otherwise, it just thought it isn't common to have this in `.gitignore`.

chanseokoh: Merging in this, but anyone feel free to restore this if needed.

PR #2312, opened 2020-02-26
Container Build Plan specification

chanseokoh:
\> [Preview](https://github.com/GoogleContainerTools/jib/blob/8ef257e49fad8509b9f0d816b48b3179c4a9f0d9/proposals/container-build-plan-spec.md) <

PR #2324, opened 2020-03-12
Set up jib-build-plan sub-project

chanseokoh:
Set up a new subproject `jib-container-build-plan`. It will host Java API (classes) for the [Container Build Plan spec](https://github.com/GoogleContainerTools/jib/pull/2312).

## Brief heads-up for what it will host

Its classes will include (basically) `AbsoluteUnixPath`, `RelativeUnixPath`, `Port`, `FilePermissions`, `LayerEntry`, `LayerConfiguration`, `ImageFormat`, etc. They will be _duplicated_ (after renaming if desired) from `jib-core` into a different package `com.google.cloud.tools.jib.buildplan`.

Since these are public API in `jib-core` that we cannot remove, this means we will have to maintain duplicate code between `jib-core` and `jib-container-build-plan`, hopefully for a short period of time. I imagine that once `jib-core` is adjusted to depend on `jib-container-build-plan`, these duplicated classes in `jib-core` may become a thin proxy to the corresponding classes in `jib-container-build-plan`. For example, `AbsoluteUnixPath` in `jib-core` could merely delegating calls to `AbsoluteUnixPath` in `jib-container-build-plan`.

I wish I could just move these classes and remove them from `jib-core`, but I think we want some stability in the `jib-core` API. Let me know if you have better ideas.


chanseokoh:
To get the idea of actual classes to be added, see this [comparison](https://github.com/GoogleContainerTools/jib/compare/build-plan-subproject...plugin-extension-1?expand=1).  Still working on it, but it's close to completion.

jib-build-plan/build.gradle
```diff
@@ -0,0 +1,45 @@
+plugins {
+  id 'net.researchgate.release'
+  id 'maven-publish'
+  id 'eclipse'
+}
+
+dependencies {
+  implementation "com.google.guava:guava:${dependencyVersions.GUAVA}"
```
loosebazooka: It makes sense that we need this, but I would think a library like this should almost have zero deps to minimize diamond dep issues.

chanseokoh: We'll need to do some more work without this, but I agree. I'll move this to `testImplementation`.

PR #2327, opened 2020-03-13
Remove Guava dependency from certain core API classes

chanseokoh:
I plan to copy these classes into the jib-build-plan project, where we won't depend on Guava.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerConfiguration.java
@@ -282,7 +282,7 @@ public class LayerConfiguration {
    * Use {@link #builder} to instantiate.
    *
    * @param name an optional name for the layer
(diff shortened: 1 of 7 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/LayerConfiguration.java
```diff
@@ -289,6 +304,10 @@ public String getName() {
    * @return the list of layer entries
    */
   public ImmutableList<LayerEntry> getLayerEntries() {
-    return layerEntries;
+    return ImmutableList.copyOf(entries);
```
loosebazooka: This is still guava no?

chanseokoh: True. We'll keep the jib-core API unchanged. When I duplicate this into jib-build-plan, the method name will change to `getEntries()` and it will return the plain `List`.

chanseokoh:
Thanks @loosebazooka!

PR #2332, opened 2020-03-16
Fix sourceProject helper

loosebazooka:
- force evaluation of dependencyProject
- actually do the right kind of removal (perhaps this changed with gradle update?)
  this will be caught by forcing evaluation now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -332,10 +332,10 @@ subprojects {
   project.ext.sourceProjects = []
   // sourceProject(Project) accepts a project and adds it as a dependency in a special manner:
   // 1. force evaluation of the project first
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -331,6 +331,7 @@ subprojects {
   // to keep track of all source projects
   project.ext.sourceProjects = []
   // sourceProject(Project) accepts a project and adds it as a dependency in a special manner:
+  // 1. force evaluation of the project first
   // 1. add the project classes as "compileOnly" and make it available to tests in "testImplementation"
```
loosebazooka: ugh numbering, I'll fix this.

PR #2337, opened 2020-03-17
Migrate deprecated LayerConfiguration and LayerEntry usage

chanseokoh:
Continuing the work for the Jib extension framework.

Refactored our codebase to use the new classes in the "Build Plan" project.

I'll add CHANGELOG in a later PR.


chanseokoh:
Now this one is ready for review.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -167,36 +170,82 @@ public JibContainerBuilder addLayer(List<Path> files, String pathInContainer) th
...
    * @return this
    */
+  @Deprecated
   public JibContainerBuilder setLayers(List<LayerConfiguration> layerConfigurations) {
-    this.layerConfigurations = new ArrayList<>(layerConfigurations);
+    this.layerConfigurations =
+        layerConfigurations
+            .stream()
+            .map(LayerConfiguration::toFileEntriesLayer)
+            .collect(Collectors.toList());
```
loosebazooka: should this call `setFileEntriesLayers` with the final result of the `stream.collect`?

chanseokoh: Makes sense. I just wanted to avoid creating an additional `ArrayList` in `setFileEntriesLayers`, but that's really negligible.

Also updated `addLayer()` to call `addFileEntriesLayer()` similarly.

chanseokoh:
I'll add CHANGELOG shortly.

PR #2338, opened 2020-03-17
Add ContainerBuildPlan as Build Plan API interface

chanseokoh:
The class works as the forefront Java interface for the proposed [Build Plan specification](https://github.com/GoogleContainerTools/jib/blob/2b2fe82ad0552ba3ce1de308a0fda24aab684365/proposals/container-build-plan-spec.md).

This setters share some similarities with those of `JibContainerBuilder`. Most Javadocs are copied over after minor tweaks.

### DIFF
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
@@ -71,7 +71,7 @@ public class ContainerBuildPlan {
      * value is ignored and the architecture of the built image follows that of the base image. The
      * default is {@code amd64}.
      *
-     * @param architectureHint image to use as a base
+     * @param architectureHint architecture value to select a base image in case of a manifest list
      * @return this
      */
     public Builder setArchitectureHint(String architectureHint) {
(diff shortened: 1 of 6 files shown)


chanseokoh:
Ready for review.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
```diff
@@ -0,0 +1,495 @@
...
+    private Map<String, String> environment = new HashMap<>();
+    private Map<String, String> labels = new HashMap<>();
+    private Set<AbsoluteUnixPath> volumes = new HashSet<>();
+    private Set<Port> exposedPorts = new HashSet<>();
+    @Nullable private String user;
+    @Nullable private AbsoluteUnixPath workingDirectory;
+    @Nullable private List<String> entrypoint;
+    @Nullable private List<String> cmd;
+
+    private List<LayerObject> layers = new ArrayList<>();
```
loosebazooka: Not sure I like the name `LayerObject`, but that's a different discussion.

chanseokoh: I don't like it either. We already have a non-public `Layer` interface in jib-core, so I just wanted to avoid name clash.

loosebazooka: Okay, well this shouldn't prevent this PR from going in.

PR #2341, opened 2020-03-18
For config generation use create and move of temp file

### LINKED ISSUE #2294: Warning from empty global Jib config

loosebazooka:
I think this is the minimum fix, but I went back and forth on creating a config class that does the config work potentially making it easier to test? (I think?) I'll post that as a followup but this should do:
fixes #2294

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
@@ -32,9 +32,11 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.AtomicMoveNotSupportedException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.format.DateTimeParseException;
@@ -46,6 +48,8 @@ import java.util.function.Consumer;
 
 /** Checks if Jib is up-to-date. */
 public class UpdateChecker {
+
+  private static final String CONFIG_FILENAME = "config.json";
 
   /** JSON template for the configuration file used to enable/disable update checks. */
   @VisibleForTesting
@@ -97,7 +101,7 @@ public class UpdateChecker {
       return Optional.empty();
     }
 
-    Path configFile = configDir.resolve("config.json");
+    Path configFile = configDir.resolve(CONFIG_FILENAME);
     Path lastUpdateCheck = configDir.resolve("lastUpdateCheck");
 
     try {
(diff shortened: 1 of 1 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -123,11 +127,24 @@ void setDisableUpdateCheck(boolean disableUpdateCheck) {
         // Generate config file if it doesn't exist
         ConfigJsonTemplate config = new ConfigJsonTemplate();
         Files.createDirectories(configDir);
-        try (OutputStream outputStream = Files.newOutputStream(configFile)) {
+        Path temp = configDir.resolve(CONFIG_FILENAME + ".tmp");
```
chanseokoh: `temp` --> `tempConfigFile`

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -123,11 +127,24 @@ void setDisableUpdateCheck(boolean disableUpdateCheck) {
...
+          // Attempts an atomic move first, and falls back to non-atomic if the file system does not
+          // support atomic moves.
+          try {
+            Files.move(
+                temp,
+                configFile,
+                StandardCopyOption.ATOMIC_MOVE,
+                StandardCopyOption.REPLACE_EXISTING);
+          } catch (AtomicMoveNotSupportedException ignored) {
+            Files.move(temp, configFile, StandardCopyOption.REPLACE_EXISTING);
```
chanseokoh: I think we don't need to do this manually. I believe `Files.move` will do this fallback.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -123,11 +127,24 @@ void setDisableUpdateCheck(boolean disableUpdateCheck) {
...
+          }
         } catch (IOException ex) {
           // If attempt to generate new config file failed, delete so we can try again next time
           log.accept(LogEvent.debug("Failed to generate global Jib config; " + ex.getMessage()));
+          Files.deleteIfExists(temp);
           Files.deleteIfExists(configFile);
```
chanseokoh: There's one problem with this. If `Files.move` fails, the existing and working config file will be deleted. I think we can just remove `Files.deleteIfExists()` if we are moving the temp file rather than directly replacing the original file. No need to touch the exist one ever. And it's not a big deal not to delete the temp file. It will be extremely rare that moving will fail, and if that happens, the presence of a `.tmp` file may give more clues, I guess?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
loosebazooka: We do this in a few places..? CacheStorageWriter mostly? but that could be an artifact from before we knew what was happening? https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java#L140

chanseokoh: Ah, sorry. Now you said it. I think this makes sense. `AtomicMoveNotSupportedException` is expected because we use the `ATOMIC_MOVE` option.

So perhaps doing `Files.move` without `ATOMIC_MOVE` is sufficient. But honestly I don't have confidence.

chanseokoh: Let's keep the current code? At least it seems safe.

loosebazooka: Which current code? This or the code before this PR?

chanseokoh: Sorry, the PR with two `Files.move()` that we also do in other places.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
loosebazooka: hrmm.. so I removed this for the regular config file, but I left in the temp file removal, maybe I should leave that in?

PR #2345, opened 2020-03-19
Add glob support for extra file permissions

### LINKED ISSUE #1200: Support globs in file permission configuration
[Original comment](https://github.com/GoogleContainerTools/jib/pull/1195#pullrequestreview-168885517)

It may be useful to set file permissions in bulk, like so:

```groovy
jib.extraDirectories.permissions = ['/*.txt':'755', '/**.sh':'555']
```

```xml
<permissions>
  <permission>
    <file>/*.txt</file>
    <mode>755</mode>
  </permission>
  <permission>
    <file>/**.sh</file>
    <mode>555</mode>
  </permission>
</permissions>
```

TadCordle:
Fixes #1200.

Can now use glob patterns when setting extra files permissions. Permissions configured with explicit paths will take priority over globs (e.g. if both `'/a/b/file': '777'` and `'**/file': '444'` are configured, `/a/b/file` will have `777` permissions, and all other files named `file` will have `444` permissions).

If a file matches multiple globs (e.g. if there's a file `/a/b/file`, and permissions are configured for both `/a/b/*` and `**/file`), ~~behavior is undefined at the moment~~ first configured match wins.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
@@ -25,8 +25,11 @@ import com.google.cloud.tools.jib.api.buildplan.FilePermissions;
 import com.google.cloud.tools.jib.api.buildplan.RelativeUnixPath;
 import com.google.cloud.tools.jib.filesystem.DirectoryWalker;
 import java.io.IOException;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.Map;
 import java.util.function.BiFunction;
@@ -36,7 +39,8 @@ import java.util.function.Predicate;
 public class JavaContainerBuilderHelper {
 
   /**
-   * Returns a {@link LayerConfiguration} for adding the extra directory to the container.
+   * Validates and returns a {@link LayerConfiguration} for adding the extra directory to the
+   * container.
    *
    * @param extraDirectory the source extra directory path
    * @param extraDirectoryPermissions map from path on container to file permissions
@@ -46,7 +50,7 @@ public class JavaContainerBuilderHelper {
    */
   public static LayerConfiguration extraDirectoryLayerConfiguration(
       Path extraDirectory,
-      Map<AbsoluteUnixPath, FilePermissions> extraDirectoryPermissions,
+      Map<String, FilePermissions> extraDirectoryPermissions,
       BiFunction<Path, AbsoluteUnixPath, Instant> modificationTimeProvider)
       throws IOException {
     LayerConfiguration.Builder builder =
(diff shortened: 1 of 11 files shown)


loosebazooka:
> However, if a file matches multiple globs (e.g. if there's a file /a/b/file, and permissions are configured for both /a/b/* and **/file), behavior is undefined at the moment. Maybe we should sort the entry set or something when searching for matches so the behavior is more predictable?

It should be deterministic, like last configured pattern wins.

TadCordle:
Actually, looks like Groovy maps are `LinkedHashMap`s, so behavior should already be predictable (first match wins). But if you think last match wins is better UX we can do that.

loosebazooka:
Oh actually I don't know, it was just a suggestion. I  was thinking we apply all rules we find to a file, but it seems like we go looking for matches and apply the first one we find -- which actually makes sense.

Should the pattern only match files? Can we accidentally end up making directories un-traversable?

TadCordle:
We still traverse over every file/directory, but some glob patterns could end up being completely skipped. e.g. this will set everything to 777:

```
extraDirectories {
  permissions = [
    '/**': '777',
    '/a/**': '444'
  ]
}
```

But I think you can avoid this by just configuring the more specific patterns first.

loosebazooka:
> And again, if you configure permissions on files/directories without using a glob, those will override any permissions set using globs.

so order doesn't matter for those?

TadCordle:
Correct.

TadCordle:
Actually, since we take the map from the gradle config and put it in a new map, we might not be maintaining order... will fix.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -58,10 +62,26 @@ public static FileEntriesLayer extraDirectoryLayerConfiguration(
...
-              FilePermissions permissions = extraDirectoryPermissions.get(pathOnContainer);
+              FilePermissions permissions =
+                  extraDirectoryPermissions.get(pathOnContainer.toString());
               if (permissions == null) {
+                // Check for matching globs
+                Path containerPath = Paths.get(pathOnContainer.toString());
+                for (Map.Entry<String, FilePermissions> entry :
+                    extraDirectoryPermissions.entrySet()) {
+                  PathMatcher pathMatcher =
+                      FileSystems.getDefault().getPathMatcher("glob:" + entry.getKey());
```
chanseokoh: This looks good. I just think we can instantiate `PathMatchers` before `new DirectoryWalker(...)`. That way, we don't have to instantiate them over and over for each file in `extraDirectory`. Then at least we won't instantiate them again within the same layer. This is a minor optimization, but I think it's worth, because

1. In practice, almost all the time, you will have some files whose permissions are not configured.
2. The optimization doesn't add new code but just amounts to moving the code outside `.walk()`.

We could even skip path strings that aren't glob, but I think this is unnecessary and rather adds more code.

WDYT?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -46,22 +52,43 @@
    */
   public static FileEntriesLayer extraDirectoryLayerConfiguration(
       Path extraDirectory,
-      Map<AbsoluteUnixPath, FilePermissions> extraDirectoryPermissions,
+      Map<String, FilePermissions> extraDirectoryPermissions,
       BiFunction<Path, AbsoluteUnixPath, Instant> modificationTimeProvider)
       throws IOException {
     FileEntriesLayer.Builder builder =
         FileEntriesLayer.builder().setName(LayerType.EXTRA_FILES.getName());
+    Map<String, PathMatcher> pathMatchers = new HashMap<>();
```
chanseokoh: Hmm... Isn't `Map<PathMatcher, Permissions>` more natural? Inside `walk()`, you iterate over this map instead of `extraDirectoryPermissions`.

TadCordle: Oh, yeah. Wasn't seeing it at the time but I get it now.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -36,7 +40,8 @@
 public class JavaContainerBuilderHelper {
 
   /**
-   * Returns a {@link FileEntriesLayer} for adding the extra directory to the container.
+   * Validates and returns a {@link FileEntriesLayer} for adding the extra directory to the
```
chanseokoh: Did we add something to "validate a layer"?

PR #2347, opened 2020-03-19
Convert LayerObject to interface

chanseokoh:


jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/LayerObject.java
```diff
@@ -27,12 +27,18 @@
   }
```
loosebazooka: I thought this would be an interface? Is that not true?

loosebazooka: I guess it's harder for deserialization huh?

chanseokoh: Good point. I thought this needed to be a class for JSON polymorphic type handling.But I found an [example using an interface](https://octoperf.com/blog/2018/02/01/polymorphism-with-jackson/), so I think interface works.

chanseokoh: > I guess it's harder for deserialization huh?

Doesn't seem so. A lot of examples use class.

PR #2350, opened 2020-03-20
Refactor out common test classes into new test project

loosebazooka:
part of #2150 

- create new test subproject (`jib-testing-common`)
- move common classes into new project (ideally this should be enough for all our cross project test infra sharing)
- checkstyle suppression for `jib-testing-common`
- NOT DONE: shared resources moved into `jib-testing-common`


loosebazooka:
oh boy, gotta fix these javadoc issues

jib-testing-common/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+}
+
+// disable javadoc on this project, it's only for testing
+tasks.javadoc.enabled = false
+
+dependencies {
+  // if you want to use classes that have dependencies in the main source
+  // sets of other projects, mark them compile only
+  compileOnly project(':jib-plugins-common')
+  compileOnly project(':jib-core')
```
chanseokoh: I don't know how this project will be used, but will source files in this project need using classes from `jib-plugins-common` and `jib-core`?

What's weird to me is that `jib-core` depends on `jib-testing-common` (albeit via `testImplementation`) but `jib-testing-common` depends on `jib-core`.

loosebazooka: `jib-core:test` can be seen as it's own project with a dependency on `jib-core`. Since the common testing infrastructure is bound to reference production classes, I think this is necessary. But if the question is: Is this PR necessary? I think it's convenient.

config/checkstyle/checkstyle-suppressions.xml
```diff
@@ -25,6 +25,7 @@
   <suppress files=".*[\\/]HelpfulSuggestions\.java" checks="MissingJavadocMethod"/>
 
   <!-- leniency for test files -->
+  <suppress files=".*[\\/]jib-testing-common[\\/].*\.java" checks ="MissingJavadocMethod"/>
```
chanseokoh: I wonder if we can start the pattern with `jib-testing-common[\\/]` without `."[\\/]`?

jib-gradle-plugin/build.gradle
```diff
@@ -23,8 +23,7 @@ dependencies {
   testImplementation "org.slf4j:slf4j-api:${dependencyVersions.SLF4J_API}"
   testImplementation "com.github.stefanbirkner:system-rules:${dependencyVersions.SYSTEM_RULES}"
 
-  testImplementation project(path:':jib-plugins-common', configuration:'tests')
-  integrationTestImplementation project(path:':jib-core', configuration:'integrationTests')
+  testImplementation project(path:':jib-testing-common')
```
chanseokoh: nit: for consistency, `testImplementation project(':jib-testing-common')` (without specifying the `path` argument name).

And does the order of `testImplementation` matter? In `jib-core`, it's the first entry.

jib-maven-plugin/build.gradle
```diff
@@ -35,8 +35,7 @@ dependencies {
   testImplementation 'org.slf4j:slf4j-api:1.7.30'
   testImplementation 'org.slf4j:slf4j-simple:1.7.30'
 
-  testImplementation project(path:':jib-plugins-common', configuration:'tests')
-  integrationTestImplementation project(path:':jib-core', configuration:'integrationTests')
+  testImplementation project(path:':jib-testing-common')
```
chanseokoh: ditto

loosebazooka:
> Should we move these too?
>
> HttpGetVerifier
> PlainHttpClient
> RequestWrapper

So these aren't shared by anyone, I guess if we want to eventually export them we can? But there's no need to now.

config/checkstyle/checkstyle-suppressions.xml
loosebazooka: I don't know if it matters, I guess this is a little more strict: `/jib-testing-common/` instead of `abc-jib-testing-common/`

PR #2351, opened 2020-03-20
Upgrade ASM and Error Prone

gliptak:


chanseokoh:
Thanks for the contribution. 2 things:

- The ASM library should be upgraded to support Java 13  & 14.
- Gradle 5.2.1 doesn't seem to support Java 14: https://github.com/gradle/gradle/issues/10248

gliptak:
@chanseokoh thanks for the pointers. I will update PR

chanseokoh:
Oh, and seems like Error Prone has [issues](https://github.com/google/error-prone/issues/1106#issuecomment-417952124) with Java 12+. Unclear if they are all fixed.

gliptak:
https://github.com/google/error-prone/pull/1107

chanseokoh:
Yeah, unfortunately, looks like we cannot add 12+ builds until they support them.

don-vip:
> Oh, and seems like Error Prone has [issues](https://github.com/google/error-prone/issues/1106#issuecomment-417952124) with Java 12+. Unclear if they are all fixed.

Nothing is fixed. I am struggling for 2 years now to submit pull requests that are ignored by error_prone developers, see https://github.com/google/error-prone/issues/1106

For JOSM we apply around 5/6 patches to make it work with recent versions of Java.

gliptak:
@chanseokoh gradle 5.x only works for 12-

this morphed into something else. Travis build is green

@don-vip maybe time to look for another tool ...

don-vip:
@gliptak for now we manage to update the tool with the patches. Maybe one day we'll drop it if Google continues to ignore recent Java. Are you working at Google, or just using their tools?

gliptak:
@don-vip just run into this while trying to bump ```Java``` version for ```jib```

chanseokoh:
@don-vip maybe try the mailing list? https://groups.google.com/g/error-prone-discuss

@loosebazooka are we okay to upgrade Gradle, or it's safer to stick to the current 5.2.1 to prevent accidental API usage?

loosebazooka:
I mean I would like to stick to 5.2.1 if we could. Which part of this PR requires gradle `5.6.4`? Is it the asm upgrade or the errorprone upgrade?

gliptak:
@chanseokoh I'm not following the "accidental API usage" concern

@loosebazooka the Gradle upgrade was in preparation for adding current JDKs

loosebazooka:
We try to stay on as low a gradle version as possible so we don't use API from newer versions of gradle -- If `XYZ` is only in gradle 5.6, we break all users who are on a version lower than 5.6. We have some testing for this, but it's not super comprehensive.

So staying at `5.2.1` keeps us compatible with more users.  We generally try to give users a year for moving up gradle version.  However there is nothing keeping the user from using `5.6.X` in their build -- that should be fully compatible with anything we use from `5.2.1`

gliptak:
Isn't the point of https://docs.gradle.org/current/userguide/gradle_wrapper.html that the pinned Gradle version is used (and automagically downloaded if needed) regardless of what Gradle (if any) is already installed by the user?

loosebazooka:
Ah so I think you're misunderstanding here. When building a gradle plugin you build with some version of the gradle api (in jib's case 5.2.1), however the gradle API is a provided dependency as is not packaged with the project, it's whatever is available at runtime -- so as long as user who has `jib` applied on their project is using gradle 5.2.1 or *higher*, it will work.

Pinning the version on the jib build creates no inherent restrictions on the version of gradle the user uses to build their project except for API compatibility and the code we have here: https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java#L85

So if the user needs 5.6.2 to build java 12 or 13, that's fine, we may not need to make *this* build use 5.6.2 unless jib introduces some explicit support for java 12 or 13.

gliptak:
@loosebazooka thank you for the explanation. Remove Gradle bump from PR

chanseokoh:
@gliptak thanks for check these! Merging.

PR #2353, opened 2020-03-23
Build plan adapter methods for JibContainerBuilder

chanseokoh:
Adds internal methods for conversion between `JibContainerBuilder` <--> `ContainerBuildPlan`. I can't think of other ways to enable this conversion while not adding public methods to `JibContainerBuilder`.

As we currently build a `JibContainerBuilder` instance (which is the front-end public API) in our plugins code, the idea for the extension framework is basically like this:

1. Construct `JibContainerBuilder` (current behavior).
1. Convert `JibContainerBuilder` to `ContainerBuildPlan` and pass it to extensions.
1. Extensions modify the `ContainerBuildPlan`.
1. Convert `ContainerBuildPlan` back to `JibContainerBuilder`.
1. Plugins code proceeds with `JibContainerBuilder` to build an image (current behavior)

Another approach would be to revamp everything around `BuildContext` and `ContainerConfiguration` in jib-core. However, I think the amount of refactoring work to do so is prohibitively enormous, and this is not practical.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -17,6 +17,7 @@
 package com.google.cloud.tools.jib.api;
 
 import com.google.cloud.tools.jib.api.buildplan.AbsoluteUnixPath;
+import com.google.cloud.tools.jib.api.buildplan.ContainerBuildPlan;
 import com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer;
 import com.google.cloud.tools.jib.api.buildplan.FileEntry;
 import com.google.cloud.tools.jib.api.buildplan.ImageFormat;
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -542,12 +568,87 @@ public JibContainer containerize(Containerizer containerizer)
...
+  /**
+   * Internal method. API end users should not use it.
+   *
+   * <p>Converts to {@link ContainerBuildPlan}. Note that not all values that this class holds can
+   * be described by a build plan, such as {@link CredentialRetriever}s for {@link RegistryImage},
+   * {@link DockerClient} for {@link DockerDaemonImage}, and output path for {@link TarImage}.
+   *
+   * @return {@link ContainerBuildPlan}
+   */
+  public ContainerBuildPlan toContainerBuildPlan() {
```
loosebazooka: Could we mask it with an interface?

loosebazooka: And have an internal impl (kinda like how gradle has ProjectInternal)

loosebazooka: hrmm.. so maybe this is okay for now

PR #2355, opened 2020-03-24
Fix some warnings

TadCordle:


jib-core/src/test/java/com/google/cloud/tools/jib/http/TestWebServer.java
```diff
@@ -175,6 +175,7 @@ private synchronized String getNextResponse(int index) {
 
   // For use to ignore (i.e., accept and do nothing) a return value from ExecutionService.submit().
   // Without "consuming" the return value this way, Error Prone will complain to use it.
+  @SuppressWarnings("unused")
   private void ignoreReturn(Future<Void> future) {
```
chanseokoh: Can we just remove this method? The only point of this method was to suppress a warning that a return value is unused. If the method itself needs `SuppressWarnings`, the whole thing is only clutter.

TadCordle: Good point

PR #2357, opened 2020-03-24
Do not disable Google HTTP Client logging for custom logger in Gradle

### LINKED ISSUE #2356: Jib Gradle plugin disables Google HTTP Client logging
There have been many reports that Gradle users have difficulty generating Google HTTP Client logging following our [instructions](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-can-i-examine-network-traffic).

- https://github.com/GoogleContainerTools/jib/issues/1917#issuecomment-523553722
- https://github.com/GoogleContainerTools/jib/issues/2335#issuecomment-600282441
- https://github.com/GoogleContainerTools/jib/issues/2207#issuecomment-573124445

From my cursory experiment, this symptom has gotten worse with more recent Gradle. This is in line with the first report that downgrading to 4.10.3 worked.

The cause was our calling [`TaskCommon.disableHttpLogging()`](https://github.com/GoogleContainerTools/jib/blob/a02a88fae7bbf54a3296c9b402ec411065b6ff47/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java#L88).

Some interesting observations:
- Looks like `disableHttpLogging()` has a lasting effect on a running Gradle daemon. Once it is called, Gradle doesn't generate logs even if I remove `disableHttpLogging()`. Stopping daemons with `./gradlew --stop` and trying again works. This explains the observation in the first report that using `--no-daemon` helps.
- If we don't call `disableHttpLogging()`, just passing `--info` to Gradle makes Google HTTP Client to log network traffic. Therefore, it is still desirable to disable logging in usual cases.

---

Additional info: https://github.com/GoogleContainerTools/jib/pull/2357#discussion_r398840977

chanseokoh:
Fixes #2356.

One of the ways to have a custom logger is through the system property `java.util.logging.config.file`. If it's defined, don't disable Google HTTP Client logging.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
@@ -108,8 +108,8 @@ class TaskCommon {
           }
         });
 
-    // Disable only when not using custom logger.
-    // https://github.com/GoogleContainerTools/jib/issues/2356
+    // Disable Google HTTP Client network logging only when 'java.util.logging.config.file' system
+    // property is undefined: https://github.com/GoogleContainerTools/jib/issues/2356
     if (System.getProperty("java.util.logging.config.file") == null) {
       // Disables Google HTTP client logging.
       java.util.logging.Logger.getLogger(HttpTransport.class.getName()).setLevel(Level.OFF);
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
@@ -108,8 +108,12 @@ class TaskCommon {
           }
         });
 
-    // Disables Google HTTP client logging.
-    java.util.logging.Logger.getLogger(HttpTransport.class.getName()).setLevel(Level.OFF);
+    // Disable only when not using custom logger.
+    // https://github.com/GoogleContainerTools/jib/issues/2356
+    if (System.getProperty("java.util.logging.config.file") == null) {
+      // Disables Google HTTP client logging.
+      java.util.logging.Logger.getLogger(HttpTransport.class.getName()).setLevel(Level.OFF);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


chanseokoh:
Travis seems to have completely stalled. Will try closing and reopening.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -108,8 +108,12 @@ static void disableHttpLogging() {
           }
         });
 
-    // Disables Google HTTP client logging.
-    java.util.logging.Logger.getLogger(HttpTransport.class.getName()).setLevel(Level.OFF);
+    // Disable only when not using custom logger.
+    // https://github.com/GoogleContainerTools/jib/issues/2356
+    if (System.getProperty("java.util.logging.config.file") == null) {
+      // Disables Google HTTP client logging.
+      java.util.logging.Logger.getLogger(HttpTransport.class.getName()).setLevel(Level.OFF);
```
loosebazooka: I'm not sure I understand how this solves the problem? Does gradle have it's own custom logger? And if we disable this, does this show that weird log information that we turned this off for?

chanseokoh: Perhaps my code comment is confusing.

To explain,

- Google HTTP Client's `HttpTransport` class uses `JUL.Logger.getLogger(HttpTransport.class.getName())` to log detailed network traffic.
- Google HTTP Client's `HttpTransport` class output detailed network traffic as INFO level.
- In Gradle, increasing general verbosity to at least `--info` allows JUL to print INFO messages, as expected.
- Therefore, running `gradle --info` generates a lot of network traffic output. This is unnecessary clutter, so we universally disabled the JUL logging of `HttpTransport` by calling `setLevel(Level.OFF)` when we start a task. We still want this.
- Therefore, our FAQ instructions to capture detailed networks logs using JUL (via `-Djava.util.logging.config.file`) doesn't work.

Does this make it clear?

loosebazooka: ooooh I see, yeah sounds good.

PR #2359, opened 2020-03-24
Sync Maven API versions in project build

chanseokoh:
Comparing [maven-core](https://search.maven.org/artifact/org.apache.maven/maven-core) and [maven-compat](https://search.maven.org/artifact/org.apache.maven/maven-compat) releases, I think they should match.

### DIFF
--- a/jib-maven-plugin/build.gradle
+++ b/jib-maven-plugin/build.gradle
@@ -12,8 +12,8 @@ dependencies {
   sourceProject project(':jib-core')
   sourceProject project(':jib-plugins-common')
 
-  implementation 'org.apache.maven:maven-plugin-api:3.5.2'
-  implementation 'org.apache.maven:maven-core:3.5.2'
+  implementation "org.apache.maven:maven-plugin-api:${dependencyVersions.MAVEN_API}"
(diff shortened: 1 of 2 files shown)


jib-maven-plugin/build.gradle
```diff
@@ -31,7 +31,7 @@ dependencies {
   testImplementation "com.github.stefanbirkner:system-rules:${dependencyVersions.SYSTEM_RULES}"
 
   testImplementation 'org.apache.maven.shared:maven-verifier:1.6'
-  testImplementation 'org.apache.maven:maven-compat:3.5.4'
```
TadCordle: So we're downgrading this? Should everything be 3.5.4 instead?

chanseokoh: This downgrades the testing dependency; the end user doesn't see any changes. And although we can upgrade, at least it's safer not to upgrade the API. For example, we can guarantee users using Maven 3.5.2 will continue to work with Jib (although I think it will definitely work even if we upgrade this to 3.5.4.) The problem occurs when we use an API only available in 3.5.4 and the user uses 3.5.2, I think.

PR #2363, opened 2020-03-26
Fix FAQ doc

chanseokoh:
`-Djib.console=plain` unnecessary with `--batch-mode` or `--console=plain`


jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -58,6 +58,7 @@
 import org.junit.contrib.java.lang.system.RestoreSystemProperties;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
+import org.mockito.Answers;
```
loosebazooka: this file seems unrelated?

chanseokoh: (blush) minor cleanup...

PR #2366, opened 2020-03-27
Return copies instead of exposing internal collections / and more

chanseokoh:
- Returns copies of collections to achieve absolute immutable. I missed some fields by mistake.
- Fixes wrong OSGi package export in `jib-build-plan`.

Misc:
- Removes the deprecated `describeContainer`.
- Ensure type before casting; we can remove `@SuppressWarnings`.

### DIFF
--- a/jib-build-plan/build.gradle
+++ b/jib-build-plan/build.gradle
@@ -25,7 +25,7 @@ jar {
     attributes 'Bundle-Vendor': 'Google LLC'
     attributes 'Bundle-DocURL': 'https://github.com/GoogleContainerTools/jib'
     attributes 'Bundle-License': 'https://www.apache.org/licenses/LICENSE-2.0'
-    attributes 'Export-Package': 'com.google.cloud.tools.jib.plugins.api.buildplan'
+    attributes 'Export-Package': 'com.google.cloud.tools.jib.api.buildplan'
   }
 }
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
@@ -604,7 +604,6 @@ public class JibContainerBuilder {
    * @throws InvalidImageReferenceException if the base image value in {@code buildPlan} is an
    *     invalid reference
    */
-  @SuppressWarnings("unchecked")
   public JibContainerBuilder applyContainerBuildPlan(ContainerBuildPlan buildPlan)
       throws InvalidImageReferenceException {
     containerBuildPlanBuilder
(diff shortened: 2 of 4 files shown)


jib-build-plan/build.gradle
```diff
@@ -25,7 +25,7 @@ jar {
     attributes 'Bundle-Vendor': 'Google LLC'
     attributes 'Bundle-DocURL': 'https://github.com/GoogleContainerTools/jib'
     attributes 'Bundle-License': 'https://www.apache.org/licenses/LICENSE-2.0'
-    attributes 'Export-Package': 'com.google.cloud.tools.jib.plugins.api.buildplan'
+    attributes 'Export-Package': 'com.google.cloud.tools.jib.api.buildplan'
```
loosebazooka: is this a modules thing?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java
```diff
@@ -640,8 +642,13 @@ public JibContainerBuilder applyContainerBuildPlan(ContainerBuildPlan buildPlan)
     baseImageConfiguration.getTarPath().ifPresent(builder::setTarPath);
     baseImageConfiguration = builder.build();
 
-    // For now, only FileEntriesLayer is supported in jib-core.
-    layerConfigurations = (List<FileEntriesLayer>) buildPlan.getLayers();
+    Function<LayerObject, FileEntriesLayer> castToFileEntriesLayer =
```
loosebazooka: Can we leave the comment in also?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -302,7 +303,9 @@ public static String getSkaffoldSyncMap(
     // since jib has already expanded out directories after processing everything, we just
     // ignore directories and provide only files to watch
     Set<Path> excludesExpanded = getAllFiles(excludes);
-    for (FileEntriesLayer layer : jibContainerBuilder.describeContainer().getFileEntriesLayers()) {
+    for (LayerObject layerObject : jibContainerBuilder.toContainerBuildPlan().getLayers()) {
+      Verify.verify(layerObject instanceof FileEntriesLayer);
```
loosebazooka: Presumably no user will ever hit this, but should we include an error message if they do?

jib-build-plan/build.gradle
chanseokoh: Yes, an OSGi thing (not JPMS).

PR #2368, opened 2020-03-27
Set up extension API sub-project for Jib Maven plugin

chanseokoh:
Setting up the Extension API sub-project for Jib Maven Plugin.


build.gradle
```diff
@@ -21,6 +21,10 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
+project('jib-maven-plugin-extension-api') {
+  apply plugin: 'java-library'
```
chanseokoh: Should I apply `java-library` to use the `api` dependencies? I didn't on `jib-build-plan`. This kind of stuff is always confusing now that we need to publish JARs from `jib-build-plan` and this sub-project.

jib-maven-plugin-extension-api/build.gradle
```diff
@@ -0,0 +1,40 @@
+plugins {
+  id 'net.researchgate.release'
+  id 'maven-publish'
+}
+
+dependencies {
+  api project(':jib-build-plan')
```
loosebazooka: So I think this part is weird... We don't actually want a direct dependency on the source like this do we?

In the end we should depend on the publicly available `jib-build-plan` obviously we can't do that now since we never released anything.

build.gradle
loosebazooka: I mean if `java` lets you define `api` dependencies, then maybe this is unnecessary?

jib-maven-plugin/build.gradle
```diff
@@ -11,6 +11,7 @@ dependencies {
   sourceProject project(':jib-build-plan')
   sourceProject project(':jib-core')
   sourceProject project(':jib-plugins-common')
+  sourceProject project(':jib-maven-plugin-extension-api')
```
loosebazooka: this one too, I'm unsure of, should we depend on what's publicly available?

jib-maven-plugin-extension-api/build.gradle
chanseokoh: If we assume we published `com.google.cloud.tools:jib-build-plan:0.1.0`, I think `api 'com.google.cloud.tools:jib-build-plan:0.1.0'` is the correct way to go, right? Because the extension dev should be able to access and use those classes in `jib-build-plan`, isn't it? (I assume the dev will only declare this dependency of this Maven extension API, after which I expect the dev can make use of all the classes in both `jib-build-plan` and `jib-maven-plugin-extension-api` plus the maven-core API.) Am I getting this right?

So, given that we haven't released `jib-build-plan`, what do you think is best to move forward at the moment, maybe a temporary measure for the sake of moving forward?

jib-maven-plugin/build.gradle
chanseokoh: Hmm... yeah, I think we should use a published one. But haven't someone said if we don't make a breaking change in `jib-maven-plugin-extension-api`, then including source will be acceptable?

How should we proceed at the moment?

build.gradle
chanseokoh: `java` doesn't define `api`. `java-library` is necessary.

chanseokoh:
I think this should work after the last two commits.

build.gradle
loosebazooka: yeah I dunno if there's a better way to do this, I don't like applying these plugins conditionally in some parent. But maybe this is okay for now.

jib-maven-plugin-extension-api/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+  api project(':jib-plugins-extension-common')
+  api "org.apache.maven:maven-core:${dependencyVersions.MAVEN_API}"
```
loosebazooka, line 9: I guess we expose maven core api in our api huh.

jib-plugins-extension-common/settings.gradle
```diff
@@ -0,0 +1 @@
+rootProject.name = 'jib-plugins-extension-common'
```
loosebazooka, line 1: I feel like this file shouldn't be here.

jib-maven-plugin-extension-api/build.gradle
chanseokoh: The API method signature will be like `extend(ContainerBuildPlan, MavenProject, MavenSession, ...)`, so I guess we need to expose the maven-core, unless there's something else workable?

jib-plugins-extension-common/settings.gradle
chanseokoh: I still don't know what this file is for. Is this required for releasing? We need to publish this sub-project.

build.gradle
chanseokoh: Oh, maybe I should put this in `jib-maven-plugin-extension-api` instead at root.

chanseokoh: Nah, there was a reason I put it at root. Looks like `java-library` needs to be applied before `java`.

loosebazooka: oh yeah I agree, I think maybe there's a consequence to us putting so much in the root buildfile :/

PR #2369, opened 2020-03-28
fix more error prone issues

loosebazooka:
- mark unused variable in jsontemplates
- rename conflicting type: Type (`org.objectweb.asm.Type` and `MainClassFinder.Result.Type`)
  - although this could also just be fixed using fqn of the types.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/MainClassFinder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/MainClassFinder.java
@@ -42,7 +42,7 @@ public class MainClassFinder {
   public static class Result {
 
     /** The type of result. */
-    public enum Type {
+    public enum Category {
 
       // Found a single main class.
       MAIN_CLASS_FOUND,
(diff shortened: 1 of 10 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/api/MainClassFinder.java
```diff
@@ -42,7 +42,7 @@
   public static class Result {
 
     /** The type of result. */
-    public enum Type {
+    public enum Category {
```
chanseokoh: Too bad, this is a public API class and the `Type` is exposed as public. We need `@depraecated` and a `CHANGELOG` update?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
```diff
@@ -43,7 +43,10 @@
   @VisibleForTesting
   class ContentDescriptorTemplate implements JsonTemplate {
 
-    @Nullable private String mediaType;
+    @SuppressWarnings("unused")
```
chanseokoh: I wonder if it is possible to configure errorpone to exclude classes of type `JsonTemplate`. But I'm not sure which one is better in the end even if it's possible

jib-core/src/main/java/com/google/cloud/tools/jib/api/MainClassFinder.java
loosebazooka: You know what, I didn't even realize this was api. Lemme go reverse this and just use fqn for asm.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
loosebazooka: Hrmm.. I'm thinking being more explicit about it means we did it intentionally, rather than wholesale.

PR #2371, opened 2020-03-30
Make useragent nullable in RegistryClient

### LINKED ISSUE #2367: registry client should accept null user agent string
RegistryClient constructor should accept null values for user agent string. 

an extension of: https://github.com/GoogleContainerTools/jib/pull/2352/commits/401887a29bc689d5ecb1a300f81d0a476968fea6

loosebazooka:
and also any underlying internal classes

fixes #2367

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
@@ -64,7 +64,7 @@ public class RegistryClient {
     private final RegistryEndpointRequestProperties registryEndpointRequestProperties;
     private final FailoverHttpClient httpClient;
 
-    @Nullable private String userAgent = "";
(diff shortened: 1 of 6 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -64,7 +64,7 @@
     private final RegistryEndpointRequestProperties registryEndpointRequestProperties;
     private final FailoverHttpClient httpClient;
 
-    private String userAgent = "";
+    @Nullable private String userAgent = "";
```
chanseokoh: This should not assign `""`. The default should be null.

loosebazooka: Makes sense.

PR #2376, opened 2020-04-01
Update gradle plugin publishing

loosebazooka:
- update plugin version
  - We had to update from 10.1 -> 11.0 because they were no longer supporting the 10.1 based publishings (https://blog.gradle.org/plugin-portal-update)
- use syntax from https://guides.gradle.org/publishing-plugins-to-gradle-plugin-portal/
  - This full syntax is required for local publishing to work correctly
- update local development workflow

I tested this release process and it seems to work fine.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,7 @@ To use a local build of the `jib-gradle-plugin`:
 
   1. Build and install `jib-gradle-plugin` into your local `~/.m2/repository`
      with `./gradlew jib-gradle-plugin:publish`
-  1. Add to your test project's `settings.gradle` the following:
+  1. Add a `pluginManagement` block to your test project's `settings.gradle` to enable reading plugins from the local maven repository. It must be the first block in the file before any `include` directives.
         ```groovy
         pluginManagement {
           repositories {
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,26 +109,23 @@ If developing with Eclipse and M2Eclipse (the Maven tooling for Eclipse), just l
 To use a local build of the `jib-gradle-plugin`:
 
   1. Build and install `jib-gradle-plugin` into your local `~/.m2/repository`
-     with `./gradlew jib-gradle-plugin:install`;
-  1. Modify your test project's `build.gradle` to look like the following:
+     with `./gradlew jib-gradle-plugin:publish`
+  1. Add to your test project's `settings.gradle` the following:
         ```groovy
-        buildscript {
-            repositories {
-                mavenLocal() // resolve in ~/.m2/repository
-                mavenCentral()
-            }
-            dependencies {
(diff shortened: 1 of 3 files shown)


jib-gradle-plugin/build.gradle
```diff
@@ -47,17 +47,20 @@ release {
...
       id = 'com.google.cloud.tools.jib'
       displayName = 'Jib'
       description = 'Containerize your Java application'
-      tags = ['google', 'java', 'containers', 'docker', 'kubernetes', 'microservices']
+      implementationClass = 'com.google.cloud.tools.jib.gradle.JibPlugin'
     }
   }
 }
-tasks.publishPlugins.dependsOn integrationTest
+//tasks.publishPlugins.dependsOn integrationTest
```
chanseokoh, line 63: Should we keep this?

jib-gradle-plugin/build.gradle
```diff
@@ -3,7 +3,7 @@ plugins {
   id 'net.researchgate.release'
   id 'com.gradle.plugin-publish'
   // for local build/test
```
chanseokoh, line 5: Can we say `// for local SNAPSHOT build/install/test`? Unrelated to this PR, yesterday I couldn't figure out what this does and removed it in one of my branches.

jib-gradle-plugin/build.gradle
```diff
@@ -74,3 +77,11 @@ eclipse.classpath.file.whenMerged {
   }
 }
 /* ECLIPSE */
+
+/* LOCAL DEV */
```
chanseokoh, line 81: How about LOCAL SNAPSHOT DEV?

jib-gradle-plugin/build.gradle
loosebazooka: oh yeah oops, I just needed it to confirm something.

jib-gradle-plugin/build.gradle
loosebazooka: you mean the `maven-publish` plugin? In other sub-projects we use it to generate the poms.

CONTRIBUTING.md
```diff
@@ -109,26 +109,23 @@ If developing with Eclipse and M2Eclipse (the Maven tooling for Eclipse), just l
...
         ```groovy
-        buildscript {
-            repositories {
-                mavenLocal() // resolve in ~/.m2/repository
-                mavenCentral()
-            }
-            dependencies {
-                classpath 'com.google.cloud.tools:jib-gradle-plugin:2.1.1-SNAPSHOT'
-            }
+        pluginManagement {
```
chanseokoh, line 115: Can we add that this must be the first block in `settings.gadle`? https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_management

loosebazooka: yeah that's a good idea

PR #2378, opened 2020-04-01
Fix ImageReference.toString() for "scratch"

chanseokoh:
`ImageReference.scratch().toString()` was returning `/scratch` (with the leading slash).

Also cleaned up some test code.


jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
```diff
@@ -316,6 +316,10 @@ public ImageReference withTag(String newTag) {
    */
   @Override
   public String toString() {
+    if (registry.isEmpty() && repository.equals("scratch") && tag.isEmpty()) {
```
loosebazooka, line 319: should we just use isScratch() directly?

PR #2380, opened 2020-04-02
Add Maven plugin extension API

chanseokoh:
The main plugin extension point is through the following method signature. The focus is on the capability to adjust a build plan prepared by Jib plugins.

```
      ContainerBuildPlan buildPlan,
      MavenProject project,
      MavenSession session,
      ExtensionLogger logger)
```

`JibPluginExtensionException` is defined to allow extensions to propagate any problems during extension execution.

`ExtensionLogger` is to allow extensions to log to the console. Note it is not dependent on any of the logging facilities in `jib-core`.


jib-maven-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/maven/extension/JibMavenPluginExtension.java
```diff
@@ -0,0 +1,49 @@
...
+   * @param session the {@link MavenSession}
+   * @param logger logger for writing log messages
+   * @return updated build plan
+   * @throws JibPluginExtensionException if an error occurs while running the plugin extension
+   */
+  ContainerBuildPlan extendContainerBuildPlan(
+      ContainerBuildPlan buildPlan,
+      MavenProject project,
+      MavenSession session,
+      ExtensionLogger logger)
```
loosebazooka, line 47: So we just inject an implementation of this that connects to the jib event system?

jib-plugins-extension-common/src/main/java/com/google/cloud/tools/jib/plugins/extension/JibPluginExtension.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.extension;
+
+/** Base type for Jib plugin extensions. */
+public interface JibPluginExtension {}
```
loosebazooka, line 20: this is used just to indentify the services to load?

jib-plugins-extension-common/src/main/java/com/google/cloud/tools/jib/plugins/extension/JibPluginExtensionException.java
```diff
@@ -0,0 +1,59 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.extension;
+
+import java.util.concurrent.ExecutionException;
+
+/** Exception while running Jib plugin extensions. */
+public class JibPluginExtensionException extends ExecutionException {
```
loosebazooka, line 22: It doesn't feel like `ExecutionException` is the right Exception to extend? Do we even need to really extend an exception like this?

jib-maven-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/maven/extension/JibMavenPluginExtension.java
```diff
@@ -0,0 +1,49 @@
...
+   * @param buildPlan original build plan prepared by the Jib Maven plugin
+   * @param project the {@link MavenProject}
+   * @param session the {@link MavenSession}
+   * @param logger logger for writing log messages
+   * @return updated build plan
+   * @throws JibPluginExtensionException if an error occurs while running the plugin extension
+   */
+  ContainerBuildPlan extendContainerBuildPlan(
+      ContainerBuildPlan buildPlan,
+      MavenProject project,
```
loosebazooka, line 45: Instead of MavenProject, MavenSession should we send a `MavenStuff` (not Stuff but something like that) that contains anything from Maven that we inject?

If we ever extend this out to include to more Maven things, like stuff from the Reactor or something, then we can just add them to MavenStuff without changing the API and older extensions will continue to work?

jib-maven-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/maven/extension/JibMavenPluginExtension.java
chanseokoh: That's right. Here's the [logger](https://github.com/GoogleContainerTools/jib/blob/6e9dfbdc8f54d28c095bf4c6349ba2d3946262f6/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenExtensionLogger.java#L33) in the prototype.

jib-plugins-extension-common/src/main/java/com/google/cloud/tools/jib/plugins/extension/JibPluginExtension.java
chanseokoh: Not really, but at least this type is required in the shared `JibPluginExtensionException` whose constructors force to pass an extension class.

jib-plugins-extension-common/src/main/java/com/google/cloud/tools/jib/plugins/extension/JibPluginExtensionException.java
chanseokoh: Ah, good catch. Leftover from previous POC iterations.

jib-maven-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/maven/extension/JibMavenPluginExtension.java
chanseokoh: `MavenBaggage`, `MavenObjects`,`MavenInterface`, `MavenFacet`, `MavenControlPlane`, or something else? I think `MavenBaggage` works and is future proof. Thoughts?

loosebazooka: I think maybe `Baggage` has a little bit of a negative connotation, but I can't really think of a better one for now. `MavenEnv`? <-- but even this doesn't sound great.

loosebazooka: `MavenBuildProperties` ?

chanseokoh: We settled on `MavenData`. I went for an `interface` instead of `class` for `MavenData`.

chanseokoh:
Merging this to put up the next PR. If anyone finds an issue, we'll have plenty of time to fix it while things are being shaped up.

PR #2381, opened 2020-04-02
Correctly parse ImageReference with both tag and digest

### LINKED ISSUE #1481: ImageReference should support tag and digest
According to [the Docker spec](https://github.com/docker/distribution/blob/0d3efadf0154c2b8a4e7b6621fff9809655cc580/reference/reference.go#L6
), it's ok to have a reference with both a tag *and* a digest such as `ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68`

See also: google/go-containerregistry#351

The following test for `ImageReferenceTest` fails
```java
  @Test
  public void testParse_tagAndDigest() throws InvalidImageReferenceException {
    String imageReferenceString =
        "ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68";
    ImageReference imageReference = ImageReference.parse(imageReferenceString);

    Assert.assertEquals("registry.hub.docker.com", imageReference.getRegistry());
    Assert.assertEquals("ubuntu", imageReference.getRepository());
    Assert.assertEquals("latest", imageReference.getTag());
    Assert.assertEquals(
        "sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68",
        imageReference.getTag());
    Assert.assertTrue(imageReference.isTagDigest());
  }
```

TadCordle:
Fixes #1481. Since `ImageReference` is part of the public API, there were a couple weird changes to keep backwards compatibility (see the jib-core changelog).


jib-core/CHANGELOG.md
```diff
@@ -5,7 +5,14 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Multiple additions to `ImageReference` to separate `tag` and `digest`. ([#1481](https://github.com/GoogleContainerTools/jib/issues/1481))
+    - `of(registry, repository, tag, digest)` to create an image from a digest
```
chanseokoh, line 9: "... from a tag and digest"?

jib-core/CHANGELOG.md
```diff
@@ -5,7 +5,14 @@ All notable changes to this project will be documented in this file.
...
+    - `isValidDigest(digest)` to check if a string is a valid digest
+    - `getDigest()` to get the digest
+    - `parse()` now supports image reference strings containing both a tag and a digest
+
 ### Changed
+- `ImageReference#isTagDigest` is now deprecated.
```
chanseokoh, line 15: nit: newline before this line

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
```diff
@@ -346,6 +387,10 @@ public String toString() {
    * @return the image reference in Docker-readable format, without hiding the tag
    */
   public String toStringWithTag() {
+    // Insert tag before digest
+    if (!Strings.isNullOrEmpty(digest)) {
+      return toString();
```
chanseokoh, line 392: I think the purpose of this method is to ensure `:latest` is added. I don't think `toString()` will add `:latest`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
```diff
@@ -360,11 +405,12 @@ public boolean equals(Object other) {
     ImageReference otherImageReference = (ImageReference) other;
     return registry.equals(otherImageReference.registry)
         && repository.equals(otherImageReference.repository)
-        && tag.equals(otherImageReference.tag);
+        && tag.equals(otherImageReference.tag)
+        && Objects.equals(digest, otherImageReference.digest);
```
chanseokoh, line 409: I know `digest` is of `Optional`, but is this different from `digest.equals(otherImageReference.digest)`?

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Glob pattern support for `jib.extraDirectories.permissions`. ([#1200](https://github.com/GoogleContainerTools/jib/issues/1200))
+- Support for image references with both a tag and a digest. ([#1481](https://github.com/GoogleContainerTools/jib/issues/1481))
```
chanseokoh, line 9: I'm curious how Jib will work and respond if there's an error for a base image, e.g.,
   - tag exists but digest doesn't match
   - digest exists but tag doesn't

Will they work? If they fail, will the error message make sense and be helpful?

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
TadCordle: I'll have to check how it's used in the rest of the codebase, but this is used for the offline mode cache directory structure. So if we add "latest" here, people may need to repopulate their cache.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
TadCordle: digest is a `@Nullable` string, so `Objects.equals` is used here to avoid NPE

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
chanseokoh: Ugh, the offline manifest caching! So we want that to be more like `toStringWithTagOrDigest()` where a digest takes precedence, right? The naming is a bit unfortunate. `toStringWithTag` doesn't make sense and is confusing. Should we just create a new method? I wish we could create an internal-only method for the offline caching.

jib-gradle-plugin/CHANGELOG.md
TadCordle: I can test to be sure, but what I think happens is:

* tag exists but digest doesn't match - this will cause `of()` to fail since it does a `Preconditions` check on the format of the tag and digest, and `parse` should throw an `InvalidImageReferenceException` since the regex won't match
* digest exists but tag doesn't - this will succeed, I modified the test to loop through every combination of registry/repository/tag/digest we have instead of separating tag and digest, which means it covers null tag + existent digest.

chanseokoh: Sorry, I should have been clear: I am talking about referring a base image in a remote registry. For example, I say `<from><image>openjdk:8@sha256:aaaaaa... </image></from>` (i.e., the digest `sha256:aaaa...` is not the right digest of the actual `openjdk:8` image.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
chanseokoh: Whether or not we deprecate this method, I think this method should be updated to fulfill the documented contract so that it returns `registry/repository:latest@sha256...` when both a tag and a digest are given. Given the name `toStringWithTag()`, returning `registry/repository@sha256...` doesn't look right to me.

I'm struggling with the name for the new method too. I don't think there's a good name to cover both tag and digest. This [grammar description](https://github.com/docker/distribution/blob/742aab907b54a367e1ac7033fb9fe73b0e7344f5/reference/reference.go#L6-L24) may be worth checking, although I don't know where the description came from. At least `toStringWithQualifier` sounds better than "identifier".

Anyways, maybe we should mark the new method as "internal"–the purpose is too specific to our use-case. @loosebazooka had some discussions around how we could add internal methods to our public API, but we couldn't find a good solution. The best I could do was to [document it](https://github.com/GoogleContainerTools/jib/pull/2353/files#diff-d390f64e06290a875d4e6a149d0f18ccR580).

TadCordle: >Given the name toStringWithTag(), returning registry/repository@sha256... doesn't look right to me.

The problem is that this was what the old behavior was, regardless of whether or not the naming was correct, since we were treating digests as tags. Is it more important to preserve old behavior in this case or change the behavior to better match the name?

TadCordle: Ok, I think given jib-core still isn't GA, we can just rename the method now instead of having to deprecate.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
```diff
@@ -137,18 +138,11 @@ public static ImageReference parse(String reference) throws InvalidImageReferenc
...
-      }
-    } else if (!Strings.isNullOrEmpty(digest)) {
-      tag = digest;
-    } else {
+    if (Strings.isNullOrEmpty(tag)) {
       tag = DEFAULT_TAG;
     }
 
-    return new ImageReference(registry, repository, tag);
+    return new ImageReference(registry, repository, tag, digest);
```
chanseokoh, line 145: I actually think we can call `of(registry, repository, tag, digest)` here and get rid of some `if` checks in this method, as `parse()` and `of()` duplicate some logic. But let's not touch this method anymore in this PR to be on the safe side.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
```diff
@@ -334,18 +360,27 @@ public String toString() {
     // Use tag if not the default tag.
     if (!DEFAULT_TAG.equals(tag)) {
       // Append with "@tag" instead of ":tag" if tag is a digest
```
chanseokoh, line 362: Remove this comment.

BTW, I think it is correct that this method prints both a (non-default) tag and a digest.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
```diff
@@ -334,18 +360,27 @@ public String toString() {
...
+   * Stringifies the {@link ImageReference}, including the default tag if no tag or digest is set.
    *
-   * @return the image reference in Docker-readable format, without hiding the tag
+   * @return the image reference in Docker-readable format, including the default tag if no tag or
+   *     digest is set.
    */
-  public String toStringWithTag() {
+  public String toStringWithQualifier() {
+    // Insert tag before digest
+    if (!Strings.isNullOrEmpty(digest)) {
```
chanseokoh, line 381: Hmm... so if both a digest and a non-default tag are given, now `toStringWithQualifier()` will print both. Are we good with that?

jib-gradle-plugin/CHANGELOG.md
chanseokoh: I'd like to check this behavior.

jib-core/src/main/java/com/google/cloud/tools/jib/api/ImageReference.java
TadCordle: Yes, I think this is correct.

chanseokoh: Just confirming: if so (e.g., returning `image:sometag@sha256...`), isn't it more natural for this method to consistently return `image:latest@sha256...` when the tag is `latest`? Currently, it returns `image@sha256...`. But if the intention is to make it return at least a tag or a digest, I can live with that. However, it's hard to predict the behavior just by reading the Javadoc.

TadCordle: Yes, this is intended. Maybe always returning a tag would be more consistent, but again this is used in offline mode and we don't want to start returning `image:latest@sha256` or people will need to repopulate the cache.

`image:sometag@sha256` wasn't possible before this PR, so in this case we could go either way. Maybe digest should just take precedence over tag instead of returning both?

chanseokoh: > Maybe digest should just take precedence over tag instead of returning both?

I actually liked this. That is, this method appends a digest (`image@sha256`) if (and only if) it's set; otherwise, it appends a tag, including `latest`.

That said, I just realized some issue with our assumption with `latest`. It is indeed true that `latest` is assumed when no tag or digest is given. And so tar, we were okay, as it was not possible to have a digest and a tag at the same time. But what if the user simply referred to a base image only by a digest? In many cases, on an actual repository, the manifest would not be attached a `latest` tag. The question is then, would it make sense for `toString()` to append an unwarranted `:latest` that is possibly incorrect in practice? Perhaps if a digest is given but not a tag, we shouldn't assume the tag is `latest`?

TadCordle: In that case should tag be `null` if a digest is set? Seems like it will need more changes outside of this method.

chanseokoh: Agreed. I'm more convinced that `latest` is the default value only when nothing is given. When a digest is given, it doesn't make sense to add `latest`.

TadCordle: Ok. Do you think we should make tag `null` when the digest is set in this PR, or wait for a followup? I'm not yet sure how many extra changes making `tag` nullable might involve.

chanseokoh: If you are committed to fix all these in a quick succession, I am fine with a follow-up. At least it should be fixed before we can make a new release. Do you think this PR is ready for merging?

TadCordle: I haven't checked [this](https://github.com/GoogleContainerTools/jib/pull/2381#discussion_r402538376) yet, but other than that, yeah I think it's ready.

chanseokoh: Alright. I think checking that also depends on how we assume `latest`, so we can check it in a follow-up.

Issue #2360, opened 2020-03-25
Option to disable adding new tags to existing images


karlmuscat-dpg:
Is there a way how to disable adding new tags to existing images?

We have a monorepo with multiple microservices, and for every build, we create a new tag. We would like to tag only those services that generate new images. Other services that have no changes (hence, their images have the same digest) do not need to get new tags.

loosebazooka:
Could you help clarify what you mean by new tags? Presumably if your tag is not dynamically generated this shouldn't be happening?

systems-dpg:
Yes, the tags are dynamically generated and their value changes for every build.

This is a typical build we have: some microservices have changes and others no. A new tag is generated, and each service is built. As it is, Jib will:
1. push images that are new with they newly created tag, and, 
2. add a tag to the existing images for the services that have no changes in them.

We would like to disable (2), ie, not add tags to existing images.

karlmuscat-dpg:
Yes, the tags are dynamically generated and their value changes for every build.

This is a typical build we have: some microservices have changes and others no. A new tag is generated, and each service is built. As it is, Jib will:

push images that are new with they newly created tag, and,
add a tag to the existing images for the services that have no changes in them.
We would like to disable (2), ie, not add tags to existing images.

loosebazooka:
So a couple of things here.
- Is this maven or gradle?
- Can you know already at build time which services need to be updated? -- and just skip jib yourself? Presumably this is much harder in a CI.
- You may already be aware: but tags on a registry are just pointers to digests. So a new tag doesn't really do anything except add noise to the entries. Could be annoying if something in your container orchestration is watching for new tags though.

I think there might be some scope for adding this feature as a flag/config option, but I'm not sure yet.

karlmuscat-dpg:
- Gradle
- Not at this time. 
- This is exactly the issue. We have some orchestrator watching the registry, and it is unnecessarily deploying the services that have no changes.

loosebazooka:
Thanks @karlmuscat-dpg this is useful information. If you're interested in submitting a design/proposal for this that we can work on together that could speed up the process -- https://github.com/GoogleContainerTools/jib/tree/master/proposals

I would start off with something less formal, like a range of ideas and we can narrow down what might work in the comments.

If that is not an option for you, someone on the jib dev team can take a look at some point in the (hopefully near) future.

karlmuscat-dpg:
Sure. I'll give it a shot.

[...]

PR #2382, opened 2020-04-03
Adds control to push tags on existing images in target registry

karlmuscat:
Relevant issues: [#2360](https://github.com/GoogleContainerTools/jib/issues/2360)

Currently Jib pushes tags for all built images, irrelevant of whether the image is already 
present on the target registry. 

As highlighted by @loosebazooka [#2360](https://github.com/GoogleContainerTools/jib/issues/2360) 
this can be particularly annoying if some automated orchestrator is monitoring the registry
and releases any new tags found. This issue is especially prevalent in cases where the tags are
auto-generated by CI systems and added with every triggered build.

The idea is to give the user the ability to control this functionality by means of a switch/option. 
There might be use-cases where the current behaviour is preferable, eg, where the tags follow semantic 
versioning, or where the tags are not automatically generated. Hence, adding a switch seems to be 
a better solution rather than changing the current behaviour.


karlmuscat:
@googlebot I signed it!

loosebazooka:
Hey @karlmuscat usually for a change like this, we start off with just the proposal, and work on that before adding code. Sorry if my instructions were unclear. As there is some cost to adding and removing configuration, we need to review the design first before doing any coding work.

karlmuscat:
@loosebazooka no problem at all. Do you need me to add some more detail to this PR to get the discussion going?

loosebazooka:
@karlmuscat yes, can you create a new PR (you can leave this open) with just the design proposal?

karlmuscat:
@loosebazooka Created new PR. Ill close this one to avoid confusion.

PR #2387, opened 2020-04-06
Proposal: Control the pushing of tags on existing images in target registry

karlmuscat:
# Proposal: Control the pushing of tags on existing images in target registry

Relevant issues: [#2360](https://github.com/GoogleContainerTools/jib/issues/2360)

## Motivation
Currently Jib pushes tags for all built images, irrelevant of whether the image is already 
present on the target registry. 

As highlighted by @loosebazooka [#2360](https://github.com/GoogleContainerTools/jib/issues/2360) 
this can be particularly annoying if some automated orchestrator is monitoring the registry
and releases any new tags found. This issue is especially prevalent in cases where the tags are
auto-generated by CI systems and added with every triggered build.

## Proposal
The idea is to give the user the ability to control this functionality by means of a switch/option. 
There might be use-cases where the current behaviour is preferable, eg, where the tags follow semantic 
versioning, or where the tags are not automatically generated. Hence, adding a switch seems to be 
a better solution rather than changing the current behaviour.


loosebazooka:
@karlmuscat any chance you could include your example workflow that uses this? The pushback I'm hearing from within the jib team is that tags are simply pointing to digests, and the trigger that looks for new tags can just verify that digests haven't changed since last build?

chanseokoh:
I think now I understand the workflow. I guess the user's workflow is that

1. For every Maven build (or at least every Jib build), tags are always dynamically generated, and whenever they are generated, they become different even if nothing in the project is changed and no single byte of the image is changed. For example, tags contain current time information.
1. Therefore, even if Jib doesn't push any new blob/layer/container config/manifest, the registry will receive newly generated tags for the same image that remained unchanged.
1. Some system monitoring the registry noticed the new tags and starts redeploying the same image.

Is this right? As @loosebazooka it'd be good to document an example workflow. I agree this feature can be useful.

karlmuscat:
The workflow is precisely as @chanseokoh described it. To be exact, we are following the Gitops model and the setup is as depicted here: 

https://www.twistlock.com/wp-content/uploads/2018/10/Weave2.png

As you can see the Deploy Agent is monitoring the Docker registry. Our CI generates a tag for every new build. The Code Repo is a monorepo made of several microservices. As properly described, if a microservice has no changes, it still gets a new tag and as a consequence the Deploy Agent releases it to the cluster. It is far easier to stop the pushing of these “unnecessary” tags than to deduce which microservice has not changed and eliminate it from the CI build.

loosebazooka:
Hey @karlmuscat ,

so here are a couple of things to consider:
1. Perhaps we can start this off as a system property rather than config parameter and then move it over as we understand its usage. `-Djib.skipExistingImages` configured in `com.google.cloud.tools.jib.global.JibSystemProperties`
1. If the property is set, the control happens at the `PushImageStep` level, potentially creating a new step if the property is set that affects `PushImageStep.makeList`?

karlmuscat:
> Perhaps we can start this off as a system property rather than config parameter and then move it over as we understand its usage. `-Djib.skipExistingImages` configured in `com.google.cloud.tools.jib.global.JibSystemProperties`

Agree.

> If the property is set, the control happens at the PushImageStep level, potentially creating a new step if the property is set that affects PushImageStep.makeList?

Are you suggesting to create a new step class that is executed in `StepsRunner.pushImages()` only if the property is set to true? Its result would then affect the execution of the `PushImageStep.makeList()` part? If this understanding is correct, why is this approach preferred over a different approach being taken in the `PushBlobStep` where the blob existence is being checked within the same step, rather than introducing a new step? I'm referring to line 74:

`if (!forcePush && registryClient.checkBlob(blobDigest).isPresent())`

In similar fashion, `PushImageStep` would be modified, somewhere around line 117, to include:

`if (skipExistingImages && registryClient.checkManifest(imageDigest).isPresent())`

loosebazooka:
I think I was overly concerned about a minor perceived optimization that maybe could add some async cost in some situations. Anyway, I think you're right, the simpler solution makes more sense here.

karlmuscat:
@loosebazooka do you want me to commit some initial changes related to this?

loosebazooka:
Oh sorry, I've been kinda slow on this:
Um we usually design out the whole feature before putting up a code change PR: Here's an example of a fairly simple one: https://github.com/GoogleContainerTools/jib/blob/master/proposals/archives/output_files.md

So to add to the current proposal, we want to explain the implementation and effect on the user experience.

Sorry if it's a little heavy weight, our process is like:
1. Finish proposal, commit it
2. Make a PR with code changes + changelog
3. Create a documentation PR (jib team can probably do this, since we don't commit until *after* release)

chanseokoh:
> I think I was overly concerned about a minor perceived optimization that maybe could add some async cost in some situations (I was concerned that we might be querying the registry too many times for multi tags -- a problem on a slower network). Anyway, I think you're right, the simpler solution makes more sense here.

Sorry, I thought this was just a small optimization that saves registry check calls but not worth adding code complexity, but now I realized this is required for correctness too.

> Are you suggesting to create a new step class that is executed in `StepsRunner.pushImages()` only if the property is set to true? Its result would then affect the execution of the `PushImageStep.makeList()` part?

So, yes. This is exactly what you need to do. Otherwise, some of the competing `checkManifest()` calls will return false while others true, so you'll end up pushing only a subset of configured tags. All `PushImageStep`s are fired roughly at the same time, so I imagine the chance of this happening in practice would actually be high. So, you should create a new step to check the manifest, and based on that result, you should either create all `PushImageStep`s or nothing.

karlmuscat:
Agree. Forgot about the additional tags possibility.

Based on this, I propose that a new step is created with the below signature:
`class CheckImageStep implements Callable<BuildResult>`, where its `call()` function performs a call to the `RegistryClient`. This step is executed in the `StepsRunner.pushImages()` method, before the `PushImageStep.makeList()` part, only if the `skipExistingImages` flag is set to true. Depending on the `BuildResult`, the `PushImageStep` futures are then created (or not).

The registry call used in `CheckImageStep` can be implemented in various ways. I propose that a new `checkManifest` function is added in `RegistryClient`. This should execute a `ManifestChecker`, also a new class. The `ManifestChecker`'s `handleHttpResponseException` class handles the `HTTP/404` scenario.

loosebazooka:
@karlmuscat looks good. Lets formalize this in the design proposal itself and we can move forward.

Again, thanks for doing this.

karlmuscat:
@loosebazooka proposal updated

proposals/tags-on-existing-images.md
```diff
@@ -0,0 +1,33 @@
...
+There might be use-cases where the current behaviour is preferable, eg, where the tags follow semantic 
+versioning, or where the tags are not automatically generated. Hence, adding a switch seems to be 
+a better solution rather than changing the current behaviour.
+
+## Proposed changes
+A new system property is to be added to control this new behaviour:
+`-Djib.skipExistingImages`
+
+A new step is created with the below signature:
+`class CheckImageStep implements Callable<BuildResult>`, where its `call()` function performs a
```
chanseokoh, line 25: We will have to see, but I feel like this step should return a manifest (as currently built in `PushImageStep.makeList()`) too, so that `PushImageStep` accepts the manifest. May need to create a "struct" class to contain manifest, imageDigest, and imageId. I think we can return imageDigest and imageId as separate fields instead of putting them into a `BuildResult` struct.

But I think these kinds of details can be sorted out when we actually work on a PR. I don't want too much bike-shedding in this proposal.

proposals/tags-on-existing-images.md
```diff
@@ -0,0 +1,33 @@
...
+
+A new step is created with the below signature:
+`class CheckImageStep implements Callable<BuildResult>`, where its `call()` function performs a 
+call to the `RegistryClient`. This step is executed in the `StepsRunner.pushImages()` method, 
+before the `PushImageStep.makeList()` part, only if the `skipExistingImages` flag is set to true. 
+Depending on the `BuildResult` of this new step, the `PushImageStep` futures are then created (or not).
+
+The registry call used in `CheckImageStep` can be implemented in various ways. It is being 
+proposed that a new `checkManifest` function is added in `RegistryClient`. This should execute 
+a `ManifestChecker`, also a new class. The `ManifestChecker`'s `handleHttpResponseException` class
```
chanseokoh, line 32: I saw in your original PR you just used `RegistryClient.pullManifest()`. What's the reason you changed your mind to create a new `ManifestChecker`?

karlmuscat: `pullManifest` throws an exception if the manifest does not exist. If used directly in the step, some exception handling would be needed. Anyone making the same check elsewhere would need to handle the same exception, possibly repeating code. 

The `ManifestChecker` seems more inline with the current code structure, eg, `checkBlob` using `BlobChecker`. Alternatively if it seems like over engineered, the `pullManifest` can be used within the `checkManifest` in the `RegistryClient`, and exception handling is centralized in the `RegistryClient`. The latter was my original choice, however it seems much cleaner handling the exception within the `handleHttpResponseException` method rather than a singluar try-catch block.

chanseokoh: Alright, let's go with `ManfiestChecker`. I looked at the `pullManifest` code, and now I think we do need a way to distinguish "manifest not found" from other errors.

chanseokoh:
Thanks @karlmuscat!

karlmuscat:
My pleasure 😉Should I start working on the code PR or will you guys be taking over now?

chanseokoh:
I don't think we'll get to work on this soon, so it's up to you!

PR #2397, opened 2020-04-09
Run Maven plugin extensions

chanseokoh:
Using the standard Service Provider Interfaces (SPI) mechanism to load and run Maven plugin extensions.

This doesn't yet involve the new extension config option (like `<extensions><extension><implementation>com.example.ExtensionClass`). The config option can be added in a later PR.

With this, we are actually enabling running and testing Maven extensions. A configuration will look like this.

```xml
<plugin>
  <version>2.1.1-SNAPSHOT</version>
  <dependencies>
    <dependency>
      <groupId>com.example</groupId>
      <artifactId>quarkus-extension-maven</artifactId>
      <version>0.1.0-SNAPSHOT</version>
    </dependency>
  </dependencies>
</plugin>
```

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
@@ -17,15 +17,19 @@
 package com.google.cloud.tools.jib.maven;
 
 import com.google.cloud.tools.jib.api.Containerizer;
+import com.google.cloud.tools.jib.api.ImageReference;
+import com.google.cloud.tools.jib.api.InvalidImageReferenceException;
 import com.google.cloud.tools.jib.api.JavaContainerBuilder;
 import com.google.cloud.tools.jib.api.JavaContainerBuilder.LayerType;
 import com.google.cloud.tools.jib.api.JibContainerBuilder;
 import com.google.cloud.tools.jib.api.LogEvent;
+import com.google.cloud.tools.jib.api.buildplan.ContainerBuildPlan;
 import com.google.cloud.tools.jib.event.events.ProgressEvent;
 import com.google.cloud.tools.jib.event.events.TimerEvent;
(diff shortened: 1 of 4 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -526,6 +533,37 @@ boolean jarRepackagedBySpringBoot() {
   @Override
   public JibContainerBuilder runPluginExtensions(JibContainerBuilder jibContainerBuilder)
       throws JibPluginExtensionException {
-    return jibContainerBuilder;
+    return runPluginExtensions(
+        ServiceLoader.load(JibMavenPluginExtension.class).iterator(), jibContainerBuilder);
```
loosebazooka, line 537: I wonder if we need to create some sort of ordering mechanism so like some file manipulation extension can run *after* a framework extension has populated all the file?

chanseokoh: Good idea. In a follow-up PR, I'll add a new extension configuration option, as I proposed in the internal doc. I think I will be able to enforce ordering with it.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -526,6 +533,37 @@ boolean jarRepackagedBySpringBoot() {
...
+  }
+
+  @VisibleForTesting
+  JibContainerBuilder runPluginExtensions(
+      Iterator<JibMavenPluginExtension> services, JibContainerBuilder jibContainerBuilder)
+      throws JibPluginExtensionException {
+    if (!services.hasNext()) {
+      log(LogEvent.debug("No Jib plugin extension discovered"));
```
loosebazooka, line 545: extension -> extensions

PR #2400, opened 2020-04-13
Set up sub-project for Gradle plugin extension and define extension API

chanseokoh:
Continued work on establish the Jib plugin extension framework.

- Sets up a sub-project for Gradle plugin extension (counterpart of #2368)
- Defines Gradle plugin extension API (counterpart of #2380)
   ```
       ContainerBuildPlan buildPlan, GradleData gradleData, ExtensionLogger logger)
   ```

One problem due to a difference from Maven: while the Maven Core API was defined with
```
dependencies {
  api "org.apache.maven:maven-core:${dependencyVersions.MAVEN_API}"
```
here the Gradle API dependency is though applying `java-gradle-plugin`.
```gradle
plugins {
  id 'java-gradle-plugin'
```
The problem (I think) is that the Gradle API version is not pinned. So the Gradle version that the extension dev uses will determined the API version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,9 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
+project('jib-gradle-plugin-extension-api') {
+  apply plugin: 'java-library'
+}
 project('jib-maven-plugin-extension-api') {
   apply plugin: 'java-library'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -21,11 +21,10 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
-project('jib-gradle-plugin-extension-api') {
-  apply plugin: 'java-library'
-}
-project('jib-maven-plugin-extension-api') {
-  apply plugin: 'java-library'
+['jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
+  project(projectName) {
+    apply plugin: 'java-library'
+  }
 }
 
 subprojects {
--- a/jib-build-plan/build.gradle
+++ b/jib-build-plan/build.gradle
@@ -29,9 +29,9 @@ jar {
   }
 }
 
+/* RELEASE */
 configureMavenRelease()
 
-/* RELEASE */
 publishing {
   publications {
     mavenJava(MavenPublication) {
--- a/jib-gradle-plugin-extension-api/build.gradle
+++ b/jib-gradle-plugin-extension-api/build.gradle
@@ -1,6 +1,4 @@
 plugins {
-  // Is this the only option? Can't we do something simiar to Maven Extension API?
-  id 'java-gradle-plugin'
   id 'net.researchgate.release'
   id 'maven-publish'
 }
@@ -8,6 +6,7 @@ plugins {
 dependencies {
   api project(':jib-build-plan')
   api project(':jib-plugins-extension-common')
+  api gradleApi()
 }
 
 jar {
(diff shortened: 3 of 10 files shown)


build.gradle
```diff
@@ -21,6 +21,9 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
+project('jib-gradle-plugin-extension-api') {
+  apply plugin: 'java-library'
+}
```
chanseokoh, line 26: Is there an elegant way to merge this with the Maven block below?

loosebazooka: you could do something like?
```
["jib-gradle-plugin-extension", "jib-maven-plugin-extension"].each { projectName ->
  project(projectName) {
    apply plugin: 'java-library'
  }
}
```
or some variation of this? I think you can call `apply` direction on the `Project` object? Nothing great I can think of.

Maybe in the subprojects section below you can check
```
if(project in [extension-api-projects list]) { 
  apply plugin: 'java-library' 
}
```

jib-gradle-plugin-extension-api/build.gradle
```diff
@@ -0,0 +1,42 @@
+plugins {
+  // Is this the only option? Can't we do something simiar to Maven Extension API?
+  id 'java-gradle-plugin'
```
loosebazooka, line 3: I don't think you need to apply the plugin, I think you can use `gradleApi()` directly in the dependencies block? (that still makes them provided dependencies though)

build.gradle
chanseokoh: The `if` check in the `subprojects` sections didn't work. I can go with the first suggestion.

jib-gradle-plugin-extension-api/build.gradle
chanseokoh: Tried `gradleApi()` in `dependencies{}`, but it doesn't even allow me to see a Gradle `Project` object in `.java` source files.

build.gradle
loosebazooka: Hrmm... this is what worked for me in the end:
```
if ([rootProject.project('jib-maven-plugin-extension-api')].contains(project)) {
  apply plugin: 'java-library'
}
```
but in the end, I mean they're all kinda gross :\

jib-gradle-plugin-extension-api/build.gradle
loosebazooka: That's weird, I feel like it should work.

https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/dsl/DependencyHandler.html#gradleApi--

jib-gradle-plugin-extension-api/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+    attributes 'Bundle-SymbolicName': 'com.google.cloud.tools.jib.gradle.extension'
+    attributes 'Bundle-Name': 'Extension API for Jib Gradle Plugin'
+    attributes 'Bundle-Vendor': 'Google LLC'
+    attributes 'Bundle-DocURL': 'https://github.com/GoogleContainerTools/jib'
+    attributes 'Bundle-License': 'https://www.apache.org/licenses/LICENSE-2.0'
+    attributes 'Export-Package': 'com.google.cloud.tools.jib.gradle.extension'
+  }
+}
+
+configureMavenRelease()
```
loosebazooka, line 27: should this go in the `/* RELEASE */` section?

jib-build-plan/build.gradle
```diff
@@ -29,9 +29,9 @@ jar {
   }
 }
 
+/* RELEASE */
 configureMavenRelease()
```
loosebazooka, line 33: oh haha, didn't realize this was a mistake I created :D

Issue #1618, opened 2019-04-11
Specify the path to the docker config file


msoub:
I'd like to specify the path to my Docker config file, by default it reads it from `${user.home}/.docker/config.json`.

I can see from `CredentialRetrieverFactory#dockerConfig(java.nio.file.Path)` that it's possible, however the documentation does not specify how to do it.

Thank you for your help.

dgageot:
Hi @msoub! Have you tried running mvn with `-Duser.home=[you decide]`?

Something like `mvn -Duser.home=/home jib:build`

chanseokoh:
I believe `-Duser.home=/home` will make Maven cache artifacts under `/home/.m2`. (Correct me if I am wrong.) That is, it changes the user home JVM-wise.

jontro:
I successfully worked around the issue.

Using jenkins docker.withRegistry will store the path in an environment variable. https://jenkins.io/doc/book/pipeline/docker/

Using kotlin dsl syntax:
```
jib {
    var credential: JavaOptional<com.google.cloud.tools.jib.api.Credential> = JavaOptional.empty()
    val path = System.getenv("DOCKER_CONFIG")
    if (path != null) {
        credential = com.google.cloud.tools.jib.registry.credentials.DockerConfigCredentialRetriever("<repository>", Paths.get(path + "/config.json")).retrieve {
            System.out.println(it.message)
        }
    }

    to {
        image = "<image>"
        if (credential.isPresent) {
            auth {
                username = credential.get().username
                password = credential.get().password
            }
        }
    }

}

```

chanseokoh:
@jontro for Jenkins, another way is to use `withCredentials()` to pass credentials through environment variables. (See https://stackoverflow.com/a/59688329/1701388 for more details.)

```
    def dockerImage
    stage('publish docker') {
        withCredentials([usernamePassword(credentialsId: 'myregistry-login', passwordVariable: 'DOCKER_REGISTRY_PWD', usernameVariable: 'DOCKER_REGISTRY_USER')]) {
            // assumes Jib is configured to use the environment variables
            sh "./mvnw -ntp jib:build"
        }
    }
```

jontro:
@chanseokoh Not sure how to use that with ecr since credentials are fetched using the aws api. This plugin https://wiki.jenkins.io/display/JENKINS/Amazon+ECR makes it easy to use with withRegistry

jontro:
[...]

PR #2402, opened 2020-04-14
Respect DOCKER_CONFIG environment variable when finding credentials

TadCordle:
Part of #1618. Maybe we can follow-up with some sort of `jib.docker.config` system property (config parameter might not make sense since this path is platform dependent).

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file.
 
 - Glob pattern support for `jib.extraDirectories.permissions`. ([#1200](https://github.com/GoogleContainerTools/jib/issues/1200))
 - Support for image references with both a tag and a digest. ([#1481](https://github.com/GoogleContainerTools/jib/issues/1481))
-- The `DOCKER_CONFIG` environment variable is now checked for retrieving credentials from the docker config. ([#1618](https://github.com/GoogleContainerTools/jib/issues/1618))
+- The `DOCKER_CONFIG` environment variable specifying the directory containing docker configs is now checked during credential retrieval. ([#1618](https://github.com/GoogleContainerTools/jib/issues/1618))
 
 ### Changed
 
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 
 - Glob pattern support for `jib.extraDirectories.permissions`. ([#1200](https://github.com/GoogleContainerTools/jib/issues/1200))
 - Support for image references with both a tag and a digest. ([#1481](https://github.com/GoogleContainerTools/jib/issues/1481))
+- The `DOCKER_CONFIG` environment variable is now checked for retrieving credentials from the docker config. ([#1618](https://github.com/GoogleContainerTools/jib/issues/1618))
 
 ### Changed
 
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
@@ -41,7 +41,8 @@ import javax.annotation.Nullable;
  *   <li>{@link CredentialRetrieverFactory#dockerCredentialHelper} for a known credential helper, if
  *       set
  *   <li>{@link CredentialRetrieverFactory#known} for known inferred credential, if set
- *   <li>{@link CredentialRetrieverFactory#dockerConfig} for {@code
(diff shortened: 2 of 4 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -166,10 +166,23 @@ public DefaultCredentialRetrievers setCredentialHelper(@Nullable String credenti
...
+    if (dockerConfigEnv != null) {
+      Path dockerConfigsPath = Paths.get(dockerConfigEnv);
+      credentialRetrievers.add(
+          credentialRetrieverFactory.dockerConfig(dockerConfigsPath.resolve(DOCKER_CONFIG_FILE)));
+      credentialRetrievers.add(
+          credentialRetrieverFactory.dockerConfig(
+              dockerConfigsPath.resolve(KUBERNETES_DOCKER_CONFIG_FILE)));
+      credentialRetrievers.add(
+          credentialRetrieverFactory.legacyDockerConfig(
+              dockerConfigsPath.resolve(LEGACY_DOCKER_CONFIG_FILE)));
```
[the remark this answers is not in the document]
TadCordle: I could pull some of this out into a helper method probably, but I think we do want to add all 9 paths to the credential retriever list, since we want to fall back to as many options as possible if retrieving the credentials fails on one source.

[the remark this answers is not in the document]
loosebazooka: oh yeah you're right, I didn't realize we could add them all.

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 
 - Glob pattern support for `jib.extraDirectories.permissions`. ([#1200](https://github.com/GoogleContainerTools/jib/issues/1200))
 - Support for image references with both a tag and a digest. ([#1481](https://github.com/GoogleContainerTools/jib/issues/1481))
+- The `DOCKER_CONFIG` environment variable is now checked for retrieving credentials from the docker config. ([#1618](https://github.com/GoogleContainerTools/jib/issues/1618))
```
chanseokoh, line 10: How about elaborate on this to message that `DOCKER_CONFIG` should be a directory containing a Docker config file?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -166,29 +166,33 @@ public DefaultCredentialRetrievers setCredentialHelper(@Nullable String credenti
...
     }
 
+    String dockerConfigEnv = environment.get("DOCKER_CONFIG");
+    if (dockerConfigEnv != null) {
+      addDockerFiles(credentialRetrievers, Paths.get(dockerConfigEnv));
+    }
+
     String homeProperty = systemProperties.getProperty("user.home");
-    String homeEnvVar = environment.get("HOME");
     if (homeProperty != null) {
```
chanseokoh, line 175: Do you think it's worth avoiding duplicates by checking if `Paths.get(homeProperty).resolve(DOCKER_DIRECTORY)` is same as `Paths.get(dockerConfigEnv)`? Or not? We've been comparing `homeEnvVar` and `homeProperty`. I just want some consistency for future readers of the code who might assume there is a reason for the inconsistency.

TadCordle: I think we can do it. It adds a little extra complexity but it's still straightforward enough.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -166,29 +166,43 @@ public DefaultCredentialRetrievers setCredentialHelper(@Nullable String credenti
...
-          credentialRetrieverFactory.legacyDockerConfig(home.resolve(LEGACY_DOCKER_CONFIG_FILE)));
+      Path homePropertyPath = Paths.get(homeProperty).resolve(DOCKER_DIRECTORY);
+      if (!checkedDockerDirs.contains(homePropertyPath)) {
+        addDockerFiles(credentialRetrievers, homePropertyPath);
+        checkedDockerDirs.add(homePropertyPath);
+      }
     }
+
+    String homeEnvVar = environment.get("HOME");
     if (homeEnvVar != null && !homeEnvVar.equals(homeProperty)) {
```
chanseokoh, line 187: With `checkedDockerDirs`, I think this can be just `if (homeEnvVar) {`?

TadCordle:
@googlebot

PR #2403, opened 2020-04-14
Also try .cmd file extension for credential helpers on Windows

### LINKED ISSUE #2399: Try .cmd file extension suffix for credential helpers on Windows
On Windows, `gcloud auth configure-docker` generates the following section in `~/.docker/config.json` to specify that the cred helper is `docker-credential-gcloud`.

```
"credHelpers": {
    "gcr.io": "gcloud",
    "us.gcr.io": "gcloud",
    "eu.gcr.io": "gcloud",
    "asia.gcr.io": "gcloud",
    "staging-k8s.gcr.io": "gcloud"
},
```

The actual binary to run is `docker-credential-gcloud.cmd`. Without `.cmd`, Jib doesn't seem able to run the binary.

```
[WARNING] The system does not have docker-credential-gcloud CLI
[WARNING]   Caused by: Cannot run program "docker-credential-gcloud": CreateProcess error=2, The system cannot find the file specified
```

`docker-compose` suffered from the same issue, and the workaround was to manually [change `gcloud` to `gcloud.cmd`](https://github.com/docker/compose/issues/5875#issuecomment-383042667) in `~/docker.config.json`. OTOH, apparently, Docker CLI has been [working without `.cmd`](https://github.com/docker/compose/issues/5875#issuecomment-383094189).

chanseokoh:
Fixes #2399.


chanseokoh:
Guess the tests don't work on Kokoro Windows. Will fix them.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
```diff
@@ -88,10 +106,30 @@ public DockerCredentialHelper(String serverUrl, Path credentialHelper) {
   public Credential retrieve()
       throws IOException, CredentialHelperUnhandledServerUrlException,
           CredentialHelperNotFoundException {
+    boolean isWindows =
+        systemProperties.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("windows");
```
briandealwis, line 110: Do you think we might compute this value and pass it into the class?  We're already computing it in `DefaultCredentialRetrievers`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
```diff
@@ -88,10 +106,30 @@ public DockerCredentialHelper(String serverUrl, Path credentialHelper) {
...
+    if (!isWindows || credentialHelper.toString().toLowerCase(Locale.ENGLISH).endsWith(".cmd")) {
+      return retrieve(Arrays.asList(credentialHelper.toString(), "get"));
+    }
+
+    try {
+      return retrieve(Arrays.asList(credentialHelper.toString() + ".cmd", "get"));
```
briandealwis, line 116: If I recall correctly, .exe files have higher precedence over .cmd files.  It would be worth checking as it would simplify this code.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/DefaultCredentialRetrievers.java
```diff
@@ -151,8 +152,11 @@ public DefaultCredentialRetrievers setCredentialHelper(@Nullable String credenti
       // If credential helper contains file separator, treat as path; otherwise treat as suffix
       if (credentialHelper.contains(FileSystems.getDefault().getSeparator())) {
         if (!Files.exists(Paths.get(credentialHelper))) {
-          throw new FileNotFoundException(
-              "Specified credential helper was not found: " + credentialHelper);
+          String osName = systemProperties.getProperty("os.name").toLowerCase(Locale.ENGLISH);
+          if (!osName.contains("windows") || !Files.exists(Paths.get(credentialHelper + ".cmd"))) {
+            throw new FileNotFoundException(
+                "Specified credential helper was not found: " + credentialHelper);
+          }
```
briandealwis, line 159: Would it make sense to push this into `DockerCredentialHelper` or `credentialRetrieverFactory.dockerCredentialHelper()`?

chanseokoh: The logic here is interpreting ahead the given configuration value to determine if it is a full path or a credential helper suffix. This semantics is tightly coupled with how we defined the `<credHelper>` configuration, so it cannot make into jib-core.

Moreover, `DockerCredentialHelper` by design doesn't do any pre-validation (including file checking); it is OK even if the credential helper binary doesn't exist or work when another credential retriever with a higher precedence provides credentials in the end.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
chanseokoh: Other file extensions including `.EXE` just don't work, and I decided to keep it that way until we get a new report.

From https://github.com/GoogleContainerTools/jib/issues/2399#issuecomment-612972912:

> Therefore, it may be necessary to also try `.EXE`, `.COM`, `.BAT`, etc. However, we had not received a report until now, so I think I'll just try `.CMD` first and watch.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
chanseokoh: So I've looked into the chain, and it goes through

1. `DefaultCredentialRetrievers.asList()`
2. `CredentialRetrieverFactory.dockerCredentialHelper(String)`
3. `CredentialRetrieverFactory.dockerCredentialHelper(Path)`
4. `CredentialRetrieverFactory.retrieveFromDockerCredentialHelper()`
5. `DockerCredentialHelper.create()` (which is `DockerCredentialHelper::new`)

It's a long chain. And I think some in the chain don't look reasonable to require passing `isWindows` in their signature. Moreover, this is relatively simple code and can be done in a self-contained manner, I think this is OK.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
chanseokoh: The current PR works, but on second thought, checking `.exe` will give a better error message if Jib can't find the binary. Otherwise, it'll be like "some-helper.exe.cmd not found" instead of "some-helper.exe not found."

PR #2406, opened 2020-04-15
Add disable information to update checker message

TadCordle:


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MojoCommon.java
```diff
@@ -80,6 +80,9 @@ static void finishUpdateChecker(
                       "\u001B[33m"
                           + ProjectInfo.GITHUB_URL
                           + "/blob/master/jib-maven-plugin/CHANGELOG.md\u001B[0m"));
+              projectProperties.log(
```
loosebazooka, line 83: This one has a slightly different format than the one above (two calls to log, rather than a single giant log)? I think it would make sense to have them be consistent?

TadCordle: This is consistent; gradle separates lines with `\n`, but maven needs to make separate log calls since each lines starts with `[INFO]`.

loosebazooka: Cool that makes sense

Issue #2148, opened 2019-11-14
Tried to pull image manifest for registry-1.docker.io/arm64v8/openjdk:8 but failed because: manifest unknown


TuringZhai:
**Environment**:

- *Jib version:1.7.0*
- *Build tool:apache-maven-3.6.1* 
- *OS: arm linux*
```  
OS name: "linux", version: "4.9.140-tegra", arch: "aarch64", family: "unix"
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.2 LTS
Release:        18.04
Codename:       bionic
```

**Description of the issue**:
 Failed to execute goal com.google.cloud.tools:jib-maven-plugin:1.7.0:build (default) on project xxx: Tried to pull image manifest for registry-1.docker.io/arm64v8/openjdk:8 but failed because: manifest unknown

**`jib-maven-plugin` Configuration**: 
```xml
 <plugin>
                <groupId>com.google.cloud.tools</groupId>
                <artifactId>jib-maven-plugin</artifactId>
                <version>1.7.0</version>
                <configuration>
                    <from>
                        <image>arm64v8/openjdk:8</image>
                        <!-- <image>openjdk:alpine</image> -->
                    </from>
                    <container>
                        <mainClass>com.xxx.xxx.xxx</mainClass>
                        <jvmFlags>
                            <jvmFlag>-server</jvmFlag>
                        </jvmFlags>
                        <!--使用该参数将镜像的创建时间与系统时间对其-->
                        <useCurrentTimestamp>true</useCurrentTimestamp>
                    </container>
                    <!--如果私有镜像仓库没有启用https，设置allowInsecureRegistries参数为true-->
                    <allowInsecureRegistries>false</allowInsecureRegistries>
                    <to>
                        <image>docker.io/xxx/${project.artifactId}</image>
                        <tags>
                            <tag>${project.version}</tag>
                        </tags>
                        <auth>
                            <username>xxx</username>
                            <password>xxx</password>
                        </auth>
                    </to>
                </configuration>
                <executions>
                    <execution>
                        <phase>package</phase>
                        <goals>
                            <goal>build</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>

```

**Log output**: 
```
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  52.376 s
[INFO] Finished at: 2019-11-13T16:18:54+08:00
[INFO] ------------------------------------------------------------------------
[...]

PR #2408, opened 2020-04-15
Add entry in faq for manifests lists

loosebazooka:
Add faq entry, so we can link to it for failing builds.

Using example provided in: #2148

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -477,7 +477,7 @@ $ docker manifest inspect openjdk:8
 ```
 
 You can then inspect the output for the specific image you want (in this example an image for the platform:`arm64/linux`)
-```
+```java
 {         
    ...
    // This whole BLOB itself is a manifest list.
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -469,14 +469,14 @@ To inspect the image that is produced from the build using Docker, you can use c
 
 ### How do I specify an image from a manifest list?
 
-By design, Jib will pick always select an image for the platform:`amd64/linux` if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list). If you need to specify a different image from a manifest list you must specify the image digest for the platform you are targeting. 
+By design, if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list), Jib will always select an image for the platform `amd64/linux` . If you need to specify a different image from a manifest list you must specify the digest for the platform you are targeting. 
 
-To determine the digest of an image in a manifest list, you can view the manifest list with:
+To view a manifest, [enable experimental docker cli](https://docs.docker.com/engine/reference/commandline/cli/#experimental-features) features and then run the [manifest inspect](https://docs.docker.com/engine/reference/commandline/manifest_inspect/) command.
 ```
 $ docker manifest inspect openjdk:8
 ```
 
-You can then inspect the output for the specific image you want (in this example an image for the platform:`arm64/linux`)
+You can then inspect the output for the specific image you want (in this example an image for the platform `arm64/linux`)
 ```java
 {         
    ...
@@ -498,7 +498,7 @@ You can then inspect the output for the specific image you want (in this example
 }
 ```
 
-You then use the digest for the target platform you desire in your `jib.to.image` in the form `"opendjdk:8@sha256:1fbd49e3fc5e53154fa93cad15f211112d899a6b0c5dc1e8661d6eb6c18b30a6"`
+You then use the digest for the target platform you desire in your `jib.to.image` in the form `"opendjdk@sha256:1fbd49e3fc5e53154fa93cad15f211112d899a6b0c5dc1e8661d6eb6c18b30a6"` or if you wish to preserve the tag for documentation purposes, you can also use the form `"opendjdk:8@sha256:1fbd49e3fc5e53154fa93cad15f211112d899a6b0c5dc1e8661d6eb6c18b30a6"`.
 
 ## Build Problems
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -28,7 +28,7 @@ If a question you have is not answered below, please [submit an issue](/../../is
 [How can I inspect the image Jib built?](#how-can-i-inspect-the-image-jib-built)\
 [I would like to run my application with a javaagent.](#i-would-like-to-run-my-application-with-a-javaagent)\
 [How can I tag my image with a timestamp?](#how-can-i-tag-my-image-with-a-timestamp)\
-[How do I specify an image from a manifest list?](#how-do-i-specify-an-image-from-a-manifest-list)
+[How do I specify a platform in the manifest list (or OCI index) of a base image?](#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image)
 
 **Build Problems**\
 [How can I diagnose problems pulling or pushing from remote registries?](#how-can-i-diagnose-problems-pulling-or-pushing-from-remote-registries)\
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -466,6 +467,38 @@ Some plugins, such as the [Docker Prepare Gradle Plugin](https://github.com/gcla
 
 To inspect the image that is produced from the build using Docker, you can use commands such as `docker inspect your/image:tag` to view the image configuration, or you can also download the image using `docker save` to manually inspect the container image. Other tools, such as [dive](https://github.com/wagoodman/dive), provide nicer UI to inspect the image.
 
+### How do I specify an image from a manifest list?
+
+By design, Jib will pick always select an image for the platform:`amd64/linux` if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list). If you need to specify a different image from a manifest list you must specify the image digest for the platform you are targeting. 
+
+To determine the digest of an image in a manifest list, you can view the manifest list with:
+```
+$ docker manifest inspect openjdk:8
```
chanseokoh, line 476: The `manifest inspect` command is an experimental one and needs to update `~/.docker/config.json`. Maybe we can add some links:
https://docs.docker.com/engine/reference/commandline/manifest_inspect/ https://docs.docker.com/engine/reference/commandline/cli/#experimental-features

docs/faq.md
```diff
@@ -466,6 +467,38 @@ Some plugins, such as the [Docker Prepare Gradle Plugin](https://github.com/gcla
 
 To inspect the image that is produced from the build using Docker, you can use commands such as `docker inspect your/image:tag` to view the image configuration, or you can also download the image using `docker save` to manually inspect the container image. Other tools, such as [dive](https://github.com/wagoodman/dive), provide nicer UI to inspect the image.
 
+### How do I specify an image from a manifest list?
+
+By design, Jib will pick always select an image for the platform:`amd64/linux` if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list). If you need to specify a different image from a manifest list you must specify the image digest for the platform you are targeting.
```
chanseokoh, line 472: "will pick always select" --> "will always select"
"for the platform:`amd64/linux`" --> "for the platform `amd64/linux`

docs/faq.md
```diff
@@ -27,7 +27,8 @@ If a question you have is not answered below, please [submit an issue](/../../is
 [What would a Dockerfile for a Jib-built image look like?](#what-would-a-dockerfile-for-a-jib-built-image-look-like)\
 [How can I inspect the image Jib built?](#how-can-i-inspect-the-image-jib-built)\
 [I would like to run my application with a javaagent.](#i-would-like-to-run-my-application-with-a-javaagent)\
-[How can I tag my image with a timestamp?](#how-can-i-tag-my-image-with-a-timestamp)
+[How can I tag my image with a timestamp?](#how-can-i-tag-my-image-with-a-timestamp)\
+[How do I specify an image from a manifest list?](#how-do-i-specify-an-image-from-a-manifest-list)
```
chanseokoh, line 31: For the title, how about "a manifest list or an OCI index"? (But we can omit OCI index in the text for brevity.)

And how about "specify OS and architecture of a base image from..."?

docs/faq.md
```diff
@@ -466,6 +467,38 @@ Some plugins, such as the [Docker Prepare Gradle Plugin](https://github.com/gcla
...
+### How do I specify an image from a manifest list?
+
+By design, Jib will pick always select an image for the platform:`amd64/linux` if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list). If you need to specify a different image from a manifest list you must specify the image digest for the platform you are targeting. 
+
+To determine the digest of an image in a manifest list, you can view the manifest list with:
+```
+$ docker manifest inspect openjdk:8
+```
+
+You can then inspect the output for the specific image you want (in this example an image for the platform:`arm64/linux`)
```
chanseokoh, line 479: "for the platform:`amd64/linux`" --> "for the platform `amd64/linux`

docs/faq.md
```diff
@@ -466,6 +467,38 @@ Some plugins, such as the [Docker Prepare Gradle Plugin](https://github.com/gcla
...
+            "architecture": "arm64",
+            "os": "linux",
+            "variant": "v8"
+         }
+      }
+   ]
+}
+```
+
+You then use the digest for the target platform you desire in your `jib.to.image` in the form `"opendjdk:8@sha256:1fbd49e3fc5e53154fa93cad15f211112d899a6b0c5dc1e8661d6eb6c18b30a6"`
```
chanseokoh, line 501: Can we add something like

"As in the example, you can preserve the tag for commenting purposes, but note that the tag will not be used but ignored."

loosebazooka: updated to show both. both are valid?

docs/faq.md
loosebazooka: I'm not really sure about the title.

`How do I specify a platform in the manifest list (or OCI index) of a base image?"

chanseokoh: Yeah, I think "platform" is a good keyword for search.

docs/faq.md
```diff
@@ -466,6 +467,38 @@ Some plugins, such as the [Docker Prepare Gradle Plugin](https://github.com/gcla
 
 To inspect the image that is produced from the build using Docker, you can use commands such as `docker inspect your/image:tag` to view the image configuration, or you can also download the image using `docker save` to manually inspect the container image. Other tools, such as [dive](https://github.com/wagoodman/dive), provide nicer UI to inspect the image.
 
+### How do I specify an image from a manifest list?
+
+By design, if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list), Jib will always select an image for the platform `amd64/linux` . If you need to specify a different image from a manifest list you must specify the digest for the platform you are targeting. 
+
+To view a manifest, [enable experimental docker cli](https://docs.docker.com/engine/reference/commandline/cli/#experimental-features) features and then run the [manifest inspect](https://docs.docker.com/engine/reference/commandline/manifest_inspect/) command.
```
chanseokoh, line 474:
```suggestion
To view a manifest, [enable experimental docker CLI](https://docs.docker.com/engine/reference/commandline/cli/#experimental-features) features and then run the [manifest inspect](https://docs.docker.com/engine/reference/commandline/manifest_inspect/) command.
```

docs/faq.md
```diff
@@ -466,6 +467,38 @@ Some plugins, such as the [Docker Prepare Gradle Plugin](https://github.com/gcla
 
 To inspect the image that is produced from the build using Docker, you can use commands such as `docker inspect your/image:tag` to view the image configuration, or you can also download the image using `docker save` to manually inspect the container image. Other tools, such as [dive](https://github.com/wagoodman/dive), provide nicer UI to inspect the image.
 
+### How do I specify an image from a manifest list?
+
+By design, if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list), Jib will always select an image for the platform `amd64/linux` . If you need to specify a different image from a manifest list you must specify the digest for the platform you are targeting.
```
chanseokoh, line 472:
```suggestion
By design, if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list), Jib will always select an image for the platform `amd64/linux`. If you need to specify a different image from a manifest list you must specify the digest for the platform you are targeting.
```

chanseokoh: We shouldn't forget to update this paragraph once we fix #1567. But I think we will forget, so should we add a link to the issue?

loosebazooka: I added a todo on the bug.

PR #2413, opened 2020-04-17
Ignore SNAPSHOT dependencies of plugin extension sub-projects when publishing

chanseokoh:
We are not yet publishing these artifacts to Maven Central. Will ignore SNAPSHOT versions for now.


chanseokoh:
Oops, wrong branch.

PR #2420, opened 2020-04-20
Add and enable new plugin extension configuration in Maven

chanseokoh:
Adds new extension configuration parameters for Maven. Extensions can additionally get custom properties. The order of executing extensions will be the order defined in `pom.xml`.

```xml
<extensions>
  <extension>
    <implementation>com.google.cloud.tools.jib.plugins.extension.maven.quarkus.QuarkusJibExtension</implementation>
    <properties>
      <customProperty>value</customProperty>
    </properties>
  </extension>
  <extension>
    <implementation>com.example.FooExtension</implementation>
  </extension>
  <!-- Not yet supported
  <extension>com.example.BarExtension</extension>
  -->
</extensions>
```

### DIFF
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
@@ -205,10 +205,8 @@ public class MavenRawConfiguration implements RawConfiguration {
     return jibPluginConfiguration.getImageJsonOutputPath();
   }
 
-  @SuppressWarnings("unchecked")
   @Override
-  public List<ExtensionConfiguration> getPluginExtensions() {
-    return (List<ExtensionConfiguration>)
-        (List<? extends ExtensionConfiguration>) jibPluginConfiguration.getPluginExtensions();
+  public List<? extends ExtensionConfiguration> getPluginExtensions() {
+    return jibPluginConfiguration.getPluginExtensions();
   }
 }
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
@@ -204,4 +204,11 @@ public class MavenRawConfiguration implements RawConfiguration {
   public Path getImageJsonOutputPath() {
     return jibPluginConfiguration.getImageJsonOutputPath();
   }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public List<ExtensionConfiguration> getPluginExtensions() {
+    return (List<ExtensionConfiguration>)
+        (List<? extends ExtensionConfiguration>) jibPluginConfiguration.getPluginExtensions();
+  }
 }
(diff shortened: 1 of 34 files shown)


chanseokoh:
Adds new extension configuration parameters for Maven. Extensions can additionally get custom properties. The order of executing extensions will be the order defined in `pom.xml`.

```xml
<extensions>
  <extension>
    <implementation>com.google.cloud.tools.jib.plugins.extension.maven.quarkus.QuarkusJibExtension</implementation>
    <properties>
      <customProperty>value</customProperty>
    </properties>
  </extension>
  <extension>
    <implementation>com.example.FooExtension</implementation>
  </extension>
  <!-- Not yet supported
  <extension>com.example.BarExtension</extension>
  -->
</extensions>
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
```diff
@@ -30,6 +30,13 @@
  */
 public interface RawConfiguration {
```
loosebazooka, line 31: Just curious? Do we have to modify raw configuration if we're already calling back into ProjectProperties?

chanseokoh: Good question. It's a viable option that I've considered too. There are pros and cons. `ProjectProperties` doesn't have access to configuration values. Therefore, we need to change the constructor of `MavenProjectProperties` and pass extension config values from each MOJO (such as `BuildImageMojo`) where it instantiate `ProjectProperties`. Then we can think of a few ways to pass the information:

- Pass `JibPluginConfiguration`. But this seems weird considering that there's already a higher abstraction of `RawConfiguration`. I don't like the divergence and incoherence in passing around config values. Also too much information and power given to `ProjectProperties`.
- Pass `MavenRawConfiguration`. Requires modifying it anyway. Better than passing `JibPluginConfiguration`, but it still feels like passing `RawConfiguration` in the constructor unnecessarily provides too much information. It only needs the extension config.
- Each MOJO passes only `ExecutionConfiguration`. Seems better than previous options, but I thought perhaps it's better to maintain consistency in obtaining config values from the common interface (`RawConfiguration`) that serves as the source of truth.
- Considerable refactoring as to who runs extensions: don't run extensions in `ProjectProperties`. `runPluginExtensions` don't use any instance values of `ProjectProperties`, so in fact, each MOJO can directly run extensions. The only reason its in `ProjectProperties` is enable the common `PluginConfigurationProcess` to call Maven- and Gradle-specific extension execution routine. If you remember, I briefly mentioned this kind of refactoring a while ago, as it was related to whether I should have a common API base project or not; I eventually decided to have `PluginConfigurationProcess` call the routine instead of each MOJO runs executions, as one of the downsides was a considerable amount of refactoring.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenRawConfiguration.java
```diff
@@ -204,4 +204,11 @@ public Path getImageIdOutputPath() {
   public Path getImageJsonOutputPath() {
     return jibPluginConfiguration.getImageJsonOutputPath();
   }
+
+  @SuppressWarnings("unchecked")
+  @Override
+  public List<ExtensionConfiguration> getPluginExtensions() {
+    return (List<ExtensionConfiguration>)
```
loosebazooka, line 211: This line seems a little crazy to me.

Should we just be using `<? extends ExtensionConfiguration>` everywhere?

PR #2426, opened 2020-04-23
Inherit USER from base image

### LINKED ISSUE #2421: USER not respected / java running as root user
**Environment**:

- jib-maven-plugin, 2.2.0
- maven
- windows

**Description of the issue**:
The java application is run as root user, the dockerfile initializes for a different user.
The application is failing due to the wrong user context, in this specific case the base image is selenium/node-firefox and user is seluser.

**Expected behavior**:
The java application is run as the user specified by the Dockerfile.

**Steps to reproduce**:

  1. use seluser as base image
  2. set the entrypoint to whoami

chanseokoh:
Fixes #2421.

I think we never inherited USER.

### DIFF
--- a/jib-core/CHANGELOG.md
+++ b/jib-core/CHANGELOG.md
@@ -27,6 +27,8 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- Fixes the problem not inheriting `USER` container configuration from a base image. ([#2426](https://github.com/GoogleContainerTools/jib/pull/2426))
+
 ## 0.13.1
 
 ### Fixed
--- a/jib-core/CHANGELOG.md
+++ b/jib-core/CHANGELOG.md
@@ -27,7 +27,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


loosebazooka:
I wonder if some discussion led us to do this? Would this break existing builds that depend on ignoring user?

chanseokoh:
Hmm... good point. I don't remember such a thing though. Let me search.

chanseokoh:
In #595, we had been gradually expanding configurations to inherit. Inheriting `User` was also [considered](https://github.com/GoogleContainerTools/jib/issues/595#issuecomment-440399731), but no further action was taken.

I do think there will be users whose images we will break with this change. `User` has a more significant impact than other configurations. However, I think we should eventually fix this. `Dockerfile` inherits it too.

loosebazooka:
Sure then this should be a 2.3.0?

loosebazooka:
Just curious though, can a user override the inherited user?

chanseokoh:
Yes, with `jib.container.user`. A test in this PR covers that.

PR #2431, opened 2020-04-27
Add and enable new plugin extension configuration in Gradle

chanseokoh:
(Gradle counterpart of #2420.)

Adds new extension configuration parameters for Gradle. Extensions can additionally get custom properties. The order of executing extensions will be the order defined in `pom.xml`.

The following example configures two extensions.
```gradle
  pluginExtensions { pluginExtension { implementation = 'com.example.Test' } }
  pluginExtensions {
    pluginExtension {
      implementation = 'com.google.cloud.tools.jib.plugins.extension.gradle.quarkus.QuarkusJibExtension'
      properties = [customProperty: 'value']
    }
  }
```
~~That is, multiple `pluginExtensions` blocks are needed to configure multiple extensions. I've consulted the [implementation](https://github.com/lacasseio/vscode-gradle-plugin/blob/105504c4908f837b252deaea112120650de1a8fb/src/main/java/io/lacasse/vscode/gradle/internal/tasks/GenerateCompileCommandsFileTask.java#L117-L121) of the [Gradle plugin for Visual Studio Code IDE](https://github.com/lacasseio/vscode-gradle-plugin#gradle-plugin-for-visual-studio-code-ide). I am not totally sure if this is the standard way, but I could not figure out a way to enable configuring extensions as a list like `pluginExtensions = [ { ... }, { ... } ]`.~~ (UPDATE: now [new syntax](https://github.com/GoogleContainerTools/jib/pull/2431#issuecomment-622452336) with latest commits.)

Originally the name of the block was `extensions { ... }`. Although it did work, one of our test classes complained that their's ambiguity in `setExtensions()`, so I went with `pluginExtensions { ... }`. Seems like an acceptable renaming. Probably I'll have to rename the Maven configuration to be consistent.


loosebazooka:
The plural on the base `jib` config of `pluginExtensions` seems weird. I think either
```
jib {
  pluginExtension {..}
  pluginExtension {..}
}
```
or (and this matches most models I see like `licenses` and `developers` in pom config)
```
jib {
  pluginExtensions {
    pluginExtension {...}
  }
}
```

chanseokoh:
> jib {
>   pluginExtensions {
>     pluginExtension {...}
>   }
> }

Trying to find a way to make this work. I'll look into this later.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -81,6 +86,8 @@
   private final SkaffoldParameters skaffold;
   private final Property<Boolean> allowInsecureRegistries;
   private final Property<String> containerizingMode;
+  private final ListProperty<ExtensionParameters> pluginExtensions;
```
TadCordle, line 89: Do we still need this, or are we fine with just `extensionParametersSpec`?

chanseokoh: This is passed to `extensionParametersSpec`, which in turns adds elements to this list.

BTW, I noticed the example in the class comment is out of date. Will update it.

Issue #1581, opened 2019-03-25
ExtraDirectory should allow for specifying a target


alkoclick:
**Description of the issue**:
The extraDirectory object is fairly simple to use but imposes an important constraint (mimicking your intended folder structure under the source). This, in turn, makes it essential to use other plugins/functionality to configure the intended directory structure. This code delays execution, introduces duplication (as you will probably already have the desired artifacts, but not in the desired configuration) and blobs the code with what could be a very pretty one-liner.

**Expected behavior**:
extraDirectories {
  paths + targets: ...
  permissions: ...
}

**Steps to reproduce**: 
N/A

**Environment**: 
All

**Additional Information**: 
I may be able to help out on this PR? 

Thanks for building an awesome tool!

chanseokoh:
Hi @alkoclick,

Thanks for the feedback. Adding `target` (which effectively defaults to `/` now) under `extraDirectory` sounds good to me. Combined with #1020, I think it'll make life easier for many people. It'd be nice if you could help us out, but before that, we need to have an internal discussion about if we really want to do this.

scripnichenko:
Feature required for my case as well!

slowr:
Any update on this @chanseokoh ?

My use case for this feature request is that I want to also deploy some libraries and configurations for tomcat when copying the war file but also some scripts.

For example:
```
[{from: `custom-conf/server.xml`, to: `usr/local/tomcat/conf/server.xml`},
{from: `custom-scripts/*.sh`, to: `usr/local/bin/`}]
```
I can copy them to the base image beforehand but the problem is that I want to be able to update and change them easily instead of recreating the whole base image. I would love to try and create a PR on this if you think this feature is good to have.

slowr:
Thanks a lot for answering!

> If you haven't noticed, the default "to" directory in the image is always `/`, so for now you will have to place your files in the following way.
> 
> ```
> <project root>/src/main/jib/usr/local/tomcat/conf/server.xml
> <project root>/src/main/jib/usr/local/bin/*.sh
> ```

I know about this solution but the problem is that I have multiple projects that share the same configuration files so I can't actually put these files inside each project. If I am missing something that can make this work for multi-project setup please tell me!

chanseokoh:
[...]

PR #2436, opened 2020-04-28
Minor refactoring to prepare for extra directory targets

TadCordle:
Towards #1581. This shouldn't change any functionality, just making changes in plugins-common so I can make the configuration changes in maven/gradle separately.


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleRawConfiguration.java
```diff
@@ -157,8 +159,12 @@ public String getCreationTime() {
   }
 
   @Override
-  public List<Path> getExtraDirectories() {
-    return jibExtension.getExtraDirectories().getPaths();
+  public Map<Path, AbsoluteUnixPath> getExtraDirectories() {
```
chanseokoh, line 162: I think the return type may depend on how we will define the new configuration. I guess one way to translate this is

```gradle
    extraDirectories {
      // not sure if you can use `/`, '\,', or `C:\` as a key name in a map.
      paths = [/path/to/extra/dir: '/target/dir'] // while maintaining backward-compatibility of string list
      permissions = [
        '/path/on/container/file1': 744,
        '/path/on/container/file2': 123
      ]
    }
```
Using a path as a map key probably won't work in Maven either. So I reckon you'll have it like

```gradle
    extraDirectories {
      paths { // directory1
         src = ...
         target = ... 
      }
      paths { // directory2
         src = ...
         target = ... 
      }
    }
```
?

I think returning `Map<Path, AbsoluteUnixPath>` here will still work even if you define a nested structure for `paths`, but I wonder if it is better to define a nested class (such as `RawConfiguration.ExtraDirectoriesConfiguration`) if we are going to have a nested structure anyway.

TadCordle: Hmm, I guess that's true, this could probably be closer to how `AuthProperty` is setup. I guess the question I have is, is `RawConfiguration` actually meant to imply the layout of the maven/gradle configurations, or is it just meant to purely be a translation layer? I just felt a map was simpler; if we add a nested class in `RawConfiguration`, we still need to translate the maven/gradle-specific nested configuration classes into the plugins-common version, so it's the same amount of work plus one extra file. The only up-side I can immediately see is that `RawConfiguration` would more closely resemble the maven/gradle configuration. Is that what we want to stick with?

TadCordle: Using a map also forces people to avoid specifying 2 target directories for the same source directory. Is that something we want to support?

chanseokoh: > I guess the question I have is, is `RawConfiguration` actually meant to imply the layout of the maven/gradle configurations, or is it just meant to purely be a translation layer?

From my point of view, `RawConfiguration` is a higher-level abstraction that doesn't necessarily have to follow the strict configuration layout. It can be independent, as long as it provides "raw config values" in a reasonable way. It's just that, because of the "raw" nature, I think in many cases it'll naturally resemble the configuration layout. The point is, we can choose whatever option that provides a better abstraction that makes sense. For example, returning a `Map` doesn't describe what the keys and the values mean; the caller should have advanced knowledge that the key is the source and the value is the target. In this sense, we could say returning a structured class has an advantage. But I am not so against using `Map`. This is up for discussions.

> Using a map also forces people to avoid specifying 2 target directories for the same source directory. Is that something we want to support?

I don't think we want to support that, at least not at the moment. Anyways, I think we can enforce this constraint when using a structured class too.

helloworlde:
@chanseokoh 
Hi, when the next time release for this optimize, we need this param now, or does there have other solution to copy file to specific path?

chanseokoh:
@helloworlde I can't say for sure about the next release date. Could be as early as next week but may be delayed further. But it'll certainly happen in this month.

Until then, you have to prepare a desired structure ahead before invoking Jib. There are various ways and setups to achieve this, for example, using the Maven Resources Plugin (the [`copy-resources` goal](https://maven.apache.org/plugins/maven-resources-plugin/examples/copy-resources.html)) or the [Copy Rename Maven Plugin](https://coderplus.github.io/copy-rename-maven-plugin/). Some examples: [here](https://github.com/gunnarmorling/quarkus-with-jib/blob/65f0ae7958b1754ec18ffeef89260bfe0ba3e362/pom.xml#L80-L100), and [here](https://github.com/gunnarmorling/quarkus-with-jib/blob/65f0ae7958b1754ec18ffeef89260bfe0ba3e362/pom.xml#L102-L119).

PR #2443, opened 2020-05-01
Use published Jib Plugins Extension dependencies

chanseokoh:
The Build Plan and Extension Common are [published and available](https://repo1.maven.org/maven2/com/google/cloud/tools/).

Once this is merged, we can go ahead and publish the Maven and Gradle Extension APIs. After they are published as well, we can update jib-maven-plugin and jib-gradle-plugin to depend only on published artifacts.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,10 @@ subprojects {
     GUAVA: '28.1-jre',
     JSR305: '3.0.2', // transitively pulled in by GUAVA
 
+    // for Build Plan and Jib Plugins Extension API
+    BUILD_PLAN: '0.1.0',
+    EXTENSION_COMMON: '0.1.0',
+
     COMMONS_COMPRESS: '1.19',
     JACKSON_DATABIND: '2.9.10',
     ASM: '7.3.1',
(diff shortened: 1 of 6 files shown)


chanseokoh:
Seems like the Google Maven Central mirror is lagging behind. I'll rebuild in the afternoon.

jib-maven-plugin-extension-api/build.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
 
 dependencies {
   api project(':jib-build-plan')
```
loosebazooka, line 7: what about build plan?

chanseokoh: ugh, good catch

build.gradle
```diff
@@ -58,6 +58,10 @@ subprojects {
     GUAVA: '28.1-jre',
     JSR305: '3.0.2', // transitively pulled in by GUAVA
 
+    // for Build Plan and Jib Plugins Extension API
+    BUILD_PLAN: '0.1.0',
```
loosebazooka, line 62: I wonder if we can use platform here in the future. Lemme take a look, this is fine for now though.

PR #2452, opened 2020-05-05
Add maven extra directory target configuration

TadCordle:
Maven side of #1581. Adds the following to Maven extra directory configuration:

```
<extraDirectories>
  <paths>
    <path>
      <from>source/directory</from>
      <into>/target/directory</into>
    </path>
    <path>old/shorthand/also/works</path>
  </paths>
</extraDirectories>
```

I totally forgot `set()` was a thing, so this ended up not being as messy as I thought it would be.


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
```diff
@@ -251,7 +252,17 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+            List<Path> pathList = new ArrayList<>();
+            for (Xpp3Dom path : paths.getChildren()) {
+              Xpp3Dom from = path.getChild("from");
+              if (from != null) {
+                pathList.add(Paths.get(from.getValue()));
+              } else {
+                pathList.add(Paths.get(path.getValue()));
+              }
+            }
+            return Collections.unmodifiableList(pathList);
```
chanseokoh, line 265: I think we can remove `unmodifiableList` for consistency. It's weird to sometimes return a modifiable list (line 242) and other times an unmodifiable one.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ExtraDirectory.java
```diff
@@ -0,0 +1,73 @@
...
+  public void setFrom(File from) {
+    this.from = from;
+  }
+
+  String getInto() {
+    return into;
+  }
+
+  @Override
+  public boolean equals(Object other) {
```
chanseokoh, line 58: Are we comparing instances of this class? Where? If not, we should probably remove `equals` and `hashCode`?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ExtraDirectory.java
```diff
@@ -0,0 +1,73 @@
...
+
+package com.google.cloud.tools.jib.maven;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.util.Objects;
+import org.apache.maven.plugins.annotations.Parameter;
+
+/** A bean that configures the source and destination of an extra directory. */
+public class ExtraDirectory {
```
chanseokoh, line 25: Can we name this `ExtraDirectoryParameters` and move it to `JibPluginConfiguration` for consistency?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ExtraDirectory.java
TadCordle: In tests, yes.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/skaffold/FilesMojoV2.java
TadCordle: I errorprone (?) complained about this same inconsistency before I added this. I thought `Collections.singletonList` and `Collectors.toList` were unmodifiable?

chanseokoh: Ah, `singletonList()` is immutable according to the Javadoc. And for `Collectors.toList()`, its Javadoc says,

> There are no guarantees on the type, mutability, serializability, or thread-safety of the List returned

I think we're good.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/ExtraDirectory.java
chanseokoh: Aha. Although it's going to be more lines for test code, I prefer removing these.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MojoCommon.java
```diff
@@ -94,15 +94,25 @@ static void finishUpdateChecker(
    * @param jibPluginConfiguration the build configuration
    * @return the list of resolved extra directories
    */
-  static List<Path> getExtraDirectories(JibPluginConfiguration jibPluginConfiguration) {
-    List<Path> paths = jibPluginConfiguration.getExtraDirectories();
-    if (!paths.isEmpty()) {
-      return paths;
+  static List<JibPluginConfiguration.ExtraDirectoryParameters> getExtraDirectories(
```
chanseokoh, line 97: Not necessary, but `import ... JibPluginConfiguration.ExtraDirectoryParameters` to import the nested type may work?

PR #2457, opened 2020-05-07
rework main class finding

### LINKED ISSUE #2456: Jib infers wrong main class if multiple main classes exist
We got a report that Jib uses the string `${start-class)` when there are multiple `public static void main()` defined. I was able to reproduce this on a Spring Boot project. I think the string `${start-class)` comes from Spring Boot.
```
$ cd examples/spring-boot
$ mvn help:effective-pom | grep start-class
                <mainClass>${start-class}</mainClass>
            <mainClass>${start-class}</mainClass>
                <mainClass>${start-class}</mainClass>
                    <mainClass>${start-class}</mainClass>
                <mainClass>${start-class}</mainClass>
            <mainClass>${start-class}</mainClass>
                  <mainClass>${start-class}</mainClass>
              <mainClass>${start-class}</mainClass>
```
We are [supposed to fail and show an error](https://github.com/GoogleContainerTools/jib/blob/eafbb96b59d05b153d943631be50733baf98754c/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java#L126-L131) if there are multiple main classes, but seems like this isn't working.

A workaround is to explicitly set the main class with `<container><mainClass>`.

UPDATE: found a related issue #170

@reechar-goog @saturnism

chanseokoh:
Fixes #2456.

The logic has changed:

1. main class explicitly configured (via `jib.container.mainClass`) but invalid
   - before: attempted to infer a main class under `src/`.
      - problems and weirdness:
         - never looked into mainClassFromJar despite checking `src/` (inconsistency)
         - configured `jib.container.mainClass` ignored if a valid main is found under `src/', no warning at all
   - after: error

2. main class not configured, invalid mainClassFromJar, and no or multiple mains found under `src/` (i.e., the only main class we could ever get is from the jar-plugin)
   - before: warned mainClassFromJar is invalid, but proceeded with it (this mainClassFromJar used to be called "backup" in our tests)
   - after: error

So, the case in #2456 falls into 2) with multiple mains, and now Jib will fail while printing

`[WARNING] 'mainClass' configured in 'maven-jar-plugin' is not a valid Java class: ${start-class}`

and

`[ERROR] ... Multiple valid main classes were found: hello.Application, hello.Foo, perhaps you should add a 'mainClass' configuration to jib-maven-plugin`

As usual, if only one (valid) main is found under `src/` while mainClassFromJar is invalid, Jib will continue to use the inferred one rather than mainClassFromJar.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.jib.plugins.common;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.api.MainClassFinder;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
 import java.io.IOException;
 import javax.annotation.Nullable;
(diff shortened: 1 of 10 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java
```diff
@@ -127,13 +102,29 @@ private static String findMainClassInClassFiles(ProjectProperties projectPropert
...
   }
 
+  /**
+   * Checks if a string is a valid Java class name.
+   *
+   * @param className the class name to check
+   * @return {@code true} if {@code className} is a valid Java class name; {@code false} otherwise
+   */
+  @VisibleForTesting
+  static boolean isValidJavaClass(String className) {
```
chanseokoh, line 120: Note this is not a new method; nothing changed.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java
```diff
@@ -40,83 +39,59 @@
...
         LogEvent.info(
             "Searching for main class... Add a 'mainClass' configuration to '"
                 + projectProperties.getPluginName()
                 + "' to improve build speed."));
-    return projectProperties.getMainClassFromJar();
-  }
 
-  private static String findMainClassInClassFiles(ProjectProperties projectProperties)
-      throws MainClassInferenceException, IOException {
+    String mainClassFromJar = projectProperties.getMainClassFromJar();
```
loosebazooka, line 71: this confused me for a sec, should we rename this to getMainClassFromJarPlugin?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/MainClassResolver.java
```diff
@@ -40,83 +39,59 @@
...
-        LogEvent.debug(
-            "Could not find a valid main class specified in "
+        LogEvent.info(
+            "Could not find a valid main class from "
                 + projectProperties.getJarPluginName()
-                + "; attempting to infer main class."));
+                + "; looking into all class files to infer main class."));
 
-    MainClassFinder.Result mainClassFinderResult =
+    MainClassFinder.Result result =
```
loosebazooka, line 90: I actually found mainClassFinderResult to be easier to read/follow. But no strong feelings here.

chanseokoh:
Oh, forgot CHANGELOG. Will add.

chanseokoh:
Merging. Feel free to update CHANGELOG afterwards if that doesn't read natural.

PR #2464, opened 2020-05-13
Enable extension-defined custom configuration

chanseokoh:
At last, this worked out really nice. Extensions can now define a custom configuration structure in their taste.

What's even more encouraging is that we can even internally cast the configuration object to the right type and present it to the extension. The extension doesn't need to cast the config object themselves.

## Extension API Changes

The interface is parameterized, and we get the config type from the new method `getExtraConfigType()`.

```diff
-public interface JibMavenPluginExtension extends JibPluginExtension {
+public interface JibMavenPluginExtension<T> extends JibPluginExtension {

+  Optional<Class<T>> getExtraConfigType();

   ContainerBuildPlan extendContainerBuildPlan(
       ContainerBuildPlan buildPlan,
       Map<String, String> properties,
+      Optional<T> extraConfig,
       MavenData mavenData,
       ExtensionLogger logger)
       throws JibPluginExtensionException;
``` 

## Gradle

```gradle
    pluginExtension { implementation = '....AnotherQuarkusJibExtension' }
    pluginExtension {
      implementation = '....AnotherQuarkusJibExtension'
      configuration {
        foo = 'bar'
      }
    }
    pluginExtension { implementation = '....QuarkusJibExtension' }
    pluginExtension {
      implementation = '....QuarkusJibExtension'
      configuration {
        potato = 'fried'
        tomato = 'rotten'
      }
    }
```
Direct assignment (`configuration = ...`) is not supported, but I think that's OK. But I think direct assignment is technically feasible if we wanted to implement that.

```
Running extension: ....AnotherQuarkusJibExtension
  --> config empty
Running extension: ....AnotherQuarkusJibExtension
  --> foo: bar
Running extension: ....QuarkusJibExtension
  --> config empty
Running extension: ....QuarkusJibExtension
  --> potato: fried
  --> tomato: rotten
```

## Maven

```xml
    <pluginExtension>
      <implementation>....AnotherQuarkusJibExtension</implementation>
    </pluginExtension>
    <pluginExtension>
      <implementation>....AnotherQuarkusJibExtension</implementation>
      <configuration implementation="....AnotherExtensionConfig">
        <foo>bar</foo>
      </configuration>
    </pluginExtension>
    <pluginExtension>
      <implementation>....QuarkusJibExtension</implementation>
    </pluginExtension>
    <pluginExtension>
      <implementation>....QuarkusJibExtension</implementation>
      <configuration implementation="....ExtensionConfig">
        <items>
          <item>potato</item>
          <item>tomato</item>
        </items>
      </configuration>
    </pluginExtension>
```

```
[INFO] Running extension: ....AnotherQuarkusJibExtension
[INFO]   --> config empty
[INFO] Running extension: ....AnotherQuarkusJibExtension
[INFO]   --> foo: bar
[INFO] Running extension: ....QuarkusJibExtension
[INFO]   --> config empty
[INFO] Running extension: ....QuarkusJibExtension
[INFO]   --> item: potato
[INFO]   --> item: tomato
```


chanseokoh:
Added tests. Ready for review now.

jib-gradle-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/gradle/extension/JibGradlePluginExtension.java
```diff
@@ -29,13 +30,28 @@
...
+public interface JibGradlePluginExtension<T> extends JibPluginExtension {
+
+  /**
+   * The type of an custom configuration defined by this extension. The configuration object is
+   * mapped from {@code <pluginExtensions><pluginExtension><configuration>}. If the extension does
+   * not wish to define a custom configuration, this method may return {@code Optional#empty()}.
+   *
+   * @return type of an extension-specific custom configuration
+   */
+  default Optional<Class<T>> getExtraConfigType() {
```
loosebazooka, line 42: I always thought default was designed to be used when extending an interface and maintain backward compatibility. It makes sense here conceptually, but I generally would avoid it?

jib-gradle-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/gradle/extension/JibGradlePluginExtension.java
```diff
@@ -44,6 +60,7 @@
   ContainerBuildPlan extendContainerBuildPlan(
       ContainerBuildPlan buildPlan,
       Map<String, String> properties,
+      Optional<T> extraConfig,
```
loosebazooka, line 63: maybe this is just config? (vs extraConfig?) I dunno

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -447,11 +446,46 @@ public JibContainerBuilder runPluginExtensions(
     }
   }
 
-  @Nullable
-  private JibGradlePluginExtension findConfiguredExtension(
-      List<JibGradlePluginExtension> extensions, String extensionClass) {
-    Predicate<JibGradlePluginExtension> matchesClassName =
-        extension -> extension.getClass().getName().equals(extensionClass);
-    return extensions.stream().filter(matchesClassName).findFirst().orElse(null);
+  @SuppressWarnings({"unchecked"})
```
loosebazooka, line 449: can we add a comment to what's unchecked here?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -447,11 +446,46 @@ public JibContainerBuilder runPluginExtensions(
...
+            () -> project,
+            new GradleExtensionLogger(this::log));
+  }
+
+  private JibGradlePluginExtension<?> findConfiguredExtension(
+      List<JibGradlePluginExtension<?>> extensions, ExtensionConfiguration config)
+      throws JibPluginExtensionException {
+    Predicate<JibGradlePluginExtension<?>> matchesClassName =
+        extension -> extension.getClass().getName().equals(config.getExtensionClass());
+    Optional<JibGradlePluginExtension<?>> found =
```
loosebazooka, line 481: is it possible to find multiple or is `findFirst` just to speed things up?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesExtensionTest.java
```diff
@@ -101,23 +117,83 @@ private FooExtensionConfig(Map<String, String> properties) {
...
+            }
+          });
+    }
+  }
+
+  // Not to be confused with Jib's plugin extension config. This class is for an extension-defined
+  // config specific to a third-party extension.
+  private static class ExtensionDefinedFooConfig {
+
+    private String potato;
```
loosebazooka, line 173: as much as I love potato and tomato(and I really really do), maybe these should be more descriptive?
`fooParam` `barParam`

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -580,11 +576,44 @@ public JibContainerBuilder runPluginExtensions(
     }
   }
 
-  @Nullable
-  private JibMavenPluginExtension findConfiguredExtension(
-      List<JibMavenPluginExtension> extensions, String extensionClass) {
-    Predicate<JibMavenPluginExtension> matchesClassName =
-        extension -> extension.getClass().getName().equals(extensionClass);
-    return extensions.stream().filter(matchesClassName).findFirst().orElse(null);
+  @SuppressWarnings({"unchecked"})
```
loosebazooka, line 579: comment on unchecked?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
chanseokoh: It's not possible (or at least it won't happen in practice) to match multiple extensions with the same class name. It means the user has added multiple extensions classes of the same name on the classpath.

jib-gradle-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/gradle/extension/JibGradlePluginExtension.java
chanseokoh: I am fine with `config`. I think one good effect about saying "extra" is to increase the chance of people preferring `properties` in simple cases? I can always change the name.

jib-gradle-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/gradle/extension/JibGradlePluginExtension.java
chanseokoh: Agreed! Also updated comments.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
loosebazooka: Ah yeah, just wondering if we should detect that and tell the user? or just ignore it.

chanseokoh: Alright, I will check if this is possible and see what to do if so.

PR #2474, opened 2020-05-18
Catch runtime exceptions from extension execution

chanseokoh:
This makes it clear to the user that what crashed is a third-party extension, not Jib.


loosebazooka:
Should we allow users to throw an exception?
Should the api include a `throws JibPluginExtensionException` ?

chanseokoh:
It already [includes `throws JibPluginExtensionException`](https://github.com/GoogleContainerTools/jib/blob/27cde263210b9da5b670a37767ef39c4c2807218/jib-maven-plugin-extension-api/src/main/java/com/google/cloud/tools/jib/maven/extension/JibMavenPluginExtension.java#L67).

loosebazooka:
Omg yeah I'm dumb

PR #2479, opened 2020-05-19
Fix potential update checker timestamp file corruption

### LINKED ISSUE #2348: Possible `lastUpdateCheck` timestamp file corruption in UpdateChecker
This looks like a very similar (or the same) problem to #2294. However, this is a fairly minor and low-priority problem and mostly benign, because we auto-generate and overwrite the file every time. It will just be that Jib may do more frequent, unintended update checks.

In [#2335](https://github.com/GoogleContainerTools/jib/issues/2335#issuecomment-601564094), I saw this log:
```
03:03:20.913 [DEBUG] [org.gradle.api.Task] Failed to read lastUpdateCheck; Text '' could not be parsed at index 0
```

I am only guessing, but the message sounds like the file was empty.

We overwrite the file with `Files.write()`, and the JDK code works by opening an `OutputStream`. So I think it can have the [same truncating issue](https://github.com/GoogleContainerTools/jib/issues/2294#issuecomment-586014889) with #2294.

```java
        try (OutputStream out = Files.newOutputStream(path, options)) {
                ...
                out.write(bytes, (len-rem), n);
```

TadCordle:
Fixes #2348. Writes the timestamp to a temp file and does an atomic move (if possible) to avoid interrupting the timestamp file overwrite.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
@@ -48,6 +48,7 @@ import java.util.function.Consumer;
 public class UpdateChecker {
 
   private static final String CONFIG_FILENAME = "config.json";
+  private static final String LAST_UPDATE_CHECK_FILENAME = "lastUpdateCheck";
 
   /** JSON template for the configuration file used to enable/disable update checks. */
   @VisibleForTesting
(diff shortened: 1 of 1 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -240,5 +233,24 @@ private static Path getConfigDir() {
...
+   * Attempts an atomic move first, and falls back to non-atomic if the file system does not support
+   * atomic moves.
+   *
+   * @param source the source path
+   * @param destination the destination path
+   * @throws IOException if the move fails
+   */
+  private static void tryAtomicMove(Path source, Path destination) throws IOException {
+    // Attempts an atomic move first, and falls back to non-atomic if the file system does not
+    // support atomic moves.
```
chanseokoh, line 246: nit: unnecessary comments given the Javadoc

TadCordle: Whoops, thought I deleted this.

TadCordle:
@googlebot

PR #2484, opened 2020-05-21
Maven release v2.3.0

loosebazooka:


jib-maven-plugin/gradle.properties
```diff
@@ -1 +1 @@
-version = 2.2.1-SNAPSHOT
+version = 2.3.1
```
TadCordle, line 1: Shouldn't this have been 2.3.1-SNAPSHOT?

loosebazooka: yeah you're right, good thing for code reviews

PR #2489, opened 2020-05-21
Fix Azure auth failure from not using identity token in docker.config auths section

### LINKED ISSUE #2488: Pushing to Azure Container Registry fails with "401 Unauthorized" even though "docker push" works
Even though `docker push francium25.azurecr.io/hello-world` works, Jib cannot push to Azure Container Registry. Note Jib does retrieve Azure credentials from `~/.docker/config.json`.
```
[INFO] Using credentials from Docker config (.../.docker/config.json) for francium25.azurecr.io/hello-world
```

Notably, `~/docker/config.json` has `identitytoken` that other `auths` entries don't have. I wonder if this is basically the same issue we've taken care of when using Azure credential helpers (#1490).
```json
    "auths": {
       "francium25.azurecr.io": {
            "auth": "...",
            "identitytoken": "..."
        },
        "quay.io": {
            "auth": "..."
        },
```

Full log:
```
[INFO] Containerizing application to francium25.azurecr.io/hello-world...
[WARNING] Base image 'gcr.io/distroless/java:8' does not use a specific image digest - build may not be reproducible
[INFO] Using credentials from Docker config (.../.docker/config.json) for francium25.azurecr.io/hello-world
[INFO] Using base image with digest: sha256:f59b26c5ecc735514a38afbf845214383c4e2ba1fdd15a76225339b8ab7da8ef
[INFO] Executing tasks:
[INFO] [=======                       ] 22.2% complete
[INFO] > pulling base image manifest
[INFO] 
[INFO] BUILD FAILURE
[INFO] Total time:  1.530 s
[INFO] Finished at: 2020-05-21T13:50:02-04:00
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:2.2.0:build (default-cli) on project helloworld: Build image failed, perhaps you should make sure your credentials for 'francium25.azurecr.io/hello-world' are set up correctly. See https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#what-should-i-do-when-the-registry-responds-with-unauthorized for help: Unauthorized for francium25.azurecr.io/hello-world: 401 Unauthorized -> [Help 1]
```

chanseokoh:
Fixes #2488.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
@@ -76,10 +76,10 @@ class DockerConfig {
    *     none exists
    */
   @Nullable
-  String getAuthFor(String registry) {
+  AuthTemplate getAuthFor(String registry) {
     Map.Entry<String, AuthTemplate> authEntry =
         findFirstInMapByKey(dockerConfigTemplate.getAuths(), getRegistryMatchersFor(registry));
-    return authEntry != null ? authEntry.getValue().getAuth() : null;
+    return authEntry != null ? authEntry.getValue() : null;
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
@@ -17,11 +17,11 @@
 package com.google.cloud.tools.jib.registry.credentials;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.client.util.Base64;
 import com.google.cloud.tools.jib.api.Credential;
 import com.google.cloud.tools.jib.api.LogEvent;
-import com.google.cloud.tools.jib.json.JsonTemplateMapper;
 import com.google.cloud.tools.jib.registry.RegistryAliasGroup;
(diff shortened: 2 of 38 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
```diff
@@ -76,10 +76,10 @@
    *     none exists
    */
   @Nullable
-  String getAuthFor(String registry) {
+  AuthTemplate getAuthFor(String registry) {
```
loosebazooka, line 79: should this be optional?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java
```diff
@@ -138,15 +139,25 @@ public Path getDockerConfigFile() {
       }
 
       // Lastly, find defined auth.
-      String auth = dockerConfig.getAuthFor(registryAlias);
+      AuthTemplate auth = dockerConfig.getAuthFor(registryAlias);
       if (auth != null) {
```
loosebazooka, line 143: then we check if present here?

chanseokoh:
> Is this really only an issue with azure? Can no one else use identity tokens?

Maybe not only an issue with Azure. But because we've not heard any error report, I think it's safer to keep using `auth` when both `auth` and `identityToken` are given and try to keep the current behavior as much as possible. That is, I'm just fixing the error in this Azure case.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfig.java
chanseokoh: Then I'd take out all `@Nullable`s from other methods in this class for consistency, but it becomes a bit involves. Besides, I think it's not that critical to migrate to Optional for this method, as it is not public.

Issue #1955, opened 2019-09-05
Consider configuring permissions and owners for files not covered in <extraDirectories>


chanseokoh:
Permissions ~~and owners~~ are configurable for files added from `<extraDirectories>`. For other files put by Jib, they are fixed (755 for directories and 644 for files, owned by root).

I think it is possible to modify permissions and owners for existing *directories* in an base image but not really possible for existing *files* unless we can somehow extract and duplicate the files in our layers? (Correct me if I'm wrong.) Knowing that, how would we know if a configured path is a directory or a file? Perhaps this configuration should apply only to the files put by Jib. Most of the time people won't know the details of which files Jib puts where, but at least people using `<appRoot>` in WAR deployment may find this useful. This WAR case is actually #1257.

That said, #1257 is a subset of this issue. The frowned workaround in #1257 actually tweaks `<extraDirectories>` to modify directory permissions.

chanseokoh:
Forgot to update here that the ownership is configurable through an extension. From https://github.com/GoogleContainerTools/jib/issues/1257#issuecomment-642889893:

---

The [Jib Extension Framework](https://github.com/GoogleContainerTools/jib-extensions) is available with the latest Jib versions. You can easily extend and tailor the Jib plugins behavior to your liking.

Since there's a very compelling reason to allow the container to be mutable at runtime by changing file ownership (e.g., running Skaffold to update class files on Kubernetes during development), we've decided to support this use case through an extension. However, changing ownership is still discouraged unless absolutely needed.
- [Maven Ownership Extension](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-ownership-extension-maven)
- [Gradle Ownership Extension]( https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-ownership-extension-gradle)

For general information about using and writing extensions, take a look at the [Jib Extensions repo](https://github.com/GoogleContainerTools/jib-extensions).

chanseokoh:
I don't remember what was the motivation for changing permissions for files put by Jib but not coming from `<extraDirectories>`. But I cannot really think of a good reason or a use case to support it at the moment, especially give than the user can now change ownership using the extension.

I am closing this issue, but if you want this feature for an interesting use case, please leave feedback.

rmannibucau:
[...]

Issue #1270, opened 2018-11-21
Unnecessary overwrites of directories, resulting changes to owners and permissions


chanseokoh:
The way we took to fix #727 and #523 was to add TAR archive entries to enumerate all parent directories of any given file and explicitly set permissions for those parent directories: PR #891. The consequence is that owner:group is always reset to `root:root` and the permission mode 755 for all the directories involved.

So, let's say I have a base Tomcat image where the owner:group and the permission of `/usr` are 12345:54321 and 750:
```
$ docker run --rm -it --entrypoint ls francium25/base-tomcat -ld /usr
drwxr-x---    1 12345    54321         4096 Sep 12 20:50 /usr
```
Now, if Jib builds an image where `/usr` is one of the parent directories of my `<appRoot>`,
```xml
          <from><image>francium25/base-tomcat</image></from>
          <container>
            <appRoot>/usr/local/tomcat/webapps/ROOT</appRoot>
          </container>
```
the owner and the permission are reset (to the values we chose).
```
$ docker run --rm -it --entrypoint ls my-image -ld /usr
drwxr-xr-x    1 root     root          4096 Jan  1  1970 /usr
```

It will be ideal to not touch the owners and permissions of existing directories.

Also note that `<appRoot>` and its parent directories will always be `root:root` and 755.

chanseokoh:
But I understand the way we reset the permissions, which resolves #727 and #523. So I'd say this issue is a low priority that we may not fix until there is a real need.

unautre:
Hello,
I think I just encountered this problem: I am building a docker image for a war on a custom wildfly image ; the app is put in the /opt/jboss/wildfly/standalone/deployments/ folder, and thus chown/chmoding all parent folders to root:root/755... The final image is not able to start as non-root.

Is there any change this bug might end up corrected, despite being low priority ? Or is there any workaround ?

loosebazooka:
If you control you base image, maybe you can control this using some sort of symlink? Are you using a standard jboss image?

unautre:
Maybe I misunderstood your comment, how would symlinks help with permissions ?

I am not using a standard jboss image, but close (we rebuilt it with minor tweaks).

A possible workaround would be pushing the exploded war somewhere else (appRoot configuration), and copy it at container start time using a custom entrypoint ; then I would be able to control permissions better.

loosebazooka:
Yeah either symlinking the directory, or a war file?

unautre:
[...]

Issue #1257, opened 2018-11-18
`appRoot` and sub dirs / files owned by root


mlbiam:
**Description of the issue**:

The resulting image has all files in `appRoot` owned by root, even when the `user` parameter is set in the configuration.  While my own container appears to work, it would be better to not have these files owned by root.  It could cause unknown results on restrictive container platforms (like OpenShift).

**Expected behavior**:

If the `user` option is set, then all the files created by the build should be owned by that user id/name for consistency.

**Steps to reproduce**: 

```
<plugin>
        <groupId>com.google.cloud.tools</groupId>
        <artifactId>jib-maven-plugin</artifactId>
        <version>0.10.0</version>
        <configuration>
          <from>
            <image>mlbiam/openunison-jib-builder:latest</image>
          </from>
          <container>
            <user>431</user>
            <appRoot>/usr/local/openunison/work/webapp</appRoot>
            <entrypoint>/usr/local/openunison/bin/run_openunison.sh</entrypoint>
          </container>
          <to>
            <image>mlbiam/testadlogin</image>
          </to>
        </configuration>
      </plugin>
```

run a build and look at `/usr/local/openunison/work/webapp` all the files are owned by root:

```
openunison@openunison-7b7b6df97b-bks92:/$ cd /usr/local/openunison/work/webapp
openunison@openunison-7b7b6df97b-bks92:~/work/webapp$ ls -lh
total 24K
drwxr-xr-x  2 root root 4.0K Jan  1  1970 META-INF
drwxr-xr-x  1 root root 4.0K Jan  1  1970 WEB-INF
drwxr-xr-x  4 root root 4.0K Jan  1  1970 auth
-rw-r--r--  1 root root 1.2K Jan  1  1970 favicon.ico
drwxr-xr-x 10 root root 4.0K Jan  1  1970 k8stoken
drwxr-xr-x 10 root root 4.0K Jan  1  1970 scale
openunison@openunison-7b7b6df97b-bks92:~/work/webapp$ id
uid=431(openunison) gid=433(openunison) groups=433(openunison)
```

**Environment**: 

```
Apache Maven 3.5.4 (1edded0938998edf8bf061f1ceb3cfdeccf443fe; 2018-06-17T14:33:14-04:00)
Maven home: /usr/local/Cellar/maven/3.5.4/libexec
[...]

PR #2494, opened 2020-05-26
Update Build Plan API to accept file ownership in layers

chanseokoh:
Towards writing an extension for #1257 and #1955, as well as a workaround for #1270.

Once merged, we need to re-publish all the libraries one after another.

### DIFF
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
@@ -170,8 +170,9 @@ public class FileEntriesLayer implements LayerObject {
      *     sourceFile}
      * @param permissions the file permissions on the container
      * @param modificationTime the file modification time
-     * @param ownership file ownership. For example, "0", "1234", "user", ":0", ":5678", ":group",
-     *     "0:0", "1234:5678", and "user:group".
+     * @param ownership file ownership. For example, "1234", "user", ":5678", ":group", "1234:5678",
+     *     and "user:group". Note that "" (empty string), ":" (single colon), "0:", ":0" are allowed
+     *     and representative of "0:0" or "root:root", but prefer an empty string for "0:0".
      * @return this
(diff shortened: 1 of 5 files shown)


jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntry.java
```diff
@@ -121,12 +158,13 @@ public boolean equals(Object other) {
     FileEntry otherFileEntry = (FileEntry) other;
     return sourceFile.equals(otherFileEntry.sourceFile)
         && extractionPath.equals(otherFileEntry.extractionPath)
-        && Objects.equals(permissions, otherFileEntry.permissions)
-        && Objects.equals(modificationTime, otherFileEntry.modificationTime);
```
chanseokoh, line 125: Perhaps `Objects.equals` was necessary in the past when `permissions` and `modificationTime` were `@Nullable`. The `java.util.Objects.equals` is implemented like the following, so we don't need it.

```java
    public static boolean equals(Object a, Object b) {
        return (a == b) || (a != null && a.equals(b));
    }
```

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
```diff
@@ -265,6 +326,10 @@ public FileEntriesLayer build() {
       DEFAULT_MODIFICATION_TIME_PROVIDER =
           (sourcePath, destinationPath) -> DEFAULT_MODIFICATION_TIME;
 
+  /** Provider that returns default file ownership ("0:0"). */
```
loosebazooka, line 329: the comment and the return value don't really line up?

chanseokoh: Fixed. Also updated the Build Plan Spec doc with the last commit.

chanseokoh:
Merging this now to start releasing early.

PR #2503, opened 2020-06-01
Add JDK12 to Travis

gliptak:
https://github.com/GoogleContainerTools/jib/pull/2351

PS Higher JDKs fail as follows

https://travis-ci.com/github/gliptak/jib/jobs/342132137

I'm not clear what update needs to happen to correct

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ matrix:
       dist: trusty # oraclejdk8 absent in Xenial
     - name: OpenJDK 11
       jdk: openjdk11
+    - name: OpenJDK 12
+      jdk: openjdk12
+
 env:
   global:
   - TERM=dumb         # for gradle
--- a/build.gradle
+++ b/build.gradle
@@ -104,7 +104,7 @@ subprojects {
   dependencies {
     // NullAway errorprone plugin
     annotationProcessor 'com.uber.nullaway:nullaway:0.7.9'
-    errorprone 'com.google.errorprone:error_prone_core:2.3.4'
+    errorprone 'com.google.errorprone:error_prone_core:2.4.0'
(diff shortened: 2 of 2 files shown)


chanseokoh:
Found this: https://jaxenter.com/gradle-6-0-release-java-13-support-164160.html

Just for the sake of testing, can you try Gradle 6.0? Not that we will upgrade to Gradle 6.0. If Gradle 6.0 works, I guess adding a JDK 13 (or JDK 14) build should wait.

chanseokoh:
@loosebazooka and we probably won't add a JDK 12 build now, right?

chanseokoh:
But I guess we could dynamically upgrade the Gradle wrapper only in a Travis build (i.e., stick to 5.6.1 in the source repo but upgrade Gradle only in JDK13/14 Travis builds) and have the the 13/14 builds non-blocking?

loosebazooka:
I mean I guess there's some value in checking if "latest" jdk works, but yeah I agree there's no reason to support 12. We could have a test for 14 for "latest" (14.0.1 or whatever)

loosebazooka:
A per our non-guaranteed, undocumented and internal policy for old gradle support: gradle 6.0 upgrade might only happen towards the end of 2020 (November?... a year after the last 5.X release)

chanseokoh:
@gliptak thanks for trying this, but I'm afraid this should wait.

- JDK12: not interested in adding JDK12. We'd rather add one build for the current latest JDK version.
- JDK13 or 14: requires Gradle 6+. However, due to Gradle API changes, our codebase needs to adopt, which will then make Gradle 5.2.1 unable to compile. I think it's not worth maintaining two separate branches.

PR #2504, opened 2020-06-01
Bring JDK current in Travis

gliptak:
```Gradle 6.0.1``` breaks ```build.gradle``` "inheritance"

https://travis-ci.com/github/gliptak/jib/jobs/342161039#L492

https://github.com/GoogleContainerTools/jib/pull/2503

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,14 @@ matrix:
       dist: trusty # oraclejdk8 absent in Xenial
     - name: OpenJDK 11
       jdk: openjdk11
+    - name: OpenJDK 12
+      jdk: openjdk12
+    - name: OpenJDK 13
+      jdk: openjdk13
+    - name: OpenJDK 14
+      jdk: openjdk14
+    - name: OpenJDK 15
+      jdk: openjdk15
 env:
   global:
   - TERM=dumb         # for gradle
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by chanseokoh** [comment]
```
* Where:
Build file '/home/travis/build/gliptak/jib/jib-gradle-plugin/build.gradle' line: 2
* What went wrong:
An exception occurred applying plugin request [id: 'java-gradle-plugin']
> Failed to apply plugin [class 'org.gradle.api.plugins.JavaLibraryPlugin']
   > Cannot create variant 'classes' after configuration ':jib-gradle-plugin:apiElements' has been resolved
```

I think `:apiElements` is from the [Java Library plugin](https://docs.gradle.org/current/userguide/java_library_plugin.html) (`java-library`). And the error seems to occur when applying `java-grdle-plugin`, whose [doc](https://docs.gradle.org/current/userguide/java_gradle_plugin.html) says "[i]t automatically applies the Java Library plugin." The error message does seem to indicate that `java-gradle-plugin` is automatically applying `java-library`. But what's wrong? I'm not a Gradle expert, so not sure what's going on.

chanseokoh:
I tried experimenting JDK 14. Unfortunately, we have integration tests where we test old Gradle versions (5.1 and 5.2, which is the minimum Gradle version we support), and these old versions do not run on JDK 13 or 14. I think building with JDK 13 or 14 should wait.

Thank you for trying this anyway!

PR #2508, opened 2020-06-03
Inject Gradle project when creating extension-specific config object

chanseokoh:
I realized we [need to inject the Gradle project](https://github.com/GoogleContainerTools/jib-extensions/pull/7/files#diff-bad2ecb1f14b7db185ced1687bb8cfa6R108-R111) when instantiating an extension-specific config object.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
@@ -469,7 +469,7 @@ public class GradleProjectProperties implements ProjectProperties {
         // config.getExtraConfiguration().get() is of type Action, so this cast always succeeds.
         // (Note generic <T> is erased at runtime.)
         Action<T> action = (Action<T>) config.getExtraConfiguration().get();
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -469,7 +469,7 @@ public JibContainerBuilder runPluginExtensions(
         // config.getExtraConfiguration().get() is of type Action, so this cast always succeeds.
         // (Note generic <T> is erased at runtime.)
         Action<T> action = (Action<T>) config.getExtraConfiguration().get();
-        extraConfig = project.getObjects().newInstance(extraConfigType.get());
+        extraConfig = project.getObjects().newInstance(extraConfigType.get(), project);
```
loosebazooka, line 472: I thought this was only if you wanted to inject the project into the extension block?

loosebazooka: I see, lets just keep the constructor with Project and document it. Having 2 options will probably confuse people?

PR #2509, opened 2020-06-03
Make jib-core expose jib-build-plan

### LINKED ISSUE #2507: Build Plan API is not transitively exposed to Jib Core
I had to explicitly add `jib-build-plan` as a dependency. I think we should use `api` instead of `implementation` in jib-core's `build.gradle` [here](https://github.com/GoogleContainerTools/jib/blob/d8c5a2128db0e3d5a2b042bb1c10399a264d8672/jib-core/build.gradle#L8).

```xml
    <dependency>
      <artifactId>jib-build-plan</artifactId>
      <version>0.2.0</version>
    </dependency>
    <dependency>
      <artifactId>jib-core</artifactId>
      <version>0.14.0</version>
    </dependency>
```

chanseokoh:
Fixes #2507.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
-['jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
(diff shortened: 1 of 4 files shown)


chanseokoh:
Ah, forgot CHANGELOG. Will add.

PR #2510, opened 2020-06-03
Resolve error prone warnings

chanseokoh:
Also suppressed [InlineFormatString](https://errorprone.info/bugpattern/InlineFormatString).


jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
```diff
@@ -157,6 +157,7 @@ public void testAllFileAndDirectories() throws IOException {
             Assert.assertTrue(layerEntry.isFile() || layerEntry.isDirectory()));
   }
 
+  @SuppressWarnings("JdkObsolete") // use of Date from TarArchiveEntry
```
loosebazooka, line 160: is this unavoidable?

build.gradle
```diff
@@ -117,6 +117,7 @@ subprojects {
     if (!name.toLowerCase().contains('test')) {
       options.errorprone {
         check('NullAway', CheckSeverity.ERROR)
+        check('InlineFormatString', CheckSeverity.OFF)
```
loosebazooka, line 120: Do we just have way too many instances of this?

loosebazooka (dismissed):

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
chanseokoh: Yeah, even the latest Apache Commons Compress returns `Date` when calling [`TarArchiveEntry.getLastModifiedDate()`](https://commons.apache.org/proper/commons-compress/javadocs/api-release/org/apache/commons/compress/archivers/tar/TarArchiveEntry.html#getLastModifiedDate--) or [`TarArchiveEntry.getModDate()`](https://commons.apache.org/proper/commons-compress/javadocs/api-release/org/apache/commons/compress/archivers/tar/TarArchiveEntry.html#getModTime--). They are the only ways to get the file modification time.

build.gradle
chanseokoh: Not too many, but the problem was that, strangely, I was not able to make `@SuppressWarnings("InlineFormatString")` work [here](https://github.com/GoogleContainerTools/jib/blob/307364160f3189e3079f7f3820be673f5d42e7ed/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JibBuildRunner.java#L49-L62). Adding the annotation right before the constant fields or before the class didn't work. Not sure why.

Another nuisance was that it forces to remove the constants on the `*Step` classes (like [this one](https://github.com/GoogleContainerTools/jib/blob/307364160f3189e3079f7f3820be673f5d42e7ed/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java#L36)). The check complains only when the string is for a format string (e.g., containing `%s`), but for consistency, we'll have to remove non-format string constants as well.

PascalSchumacher: >but the problem was that, strangely, I was not able to make @SuppressWarnings("InlineFormatString") work here

I guess that was https://github.com/google/error-prone/issues/1650

chanseokoh: @PascalSchumacher ah, a bug in 2.4.0. Thanks for letting us know!

chanseokoh: I'll hold off merging whit until they release a new version that fixes the `@SuppressWarnings("InlineFormatString")` issue.

chanseokoh:
Now that we upgrade Error Prone to 2.5.1 in #3005, I re-purposed this PR to fix warnings from the latest version.

loosebazooka (approved):

PR #2513, opened 2020-06-04
Ports.parse in api should return Set, not ImmutableSet

### LINKED ISSUE #2512: Ports in api package should not expose api with ImmutableSet from guava
https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/api/Ports.java#L48

return type should be `Set`

loosebazooka:
fixes #2512

### DIFF
--- a/jib-core/CHANGELOG.md
+++ b/jib-core/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
-- `Ports.parse(List<String> ports)` in api now returns `Set` instead of `ImmutableSet` ([#2513](https://github.com/GoogleContainerTools/jib/pull/2513))
+- `Ports.parse(List<String> ports)` now returns `Set` instead of `ImmutableSet` ([#2513](https://github.com/GoogleContainerTools/jib/pull/2513))
 - Previous locally cached application layers will be ignored because of changes to the caching selectors. ([#2499](https://github.com/GoogleContainerTools/jib/pull/2499))
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


jib-core/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- `Ports.parse(List<String> ports)` in api now returns `Set` instead of `ImmutableSet` ([#2513](https://github.com/GoogleContainerTools/jib/pull/2513))
```
chanseokoh, line 12:
```suggestion
- `Ports.parse(List<String> ports)` now returns `Set` instead of `ImmutableSet` ([#2513](https://github.com/GoogleContainerTools/jib/pull/2513))
```

jib-core/src/main/java/com/google/cloud/tools/jib/api/Ports.java
```diff
@@ -45,7 +46,7 @@
    * @return the ports as a list of {@link Port}
    * @throws NumberFormatException if any of the ports are in an invalid format or out of range
    */
-  public static ImmutableSet<Port> parse(List<String> ports) throws NumberFormatException {
+  public static Set<Port> parse(List<String> ports) throws NumberFormatException {
     ImmutableSet.Builder<Port> result = new ImmutableSet.Builder<>();
```
chanseokoh, line 50: I think this should return a mutable set as an API.

loosebazooka: Hrmm... I guess this isn't really a value class. Makes sense, I'll change the underlying implementation.

PR #2515, opened 2020-06-08
Proposal for jib cli buildfile

loosebazooka:


proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+# global file properties applied to all layers
+fileProperties:
+  filePermissions: "644"
+  directoryPermissions: "755"
+  user: "0"
+  group: "0"
+  timestamp: "0"
+layers:
+  - name: "scripts and classes"
+    # file properties only applied to layer this layer "scripts and classes"
```
chanseokoh, line 48:
```suggestion
    # file properties only applied to this layer "scripts and classes"
```

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+- All properties in `Copy` are applied first
+- Any property not in `Copy` are applied from `Layer`
+- Any property not in `Copy` or `Layer` are applied from `Global`
+- Any property not defined anywhere use jib system defaults.
+
+
+## Extented features (not included in this build)
+
+### Other time options
+* `actual`: use timestamp from file on disk
```
chanseokoh, line 107: TBH, I don't like this option. The Jib plugins deliberately exclude this option, as this completely nullifies caching and reproducibility almost all the time.

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+* `now`: use time of build
+
+### Templating
+Allow passing of values into build to be replaced in buildfile
```
chanseokoh, line 111: As we discussed, I imagine eventually we will end up implementing this, but I think we would keep this as a low priority for some time. I guess there are numerous templating solutions, and I think people will have no issue using them with a script considering that this is a CLI tool.

BTW, templating is like a feature that tools like Jib CLI can support independent of this spec; it doesn't need to be part of this buildfile spec.

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+  - "myjar.jar"
+
+# global file properties applied to all layers
+fileProperties:
+  filePermissions: "644"
+  directoryPermissions: "755"
+  user: "0"
+  group: "0"
+  timestamp: "0"
+layers:
```
chanseokoh, line 46: The build plan allows adding a layer archive tarball. Perhaps Jib CLI can support that too?

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+
+Specification for a buildfile describing building a container image. This buildfile can be
+used by the jib-cli to generate a container. It is translated directly into a buildplan and
+passed to the builder.
+
+```
+# "FROM" with option to use os/architecture for manifest lists
+baseImage: "gcr.io/distroless/java:8"
+osHint: linux
+architectureHint: amd64
```
briandealwis, line 11: Why `Hint`?  It could signify something optional that could be ignored?  Why not `os` and `arch`?

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
+# Jib CLI Buildfile Specification
+
+Specification for a buildfile describing building a container image. This buildfile can be
+used by the jib-cli to generate a container. It is translated directly into a buildplan and
+passed to the builder.
```
briandealwis, line 5: Is this YAML?  Should be explicit.

proposals/buildfile.md
briandealwis: There are use-cases for this.  It shouldn't be the default, and these reasons can be expressed.  But we shouldn't prevent it.

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+      - from: "build/other"
+        to: "/app"
+```
+
+## Layers
+
+`layers` are a list of layer directives, each directive consisting of 3 parts
+* `name`: the name of the layer (metadata)
+* `filesProperites`: on disk metadata for all files in the layer
+* `files`: a list of copy directies
```
briandealwis, line 81:
```suggestion
* `files`: a list of _copy directives_ (see below)
```

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+fileProperties:
+  filePermissions: "644"
+  directoryPermissions: "755"
+  user: "0"
+  group: "0"
+  timestamp: "0"
+layers:
+  - name: "scripts and classes"
+    # file properties only applied to layer this layer "scripts and classes"
+    fileProperties:
```
briandealwis, line 49: Seems odd to have "fileProperties" contain "directoryPermissions"?  Could this be named "properties"?

proposals/buildfile.md
```diff
@@ -0,0 +1,118 @@
...
+* `fileproperties`: on disk metadata for all files in this copy directive
+
+### FileProperties
+
+File properties are available at 3 levels
+* Global: applies to all layers
+* Layer: applies to a single layer
+* Copy: applies to a single copy directive
+
+Properties can be defined at any level and are applied in a cascading fashion.
```
briandealwis, line 97: Might it be clearer to say that each property (permission, owner, group, timestamp) is resolved in the following order?

proposals/buildfile.md
chanseokoh: Ah, yeah, I remembered our last discussion on this. Let's keep this.

proposals/buildfile.md
loosebazooka: yeah i'm looking into how to support that parsing with the yaml parser I'm using

proposals/buildfile.md
loosebazooka: I just copied this part from the buildplan. @chanseokoh why did we use *hint* there?

proposals/buildfile.md
loosebazooka: yeah that was weird. but directories are files no? I don't know exactly how this part should be named.

proposals/buildfile.md
chanseokoh: The user must provide a value to this field when the given base image is a manifest list. However, when the base image is an image, a tool must inherit the value from the base image while ignoring this field (or the tool may decided to throw an error).

As a general spec describing a plan to build an image in a declarative manner, I wanted to avoid possible confusion that someone mistakenly assumes that by specifying `os: arm64`, they will create an ARM64 image. Before you can actually check the base image reference at runtime, you won't know the final outcome.

briandealwis: > However, when the base image is an image, a tool must inherit the value from the base image while ignoring this field (or the tool may decided to throw an error).

Where is this requirement come from?
 
Given that the os/arch is normally inherited, most users should never need to use this unless they're either creating a scratch image or if they mean to override the settings from a different image.  There are (were?) some ARM images on Docker Hub that identified as being amd64.  I believe distroless/base has no actual machine-specific code.

chanseokoh: > if they mean to override the settings from a different image.

Someone doing this doesn't make sense to me. It means an image has a wrong metadata, which should be fixed. I don't know if some Docker Hub images had wrong image metadata, but that's an error from the image creator.

In any case, this is a spec, and for this part, I think it doesn't have to mandate a certain interpretation in an implementation. A tool may provide an option like `--force-os-arch` (which I think is really unnecessary in practice.) But for the reason of the potential confusion I explained, I think `osHint` (or something else if there's a better term than `Hint`) is better than `os`

loosebazooka: Yeah I agree that correcting someone else's error is really our problem. But yeah I remember now that using *hint* here implies that you're not setting anything.

proposals/buildfile.md
```diff
@@ -0,0 +1,128 @@
...
+          - "**/*.class"
+    # another layer, only globally defined file permissions are applied here
+  - name: "other"
+    files:
+      - from: "build/other"
+        to: "/app"
+    # a tar layer 
+  - name: "some tar layer"
+    mediatype: "some/mediatype"
+    tar: "build/generated.tar"
```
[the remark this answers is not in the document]
loosebazooka: gotcha, maybe archive?

proposals/buildfile.md
```diff
@@ -0,0 +1,128 @@
...
+          - "**/moose.class"
+        includes:
+          - "**/*.class"
+    # another layer, only globally defined file permissions are applied here
+  - name: "other"
+    files:
+      - from: "build/other"
+        to: "/app"
+    # a tar layer 
+  - name: "some tar layer"
```
briandealwis, line 75: It'd be nice to support a foreign-layer and/or URL.  Ideally in such a way that it doesn't need to change to switch between `format: Docker` and `format: OCI`.

briandealwis: > Ideally in such a way that it doesn't need to change to switch between format: Docker and format: OCI.

Assuming that is a goal.  I'm not sure it makes sense, especially if you're looking to explicitly support `mediaType`s.

loosebazooka: ah yes, forgot about this. Will add that in too.

proposals/buildfile.md
```diff
@@ -0,0 +1,128 @@
+# Jib CLI Buildfile Specification
+
+Specification for a YAML buildfile describing building a container image. This buildfile can be
+used by the jib-cli to generate a container. It is translated directly into a buildplan and
+passed to the builder.
+
+```
+# "FROM" with option to use os/architecture for manifest lists
+baseImage: "gcr.io/distroless/java:8"
+baseImageOs: linux
```
briandealwis, line 10: There are other properties to select an image from a manifest list / [image index](https://github.com/opencontainers/image-spec/blob/master/image-index.md).  Maybe this can be made more explicit as an map.  Something like:
```
from:
  image: "gcr.io/distroless/java:8"
  platform:
    os: linux
    arch: arm
    variant:  v6
```

It would be very useful to try expressing your format as a JSON schema so that it can be used to provide auto-complete support by IDEs.  @dgageot wrote some scripts for [Skaffold to turn Go-based structures into JSON schema](https://github.com/GoogleContainerTools/skaffold/blob/master/hack/schemas/main.go).

proposals/buildfile.md
chanseokoh: For a foreign layer (which must use a URL), I think `mediaType` already covers that?
```
"mediaType": "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
```
And a tool could just support URL in the `archive` (or `archiveFile`) field if they want?

proposals/buildfile.md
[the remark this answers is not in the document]
chanseokoh: That works.

proposals/buildfile.md
```diff
@@ -0,0 +1,128 @@
...
+
+user: root
+workingDirectory: "/somewhere"
+entrypoint:
+  - "java"
+  - "-jar"
+cmd:
+  - "myjar.jar"
+
+# global file properties applied to all layers
```
briandealwis, line 39: Including existing layers?  Maybe `layerDefaults`?

But maybe this is better handled by YAML aliases/anchors?  (Here's a [nice overview](https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd))

proposals/buildfile.md
```diff
@@ -0,0 +1,128 @@
...
+  - "456"
+  - "789/tcp"
+
+user: root
+workingDirectory: "/somewhere"
+entrypoint:
+  - "java"
+  - "-jar"
+cmd:
+  - "myjar.jar"
```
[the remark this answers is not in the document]
loosebazooka: The parser itself doesn't do much, it allows lists of strings and that's it. Validation (beyond simple type validation) will probably happen at a different level.

proposals/buildfile.md
loosebazooka: maybe url is the right value?

loosebazooka: actually archive seems fine for now.

proposals/buildfile.md
loosebazooka: It looks like anchors are kind of a "UI" thing, the parser should support it? This isn't something we have to implement in our data model?

proposals/buildfile.md
```diff
@@ -0,0 +1,187 @@
+# Jib CLI Buildfile Specification
+
+Specification for a YAML buildfile describing building a container image. This buildfile can be
+used by the jib-cli to generate a container. It is translated directly into a buildplan and
+passed to the builder.
+
+```yaml
+# "FROM" with option to use os/architecture for manifest lists
```
loosebazooka, line 8: `kind:` for type and versioning

proposals/buildfile.md
```diff
@@ -0,0 +1,187 @@
...
+
+# global file properties applied to all layers
+layerProperties:
+  filePermissions: "644"
+  directoryPermissions: "755"
+  user: "0"
+  group: "0"
+  timestamp: "0"
+layers:
+  - name: "scripts and classes"
```
loosebazooka, line 62: Platform specific layers?
```
platforms:
  - os:
    architecture:
```

proposals/buildfile.md
```diff
@@ -0,0 +1,187 @@
...
+Jib has some default behavior on inheritance of config parameters from the base image.
+Perhaps this needs to be configurable
+
+(this is just an exploration, open to some ideas here)
+
+For all values in config of the base image, allow inheritance.
+```
+baseImage:
+  from: "gcr.io/birds/goose"
+  inherit:
```
loosebazooka, line 162: alternative is to configure things for removal
strip:
  - environment
  - volumes

proposals/buildfile.md
```diff
@@ -0,0 +1,187 @@
...
+  - name: "scripts and classes"
+    # file properties only applied to this layer "scripts and classes"
+    properties:
+      filePermissions: "333"
+      directoryPermissions: "777"
+      user: "goose"
+      group: "3"
+      timestamp: "2020-06-03T19:31:50+00:00"
+    files:
+      - from: "target/scripts"
```
loosebazooka, line 71: file copy behavior needs to be defined

loosebazooka: target/bin/script.amd64
->
/app/bin/script (script is a file)

loosebazooka: /app/bin/script/ is a directory
/app/bin/script is a file, if from is a file
/app/bin/script is a directory, if from is a directory

loosebazooka: what about the behavior of docker add/copy?

loosebazooka: symlinks?
- expand out symlinks, they will be regular files on destination.
- preserve if contained within the layer

alternatives
- preserve and hope the file exists on the destination system?

symlink features: expand, preserve (check rsync behavior)

proposals/buildfile.md
```diff
@@ -0,0 +1,187 @@
...
+      - from: "build/other"
+        to: "/app"
+
+    # a archive layer using a tar, default mediaType
+  - name: "some tar layer"
+    archive: "build/generated.tar"
+
+    # a foreign layer using the optional mediatype for archive layers
+  - name: "some foreign layer"
+    mediaType: "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
```
loosebazooka, line 97: can this just be keyword "foreign?"
should we remove mediatype

proposals/buildfile.md
```diff
@@ -0,0 +1,187 @@
...
+    files:
+      - from: "build/other"
+        to: "/app"
+
+    # a archive layer using a tar, default mediaType
+  - name: "some tar layer"
+    archive: "build/generated.tar"
+
+    # a foreign layer using the optional mediatype for archive layers
+  - name: "some foreign layer"
```
loosebazooka, line 96: what about history entry?

proposals/buildfile.md
loosebazooka: reworked this section.

proposals/buildfile.md
loosebazooka: reworked this section

proposals/buildfile.md
loosebazooka: added section for supporting this

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
+# Jib CLI Buildfile Specification
+
+Specification for a YAML buildfile describing building a container image. This buildfile can be
+used by the jib-cli to generate a container. It is translated directly into a buildplan and
+passed to the builder.
+
+```yaml
+apiVersion: jib/v0alpha1
```
briandealwis, line 8:
```suggestion
apiVersion: jib/v1alpha1
```
`alpha` and `beta` are like Maven `-SNAPSHOT`: `v1alphaX` &rarr; `v1betaX` &rarr; `v1`.
Otherwise you're saying when this is complete then it becomes `v0`.

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+      os: "linux"
+      os.version: "a"
+      os.features:
+        - "b1"
+        - "b2"
+      variant: "c"
+      features:
+        - "d1"
+        - "d2"
+    - architecutre: amd64
```
briandealwis, line 25:
```suggestion
    - architecture: amd64
```

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+passed to the builder.
+
+```yaml
+apiVersion: jib/v0alpha1
+kind: Config
+
+# "FROM" with detail for manifest lists or multiple architectures
+from:
+  image: "gcr.io/distroless/java:8"
+  platforms:
```
briandealwis, line 14:
```suggestion
  # optional: if missing, then defaults to `linux/amd64`
  platforms:
```

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+
+Specification for a YAML buildfile describing building a container image. This buildfile can be
+used by the jib-cli to generate a container. It is translated directly into a buildplan and
+passed to the builder.
+
+```yaml
+apiVersion: jib/v0alpha1
+kind: Config
+
+# "FROM" with detail for manifest lists or multiple architectures
```
briandealwis, line 11: So what happens if the user does `jib build ./buildplan.yaml gcr.io/project/image` and `gcr.io/project/image` has never existed?  Do we always create a manifest list with a single image?  Do we only create a manifest list if there is > 1 platform specified?

What happens if `gcr.io/project/image` does exist and it's a manifest for a different platform (not a manifest list)?  Do we replace it?  Do we "promote" the existing image to a component of a manifest list?

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+from:
+  image: "gcr.io/distroless/java:8"
+  platforms:
+    - architecture: "arm"
+      os: "linux"
+      os.version: "a"
+      os.features:
```
briandealwis, line 18: It would seem like `os.version` and `os.features` should be subitems of `os`:
```yaml
os:
  id: "linux"
  version: "a"
  features: ["b1", "b2"]
```

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+      os: darwin
+
+# potentially simple form of "FROM" (based on ability to define schema)
+from: "gcr.io/distroless/java:8"
+
+creationTime: 0 # millis since epoch or iso8601 creation time
+format: Docker # Docker or OCI
+
+environment:
+  "MY_KEY1": "value"
```
briandealwis, line 35: I haven't done an exhaustive search but it feels to me most other environments use an array of strings:
```yaml
environment:
  - "MYKEY1=value"
```

That said, map-style allows using aliases and anchors so 🤷

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+          excludes:
+            - "**/goose.class"
+            - "**/moose.class"
+          includes:
+            - "**/*.class"
+
+      # another layer, only globally defined file permissions are applied here
+    - name: "other"
+      files:
+        - from: "build/other"
```
briandealwis, line 92: I was thrown off by this: I thought the `from` was referring to a base image.  WDYT about `source` and `destination`?

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+* `from` *required*: a file, or directory on disk to copy from
+* `includes`: only includes the patterns matched in this parameter
+* `exludes`: excludes all files (higher prescendence than `includes`) matched in this parameter
+* `properties`: FilesProperties to represent on disk metadata for all files in this copy directive
+
+### FileProperties
+
+A list of properties that can be user specified for each file in a file layer
+* filePermissions: An octal representing permissions for files
+* directoryPermissions: An octal representing permissions for directories
```
briandealwis, line 133: Note that YAML has native support for representing hex and octal numbers  (https://yaml.org/refcard.html)

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+The paratmeters will be overwritten:
+- `user`
+- `workingDirectory`
+- `entrypoint`
+- `cmd`
+
+If we start getting specific user requests to control this, we can explore
+inheritance control in the future.
+
+## Extented features (not included in this build)
```
briandealwis, line 170: ?
```suggestion
## Extended features (not included in this build)
```

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+
+The behavior of the buildfile values post-inheritance must be considered
+
+These parameters will allow appending to the base image value:
+- `environment`
+- `volumes`
+- `labels`
+- `exposedPorts`
+
+The paratmeters will be overwritten:
```
briandealwis, line 161:
```suggestion
These parameters will be overwritten:
```

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
...
+- Any properties not in `Copy` or `Layer` are applied from `Global`
+- Any properties not defined anywhere use jib system defaults.
+
+### Base image value inheritance
+The value(s) defined in the base image are preserved and propogated into the
+config of the new container.
+
+The behavior of the buildfile values post-inheritance must be considered
+
+These parameters will allow appending to the base image value:
```
briandealwis, line 155: Separate out `environment` and `labels`: they will override values with the same key.

loosebazooka: ah good call.

proposals/buildfile.md
loosebazooka: Currently parsing as string in my prototype, but I'll see what the parser allows.

proposals/buildfile.md
loosebazooka: oh yeah this was just filler. I'll update.

proposals/buildfile.md
```diff
@@ -0,0 +1,186 @@
+# Jib CLI Buildfile Specification
+
+Specification for a YAML buildfile describing building a container image. This buildfile can be
+used by the jib-cli to generate a container. It is translated directly into a buildplan and
+passed to the builder.
+
+```yaml
+apiVersion: jib/v0alpha1
+kind: Config
```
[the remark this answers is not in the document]
loosebazooka: :+1:

proposals/buildfile.md
loosebazooka: yeah that's just how they're represented in the specs, so I was mirroring them. not sure what the right thing in here, leaning to keep os.version

proposals/buildfile.md
loosebazooka: Yeah I dunno, I have this parsing correctly like this, and it seems to make sense. I would prefer to keep it this way, unless the ecosystem has decided otherwise.

proposals/buildfile.md
loosebazooka: src, dest makes sense

proposals/buildfile.md
loosebazooka: While this behavior will be determined by the underlying jib system, but yeah, it makes sense to at least document behavior here. I would think that each build with multiple platforms will create a new manifest list. It will not append.

proposals/buildfile.md
loosebazooka: https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list-field-descriptions

proposals/buildfile.md
loosebazooka: Thinking about it, sure we can accept `0777` instead of `777`, but I think we already have parsing logic to handle this. And `777` feels like a better UX to me even if not technically a correct octal match.

chanseokoh: Hmm... yeah I think the string type is better. 0777 can be confusing, because you can actual do things like `chmod 4755 setCtrls.sh` and in this context the leading 0 could be interpreted as that you want to clear the first digit.

proposals/buildfile.md
loosebazooka: and all writes are overwrites

proposals/buildfile.md
loosebazooka: added section about behavior for builds with multiple platforms

proposals/buildfile.md
loosebazooka: Jib currently uses a map style config. Will leave like this for now.

proposals/buildfile.md
loosebazooka: added a section about multiple platforms.

loosebazooka:
I think we can do a review of this? This is basically implemented as is.

chanseokoh (approved):

PR #2531, opened 2020-06-16
New option to skip pushing images (manifests) if the image (manifest) already exists in the target registry

### LINKED ISSUE #2360: Option to disable adding new tags to existing images

karlmuscat:
Fixes #2360.

See #2387.


chanseokoh:
Sorry for the delay. Appreciate your patience.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestCheckerIntegrationTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.Test;
+
+/** Integration tests for {@link ManifestChecker}. */
+public class ManifestCheckerIntegrationTest {
+
+  /** A known manifest list sha for openjdk:11-jre-slim. */
+  public static final String KNOWN_MANIFEST =
+      "sha256:8ab7b3078b01ba66b937b7fbe0b9eccf60449cc101c42e99aeefaba0e1781155";
+
+  /** A known manifest list sha for openjdk:11-jre-slim. */
```
loosebazooka, line 40: this comment seems to be off? maybe should be **un**known ?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestCheckerIntegrationTest.java
```diff
@@ -0,0 +1,81 @@
...
+  public static final String UNKNOWN_MANIFEST =
+      "sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+
+  @ClassRule public static final LocalRegistry localRegistry = new LocalRegistry(5000);
+
+  @BeforeClass
+  public static void setUp() throws IOException, InterruptedException {
+    localRegistry.pullAndPushToLocal("busybox", "busybox");
```
loosebazooka, line 48: is this necessary?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
```diff
@@ -0,0 +1,190 @@
+/*
+ * Copyright 2018 Google LLC.
```
loosebazooka, line 2: 2018 -> 2020

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
```diff
@@ -0,0 +1,190 @@
...
+  }
+
+  @Nullable
+  @Override
+  public BlobHttpContent getContent() {
+    return null;
+  }
+
+  @Override
+  public List<String> getAccept() {
```
loosebazooka, line 70: I don't think we should be duplicating code like this, the code in ManifestPuller and this can go out of sync. Is there a way provide a shared resource that both can refrence?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
```diff
@@ -47,23 +45,25 @@
...
         ProgressEventDispatcher progressEventDispatcher =
             progressEventDispatcherFactory.create("launching manifest pushers", tags.size())) {
 
-      // Gets the image manifest to push.
-      BuildableManifestTemplate manifestTemplate =
-          new ImageToJsonTranslator(builtImage)
-              .getManifestTemplate(
-                  buildContext.getTargetFormat(), containerConfigurationDigestAndSize);
+      if (JibSystemProperties.skipExistingImages()
+          && registryClient.checkImage(imageDigest.toString()).isPresent()) {
```
chanseokoh, line 61: This check is still suffering from the concurrency issue I mentioned in https://github.com/GoogleContainerTools/jib/pull/2387#issuecomment-612192092. As we discussed, you should follow the [design outlined in the proposal]( https://github.com/GoogleContainerTools/jib/pull/2387/files#diff-7a6f7ffce8c0482a6cb7542dd553fd25R24-R28) to create a separate `-Step` class to call `registryClient.checkImage()`. If the program control already reached here, it means multiple `PushImageStep`s are scheduled and running.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestCheckerIntegrationTest.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2018 Google LLC.
```
chanseokoh, line 2: 2020

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -218,7 +219,60 @@ public void testSteps_forBuildToDockerRegistry_multipleTags()
...
+
+    // Test that the pull request throws an exception, indicating that the new tag was not pushed.
+    String imageReference3 = "localhost:5000/testimagerepo:new_testtag";
+    try {
+      localRegistry.pull(imageReference3);
+      Assert.fail(
+          "jib.skipExistingImages was enabled and digest was already pushed, "
+              + "hence testtag2 shouldn't have been pushed.");
+    } catch (RuntimeException ignore) {
+      // As expected, registry throws exception that manifest is unknown.
```
chanseokoh, line 263: Please catch the specific exception and verify the exception message (`assertEquals(..., ex.getMessage())` or `assertThat(ex.getMessage(), ....containsString(...))` for example).

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -218,7 +219,60 @@ public void testSteps_forBuildToDockerRegistry_multipleTags()
...
+    }
+
+    // Test that both images have the same properties.
+    Assert.assertEquals(image1.getDigest(), image2.getDigest());
+    Assert.assertEquals(image1.getImageId(), image2.getImageId());
+
+    // Cleanup. Disable option to not disrupt other tests.
+    System.setProperty(JibSystemProperties.SKIP_EXISTING_IMAGES, "false");
```
chanseokoh, line 271: We use [`@Rule RestoreSystemProperties`](https://github.com/GoogleContainerTools/jib/blob/826b5bbff1192ec3bf411bf4d8a14bf52ad157c4/jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildContextTest.java#L57) to clean up system properties, because cleaning this way is fragile unless it's wrapper around `try { ... } finally { ...}`.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -218,7 +219,60 @@ public void testSteps_forBuildToDockerRegistry_multipleTags()
...
+    Assert.assertEquals(
+        "Hello, world. An argument.\n", new Command("docker", "run", "--rm", imageReference).run());
+
+    // Test that any additional tags have also been pushed with the original image.
+    String imageReference2 = "localhost:5000/testimagerepo:testtag2";
+    localRegistry.pull(imageReference2);
+    assertDockerInspect(imageReference2);
+    Assert.assertEquals(
+        "Hello, world. An argument.\n",
+        new Command("docker", "run", "--rm", imageReference2).run());
```
chanseokoh, line 246: I think we can skip this testing, as the target of this test method is to test `SKIP_EXISTING_IMAGES`. Our integration are increasing running more Docker images and getting slower.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -218,7 +219,60 @@ public void testSteps_forBuildToDockerRegistry_multipleTags()
...
+    JibContainer image2 =
+        buildRegistryImage(
+            ImageReference.of("gcr.io", "distroless/java", DISTROLESS_DIGEST),
+            ImageReference.of("localhost:5000", "testimagerepo", "new_testtag"),
+            Collections.emptyList());
+
+    // Test that the pull request throws an exception, indicating that the new tag was not pushed.
+    String imageReference3 = "localhost:5000/testimagerepo:new_testtag";
+    try {
+      localRegistry.pull(imageReference3);
```
chanseokoh, line 258:
```suggestion
    try {
      localRegistry.pull("localhost:5000/testimagerepo:new_testtag");
```

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
chanseokoh: Agreed. I really think we should not duplicate this much amount of code.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java
```diff
@@ -367,6 +367,31 @@ private Authorization refreshBearerAuth(@Nullable String wwwAuthenticate)
...
+   * @param <T> child type of ManifestTemplate
+   * @param manifestTemplateClass the specific version of manifest template to check
+   * @param imageDigest the image digest to check for
+   * @return the image's {@link ManifestAndDigest} if the image exists on the registry, or {@link
+   *     Optional#empty()} if it doesn't
+   * @throws IOException if communicating with the endpoint fails
+   * @throws RegistryException if communicating with the endpoint fails
+   */
+  public <T extends ManifestTemplate> Optional<ManifestAndDigest<T>> checkImage(
+      String imageDigest, Class<T> manifestTemplateClass) throws IOException, RegistryException {
```
chanseokoh, line 382: I didn't find any usage of passing `manifestTemplateClass` in this PR. Maybe just
```
  public Optional<ManifestAndDigest<ManifestTemplate>> checkImage(String imageDigest)
```
?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
karlmuscat: Do you mean that all the test is unnecessary, or just the part where `testtag` and `testtag2` are pushed?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
karlmuscat: The `PushImageStep` steps are being created after the condition is evaluated to false. If it evaluates to true, no steps are created and an empty list is returned. It seems like it achieves the same behaviour but with simpler code. If you think that adding a new step is better, Ill change it.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
karlmuscat: The exception thrown is a `RuntimeException`. I'll add assertions on the message though.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
karlmuscat: I don't see an easy way how to extract common code from the 2 classes, given that they implement the `RegistryEndpointProvider` interface with 2 different types. A simple option would be to change the implementation of `RegistryClient.checkImage` to use `RegistryClient.pullManifest` and catch the exception, like so:

```
  public Optional<ManifestAndDigest<ManifestTemplate>> checkImage(String imageDigest) {
    try {
      return Optional.ofNullable(pullManifest(imageDigest, ManifestTemplate.class));
    } catch (RegistryErrorException e) {
      return Optional.empty();
    }
  }
```

What do you think?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
chanseokoh: Sorry, I mean this "Hello, world" assertion on this line and and `assertDockerInpsect()` on line 243. Also remove them on line 236 and 237.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
chanseokoh: Sounds good. Please verify message.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
chanseokoh: Just in case, I will "unresolve" conversations (include this) until I can verify actual changes. :)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
chanseokoh: Oh, I jumped into the conclusion too early after I saw this check in this file.

However, you still need to create a separate Step for a blocking network call. `makeList()` should just create and return necessary Steps instantaneously and should never block thread scheduling in `StepsRunner`.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
chanseokoh: Would you mind if I push changes to your PR?

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
```diff
@@ -0,0 +1,190 @@
...
+
+    // Finds a MANIFEST_BLOB_UNKNOWN error response code.
+    if (responseException.getContent() == null) {
+      return Optional.empty();
+    }
+
+    ErrorCodes errorCode = ErrorResponseUtil.getErrorCode(responseException);
+    if (errorCode == ErrorCodes.MANIFEST_UNKNOWN) {
+      return Optional.empty();
+    }
```
chanseokoh, line 185: Could you explain these two checks (`responseException.getContext()` and `ErrorCodes.MANIFEST_UNKNOWN`)? I am asking because we've seen different registries have different behaviors that are not really conformant to the Registry API standard, and would like to know if these checks are specific to a certain registry implementation.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
chanseokoh: Ah, I was looking at an earlier commit. I will make undo "unresolving".

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
karlmuscat: Why would I? 😄

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
chanseokoh: nit: per our style, it should be `ex` instead of `e`:

```java
 } catch (RuntimeException ex) {
```

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
karlmuscat: I have followed the same pattern done in `BlobChecker`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
karlmuscat: Ok. Can you have a look at commit [987c58c](https://github.com/GoogleContainerTools/jib/pull/2531/commits/987c58cfec454be74a4c14c387c6e3446d43c47b)? I had initially introduced a step before the `makeList` call. You should look at `CheckImageStep`, `PushImageStep` and `StepsRunner`. Should I change it to something like that?

chanseokoh: The direction seems right. However, we maintain complete separation between business logic (what Steps do) and the async handling (what `StepsRunner` do). That is, `StepsRunner` doesn't care what each Step does or implies but is only interested in scheduling Steps, passing parameters to them as necessary, and ultimately storing results in `results.*`. All the business logic ([building a manifest and computing its digest](https://github.com/GoogleContainerTools/jib/pull/2531/commits/987c58cfec454be74a4c14c387c6e3446d43c47b#diff-18df945d9e250304dbf7dd948486e16fR395-R401) and [the conditional skip](https://github.com/GoogleContainerTools/jib/pull/2531/commits/987c58cfec454be74a4c14c387c6e3446d43c47b#diff-18df945d9e250304dbf7dd948486e16fR410-R414)) should go out of `StepsRunner`. For example, `StepsRunner` code shouldn't have conditional Step scheduling based on the result of some step. So, probably you should put the manifest construction and digest computation into `CheckImageStep`. The step would then return an `Optional<Manifest...>`, and `StepsRunner` would just store the result. (If deemed necessary, `CheckImageStep` could return the manifest it built, but because anyone can easily build the same manifest later, this may be unnecessary.) This manifest check result should probably be passed into `makeList()` where it returns either an empty list or multiple `PushImageStep` instances based on the result. The responsibility of `makeList()` should just be about properly instantiating necessary `-Step` instances where all the core business logic goes in.

karlmuscat: Pushed new changes for this.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
chanseokoh: Alright, I will look the Registry API specification later for this part and will probably take care of this one way or another in a follow-up PR. But I think at least this works and doesn't actually break things, since this is just an extra manifest check; the worst case would be getting an error later. So I think we are good to go.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
chanseokoh: Discussed this with @loosebazooka. I can take care of code duplication in a follow-up PR.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java
```diff
@@ -47,16 +48,25 @@
...
-                buildContext.getEventHandlers(), "Preparing manifest pushers");
+            new TimerEventDispatcher(eventHandlers, "Preparing manifest pushers");
         ProgressEventDispatcher progressEventDispatcher =
             progressEventDispatcherFactory.create("launching manifest pushers", tags.size())) {
 
+      if (JibSystemProperties.skipExistingImages() && manifestAlreadyExists) {
+        eventHandlers.dispatch(
+            LogEvent.info("Skipping pushing manifest; manifest already exists."));
+
+        return ImmutableList.<PushImageStep>builder().build();
```
chanseokoh, line 67:
```suggestion
        return ImmutableList.of();
```

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
chanseokoh: I've just taken care of this in this PR.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/CheckImageStep.java
```diff
@@ -0,0 +1,93 @@
...
+/** Checks the existence of a manifest. */
+class CheckImageStep implements Callable<Optional<ManifestAndDigest<ManifestTemplate>>> {
+
+  private static final String DESCRIPTION = "Checking existence of manifest";
+
+  BuildContext buildContext;
+  ProgressEventDispatcher.Factory progressEventDispatcherFactory;
+  RegistryClient registryClient;
+  BlobDescriptor containerConfigurationDigestAndSize;
+  Image image;
```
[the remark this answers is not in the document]
chanseokoh: Done.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/ManifestChecker.java
chanseokoh: This looks good.

karlmuscat:
My pleasure. Thank you for accepting my proposal. Our architecture will be much leaner with this new feature.

PR #2538, opened 2020-06-22
Temporary fix for registry not having htpassword

loosebazooka:
This is a temporary workaround. Pin dockerhub registry contianer image to 2.7.0.

core issue: https://github.com/docker/distribution-library-image/issues/106


loosebazooka:
Yeah, I think the final solution might be to correctly do the right auth. I'll look into that.

PR #2539, opened 2020-06-23
Programmatically generate htpasswd file

### LINKED ISSUE #2537: Tests failing after official Docker Distribution registry (`registry:2`) removed `htpasswd`
Our integration tests run
```
docker run --rm --entrypoint htpasswd registry:2 -Bbn username password
```

On my desktop where it cached an old version of `registry:2`, the command works.
```
$ docker run --rm --entrypoint htpasswd registry:2 -Bbn username password
username:$2y$05$RO8TSDjphyaccxuztK6L.uQrJ0csk.YpAhfvS7cm2d5Z/cI3PAZm6

$ docker images
REPOSITORY                               TAG                 IMAGE ID            CREATED             SIZE
registry                                 2                   708bc6af7e5e        4 months ago        25.8MB
```

However, the new `registry:2` image they created 27 hours ago fails, as it doesn't have `htpasswd`.
```
$ docker run --rm --entrypoint htpasswd registry:2 -Bbn username password
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"htpasswd\": executable file not found in $PATH": unknown.
$ docker images
REPOSITORY               TAG                  IMAGE ID            CREATED             SIZE
registry                 2                    2d4f4b5309b1        27 hours ago        26.2MB
```

Here's a sample [Kokoro build failure log](https://source.cloud.google.com/results/invocations/0921a6bc-ce79-4cbe-8fa3-04846176ea21/log).

loosebazooka:
Turn out you must use bcrypt. md5 doesn't work for registry even if the htpasswd file is right.

The jbcrypt java library uses $2a$, not $2y$ (described in htpasswd docs: https://httpd.apache.org/docs/2.4/misc/password_encryptions.html), but $2y$ vs $2a$ doesn't seem to matter (integration tests passing).

core issue: docker/distribution-library-image#106
longterm solution for: #2538 

fixes #2537

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
@@ -30,6 +30,7 @@ import java.util.Arrays;
 import java.util.UUID;
 import javax.annotation.Nullable;
 import org.junit.rules.ExternalResource;
+import org.mindrot.jbcrypt.BCrypt;
 
 /** Runs a local registry. */
 public class LocalRegistry extends ExternalResource {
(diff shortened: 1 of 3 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -68,19 +69,11 @@ public void start() throws IOException, InterruptedException {
...
-                  "run",
-                  "--rm",
-                  "--entrypoint",
-                  "htpasswd",
-                  "registry:2.7.0", // TODO: correctly fix this when using latest
-                  "-Bbn",
-                  username,
-                  password)
-              .run();
+      // since registry:2 no longer contains htpasswd, generate the htpasswd file manually
```
chanseokoh, line 72: I think we can just say
```suggestion
      // Equivalent of "$ htpasswd -nbB username password".
```
There's no reason to use or consider `registry:2` at all. The image just happened to have `htpasswd` installed.

PR #2568, opened 2020-07-06
Adding the `from.platforms` Jib config parameter to jib-maven-plugin and to the jib-gradle-plugin

louismurerwa:
Toward #1567

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -48,6 +48,7 @@ public class BaseImageParameters {
     platforms.add(platform);
   }
 
+  @Input
   public ListProperty<PlatformParameters> getPlatforms() {
     return platforms;
   }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -48,7 +48,6 @@ public class BaseImageParameters {
     platforms.add(platform);
   }
 
-  @Input
   ListProperty<PlatformsParameters> getPlatforms() {
     return platforms;
   }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -43,12 +43,12 @@ public class BaseImageParameters {
         objectFactory.newInstance(PlatformParametersSpec.class, objectFactory, platforms);
 
     PlatformParameters platform = new PlatformParameters();
-    platform.os = "linux";
-    platform.architecture = "amd64";
+    platform.setOs("linux");
+    platform.setArchitecture("amd64");
     platforms.add(platform);
   }
 
-  @Input
+  @Nested
   @Optional
   public ListProperty<PlatformParameters> getPlatforms() {
     return platforms;
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -30,16 +30,16 @@ import org.gradle.api.tasks.Optional;
 public class BaseImageParameters {
 
   private final AuthParameters auth;
-  @Nullable private String credHelper;
   @Nullable private String image;
-  private final PlatformParametersSpec platformsParametersSpec;
+  @Nullable private String credHelper;
+  private final PlatformParametersSpec platformParametersSpec;
   private final ListProperty<PlatformParameters> platforms;
 
   @Inject
   public BaseImageParameters(ObjectFactory objectFactory) {
     auth = objectFactory.newInstance(AuthParameters.class, "from.auth");
     platforms = objectFactory.listProperty(PlatformParameters.class).empty();
-    platformsParametersSpec =
+    platformParametersSpec =
         objectFactory.newInstance(PlatformParametersSpec.class, objectFactory, platforms);
 
     PlatformParameters platform = new PlatformParameters();
@@ -55,7 +55,7 @@ public class BaseImageParameters {
   }
 
   public void platforms(Action<? super PlatformParametersSpec> action) {
-    action.execute(platformsParametersSpec);
+    action.execute(platformParametersSpec);
   }
 
   @Input
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -49,6 +49,7 @@ public class BaseImageParameters {
   }
 
   @Input
+  @Optional
   public ListProperty<PlatformParameters> getPlatforms() {
     return platforms;
   }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -32,27 +32,27 @@ public class BaseImageParameters {
   private final AuthParameters auth;
   @Nullable private String credHelper;
   @Nullable private String image;
-  private final PlatformsParametersSpec platformsParametersSpec;
-  private ListProperty<PlatformsParameters> platforms;
+  private final PlatformParametersSpec platformsParametersSpec;
+  private final ListProperty<PlatformParameters> platforms;
 
   @Inject
   public BaseImageParameters(ObjectFactory objectFactory) {
     auth = objectFactory.newInstance(AuthParameters.class, "from.auth");
-    platforms = objectFactory.listProperty(PlatformsParameters.class).empty();
+    platforms = objectFactory.listProperty(PlatformParameters.class).empty();
     platformsParametersSpec =
-        objectFactory.newInstance(PlatformsParametersSpec.class, objectFactory, platforms);
+        objectFactory.newInstance(PlatformParametersSpec.class, objectFactory, platforms);
 
-    PlatformsParameters platform = new PlatformsParameters();
+    PlatformParameters platform = new PlatformParameters();
     platform.os = "linux";
     platform.architecture = "amd64";
     platforms.add(platform);
   }
 
-  ListProperty<PlatformsParameters> getPlatforms() {
+  public ListProperty<PlatformParameters> getPlatforms() {
     return platforms;
   }
 
-  public void platforms(Action<? super PlatformsParametersSpec> action) {
+  public void platforms(Action<? super PlatformParametersSpec> action) {
     action.execute(platformsParametersSpec);
   }
 
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -21,6 +21,7 @@ import javax.annotation.Nullable;
 import javax.inject.Inject;
 import org.gradle.api.Action;
 import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.ListProperty;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Nested;
 import org.gradle.api.tasks.Optional;
@@ -29,13 +30,31 @@ import org.gradle.api.tasks.Optional;
 public class BaseImageParameters {
 
   private final AuthParameters auth;
-
-  @Nullable private String image;
   @Nullable private String credHelper;
+  @Nullable private String image;
+  private final PlatformsParametersSpec platformsParametersSpec;
+  private ListProperty<PlatformsParameters> platforms;
 
   @Inject
   public BaseImageParameters(ObjectFactory objectFactory) {
     auth = objectFactory.newInstance(AuthParameters.class, "from.auth");
+    platforms = objectFactory.listProperty(PlatformsParameters.class).empty();
+    platformsParametersSpec =
+        objectFactory.newInstance(PlatformsParametersSpec.class, objectFactory, platforms);
+
+    PlatformsParameters platform = new PlatformsParameters();
+    platform.os = "linux";
+    platform.architecture = "amd64";
+    platforms.add(platform);
+  }
+
+  @Input
+  ListProperty<PlatformsParameters> getPlatforms() {
+    return platforms;
+  }
+
+  public void platforms(Action<? super PlatformsParametersSpec> action) {
+    action.execute(platformsParametersSpec);
   }
 
   @Input
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
@@ -37,6 +37,12 @@ import org.gradle.api.tasks.Optional;
  *   from {
  *     image = 'gcr.io/my-gcp-project/my-base-image'
  *     credHelper = 'gcr'
+ *     platforms {
+ *       platform {
+ *         os = 'value1'
+ *         architecture = 'value2'
+ *       }
+ *     }
  *   }
  *   to {
  *     image = 'gcr.io/gcp-project/my-app:built-with-jib'
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
@@ -39,8 +39,8 @@ import org.gradle.api.tasks.Optional;
  *     credHelper = 'gcr'
  *     platforms {
  *       platform {
- *         os = 'value1'
- *         architecture = 'value2'
+ *         os = 'linux'
+ *         architecture = 'amd64'
  *       }
  *     }
  *   }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformParameters.java
@@ -28,15 +28,21 @@ public class PlatformParameters implements PlatformConfiguration {
 
   @Input
   @Override
-  @Nullable
   public Optional<String> getOs() {
     return Optional.ofNullable(this.os);
   }
 
+  public void setOs(String os) {
+    this.os = os;
+  }
+
   @Input
   @Override
-  @Nullable
   public Optional<String> getArchitecture() {
     return Optional.ofNullable(this.architecture);
+  }
+
+  public void setArchitecture(String architecture) {
+    this.architecture = architecture;
   }
 }
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformParameters.java
@@ -16,11 +16,14 @@
 
 package com.google.cloud.tools.jib.gradle;
 
+import com.google.cloud.tools.jib.plugins.common.RawConfiguration.PlatformConfiguration;
+import java.util.Optional;
 import javax.annotation.Nullable;
 import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.Internal;
 
 /** Configuration of a platform. */
-public class PlatformParameters {
+public class PlatformParameters implements PlatformConfiguration {
   @Nullable private String os;
   @Nullable private String architecture;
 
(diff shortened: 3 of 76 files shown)


jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -135,6 +150,15 @@ public PermissionConfiguration() {}
     @Nullable @Parameter private String credHelper;
 
     @Parameter private FromAuthConfiguration auth = new FromAuthConfiguration();
+
+    @Parameter private List<PlatformsConfiguration> platforms;
+
+    /** Configuration for {@code platforms} parameter. */
+    public FromConfiguration() {
+      platforms = new ArrayList<>();
```
loosebazooka, line 158: Can the `platforms` list not be empty?

If we were to initialize this to an empty list at line 154, would it break the default beahavior?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -127,6 +128,20 @@ public PermissionConfiguration() {}
     }
   }
 
+  /** Configuration for {@code platform} parameter. */
```
loosebazooka, line 131: Add comments about default perhaps?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -135,6 +150,15 @@ public PermissionConfiguration() {}
     @Nullable @Parameter private String credHelper;
 
     @Parameter private FromAuthConfiguration auth = new FromAuthConfiguration();
+
+    @Parameter private List<PlatformsConfiguration> platforms;
+
+    /** Configuration for {@code platforms} parameter. */
```
loosebazooka, line 156: This comment seems out of place?

loosebazooka: Right maybe change it to `/** Constructor for defaults . **/` or something?

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: What would an appropriate comment look like for this constructor?Would this work?

`/** Initializing a default platform in case a user doesn't pass in one **/

loosebazooka: I think for IDE users, just seeing the pop up of: 
```
/** Configuration for {@code platform} parameter. Defaults to amd64/linux. **/
```
 is helpful, since they wont have to go into the code and see what the defaults are?

But commenting sometimes feels pretty subjective. This is just how I would do it.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -127,6 +128,20 @@ public PermissionConfiguration() {}
     }
   }
 
+  /** Configuration for {@code platform} parameter. */
+  public static class PlatformsConfiguration {
+    @Parameter private String os = "linux";
+    @Parameter private String architecture = "amd64";
+
+    public String getOs() {
```
chanseokoh, line 136: `getOs()` and `getArchitecture()` may not need to be `public`? I see some getters in other nested `-Configuration` classes don't have `public`. We can start with as much limited visibility as possible.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: The idea is to have a default platform amd64/linux in the platforms any time a user fails to specify the platforms.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -330,6 +354,16 @@ protected void checkJibVersion() throws MojoExecutionException {
     MojoCommon.checkJibVersion(descriptor);
   }
 
+  /**
+   * Gets the specified platforms.
+   *
+   * @return the specified platforms
+   */
+  @Nullable
```
chanseokoh, line 362: At the current state, I don't think this method can ever return null. In that case, we should remove `@Nullable`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: I agree commenting is subjective sometimes.I think your suggested comment works well in this case.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: Done

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: We never use these methods locally so making them ` private ` throws ` unused  method warnings` .

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: Thats true .Done.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
chanseokoh: Java has so-called "package-private" in addition to `public`, `protected`, and `private`. Just remove the `public` keyword, and the method will also be accessible from any class in the same Java package.

louismurerwa: Nicee !! This is an interesting concept Java concept.
Done.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -127,6 +128,20 @@ public PermissionConfiguration() {}
     }
   }
 
+  /** Configuration for {@code platform} parameter. Defaults to amd64/linux. * */
+  public static class PlatformsConfiguration {
+    @Parameter private String os = "linux";
+    @Parameter private String architecture = "amd64";
```
chanseokoh, line 134: As discussed in the design doc, if we are going to raise an error when one of these are missing, I think we have to make these `@Nullable` and not assign default values.

Need to add `@Nullable` in getters too.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -127,6 +128,20 @@ public PermissionConfiguration() {}
     }
   }
 
+  /** Configuration for {@code platform} parameter. Defaults to amd64/linux. * */
```
chanseokoh, line 131: Therefore,
```suggestion
  /** Configuration for {@code platform} parameter. */
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
chanseokoh: Another way to create a default singleton (immutable) list:
```java
PlatformsConfiguration platform = new PlatformsConfiguration();
platform.os = ...
platform.version = ...
platforms = Collections.singletonList(platform);
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: Done

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: Makes sense !! Done

chanseokoh: You can mark a GitHub conversion "resolved" yourself if you believe the issue is cleanly resolved (but only after you pushed new changes).

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
louismurerwa: Done

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -135,6 +151,16 @@ public PermissionConfiguration() {}
     @Nullable @Parameter private String credHelper;
 
     @Parameter private FromAuthConfiguration auth = new FromAuthConfiguration();
+
+    @Parameter private List<PlatformsConfiguration> platforms;
+
+    /** Constructor for defaults . * */
```
chanseokoh, line 157:
```suggestion
    /** Constructor for defaults. */
```

chanseokoh:
Oh, I thought you'd merge this and start a new PR for further changes. Whatever works, I'm good with it.

And it's a good practice to put some relevant description in the first PR comment.

louismurerwa:
I added the ` getPlaforms ` method to the ` RawConfiguration ` interface .This interface is has not be impleneted by ` GradleRawConfiguration ` yet some of the tests will fail.I am working on implementing the Gradle side right now but feel free to review the current changes .

chanseokoh:
The `RawConfiguration` changes look good to me.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
```diff
@@ -40,6 +40,13 @@
     Optional<Object> getExtraConfiguration();
   }
 
+  static interface PlatformsConfiguration {
```
chanseokoh, line 43: `PlatformsConfiguration` --> `PlatformConfiguration` (singular)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformsParametersSpec.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package com.google.cloud.tools.jib.gradle;
+
+import javax.inject.Inject;
+import org.gradle.api.Action;
+import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.ListProperty;
+
+/** Allows to add {@link PlatformsParameters} objects to the list property of the same type. */
```
chanseokoh, line 24:
```suggestion
/** Allows to add {@link PlatformParameters} objects to the list property of the same type. */
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformsParametersSpec.java
```diff
@@ -0,0 +1,47 @@
...
+public class PlatformsParametersSpec {
```
chanseokoh, line 25:
```suggestion
public class PlatformParametersSpec {
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformsParameters.java
```diff
@@ -0,0 +1,36 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.gradle;
+
+import com.google.cloud.tools.jib.plugins.common.RawConfiguration.PlatformsConfiguration;
+import javax.annotation.Nullable;
+
+/** Configuration of a platform. */
+public class PlatformsParameters implements PlatformsConfiguration {
```
chanseokoh, line 23:
```suggestion
public class PlatformParameters implements PlatformConfiguration {
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformsParameters.java
```diff
@@ -0,0 +1,36 @@
...
+  @Nullable String os;
+  @Nullable String architecture;
+
+  @Override
+  public java.util.Optional<String> getOs() {
+    return java.util.Optional.ofNullable(this.os);
+  }
```
chanseokoh, line 30: There's no other class with the same name `Optional` used in this file, so you don't have to fully qualify the class. Changing the code like below will make your IDE automatically add `import java.util.Optional` at the top.

```suggestion
  @Override
  public Optional<String> getOs() {
    return Optional.ofNullable(this.os);
  }
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformsParameters.java
```diff
@@ -0,0 +1,36 @@
...
+
+  @Override
+  public java.util.Optional<String> getArchitecture() {
+    return java.util.Optional.ofNullable(this.architecture);
```
chanseokoh, line 34: ditto

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -37,6 +37,12 @@
  *   from {
  *     image = 'gcr.io/my-gcp-project/my-base-image'
  *     credHelper = 'gcr'
+ *     platforms {
+ *       platform {
+ *         os = 'value1'
+ *         architecture = 'value2'
```
chanseokoh, line 43:
```suggestion
 *         os = 'linux'
 *         architecture = 'amd64'
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -29,13 +30,30 @@
 public class BaseImageParameters {
 
   private final AuthParameters auth;
-
-  @Nullable private String image;
   @Nullable private String credHelper;
+  @Nullable private String image;
+  private final PlatformsParametersSpec platformsParametersSpec;
+  private ListProperty<PlatformsParameters> platforms;
```
chanseokoh, line 36: I guess this can be `final`. (`final` in Java means the reference `platforms` must be assigned when the class is instantiated and cannot be changed to something else later.)
```suggestion
  private final ListProperty<PlatformsParameters> platforms;
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
louismurerwa: Done

louismurerwa:
I am getting a `  Warning: Task type 'com.google.cloud.tools.jib.gradle.BuildDockerTask': property 'jib.from.platforms' is not annotated with an input or output annotation. ` when I run ` ./gradlew check ` .What is the output/input annotation?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -29,13 +30,30 @@
...
+    platforms = objectFactory.listProperty(PlatformParameters.class).empty();
+    platformsParametersSpec =
+        objectFactory.newInstance(PlatformParametersSpec.class, objectFactory, platforms);
+
+    PlatformParameters platform = new PlatformParameters();
+    platform.os = "linux";
+    platform.architecture = "amd64";
+    platforms.add(platform);
+  }
+
```
chanseokoh, line 50: You need to add some annotations like [this](https://github.com/GoogleContainerTools/jib/blob/00dbfee080e66c238bd1f1168205cab9e96e11b7/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java#L218-L219). For example, `@Optional` tells the Gradle system that `from.platforms` is an optional config that the user doesn't have to specify.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformParameters.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.cloud.tools.jib.plugins.common.RawConfiguration.PlatformConfiguration;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/** Configuration of a platform. */
+public class PlatformParameters implements PlatformConfiguration {
+  @Nullable String os;
+  @Nullable String architecture;
+
+  @Override
```
chanseokoh, line 28: Needs to add `@Input` as in [`ExtensionsParameters`](https://github.com/GoogleContainerTools/jib/blob/00dbfee080e66c238bd1f1168205cab9e96e11b7/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtensionParameters.java#L34). This tells Gradle that `platform.os` is an input parameter.

And I think you may need setters. Please check if setting `platform.os = ...` in `build.gradle` actually works.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformParameters.java
```diff
@@ -0,0 +1,37 @@
...
+  public Optional<String> getOs() {
+    return Optional.ofNullable(this.os);
+  }
+
+  @Override
```
chanseokoh, line 33: ditto

chanseokoh:
FTR: we realized there was a really tricky issue on the Gradle side, so I needed to take care of that part.

@louismurerwa please verify there's no issue when running a snapshot version.

chanseokoh:
@louismurerwa actually, I think we can go simpler with renaming the interface method names to `getOsName()` and `getOsArchitecture()`. I thought this would change the Maven config UI, but having extra getters doesn't seem to matter. But please verify both Maven and Gradle snapshot versions.

louismurerwa:
> @louismurerwa actually, I think we can go simpler with renaming the interface method names to `getOsName()` and `getOsArchitecture()`. I thought this would change the Maven config UI, but having extra getters doesn't seem to matter. But please verify both Maven and Gradle snapshot versions.

louismurerwa:
Gradle plugin built and ran successfully

louismurerwa:
> > @louismurerwa actually, I think we can go simpler with renaming the interface method names to `getOsName()` and `getOsArchitecture()`. I thought this would change the Maven config UI, but having extra getters doesn't seem to matter. But please verify both Maven and Gradle snapshot versions.

The plugins are compiling and executing without errors on my local environment.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -29,13 +30,32 @@
 public class BaseImageParameters {
 
   private final AuthParameters auth;
-
-  @Nullable private String image;
   @Nullable private String credHelper;
+  @Nullable private String image;
+  private final PlatformParametersSpec platformsParametersSpec;
```
chanseokoh, line 35: platformsParametersSpec --> platformParametersSpec

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -29,13 +30,32 @@
 public class BaseImageParameters {
 
   private final AuthParameters auth;
-
-  @Nullable private String image;
   @Nullable private String credHelper;
+  @Nullable private String image;
```
chanseokoh, line 34: nit: I see the order of existing fields has changed. Let's revert this.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -127,6 +128,22 @@ public PermissionConfiguration() {}
     }
   }
 
+  /** Configuration for {@code platform} parameter. */
+  public static class PlatformsParameters implements PlatformConfiguration {
```
chanseokoh, line 132: PlatformsParameters --> PlatformParameters

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
louismurerwa: Do you mean putting image on top of credHelper?

chanseokoh: Yes. This is just an unnecessary change. No need to change unrelated code in this PR.

louismurerwa: Done !!

PR #2571, opened 2020-07-08
Add yaml dependencies and FilePropertiesSpec

loosebazooka:
- add jackson databind, yaml dependencies
- add first parseable portion of buildfile
- add `Instants` class which may be removed later

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesSpec.java
@@ -1,0 +1,50 @@
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.cloud.tools.jib.api.buildplan.FilePermissions;
+
+import javax.annotation.Nullable;
+import java.time.Instant;
+import java.util.Optional;
+
+public class FilePropertiesSpec {
+  @Nullable private FilePermissions filePermissions;
+  @Nullable private FilePermissions directoryPermissions;
+  @Nullable private String user;
+  @Nullable private String group;
+  @Nullable private Instant timestamp;
+
+  @JsonCreator
+  public FilePropertiesSpec(@JsonProperty("filePermissions") String filePermissions,
+                            @JsonProperty("directoryPermissions") String directoryPermissions,
+                            @JsonProperty("user") String user,
+                            @JsonProperty("group") String group,
+                            @JsonProperty("timestamp") String timestamp) {
+    this.filePermissions = filePermissions == null ? null : FilePermissions.fromOctalString(filePermissions);
+    this.directoryPermissions = directoryPermissions == null ? null : FilePermissions.fromOctalString(directoryPermissions);
+    this.user = user;
+    this.group = group;
+    this.timestamp = (timestamp == null) ? null : Instants.fromMillisOrIso8601(timestamp, "timestamp");
+  }
+
+  public Optional<FilePermissions> getFilePermissions() {
+    return Optional.ofNullable(filePermissions);
+  }
+
+  public Optional<FilePermissions> getDirectoryPermissions() {
+    return Optional.ofNullable(directoryPermissions);
+  }
+
+  public Optional<String> getUser() {
+    return Optional.ofNullable(user);
+  }
+
+  public Optional<String> getGroup() {
+    return Optional.ofNullable(group);
+  }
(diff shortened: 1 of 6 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Instants.java
```diff
@@ -0,0 +1,54 @@
...
+
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+
+/** Helper class to convert various strings to Instants. */
+public class Instants {
```
chanseokoh, line 25: Can this be package-private?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Instants.java
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Parses a string into Instant, string must be time in millis or iso8601 datetime.
+   *
+   * @param time in millis or is8601 format
+   * @param fieldName name of field being parse (for error messaging)
+   * @return Instant value of parsed time
+   */
+  public static Instant fromMillisOrIso8601(String time, String fieldName) {
```
chanseokoh, line 33: package-private?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesSpec.java
```diff
@@ -0,0 +1,91 @@
...
+ * group: bar
+ * timestamp: 100
+ * }</pre>
+ */
+public class FilePropertiesSpec {
+  @Nullable private FilePermissions filePermissions;
+  @Nullable private FilePermissions directoryPermissions;
+  @Nullable private String user;
+  @Nullable private String group;
+  @Nullable private Instant timestamp;
```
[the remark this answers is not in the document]
loosebazooka: Yeah actually I think so, I was using a different jackson mechanism to load the values, but since I'm using `@JsonCreator` now, final seems possible, let me test it out.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesSpecTest.java
```diff
@@ -0,0 +1,108 @@
...
+public class FilePropertiesSpecTest {
+
+  private static ObjectMapper filePropertiesSpecMapper;
+
+  @BeforeClass
+  public static void createObjectMapper() {
+    filePropertiesSpecMapper =
+        new ObjectMapper(new YAMLFactory())
+            .configure(DeserializationFeature.FAIL_ON_IGNORED_PROPERTIES, true);
+  }
```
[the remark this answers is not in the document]
loosebazooka: Actually when I took apart these tests, I meant to use `(FAIL_ON_UNKNOWN_PROPERTIES, true)`, but it looks like the default value for those is true anyway. So this doesn't do anything useful. Moving to field declaration. Good Catch.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesSpecTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+  @Test
+  public void testFilePropertiesSpec_full() throws JsonProcessingException {
+    String data =
+        ""
```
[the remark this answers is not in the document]
loosebazooka: yeah I think so when I wrote this, but I don't think that's a great reason. Removed this.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesSpecTest.java
```diff
@@ -0,0 +1,114 @@
...
+import java.time.Instant;
+import org.hamcrest.CoreMatchers;
+import org.hamcrest.MatcherAssert;
+import org.junit.Assert;
+import org.junit.Test;
+
+/** Tests for {@link FilePropertiesSpec}. */
+public class FilePropertiesSpecTest {
+
+  private static ObjectMapper filePropertiesSpecMapper = new ObjectMapper(new YAMLFactory());
```
chanseokoh, line 34: nit: `final`

loosebazooka: oops, done

Issue #2570, opened 2020-07-08
[jib-cli] buildfile tracking bug


loosebazooka:
- [x] Proposal (#2515) 

Parser
- [x] Yaml Parsing Dependencies (#2571)
- [x] PropertiesSpec (#2571)
- [x] CopySpec (#2579)
- [x] LayerSpec, ArchiveLayerSpec, FileLayerSpec (#2580)
- [x] PlatformSpec (#2581) (#2624)
- [x] BaseImageSpec (#2588)
- [x] LayersSpec (#2590)
- [x] BuildFileSpec (#2598) (#2624) 

Processors
- [x] File Permissions Stack (#2645) (#2683)
- [x] LayersSpec -> List<LayerObject> (#2667) (#2700)
- [x] JibContainerBuilder Generator (#2735)

Open Questions?
- [x] Disallow null and empty values on requires parameters (#2606)
- [x] Disallow empty values on optional parameters (#2649)
- [x] Disallow null or empty values in collections (#2649)
- [x] Use of optional in non-api classes (https://github.com/GoogleContainerTools/jib/pull/2581#discussion_r453899698)
- [x] Can we throw subclass of JsonProcessingException? (https://github.com/GoogleContainerTools/jib/pull/2598#discussion_r456577550)
- [x] Empty collections vs Optional. (https://github.com/GoogleContainerTools/jib/pull/2598#discussion_r456577550) (#2609)

Opened some new bugs for any remaining issues

PR #2580, opened 2020-07-10
Add LayerSpec, ArchiveLayerSpec, FileLayerSpec for buildfile

loosebazooka:
part of #2570

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpec.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * A yaml block for specifying archive layers
+ *
+ * <p>Example use of this yaml snippet.
+ *
+ * <pre>{@code
+ * name: "my tar layer"
+ * archive: output/mytar.tgz
+ * // optional mediatype
+ * mediaType: application/vnd.docker.image.rootfs.diff.tar.gzip
+ * }</pre>
+ */
+@JsonDeserialize(using = JsonDeserializer.None.class) // required since LayerSpec overrides this
+public class ArchiveLayerSpec implements LayerSpec {
+
+  private String name;
+  // TODO: arhive should maybe be a uri to support file paths or urls
+  private Path archive;
+  @Nullable private String mediaType;
+
+  /**
+   * Constructor for use by jackson to populate this object.
+   *
+   * @param name a unique name for this layer
+   * @param archive a path to an archive file
+   * @param mediaType the media type of the file
+   */
+  @JsonCreator
+  public ArchiveLayerSpec(
+      @JsonProperty(value = "name", required = true) String name,
+      @JsonProperty(value = "archive", required = true) String archive,
+      @JsonProperty("mediaType") String mediaType) {
+    this.name = name;
+    this.archive = Paths.get(archive);
+    this.mediaType = mediaType;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public Path getArchive() {
+    return archive;
+  }
+
+  public Optional<String> getMediaType() {
+    return Optional.ofNullable(mediaType);
+  }
+}
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpec.java
@@ -26,7 +26,7 @@ import java.util.Optional;
 import javax.annotation.Nullable;
 
 /**
- * A yaml block for specifying archive layers
+ * A yaml block for specifying archive layers.
  *
  * <p>Example use of this yaml snippet.
  *
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FileLayerSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FileLayerSpec.java
@@ -48,7 +48,7 @@ public class FileLayerSpec implements LayerSpec {
    * Constructor for use by jackson to populate this object.
    *
    * @param name a unique name for this layer
-   * @param files a path to an archive file
+   * @param files a list of {@link CopySpec} describing files to add to the layer
    * @param properties a {@link FilePropertiesSpec} that applies to all files in this layer
    */
   @JsonCreator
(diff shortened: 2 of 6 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpec.java
```diff
@@ -0,0 +1,76 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * A yaml block for specifying archive layers
```
chanseokoh, line 29: I do remember the checkstyle complained if having a Javadoc not ending with a period. :man_shrugging:

```suggestion
 * A yaml block for specifying archive layers.
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FileLayerSpec.java
```diff
@@ -0,0 +1,75 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * A yaml block for specifying files layers
```
chanseokoh, line 28:
```suggestion
 * A yaml block for specifying files layers.
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FileLayerSpec.java
```diff
@@ -0,0 +1,75 @@
...
+   * Constructor for use by jackson to populate this object.
+   *
+   * @param name a unique name for this layer
+   * @param files a path to an archive file
+   * @param properties a {@link FilePropertiesSpec} that applies to all files in this layer
+   */
+  @JsonCreator
+  public FileLayerSpec(
+      @JsonProperty(value = "name", required = true) String name,
+      @JsonProperty(value = "files", required = true) List<CopySpec> files,
```
chanseokoh, line 57: Should we make a copy?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayerSpec.java
```diff
@@ -0,0 +1,54 @@
...
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import java.io.IOException;
+
+/**
+ * Polymorphic yaml LayerSpec interface with custom deserializer, can parse both {@link
+ * ArchiveLayerSpec} and {@link FileLayerSpec}.
+ */
+@JsonDeserialize(using = LayerSpec.Deserializer.class)
+public interface LayerSpec {
+  class Deserializer extends StdDeserializer<LayerSpec> {
```
chanseokoh, line 32: Probably `LayerSpec.Deserializer` is publicly accessible, but I guess it doesn't matter because all the buildfile classes are internal and not public API, right?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayerSpec.java
```diff
@@ -0,0 +1,54 @@
...
+    public Deserializer() {
+      super(LayerSpec.class);
+    }
+
+    /**
+     * Deserialize based on the presence of "archive", if present, consider the layer to be of type
+     * {@link ArchiveLayerSpec}, else a {@link FileLayerSpec}.
+     */
+    @Override
+    public LayerSpec deserialize(JsonParser jp, DeserializationContext txt) throws IOException {
```
chanseokoh, line 43: And how about
jp --> parser
txt --> context
n --> node
?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpec.java
loosebazooka: Weird that it didn't catch these?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayerSpec.java
loosebazooka: That's a good point, but since interfaces don't allow the private modifier, I think this is fine. Otherwise we would have to add another class to do this?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FileLayerSpec.java
loosebazooka: So I thought about that. This is not a public api, and we will be converting it directly into a build plan. No one will really be processing it in the meantime, so I think it's fine to leave like that.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpecTest.java
```diff
@@ -0,0 +1,73 @@
...
+    } catch (JsonProcessingException jpe) {
+      MatcherAssert.assertThat(
+          jpe.getMessage(), CoreMatchers.startsWith("Missing required creator property 'name'"));
+    }
+  }
+
+  /**
+   * With {@link LayerSpec.Deserializer#deserialize} this test seems pointless, but it still helps
+   * define the behavior of this class.
+   */
```
chanseokoh, line 60: nit: I suggest this as code comment rather than a Javadoc.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FileLayerSpec.java
```diff
@@ -0,0 +1,75 @@
...
+public class FileLayerSpec implements LayerSpec {
+  private String name;
+  private List<CopySpec> files;
+  @Nullable private FilePropertiesSpec properties;
+
+  /**
+   * Constructor for use by jackson to populate this object.
+   *
+   * @param name a unique name for this layer
+   * @param files a path to an archive file
```
chanseokoh, line 51: incorrect doc?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayerSpec.java
```diff
@@ -0,0 +1,55 @@
...
+@JsonDeserialize(using = LayerSpec.Deserializer.class)
+public interface LayerSpec {
+  class Deserializer extends StdDeserializer<LayerSpec> {
+
+    public Deserializer() {
+      super(LayerSpec.class);
+    }
+
+    /**
+     * Deserialize based on the presence of "archive", if present, consider the layer to be of type
```
chanseokoh, line 39: I remember this has to start with "Deserialize**s** ...". And the comment reads a bit weird to me.

loosebazooka: updated this comment

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayerSpec.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Deserializes to {@link ArchiveLayerSpec} if yaml contains "archive" field, to {@link
+     * FileLayerSpec} if yaml contains "files" field or throws {@link IOException} if neither is
+     * found or no "name" was specified for the layer entry.
+     */
+    @Override
+    public LayerSpec deserialize(JsonParser jsonParser, DeserializationContext context)
+        throws IOException {
```
chanseokoh, line 45: Sorry, one final comment, then LGTM. I just noticed that the `abstract` super method throws `IOException` and `JsonProcessingException`. Isn't throwing `JsonProcessingException` more suitable for the errors you are checking?

loosebazooka: JsonProcessingException has no public constructor unfortunately :cry: .

loosebazooka: maybe there's some other subclass that I can use. I'll add it to the list of things to check on the tracking bug.

chanseokoh: Huh, interesting. I am fine going with `IOException` then.

PR #2581, opened 2020-07-11
Add PlatformSpec for buildfile

loosebazooka:
part of #2570

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * A yaml block for specifying platforms
+ *
+ * <p>Example use of this yaml snippet.
+ *
+ * <pre>{@code
+ * architecture: amd64
+ * os: linux
+ * os.version: 1.0.0
+ * os.features:
+ *   - headless
+ * variant: amd64v10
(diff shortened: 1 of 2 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
```diff
@@ -0,0 +1,99 @@
...
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/**
+ * A yaml block for specifying platforms
```
chanseokoh, line 26:
```suggestion
 * A yaml block for specifying platforms.
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
```diff
@@ -0,0 +1,99 @@
...
+   * @param osFeatures a list of operating system features
+   */
+  @JsonCreator
+  public PlatformSpec(
+      @JsonProperty("architecture") String architecture,
+      @JsonProperty("os") String os,
+      @JsonProperty("os.version") String osVersion,
+      @JsonProperty("os.features") List<String> osFeatures,
+      @JsonProperty("variant") String variant,
+      @JsonProperty("features") List<String> features) {
```
chanseokoh, line 67: For every class in jib-build-plan, I copied collections to make it truly immutable. Should it do the same?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
```diff
@@ -0,0 +1,99 @@
...
+  public Optional<String> getOs() {
+    return Optional.ofNullable(os);
+  }
+
+  public Optional<String> getOsVersion() {
+    return Optional.ofNullable(osVersion);
+  }
+
+  public Optional<List<String>> getOsFeatures() {
+    return Optional.ofNullable(osFeatures);
```
chanseokoh, line 89: Likewise, it's an overhead, but every class in jib-build-plan returns a copy.

loosebazooka: Don't think users will be interacting with this code. Perhaps I could skip `Optional` as well? I see how it flows when I'm building the buildplan.

chanseokoh: Honestly, I am still unsure on this `Optional` topic and would love to get an expert guide. Generally speaking, getting rid of null (and null checks) by using `Optional` to denote something (especially for public API) is absent sounds like a good idea. However, sometimes it results in more bloated code that is really difficult to tolerate. And I'm also unclear if there should be a difference when using `Optional` as a return value versus as method arguments.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
chanseokoh: Saw the comment in the other thread. Marking resolved.

PR #2583, opened 2020-07-14
Fix integration tests

chanseokoh:
Distroless added one more layer, so fixing the assertions. Also pinning down an image with a digest.


chanseokoh:
Ah, wait. Need to fix Maven tests too.

PR #2584, opened 2020-07-14
Adding the platform config methods to the ContainerBuildPlan api

louismurerwa:
This PR adds the ` add / set/ get platform ` methods to the Container Builder Api .

### DIFF
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
@@ -75,6 +75,9 @@ public class ContainerBuildPlan {
      * @return this
      */
     public Builder addPlatform(String os, String architecture) {
+      if (platforms.isEmpty()) {
+        platforms.add(new Platform("linux", "amd64"));
+      }
       platforms.add(new Platform(os, architecture));
       return this;
     }
@@ -418,10 +421,10 @@ public class ContainerBuildPlan {
   }
 
   /**
-   * Creates and returns a default platform if the user doesn't specify a platform. Returns a list
-   * of platforms to be used in the build plan.
+   * Creates and returns a default platform if the user hasn't added or set any platforms ,else
+   * returns a list of user specified platforms .
    *
-   * @return platforms a list of desired platforms.
+   * @return platforms a list of user specified platforms.
    */
   public List<Platform> getPlatforms() {
     if (platforms.isEmpty()) {
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
@@ -418,9 +418,8 @@ public class ContainerBuildPlan {
   }
 
   /**
-   * Creates and adds the user specified platform to the platforms list if user opts to use
-   * setOsHint and setArchitectureHint methods.Else the method just returns the platforms list set
-   * by the user using addPlatform or setPlatforms methods.
+   * Creates and returns a default platform if the user doesn't specify a platform. Returns a list
+   * of platforms to be used in the build plan.
    *
    * @return platforms a list of desired platforms.
    */
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
@@ -102,7 +102,7 @@ public class ContainerBuildPlan {
      */
     public Builder setPlatforms(Set<Platform> platforms) throws IllegalArgumentException {
       if (platforms.isEmpty()) {
-        throw new IllegalArgumentException("platforms list cannot be empty.");
+        throw new IllegalArgumentException("platforms set cannot be empty.");
       }
       this.platforms = new LinkedHashSet<>(platforms);
       return this;
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
@@ -20,6 +20,7 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -34,9 +35,12 @@ public class ContainerBuildPlan {
   public static class Builder {
 
     private String baseImage = "scratch";
-    private List<Platform> platforms = new ArrayList<>();
     private Instant creationTime = Instant.EPOCH;
     private ImageFormat format = ImageFormat.Docker;
+
+    // note that a LinkedHashSet instead of HashSet has been used so as to preserve the platform
+    // order
+    private Set<Platform> platforms = new LinkedHashSet<>();
 
     // image execution parameters
     private Map<String, String> environment = new HashMap<>();
@@ -70,6 +74,9 @@ public class ContainerBuildPlan {
      * value is ignored and the platform of the built image follows that of the base image. The
      * default is {@code linux amd64 }.
      *
+     * <p>Note that the build plan starts with amd64/linux as the default platform. If you want to
+     * reset the default platform instead of adding a new one, use setPlatforms()"
+     *
      * @param os value to select a base image in case of a manifest list
      * @param architecture value to select a base image in case of a manifest list
      * @return this
@@ -94,12 +101,11 @@ public class ContainerBuildPlan {
      * @return this
      * @throws IllegalArgumentException if a user passes in an empty platform list
      */
-    public Builder setPlatforms(List<Platform> platforms) throws IllegalArgumentException {
+    public Builder setPlatforms(Set<Platform> platforms) throws IllegalArgumentException {
       if (platforms.isEmpty()) {
-        throw new IllegalArgumentException(
-            "platforms list cannot be empty.Please pass in a non-empty platforms list.");
+        throw new IllegalArgumentException("platforms list cannot be empty.");
       }
-      this.platforms = new ArrayList<>(platforms);
+      this.platforms = new LinkedHashSet<>(platforms);
       return this;
     }
 
@@ -371,7 +377,7 @@ public class ContainerBuildPlan {
   }
 
   private final String baseImage;
-  private final List<Platform> platforms;
+  private final Set<Platform> platforms;
   private final Instant creationTime;
   private final ImageFormat format;
 
@@ -389,7 +395,7 @@ public class ContainerBuildPlan {
 
   private ContainerBuildPlan(
       String baseImage,
-      List<Platform> platforms,
+      Set<Platform> platforms,
       Instant creationTime,
       ImageFormat format,
       Map<String, String> environment,
@@ -426,11 +432,11 @@ public class ContainerBuildPlan {
    *
    * @return platforms a list of user specified platforms.
    */
-  public List<Platform> getPlatforms() {
+  public Set<Platform> getPlatforms() {
     if (platforms.isEmpty()) {
       platforms.add(new Platform("linux", "amd64"));
     }
-    return new ArrayList<>(platforms);
+    return new LinkedHashSet<>(platforms);
   }
 
   public ImageFormat getFormat() {
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
@@ -34,8 +34,6 @@ public class ContainerBuildPlan {
   public static class Builder {
 
     private String baseImage = "scratch";
-    private String architectureHint = "amd64";
-    private String osHint = "linux";
     private List<Platform> platforms = new ArrayList<>();
     private Instant creationTime = Instant.EPOCH;
     private ImageFormat format = ImageFormat.Docker;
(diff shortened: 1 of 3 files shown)


jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/Platform.java
```diff
@@ -0,0 +1,37 @@
...
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.api.buildplan;
+
+import javax.annotation.Nonnull;
+
+public class Platform {
+  @Nonnull private String os;
+  @Nonnull private String architecture;
```
chanseokoh, line 23: We don't use `@Nonnull`. Everything not `@Nullable` is assumed non-null.

Also, always add `final` if field references never change.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/Platform.java
```diff
@@ -0,0 +1,37 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.api.buildplan;
+
+import javax.annotation.Nonnull;
+
+public class Platform {
```
chanseokoh, line 21: Add `@Immutable` as other classes.

Also our repo policy is to add a Javadoc for a public class. Can be a short one.

jib-build-plan/src/test/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlanTest.java
```diff
@@ -121,10 +127,12 @@ private ContainerBuildPlan createSamplePlan() {
         FileEntriesLayer.builder()
             .addEntry(Paths.get("/src/file/foo"), AbsoluteUnixPath.get("/path/in/container"))
             .build();
+    Platform platform = new Platform("os1", "arch1");
     return ContainerBuildPlan.builder()
         .setBaseImage("base/image")
         .setArchitectureHint("arch")
         .setOsHint("os")
+        .setPlatforms(Arrays.asList(platform))
```
chanseokoh, line 135: nit: let's just inline this.
```suggestion
        .setPlatforms(Arrays.asList(new Platform("os1", "arch1")))
```

chanseokoh:
@loosebazooka this PR will require more changes.

@louismurerwa it's a good practice to add a short summary in the PR description.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
```diff
@@ -422,6 +459,20 @@ public String getOsHint() {
...
+   * setOsHint and setArchitectureHint methods.Else the method just returns the platforms list set
+   * by the user using addPlatform or setPlatforms methods.
+   *
+   * @return platforms a list of desired platforms.
+   */
+  public List<Platform> getPlatforms() {
+    if (platforms.isEmpty()) {
+      platforms.add(new Platform(osHint, architectureHint));
+    }
+    return new ArrayList<>(platforms);
```
chanseokoh, line 473: We need to think about if this behavior is logical. For example,

- case 1
  1. Someone calls `getPlatforms()` at the beginning. It returns linux/amd64.
  2. they call `addPlatform(linux/arm64)`.
  3. `getPlatforms()` returns linux/am64.

- case 2
  1. Someone calls `setPlatforms( [ linux/amd64 ] )`.
  2. `getPlatforms()` returns linux/amd64.
  3. They call `addPlatform(linux/arm64)`.
  4. `getPlatforms()` returns [ linux/am64, linux/arm64 ].

I think it's a bit counter-intuitive and may become a surprise.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
```diff
@@ -93,6 +94,38 @@ public Builder setOsHint(String osHint) {
...
+     * the images matching the given platforms. If the base image reference is an image manifest,
+     * this value is ignored and the platform of the built image follows that of the base image. The
+     * default is {@code linux amd64 }.
+     *
+     * @param platforms a list of platform objects to be used to select base images in case of a
+     *     manifest list
+     * @return this
+     */
+    public Builder setPlatforms(List<Platform> platforms) {
+      this.platforms = new ArrayList<>(platforms);
```
chanseokoh, line 125: Should we guard against passing an empty input?

```java
    if (platforms.isEmpty()) {
      throw new IllegalArgumentException("some error message telling the caller that the list cannot be empty");
    }
```

If we disallow an empty list, it should also be documented in the Javadoc.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
louismurerwa: We need to think about if this behavior is logical. For example,

- case 1
  1. The user already knows that  linux/amd64 is a default platform so I think this is logical
  2. they call `addPlatform(linux/arm64)`.
  3. `getPlatforms()` returns linux/am64 - This is logical to some extent , if `getPlatforms()` returns [[linux/am64] ,[linux/amd64]] ` that means the user is always forced to build the ` [linux,amd64] ` even though he doesnt want it.

I agree with you that case 1 and case 2 are a bit illogical once you pair them but the current implementation of case 1 makes sense to some extent , we dont want the user to get stuck with the default platform.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
louismurerwa: Good Point !! I think its necessary to disallow empty lists !! It ensures that we always have a value in platforms.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/Platform.java
louismurerwa: Intresting !! Done

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
```diff
@@ -93,6 +94,38 @@ public Builder setOsHint(String osHint) {
...
+     * the image matching the given platform. If the base image reference is an image manifest, this
+     * value is ignored and the platform of the built image follows that of the base image. The
+     * default is {@code linux amd64 }.
+     *
+     * @param os value to select a base image in case of a manifest list
+     * @param architecture value to select a base image in case of a manifest list
+     * @return this
+     */
+    public Builder addPlatform(String os, String architecture) {
+      this.platforms.add(new Platform(os, architecture));
```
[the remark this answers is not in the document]
louismurerwa: Since we have decided to remove ` get/setOs/ArchitectureHint() ` all together  , I wont be working on the ` @deprecated ` feature.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
louismurerwa: @chanseokoh What are your thoughts? Should i make case 1 point 3 more like case 2 point 3?

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
chanseokoh: In our case, we have set up a system for checking null, so unless marked with `@Nullable`, our code assumes it's never null. Of course, because this is a public API and our internal null checking system can't prevent the user from passing null, but I think it's fine. They should understand that our API doesn't accept null unless marked with `@Nullable`.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
loosebazooka: How do we see people using getArchitectureHint/getOs now? And how do we see people using getPlatform?

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
loosebazooka: You can also use `Preconditions.checkArgument()`

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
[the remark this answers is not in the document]
chanseokoh: I was taking about qualifying `platforms` with the `this` keyword. This line can just be `platforms.add(...)`. You will have to qualify with `this` if there is another variable with the same name to distinguish between variables, but this isn't just a case.

And please everything related to `get/setOs/ArchitectureHint()` in this PR. We don't want to check in code to HEAD that is incomplete or temporarily broken. The assumption is the code at HEAD should work correctly at any point of time.

louismurerwa:
Some of the tests might fail because I removed ` get/set/Os/ArchHint ` which are also used in the file ` JibCOntainerBuildPlan `

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
chanseokoh: I expect no external users using get/setOs/ArchitecureHint as of now, since they have no effect.

Thinking about this, I think `addPlatform()` should work in a consistent manner regardless of whether it's called for the first time or the second time. That is, considering the nature of the method name "add", I think it makes sense to always add a new platform.

At the same time, I'd like to have amd64/linux as the default. We can clearly document in the Javadoc of `addPlatform()` something along the line of saying "note that the build plan starts with amd64/linux as the default platform. If you want to reset the default platform instead of adding a new one, use setPlatforms()" or something like that.

And I remember we had a discussion about handling duplicate values. I think we can take of it easily if we use `Set<Platform>` instead of `List<Platform>`. And this behavior should also be documented. (For an actual `Set` implementation, we want to choose `LinkedHashSet` instead of, e.g., `HashSet` in order to preserve the element order.) But just using `Set` doesn't automatically make it remove duplicates, because Java will always treat one `Platform` instance is different from another instance even if their OS and arch values are same. We need to tell Java that if two different `Platform` instances are essentially "equal" if their internal values are same. To do so, you need to implement `Platform.equals()` and `Platform.hashCode()`. Take a look at the sibling `FileEntry` class.

chanseokoh:
> Some of the tests might fail because I removed `get/set/Os/ArchHint` which are called in the file `JibCOntainerBuildPlan`

That won't, actually. Other modules use a published jib-build-plan, so this code change doesn't affect them.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
```diff
@@ -65,31 +64,42 @@ public Builder setBaseImage(String baseImage) {
...
+     *     manifest list
      * @return this
+     * @throws IllegalArgumentException if a user passes in an empty platform list
      */
-    public Builder setOsHint(String osHint) {
-      this.osHint = osHint;
+    public Builder setPlatforms(List<Platform> platforms) throws IllegalArgumentException {
+      if (platforms.isEmpty()) {
+        throw new IllegalArgumentException(
+            "platforms list cannot be empty.Please pass in a non-empty platforms list.");
```
chanseokoh, line 100:
```suggestion
            "platforms list cannot be empty.");
```

louismurerwa: Done

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
```diff
@@ -34,11 +35,13 @@
...
 
     private String baseImage = "scratch";
-    private String architectureHint = "amd64";
-    private String osHint = "linux";
     private Instant creationTime = Instant.EPOCH;
     private ImageFormat format = ImageFormat.Docker;
 
+    // note that a LinkedHashSet instead of HashSet has been used so as to preserve the platform
+    // order
+    private Set<Platform> platforms = new LinkedHashSet<>();
```
chanseokoh, line 43:
```java
    private Set<Platform> platforms = new LinkedHashSet<>(Collections.singleton(new Platform("linux", "amd64")));
```

You can then remove all the "if empty then add amd64/linux as a default" code in multiple places.

louismurerwa: Good Point

louismurerwa:
I pushed new updates to add add/getPlatforms test on a branch ` ContainerBuildPlan ` that I used to create this PR but the PR is not getting updated.Any ideas to solve that @loosebazooka ?

chanseokoh:
LGTM. I want to update Javadocs and add CHANGELOG, but I think I can do it later.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
```diff
@@ -65,31 +70,41 @@ public Builder setBaseImage(String baseImage) {
...
+     * @param platforms is a list of platform objects to be used to select base images in case of a
+     *     manifest list
      * @return this
+     * @throws IllegalArgumentException if a user passes in an empty platform list
      */
-    public Builder setOsHint(String osHint) {
-      this.osHint = osHint;
+    public Builder setPlatforms(Set<Platform> platforms) throws IllegalArgumentException {
+      if (platforms.isEmpty()) {
+        throw new IllegalArgumentException("platforms list cannot be empty.");
```
louismurerwa, line 105: ` platforms set cannot be empty ` - I will correct this

jib-build-plan/src/test/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlanTest.java
```diff
@@ -116,15 +115,37 @@ public void testToBuilder() {
...
+        ImmutableSet.of(new Platform("linux", "amd64"), new Platform("testOS", "testArchitecture")),
+        plan.getPlatforms());
+  }
+
+  @Test
+  public void testSetPlatforms_EmptyPlatformsSet() {
+    try {
+      ContainerBuildPlan.builder().setPlatforms(Collections.emptySet()).build();
+      Assert.fail();
+    } catch (IllegalArgumentException e) {
```
chanseokoh, line 135: e --> ex per Google Java style

PR #2587, opened 2020-07-15
Minor test refactoring

loosebazooka:
Just moves some things around, normalizes naming and creation of object mapper.

### DIFF
--- a/jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/TestObjectMapper.java
+++ b/jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/TestObjectMapper.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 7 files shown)


jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/TestObjectMapper.java
```diff
@@ -0,0 +1,31 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+
+/**
+ * Helper class to create object mappers for all tests that match what we use in the actual code.
+ */
+public class TestObjectMapper {
+
+  public static ObjectMapper newObjectMapper() {
+    // TODO: in future, create this mapper from the mapper creation in the main program
```
chanseokoh, line 28: I have trouble understanding what this is saying. (Specifically, I don't get what "mapper creation" and "main program" mean here.) And why would one want to do so?

loosebazooka: Valid question, for some restrictions/features, I believe you can only apply them on the mapper during creation and not on the class with an annotation.

I guess we still wouldn't need this test mapper, could probably just call the mapper creator from the tests directly.

loosebazooka:
This is not required anymore, closing.

PR #2590, opened 2020-07-16
Add layers spec

loosebazooka:
part of #2570

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayersSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayersSpec.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayersSpec.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class LayersSpec {
+  private List<LayerSpec> entries;
+  @Nullable private FilePropertiesSpec properties;
+
+  /**
+   * Constructor for use by jackson to populate this object.
+   *
+   * @param entries octal string for directory permissions
+   * @param properties octal string for file permissions
```
chanseokoh, line 45: Javadoc needs fixing.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/LayersSpec.java
```diff
@@ -0,0 +1,62 @@
...
+   */
+  @JsonCreator
+  public LayersSpec(
+      @JsonProperty(value = "entries", required = true) List<LayerSpec> entries,
+      @JsonProperty("properties") FilePropertiesSpec properties) {
```
chanseokoh, line 50: Unrelated, now I come to realize probably we can also use this same `@JsonCreator` pattern in our `jib-core`, right? I think there are -`Template` JSON classes that have a dummy 0-arg constructor in order to have another field-initializing constructor.

loosebazooka: I believe some of the templating stuff is also used to write down java -> json, not really clear to me if that matters yet, I'll take a look.

Issue #2585, opened 2020-07-15
Jib is not sending client certificate regardless of  the `javax.net.ssl.keyStore` property.


jacksierkstra:
**Environment**:
- *Jib version:* 2.4.0
- *Build tool:* Maven 
- *OS:* Windows 10

**Description of the issue**:
The issue has been filed before, but in https://github.com/googleapis/google-http-java-client/issues/904#issuecomment-564625366. The problem is that whenever you add the jvm property: **javax.net.ssl.keyStore**, it is not picked up by the eventual HttpClient. 

I did some research and it seems that `org.apache.http.conn.ssl.SSLConnectionSocketFactory#getSocketFactory()` is always called whenever the plugin is constructing a HttpClient. If `org.apache.http.conn.ssl.SSLConnectionSocketFactory#getSystemSocketFactory()` would be called, the property: **javax.net.ssl.keyStore** would be taken into consideration and therefore mutual SSL authentication would be possible.

The problem is that `org.apache.http.ssl.SSLContexts#createDefault()` is always constructing an `SSLContext` without a keymanager and/or trustmanager and that makes mutual SSL impossible.

**Expected behavior**:
Whenever I set the property: **javax.net.ssl.keyStore**, I expect that the keystore would be set on the `SSLContext`.

**Steps to reproduce**:

  1. Create a project with Jib Maven plugin. 
  2. Create a custom keystore.
  3. Run `mvn -Djavax.net.debug=ssl,handshake -Djavax.net.ssl.keyStore=/path/to/keystore -Djavax.net.ssl.keyStorePassword=changeit jib:build`

loosebazooka:
I'm not familiar with keystore stuff. Perhaps @chanseokoh can take a look when's back from vacation.

jacksierkstra:
Ok, no problem. To give you a little bit extra information: when `com.google.cloud.tools.jib.http.FailoverHttpClient` is constructed, and secure transport is being initiated, it uses this method:
```java
  private static HttpTransport getSecureHttpTransport() {
    // Do not use NetHttpTransport. It does not process response errors properly.
    // See https://github.com/google/google-http-java-client/issues/39
    //
    // A new ApacheHttpTransport needs to be created for each connection because otherwise HTTP
    // connection persistence causes the connection to throw NoHttpResponseException.
    return new ApacheHttpTransport();
  }
```
When `new ApacheHttpTransport()` is being called, it constructs a new default http client. The creation of that client looks as follows:
```java
[...]

PR #2592, opened 2020-07-16
Use SystemSocketFactory instead of the default factory for SSL socket factory

jacksierkstra:
Solves: https://github.com/GoogleContainerTools/jib/issues/2585

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
@@ -42,6 +42,7 @@ import java.util.function.Supplier;
 import javax.net.ssl.SSLException;
 import org.apache.http.conn.ssl.NoopHostnameVerifier;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
+import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.conn.SystemDefaultRoutePlanner;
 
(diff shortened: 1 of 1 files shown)


jacksierkstra:
@googlebot I signed it!

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -97,7 +101,17 @@ private static HttpTransport getSecureHttpTransport() {
...
+        HttpClientBuilder.create()
+            .useSystemProperties()
+            .setSSLSocketFactory(SSLConnectionSocketFactory.getSystemSocketFactory())
+            .setMaxConnTotal(200)
+            .setMaxConnPerRoute(20)
+            .setConnectionTimeToLive(-1, TimeUnit.MILLISECONDS)
+            .setRoutePlanner(new SystemDefaultRoutePlanner(ProxySelector.getDefault()))
+            .disableRedirectHandling()
+            .disableAutomaticRetries();
+    return new ApacheHttpTransport(httpClientBuilder.build());
```
chanseokoh, line 114: Thanks for your contribution!

As in `getInsecureHttpTransport()` below, can't this just be
```java
          ApacheHttpTransport.newDefaultHttpClientBuilder()
              .setSSLSocketFactory(SSLConnectionSocketFactory.getSystemSocketFactory());
```
? This is exactly [the reason for asking Google HTTP Client to add `newDefaultHttpClientBuilder()`](https://github.com/googleapis/google-http-java-client/issues/578) in the past.

jacksierkstra: Yes, that would work and is cleaner. I applied your suggestion.

chanseokoh:
Cool. The code is not formatted correctly. Try running `./gradlew googleJavaFormat`.

PR #2598, opened 2020-07-17
Add BuildFileSpec for buildfile

loosebazooka:
part of #2570

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
@@ -81,8 +81,8 @@ public class BuildFileSpec {
    * Constructor for use by jackson to populate this object.
    *
    * @param apiVersion the api version of this buildfile
-   * @param kind the kind (always BuildFile)
-   * @param from a {@link BaseImageSpec} for base image
+   * @param kind the type of configuration file (always BuildFile)
+   * @param from a {@link BaseImageSpec} for specifying the base image
    * @param creationTime in milliseconds since epoch or ISO 8601 datetime
    * @param format of the container, valid values in {@link ImageFormat}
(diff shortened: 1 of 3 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
```diff
@@ -0,0 +1,197 @@
...
+public class BuildFileSpec {
+  private String apiVersion;
+  private String kind;
+  @Nullable private BaseImageSpec from;
+  @Nullable private Instant creationTime;
+  @Nullable private ImageFormat format;
+  @Nullable private Map<String, String> environment;
+  @Nullable private Map<String, String> labels;
+  @Nullable private Set<AbsoluteUnixPath> volumes;
+  @Nullable private Set<Port> exposedPorts;
```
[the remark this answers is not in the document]
loosebazooka: So I'm working under the assumption that defaults come from the build plan. However, this is something we can discuss. Added to tracking bug.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
```diff
@@ -0,0 +1,197 @@
...
+  @Nullable private List<String> cmd;
+
+  @Nullable private LayersSpec layers;
+
+  /**
+   * Constructor for use by jackson to populate this object.
+   *
+   * @param apiVersion the api version of this buildfile
+   * @param kind the kind (always BuildFile)
+   * @param from a {@link BaseImageSpec} for base image
```
mpeddada1, line 85: is it possible to get a bit more context into these two variables?

loosebazooka: kind and from?

PR #2603, opened 2020-07-20
Processing Raw Platforms Configs From the Plugins

louismurerwa:
This PR processes the raw platforms configurations

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -28,6 +28,7 @@ import com.google.cloud.tools.jib.plugins.common.InvalidContainerVolumeException
 import com.google.cloud.tools.jib.plugins.common.InvalidContainerizingModeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidCreationTimeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidFilesModificationTimeException;
+import com.google.cloud.tools.jib.plugins.common.InvalidPlatformConfigurationException;
 import com.google.cloud.tools.jib.plugins.common.InvalidWorkingDirectoryException;
 import com.google.cloud.tools.jib.plugins.common.MainClassInferenceException;
 import com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor;
@@ -125,6 +126,11 @@ public class BuildDockerTask extends DefaultTask implements JibTask {
       throw new GradleException(
           "container.workingDirectory is not an absolute Unix-style path: "
               + ex.getInvalidPathValue(),
+          ex);
+
+    } catch (InvalidPlatformConfigurationException ex) {
+      throw new GradleException(
+          "container.platforms contains a null architecture or os name: " + ex.getInvalidPlatform(),
           ex);
 
     } catch (InvalidContainerVolumeException ex) {
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -130,7 +130,9 @@ public class BuildDockerTask extends DefaultTask implements JibTask {
 
     } catch (InvalidPlatformException ex) {
       throw new GradleException(
-          "from.platforms contains a platformConfiguration that is missing required fields or has invalid values: "
+          "from.platforms contains a platform configuration that is missing required values or has invalid values: "
+              + ex.getMessage()
+              + ": "
               + ex.getInvalidPlatform(),
           ex);
 
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -130,7 +130,8 @@ public class BuildDockerTask extends DefaultTask implements JibTask {
 
     } catch (InvalidPlatformConfigurationException ex) {
       throw new GradleException(
-          "container.platforms contains a null architecture or os name: " + ex.getInvalidPlatform(),
+          "from.platforms is missing required fields or has invalid values: "
+              + ex.getInvalidPlatform(),
           ex);
 
     } catch (InvalidContainerVolumeException ex) {
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
@@ -28,7 +28,7 @@ import com.google.cloud.tools.jib.plugins.common.InvalidContainerVolumeException
 import com.google.cloud.tools.jib.plugins.common.InvalidContainerizingModeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidCreationTimeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidFilesModificationTimeException;
-import com.google.cloud.tools.jib.plugins.common.InvalidPlatformConfigurationException;
+import com.google.cloud.tools.jib.plugins.common.InvalidPlatformException;
 import com.google.cloud.tools.jib.plugins.common.InvalidWorkingDirectoryException;
 import com.google.cloud.tools.jib.plugins.common.MainClassInferenceException;
 import com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor;
@@ -128,9 +128,9 @@ public class BuildDockerTask extends DefaultTask implements JibTask {
               + ex.getInvalidPathValue(),
           ex);
 
-    } catch (InvalidPlatformConfigurationException ex) {
+    } catch (InvalidPlatformException ex) {
       throw new GradleException(
-          "from.platforms is missing required fields or has invalid values: "
+          "from.platforms contains a platformConfiguration that is missing required fields or has invalid values: "
               + ex.getInvalidPlatform(),
           ex);
 
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -28,6 +28,7 @@ import com.google.cloud.tools.jib.plugins.common.InvalidContainerVolumeException
 import com.google.cloud.tools.jib.plugins.common.InvalidContainerizingModeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidCreationTimeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidFilesModificationTimeException;
+import com.google.cloud.tools.jib.plugins.common.InvalidPlatformConfigurationException;
 import com.google.cloud.tools.jib.plugins.common.InvalidWorkingDirectoryException;
 import com.google.cloud.tools.jib.plugins.common.MainClassInferenceException;
 import com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor;
@@ -106,7 +107,10 @@ public class BuildDockerMojo extends JibPluginConfiguration {
           "<container><workingDirectory> is not an absolute Unix-style path: "
               + ex.getInvalidPathValue(),
           ex);
-
+    } catch (InvalidPlatformConfigurationException ex) {
+      throw new MojoExecutionException(
+          "container.platforms contains a null architecture or os name: " + ex.getInvalidPlatform(),
+          ex);
     } catch (InvalidContainerVolumeException ex) {
       throw new MojoExecutionException(
           "<container><volumes> is not an absolute Unix-style path: " + ex.getInvalidVolume(), ex);
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -109,7 +109,9 @@ public class BuildDockerMojo extends JibPluginConfiguration {
           ex);
     } catch (InvalidPlatformException ex) {
       throw new MojoExecutionException(
-          "<from><platforms> contains a platformConfiguration that is missing required fields or has invalid values: "
+          "<from><platforms> contains a platform configuration that is missing required values or has invalid values: "
+              + ex.getMessage()
+              + ": "
               + ex.getInvalidPlatform(),
           ex);
     } catch (InvalidContainerVolumeException ex) {
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -28,7 +28,7 @@ import com.google.cloud.tools.jib.plugins.common.InvalidContainerVolumeException
 import com.google.cloud.tools.jib.plugins.common.InvalidContainerizingModeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidCreationTimeException;
 import com.google.cloud.tools.jib.plugins.common.InvalidFilesModificationTimeException;
-import com.google.cloud.tools.jib.plugins.common.InvalidPlatformConfigurationException;
+import com.google.cloud.tools.jib.plugins.common.InvalidPlatformException;
 import com.google.cloud.tools.jib.plugins.common.InvalidWorkingDirectoryException;
 import com.google.cloud.tools.jib.plugins.common.MainClassInferenceException;
 import com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor;
@@ -107,9 +107,9 @@ public class BuildDockerMojo extends JibPluginConfiguration {
           "<container><workingDirectory> is not an absolute Unix-style path: "
               + ex.getInvalidPathValue(),
           ex);
-    } catch (InvalidPlatformConfigurationException ex) {
+    } catch (InvalidPlatformException ex) {
       throw new MojoExecutionException(
-          "<from><platforms> is missing required fields or has invalid values: "
+          "<from><platforms> contains a platformConfiguration that is missing required fields or has invalid values: "
               + ex.getInvalidPlatform(),
           ex);
     } catch (InvalidContainerVolumeException ex) {
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
@@ -109,7 +109,8 @@ public class BuildDockerMojo extends JibPluginConfiguration {
           ex);
     } catch (InvalidPlatformConfigurationException ex) {
       throw new MojoExecutionException(
-          "container.platforms contains a null architecture or os name: " + ex.getInvalidPlatform(),
+          "<from><platforms> is missing required fields or has invalid values: "
+              + ex.getInvalidPlatform(),
           ex);
     } catch (InvalidContainerVolumeException ex) {
       throw new MojoExecutionException(
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidPlatformConfigurationException.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidPlatformConfigurationException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Google LLC.
+ * Copyright 2020 Google LLC.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -18,11 +18,11 @@ package com.google.cloud.tools.jib.plugins.common;
 
 /**
  * Indicates that the {@code container.platforms} config value has at least one invalid platform.
- * (The platform misses either an architecture value or os value or both).
+ * (For example, a platform misses a required field or has an invalid value.).
  */
 public class InvalidPlatformConfigurationException extends Exception {
 
-  private String platform;
+  private final String platform;
 
   InvalidPlatformConfigurationException(String message, String platform, Throwable cause) {
     super(message, cause);
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidPlatformConfigurationException.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidPlatformConfigurationException.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.common;
+
+/**
+ * Indicates that the {@code container.platforms} config value has at least one invalid platform.
+ * (The platform misses either an architecture value or os value or both).
+ */
+public class InvalidPlatformConfigurationException extends Exception {
+
+  private String platform;
+
+  InvalidPlatformConfigurationException(String message, String platform, Throwable cause) {
+    super(message, cause);
+    this.platform = platform;
+  }
+
+  public String getInvalidPlatform() {
+    return platform;
+  }
+}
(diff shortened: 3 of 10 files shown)


jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -104,11 +107,34 @@ private static void assertExtractionPathsUnordered(
   @Mock private AuthProperty authProperty;
   @Mock private Consumer<LogEvent> logger;
 
+  /** Configuration for {@code platform} parameter. */
+  public static class PlatformParametersTest implements PlatformConfiguration {
```
chanseokoh, line 111: No need for a comment. And this class isn't going to be used outside of this test class, you don't want to make it `public`. And the name `-Test` doesn't sound right.

```suggestion
  private static class TestPlatformConfiguration implements PlatformConfiguration {
```

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -104,11 +107,34 @@ private static void assertExtractionPathsUnordered(
...
+      return Optional.ofNullable(architecture);
+    }
+  }
+
   @Before
   public void setUp() throws IOException, InvalidImageReferenceException, InferredAuthException {
     Mockito.when(rawConfiguration.getFromAuth()).thenReturn(authProperty);
     Mockito.when(rawConfiguration.getEntrypoint()).thenReturn(Optional.empty());
     Mockito.when(rawConfiguration.getAppRoot()).thenReturn("/app");
+    Mockito.<List<? extends PlatformConfiguration>>when(rawConfiguration.getPlatforms())
```
chanseokoh, line 136: Probably `Mockito.when(...)` will just work.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -837,6 +874,31 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
       Assert.assertEquals(12, ex.getProjectMajorJavaVersion());
     }
   }
+  //
```
chanseokoh, line 877:
```suggestion

```

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -837,6 +874,31 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
...
+  @Test
+  public void testGetPlatformsSet() throws InvalidPlatformConfigurationException {
+    Mockito.<List<? extends PlatformConfiguration>>when(rawConfiguration.getPlatforms())
```
chanseokoh, line 880: ditto

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -837,6 +874,31 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
...
+
+    Assert.assertEquals(
+        ImmutableSet.of(new Platform("amd64", "linux")),
+        PluginConfigurationProcessor.getPlatformsSet(rawConfiguration));
+  }
+
+  @Test
+  public void testInvalidPlatformsSet() throws InvalidPlatformConfigurationException {
+    PlatformParametersTest platform = new PlatformParametersTest(null, "linux");
+    Mockito.<List<? extends PlatformConfiguration>>when(rawConfiguration.getPlatforms())
```
chanseokoh, line 891: ditto

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -104,11 +107,34 @@ private static void assertExtractionPathsUnordered(
   @Mock private AuthProperty authProperty;
   @Mock private Consumer<LogEvent> logger;
 
+  /** Configuration for {@code platform} parameter. */
+  public static class PlatformParametersTest implements PlatformConfiguration {
+    @Nullable private String os;
+    @Nullable private String architecture;
+
+    PlatformParametersTest(String architecture, String os) {
```
chanseokoh, line 115:
```suggestion
    PlatformParametersTest(@Nullable String architecture, @Nullable String os) {
```

Although we don't strictly enforce using `@Nullable annotation in test classes, let's annotate these properly since you used `@Nullable` for the fields.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidPlatformConfigurationException.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2018 Google LLC.
```
chanseokoh, line 2: 2020

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -837,6 +874,31 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
...
+    Mockito.<List<? extends PlatformConfiguration>>when(rawConfiguration.getPlatforms())
+        .thenReturn(Arrays.asList(platform));
+
+    try {
+      PluginConfigurationProcessor.getPlatformsSet(rawConfiguration);
+      Assert.fail();
+    } catch (InvalidPlatformConfigurationException ex) {
+      Assert.assertEquals(platform.toString(), ex.getMessage());
+      Assert.assertEquals(platform.toString(), ex.getInvalidPlatform());
+    }
```
chanseokoh, line 900: Don't use `platform.toString()` but use an actual `String` literal to verify the returned value.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -127,6 +128,11 @@ public void buildDocker()
               + ex.getInvalidPathValue(),
           ex);
 
+    } catch (InvalidPlatformConfigurationException ex) {
+      throw new GradleException(
+          "container.platforms contains a null architecture or os name: " + ex.getInvalidPlatform(),
```
chanseokoh, line 133: How about `from.platform misses a required field or has an invalid value: `?

Note it's `from.platform`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -106,7 +107,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
           "<container><workingDirectory> is not an absolute Unix-style path: "
               + ex.getInvalidPathValue(),
           ex);
-
+    } catch (InvalidPlatformConfigurationException ex) {
+      throw new MojoExecutionException(
+          "container.platforms contains a null architecture or os name: " + ex.getInvalidPlatform(),
```
chanseokoh, line 112: `<from><platforms>` We use brackets for Maven.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -611,6 +627,38 @@ static String getDefaultBaseImage(ProjectProperties projectProperties)
...
+        Preconditions.checkNotNull(
+            platformConfiguration.getOsName(), "platforms set cannot be empty");
+        Platform platform =
+            new Platform(
+                platformConfiguration.getArchitectureName().orElse(null),
+                platformConfiguration.getOsName().orElse(null));
+        platforms.add(platform);
+      } catch (IllegalArgumentException exception) {
+        throw new InvalidPlatformConfigurationException(
+            platformConfiguration.toString(), platformConfiguration.toString(), exception);
```
chanseokoh, line 655: `platformConfiguration.toString()` will returns some weird string value (for example, `PlatformConfiguration@6d06d69c`), unless the class (or any of its super-classes) overrode and implemented `toString()`.

And I think we don't want to implement `toString()` in `PlatformConfiguration` implementations. So probably we should assemble a human-readable string form here. For example,

`"architecture=" + platformConfiguration.getArchitectureName().orElse("<missing>") + ", os=" + platformConfiguration.getOsName().orElse("<missing>")`

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -838,6 +875,31 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
     }
   }
 
+  @Test
+  public void testGetValidPlatformsList() throws InvalidPlatformConfigurationException {
```
chanseokoh, line 879: testGetPlatformsSet()

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -838,6 +875,31 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
...
+    Mockito.<List<? extends PlatformConfiguration>>when(rawConfiguration.getPlatforms())
+        .thenReturn(Arrays.asList(new PlatformParametersTest("testArchitecture", "testOs")));
+
+    Assert.assertEquals(
+        ImmutableSet.of(new Platform("testArchitecture", "testOs")),
+        PluginConfigurationProcessor.getPlatformsSet(rawConfiguration));
+  }
+
+  @Test
+  public void testInvalidPlatformsList() throws InvalidPlatformConfigurationException {
```
chanseokoh, line 889: testGetPlatformsSet_osMissing()

That said, I'd add testGetPlatformsSet_architectureMissing() as well.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -92,6 +95,8 @@
    * @throws IOException if an error occurs creating the container builder
    * @throws InvalidWorkingDirectoryException if the working directory specified for the build is
    *     invalid
+   * @throws InvalidPlatformConfigurationException if there exists a plaformConfiguration with a
+   *     null architectureName or osName
```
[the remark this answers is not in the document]
louismurerwa: I think this sounds better since a platformConfig is the one missing values .

` if there exists a platformConfiguration in the platforms list that is missing required fields or has invalid values `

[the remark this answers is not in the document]
chanseokoh: Sounds good. I'd go with `... exists a {@link PlatformConfiguration} ...` in that case.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
louismurerwa: Done

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
louismurerwa: Done

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
louismurerwa: I see what you mean .I agree with you on the string implementation. Done

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -104,11 +107,34 @@ private static void assertExtractionPathsUnordered(
   @Mock private AuthProperty authProperty;
   @Mock private Consumer<LogEvent> logger;
 
+  /** Configuration for {@code platform} parameter. */
+  public static class PlatformParametersTest implements PlatformConfiguration {
+    @Nullable private String os;
+    @Nullable private String architecture;
```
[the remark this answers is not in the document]
chanseokoh: Sorry, I initially removed `@Nullable` as I mistakenly thought they were just copied blindly. Updated my comment to bring them back.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
louismurerwa: We need to chat in person about this implementation ,I might some help to change it.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
chanseokoh: Oh, I just noticed that you can remove the `throws InvalidPlatformConfigurationException` clause from the method signature.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
chanseokoh: Ah, just realized we can't remove it. I think in our test code, we can just say `Mockito.<List<?>>`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -611,6 +628,52 @@ static String getDefaultBaseImage(ProjectProperties projectProperties)
...
+          "architecture="
+              + platformConfiguration.getArchitectureName().orElse("<missing>")
+              + ", os="
+              + platformConfiguration.getOsName().orElse("<missing>");
+
+      if (!platformConfiguration.getArchitectureName().isPresent()) {
+        throw new InvalidPlatformConfigurationException(
+            "platform configuration is missing an architecture value",
+            platformToString,
+            new IllegalArgumentException());
```
louismurerwa, line 655: Is this correct way of throwing a `InvalidPlatformConfigurationException` when I didnt catch any exceptions?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
louismurerwa: The moment I remove it , the file raises the 'unhandled exception error `

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
louismurerwa: Gotcha

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -611,6 +627,38 @@ static String getDefaultBaseImage(ProjectProperties projectProperties)
...
+      try {
+        Preconditions.checkNotNull(
+            platformConfiguration.getArchitectureName(), "platforms set cannot be empty");
+        Preconditions.checkNotNull(
+            platformConfiguration.getOsName(), "platforms set cannot be empty");
+        Platform platform =
+            new Platform(
+                platformConfiguration.getArchitectureName().orElse(null),
+                platformConfiguration.getOsName().orElse(null));
+        platforms.add(platform);
```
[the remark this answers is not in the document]
louismurerwa: Nice that worked

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -611,6 +627,38 @@ static String getDefaultBaseImage(ProjectProperties projectProperties)
...
+   *     null architectureName or osName
+   */
+  @VisibleForTesting
+  static Set<Platform> getPlatformsSet(RawConfiguration rawConfiguration)
+      throws InvalidPlatformConfigurationException {
+    Set<Platform> platforms = new LinkedHashSet<>();
+    for (PlatformConfiguration platformConfiguration : rawConfiguration.getPlatforms()) {
+      try {
+        Preconditions.checkNotNull(
+            platformConfiguration.getArchitectureName(), "platforms set cannot be empty");
```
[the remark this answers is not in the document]
louismurerwa: What `exception` should i pass to ` InvalidPlatformConfigurationException ` since I didnt catch any here ?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: We can remove `Throwable cause` from the contructor, as we don't get any exception from another party.
```java
  InvalidPlatformConfigurationException(String message, String platform, Throwable cause) {
```
And for consistency, let's rename the exception class to `InvalidPlatformException`. And when you rename it, you need to update all the Javadocs too.

And because now we are ensure that `getArchitectureName()` and `getOsName()` are always present, there is no need to catch `NoSuchElementException` that will never be thrown. Remove the `try-catch` block below.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -611,6 +628,52 @@ static String getDefaultBaseImage(ProjectProperties projectProperties)
...
+
+      if (!platformConfiguration.getArchitectureName().isPresent()) {
+        throw new InvalidPlatformConfigurationException(
+            "platform configuration is missing an architecture value",
+            platformToString,
+            new IllegalArgumentException());
+      }
+      if (!platformConfiguration.getOsName().isPresent()) {
+        throw new InvalidPlatformConfigurationException(
+            "platform configuration is missing an os value",
```
chanseokoh, line 659: nit: os --> OS

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -838,6 +874,58 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
...
+
+    try {
+      PluginConfigurationProcessor.getPlatformsSet(rawConfiguration);
+      Assert.fail();
+    } catch (InvalidPlatformConfigurationException ex) {
+      String message =
+          "architecture="
+              + platform.getArchitectureName().orElse("<missing>")
+              + ", os="
+              + platform.getOsName().orElse("<missing>");
```
chanseokoh, line 923: `message = "architecture=testArchitecture, os=<missing>"` since we are verifying the return value of `ex.getInvalidPlatform()`.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -838,6 +874,58 @@ public void testGetJavaContainerBuilderWithBaseImage_java12NoBaseImage()
...
+      throws InvalidPlatformConfigurationException {
+    TestPlatformConfiguration platform = new TestPlatformConfiguration(null, "testOs");
+    Mockito.<List<? extends PlatformConfiguration>>when(rawConfiguration.getPlatforms())
+        .thenReturn(Arrays.asList(platform));
+
+    try {
+      PluginConfigurationProcessor.getPlatformsSet(rawConfiguration);
+      Assert.fail();
+    } catch (InvalidPlatformConfigurationException ex) {
+      String message =
```
chanseokoh, line 898: ditto

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
chanseokoh: With the latest commit, remove it from `testGetPlatformsSet_architectureMissing` and `testGetPlatformsSet_osMissing`.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
louismurerwa: That implemantain works

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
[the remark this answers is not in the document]
louismurerwa: Since we are testing the PlatformConfiguration which might contain nulls I had made the os and architecture ` @Nullable `

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
louismurerwa: I ended up changing the constructor

louismurerwa:
I believe we need one more test to check if the `getOSName()` and `getArchitectureName` are valid platform variables. For example is a user passes if ` osname = louis ` ,we should throw an `invalid platform exception `.

This check is hard to implement because it reqires us to know all the possible arch/os names and store them in a data structure which we use to cross check the user input

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -127,6 +128,12 @@ public void buildDocker()
               + ex.getInvalidPathValue(),
           ex);
 
+    } catch (InvalidPlatformException ex) {
+      throw new GradleException(
+          "from.platforms contains a platformConfiguration that is missing required fields or has invalid values: "
```
chanseokoh, line 133: So, I see we have a difference from other `Invalid*Exception` in that we provide better exception messages when instantiating `InvalidPlatformException`. Then we are not making use of the exception message. In this case, I think we can provide a more helpful and detailed message by

"from.platforms contains a platform configuration that is missing required values or has invalid values: " + ex.getMessage() + ": " + ex.getInvalidPlatform()

Note "platformConfiguration" --> "platform configuration". The message will be shown to the end user, and `platformConfiguration` is an internal Java variable name in our code that the user doesn't know about. Similarly, I think "required values" is more appropriate for the end user.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/BuildDockerMojo.java
```diff
@@ -106,7 +107,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
           "<container><workingDirectory> is not an absolute Unix-style path: "
               + ex.getInvalidPathValue(),
           ex);
-
+    } catch (InvalidPlatformException ex) {
+      throw new MojoExecutionException(
+          "<from><platforms> contains a platformConfiguration that is missing required fields or has invalid values: "
```
chanseokoh, line 112: same for Maven

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/InvalidPlatformException.java
```diff
@@ -0,0 +1,35 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.plugins.common;
+
+/**
+ * Indicates that the {@code container.platforms} config value has at least one invalid platform.
+ * (For example, a platform misses a required field or has an invalid value.).
```
chanseokoh, line 21:
```suggestion
 * (For example, a platform misses a required field or has an invalid value.)
```

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -104,11 +107,33 @@ private static void assertExtractionPathsUnordered(
   @Mock private AuthProperty authProperty;
   @Mock private Consumer<LogEvent> logger;
 
+  private static class TestPlatformConfiguration implements PlatformConfiguration {
+    @Nullable private final String os;
+    @Nullable private final String architecture;
+
+    TestPlatformConfiguration(String architecture, String os) {
```
chanseokoh, line 114:
```suggestion
    private TestPlatformConfiguration(@Nullable String architecture, @Nullable String os) {
```

Issue #2602, opened 2020-07-20
make our dependency version stuff more friendly to dependabot


loosebazooka:
Right now in `build.gradle` we use [dependency versions](https://github.com/GoogleContainerTools/jib/blob/master/build.gradle#L51-L81).
If we use full dependency data, maybe the dependabot would work better.

```
project.ext.dependencyVersions = [
  GOOGLE_HTTP_CLIENT: '1.34.0',
  ...
```

could maybe be 

```
project.ext.dependencyStrings = [
  GOOGLE_HTTPS_CLIENT: "com.google.http-client:google-http-client:1.34.0"
  ...
]
```

We would probably need a splitter on `:` to make sure [`ensureTransitiveDependencyOverrides`](https://github.com/GoogleContainerTools/jib/blob/master/build.gradle#L85-L102) still works.

mpeddada1:
Fixed with #2602

PR #2605, opened 2020-07-21
Modify dependencies info to include the full dependency path instead of just the version

mpeddada1:
Modify the main build.gradle and related build files to reference the full dependency data instead of just the version. This change attempts to make our dependency info more friendly with dependabot, which is currently unable to parse through our dependencies.
For #2602

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,44 +48,44 @@ subprojects {
 
   /* PROJECT DEPENDENCY VERSIONS */
   // define all common versioned dependencies here
-  project.ext.dependencyVersions = [
+  project.ext.dependencyStrings = [
     // For Google libraries, check <http-client-bom.version>, <google.auth.version>, <guava.version>,
     // ... in https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/pom.xml
     // for best compatibility.
-    GOOGLE_HTTP_CLIENT: '1.34.0',
-    GOOGLE_HTTP_CLIENT_APACHE_V2: '1.34.0',
-    GOOGLE_AUTH_LIBRARY_OAUTH2_HTTP: '0.18.0',
-    GUAVA: '28.1-jre',
-    JSR305: '3.0.2', // transitively pulled in by GUAVA
+    GOOGLE_HTTP_CLIENT: "com.google.http-client:google-http-client:1.34.0",
+    GOOGLE_HTTP_CLIENT_APACHE_V2: "com.google.http-client:google-http-client-apache-v2:1.34.0",
+    GOOGLE_AUTH_LIBRARY_OAUTH2_HTTP: "com.google.auth:google-auth-library-oauth2-http:0.18.0",
+    GUAVA: "com.google.guava:guava:28.1-jre",
+    JSR305: "com.google.code.findbugs:jsr305:3.0.2", // transitively pulled in by GUAVA
 
     // for Build Plan and Jib Plugins Extension API
-    BUILD_PLAN: '0.3.1',
-    EXTENSION_COMMON: '0.2.0',
-    GRADLE_EXTENSION: '0.3.0',
-    MAVEN_EXTENSION: '0.3.0',
+    BUILD_PLAN: "com.google.cloud.tools:jib-build-plan:0.3.1",
+    EXTENSION_COMMON: "com.google.cloud.tools:jib-plugins-extension-common:0.2.0",
+    GRADLE_EXTENSION: "com.google.cloud.tools:jib-gradle-plugin-extension-api:0.3.0",
+    MAVEN_EXTENSION: "com.google.cloud.tools:jib-maven-plugin-extension-api:0.3.0",
 
-    COMMONS_COMPRESS: '1.19',
-    JACKSON_DATABIND: '2.11.0',
-    JACKSON_DATAFORMAT_YAML: '2.11.0',
-    ASM: '7.3.1',
-    PICOCLI: '4.2.0',
+    COMMONS_COMPRESS: "org.apache.commons:commons-compress:1.19",
+    JACKSON_DATABIND: "com.fasterxml.jackson.core:jackson-databind:2.11.0",
+    JACKSON_DATAFORMAT_YAML: "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.11.0",
+    ASM: "org.ow2.asm:asm:7.3.1",
+    PICOCLI: "info.picocli:picocli:4.2.0",
 
-    MAVEN_API: '3.5.2',
+    MAVEN_API: "org.apache.maven:maven-plugin-api:3.5.2",
 
     //test
-    JUNIT: '4.13',
-    MOCKITO_CORE: '3.2.4',
-    SLF4J_API: '1.7.25',
-    SYSTEM_RULES: '1.19.0',
-    JBCRYPT: '0.4',
+    JUNIT: "junit:junit:4.13",
+    MOCKITO_CORE: "org.mockito:mockito-core:3.2.4",
+    SLF4J_API: "org.slf4j:slf4j-api:1.7.25",
+    SYSTEM_RULES:  "com.github.stefanbirkner:system-rules:1.19.0",
+    JBCRYPT: "org.mindrot:jbcrypt:0.4",
   ]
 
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -48,44 +48,46 @@ subprojects {
...
+  project.ext.dependencyStrings = [
     // For Google libraries, check <http-client-bom.version>, <google.auth.version>, <guava.version>,
     // ... in https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/pom.xml
     // for best compatibility.
-    GOOGLE_HTTP_CLIENT: '1.34.0',
-    GOOGLE_HTTP_CLIENT_APACHE_V2: '1.34.0',
-    GOOGLE_AUTH_LIBRARY_OAUTH2_HTTP: '0.18.0',
-    GUAVA: '28.1-jre',
-    JSR305: '3.0.2', // transitively pulled in by GUAVA
+    GOOGLE_HTTP_CLIENT: "com.google.http-client:google-http-client:1.34.0",
```
chanseokoh, line 55: Really super-nit, but we try to maintain consistency regarding `'` and `"`. In Gradle (which is Groovy language), we use `'` for static strings like this one, while using `"` for strings containing expressions to evaluate (such as `"some text with a ${variable}"`).

build.gradle
```diff
@@ -48,44 +48,46 @@ subprojects {
...
+    JBCRYPT: "org.mindrot:jbcrypt:0.4",
   ]
 
   // Use this to ensure we correctly override transitive dependencies
   // TODO: There might be a plugin that does this
   task ensureTransitiveDependencyOverrides {
     def rules = [
-        'google-http-client': dependencyVersions.GOOGLE_HTTP_CLIENT,
-        'google-http-client-apache-v2': dependencyVersions.GOOGLE_HTTP_CLIENT_APACHE_V2,
+        'google-http-client': dependencyStrings.GOOGLE_HTTP_CLIENT.split(':')[-1],
```
loosebazooka, line 89: I wonder if this name `google-https-client` should be derived from the `GOOGLE_HTTP_CLIENT` value??

Maybe you could use a closure to create a map from a list?
rules = [GOOGLE_HTTP_CLIENT, GOOGLE_HTTP_CLIENT_APACHE_V2] .collectEntries {
  [calculate name: calculate version]
}
```

I'm only vaguely familiar with groovy transforms, might be worth looking up

File: jib-maven-plugin/build.gradle [hunk 3]
```diff
@@ -12,10 +12,10 @@ dependencies {
...
-  implementation "com.google.cloud.tools:jib-maven-plugin-extension-api:${dependencyVersions.MAVEN_EXTENSION}"
+  implementation dependencyStrings.MAVEN_EXTENSION
 
-  implementation "org.apache.maven:maven-plugin-api:${dependencyVersions.MAVEN_API}"
-  implementation "org.apache.maven:maven-core:${dependencyVersions.MAVEN_API}"
+  implementation dependencyStrings.MAVEN_API
+  implementation dependencyStrings.MAVEN_CORE
 
   // compileOnly + testImplementation equivalent to "provided"
   compileOnly 'org.apache.maven.plugin-tools:maven-plugin-annotations:3.6.0'
```
loosebazooka, line 21: It might be worth moving all these values into the root build.gradle?

build.gradle
mpeddada1: Oh I didn't know you could do that, thanks!

jib-maven-plugin/build.gradle
mpeddada1: Done

build.gradle
```diff
@@ -52,43 +52,50 @@ subprojects {
...
+    JACKSON_DATAFORMAT_YAML: 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.11.0',
+    ASM: 'org.ow2.asm:asm:7.3.1',
+    PICOCLI: 'info.picocli:picocli:4.2.0',
+
+    MAVEN_API: 'org.apache.maven:maven-plugin-api:3.5.2',
+    MAVEN_CORE: 'org.apache.maven:maven-core:3.5.2',
+    MAVEN_COMPAT: 'org.apache.maven:maven-compat:3.5.2',
+    MAVEN_ANNOTATIONS_LATEST: 'org.apache.maven.plugin-tools:maven-plugin-annotations:3.6.0',
+    MAVEN_ANNOTATIONS_OLD: 'org.apache.maven.plugin-tools:maven-plugin-annotations:3.5',
+    MAVEN_TESTING_HARNESS: 'org.apache.maven.plugin-testing:maven-plugin-testing-harness:3.3.0',
```
loosebazooka, line 78: can we move 78-79 to the test commented block below?

build.gradle
```diff
@@ -52,43 +52,50 @@ subprojects {
...
 
   // Use this to ensure we correctly override transitive dependencies
   // TODO: There might be a plugin that does this
   task ensureTransitiveDependencyOverrides {
-    def rules = [
-        'google-http-client': dependencyStrings.GOOGLE_HTTP_CLIENT.split(':')[-1],
-        'google-http-client-apache-v2': dependencyStrings.GOOGLE_HTTP_CLIENT_APACHE_V2.split(':')[-1],
-    ]
+    def dependenciesList = [dependencyStrings.GOOGLE_HTTP_CLIENT, dependencyStrings.GOOGLE_HTTP_CLIENT_APACHE_V2]
+    def rules = dependenciesList.collectEntries{[/*name: */ it.split(':')[1], /*version: */ it.split(':')[-1]]}
```
loosebazooka, line 98: I think using `[2]` instead of `[-1]` will help us catch bugs better?

build.gradle
mpeddada1: Thanks for pointing this out! Fixed the style issue.

jib-maven-plugin/build.gradle
```diff
@@ -12,32 +12,32 @@ dependencies {
...
+  testImplementation dependencyStrings.SLF4J_API_OLD
+  testImplementation dependencyStrings.SYSTEM_RULES
 
-  testImplementation 'org.apache.maven.shared:maven-verifier:1.7.2'
-  testImplementation "org.apache.maven:maven-compat:${dependencyVersions.MAVEN_API}"
-  testImplementation 'org.slf4j:slf4j-api:1.7.30'
-  testImplementation 'org.slf4j:slf4j-simple:1.7.30'
+  testImplementation dependencyStrings.MAVEN_VERIFIER
+  testImplementation dependencyStrings.MAVEN_COMPAT
+  testImplementation dependencyStrings.SLF4J_API_LATEST
```
loosebazooka, line 37: hrmm weird that these are both in here, can you file a new issue to handle standardizing the SLF4J apis?

mpeddada1: Sure! Filed issue #2608

build.gradle
mpeddada1: Good point, done.

build.gradle
mpeddada1: Done.

build.gradle
```diff
@@ -48,45 +48,52 @@ subprojects {
...
 
   // Use this to ensure we correctly override transitive dependencies
   // TODO: There might be a plugin that does this
   task ensureTransitiveDependencyOverrides {
-    def rules = [
-        'google-http-client': dependencyVersions.GOOGLE_HTTP_CLIENT,
-        'google-http-client-apache-v2': dependencyVersions.GOOGLE_HTTP_CLIENT_APACHE_V2,
-    ]
+    def dependenciesList = [dependencyStrings.GOOGLE_HTTP_CLIENT, dependencyStrings.GOOGLE_HTTP_CLIENT_APACHE_V2]
+    def rules = dependenciesList.collectEntries{[/*name: */ it.split(':')[1], /*version: */ it.split(':')[2]]}
```
loosebazooka, line 96: Sorry this seems like python style comments, which is fine.
I think `/*name*/` , `/*version*/` is enough?

mpeddada1: Oh sorry, I removed the colon from the comment. Let me know if you want the comment to be moved outside of `collectEntries`

loosebazooka:
Not sure why the travis builds aren't coming through? I'll re-trigger them

PR #2606, opened 2020-07-22
Add validation to required yaml properties

loosebazooka:
- Only check required values
- Add validator class with standard error messages
- Add in tests for value validation

Why not `@JsonInclude`?
Because `@JsonCreator` and `@JsonInclude` do not work at all
with eachother. We use `@JsonCreator` since it handles **required** fields
and we can validate non-null, non-empty at creation time. Using `@JsonInclude`
with jackson bean initialized of fields (vs creator) would give us access
to automatic non-null, non-empty validation, but we would have to push
our own **required** infrastructure. This cannot happen at yaml parse time, and
would happen after parsing in some sort of post-parsing step (or there is some
jackson validation framework that I was unable to find). I find the
`@JsonCreator` workflow slightly cleaner.

part of #2570

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Validator.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Validator.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import com.google.common.base.Preconditions;
+import java.util.Collection;
+
+/**
+ * Utility helper class to detect errors in parsed yaml values. This class is mostly concerned with
+ * error message formatting, checking is delegated to guava.
+ */
+public class Validator {
+
+  /**
+   * Checks if an object reference is not null.
+   *
+   * @param value the object in question
+   * @param propertyName the equivalent 'yaml' property name
+   * @throws NullPointerException if {@code value} is null
+   */
+  public static void checkNotNull(Object value, String propertyName) {
+    Preconditions.checkNotNull(value, "Property '" + propertyName + "' cannot be null");
+  }
+
+  /**
+   * Checks if string is null, empty or only whitespace.
+   *
+   * @param value the string in question
+   * @param propertyName the equivalent 'yaml' property name
+   * @throws NullPointerException if {@code value} is null
+   * @throws IllegalArgumentException if {@code value} is empty or only whitespace
+   */
+  public static void checkNotEmpty(String value, String propertyName) {
+    checkNotNull(value, propertyName);
+    Preconditions.checkArgument(
+        !value.trim().isEmpty(), "Property '" + propertyName + "' cannot be empty");
+  }
+
+  /**
+   * Checks if a collection is null or empty.
+   *
+   * @param value the string in question
+   * @param propertyName the equivalent 'yaml' property name
+   * @throws NullPointerException if {@code value} is null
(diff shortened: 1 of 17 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Validator.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Checks if an object reference is not null.
+   *
+   * @param value the object in question
+   * @param propertyName the equivalent 'yaml' property name
+   * @throws NullPointerException if {@code value} is null
+   */
+  public static void checkNotNull(Object value, String propertyName) {
+    Preconditions.checkNotNull(value, "Property '" + propertyName + "' cannot be null");
```
chanseokoh, line 36: Should we annotate `@Nullable` on `value`?

BTW, Guava `checkNotNull` returns the input object like the following, although we don't need to follow the same pattern, of course.

```java
this.something = Preconditions.checkNotNull(something);
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Validator.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Checks if string is null, empty or only whitespace.
+   *
+   * @param value the string in question
+   * @param propertyName the equivalent 'yaml' property name
+   * @throws NullPointerException if {@code value} is null
+   * @throws IllegalArgumentException if {@code value} is empty or only whitespace
+   */
+  public static void checkNotEmpty(String value, String propertyName) {
```
chanseokoh, line 47: `@Nullable` arg?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpecTest.java
```diff
@@ -46,26 +46,78 @@ public void testArchiveLayerSpec_nameRequired() {
...
   }
 
+  @Test
+  public void testArchiveLayerSpec_nameNonNull() {
+    String data = "name: null\n" + "archive: out/archive";
+
+    try {
+      mapper.readValue(data, ArchiveLayerSpec.class);
+      Assert.fail();
+    } catch (JsonProcessingException jpe) {
```
chanseokoh, line 64: Interesting. So the mapper catches runtime exceptions and wraps it with `JsonProcessingException` after attaching the underlying runtime exception message?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Validator.java
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Checks if a collection is null or empty.
+   *
+   * @param value the string in question
+   * @param propertyName the equivalent 'yaml' property name
+   * @throws NullPointerException if {@code value} is null
+   * @throws IllegalArgumentException if {@code value} is empty
+   */
+  public static void checkNotEmpty(Collection<?> value, String propertyName) {
```
chanseokoh, line 61: `@Nullable` arg?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Validator.java
```diff
@@ -0,0 +1,82 @@
...
+   * @param value the string in question
+   * @param propertyName the equivalent 'yaml' property name
+   * @param expectedValue the value we expect {@code value} to be
+   * @throws NullPointerException if {@code value} is null
+   * @throws IllegalArgumentException if {@code value} is not equal to {@code expectedValue}
+   */
+  public static void checkEquals(String value, String propertyName, String expectedValue) {
+    checkNotNull(value, propertyName);
+    Preconditions.checkArgument(
+        expectedValue.equals(value),
```
chanseokoh, line 79: `value.equals(expectedValue)` for extra safety since we can't prevent the caller from passing null to `expectedValue`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Validator.java
loosebazooka: Yeah I thought about that, it doesn't seem necessary here.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/ArchiveLayerSpecTest.java
loosebazooka: Yes, any exception is always wrapped in JsonProcessingException.

PR #2609, opened 2020-07-23
Convert most collections to empty instead of optional

loosebazooka:
- entrypoint and cmd are still optional as we have to
  differentiate between the user trying to erase a
  base image value vs not setting anything.

Doing some more thinking and reading about this (https://github.com/GoogleContainerTools/jib/pull/2598#discussion_r456577550). It seems like using optional for collections is probably not great, except in the case of entrypoint and cmd.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
@@ -23,6 +23,8 @@ import com.google.cloud.tools.jib.api.buildplan.AbsoluteUnixPath;
 import com.google.cloud.tools.jib.api.buildplan.ImageFormat;
 import com.google.cloud.tools.jib.api.buildplan.Port;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
@@ -66,12 +68,14 @@ public class BuildFileSpec {
   @Nullable private BaseImageSpec from;
   @Nullable private Instant creationTime;
   @Nullable private ImageFormat format;
-  @Nullable private Map<String, String> environment;
-  @Nullable private Map<String, String> labels;
-  @Nullable private Set<AbsoluteUnixPath> volumes;
-  @Nullable private Set<Port> exposedPorts;
+  private Map<String, String> environment;
+  private Map<String, String> labels;
+  private Set<AbsoluteUnixPath> volumes;
+  private Set<Port> exposedPorts;
   @Nullable private String user;
   @Nullable private AbsoluteUnixPath workingDirectory;
+  // these two specific collection can be null as they can be set to empty to override
+  // the base image
   @Nullable private List<String> entrypoint;
   @Nullable private List<String> cmd;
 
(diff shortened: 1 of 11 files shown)


chanseokoh:
> * entrypoint and cmd are still optional as we have to
>   differentiate between the user trying to erase a
>   base image value vs not setting anything.

`Optional` makes sense for them. Just in case, I believe jib-core (through jib-build-plan) doesn't allow "erasing" entrypoint or cmd (although I don't remember if they can be set to an empty list). Rather, non-existence means inheriting them from a base image.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
```diff
@@ -66,12 +68,14 @@
...
-  @Nullable private Set<AbsoluteUnixPath> volumes;
-  @Nullable private Set<Port> exposedPorts;
+  private Map<String, String> environment;
+  private Map<String, String> labels;
+  private Set<AbsoluteUnixPath> volumes;
+  private Set<Port> exposedPorts;
   @Nullable private String user;
   @Nullable private AbsoluteUnixPath workingDirectory;
+  // these two specific collection can be null as they can be set to empty to override
+  // the base image
```
chanseokoh, line 78: This confuses me. It says "can be null" and then "set to empty". Does "empty" mean an empty list? And what exactly "override" means?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpec.java
```diff
@@ -124,14 +128,13 @@ public BuildFileSpec(
...
     }
-    this.environment = environment;
-    this.labels = labels;
-    if (volumes != null) {
-      this.volumes = volumes.stream().map(AbsoluteUnixPath::get).collect(Collectors.toSet());
-    }
-    if (exposedPorts != null) {
-      this.exposedPorts = Ports.parse(exposedPorts);
-    }
+    this.environment = (environment == null) ? ImmutableMap.of() : environment;
```
chanseokoh, line 131: Use of `ImmutableMap` makes me wonder if it is possible to make all the fields of the class `final`? (Although using `final` on all the fields doesn't make this class an immutable one, but there is no reason not to do so if possible.) Or, is it not feasible due to JSON parsing?

loosebazooka: No final should be on it. I guess I just never got around to doing it.

loosebazooka:
> Optional makes sense for them. Just in case, I believe jib-core (through jib-build-plan) doesn't allow "erasing" entrypoint or cmd (although I don't remember if they can be set to an empty list). Rather, non-existence means inheriting them from a base image.

So I just reread this comment. If erasure is not allowed at all, then these don't need to be optional. I need to see if they can be set to empty though.

chanseokoh:
> So I just reread this comment. If erasure is not allowed at all, then these don't need to be optional. I need to see if they can be set to empty though.

I should say I don't remember this correctly. And I just meant by "erasing" the technicality of setting the `Entrypoint` JSON property of a container configuration to  the `null` value (which technically is different from an empty list). It may be that setting it to an empty list is supported, while setting it to `null` isn't. But maybe a container runtime may not distinguish them.

loosebazooka:
There is an old discussion about this at rules_docker: https://github.com/bazelbuild/rules_docker/issues/379

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
```diff
@@ -85,15 +86,15 @@ public PlatformSpec(
     return Optional.ofNullable(osVersion);
   }
 
-  public Optional<List<String>> getOsFeatures() {
-    return Optional.ofNullable(osFeatures);
+  public List<String> getOsFeatures() {
```
mpeddada1, line 89: Sorry, can we return ImmutableList<String> here instead of List? Might be more secure to pass around immutable collections.

loosebazooka: That's a good question. Usually we use the base type interface on return values. That keeps implementation details from leaking out. However having said that, this is not an api that is exposed externally, so it could be okay. The other issue is that we don't know what kind of list the yaml parser returns, so we would have to wrap the list generated in the constructor  (https://github.com/GoogleContainerTools/jib/pull/2609/files#diff-784a41ae78bc030a7b61cd50c6a28e42R72) with a ImmutableList type.

So there's a lot going on here, and there's no obvious correct answer. It could go either way. In general though, we use the base type interface, and I just happened to continue to use that pattern.

mpeddada1: Thanks for the explanation! That makes sense:)

PR #2615, opened 2020-07-23
Adding support for PullBaseImageStep to search a manifest list using a user specified os/arch pair

louismurerwa:
This PR adds support for a user to specify the platform he/she wants to use when building an image

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.jib.builder.steps;
 
-import com.google.api.client.util.Preconditions;
 import com.google.cloud.tools.jib.api.Credential;
 import com.google.cloud.tools.jib.api.DescriptorDigest;
 import com.google.cloud.tools.jib.api.ImageReference;
@@ -270,9 +269,6 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
       throws IOException, RegistryException {
 
     Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
-    Preconditions.checkArgument(
-        platforms.size() == 1,
-        "multiple base image platforms is not yet supported; please specify only one platform");
     Platform platform = platforms.iterator().next();
     String architecture = platform.getArchitecture();
     String os = platform.getOs();
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.jib.builder.steps;
 
+import com.google.api.client.util.Preconditions;
 import com.google.cloud.tools.jib.api.Credential;
 import com.google.cloud.tools.jib.api.DescriptorDigest;
 import com.google.cloud.tools.jib.api.ImageReference;
@@ -269,6 +270,9 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
       throws IOException, RegistryException {
 
     Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
+    Preconditions.checkArgument(
+        platforms.size() == 1,
+        "multiple base image platforms is not yet supported; please specify only one platform");
     Platform platform = platforms.iterator().next();
     String architecture = platform.getArchitecture();
     String os = platform.getOs();
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -294,7 +294,7 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
               + os
               + ". If your intention was to specify a platform for your image,"
               + " see https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image"
-              + " to learn more about specifyng a platform";
+              + " to learn more about specifying a platform";
 
       eventHandlers.dispatch(LogEvent.error(errorMessage));
       throw new RegistryException(errorMessage);
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -22,6 +22,7 @@ import com.google.cloud.tools.jib.api.ImageReference;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.api.RegistryException;
 import com.google.cloud.tools.jib.api.RegistryUnauthorizedException;
+import com.google.cloud.tools.jib.api.buildplan.Platform;
 import com.google.cloud.tools.jib.blob.Blobs;
 import com.google.cloud.tools.jib.builder.ProgressEventDispatcher;
 import com.google.cloud.tools.jib.builder.TimerEventDispatcher;
@@ -46,9 +47,11 @@ import com.google.cloud.tools.jib.json.JsonTemplateMapper;
 import com.google.cloud.tools.jib.registry.ManifestAndDigest;
 import com.google.cloud.tools.jib.registry.RegistryClient;
 import com.google.cloud.tools.jib.registry.credentials.CredentialRetrievalException;
+import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import javax.annotation.Nullable;
 
@@ -71,6 +74,8 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
 
   private final BuildContext buildContext;
   private final ProgressEventDispatcher.Factory progressEventDispatcherFactory;
+  private String architecture = "amd64";
+  private String os = "linux";
 
   PullBaseImageStep(
       BuildContext buildContext, ProgressEventDispatcher.Factory progressEventDispatcherFactory) {
@@ -260,14 +265,22 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
   }
 
   /**
-   * Looks through a manifest list for any amd64/linux manifest and downloads and returns the first
-   * manifest it finds.
+   * Looks through a manifest list for the user specified arch/os manifest and downloads and returns
+   * the first manifest it finds.
    */
-  private ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
+  @VisibleForTesting
+  ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
       RegistryClient registryClient, V22ManifestListTemplate manifestListTemplate)
       throws RegistryException, IOException {
 
-    List<String> digests = manifestListTemplate.getDigestsForPlatform("amd64", "linux");
+    if (buildContext.getContainerConfiguration() != null) {
+      Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
+      Platform platform = platforms.stream().findFirst().get();
+      architecture = platform.getArchitecture();
+      os = platform.getOs();
+    }
+
+    List<String> digests = manifestListTemplate.getDigestsForPlatform(architecture, os);
     if (digests.size() == 0) {
       String errorMessage =
           buildContext.getBaseImageConfiguration().getImage()
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -271,32 +271,27 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
       RegistryClient registryClient, V22ManifestListTemplate manifestListTemplate)
       throws RegistryException, IOException {
 
-    String architecture = "amd64";
-    String os = "linux";
-
-    if (buildContext.getContainerConfiguration() != null) {
-      Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
-      Platform platform = platforms.iterator().next();
-      architecture = platform.getArchitecture();
-      os = platform.getOs();
-    }
+    Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
+    Platform platform = platforms.iterator().next();
+    String architecture = platform.getArchitecture();
+    String os = platform.getOs();
 
     buildContext
         .getEventHandlers()
         .dispatch(
-            LogEvent.error(
+            LogEvent.lifecycle(
                 "Searching the manifest list for the platfrom, architecture =  "
                     + architecture
                     + ", os = "
                     + os));
 
-    List<String> digests = manifestListTemplate.getDigestsForPlatform(os, architecture);
+    List<String> digests = manifestListTemplate.getDigestsForPlatform(architecture, os);
     if (digests.size() == 0) {
       String errorMessage =
           buildContext.getBaseImageConfiguration().getImage()
-              + " is a manifest list, but the list does not contain an image manifest for "
+              + " is a manifest list, but the list does not contain an image manifest for the platform arch = "
               + architecture
-              + "/"
+              + ", os = "
               + os
               + ". If your intention was to specify a platform for your image,"
               + " see https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image"
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -204,7 +204,7 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
     if (manifestTemplate instanceof V22ManifestListTemplate) {
       eventHandlers.dispatch(
           LogEvent.lifecycle(
-              "The base image reference is manifest list, searching for linux/amd64"));
+              "The base image reference is manifest list, searching for the user specified platform"));
       manifestAndDigest =
           obtainPlatformSpecificImageManifest(
               registryClient, (V22ManifestListTemplate) manifestTemplate);
@@ -273,12 +273,22 @@ class PullBaseImageStep implements Callable<ImageAndRegistryClient> {
       RegistryClient registryClient, V22ManifestListTemplate manifestListTemplate)
       throws RegistryException, IOException {
 
+    LogEvent.lifecycle("Found Manifest List ");
     if (buildContext.getContainerConfiguration() != null) {
       Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
       Platform platform = platforms.stream().findFirst().get();
       architecture = platform.getArchitecture();
       os = platform.getOs();
     }
+
+    buildContext
+        .getEventHandlers()
+        .dispatch(
+            LogEvent.lifecycle(
+                "Searching the manifest list for the platfrom , architecture =  "
+                    + architecture
+                    + " , os = "
+                    + os));
 
     List<String> digests = manifestListTemplate.getDigestsForPlatform(architecture, os);
     if (digests.size() == 0) {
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,14 +265,22 @@ private Image pullBaseImage(
...
    */
-  private ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
+  @VisibleForTesting
+  ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
       RegistryClient registryClient, V22ManifestListTemplate manifestListTemplate)
       throws RegistryException, IOException {
 
-    List<String> digests = manifestListTemplate.getDigestsForPlatform("amd64", "linux");
+    if (buildContext.getContainerConfiguration() != null) {
+      Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
```
louismurerwa, line 277: I was thinking of checking if the platforms set is empty here but i realized that our front end `ContainerBuildPlan ` does the job for us.Should I add a check or should I trust our api?

chanseokoh: Good question. Because `ContainerBuildPlan` never returns an empty list and we have added a test to ensure it's non-empty, in this case, I think I'm fine with assuming it's non-empty _given that `getContainerConfiguration()` is not null_.

We can certainly add more code to guard against an empty set and raise an error which may also be a good practice in general (depends on situations). But the integration between `ContainerBuildPlan` and the jib-core is pretty tight and and internal, I think assuming non-empty is OK _given that `getContainerConfiguration()` is not null_.`

That said, I noticed `getContainerConfiguration()` is `@Nullable`. I think we should figure out when this can be null and why.

Lastly, using streams works, but for just getting the first element, I'd go with
```java
    platforms.iterator().next();
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,14 +265,22 @@ private Image pullBaseImage(
...
+      Platform platform = platforms.stream().findFirst().get();
+      architecture = platform.getArchitecture();
+      os = platform.getOs();
+    }
```
chanseokoh, line 281: I find no reason to declare these as method fields.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: > That said, I noticed `getContainerConfiguration()` is `@Nullable`. I think we should figure out when this can be null and why.

See #2618 and #2619.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
louismurerwa: Thats the only way we can access the variable inside the platform object.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
louismurerwa: Once #2619 is merged ,then I will need to adjust my code a little bit

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,14 +265,22 @@ private Image pullBaseImage(
...
+
+    List<String> digests = manifestListTemplate.getDigestsForPlatform(architecture, os);
```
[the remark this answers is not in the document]
louismurerwa: Sounds Good I just added it

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: What I meant is basically this:
```java
String architecture = "amd64"; // no need to have them as method fields, just local variables
String os = "linux";
if (...) {
  ...
  architecture = platform.getArchitecture();
  os = platforms.getOs();
}
List<String> digests = manifestListTemplate.getDigestsForPlatform(architecture, os);
```

Even with this, it's still not ideal because this `PullBaseImageStep` is yet another entity having an opinion of what a default should be. But this issue will be gone once #2619 is merged.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,22 +263,40 @@ private Image pullBaseImage(
...
 
-    List<String> digests = manifestListTemplate.getDigestsForPlatform("amd64", "linux");
+    Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
+    Platform platform = platforms.iterator().next();
+    String architecture = platform.getArchitecture();
+    String os = platform.getOs();
+
+    buildContext
+        .getEventHandlers()
+        .dispatch(
```
chanseokoh, line 281:
```suggestion
    EventHandlers eventHandlers = buildContext.getEventHandlers();
    eventHandlers.dispatch(
```

Then on line 300, you can reuse `eventHandlers`.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -127,4 +138,80 @@ public void testCall_offlineMode_cached()
 
     Mockito.verify(buildContext, Mockito.never()).newBaseImageRegistryClientFactory();
   }
+
```
louismurerwa, line 141: i'm skeptical about my implementation for this testing method .

Basically its 

1.  Pass in a mock manifest list to the `testObtainPlatformSpecificImageManifest` method.
2.  Method searches for the digest matching the input platform
3.  Method calls registry client with with the obtained digest to get the mock manifest

Looking for feedback

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,22 +263,40 @@ private Image pullBaseImage(
...
-              + " If your intention was to use a non-amd64/linux base image,"
+              + " is a manifest list, but the list does not contain an image manifest for the platform arch = "
+              + architecture
+              + ", os = "
+              + os
+              + ". If your intention was to specify a platform for your image,"
               + " see https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image"
-              + " to learn how to specify a manifest instead of a manifest list, until Jib fixes"
-              + " https://github.com/GoogleContainerTools/jib/issues/1567 to allow specifying architecture and OS.";
+              + " to learn more about specifyng a platform using the platform tag";
```
chanseokoh, line 298: "tag" is a confusing term, so let's just skip it.
```suggestion
              + " to learn more about specifying a platform";
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,22 +263,40 @@ private Image pullBaseImage(
...
+
+    List<String> digests = manifestListTemplate.getDigestsForPlatform(architecture, os);
     if (digests.size() == 0) {
       String errorMessage =
           buildContext.getBaseImageConfiguration().getImage()
-              + " is a manifest list, but the list does not contain an image manifest for amd64/linux."
-              + " If your intention was to use a non-amd64/linux base image,"
+              + " is a manifest list, but the list does not contain an image manifest for the platform arch = "
+              + architecture
+              + ", os = "
```
chanseokoh, line 294:
```suggestion
              + " is a manifest list, but the list does not contain an image manifest for the platform architecture="
              + architecture
              + ", os="
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,22 +263,40 @@ private Image pullBaseImage(
...
+    String architecture = platform.getArchitecture();
+    String os = platform.getOs();
+
+    buildContext
+        .getEventHandlers()
+        .dispatch(
+            LogEvent.lifecycle(
+                "Searching the manifest list for the platfrom, architecture =  "
+                    + architecture
+                    + ", os = "
```
chanseokoh, line 285:
```suggestion
            LogEvent.lifecycle(
                "Searching the manifest list for the platform, architecture="
                    + architecture
                    + ", os="
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,22 +263,40 @@ private Image pullBaseImage(
...
+   * the first manifest it finds.
    */
-  private ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
+  @VisibleForTesting
+  ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
       RegistryClient registryClient, V22ManifestListTemplate manifestListTemplate)
       throws RegistryException, IOException {
 
-    List<String> digests = manifestListTemplate.getDigestsForPlatform("amd64", "linux");
+    Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
```
chanseokoh, line 274: I think we need to throw an error if there is more than one platform. Something along the line of

```
Preconditions.checkArgument(platforms.size() == 1, "multiple base image platforms is not yet support; please specify only one platform");
```

You can do this in a separate PR if you want. Regardless, you need to add a test.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -127,4 +138,80 @@ public void testCall_offlineMode_cached()
...
+            + "    }\n"
+            + "}";
+    Mockito.when(buildContext.getContainerConfiguration()).thenReturn(containerConfiguration);
+    Mockito.when(containerConfiguration.getPlatforms())
+        .thenReturn(ImmutableSet.of(new Platform("testArchitecture", "testOS")));
+    V22ManifestListTemplate testManifestList =
+        JsonTemplateMapper.readJson(testManifestListTemplate, V22ManifestListTemplate.class);
+
+    V22ManifestTemplate testManifest =
+        new ObjectMapper().readValue(testManifestTemplate, V22ManifestTemplate.class);
```
chanseokoh, line 189: Be consistent and use either `JsonTemplateMapper` or `new ObjectMapper()` across the file.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -127,4 +138,80 @@ public void testCall_offlineMode_cached()
 
     Mockito.verify(buildContext, Mockito.never()).newBaseImageRegistryClientFactory();
   }
+
+  @Test
+  public void testObtainPlatformSpecificImageManifest()
+      throws IOException, InterruptedException, GeneralSecurityException, URISyntaxException,
+          RegistryException {
```
chanseokoh, line 145:
```suggestion
  public void testObtainPlatformSpecificImageManifest() throws IOException, RegistryException {
```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -127,4 +138,80 @@ public void testCall_offlineMode_cached()
...
+            + "            \"os\": \"testOS\"\n"
+            + "         }\n"
+            + "      },\n"
+            + "      {\n"
+            + "         \"mediaType\": \"application/vnd.docker.distribution.manifest.v2+json\",\n"
+            + "         \"size\": 425,\n"
+            + "         \"digest\": \"sha256:5bb8e50aa2edd408bdf3ddf61efb7338ff34a07b762992c9432f1c02fc0e5e62\",\n"
+            + "         \"platform\": {\n"
+            + "            \"architecture\": \"Architecture\",\n"
+            + "            \"os\": \"OS\"\n"
```
chanseokoh, line 166:
```suggestion
            + "            \"architecture\": \"targetArchitecture\",\n"
            + "            \"os\": \"targetOS\"\n"
```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -127,4 +138,80 @@ public void testCall_offlineMode_cached()
...
+            + "    \"mediaType\": \"application/vnd.docker.distribution.manifest.v2+json\",\n"
+            + "    \"config\": {\n"
+            + "        \"mediaType\": \"application/vnd.docker.container.image.v1+json\",\n"
+            + "        \"size\": 424,\n"
+            + "        \"digest\": \"sha256:5bb8e50aa2edd408bdf3ddf61efb7338ff34a07b762992c9432f1c02fc0e5e62\"\n"
+            + "    }\n"
+            + "}";
+    Mockito.when(buildContext.getContainerConfiguration()).thenReturn(containerConfiguration);
+    Mockito.when(containerConfiguration.getPlatforms())
+        .thenReturn(ImmutableSet.of(new Platform("testArchitecture", "testOS")));
```
chanseokoh, line 184: I think it's more meaningful to make the code return the second manifest from the list.
```suggestion
        .thenReturn(ImmutableSet.of(new Platform("targetArchitecture", "targetOS")));
```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -127,4 +138,80 @@ public void testCall_offlineMode_cached()
...
+    V22ManifestListTemplate testManifestList =
+        JsonTemplateMapper.readJson(testManifestListTemplate, V22ManifestListTemplate.class);
+
+    V22ManifestTemplate testManifest =
+        new ObjectMapper().readValue(testManifestTemplate, V22ManifestTemplate.class);
+
+    ManifestAndDigest<?> manifestAndDigest =
+        new ManifestAndDigest(testManifest, Digests.computeJsonDigest(testManifest));
```
chanseokoh, line 192:
```suggestion
        new ManifestAndDigest<>(testManifest, Digests.computeJsonDigest(testManifest));
```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
chanseokoh: Yeah, think about what you are testing. Most of the assertions that verifies the manifest content is unnecessary, because what you are testing is whether the method can find the digest to pass to `registryClient.pullManifest()` and whether it returns the same manifest returned by `registryClient`. `registryClient.pullManifest` doesn't even need to return an actual manifest; a mock will work. I can talk about this later.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,23 +263,40 @@ private Image pullBaseImage(
...
+    Platform platform = platforms.iterator().next();
+    String architecture = platform.getArchitecture();
+    String os = platform.getOs();
+
+    EventHandlers eventHandlers = buildContext.getEventHandlers();
+    eventHandlers.dispatch(
+        LogEvent.info(
+            "Searching the manifest list for the platform, architecture="
```
chanseokoh, line 282: So, perhaps

LogEvent.lifecycle("The base image reference is manifest list, searching for architecture=...")

(Probably "for the platform" is redundant too. Generally speaking, program/tool messages should be concise.)

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -127,4 +132,49 @@ public void testCall_offlineMode_cached()
...
+    ManifestAndDigest<?> manifest = Mockito.mock(ManifestAndDigest.class);
+    Mockito.<ManifestAndDigest<?>>when(
+            registryClient.pullManifest(
+                "sha256:222222222222222222222222222222222222222222222222222222222222222222"))
+        .thenReturn(manifest);
+
+    ManifestAndDigest<?> returnManifest =
+        pullBaseImageStep.obtainPlatformSpecificImageManifest(registryClient, manifestList);
+
+    Assert.assertEquals(manifest, returnManifest);
```
chanseokoh, line 178: nit: `assertEquals` may work, but we can use `assertSame` to check if two object references are same (i.e., if `manifest == returnManifest` evaluates to true, meaning the references are pointing to the same single instance in the memory). OTOH, `assertEquals` may return when two (possible separate) object instances are deemed "equal" (whose definition is given by the class implementation of `.equals()`) even if `manifest != returnManifest`.

louismurerwa: Good point !! Object comparison requires ` .equals() `

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
louismurerwa: @chanseokoh I am going to add this ` check `  in this PR

chanseokoh: Let's do it in a separate PR. We may choose a different place for the check.

chanseokoh: Sorry, let's not* do it.

chanseokoh: Oh, wait. What I said was right. Let's not do it here but in a separate PR.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -260,23 +261,43 @@ private Image pullBaseImage(
...
-  private ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
+  @VisibleForTesting
+  ManifestAndDigest<?> obtainPlatformSpecificImageManifest(
       RegistryClient registryClient, V22ManifestListTemplate manifestListTemplate)
-      throws RegistryException, IOException {
+      throws IOException, RegistryException {
 
-    List<String> digests = manifestListTemplate.getDigestsForPlatform("amd64", "linux");
+    Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
+    Preconditions.checkArgument(
```
louismurerwa, line 273: I added a new `check` here after you approved the pr

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
louismurerwa: Ok let me revert my commit

chanseokoh: Sounds good. Avoid significantly changing the original intention of a PR (e.g., adding a new feature that can be done in a separate PR) (unless necessary) when people started reviewing.

For the check, I'm thinking maybe we can check and raise an error at an earlier point. We should also add a test too anyway.

louismurerwa: Yes the feature requires a test or two to confirm its functionality .At which point ,do you think we could do the check?

chanseokoh: `JibContainerBuilder.containerize()` came to my mind, but I haven't given much thought actually. I'll need to look into the code, but in the meantime, you can continue work on other stuff.

chanseokoh: I think we can validate the input in `JibContainerBuilder.containerize()` and that `UnsupportedOperationException` could work. But if you think there's a better place, we can discuss.

```java
    try (BuildContext buildContext = toBuildContext(containerizer);
            new TimerEventDispatcher(
                buildContext.getEventHandlers(), containerizer.getDescription())) {
      if (...) {
        throw UnsupportedOperationException("... not yet supported ...");
      }
      logSources(buildContext.getEventHandlers());
```

louismurerwa: Single Platform Specification Support #2641

PR #2619, opened 2020-07-23
Make ContainerConfiguration required when building BuildContext

### LINKED ISSUE #2618: Make ContainerConfiguration required when instantiatating BuildContext
Currently `BuildContext.getContainerConfiguration()` is `@Nullable`. However, I realized that we [always provide a container configuration instance](https://github.com/GoogleContainerTools/jib/blob/6e54252d348eaa3c56dc3aa193db2c7a64770dbf/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java#L717) when building a `BuildContext`.

I think we can make `ContainerConfiguration` a required field.

chanseokoh:
Fixes #2618.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
@@ -39,9 +39,6 @@ import javax.annotation.Nullable;
 /** Immutable configuration options for the container. */
 public class ContainerConfiguration {
 
-  /** The default creation time of the container (constant to ensure reproducibility by default). */
-  public static final Instant DEFAULT_CREATION_TIME = Instant.EPOCH;
-
   /** Builder for instantiating a {@link ContainerConfiguration}. */
   public static class Builder {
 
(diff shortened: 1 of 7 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/configuration/ContainerConfiguration.java
```diff
@@ -39,17 +39,14 @@
 /** Immutable configuration options for the container. */
 public class ContainerConfiguration {
 
-  /** The default creation time of the container (constant to ensure reproducibility by default). */
-  public static final Instant DEFAULT_CREATION_TIME = Instant.EPOCH;
```
loosebazooka, line 43: I wonder if we should keep this?

chanseokoh: Okay. The class is not a public API, so I moved it to the builder and made it `private`.

FYI, eventually we should remove the entire class and replace it with `ContainerBuildPlan`. They are basically identical.

Issue #2627, opened 2020-07-25
AWS ECR with jib.to.auth.username does not work


jomach:
**Environment**:

- *Jib version: 2.4.0* 
- *Build tool: maven 3.6.3*
- *OS: MacOs* 

**Description of the issue**:

Running the command: 
`mvn compile com.google.cloud.tools:jib-maven-plugin:2.4.0:build Djib.to.auth.username=${AWS_ECR_KEY_USER} -Djib.to.auth.password=${AWS_ECR_SECRET} -Dimage=<MY IMAGE>` does not work and I get a 401 from the ECR registry 

**Expected behavior**: Should build

**Steps to reproduce**:
Create a mvn project with : 

```
<build>
<plugins>
    <plugin>
        <groupId>com.google.cloud.tools</groupId>
        <artifactId>jib-maven-plugin</artifactId>
        <version>${jib.version}</version>
        <executions>
            <execution>
                <phase>package</phase>
                <goals>
                    <goal>build</goal>
                </goals>
            </execution>
        </executions>
        <configuration>
            <from>
                <image>gcr.io/distroless/java:8</image>
            </from>
            <to>
                <image>
                    imageURI:/${project.artifactId}:${project.version}
                </image>
                <tags>latest</tags>
            </to>
        </configuration>
    </plugin>
</plugins>
</build>
```

**Log output**: are set up correctly. See https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#what-should-i-do-when-the-registry-responds-with-unauthorized for help: Unauthorized forimageURI: 401 Unauthorized -> [Help 1]

AWS_ECR_KEY_USER and AWS_ECR_SECRET are my aws credentials this seems not to work. If I don't pass the -Djib.to.auth.username it will grab it from amazon-ecr-credential-helper which in my case works and the credentials are the same. That's why I think this is a bug.

jomach:
I tracked it down to : 

```
  private T call(URL url) throws IOException, RegistryException {
    String serverUrl = registryEndpointRequestProperties.getServerUrl();
    String imageName = registryEndpointRequestProperties.getImageName();

    Request.Builder requestBuilder =
        Request.builder()
            .setUserAgent(userAgent)
            .setHttpTimeout(JibSystemProperties.getHttpTimeout())
            .setAccept(registryEndpointProvider.getAccept())
            .setBody(registryEndpointProvider.getContent())
            .setAuthorization(authorization);
```
[...]

PR #2628, opened 2020-07-25
Update README on authenticating using jib.to.auth.username|password with major cloud registries

jomach:
Update docs for hint about aws ecr 

Helps on #2627

### DIFF
--- a/jib-maven-plugin/README.md
+++ b/jib-maven-plugin/README.md
@@ -519,7 +519,7 @@ Property | Description
 
 e.g. `mvn compile jib:build -Djib.to.auth.username=user -Djib.to.auth.password=pass`
 
-**Note:** This method of authentication should be used only as a last resort, as it is insecure to make your password visible in plain text.
+**Note:** This method of authentication should be used only as a last resort, as it is insecure to make your password visible in plain text. Furthermore the username for the registry AWS ECR is always AWS and the password you need to get it using the command: `aws ecr get-login-password`. Passing aws keys and secret does not work.
 
 #### Using Maven Settings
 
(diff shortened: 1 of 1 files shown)


jib-maven-plugin/README.md
```diff
@@ -519,7 +519,7 @@ Property | Description
 
 e.g. `mvn compile jib:build -Djib.to.auth.username=user -Djib.to.auth.password=pass`
 
-**Note:** This method of authentication should be used only as a last resort, as it is insecure to make your password visible in plain text.
+**Note:** This method of authentication should be used only as a last resort, as it is insecure to make your password visible in plain text. Furthermore the username for the registry AWS ECR is always AWS and the password you need to get it using the command: `aws ecr get-login-password`. Passing aws keys and secret does not work.
```
chanseokoh, line 522:
```suggestion
**Note:** This method of authentication should be used only as a last resort, as it is insecure to make your password visible in plain text. Note that often cloud registries (such as Google GCR, Amazon ECR, Azure ACR) do not accept "user credentials" (such as Gmail account name and password) but require different forms of credentials. For example, you may use [`oauth2accesstoken` or `_json_key`](https://cloud.google.com/container-registry/docs/advanced-authentication) as the username for GCR depending on the authentication method, and [`AWS`](https://serverfault.com/questions/1004915/what-is-the-proper-way-to-log-in-to-ecr) for ECR. For ACR, you may use a [_service principle_](https://docs.microsoft.com/en-us/azure/container-registry/container-registry-auth-service-principal).
```

chanseokoh:
@jomach thanks for improving the doc!

PR #2634, opened 2020-07-28
Allowing PullBaseImageStep to pull multiple images corresponding to the platforms specified by the user

louismurerwa:
This is the first step torwards multi-platform image building.
This PR allows Jib to pull down multiple base images which correspond to user specified platforms

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -58,7 +58,7 @@ import java.util.Set;
 import java.util.concurrent.Callable;
 import javax.annotation.Nullable;
 
-/** Pulls the base image manifest. */
+/** Pulls the base image manifests for the specified platforms. */
 class PullBaseImageStep implements Callable<List<ImageAndRegistryClient>> {
 
   private static final String DESCRIPTION = "Pulling base image manifest";
@@ -174,12 +174,12 @@ class PullBaseImageStep implements Callable<List<ImageAndRegistryClient>> {
   }
 
   /**
-   * Pulls the base image.
+   * Pulls the base images specified in the platforms list.
    *
    * @param registryClient to communicate with remote registry
    * @param progressEventDispatcher the {@link ProgressEventDispatcher} for emitting {@link
    *     ProgressEvent}s
-   * @return the pulled image
+   * @return the list of pulled images
    * @throws IOException when an I/O exception occurs during the pulling
    * @throws RegistryException if communicating with the registry caused a known error
    * @throws LayerCountMismatchException if the manifest and configuration contain conflicting layer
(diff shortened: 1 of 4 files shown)


louismurerwa:
This PR is not yet ready for reviews

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -187,76 +188,94 @@ public ImageAndRegistryClient call()
...
-    // handling it normally
+    // special handling if we happen upon a manifest list, extract manifests handle them normally
     if (manifestTemplate instanceof V22ManifestListTemplate) {
-      manifestAndDigest =
-          obtainPlatformSpecificImageManifest(
-              registryClient, (V22ManifestListTemplate) manifestTemplate);
-      manifestTemplate = manifestAndDigest.getManifest();
+      Set<Platform> platforms = buildContext.getContainerConfiguration().getPlatforms();
+      Iterator<Platform> platformsIterator = platforms.iterator();
+      while (platformsIterator.hasNext()) {
```
chanseokoh, line 207: This can be
```java
for (Platform platform : platforms) {
   ...
}

File: jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java [hunk 2]
```diff
@@ -187,76 +188,94 @@ public ImageAndRegistryClient call()
...
-              buildableManifestTemplate, containerConfigurationTemplate);
-        }
-      default:
-        throw new IllegalStateException(
-            "Unknown manifest schema version: " + manifestTemplate.getSchemaVersion());
+                  buildContext.getBaseImageConfiguration().getImage(), v21ManifestTemplate);
+          imagesAndRegistryClient.add(
+              new ImageAndRegistryClient(
+                  JsonToImageTranslator.toImage(v21ManifestTemplate), registryClient));
+          // fall through
```
chanseokoh, line 230: I'll assume this PR is not ready and you'll fix this.

louismurerwa: Yes I put break statements over there

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -58,18 +58,18 @@
...
-  static class ImageAndRegistryClient {
+  static class BaseImagesAndRegistryClient {
 
-    final Image image;
+    final List<Image> baseImages;
     @Nullable final RegistryClient registryClient;
 
-    ImageAndRegistryClient(Image image, @Nullable RegistryClient registryClient) {
-      this.image = image;
+    BaseImagesAndRegistryClient(List<Image> baseImages, @Nullable RegistryClient registryClient) {
```
chanseokoh, line 71: Although the new name works, I think we can keep this general by `ImagesAndRegistryClient`. It's just a tuple of two things (an `Image` list and a `RegistryClient`) only required in Java to return multiple objects. If we give the name "Base", then other people who don't look into the implementation of this tuple will assume that there must be something specific in the class that has to be tied to base images. As a class, I think the name should just reflect what the class actually represents, not how it must be used with only a subset of semantics. For example, a (bad) analogy would be that I rename a general `Integer` class to `PositiveInteger` while not changing anything about the class implementation only because I will never use the class to store a non-positive integer throughout my application lifecycle. Rather, I should give the `positiveInteger` name to an instance of the `Integer` class (e.g., `Integer positiveInteger = 30;`). Does that make sense?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -187,76 +187,88 @@ public ImageAndRegistryClient call()
...
-                  registryClient.pullBlob(
-                      containerConfigurationDigest,
-                      progressEventDispatcherWrapper::setProgressTarget,
-                      progressEventDispatcherWrapper::dispatchProgress));
-
-          ContainerConfigurationTemplate containerConfigurationTemplate =
-              JsonTemplateMapper.readJson(
-                  containerConfigurationString, ContainerConfigurationTemplate.class);
+    for (ManifestAndDigest<?> manifestAndDigestInstance : manifestAndDigestList) {
+      manifestTemplate = manifestAndDigestInstance.getManifest();
```
chanseokoh, line 215: As we discussed yesterday, I think now it makes sense to factor out the code inside the loop as a separate method.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
louismurerwa: Yes I get what you mean.The images stored in these classes are no different form regular images so there is no need to rename the class.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
louismurerwa: Refactoring Done

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -63,7 +63,7 @@
           new IllegalStateException("invalid usage; required step not configured"));
     }
 
-    private Future<ImageAndRegistryClient> baseImageAndRegistryClient = failedFuture();
+    private Future<ImagesAndRegistryClient> imagesAndRegistryClient = failedFuture();
```
chanseokoh, line 66: So in this case, we should name it `baseImages...` as an instance that belongs to the `ImagesAndRegistryClient` "class". This instance will all contain base images.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -187,25 +187,97 @@ public ImageAndRegistryClient call()
    * @throws BadContainerConfigurationFormatException if the container configuration is in a bad
    *     format
    */
-  private Image pullBaseImage(
+  private ImagesAndRegistryClient pullBaseImages(
```
chanseokoh, line 190: I think it makes sense for this method to return `List<Image>`. Basically, not slipping the responsibilty of returning `ImagesAndRegistryClient` to this "helper" method, which can only have a local focus on downloading base images if viewed in isolation.

chanseokoh:
When you're ready, make the PR out of draft.

PR #2641, opened 2020-07-29
Single Platform Specification Support

louismurerwa:
This PR restricts Jib to support a single platform specification only .
Jib will throw an error if a user specifies more than one platform .

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/api/JibContainerBuilderTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/api/JibContainerBuilderTest.java
@@ -38,6 +38,7 @@ import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Consumer;
 import org.hamcrest.CoreMatchers;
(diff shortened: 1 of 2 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/api/JibContainerBuilderTest.java
```diff
@@ -362,4 +363,23 @@ public void setApplyContainerBuildPlan()
...
+          RegistryException, IOException, ExecutionException {
+    try {
+      ImageConfiguration imageConfiguration =
+          ImageConfiguration.builder(ImageReference.parse("base/image")).build();
+      new JibContainerBuilder(imageConfiguration, spyBuildContextBuilder)
+          .setPlatforms(
+              ImmutableSet.of(
+                  new Platform("testArchitecture", "testOS"),
+                  new Platform("testArchitecture1", "testOS2")))
+          .containerize(Containerizer.to(RegistryImage.named("target/image")));
```
chanseokoh, line 379: Generally you should limit the scope of `try { } catch` for this type of exception-checking tests, because theoretically other code (e.g., `ImageConfiguration.build()`) may also throw `UnsupportedOperationException`. We are checking the exception message which is very specific, but in the worst case, it's possible that a test becomes useless to catch a bug if the test doesn't check the message or the message is very generic.

```java
      ImageConfiguration imageConfiguration =
          ImageConfiguration.builder(ImageReference.parse("base/image")).build();
      JibContainerBuilder containerBuilder = new JibContainerBuilder(imageConfiguration, spyBuildContextBuilder)
          .setPlatforms(
              ImmutableSet.of(
                  new Platform("testArchitecture", "testOS"),
                  new Platform("testArchitecture1", "testOS2")))
    try {
          containerBuilder.containerize(Containerizer.to(RegistryImage.named("target/image")));
```

and I guess `new Platform("arch1", "os1"), new Platform("arch2", "os2")` could save a few lines (which isn't important at all but I think it's an easy way to save some of the expensive screen-estate).

chanseokoh: Updated the code.

louismurerwa: Good Catch

PR #2643, opened 2020-07-29
Use explicit interface instead of BiFunction  

### LINKED ISSUE #2638: Convert raw BiFunction into named interfaces
We have two transformation helpers:
1. `filePermissionsProvider: BiFunction<Path, AbsoluteUnixPath, FilePermissions>`
2. `modificationTimeProvider: BiFunction<Path, AbsoluteUnixPath, Instant>`

These work fine, but having named interfaces will allow us to document their use better on the interface rather than the calling function.

- Create two new named interfaces:
  1. `FilePermissionsProvider`
  2. `ModificationTimeProvider`
- Create new methods that use these named interfaces instead of BiFunction
- Deprecate (annotation) the old methods that use BiFunction directly

mpeddada1:
Fixes #2638

### DIFF
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
@@ -280,9 +280,8 @@ public class FileEntriesLayer implements LayerObject {
         OwnershipProvider ownershipProvider)
         throws IOException {
       FilePermissions permissions = filePermissionProvider.get(sourceFile, pathInContainer);
-      Instant modificationTime =
-          modificationTimeProvider.getFileModificationTime(sourceFile, pathInContainer);
-      String ownership = ownershipProvider.getFileOwnership(sourceFile, pathInContainer);
+      Instant modificationTime = modificationTimeProvider.get(sourceFile, pathInContainer);
+      String ownership = ownershipProvider.get(sourceFile, pathInContainer);
       addEntry(sourceFile, pathInContainer, permissions, modificationTime, ownership);
       if (!Files.isDirectory(sourceFile)) {
         return this;
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
@@ -100,31 +100,6 @@ public class FileEntriesLayer implements LayerObject {
     }
 
     /**
-     * Adds an entry to the layer. Only adds the single source file to the exact path in the
-     * container file system.
-     *
-     * <p>For example, {@code newAddEntry(Paths.get("myfile"),
-     * AbsoluteUnixPath.get("/path/in/container"))} adds a file {@code myfile} to the container file
-     * system at {@code /path/in/container}.
-     *
-     * <p>For example, {@code newAddEntry(Paths.get("mydirectory"),
-     * AbsoluteUnixPath.get("/path/in/container"))} adds a directory {@code mydirectory/} to the
-     * container file system at {@code /path/in/container/}. This does <b>not</b> add the contents
-     * of {@code mydirectory}.
-     *
-     * @param sourceFile the source file to add to the layer
-     * @param pathInContainer the path in the container file system corresponding to the {@code
-     *     sourceFile}
-     * @return this
-     */
-    public Builder newAddEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
-      return addEntry(
-          sourceFile,
-          pathInContainer,
-          FILE_PERMISSIONS_PROVIDER.retrieveFilePermission(sourceFile, pathInContainer));
-    }
-
-    /**
      * Adds an entry to the layer with the given permissions. Only adds the single source file to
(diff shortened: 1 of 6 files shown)


jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
```diff
@@ -89,7 +89,9 @@ public Builder addEntry(FileEntry entry) {
      * @param pathInContainer the path in the container file system corresponding to the {@code
      *     sourceFile}
      * @return this
+     * @deprecated Use {@link #newAddEntry(Path, AbsoluteUnixPath)}
      */
+    @Deprecated
```
loosebazooka, line 94: I'm not sure we have to deprecate everything in here. If the new methods do the same thing, there is no reason to make the new methods.

We might only want to really change the methods that expose the bifunction in the api?

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
```diff
@@ -300,6 +315,170 @@ public Builder addEntryRecursive(
...
+     * container file system at {@code /path/in/container/}. This does <b>not</b> add the contents
+     * of {@code mydirectory}.
+     *
+     * @param sourceFile the source file to add to the layer
+     * @param pathInContainer the path in the container file system corresponding to the {@code
+     *     sourceFile}
+     * @return this
+     */
+    // TODO: Rename this method once deprecated methods have been removed.
+    public Builder newAddEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
```
loosebazooka, line 337: In general we try not to do this: Create a new method with a new name and change the name when deprecating an old method. Someone trying to use this new api will inevitably have to change their code again when we rename.

chanseokoh:
Additionally, I think it's good to make deprecated methods to directly call the new method (i.e., deprecated methods just delegate to the new methods). It removes code duplicate, which is meaningful when we have to change implementations, and is a proof that it will continue to work in the same way when the user migrates their code.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
mpeddada1: Ah that is good to know. Agreed, this would be a hassle for the user of the method.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
```diff
@@ -133,7 +132,7 @@ public Builder addEntry(
       return addEntry(
           sourceFile,
```
mpeddada1, line 133: `addEntry(Path, AbsoluteUnixPath, Instant)` is currently not used. Should it still remain in the class?

chanseokoh: I'd say yes. `c.g.c.t.j.api.buildplan` is the public API of the "Container Build Plan" library. We may not use the method internally, but external users may be using it (and they always do).

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
```diff
@@ -94,7 +93,7 @@ public Builder addEntry(Path sourceFile, AbsoluteUnixPath pathInContainer) {
...
           pathInContainer,
-          DEFAULT_FILE_PERMISSIONS_PROVIDER.apply(sourceFile, pathInContainer));
+          DEFAULT_FILE_PERMISSIONS_PROVIDER.getFilePermissions(sourceFile, pathInContainer));
```
chanseokoh, line 96: Now I think the name `getFilePermissions()` is redundant and we can just go with `FilePermissionsProvider.get()`. I've seen this naming pattern in many places. @loosebazooka WDYT?

loosebazooka: Yeah `get` makes sense to me.

loosebazooka:
I realized there's some other BiFunction calls in the code base that require us to publish this updated BuildPlan before we change. So once this PR is in, we can do a publish and release and then update those parts.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
loosebazooka: Lets put that change in here and then merge the PR.

mpeddada1: That's a good idea.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FileEntriesLayer.java
mpeddada1: Oh I see! Thanks for the explanation:)

loosebazooka:
lets also make the other one's `get` as well. `OwnershipProvider`, `ModificationTimeProvider`

PR #2645, opened 2020-07-30
Add FilePropertiesStack helper class

loosebazooka:
This keeps track of the various cascading properties of file
layers at different levels. Allows pushing and popping, but its
true value is calculating and providing the various file permissions
and properties to whatever bit of code is populating a layer.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
```diff
@@ -0,0 +1,114 @@
...
+    filePermissions = null;
+    directoryPermissions = null;
+    modificationTime = null;
+    ownership = null;
+    user = null;
+    group = null;
+
+    // the item with the lowest index has the lowest priority
+    for (FilePropertiesSpec properties : stack) {
+      properties.getFilePermissions().ifPresent(permissions -> filePermissions = permissions);
```
chanseokoh, line 75: How about `permissions = properties.getFilePermissions.orElse(permissions)`? Or, how about just declaring the field as `Optional` and assign them directly? (The getter return types can remain `@Nullable` though.)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
```diff
@@ -0,0 +1,114 @@
...
+      properties
+          .getDirectoryPermissions()
+          .ifPresent(permissions -> directoryPermissions = permissions);
+      properties.getTimestamp().ifPresent(timestamp -> modificationTime = timestamp);
+      properties.getUser().ifPresent(user -> this.user = user);
+      properties.getGroup().ifPresent(group -> this.group = group);
```
chanseokoh, line 81: I usually pay special attention to `String` or collections when combined `Optional`, as they can be empty. Is it guaranteed that `user` is a non-empty string (if present)? Otherwise, it's possible that a lower stack has, say, `"1234"` and a higher `""` (empty string), resulting in clearing `"1234"`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+    // ownership calculations
+    if (group == null) {
```
chanseokoh, line 84: So if a lower stack had `user=1234` but a higher stack `group=5678`, it will lost `1234`. Probably this is what we intend, but I just wanted to confirm. Maybe good to have a test to formalize the intention.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
```diff
@@ -0,0 +1,114 @@
...
+      ownership = user;
+    } else {
+      if (user == null) {
+        ownership = ":" + group;
+      } else {
+        ownership = user + ":" + group;
+      }
+    }
```
chanseokoh, line 92: To me, the following is more readable in this case.
```suggestion
    } else if (user == null) {
      ownership = ":" + group;
    } else {
      ownership = user + ":" + group;
    }
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
```diff
@@ -0,0 +1,114 @@
...
+  private List<FilePropertiesSpec> stack = new ArrayList<>(3);
+
+  @Nullable private FilePermissions filePermissions;
+  @Nullable private FilePermissions directoryPermissions;
+  @Nullable private Instant modificationTime;
+  @Nullable private String ownership;
+
+  // intermediate values
+  @Nullable private String user;
+  @Nullable private String group;
```
chanseokoh, line 44: These don't seem necessary. In `updateProperties()`, you always start with `null`. These can be local to the method.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
loosebazooka: OrElse has a nice flow to it.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
loosebazooka: Right, so something I didn't include in this PR was enforcing empty in the json parser. I have that in a separate branch. But basically empty items will not be allowed.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
loosebazooka: I'm not sure I understand, the ownership calculation is outside the loop. So once `group` and `user` are finalized, then the values are converted into an ownership string.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
loosebazooka: Yeah, I can't remember why I did this, good catch.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
chanseokoh: Ah, I misunderstood. Yeah, the calculation is outside the loop.

Just so you know, jib-core will probably accept `:` (a single colon) as an ownership string, but I think it makes sense to return null as done here when both `user` and `group` are null.

loosebazooka:
@mpeddada1 @chanseokoh if someone could take a look and approve?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStackTest.java
```diff
@@ -0,0 +1,142 @@
...
+        FilePermissions.fromOctalString("111"), testStack.getDirectoryPermissions());
+    Assert.assertEquals("1:1", testStack.getOwnership());
+    Assert.assertEquals(Instant.ofEpochMilli(1), testStack.getModificationTime());
+  }
+
+  @Test
+  public void testPush_stacking() {
+    FilePropertiesStack testStack = new FilePropertiesStack();
+
+    testStack.push(new FilePropertiesSpec("111", "111", "1", "1", "1"));
```
mpeddada1, line 43: super nit: could be useful to add inline comments for these parameters? like /*filePermissions=*/ etc.

loosebazooka: In general we avoid these kind of comments. Usually that kind of information is provided by IDE, you can quickly determine what each parameter is.

PR #2646, opened 2020-07-30
obtainBaseImageLayers for multiple images

louismurerwa:
This is a draft PR for design review
Stylistics changes are still in progress
This PR adds the ability for Jib to pull and store baseImagelayers for multiple images

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -37,7 +37,9 @@ import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -64,7 +66,7 @@ public class StepsRunner {
     }
 
     private Future<ImagesAndRegistryClient> baseImageAndRegistryClient = failedFuture();
-    private Future<List<Future<PreparedLayer>>> baseImageLayers = failedFuture();
+    private Future<Map<Image, List<Future<PreparedLayer>>>> baseImageLayers = failedFuture();
     @Nullable private List<Future<PreparedLayer>> applicationLayers;
     private Future<Image> builtImage = failedFuture();
     private Future<RegistryClient> targetRegistryClient = failedFuture();
@@ -271,14 +273,28 @@ public class StepsRunner {
   }
 
   private void assignLocalImageResult(Future<LocalImage> localImage) {
-    results.baseImageLayers = executorService.submit(() -> localImage.get().layers);
+
+    Future<List<Future<PreparedLayer>>> localBaseImageLayers =
+        executorService.submit(() -> localImage.get().layers);
+
     results.baseImageAndRegistryClient =
         executorService.submit(
-            () ->
-                LocalBaseImageSteps.returnImageAndRegistryClientStep(
-                        realizeFutures(results.baseImageLayers.get()),
-                        localImage.get().configurationTemplate)
-                    .call());
+            () -> {
+              return LocalBaseImageSteps.returnImageAndRegistryClientStep(
+                      realizeFutures(localBaseImageLayers.get()),
+                      localImage.get().configurationTemplate)
+                  .call();
+            });
+
+    results.baseImageLayers =
+        executorService.submit(
+            () -> {
+              Map<Image, List<Future<PreparedLayer>>> baseImageLayers = new HashMap<>();
+              baseImageLayers.put(
+                  results.baseImageAndRegistryClient.get().images.get(0),
+                  localBaseImageLayers.get());
+              return baseImageLayers;
+            });
   }
 
   private void pullBaseImage() {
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -271,14 +273,28 @@ private void extractTar() {
...
+        executorService.submit(() -> localImage.get().layers);
+
     results.baseImageAndRegistryClient =
         executorService.submit(
-            () ->
-                LocalBaseImageSteps.returnImageAndRegistryClientStep(
-                        realizeFutures(results.baseImageLayers.get()),
-                        localImage.get().configurationTemplate)
-                    .call());
+            () -> {
```
chanseokoh, line 282: And probably you don't need `{ return ... }` but keep the original syntactic sugar `() -> expr`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -271,14 +273,28 @@ private void extractTar() {
...
+                  .call();
+            });
+
+    results.baseImageLayers =
+        executorService.submit(
+            () -> {
+              Map<Image, List<Future<PreparedLayer>>> baseImageLayers = new HashMap<>();
+              baseImageLayers.put(
+                  results.baseImageAndRegistryClient.get().images.get(0),
+                  localBaseImageLayers.get());
```
chanseokoh, line 295: `localImage.get().layers` should work. No need to defined `localBaseImageLayers` on line 277.

and this will probably work:
```java
.submit( () -> Collections.singletonMap(
            results.baseImageAndRegistryClient.get().images.get(0),
            localImage.get().layers) );
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -292,23 +308,31 @@ private void pullBaseImage() {
...
-                            results.baseImageAndRegistryClient.get().registryClient)
-                        : ObtainBaseImageLayerStep.makeListForSelectiveDownload(
-                            buildContext,
-                            childProgressDispatcherFactory,
-                            results.baseImageAndRegistryClient.get().images.get(0),
-                            results.baseImageAndRegistryClient.get().registryClient,
-                            results.targetRegistryClient.get())));
+            () -> {
+              Map<Image, List<Future<PreparedLayer>>> baseImageLayers = new HashMap<>();
+              for (Image image : results.baseImageAndRegistryClient.get().images) {
```
chanseokoh, line 315: I realized we could have increased concurrency further. `PullBaseImageStep` is now designed to return all manifests all at once, meaning that downloading base image layers can start only after _all_ manifests have been downloaded. Moreover, in `PullBaseImageStep`, we will probably end up having to download all manifests synchronously and sequentially. Theoretically and ideally we should be able to concurrently download manifests as well as start downloading base image layers while there are still pending manifest downloads. However, because this requires significant redesign of `PullBaseImageStep` (i.e., we should schedule multiple threads (multiple step instances) each downloading an assigned manifest, resulting in `List<Future<Image>>` instead of the current `List<Image>`) and because manifests are small and generally quick to download, I think the current implementation is reasonable. We just need to keep track of this issue so that we can eventually implement this.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa: Adding back the ` sugar ` so that our code can be ` sweet ` again hahah . Done.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa: This is a very good point .I hadn't realized that we are using a future<List<Image>> instead of List<Future<Image>> . As you said the manifest files are really small and thus there is a lower chance of them slowing down the build process I support the fact that we can refactor this later.I have put this point on the list for items to be improved once a working multi-image is completed.

chanseokoh: > because manifests are small and generally quick to download, I think the current implementation is reasonable.

For your insight, this is somewhat acceptable, but we've seen many, many Jib users working in a slow network where it takes, e.g., 4~5 seconds to download a manifest. So, if you are sequentially downloading 3 manifests, they will add up to ~15 seconds. 15 seconds would still be a small fraction considering that the entire build would take far more time in such a slow network, but it will be a significant bottleneck, e.g., when not pushing to a remote registry.

chanseokoh: Leaving this comment thread open, for easy reference in the future. You don't have to obsessed with collapsing every comment thread.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -64,7 +66,7 @@
     }
 
     private Future<ImagesAndRegistryClient> baseImageAndRegistryClient = failedFuture();
-    private Future<List<Future<PreparedLayer>>> baseImageLayers = failedFuture();
+    private Future<Map<Image, List<Future<PreparedLayer>>>> baseImageLayers = failedFuture();
```
louismurerwa, line 69: What is the best name for this map object that maps Image - baseImageLayers ? I was thinking of ` imageLayers` or ` baseImagesAndLayers `

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa: I always thought it was a requirement to close all of them hahaha I will leave this one open

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: `baseImagesAndLayers` sounds good to me.

chanseokoh: After addressing the comments, I think you can take this out of the draft status.

louismurerwa: Sounds Good

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -271,14 +273,22 @@ private void extractTar() {
   }
 
   private void assignLocalImageResult(Future<LocalImage> localImage) {
-    results.baseImageLayers = executorService.submit(() -> localImage.get().layers);
     results.baseImageAndRegistryClient =
```
chanseokoh, line 276: I didn't bother in the last PR as the overall implementation is in the transitioning state from a single image to multiple images, but given the current state, I think we should rename this to baseImage**s**AndRegistryClient to be consistent.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -271,14 +273,22 @@ private void extractTar() {
...
                         localImage.get().configurationTemplate)
                     .call());
+
+    results.baseImagesAndLayers =
+        executorService.submit(
+            () -> {
+              Map<Image, List<Future<PreparedLayer>>> baseImageAndLayers = new HashMap<>();
+              baseImageAndLayers.put(
+                  results.baseImageAndRegistryClient.get().images.get(0), localImage.get().layers);
+              return baseImageAndLayers;
```
chanseokoh, line 290: `Collections.signletonMap` (https://github.com/GoogleContainerTools/jib/pull/2646#discussion_r463274516)?

louismurerwa: I had missed this review

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -271,44 +274,56 @@ private void extractTar() {
...
-                            results.baseImageAndRegistryClient.get().images.get(0),
-                            results.baseImageAndRegistryClient.get().registryClient)
-                        : ObtainBaseImageLayerStep.makeListForSelectiveDownload(
-                            buildContext,
-                            childProgressDispatcherFactory,
-                            results.baseImageAndRegistryClient.get().images.get(0),
-                            results.baseImageAndRegistryClient.get().registryClient,
-                            results.targetRegistryClient.get())));
+            () -> {
+              Map<Image, List<Future<PreparedLayer>>> baseImageAndLayers = new HashMap<>();
```
chanseokoh, line 307: baseImage*s*AndLayers

PR #2647, opened 2020-07-30
Creating interfaces that will be used to replace BiFunction

mpeddada1:
This pr serves as part 1 of the process to replace BiFunction usage with explicit interfaces.
For #2638

### DIFF
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FilePermissionsProvider.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FilePermissionsProvider.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.jib.api.buildplan;
 import java.nio.file.Path;
 
 /** Serves as a base class for providing file permissions on a container. */
+@FunctionalInterface
 public interface FilePermissionsProvider {
 
   /**
@@ -30,5 +31,5 @@ public interface FilePermissionsProvider {
    *     sourcePath.
    * @return the permissions to be set for the file.
    */
-  public FilePermissions retrieveFilePermissions(Path sourcePath, AbsoluteUnixPath destinationPath);
+  public FilePermissions getFilePermissions(Path sourcePath, AbsoluteUnixPath destinationPath);
 }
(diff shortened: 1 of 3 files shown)


jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FilePermissionsProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.api.buildplan;
+
+import java.nio.file.Path;
+
+/** Serves as a base class for determining file permissions on a container. */
```
loosebazooka, line 21: Remove "Servies as a base class" as an interface is not really a "base class"

Maybe it could be something like "Interface for providing rules to determine file permission on a container"

loosebazooka: and similarly for the other interfaces?

mpeddada1: That is much better, thanks.

jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/FilePermissionsProvider.java
```diff
@@ -0,0 +1,35 @@
...
+public interface FilePermissionsProvider {
+
+  /**
+   * Returns the file permissions that should be set for a path, given the source path and
+   * destination path on a container.
+   *
+   * @param sourcePath the source file.
+   * @param destinationPath the destination path in the container file system corresponding to
+   *     sourcePath.
+   * @return the permissions to be set for the file.
```
chanseokoh, line 32: style super-nit

Although not really enforced, I generally follow these guidelines.

> When writing the comments themselves, in general, start with a phrase and follow it with sentences if they are needed.
> 
> - When writing a phrase, do not capitalize and do not end with a period:
> ```
> @param x the x-coordinate, measured in pixels
> ```
> 
> - When writing a phrase followed by a sentence, do not capitalize the phrase, but end it with a period to distinguish it from the start of the next sentence:
> ```
> @param x the x-coordinate. Measured in pixels.
> ```
> 
> - If you prefer starting with a sentence, capitalize it and end it with a period:
> ```
> @param x Specifies the x-coordinate, measured in pixels.
> ```
> 
> - When writing multiple sentences, follow normal sentence rules:
> ```
> @param x Specifies the x-coordinate. Measured in pixels.
> ```
>

mpeddada1: Ah this does make it easier to understand what the parameters mean.

PR #2652, opened 2020-07-31
build multiple base Images concurrently

louismurerwa:
This PR adds the ability for JIB to build multiple Images concurrently .

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -37,7 +37,6 @@ import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -358,22 +358,27 @@ private void buildAndCacheApplicationLayers() {
...
-                    .call());
+            () -> {
+              List<Future<Image>> builtImages = new ArrayList<>();
+              for (Image image : results.baseImagesAndLayers.get().keySet()) {
+                Future<Image> builtImage =
+                    executorService.submit(
+                        () ->
+                            new BuildImageStep(
+                                    buildContext,
+                                    childProgressDispatcherFactory,
```
chanseokoh, line 371: Just realized this. I actually think this code (and the code in `obtainBaseImageLayers` that we updated) will break once we have multiple `Images` due to `childProgressDispatcherFactory` (the progress reporting thing), which I haven't explained. Will need to fix it when the time comes. Will be easy to fix, so not a concern at all.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -358,22 +358,27 @@ private void buildAndCacheApplicationLayers() {
...
+                                    image,
+                                    realizeFutures(
+                                        Verify.verifyNotNull(
+                                            results.baseImagesAndLayers.get().get(image))),
+                                    realizeFutures(Verify.verifyNotNull(results.applicationLayers)))
+                                .call());
+                builtImages.add(builtImage);
+              }
```
chanseokoh, line 379: You can use `.entrySet()` to iterator over a Map. An entry is a (key, value) pair, so no need to call `Map.get(key)` to get the value.

```suggestion
              for (Map.Entry<Image, List<Future<PreparedLayer>>> entry :
                  results.baseImagesAndLayers.get().entrySet()) {
                Future<Image> builtImage =
                    executorService.submit(
                        () ->
                            new BuildImageStep(
                                    buildContext,
                                    childProgressDispatcherFactory,
                                    entry.getKey(), // base image
                                    realizeFutures(entry.getValue()), // base image layers
                                    realizeFutures(Verify.verifyNotNull(results.applicationLayers)))
                                .call());
                builtImages.add(builtImage);
              }
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -358,22 +358,27 @@ private void buildAndCacheApplicationLayers() {
   private void buildImage() {
```
chanseokoh, line 358: Rename it to `buildImages()`? Also, rename other methods too (`pullBaseImage()` --> `pullBaseImages()`)?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa: Sounds Good .I will be on the look out for the breaks.

PR #2654, opened 2020-08-03
Update build-plan CHANGELOG.md

mpeddada1:

### DIFF
--- a/jib-build-plan/CHANGELOG.md
+++ b/jib-build-plan/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 ### Changed
+- Replaced `BiFunction` usage in `FileEntriesLayer` with `FilePermissionsProvider`, `ModificationTimeProvider` and `OwnershipProvider`. ([#2638](https://github.com/GoogleContainerTools/jib/issues/2638))
 
 ### Fixed
 
@@ -19,7 +20,6 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Removed `get/setOsHint()` and `get/setArchitectureHint()` in favor of `get/setPlatforms()` and `addPlatform()`. ([#2584](https://github.com/GoogleContainerTools/jib/pull/2584))
-- Replaced `BiFunction` usage in `FileEntriesLayer` with `FilePermissionsProvider`, `ModificationTimeProvider` and `OwnershipProvider`. ([#2638](https://github.com/GoogleContainerTools/jib/issues/2638))
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


loosebazooka:
@chanseokoh have we done a release with any of these changes before? I don't see any versioning in here?

chanseokoh:
Ah, yeah, sorry. I forgot to update the template. The current `[unreleased]` entries need to be under 0.3.1. (Strictly speaking, we should have both 0.3.0 and 0.3.1, but I think it's fine to go with a single 0.3.1.)

chanseokoh:
Good catch!

jib-build-plan/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 - Removed `get/setOsHint()` and `get/setArchitectureHint()` in favor of `get/setPlatforms()` and `addPlatform()`. ([#2584](https://github.com/GoogleContainerTools/jib/pull/2584))
+- Replaced `BiFunction` usage in `FileEntriesLayer` with `FilePermissionsProvider`, `ModificationTimeProvider` and `OwnershipProvider`. ([#2643](https://github.com/GoogleContainerTools/jib/pull/2643))
```
chanseokoh, line 14: I think linking #2638 is better for the user looking for more information. #2643 doesn't have any description.

jib-build-plan/CHANGELOG.md
```diff
@@ -5,12 +5,21 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
 - Added `Platform` class representing an image platform. ([#2584](https://github.com/GoogleContainerTools/jib/pull/2584))
 - Added `get/setPlatforms()` and `addPlatform()` to `ContainerBuildPlan` for setting and getting image platforms. ([#2584](https://github.com/GoogleContainerTools/jib/pull/2584))
 
 ### Changed
 
 - Removed `get/setOsHint()` and `get/setArchitectureHint()` in favor of `get/setPlatforms()` and `addPlatform()`. ([#2584](https://github.com/GoogleContainerTools/jib/pull/2584))
+- Replaced `BiFunction` usage in `FileEntriesLayer` with `FilePermissionsProvider`, `ModificationTimeProvider` and `OwnershipProvider`. ([#2638](https://github.com/GoogleContainerTools/jib/issues/2638))
```
chanseokoh, line 22: Now this one should go under `[unreleased]`. 0.3.1 is already released and live.

mpeddada1: Oh right. Sorry about that!

loosebazooka:
Yeah we probably should've provided some better guidance on this file. Anyway, while you're in here can you update the top to look more like https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/CHANGELOG.md#change-log including a title and `[unreleased]` header and stuff

loosebazooka:
I think we still need the "changelog" title and subtitle, and the "unreleased" subsection header?

PR #2656, opened 2020-08-03
Adding functionality to push base image layers and push container config to the registry

louismurerwa:
1. This PR pushes multiple container config JSONs: loops through the built images and schedules multiple pushes in `pushContainerConfiguration()`. Note care has been taken to not block scheduling threads while looping by using the `Future<Image> ` as the key in the resulting `pushResults` map.
2.This PR also pushes multiple baseImageLayers: loops through the `results.baseImagesAndLayers.get()` and schedules multiple pushes in `pushBaseImageLayers()`. Note  the resulting results from the pushes are being stored in a list for now but in the next PR we will update the resulting `pushResults` list into a map so as to track the `builtImage/baseImageLayers` key-pair values.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -393,17 +393,17 @@ public class StepsRunner {
         executorService.submit(
             () -> {
               Map<Future<Image>, Future<BlobDescriptor>> pushResults = new HashMap<>();
-              for (Future<Image> buildImage : results.builtImages.get()) {
-                Future<BlobDescriptor> containerConfigurationPushResult =
+              for (Future<Image> builtImage : results.builtImages.get()) {
+                Future<BlobDescriptor> configPushResult =
                     executorService.submit(
                         () ->
                             new PushContainerConfigurationStep(
                                     buildContext,
                                     childProgressDispatcherFactory,
                                     results.targetRegistryClient.get(),
-                                    buildImage.get())
+                                    builtImage.get())
                                 .call());
-                pushResults.put(buildImage, containerConfigurationPushResult);
+                pushResults.put(builtImage, configPushResult);
               }
               return pushResults;
             });
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -73,8 +73,8 @@ public class StepsRunner {
     private Future<RegistryClient> targetRegistryClient = failedFuture();
     private Future<List<List<Future<BlobDescriptor>>>> baseImageLayerPushResults = failedFuture();
     private Future<List<Future<BlobDescriptor>>> applicationLayerPushResults = failedFuture();
-    private Future<Map<Image, Future<BlobDescriptor>>> containerConfigurationPushResults =
-        failedFuture();
+    private Future<Map<Image, Future<BlobDescriptor>>>
+        builtImagesAndContainerConfigurationPushResults = failedFuture();
     private Future<BuildResult> buildResult = failedFuture();
     private Future<Optional<ManifestAndDigest<ManifestTemplate>>> manifestCheckResult =
         failedFuture();
@@ -389,7 +389,7 @@ public class StepsRunner {
     ProgressEventDispatcher.Factory childProgressDispatcherFactory =
         Verify.verifyNotNull(rootProgressDispatcher).newChildProducer();
 
-    results.containerConfigurationPushResults =
+    results.builtImagesAndContainerConfigurationPushResults =
         executorService.submit(
             () -> {
               Map<Image, Future<BlobDescriptor>> containerConfigurationPushResults =
@@ -439,7 +439,7 @@ public class StepsRunner {
                         results.targetRegistryClient.get(),
                         Verify.verifyNotNull(
                                 results
-                                    .containerConfigurationPushResults
+                                    .builtImagesAndContainerConfigurationPushResults
                                     .get()
                                     .get(results.builtImages.get().get(0).get()))
                             .get(),
@@ -465,7 +465,7 @@ public class StepsRunner {
                           results.targetRegistryClient.get(),
                           Verify.verifyNotNull(
                                   results
-                                      .containerConfigurationPushResults
+                                      .builtImagesAndContainerConfigurationPushResults
                                       .get()
                                       .get(results.builtImages.get().get(0).get()))
                               .get(),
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -332,17 +333,21 @@ private void pushBaseImageLayers() {
...
+                List<Future<BlobDescriptor>> baseImageLayerPushResult =
+                    scheduleCallables(
+                        PushLayerStep.makeList(
+                            buildContext,
+                            childProgressDispatcherFactory,
+                            results.targetRegistryClient.get(),
+                            Verify.verifyNotNull(baseImageLayers)));
+                baseImageLayerPushResults.add(baseImageLayerPushResult);
+              }
+              return baseImageLayerPushResults;
```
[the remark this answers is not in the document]
louismurerwa: Good Point .I was thinking of changing it to a map at a later stage when we start pushing the image.Right now I will keep it as a list so as to make line 457 easy to integrate at this stage.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -384,15 +389,26 @@ private void pushContainerConfiguration() {
...
         executorService.submit(
-            () ->
-                new PushContainerConfigurationStep(
-                        buildContext,
-                        childProgressDispatcherFactory,
-                        results.targetRegistryClient.get(),
-                        results.builtImages.get().get(0).get())
-                    .call());
+            () -> {
+              Map<Image, Future<BlobDescriptor>> containerConfigurationPushResults =
```
chanseokoh, line 395: Let's rename this to `builtImagesAndContainerConfigurationPushResults`.

Using a `Map` is convenient, but compared to using a dedicated class, it's difficult to know what the keys and the values mean and what is the relation between them without documentation or descriptive naming. For a short time, I had to go over the code to see if `Image` in the map is a base image or a built image. We may keep the current name, but in that case, we should add a clear Javadoc to the field.

louismurerwa: I think its a good idea change the name to ` builtImagesAndContainerConfigurationPushResults `.I will also change image to `baseImage` or to `builtImage`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -332,17 +333,21 @@ private void pushBaseImageLayers() {
...
-                        Verify.verifyNotNull(
-                            results
-                                .baseImagesAndLayers
-                                .get()
-                                .get(results.baseImagesAndRegistryClient.get().images.get(0))))));
+            () -> {
+              List<List<Future<BlobDescriptor>>> baseImageLayerPushResults = new ArrayList<>();
+              for (List<Future<PreparedLayer>> baseImageLayers :
+                  results.baseImagesAndLayers.get().values()) {
+                List<Future<BlobDescriptor>> baseImageLayerPushResult =
```
chanseokoh, line 340: In a local scope, I think we don't always have to be fully verbose. IMO, reasonable names help reducing cognitive load when reading code, especially in a local scope. So, I prefer
```java
              List<List<Future<BlobDescriptor>>>  pushResults = new ArrayList<>();
              for (List<Future<PreparedLayer>> layers : results.baseImagesAndLayers.get().values()) {
                List<Future<BlobDescriptor>> pushResult =
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: As mentioned in the other comment, how about
```
Map<Image, Future<BlobDescriptor>> pushResults =
```
? (I think `buildImage` in `for (Future<Image> builtImage` : ...` is good though.)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -384,15 +389,26 @@ private void pushContainerConfiguration() {
...
+                    executorService.submit(
+                        () ->
+                            new PushContainerConfigurationStep(
+                                    buildContext,
+                                    childProgressDispatcherFactory,
+                                    results.targetRegistryClient.get(),
+                                    builtImage.get())
+                                .call());
+                containerConfigurationPushResults.put(
+                    builtImage.get(), containerConfigurationPushResult);
```
chanseokoh, line 408: So I realized a problem with this. `buildImage` in the loop is a `Future`, so calling `builtImage.get()` means blocking until the image is built. Rather, we don't want the loop to block. We may need to put a `Future<Image>` instead of `Image` as a key?

chanseokoh:
Don't forget to update the description.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: An alternative is to make `PushContainerConfigurationStep` return `Image` and `BlobDescriptor` together, but I think that is more involved.

louismurerwa: I see what you mean .Since we have already called `builtImage.get()` in the same loop does it really mater to call it again.

chanseokoh: Note we don't call `buitlImage.get()` inside loop directly, but rather tells `executorService` to spawn a new thread to run the lambda which will later cal lit.

louismurerwa: Ohh I see the logic now

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -384,15 +389,24 @@ private void pushContainerConfiguration() {
...
-            () ->
-                new PushContainerConfigurationStep(
-                        buildContext,
-                        childProgressDispatcherFactory,
-                        results.targetRegistryClient.get(),
-                        results.builtImages.get().get(0).get())
-                    .call());
+            () -> {
+              Map<Future<Image>, Future<BlobDescriptor>> pushResults = new HashMap<>();
+              for (Future<Image> buildImage : results.builtImages.get()) {
```
chanseokoh, line 396:
```suggestion
              for (Future<Image> builtImage : results.builtImages.get()) {
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -384,15 +389,24 @@ private void pushContainerConfiguration() {
...
+                Future<BlobDescriptor> containerConfigurationPushResult =
```
chanseokoh, line 397:
```suggestion
                Future<BlobDescriptor> configPushResult =
```

PR #2663, opened 2020-08-05
Adding ability check the registry for manifests and push `builtImages` concurrently to the server

louismurerwa:
This PR allows jib to pushes the multiple manifest files of the built images to the registry in the case that the manifest dont already exist in the registry.


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -331,20 +333,22 @@ private void pushBaseImageLayers() {
...
+    results.builtImagesAndLayerPushResults =
         executorService.submit(
             () -> {
-              List<List<Future<BlobDescriptor>>> pushResults = new ArrayList<>();
-              for (List<Future<PreparedLayer>> baseImageLayers :
-                  results.baseImagesAndLayers.get().values()) {
+              Map<Future<Image>, List<Future<BlobDescriptor>>> pushResults = new HashMap<>();
+
+              for (Map.Entry<Future<Image>, List<Future<PreparedLayer>>> entry :
+                  results.builtImagesAndLayers.get().entrySet()) {
```
chanseokoh, line 342: I don't think this is correct. As the original code, it should depend on `baseImagesAndLayers`. Depending on `builtImages...` means that this code will block until you complete building `Image`s.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -331,20 +333,22 @@ private void pushBaseImageLayers() {
     ProgressEventDispatcher.Factory childProgressDispatcherFactory =
         Verify.verifyNotNull(rootProgressDispatcher).newChildProducer();
 
-    results.baseImageLayerPushResults =
+    results.builtImagesAndLayerPushResults =
```
chanseokoh, line 336: builtImagesAndBaseImageLayerPushResults

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -379,9 +384,9 @@ private void buildImages() {
                                     realizeFutures(Verify.verifyNotNull(entry.getValue())),
                                     realizeFutures(Verify.verifyNotNull(results.applicationLayers)))
                                 .call());
-                builtImages.add(builtImage);
+                builtImagesAndLayers.put(builtImage, entry.getValue());
```
chanseokoh, line 387: Since we won't use `builtImagesAndLayers` in `pushBaseImageLayers()` (see my other comment), perhaps we don't need to change `buildImages()` code?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -439,80 +444,122 @@ private void checkImageInTargetRegistry() {
...
                                     .get()
-                                    .get(results.builtImages.get().get(0)))
+                                    .values()
+                                    .iterator()
+                                    .next())
                             .get(),
-                        results.builtImages.get().get(0).get())
+                        Verify.verifyNotNull(
+                                results
+                                    .builtImagesAndContainerConfigurationPushResults
```
chanseokoh, line 453: I think we should keep using `builtImages`. I think using `builtImagesAndContainerConfigurationPushResults` to get an `Image` causes confusion and feels like abusing.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -439,80 +444,122 @@ private void checkImageInTargetRegistry() {
                                 results
                                     .builtImagesAndContainerConfigurationPushResults
                                     .get()
-                                    .get(results.builtImages.get().get(0)))
+                                    .values()
```
chanseokoh, line 447: Do we really need this change? I was aware that we can always use `.iterator().next()` to retrieve a value from a single-entry map, but you've been so far consistently calling `Map.get()`. Mixing patterns makes it harder to understand code.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -439,80 +444,122 @@ private void checkImageInTargetRegistry() {
...
-                      Verify.verifyNotNull(
-                              results
-                                  .builtImagesAndContainerConfigurationPushResults
-                                  .get()
-                                  .get(results.builtImages.get().get(0)))
-                          .get()
-                          .getDigest())
-                  // Manifest pushers return the same BuildResult.
-                  : manifestPushResults.get(0).get();
+                                      .get(entry.getKey())));
```
chanseokoh, line 483: It takes some seconds to get what `entry.getKey()` means. I suggest declaring a local variable `Future<Image> builtImage = entry.getKey()` before this `realizeFutures()`. The same goes for `entry.getValue()`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -439,80 +444,122 @@ private void checkImageInTargetRegistry() {
...
+                          new LoadDockerStep(
+                                  buildContext,
+                                  childProgressDispatcherFactory,
+                                  dockerClient,
+                                  results
+                                      .builtImagesAndLayers
+                                      .get()
+                                      .keySet()
+                                      .iterator()
+                                      .next()
```
chanseokoh, line 529: ditto about the `.iterator().next()` practice.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -439,80 +444,122 @@ private void checkImageInTargetRegistry() {
...
+                      () ->
+                          new WriteTarFileStep(
+                                  buildContext,
+                                  childProgressDispatcherFactory,
+                                  outputPath,
+                                  results
+                                      .builtImagesAndLayers
+                                      .get()
+                                      .keySet()
+                                      .iterator()
```
chanseokoh, line 556: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa: The logic here is to be able to keep track of `builtimage -> baseImageLayerPushResults` so that we can be able to use `realizeFutures` on the correct `baseImageLayerPushResults` in the `pushImages` step.

The reason why I am depending on `builtImages` here is because inorder to be able to match `builtImage -> baseImageLayerPushResults` you need to have the `builtImage -> baseImageLayer` mapping first which I create in the `buildimage` function.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa, line 483: @chanseokoh This is the line that has caused all the changes in the structure of the code .Do we really need to `realizeFutures(baseImageLayerPushResults) ` here ?

Is there any way we could for-go realizing these futures here? 

Once we have answered this question ,I can easily refactor the code.

louismurerwa: @chanseokoh This is the line that has caused all the changes in the structure of the code .Do we really need to `realizeFutures(baseImageLayerPushResults) ` here ?

Is there any way we could for-go realizing these futures here? 

Once we have answered this question ,I can easily refactor the code.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: I can understand why it ended up like this, but I think this is not correct for a few reasons. I will just mention one reason that can alone make this not acceptable: 

`results.builtImagesAndLayers` is the result coming from scheduling a thread in `buildImages()`. And calling `results.builtImagesAndLayers.get()` here means you will wait until the thread scheduled in `buildImages()` runs to completion. Conceptually, it means you'll wait until "images" are built. The code will now speak that "in order to be able to push base image layers, I need to build images first." OTOH, previously, this `pushBaseImageLayers()` didn't have any dependency on `buildImages()`. Pushing base image layers can start as soon as we have base image layers. But with this change, you can start pushing only after the thread from `buildImages()` completed. This is a significant departure from the current async dependencies behaviors. Whenever we change the original intention, there should be a clear reason; just to allow making some other parts technically easy to make it work with minimal changes can't be a reason.

louismurerwa: That makes sense - this implementation does slow down things.I'l revert this to the previous implementation and we can make changes starting from there.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: Every line of code is there for a reason. We have`realizeFutures()` in the original code, so I assume there must be a reason. If you can't or don't know if what the existing code does or why it is there, you should assume it's necessary. Even when you think some code is unnecessary, you need to think through and really spend a lot of time before taking out some code.

In this case, I think it's definitely necessary. You can push a manifest only after confirmed that you've done pushing layers. Otherwise, a registry will return "bad manifest push; layers described in the manifest doesn't exist." And I can see the only place where we confirm the base image push results is here.

louismurerwa:
Still working on responding to reviews

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -331,20 +332,23 @@ private void pushBaseImageLayers() {
     ProgressEventDispatcher.Factory childProgressDispatcherFactory =
         Verify.verifyNotNull(rootProgressDispatcher).newChildProducer();
 
-    results.baseImageLayerPushResults =
+    results.baseImagesAndbaseImageLayerPushResults =
```
chanseokoh, line 335:
```suggestion
    results.baseImagesAndBaseImageLayerPushResults =
```
but because this time it's all about base images, I think we can go with
```suggestion
    results.baseImagesAndLayerPushResults =
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -379,21 +383,21 @@ private void buildImages() {
                                     realizeFutures(Verify.verifyNotNull(entry.getValue())),
                                     realizeFutures(Verify.verifyNotNull(results.applicationLayers)))
                                 .call());
-                builtImages.add(builtImage);
+                builtImagesAndBaseImages.put(builtImage, entry.getKey());
```
chanseokoh, line 386:
```suggestion
                builtImagesAndBaseImages.put(builtImage, entry.getKey() /* base Image */);
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +434,138 @@ private void checkImageInTargetRegistry() {
...
-                                    .builtImagesAndContainerConfigurationPushResults
-                                    .get()
-                                    .get(results.builtImages.get().get(0)))
-                            .get(),
-                        results.builtImages.get().get(0).get())
-                    .call());
+            () -> {
+              Map.Entry<Future<Image>, Image> entry =
+                  results.builtImagesAndBaseImages.get().entrySet().iterator().next();
+              Future<Image> builtImage = entry.getKey();
```
chanseokoh, line 440: You only need the key, so no need for `entrySet()`. And let's add a check; I'm certain thecheck will become useful going forward.
```suggestion
              Verify.verify(results.builtImagesAndBaseImages.get().size() == 1);
              Future<Image> builtImage = results.builtImagesAndBaseImages.get().keySet().iterator().next();
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +434,138 @@ private void checkImageInTargetRegistry() {
...
-                new LoadDockerStep(
-                        buildContext,
-                        childProgressDispatcherFactory,
-                        dockerClient,
-                        results.builtImages.get().get(0).get())
-                    .call());
+            () -> {
+              Map.Entry<Future<Image>, Image> entry =
+                  results.builtImagesAndBaseImages.get().entrySet().iterator().next();
+              Future<Image> builtImage = entry.getKey();
```
chanseokoh, line 528:
```suggestion
              Verify.verify(results.builtImagesAndBaseImages.get().size() == 1, "multi-platform image building not supported when pushing to Docker engine");
              Future<Image> builtImage = results.builtImagesAndBaseImages.get().keySet().iterator().next();
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +434,138 @@ private void checkImageInTargetRegistry() {
...
+
+              List<Future<BuildResult>> buildResults = new ArrayList<>();
+              Future<BuildResult> buildResult =
+                  executorService.submit(
```
chanseokoh, line 532: Can be simplified by `return Collections.singletonList(executorService.submit(...))`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +434,138 @@ private void checkImageInTargetRegistry() {
...
-                new WriteTarFileStep(
-                        buildContext,
-                        childProgressDispatcherFactory,
-                        outputPath,
-                        results.builtImages.get().get(0).get())
-                    .call());
+            () -> {
+              Map.Entry<Future<Image>, Image> entry =
+                  results.builtImagesAndBaseImages.get().entrySet().iterator().next();
+              Future<Image> builtImage = entry.getKey();
```
chanseokoh, line 554:
```suggestion
              Verify.verify(results.builtImagesAndBaseImages.get().size() == 1, "multi-platform image building not supported when building a local tar image");
              Future<Image> builtImage = results.builtImagesAndBaseImages.get().keySet().iterator().next();
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +434,138 @@ private void checkImageInTargetRegistry() {
...
+
+              List<Future<BuildResult>> buildResults = new ArrayList<>();
```
chanseokoh, line 556: Ditto. No need to create an array.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +435,133 @@ private void checkImageInTargetRegistry() {
...
+              return Collections.singletonList(
+                  executorService.submit(
+                      () ->
+                          new LoadDockerStep(
+                                  buildContext,
+                                  childProgressDispatcherFactory,
+                                  dockerClient,
+                                  builtImage.get())
+                              .call()));
+            });
```
chanseokoh, line 539: I think we can block until we get the built `Image`?

```java
              Image builtImage =
                  results.builtImagesAndBaseImages.get().keySet().iterator().next().get();
```

Next, I think we should avoid scheduling another thread. We can use `Futures.immediateFuture` (a utility method to directly instantiate a `Future` from a concrete value) that just wraps a value instead of actually scheduling a thread.

```java
              Image builtImage =
                  results.builtImagesAndBaseImages.get().keySet().iterator().next().get();

              return Collections.singletonList(
                  Futures.immediateFuture(
                      new LoadDockerStep(
                              buildContext,
                              childProgressDispatcherFactory,
                              dockerClient,
                              builtImage)
                          .call()));
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +435,133 @@ private void checkImageInTargetRegistry() {
...
-                    .call());
+            () -> {
+              Verify.verify(
+                  results.builtImagesAndBaseImages.get().size() == 1,
+                  "multi-platform image building not supported when building a local tar image");
+              Future<Image> builtImage =
+                  results.builtImagesAndBaseImages.get().keySet().iterator().next();
+
+              return Collections.singletonList(
+                  executorService.submit(
```
chanseokoh, line 556: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: And I think this is more readable:
```java
              BuildResult buildResult =
                  new LoadDockerStep(
                          buildContext, childProgressDispatcherFactory, dockerClient, builtImage)
                      .call();
              return Collections.singletonList(Futures.immediateFuture(buildResult));
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -430,89 +435,133 @@ private void checkImageInTargetRegistry() {
...
+                              scheduleCallables(
+                                  PushImageStep.makeList(
+                                      buildContext,
+                                      childProgressDispatcherFactory,
+                                      results.targetRegistryClient.get(),
+                                      Verify.verifyNotNull(
+                                              results
+                                                  .builtImagesAndContainerConfigurationPushResults
+                                                  .get()
+                                                  .get(builtImage))
```
chanseokoh, line 491: This is used on line 500, so I suggest refactor out into a local variable. I think it will considerable improve readability.

louismurerwa:
LGTM

Issue #413, opened 2018-06-18
Consider options for setting image creation time while maintaining reproducibility


TadCordle:
Followup from discussion: https://github.com/GoogleContainerTools/jib/pull/412#discussion_r196172710

patflynn:
this sounds infeasible.

TadCordle:
Why? I think this thread brings up some feasible suggestions https://groups.google.com/forum/#!msg/istio-dev/nuv6UaBFGDk/sR-dfoc9CgAJ

If you don't think it's worth it we can close this issue.

briandealwis:
Tycho, a set of extensions to Maven for building Eclipse/OSGi bundles that we use for CT4E, generates a timestamp using a pluggable provider.  Tycho provides a [git-based timestamp provider](https://github.com/eclipse/tycho.extras/tree/master/tycho-buildtimestamp-jgit) that uses [JGit](https://github.com/eclipse/jgit) to determine the timestamp of the last commit within a project directory.

[This is it in action](https://github.com/GoogleCloudPlatform/google-cloud-eclipse/blob/master/pom.xml#L254) in CT4E: the `tycho-packaging-plugin` is responsible for generating the .jar files with the version and timestamp.  [This is the class](https://github.com/eclipse/tycho/blob/master/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java) that actually generates the version and timestamp.

coollog:
Since there are a lot of issues asking about this, we should probably provide a solution for users. Proposal:

Add a `useCurrentTimestamp` configuration that would set `created` on the container to the current time. For example, in Maven:

```xml
<configuration>
  <container>
    <useCurrentTimestamp>true</useCurrentTimestamp>
  </container>
</configuration>
```

When using this configuration, the build log will also display a warning saying that the image is not reproducible because `useCurrentTimestamp` is enabled.

This is not letting users add their own timestamps because we would have to most likely implement custom parsers to convert into the [format that is necessary for the container configuration](https://tools.ietf.org/html/rfc3339#section-5.6).

lovelivestyle1102:
It doesn't work, when I config just like this : 
            <plugin>
                <groupId>com.google.cloud.tools</groupId>
                <artifactId>jib-maven-plugin</artifactId>
                <version>0.9.6</version>
                <configuration>
                    <to>
                        <image>myimage</image>
                    </to>

                   <container>
[...]

PR #2664, opened 2020-08-05
Fix FAQ documentation regarding useCurrentTimestamp

arlowhite:
0.9.7 changed configuration, see: https://github.com/GoogleContainerTools/jib/issues/413

creationTime no longer works.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -113,10 +113,10 @@ See [Extended Usage](../jib-gradle-plugin#extended-usage) for the `container.for
 
 ### Why is my image created 48+ years ago?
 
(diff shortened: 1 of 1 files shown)


arlowhite:
I have not tested the Gradle config.

Also, does this last sentence need to be updated?
> Note that the modification time of the files in the built image put by Jib will still be 1 second past the epoch. The file modification time can be configured using <container><filesModificationTime> (Maven) or jib.container.filesModificationTime (Gradle).

arlowhite:
Other documentation needs updating too, such as
https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#quickstart

chanseokoh:
The current documentation should be correct. We deprecated `jib.container.useCurrentTimestamp` in favor of `creationTime` a while ago.: https://stackoverflow.com/a/61163067/1701388

arlowhite:
Sorry, you're right. I started with https://www.baeldung.com/jib-dockerizing which is really old (version 0.9.10)
I should have checked the latest version or followed the official Jib quickstart. I'll ping baeldung about it

chanseokoh:
Thank you!

PR #2666, opened 2020-08-06
Allow main() to be synthetic in MainClassFinder

frost13it:
Kotlin compiler allows main() method to take no arguments:
```kotlin
fun main() {
    println("Hello, World!")
}
```

Under the hood it creates a synthetic method with a signature that is acceptable by JVM (`main(java.lang.String[])`):
```
  public static final void main();
    descriptor: ()V
    flags: (0x0019) ACC_PUBLIC, ACC_STATIC, ACC_FINAL
    Code:
      stack=2, locals=2, args_size=0
         0: ldc           #11                 // String Hello, World!
         2: astore_0
         3: iconst_0
         4: istore_1
         5: getstatic     #17                 // Field java/lang/System.out:Ljava/io/PrintStream;
         8: aload_0
         9: invokevirtual #23                 // Method java/io/PrintStream.println:(Ljava/lang/Object;)V
        12: return
      LineNumberTable:
        line 2: 0
        line 3: 12

  public static void main(java.lang.String[]);
    descriptor: ([Ljava/lang/String;)V
    flags: (0x1009) ACC_PUBLIC, ACC_STATIC, ACC_SYNTHETIC
    Code:
      stack=0, locals=1, args_size=1
         0: invokestatic  #9                  // Method main:()V
         3: return
```

But jib does not recognize such a class as the application's entrypoint.

This PR adds `ACC_SYNTHETIC` to list of allowed `main()` access flags.

The `HelloWorldKt.class` in tests is compiled from the code above with kotlinc-jvm 1.3.72 using default settings.


frost13it:
@googlebot I signed it!

chanseokoh:
@GoogleContainerTools/java-tools-build added CHANGELOG enties.

chanseokoh:
@frost13it Jib Gradle 2.5.0 and Jib Maven 2.5.2 have been released with your fix. Thanks for your contribution!

frost13it:
@chanseokoh Thanks for such a quick release!

PR #2667, opened 2020-08-06
Preserve trailing slash metadata

loosebazooka:
Preserve trailing slash information in `dest` property. This should not be done in `AbsoluteUnixPath` because that is agnostic to the filesystem. It merely contains a list of paths. Preserve the information here and use within the `jib-cli` project.


loosebazooka:
I'm not super fond of my naming choices, but I can revisit later given it's mostly internal.

PR #2670, opened 2020-08-06
Update FAQ and plugin reference docs about selecting base image platform from a manifest list

chanseokoh:
Should be merge only after releasing Jib 2.5.0.

The FAQ needs to be correct and helpful, as we show the link to the FAQ in our Jib error message if Jib was unable to find a matching platform. @louismurerwa

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -469,7 +469,7 @@ To inspect the image that is produced from the build using Docker, you can use c
 
 ### How do I specify a platform in the manifest list (or OCI index) of a base image?
 
-Jib 2.5.0 added an _incubating feature_ that provides limited support for selecting a base image with the describe platform from a manifest list. For example,
+Jib 2.5.0 added an _incubating feature_ that provides limited support for selecting a base image with the desired platform from a manifest list. For example,
 
 ```xml
   <from>
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -469,18 +469,46 @@ To inspect the image that is produced from the build using Docker, you can use c
 
 ### How do I specify a platform in the manifest list (or OCI index) of a base image?
 
-By design, if the target image reference is a [manifest list](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list), Jib will always select an image for the platform `amd64/linux`. If you need to specify a different image from a manifest list you must specify the digest for the platform you are targeting.
+Jib 2.5.0 added an _incubating feature_ that provides limited support for selecting a base image with the describe platform from a manifest list. For example,
 
-To view a manifest, [enable experimental docker CLI](https://docs.docker.com/engine/reference/commandline/cli/#experimental-features) features and then run the [manifest inspect](https://docs.docker.com/engine/reference/commandline/manifest_inspect/) command.
+```xml
+  <from>
+    <image>... image reference to a manifest list ...</image>
+    <platforms>
+      <platform>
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -469,18 +469,46 @@ To inspect the image that is produced from the build using Docker, you can use c
...
-You can then inspect the output for the specific image you want (in this example an image for the platform `arm64/linux`)
-```java
+The default platform is "amd64/linux" if not specified, whose behavior is backward-compatible.
+
+As an incubating feature, there are certain limitations:
+- OCI image indices are not supported (as opposed to Docker manifest lists).
+- Supports specifying only one platform.
+- Only `architecture` and `os` are supported. If there are multiple base images with given architecture and os, the first image will be selected.
+
+Make sure to specify a manifest _list_ in `<from><image>` (whether by a tag name or a digest (`@sha256:...`)). For troubleshooting, you may want to check what platforms a manifest list contains. To view a manifest, [enable experimental docker CLI](https://docs.docker.com/engine/reference/commandline/cli/#experimental-features) features and then run the [manifest inspect](https://docs.docker.com/engine/reference/commandline/manifest_inspect/) command.
```
chanseokoh, line 505: @louismurerwa what happens as of now if the user is currently specifying a manifest (as opposed to a manifest list) of a non-default platform with Jib 2.4.0 and upgrades to 2.5.0? Will it fail or just shows a warning?

louismurerwa: User specifies a manifest of a non-default platform
Jib 2.4.0 - This is the old Jib - it just builds using the manifest it found - it doesnt care about the platform of the manifest
Jib 2.5.0 - This is the the new Jib release - It also builds using the manifest it finds -
https://github.com/GoogleContainerTools/jib/blob/4fae9f105d0c4f4c3796607f047a06f54b546668/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L234

chanseokoh: I see. So there's no change on that front; no warning or error, right? I know you've been keep tracking of tasks to do in a separate doc. If you prefer, you can use GitHub as a tracking tool by creating issues. But eventually, I planned to have you file issues anyway.

louismurerwa: I will track this in the doc and file an issue at the end of the summer if I dont fix it then.

chanseokoh: Sounds good. Other than that, do the overall doc changes look OK?

chanseokoh:
@GoogleContainerTools/java-tools-build since 2.5.0 is released, we need to update the docs soon.

docs/faq.md
```diff
@@ -469,18 +469,46 @@ To inspect the image that is produced from the build using Docker, you can use c
...
 ```
 
-You can then inspect the output for the specific image you want (in this example an image for the platform `arm64/linux`)
-```java
+The default platform is "amd64/linux" if not specified, whose behavior is backward-compatible.
+
+As an incubating feature, there are certain limitations:
+- OCI image indices are not supported (as opposed to Docker manifest lists).
+- Supports specifying only one platform.
+- Only `architecture` and `os` are supported. If there are multiple base images with given architecture and os, the first image will be selected.
```
louismurerwa, line 503: Only `architecture` and `os` are supported. If there are multiple base images with given architecture and os in the manifest list, the first image will be selected.

Issue #2650, opened 2020-07-31
Can't push image from local machine to public Docker Hub Registry getting 401 error


uthircloud:
I have simple Spring Boot app. I tried to create an docker image of it using jib-maven plugin. But when I build this app its failing with 401 Unauthorized exception. But I am giving all the configurations correctly in logs I can see its been read correctly but while pushing its failing with I/O error for image..

<plugin>
    <groupId>com.google.cloud.tools</groupId>
    <artifactId>jib-maven-plugin</artifactId>
    <version>2.4.0</version>
    <configuration>
         <from>
            <image>openjdk:latest</image>
            <auth>
              <username>userName</username>
              <password>password</password>
            </auth>
        </from> 
         <to>
            <image><userName>/java-spring/${project.artifactId}:latest</image>
            <auth>
              <username>userName</username>
              <password>password</password>
            </auth>
        </to> 
        <container>
            <jvmFlags>
                <jvmFlag>-Xms512m</jvmFlag>
                <jvmFlag>-Xmx512m</jvmFlag>
            </jvmFlags>
        </container>
    </configuration> 
   <executions>
    <execution>
      <phase>package</phase>
      <goals>
        <goal>build</goal>
      </goals>
    </execution>
  </executions>
</plugin>

I am using mvn clean install -Djib.httpTimeout=0 -X while building

[DEBUG] bearer auth succeeded for registry-1.docker.io/<userName>/java-spring/singin
[DEBUG] TIMED   Authenticating push to registry-1.docker.io : 3433.0 ms
[DEBUG] TIMING  Preparing layer pushers
[DEBUG] TIMED   Preparing layer pushers : 2.0 ms
[DEBUG] TIMING  Pushing BLOB digest: sha256:6a1d32d7f6ff4a41815f4be581691138a588a3e8b4faa839e1e80f7e51b2cd6f, size: 1544
[DEBUG] TIMING  Pushing BLOB digest: sha256:fe3d76ed81f8e2c7419641c7accf6a0f1c4d15a5b5e344d5848422d7e481edd2, size: 314
[DEBUG] refreshing bearer auth token for registry-1.docker.io/<userName>/java-spring/singin...
[DEBUG] refreshing bearer auth token for registry-1.docker.io/<userName>/java-spring/singin...
[DEBUG] TIMED   Pulling base image manifest : 4079.0 ms
[DEBUG] TIMING  Preparing base image layer pullers
[...]

PR #2675, opened 2020-08-06
Update FAQ about 401 Unauthorized to mention the case of Docker Hub multi-level repository names

chanseokoh:
Helps #2650.


docs/faq.md
```diff
@@ -536,7 +536,9 @@ If the registry returns `401 Unauthorized` or `"code":"UNAUTHORIZED"`, it is oft
 * Different auth configurations exist in multiple places, and Jib is not picking up the auth information you are working on.
 * You configured a credential helper, but the helper is not on `$PATH`. This is especially common when running Jib inside IDE where the IDE binary is launched directly from an OS menu and does not have access to your shell's environment.
 * Configured credentials have access to the base image repository but not to the target image repository (or vice versa).
-* Typos in username, password, image names, or registry names.
+* Typos in username, password, image names, repository names, or registry names. (This is surprisingly common.)
```
loosebazooka, line 539: Don't think we need the parenthesis around `This is surprisingly common`. The use of the word surprisingly here has a strange feel to me as well? Perhaps, "This is a very common error."

chanseokoh: Ah, that sounds less colloquial and better.

PR #2683, opened 2020-08-08
Add defaults to FilePropertiesStack

loosebazooka:
The way we use `FileEntriesLayer.addEntry` requires non-null values to be entered. So work with defaults directly here, so we can always pass all properties to that method instead of complicated logic to determine based on null values which `addEntry` to call.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
@@ -86,15 +86,7 @@ class FilePropertiesStack {
       group = properties.getGroup().orElse(group);
     }
     // ownership calculations
-    if (group == null && user == null) {
-      ownership = "";
-    } else if (group == null && /* Nullaway */ user != null) {
-      ownership = user;
-    } else if (user == null) {
-      ownership = ":" + group;
-    } else {
-      ownership = user + ":" + group;
-    }
+    ownership = (user != null ? user : "") + (group != null ? ":" + group : "");
   }
 
   public FilePermissions getFilePermissions() {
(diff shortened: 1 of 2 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
```diff
@@ -34,10 +34,23 @@
...
+  /** Create new FilePropertiesStack with defaults. */
+  public FilePropertiesStack() {
+    setDefaults();
+  }
+
+  private void setDefaults() {
+    filePermissions = FilePermissions.DEFAULT_FILE_PERMISSIONS;
+    directoryPermissions = FilePermissions.DEFAULT_FOLDER_PERMISSIONS;
+    modificationTime = FileEntriesLayer.DEFAULT_MODIFICATION_TIME;
+    // TODO: get default from FileEntriesLayer (requires buildplan release)
```
chanseokoh, line 51: Do we need another release? 0.5.0? I don't recall new changes since 0.4.0.

loosebazooka: I mean if we were to define a default in the build plan, this would be required.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
```diff
@@ -76,7 +86,9 @@ private void updateProperties() {
       group = properties.getGroup().orElse(group);
     }
     // ownership calculations
-    if (group == null) {
+    if (group == null && user == null) {
```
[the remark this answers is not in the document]
loosebazooka: yeah these look better, lemme update

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/FilePropertiesStack.java
loosebazooka: The thing is, I think there might be some other minor additions to the buildplan that might come up, so we can hold off on this.

PR #2699, opened 2020-08-12
Fix child progress dispatcher factory

louismurerwa:
This PR fixes the `java.lang.IllegalStateException` resulting from the multiple use of a single progress dispatcher factory.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -320,17 +320,18 @@ public class StepsRunner {
 
               Map<Image, List<Future<PreparedLayer>>> baseImagesAndLayers = new HashMap<>();
               for (Image image : results.baseImagesAndRegistryClient.get().images) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 List<Future<PreparedLayer>> layers =
                     scheduleCallables(
                         layersRequiredLocally
                             ? ObtainBaseImageLayerStep.makeListForForcedDownload(
                                 buildContext,
-                                progressDispatcher.newChildProducer(),
+                                progressDispatcherFactory,
                                 image,
                                 results.baseImagesAndRegistryClient.get().registryClient)
                             : ObtainBaseImageLayerStep.makeListForSelectiveDownload(
                                 buildContext,
-                                progressDispatcher.newChildProducer(),
+                                progressDispatcherFactory,
                                 image,
                                 results.baseImagesAndRegistryClient.get().registryClient,
                                 results.targetRegistryClient.get()));
@@ -360,6 +361,7 @@ public class StepsRunner {
               Map<Image, List<Future<BlobDescriptor>>> pushResults = new HashMap<>();
               for (Map.Entry<Image, List<Future<PreparedLayer>>> entry :
                   results.baseImagesAndLayers.get().entrySet()) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 Image baseImage = entry.getKey();
                 List<Future<PreparedLayer>> baseImageLayers = entry.getValue();
 
@@ -367,7 +369,7 @@ public class StepsRunner {
                     scheduleCallables(
                         PushLayerStep.makeList(
                             buildContext,
-                            progressDispatcher.newChildProducer(),
+                            progressDispatcherFactory,
                             results.targetRegistryClient.get(),
                             Verify.verifyNotNull(baseImageLayers)));
                 pushResults.put(baseImage, baseImageLayerPushResult);
@@ -404,12 +406,13 @@ public class StepsRunner {
               Map<Future<Image>, Image> builtImagesAndBaseImages = new HashMap<>();
               for (Map.Entry<Image, List<Future<PreparedLayer>>> entry :
                   results.baseImagesAndLayers.get().entrySet()) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 Future<Image> builtImage =
                     executorService.submit(
                         () ->
                             new BuildImageStep(
                                     buildContext,
-                                    progressDispatcher.newChildProducer(),
+                                    progressDispatcherFactory,
                                     entry.getKey(), // base Image
                                     realizeFutures(
                                         Verify.verifyNotNull(entry.getValue())), // layers
@@ -440,12 +443,13 @@ public class StepsRunner {
 
               Map<Future<Image>, Future<BlobDescriptor>> pushResults = new HashMap<>();
               for (Future<Image> builtImage : results.builtImagesAndBaseImages.get().keySet()) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 Future<BlobDescriptor> configPushResult =
                     executorService.submit(
                         () ->
                             new PushContainerConfigurationStep(
                                     buildContext,
-                                    progressDispatcher.newChildProducer(),
+                                    progressDispatcherFactory,
                                     results.targetRegistryClient.get(),
                                     builtImage.get())
                                 .call());
@@ -514,9 +518,9 @@ public class StepsRunner {
               List<Future<BuildResult>> buildResults = new ArrayList<>();
               for (Map.Entry<Future<Image>, Image> entry :
                   results.builtImagesAndBaseImages.get().entrySet()) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 buildResults.add(
-                    pushImage(
-                        entry.getKey(), entry.getValue(), progressDispatcher.newChildProducer()));
+                    pushImage(entry.getKey(), entry.getValue(), progressDispatcherFactory));
               }
               return buildResults;
             });
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -404,7 +404,8 @@ public class StepsRunner {
               Map<Future<Image>, Image> builtImagesAndBaseImages = new HashMap<>();
               for (Map.Entry<Image, List<Future<PreparedLayer>>> entry :
                   results.baseImagesAndLayers.get().entrySet()) {
-                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
+                ProgressEventDispatcher.Factory progressDispatcherFactory =
+                    progressDispatcher.newChildProducer();
                 Future<Image> builtImage =
                     executorService.submit(
                         () ->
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -304,22 +304,28 @@ private void pullBaseImages() {
   private void obtainBaseImageLayers(boolean layersRequiredLocally) {
     ProgressEventDispatcher.Factory childProgressDispatcherFactory =
         Verify.verifyNotNull(rootProgressDispatcher).newChildProducer();
+
     results.baseImagesAndLayers =
         executorService.submit(
             () -> {
+              ProgressEventDispatcher progressDispatcher =
```
chanseokoh, line 311: This is a temporary workaround to overcome the runtime failure. But I can accept this to move forward. Eventually, we should implement a permanent fix. Please keep track of this issue too.

```suggestion
              // TODO: ideally, progressDispatcher should be closed at the right moment, after the scheduled threads have completed. However, it can be tricky and cumbersome to track completion, so it may just be better to delay closing until everything ends. At least, we must ensure that it's not closed prematurely. (Garbage collection doesn't auto-close it wit the current implementation.)
              ProgressEventDispatcher progressDispatcher =
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -377,6 +392,15 @@ private void buildImages() {
...
         executorService.submit(
             () -> {
+              // TODO: ideally, progressDispatcher should be closed at the right moment, after the
+              // scheduled threads have completed. However, it can be tricky and cumbersome to track
+              // completion, so it may just be better to delay closing until everything ends. At
+              // least, we must ensure that it's not closed prematurely. (Garbage collection doesn't
+              // auto-close it wit the current implementation.)
+              ProgressEventDispatcher progressDispatcher =
+                  childProgressDispatcherFactory.create(
+                      "scheduling building images", results.baseImagesAndLayers.get().size());
```
chanseokoh, line 402: "building manifests" sounds more informative to the user.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -404,14 +428,24 @@ private void pushContainerConfigurations() {
...
             () -> {
+              // TODO: ideally, progressDispatcher should be closed at the right moment, after the
+              // scheduled threads have completed. However, it can be tricky and cumbersome to track
+              // completion, so it may just be better to delay closing until everything ends. At
+              // least, we must ensure that it's not closed prematurely. (Garbage collection doesn't
+              // auto-close it wit the current implementation.)
+              ProgressEventDispatcher progressDispatcher =
+                  childProgressDispatcherFactory.create(
+                      "scheduling pushing container configurations",
+                      results.builtImagesAndBaseImages.get().keySet().size());
```
chanseokoh, line 439:
```suggestion
                      results.builtImagesAndBaseImages.get().size());
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -465,13 +499,24 @@ private void pushImages() {
...
             () -> {
+              // TODO: ideally, progressDispatcher should be closed at the right moment, after the
+              // scheduled threads have completed. However, it can be tricky and cumbersome to track
+              // completion, so it may just be better to delay closing until everything ends. At
+              // least, we must ensure that it's not closed prematurely. (Garbage collection doesn't
+              // auto-close it wit the current implementation.)
+              ProgressEventDispatcher progressDispatcher =
+                  childProgressDispatcherFactory.create(
+                      "scheduling pushing images",
+                      results.builtImagesAndBaseImages.get().keySet().size());
```
chanseokoh, line 510:
```suggestion
                      results.builtImagesAndBaseImages.get().size());
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -465,13 +499,24 @@ private void pushImages() {
...
         executorService.submit(
             () -> {
+              // TODO: ideally, progressDispatcher should be closed at the right moment, after the
+              // scheduled threads have completed. However, it can be tricky and cumbersome to track
+              // completion, so it may just be better to delay closing until everything ends. At
+              // least, we must ensure that it's not closed prematurely. (Garbage collection doesn't
+              // auto-close it wit the current implementation.)
+              ProgressEventDispatcher progressDispatcher =
+                  childProgressDispatcherFactory.create(
+                      "scheduling pushing images",
```
chanseokoh, line 509: pushing manifests

chanseokoh:
There's certainly one more thing to consider: https://github.com/GoogleContainerTools/jib/pull/1799#issue-290385108

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -379,7 +409,7 @@ private void buildImages() {
                         () ->
                             new BuildImageStep(
                                     buildContext,
-                                    childProgressDispatcherFactory,
+                                    progressDispatcher.newChildProducer(),
```
chanseokoh, line 412: https://github.com/GoogleContainerTools/jib/pull/2699#issuecomment-673148403

Almost all of classes are immutable, so we didn't have to think about synchronization when accessing instances in the multi-threaded environment. However, `progressDispatcher` is an exception (understandably as a class to report progress updates to the user).

So, it is unsafe to access `progressDispatcher` from multiple threads. That is, calling `newChildProducer()` here needs synchronization; it's being called in multiple threads launched through the loop.

The simplest solution I think is to call `newChildProducer()` outside `executorService.submit()` on line 408.

louismurerwa: I see so that means creating a variable to hold the 'progressDispatcher.newChildProducer()` outside of the loop and then use the variable instead in the threads?

chanseokoh: Yes. You can actually see it's what the current code does in `StepsRunner`. For example,

```java
  private void pushApplicationLayers() {

    results.applicationLayerPushResults =
        executorService.submit(
            () ->
                scheduleCallables(..., childProgressDispatcherFactory, ...)
  }
```

You won't need `Verify.verifyNotNull()` in your case. For naming, since `childProgressDispatcherFactory` is already taken, I think `progressDispatcherFactory` works.

louismurerwa: Yes from the looks of it ,its creating a `newProducer() ` out side of the `executorSubmiter`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -459,13 +503,24 @@ private void pushImages() {
...
+
               realizeFutures(results.applicationLayerPushResults.get());
 
               List<Future<BuildResult>> buildResults = new ArrayList<>();
               for (Map.Entry<Future<Image>, Image> entry :
                   results.builtImagesAndBaseImages.get().entrySet()) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 buildResults.add(
-                    pushImage(entry.getKey(), entry.getValue(), childProgressDispatcherFactory));
+                    pushImage(entry.getKey(), entry.getValue(), progressDispatcherFactory));
```
chanseokoh, line 523: It's ok as long as you are not accessing `progressDispatcher` concurrently. Let's keep the previous code.
```suggestion
                    pushImage(entry.getKey(), entry.getValue(), progressDispatcherFactory.newChildProducer()));
```

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -336,17 +348,28 @@ private void pushBaseImageLayers() {
...
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 Image baseImage = entry.getKey();
                 List<Future<PreparedLayer>> baseImageLayers = entry.getValue();
 
                 List<Future<BlobDescriptor>> baseImageLayerPushResult =
                     scheduleCallables(
                         PushLayerStep.makeList(
                             buildContext,
-                            childProgressDispatcherFactory,
+                            progressDispatcherFactory,
```
chanseokoh, line 372: Ditto, for simplicity, just `progressDispatcher.newChildProducer()`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -304,22 +304,34 @@ private void pullBaseImages() {
...
               Map<Image, List<Future<PreparedLayer>>> baseImagesAndLayers = new HashMap<>();
               for (Image image : results.baseImagesAndRegistryClient.get().images) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
                 List<Future<PreparedLayer>> layers =
                     scheduleCallables(
                         layersRequiredLocally
                             ? ObtainBaseImageLayerStep.makeListForForcedDownload(
                                 buildContext,
-                                childProgressDispatcherFactory,
+                                progressDispatcherFactory,
```
chanseokoh, line 329: ditto

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -371,15 +394,25 @@ private void buildImages() {
...
+              // least, we must ensure that it's not closed prematurely. (Garbage collection doesn't
+              // auto-close it wit the current implementation.)
+              ProgressEventDispatcher progressDispatcher =
+                  childProgressDispatcherFactory.create(
+                      "scheduling building manifests", results.baseImagesAndLayers.get().size());
+
               Map<Future<Image>, Image> builtImagesAndBaseImages = new HashMap<>();
               for (Map.Entry<Image, List<Future<PreparedLayer>>> entry :
                   results.baseImagesAndLayers.get().entrySet()) {
+                Factory progressDispatcherFactory = progressDispatcher.newChildProducer();
```
chanseokoh, line 409: `ProgressEventDispatcher.Factory progress... = ...`

Error Prone usually complains if a subclass name is too generic.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa: I see so its only in the `executorSubmiter`

PR #2700, opened 2020-08-13
Layers creator classes

loosebazooka:
There are still some quirks to layer creation that we will have
to work out
- includes on file patterns can end up "excluding" directories
  which will have to be readded, or we can ignore includes
  directives on directories and end up with empty directories
  everywhere
- directories between a target and root will be configured to
  have the default properties of ReproducibleLayerBuilder rather
  that what is configured by the user. This can be dealt with
  by populating directories from root.
- only handles `FileEntriesLayers` types

part of #2570

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
@@ -106,14 +106,9 @@ class Layers {
                         if (includes.isEmpty()) {
                           return true;
                         }
-                        // TODO: for directories that fail to match the "include" directive on, if a
-                        // file gets
-                        // TODO: included, ensure parents are populated correctly
                         // TODO: if <dest>/path/to/file.txt is included because of a pattern like
-                        // **/file.txt
-                        // TODO: ensure we create <dest>/path and <dest>/path/to with the correct
-                        // directory
-                        // TODO: properties here
+                        // TODO: **/file.txt, ensure we create <dest>/path and <dest>/path/to with
+                        // TODO: the correct directory properties here
                         // if there are includes directives, only include those specified
                         for (PathMatcher matcher : includes) {
                           if (matcher.matches(path)) {
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import com.google.cloud.tools.jib.api.buildplan.AbsoluteUnixPath;
+import com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer;
+import com.google.cloud.tools.jib.api.buildplan.FileEntry;
+import com.google.cloud.tools.jib.api.buildplan.LayerObject;
+import com.google.common.base.Predicate;
+import java.io.IOException;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/** Class to between different layer representations. */
+class Layers {
+
+  /**
+   * Convert a layer spec to a list of layer objects.
+   *
+   * <p>Does not handle missing directories for files added via this method. We can either prefill
+   * directories here, or allow passing of the file entry information directly to the reproducible
+   * layer builder
+   *
+   * @param buildRoot the directory to resolve relative paths, usually the directory where the build
+   *     config file is located
+   * @param layersSpec a layersSpec containing configuration for all layers
+   * @return a {@link List} of {@link LayerObject} to use as part of a buildplan
+   * @throws IOException if traversing a directory fails
+   */
+  static List<LayerObject> toLayers(Path buildRoot, LayersSpec layersSpec) throws IOException {
+    List<LayerObject> layers = new ArrayList<>();
+
+    FilePropertiesStack filePropertiesStack = new FilePropertiesStack();
+    // base properties
+    layersSpec.getProperties().ifPresent(filePropertiesStack::push);
+
+    for (LayerSpec entry : layersSpec.getEntries()) {
+      // each loop is a new layer
+      if (entry instanceof FileLayerSpec) {
+        FileEntriesLayer.Builder layerBuiler = FileEntriesLayer.builder();
+
+        FileLayerSpec fileLayer = (FileLayerSpec) entry;
+        layerBuiler.setName(fileLayer.getName());
+        // layer properties
+        fileLayer.getProperties().ifPresent(filePropertiesStack::push);
+        for (CopySpec copySpec : ((FileLayerSpec) entry).getFiles()) {
+          // copy spec properties
+          copySpec.getProperties().ifPresent(filePropertiesStack::push);
+
+          // relativize all paths to the buildRoot location
+          Path rawSrc = copySpec.getSrc();
+          Path src = rawSrc.isAbsolute() ? rawSrc : buildRoot.resolve(rawSrc);
+          AbsoluteUnixPath dest = copySpec.getDest();
+
+          if (Files.isRegularFile(src)) { // regular file
+            if (copySpec.getExcludes().size() > 0 || copySpec.getIncludes().size() > 0) {
+              throw new UnsupportedOperationException(
+                  "Cannot apply includes/excludes on single file copy directives.");
+            }
+            AbsoluteUnixPath target = dest;
+            if (copySpec.isDestEndsWithSlash()) {
+              target = copySpec.getDest().resolve(src.getFileName());
+            }
+            layerBuiler.addEntry(
+                src,
+                target,
+                filePropertiesStack.getFilePermissions(),
+                filePropertiesStack.getModificationTime(),
+                filePropertiesStack.getOwnership());
+          } else if (Files.isDirectory(src)) { // directory
+            List<PathMatcher> excludes =
+                copySpec
+                    .getExcludes()
+                    .stream()
+                    .map(
+                        pattern ->
+                            pattern.endsWith("/") || pattern.endsWith("\\")
+                                ? pattern + "**"
+                                : pattern)
+                    .map(exclude -> FileSystems.getDefault().getPathMatcher("glob:" + exclude))
+                    .collect(Collectors.toList());
+            List<PathMatcher> includes =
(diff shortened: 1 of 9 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,179 @@
...
+                              if (matcher.matches(path)) {
+                                return false;
+                              }
+                            }
+                            // if there are no includes directives, include everything
+                            if (includes.isEmpty()) {
+                              return true;
+                            }
+                            // TODO: for directories that fail to match the "include" directive on
+                            // TODO: files, populate the directories somehow or just never apply
```
mpeddada1, line 129: could you elaborate more on this or provide an example?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,179 @@
...
+          AbsoluteUnixPath dest = copySpec.getDest();
+
+          if (Files.isRegularFile(src)) { // regular file
+            if (copySpec.getExcludes().size() > 0 || copySpec.getIncludes().size() > 0) {
+              throw new UnsupportedOperationException(
+                  "Cannot apply includes/excludes on single file copy directives.");
+            }
+            AbsoluteUnixPath target = dest;
+            if (copySpec.isDestEndsWithSlash()) {
+              target = copySpec.getDest().resolve(src.getFileName());
```
chanseokoh, line 82: Just do `dest = dest.resolve(src.getFileName())` and use `dest` below?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,179 @@
...
+          // copy spec properties
+          copySpec.getProperties().ifPresent(filePropertiesStack::push);
+
+          // relativize all paths to the buildRoot location
+          Path rawSrc = copySpec.getSrc();
+          Path src = rawSrc.isAbsolute() ? rawSrc : buildRoot.resolve(rawSrc);
+          AbsoluteUnixPath dest = copySpec.getDest();
+
+          if (Files.isRegularFile(src)) { // regular file
+            if (copySpec.getExcludes().size() > 0 || copySpec.getIncludes().size() > 0) {
```
chanseokoh, line 76: `! ... isEmpty()`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,179 @@
...
+                copySpec
+                    .getExcludes()
+                    .stream()
+                    .map(
+                        pattern ->
+                            pattern.endsWith("/") || pattern.endsWith("\\")
+                                ? pattern + "**"
+                                : pattern)
+                    .map(exclude -> FileSystems.getDefault().getPathMatcher("glob:" + exclude))
+                    .collect(Collectors.toList());
```
chanseokoh, line 101: Maybe this part can be factored out as `toPathMatchers(List<String> globPatterns)` (or `globPatternsToPathMatchers()`) and reused?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,179 @@
...
+                  .forEach(layerBuiler::addEntry);
+            }
+          } else { // other
+            throw new UnsupportedOperationException(
+                "Cannot create FileLayers from non-file, non-directory: " + src.toString());
+          }
+          copySpec.getProperties().ifPresent(ignored -> filePropertiesStack.pop());
+        }
+        fileLayer.getProperties().ifPresent(ignored -> filePropertiesStack.pop());
+        layers.add(layerBuiler.build());
```
chanseokoh, line 171: Worth thinking about allowing to add an empty layer. But I think Jib will simply drop empty layers anyway, if I'm not mistaken.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
loosebazooka: I would like to limit how much I change dest, rather not it be a changed var. (I should use final, but not google style)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
loosebazooka: yeah, also wondering if there's a way to combine path matchers like spring has for https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/PathMatcher.html#combine-java.lang.String-java.lang.String-

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
loosebazooka: I think I'll just use a `?:` for this?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,179 @@
...
+                              dest.resolve(relative),
+                              filePropertiesStack.getFilePermissions(),
+                              filePropertiesStack.getModificationTime(),
+                              filePropertiesStack.getOwnership());
+                        } else {
+                          throw new UnsupportedOperationException(
+                              "Cannot create FileLayers from non-file, non-directory: "
+                                  + path.toString());
+                        }
+                      })
```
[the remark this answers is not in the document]
loosebazooka: I cleaned it up a little.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,179 @@
...
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/** Class to between different layer representations. */
```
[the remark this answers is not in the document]
loosebazooka: lol, yes oops

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
loosebazooka: hrmm.. yeah lemme go update this comment

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
loosebazooka: added todo, and bug in tracker bug

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/LayersTest.java
```diff
@@ -0,0 +1,263 @@
...
+              });
+      System.out.println("EXCPECTED");
+      expectedLayerEntries.forEach(
+          entry -> {
+            System.out.println("src: " + entry.getSourceFile());
+            System.out.println("dest: " + entry.getExtractionPath());
+            System.out.println("permission: " + entry.getPermissions().toOctalString());
+            System.out.println("time: " + entry.getModificationTime());
+            System.out.println("ownership: " + entry.getOwnership());
+          });
```
mpeddada1, line 234: Does calling assertEquals not log the actual vs expected value? Maybe Assert.assertThat(ImmutableSet.copyOf(layer.getEntries)).containsExactly(expectedEntry1, expectedEntry2) could help.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -0,0 +1,163 @@
...
+   * directories here, or allow passing of the file entry information directly to the reproducible
+   * layer builder
+   *
+   * @param buildRoot the directory to resolve relative paths, usually the directory where the build
+   *     config file is located
+   * @param layersSpec a layersSpec containing configuration for all layers
+   * @return a {@link List} of {@link LayerObject} to use as part of a buildplan
+   * @throws IOException if traversing a directory fails
+   */
+  static List<LayerObject> toLayers(Path buildRoot, LayersSpec layersSpec) throws IOException {
```
mpeddada1, line 50: For my understanding, this method will be expanded to support other layers in addition to FileLayer in the future?

loosebazooka: Correct, but I'm not sure what needs to be updated here?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/LayersTest.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.Set;
+import org.hamcrest.CoreMatchers;
+import org.hamcrest.MatcherAssert;
+import org.junit.Assert;
+import org.junit.Test;
+
+/** Tests for {@link Layers}. */
+public class LayersTest {
+
+  public static final String LAYERS_TEST_RESOURCE_DIR = "buildfiles/layers/";
```
chanseokoh, line 46: private?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/LayersTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+  public List<FileEntriesLayer> parseLayers(Path testDir, int expectedLayerCount)
```
chanseokoh, line 48: private static?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/LayersTest.java
```diff
@@ -0,0 +1,263 @@
...
+    Assert.assertEquals(expectedLayerCount, layers.size());
+    List<FileEntriesLayer> fileEntriesLayers = new ArrayList<>(expectedLayerCount);
+    for (LayerObject layerObject : layers) {
+      MatcherAssert.assertThat(layerObject, CoreMatchers.instanceOf(FileEntriesLayer.class));
+      fileEntriesLayers.add((FileEntriesLayer) layerObject);
+    }
+    return fileEntriesLayers;
+  }
+
+  private Path getLayersTestRoot(String testName) throws URISyntaxException {
```
chanseokoh, line 67: static?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/LayersTest.java
mpeddada1: Right. Yeah that's a valid point. It's worth adding (to avoid catching and throwing the same exception) but it can also be done in a different PR since this is one already pretty big:)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
mpeddada1: Oh interesting, I didn't know empty layers were possible

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
mpeddada1: Nice, might be worth adding a line to the PR summary or a TODO at beginning of the method that mentions that the change is for FileLayer.

loosebazooka: Added to PR description

PR #2707, opened 2020-08-18
Prepare JSON template classes for new manifest cache design

chanseokoh:
1. Renaming `ManifestAndConfig` to `ManifestAndConfigTemplate`.
   - The class is a simple pair of manifest and config. In the new design, it will be used as a JSON template for de-/serialization.
   - As a JSON template, converted `Optional<>` to `@Nullable`.

1. New class: `ImageMetadataTemplate`.
   - Holds
      1. a manifest list (if not Docker schema 1). Can be null (when a base image reference is not a manifest list).
      2. a list of manifest+config pairs (`List<ManifestAndConfigTemplate>`).
   For example,
   ```
   {
     "manifestList": {  # manifest list content
       "@class": "com.google.cloud.tools.jib.image.json.V22ManifestListTemplate",
       "manifests": [ { ... }, { ... }, { ... }, ... ]
     },

     "manifestsAndConfigs": [
       {
         "manifest": {
           "@class": "com.google.cloud.tools.jib.image.json.V22ManifestTemplate",
           "schemaVersion": 2,
           "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
           "config": { ... }
           ...
         },
         "config": { ... }
       },
       { ... },
       ...
     ]
   ```

1. `OciImageIndex` now inherits `ManifestTemplate` (previously `JsonTemplate`). This is to have a slightly more specialized class as a common super class of `OciImageIndex` and `V22ManifestListTemplate`.
   - Note currently `ManfiestTemplate` is also a super class of `OciManifestTemplate`, `V21ManifestTemplate` and `V22ManifestTemplate`. It does make sense to consider all of these manifest list classes and manifest classes as `ManifestTemplate`.

@louismurerwa

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
@@ -23,7 +23,7 @@ import com.google.cloud.tools.jib.api.ImageReference;
 import com.google.cloud.tools.jib.blob.Blobs;
 import com.google.cloud.tools.jib.filesystem.LockFile;
 import com.google.cloud.tools.jib.image.json.ContainerConfigurationTemplate;
-import com.google.cloud.tools.jib.image.json.ManifestAndConfig;
+import com.google.cloud.tools.jib.image.json.ManifestAndConfigTemplate;
 import com.google.cloud.tools.jib.image.json.ManifestTemplate;
 import com.google.cloud.tools.jib.image.json.OciManifestTemplate;
 import com.google.cloud.tools.jib.image.json.V21ManifestTemplate;
@@ -84,7 +84,7 @@ class CacheStorageReader {
    * @throws IOException if an I/O exception occurs
    * @throws CacheCorruptedException if the cache is corrupted
    */
-  Optional<ManifestAndConfig> retrieveMetadata(ImageReference imageReference)
+  Optional<ManifestAndConfigTemplate> retrieveMetadata(ImageReference imageReference)
       throws IOException, CacheCorruptedException {
     Path imageDirectory = cacheStorageFiles.getImageDirectory(imageReference);
     Path manifestPath = imageDirectory.resolve("manifest.json");
@@ -110,7 +110,7 @@ class CacheStorageReader {
 
       if (schemaVersion == 1) {
         return Optional.of(
-            new ManifestAndConfig(
+            new ManifestAndConfigTemplate(
                 JsonTemplateMapper.readJsonFromFile(manifestPath, V21ManifestTemplate.class),
                 null));
       }
@@ -139,7 +139,7 @@ class CacheStorageReader {
         ContainerConfigurationTemplate config =
             JsonTemplateMapper.readJsonFromFile(configPath, ContainerConfigurationTemplate.class);
 
-        return Optional.of(new ManifestAndConfig(manifestTemplate, config));
+        return Optional.of(new ManifestAndConfigTemplate(manifestTemplate, config));
       }
       throw new CacheCorruptedException(
           cacheStorageFiles.getCacheDirectory(),
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ManifestAndConfigTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ManifestAndConfigTemplate.java
@@ -16,16 +16,32 @@
 
 package com.google.cloud.tools.jib.image.json;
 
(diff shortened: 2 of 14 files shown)


chanseokoh:
> 3. `OciImageIndex` now inherits `ManifestTemplate` (previously `JsonTemplate`).

Note `ManifestTemplate` is not much significantly different from `JsonTemplate`.
```java
/** Parent class for image manifest JSON templates. */
public interface ManifestTemplate extends JsonTemplate {
  int getSchemaVersion();
}
```
Also the only place where `OciImageIndex` is used is `ImageTarball.ociWriteTo()` to write a local OCI-format image tarball.
https://github.com/GoogleContainerTools/jib/blob/3ba06f929cb1e6a77603d13f220c5d77d869aa78/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageTarball.java#L113-L118

loosebazooka:
I wonder if we're using explicit type information, then it makes sense to only use a complex json cache entry for manifest list/oci index and simpler entries for single images?

chanseokoh:
Ah, that's a very natural idea. There's one thing related though. For the cache retrieval method `Cache.retrieveMetadata()`, I'm making it return the entire metadata info (both a manifest list and a manifest+config list) with `ImageMetadataTemplate`:
```java
  public Optional<ImageMetadataTemplate> retrieveMetadata(ImageReference imageReference)
```
(Did not want to create another class for a return type that is fundamentally same as `ImageMetadataTemplate`).

So `retrieveMetadata()` needs to either 1) reassemble this complex metadata structure again; or 2) return a nominal common super class of a manifest list+manifests structure and a single manifest structure. For 2), I guess the caller would then need to check the actual sub-type with `instanceof` and cast it to either a manifest list+manifests type or just a single manifest type, which may not a bad thing.

Initially I thought returning `ImageMetadataTemplate` is reasonable and simple in implementation. There are certainly some disadvantages though; some assumptions that cannot be enforced with strong typing need to be made. For example, an absence of `ImageMetadataTemplate.manifestList` means `ImageMetadataTemplate.manifestsAndConfigs` should be a single-element list. I'm not so decisive of what's the best way or whether there is a better solution. But at least it seems like having a single `ImageMetadataTemplate` type is simpler in implementation and resulting in creating a less number of JSON-related types that are only required to exist due to the cache.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
```diff
@@ -110,7 +110,7 @@
 
       if (schemaVersion == 1) {
         return Optional.of(
-            new ManifestAndConfig(
+            new ManifestAndConfigTemplate(
```
loosebazooka, line 113: It seems strange that we would generate a JsonTemplate type from separately parsed out jsonTemplates? Is this something that we would need to change eventually? 

Can we go directly to ManifestAndConfigTemplate directly from file?

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ManifestAndConfigTemplate.java
```diff
@@ -16,16 +16,32 @@
...
 
-  private final ManifestTemplate manifest;
-  @Nullable private final ContainerConfigurationTemplate config;
+  @JsonTypeInfo(
+      use = JsonTypeInfo.Id.CLASS,
+      include = JsonTypeInfo.As.PROPERTY,
+      property = "@class")
+  @JsonSubTypes({
+    @JsonSubTypes.Type(value = OciManifestTemplate.class),
+    @JsonSubTypes.Type(value = V21ManifestTemplate.class),
```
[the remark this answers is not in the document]
chanseokoh: We don't actually do a JSON-to-JSON conversion. We are currently [persisting `V21ManifestTemplate`](https://github.com/GoogleContainerTools/jib/blob/79d96f32dfa88a97896478d4afb6333976f3fef0/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L303-L304) in the cache.

It's just that we can properly instantiate a Java `Image` instance from both V21 and V22.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
chanseokoh: In the next PR, all of these parsing and conversion operations will be gone, and this `retrieveMetadata()` will directly de-serialize and return `ImageMetadataTemplate` without any processing.

Issue #2702, opened 2020-08-14
NPE during Preconditions validation   


vgalloy:
**Environment**:

- *Jib version:*  2.5.0
- *Build tool:* gradle 6.6
- *OS:* Ubuntu 20.04.1 LTS
- *Java:* openjdk version "14" 2020-03-17  

**Description of the issue**:
`ImmutableCollections#contains(Object)` throw a `NullPointerException` if the argument is `null`. Checking if collection contains `null` will always throw a `NullPointerException`
Example 
`Arrays.asList("java").contains(null)` => `false`
`List.of("java").contains(null)` => `NullPointerException`

There is several place where the following pattern is apply :
`Preconditions.checkArgument(!entrypoint.contains(null), "entrypoint contains null elements");`

Passing an ImmutableCollections as `entrypoint` always fail 

**Expected behavior**:
No `NullPointerException` exception

**Steps to reproduce**:
1. ./gradlew jibDockerBuild

**`jib-gradle-plugin` Configuration**: 
```groovy
plugins {
  id 'java-library'
  id 'com.google.cloud.tools.jib' version '2.5.0'
}

java {
  targetCompatibility = JavaVersion.VERSION_11
  sourceCompatibility = JavaVersion.VERSION_11
}

jib {
  container {
    entrypoint = List.of('java', '-jar', 'my-jar')
  }
}
```

**Log output**: 
```
Caused by: java.lang.NullPointerException
        at com.google.cloud.tools.jib.configuration.ContainerConfiguration$Builder.setEntrypoint(ContainerConfiguration.java:265)
        at com.google.cloud.tools.jib.api.JibContainerBuilder.setEntrypoint(JibContainerBuilder.java:274)
        at com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor.processCommonConfiguration(PluginConfigurationProcessor.java:403)
        at com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor.processCommonConfiguration(PluginConfigurationProcessor.java:455)
        at com.google.cloud.tools.jib.plugins.common.PluginConfigurationProcessor.createJibBuildRunnerForDockerDaemonImage(PluginConfigurationProcessor.java:132)
        at com.google.cloud.tools.jib.gradle.BuildDockerTask.buildDocker(BuildDockerTask.java:110)
```

**Additional Information**: 
I know that not a conventional way to write a `List`, I'm just trying to provided the shorter possible example.

[...]

PR #2708, opened 2020-08-18
Update CHANGELOG for #2702

mpeddada1:


jib-gradle-plugin/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 ### Fixed
+- Fixed `NullPointerException` during preconditions validation (in Java 9+) when checking for presence of `null` value in ImmutableCollection.([#2702](https://github.com/GoogleContainerTools/jib/issues/2702))
```
chanseokoh, line 11: Add a blank line after `### Fixed`.

The message sounds too technical and not so helpful to the end user. The CHANGELOG is for the end user where they want to quickly find relevant bug fixes and versions that matter to them.

```suggestion
- Fixed `NullPointerException` during input validation (in Java 9+) when configuring Jib parameters using certain immutable collections (such as `List.of()`). ([#2702](https://github.com/GoogleContainerTools/jib/issues/2702))
```

jib-maven-plugin/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 
 ### Fixed
+- Fixed `NullPointerException` during preconditions validation (in Java 9+) when checking for presence of `null` value in ImmutableCollection.([#2702](https://github.com/GoogleContainerTools/jib/issues/2702))
```
chanseokoh, line 11: I think we can skip Maven, as I don't think this occurs in Maven.

jib-gradle-plugin/CHANGELOG.md
mpeddada1: This is much better! Thank you

jib-maven-plugin/CHANGELOG.md
mpeddada1: Ah you're right, good catch.

PR #2709, opened 2020-08-19
Check manifest in target registry

louismurerwa:
This PR adds the ability to check if the manifest template exists in the target registry before pushing the image/images to the registry .


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -499,16 +499,11 @@ private void checkImageInTargetRegistry() {
         executorService.submit(
             () -> {
               Verify.verify(results.builtImagesAndBaseImages.get().size() == 1);
```
chanseokoh, line 501: Since we've removed the `iterator().next()` code below and the `CheckImageStep` code can fully work without the assumption that there should be a single image, I think we can remove this verify line.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/CheckImageStep.java
```diff
@@ -44,31 +40,23 @@
   private final BuildContext buildContext;
   private final ProgressEventDispatcher.Factory progressEventDispatcherFactory;
   private final RegistryClient registryClient;
-  private final BlobDescriptor containerConfigurationDigestAndSize;
-  private final Image image;
+  private final ManifestTemplate manifestTemplate;
 
   CheckImageStep(
```
chanseokoh, line 45: At this point, I think we'd better rename this to `CheckManifestStep()`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
louismurerwa: Good Point

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -497,20 +497,13 @@ private void checkImageInTargetRegistry() {
...
-              return new CheckImageStep(
-                      buildContext,
-                      childProgressDispatcherFactory,
-                      results.targetRegistryClient.get(),
-                      Verify.verifyNotNull(containerConfigPushResult).get(),
-                      builtImage.get())
-                  .call();
-            });
+            () ->
+                new CheckManifestStep(
```
chanseokoh, line 501: Let's rename `checkImageInTargetRegistry()` to `checkManifestInTargetRegistry()` on line 494.

PR #2710, opened 2020-08-19
Rename RegistryClient.checkimage() to RegistryClient.checkManifest()

chanseokoh:
The method pulls the manifest referred to by a tag or digest (if exists). The pulled manifest can also be a manifest list instead of an image manifest, so renaming it to `checkManifest()`.


louismurerwa:
LGTM

PR #2711, opened 2020-08-20
Implement new manifest cache design: cache manifest lists and manifest+config pairs

chanseokoh:
(See #2707 about the new JSON data structure (`ImageMetadataTemplate`) for the cache file `manifests_configs.json`.)

The new cache supports storing and retrieving both OCI and Docker manifests and manifest lists.

- If a base image ref is not a manifest list, Jib will result in caching a single-element `ImageMetadataTemplate.manifestsAndConfigs` without caching a manifest list (null `ImageMetadataTemplate.manifestList`).
- If a base image ref is a manifest list, Jib will cache the entire original manifest list. However, for actual manifests, it will cache only a subset that are downloaded and used.

#### Cache API Changes

- Current manifest cache API
   - Returns and writes a single manifest and config pairs (`ManifestAndConfigTemplate`).
   ```java
     public Optional<ManifestAndConfigTemplate> retrieveMetadata(ImageReference imageReference)
   ```
   ```java
     public void writeMetadata(
         ImageReference imageReference,
         BuildableManifestTemplate manifestTemplate,
         ContainerConfigurationTemplate containerConfigurationTemplate)
   ```

- New API
   - Returns and writes a manifest list and a list of manifest and config pairs (`ImageMetadataTemplate`).
   ```java
     public Optional<ImageMetadataTemplate> retrieveMetadata(ImageReference imageReference)
   ```
   ```java
     public void writeMetadata(
         ImageReference imageReference,
         @Nullable ManifestTemplate manifestList,
         List<BuildableManifestTemplate> manifests,
         List<ContainerConfigurationTemplate> containerConfigurations)
   ```

#### Updating Manifest Cache File

The implementation replaces `manifests_configs.json` whenever running Jib online. Jib will store only those manifests and configs Jib downloaded and used in the last build–no accumulation of past manifests and configs.

#### Cache Invalidation in the New Release

The new cache file is `manifests_configs.json`. Existing `manifest.json` and `config.json` will not be used. The user upgrading Jib will lose previous manifest and config caches.


chanseokoh:
Will update CHANGELOG and do some more manual end-to-end tests.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -215,139 +218,133 @@ private boolean checkImagePlatform(Image image) {
...
+
     ManifestAndDigest<?> manifestAndDigest =
-        registryClient.pullManifest(buildContext.getBaseImageConfiguration().getImageQualifier());
+        registryClient.pullManifest(baseImageConfig.getImageQualifier());
     ManifestTemplate manifestTemplate = manifestAndDigest.getManifest();
 
+    if (manifestTemplate instanceof V21ManifestTemplate) {
+      V21ManifestTemplate v21Manifest = (V21ManifestTemplate) manifestTemplate;
+      cache.writeMetadata(baseImageConfig.getImage(), v21Manifest);
+      return Collections.singletonList(JsonToImageTranslator.toImage(v21Manifest));
```
louismurerwa, line 231: Maybe we should check if the `platform` of the image being returned matches the `platform` specified by the user . We can use the `checkImage`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -215,139 +218,133 @@ private boolean checkImagePlatform(Image image) {
...
+                registryClient, (V22ManifestListTemplate) manifestList, platform);
+
+        manifests.add((BuildableManifestTemplate) manifestAndDigest.getManifest());
+        containerConfigs.add(
+            pullContainerConfigJson(manifestAndDigest, registryClient, progressEventDispatcher));
       }
-      return images.build();
+
+    } else {
+      // V22ManifestTemplate or OciManifestTemplate
```
louismurerwa, line 251: What about the `OciManifestListTemplate` ?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: This is one of the TODO items you're tracking. We should handle it properly, but as a separate PR on its own.

However, note that we shouldn't fail a build even if the platform doesn't match.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: That's why I added the comment
```
// TODO: support OciIndexTemplate once AbstractManifestPuller starts to accept it.
```
below. (Note `OciIndexTemplate` is what you call by "OCI manifest list" here.) Supporting an OCI image index is another TODO item that you're tracking.

louismurerwa: Makes sense ,  I wasn't aware of the fact that `OciIndexTemplate` = `Oci manifest list`

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -215,139 +218,133 @@ private boolean checkImagePlatform(Image image) {
...
       throws IOException, RegistryException {
     EventHandlers eventHandlers = buildContext.getEventHandlers();
+    String message = "Searching for architecture=%s, os=%s in the base image manifest list";
     eventHandlers.dispatch(
-        LogEvent.lifecycle(
-            "The base image reference is a manifest list, searching for architecture="
-                + platform.getArchitecture()
-                + ", os="
-                + platform.getOs()));
+        LogEvent.lifecycle(String.format(message, platform.getArchitecture(), platform.getOs())));
```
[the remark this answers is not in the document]
chanseokoh: I had the same question. It has been lifecycle from the start, and I guess the reasoning is to let the user easily see that the base image they use is a manifest list and not a manifest. I think that's helpful. But now that we start supporting configuring platforms, I also think we don't have to be verbose on this. Let's go for INFO.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java
```diff
@@ -296,42 +302,59 @@ CachedLayer writeTarLayer(DescriptorDigest diffId, Blob compressedBlob) throws I
...
-   * @param containerConfiguration the container configuration
+   * @param manifestList the V2.2 manifest list or OCI image index. Can be null.
+   * @param manifests the V2.2 or OCI manifests
+   * @param containerConfigurations the container configurations
    */
   void writeMetadata(
       ImageReference imageReference,
-      BuildableManifestTemplate manifestTemplate,
-      ContainerConfigurationTemplate containerConfiguration)
+      @Nullable ManifestTemplate manifestList,
```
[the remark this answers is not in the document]
chanseokoh: I have been working on changing this to accept `ImageMetadataTemplate` since our last meeting. Still working on it. Because you've already gone through this PR, maybe it's better to avoid further changing this PR a lot. I can do the API refactoring in a follow-up.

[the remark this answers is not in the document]
chanseokoh: I decided to change the API in this PR. It enables refactoring the code in a more readable way, I think.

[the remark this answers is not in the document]
chanseokoh: Ugh, failed to split the PR. Let me clean this up.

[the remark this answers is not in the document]
chanseokoh: Should be good now.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -215,139 +218,135 @@ private boolean checkImagePlatform(Image image) {
...
+      BuildableManifestTemplate imageManifest = (BuildableManifestTemplate) manifestTemplate;
+      ContainerConfigurationTemplate containerConfig =
+          pullContainerConfigJson(manifestAndDigest, registryClient, progressEventDispatcher);
+      cache.writeMetadata(baseImageConfig.getImage(), imageManifest, containerConfig);
+      return Collections.singletonList(
+          JsonToImageTranslator.toImage(imageManifest, containerConfig));
+    }
+
+    // TODO: support OciIndexTemplate once AbstractManifestPuller starts to accept it.
+    Verify.verify(manifestTemplate instanceof V22ManifestListTemplate);
```
loosebazooka, line 247: Does this then throw an error for `OciIndexTemplate`? Perhaps we can give the user a better error message (I thought `Verify` is for catching programming/integration issues?)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -215,139 +218,135 @@ private boolean checkImagePlatform(Image image) {
...
-                  containerConfigurationTemplate);
-
-          return JsonToImageTranslator.toImage(
-              buildableManifestTemplate, containerConfigurationTemplate);
-        }
+      ProgressEventDispatcher progressDispatcher)
+      throws IOException, LayerPropertyNotFoundException, UnknownManifestFormatException {
+    BuildableManifestTemplate manifest =
+        (BuildableManifestTemplate) manifestAndDigest.getManifest();
+    Preconditions.checkArgument(manifest.getSchemaVersion() == 2);
```
loosebazooka, line 329: Similarly here, what happens when schema version is not `2`, does this just kill our build with a confusing error message?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
```diff
@@ -66,32 +69,52 @@ private Cache(CacheStorageFiles cacheStorageFiles) {
...
+    cacheStorageWriter.writeMetadata(imageReference, metadata);
+  }
+
+  /**
+   * Saves a schema 2 manifest for an image reference. This is a simple wrapper around {@link
+   * #writeMetadata(ImageReference, ImageMetadataTemplate)} to save a single manifest without a
+   * manifest list.
+   *
+   * @param imageReference the image reference to save the manifest for
+   * @param manifest the V2.2 or OCI manifest
```
loosebazooka, line 90: I thought we also write 2.1 manifests down to cache?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: Basically the refactored code doesn't change the current behavior in this regard. The current code would simply break with `ClassCastException` if `manifestTemplate` could ever be of `OciIndexTemplate`, as it can't cast `OciInexTemplate` into `BuildableManifestTemplate` in `jsonManifestToImage()`. That is, it's an impossible condition; we never see `OciIndexTemplate`, as the manifest puller doesn't tell a registry that it can accept an OCI image index. All the registries in practice so far have been returning a Docker manifest list.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: Will kill the build. It's an impossible condition the user should never hit. If we see it, it's a bug for us to fix. That is, we should pull a container config only when it's a schema 2 (i.e., when `OciImageTemplate` or `V22ManifestTemplate`), because the V2.1 manifest embeds a container config.

But actually, `BuildableManifestTemplate` is always schema version 2. It's only for helping us locate places to update more easily in the future when a schema version 3 is introduced in the spec.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/Cache.java
chanseokoh: Yes, we have a dedicated API for V2.1 on  line 113 (line 92 in old code).

chanseokoh:
> Will update CHANGELOG and do some more manual end-to-end tests.

I will add CHANGELOG once this is fully implemented.

PR #2714, opened 2020-08-20
Add explicit toString methods to FileEntry and FilePermissions

loosebazooka:


jib-build-plan/src/test/java/com/google/cloud/tools/jib/api/buildplan/FileEntryTest.java
```diff
@@ -0,0 +1,15 @@
+package com.google.cloud.tools.jib.api.buildplan;
```
chanseokoh, line 1: copyright

loosebazooka:
ah, oops, I got a little excited, will cleanup

chanseokoh:
Ah, the test fails on Windows because of `/`-vs-`\`.

PR #2721, opened 2020-08-24
Extending the ManifestTemplate  interface by adding getManifestMediaType

louismurerwa:


jib-core/src/test/resources/core/json/ociindex.json
```diff
@@ -1 +1 @@
-{"schemaVersion":2,"manifests":[{"mediaType":"application/vnd.oci.image.manifest.v1+json","digest":"sha256:8c662931926fa990b41da3c9f42663a537ccd498130030f9149173a0493832ad","size":1000,"annotations":{"org.opencontainers.image.ref.name":"regis.try/repo:tag"}}]}
\ No newline at end of file
+{"schemaVersion":2,"mediaType":"application/vnd.oci.image.index.v1+json","manifests":[{"mediaType":"application/vnd.oci.image.manifest.v1+json","digest":"sha256:8c662931926fa990b41da3c9f42663a537ccd498130030f9149173a0493832ad","size":1000,"annotations":{"org.opencontainers.image.ref.name":"regis.try/repo:tag"}}]}
```
chanseokoh, line 1: Avoid introducing unnecessary chanages.

louismurerwa: This is a necessary change that makes the tests pass - I can explain more in person.

chanseokoh: FTR: `OciIndexTemplateTest.testToJson()` became broken only because it compared the result of some operation with this JSON content. We should be good with this change to the test data.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/OciIndexTemplate.java
```diff
@@ -30,6 +30,7 @@
  * <pre>{@code
  * {
  *   "schemaVersion": 2,
+ *   "mediaType":"application/vnd.oci.image.index.v1+json",
```
chanseokoh, line 33: super nit
```suggestion
 *   "mediaType": "application/vnd.oci.image.index.v1+json",
```

PR #2730, opened 2020-08-26
Implement new manifest cache fully capable of multi-platform image building

chanseokoh:
Follow-up to #2711 to complete the new cache design to support multi-platform image building.

- Adds `UnlistedPlatformInManifestListException`.
- No platform verification of a cached manifest any more. `getCachedBaseImages()` returns cached images only when Jib has cached all the manifests matching the configured platforms.
- When caching a manifest list and manifests, `PullBaseImageStep` also stores additional information of manifest digests. This is to enable looking up a manifest from a manifest list by digest. For example,
```
{
  "manifestList": {  # manifest list content
    "manifests": [
      {  # manifest descriptor 1
        "digest": "sha256:a157906...",
        "platform": { ... },
        ...
      },
      {
        "digest": "sha256:...",
        "platform": { ... },
        ...
      }
    ]
  },

  "manifestsAndConfigs": [
    {
      "manifestDigest": "sha256:a1579064....",  <-- manifest digest for lookup
      "manifest": { ... },
      "config": { ... }
    },
    { ... },
    ...
  ]     
```


chanseokoh:
Ready for review. See the PR description for high-level changes.

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ManifestAndConfigTemplate.java
```diff
@@ -43,13 +43,36 @@
   @SuppressWarnings("unused")
   private ManifestAndConfigTemplate() {}
 
+  /**
+   * Creates an instance.
+   *
+   * @param manifest the manifest
```
loosebazooka, line 49: maybe "the registry manifest"?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -105,28 +106,21 @@ public ImagesAndRegistryClient call()
...
-      Optional<Image> image = getCachedBaseImage();
-      if (image.isPresent()) {
-        if (!checkImagePlatform(image.get())) {
-          throw new IllegalStateException(
-              "The cached base image manifest does not match the configured platform due to the "
-                  + "current implementation of limited platform support. As a workaround, re-run "
-                  + "Jib online once to re-cache the right image manifest.");
-        }
-        return new ImagesAndRegistryClient(Collections.singletonList(image.get()), null);
+      List<Image> images = getCachedBaseImages();
```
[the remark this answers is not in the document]
chanseokoh: > where we don't have all the necessary platform information for an image/platform that wasn't previously referenced?

In this case, the `getCachedBaseImages()` in this PR returns an empty list (i.e., cache miss). Offline building will fail and suggest to re-run online once. The digest reference case will fall through to normal operation (contacting the base image repository).

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/ManifestAndConfigTemplate.java
chanseokoh: I think "the image manifest" is more suitable.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -351,35 +330,67 @@ private ContainerConfigurationTemplate pullContainerConfigJson(
...
+      Verify.verify(manifestsAndConfigs.size() == 1);
+      ManifestTemplate manifest = manifestsAndConfigs.get(0).getManifest();
+      if (manifest instanceof V21ManifestTemplate) {
+        return Collections.singletonList(
+            JsonToImageTranslator.toImage((V21ManifestTemplate) manifest));
+      }
+      return Collections.singletonList(
+          JsonToImageTranslator.toImage(
+              (BuildableManifestTemplate) Verify.verifyNotNull(manifest),
+              Verify.verifyNotNull(manifestsAndConfigs.get(0).getConfig())));
```
mpeddada1, line 370: So this section is for when the base image reference is a single manifest?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -351,35 +330,67 @@ private ContainerConfigurationTemplate pullContainerConfigJson(
   }
 
   /**
-   * Retrieves the cached base image.
+   * Retrieves the cached base images. If a base image reference is not a manifest list, returns a
+   * single image (if cached). If a manifest list, returns all the images matching the configured
+   * platforms in the manifest list but only when all of the images are cached.
```
mpeddada1, line 335: Why do we make sure to return the images matching the configured platforms in the manifest list only if *all* the images are cached?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -351,35 +330,67 @@ private ContainerConfigurationTemplate pullContainerConfigJson(
...
+              .stream()
+              .filter(entry -> manifestDigest.equals(entry.getManifestDigest()))
+              .findFirst();
+      if (!manifestAndConfigFound.isPresent()) {
+        return Collections.emptyList();
+      }
+
+      ManifestTemplate manifest = Verify.verifyNotNull(manifestAndConfigFound.get().getManifest());
+      ContainerConfigurationTemplate containerConfig =
+          Verify.verifyNotNull(manifestAndConfigFound.get().getConfig());
```
mpeddada1, line 390: I may be missing something here but what's the reasoning behind checking if these values are null?

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: If we have cached only a subset of what we really need (i.e., Jib needs build, says, three `Image`s, but the cache has only two `Image`s), then it's not a cache hit as a whole, thus a cache miss.

Theoretically you could say, Jib could re-download only the missing `Image`s, but note that performance is not the reason for caching. It's about whether we can completely avoid accessing the Internet or not. So, if any image is missing in the cache, I am saying "cache miss."

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: Exactly. When the image reference is an image manifest and not a manifest list. In the cache metadata, `manifestList` is null in this case.

mpeddada1: Ah gotcha, is there a particular reason why we chose `null` instead of an empty list for manifestList?

chanseokoh: Good question. A "manifest list" is not a `List` in Java but a class called `ManifestTemplate`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: It's our assumption and how we designed the cache to work. If any `ManifestAndConfig` entry (which is just a pair of a manifest and a config) is saved (cached), their values should never be null; otherwise, something got terribly wrong.

Now, we then use `Verify.verifyNotNull()` to signal the NullAway check (it's a check we set up in our repo to catch potential NPE) that `manifest` and `containerConfig` are not supposed to be null. Otherwise, because `JsonToImageTranslator.toImage()` doesn't except `@Nullable` parameters, NullAway will complain that you're passing potentially null parameters.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
mpeddada1: Right my bad, thanks for clarifying:)

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
mpeddada1: Oh I see! Makes sense

PR #2735, opened 2020-08-28
Read a buildfile and convert it into a JibContainerBuilder

loosebazooka:
Reads a buildfile from file and converts it into a JibContainerBuilder, this will be used by the CLI to then generate a container.

Requires #2763

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
@@ -43,9 +43,10 @@ public class BuildFiles {
    */
   public static JibContainerBuilder toJibContainerBuilder(Path buildFilePath)
       throws InvalidImageReferenceException, IOException {
-    ObjectMapper om = new ObjectMapper(new YAMLFactory());
+    ObjectMapper yamlObjectMapper = new ObjectMapper(new YAMLFactory());
     BuildFileSpec buildFile =
-        om.readValue(Files.newBufferedReader(buildFilePath, Charsets.UTF_8), BuildFileSpec.class);
+        yamlObjectMapper.readValue(
+            Files.newBufferedReader(buildFilePath, Charsets.UTF_8), BuildFileSpec.class);
     Path projectRoot = buildFilePath.toAbsolutePath().getParent();
 
     JibContainerBuilder containerBuilder;
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.cli.buildfile;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.google.cloud.tools.jib.api.InvalidImageReferenceException;
+import com.google.cloud.tools.jib.api.Jib;
+import com.google.cloud.tools.jib.api.JibContainerBuilder;
+import com.google.cloud.tools.jib.api.buildplan.Platform;
+import com.google.common.base.Charsets;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.stream.Collectors;
+
+/** Class to convert BuildFiles to build container representations. */
+public class BuildFiles {
+
+  /**
+   * Read a buildfile from disk and generate a JibContainerBuilder instance. All parsing of files
+   * considers the directory the buildfile is located in as the working directory.
+   *
+   * @param buildFilePath a file containing the build definition
+   * @return a {@link JibContainerBuilder} generated from the contents of {@code buildFilePath}
+   * @throws IOException if an I/O error occurs opening the file, or an error occurs while
+   *     traversing files on the filesystem
+   * @throws InvalidImageReferenceException if the baseImage reference can not be parsed
+   */
+  public static JibContainerBuilder toJibContainerBuilder(Path buildFilePath)
+      throws InvalidImageReferenceException, IOException {
+    ObjectMapper om = new ObjectMapper(new YAMLFactory());
+    BuildFileSpec buildFile =
+        om.readValue(Files.newBufferedReader(buildFilePath, Charsets.UTF_8), BuildFileSpec.class);
+    Path projectRoot = buildFilePath.toAbsolutePath().getParent();
+
+    JibContainerBuilder containerBuilder;
+    if (buildFile.getFrom().isPresent()) {
+      BaseImageSpec from = buildFile.getFrom().get();
+      containerBuilder = Jib.from(buildFile.getFrom().get().getImage());
+      if (!from.getPlatforms().isEmpty()) {
+        containerBuilder.setPlatforms(
+            from.getPlatforms()
+                .stream()
+                .map(
+                    platformSpec ->
+                        new Platform(platformSpec.getArchitecture(), platformSpec.getOs()))
+                .collect(Collectors.toSet()));
+      }
+    } else {
+      containerBuilder = Jib.fromScratch();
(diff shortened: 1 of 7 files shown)


jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class BuildFilesTest {
+
+  @Rule public TemporaryFolder tmp = new TemporaryFolder();
+
+  @Test
+  public void testToJibContainerBuilder_allProperties()
```
mpeddada1, line 48: have we considered using Truth (https://truth.dev/) for our assertions? It could offer clearer error messaging and help in understanding the actual vs expected values --> `assertThat(actual).isEqualTo(expected)` (for example, the _actual_ is always on the left-hand side in Truth so we don't need to remember the correct parameter ordering)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -47,8 +47,8 @@
    * @return a {@link List} of {@link LayerObject} to use as part of a buildplan
    * @throws IOException if traversing a directory fails
    */
-  static List<LayerObject> toLayers(Path buildRoot, LayersSpec layersSpec) throws IOException {
-    List<LayerObject> layers = new ArrayList<>();
+  static List<FileEntriesLayer> toLayers(Path buildRoot, LayersSpec layersSpec) throws IOException {
```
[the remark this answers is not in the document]
loosebazooka: It can be, it doesn't really need to be exposed outside of this package though. Since this isn't a library it doesn't matter all that much. But it really should only be used for conversion within the package.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -0,0 +1,85 @@
...
+  public static JibContainerBuilder toJibContainerBuilder(Path buildFilePath)
+      throws InvalidImageReferenceException, IOException {
+    ObjectMapper om = new ObjectMapper(new YAMLFactory());
+    BuildFileSpec buildFile =
+        om.readValue(Files.newBufferedReader(buildFilePath, Charsets.UTF_8), BuildFileSpec.class);
+    Path projectRoot = buildFilePath.toAbsolutePath().getParent();
+
+    JibContainerBuilder containerBuilder;
+    if (buildFile.getFrom().isPresent()) {
+      BaseImageSpec from = buildFile.getFrom().get();
```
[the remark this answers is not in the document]
loosebazooka: I would say in this context `from` is a better description of that this variable represents.

jib-cli/src/test/resources/buildfiles/projects/allProperties/jib.yaml
```diff
@@ -0,0 +1,58 @@
...
+kind: BuildFile
+
+# "FROM" with detail for manifest lists or multiple architectures
+from:
+  image: "ubuntu"
+  # optional: if missing, then defaults to `linux/amd64`
+  platforms:
+    - architecture: "arm"
+      os: "linux"
+      os.version: "a"
```
loosebazooka, line 12: I realized, these values "os.version", "os.features", "variant", "features" are all ignored in the current implementation, because we don't support it at lower levels of jib.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -0,0 +1,86 @@
...
+   * @throws IOException if an I/O error occurs opening the file, or an error occurs while
+   *     traversing files on the filesystem
+   * @throws InvalidImageReferenceException if the baseImage reference can not be parsed
+   */
+  public static JibContainerBuilder toJibContainerBuilder(Path buildFilePath)
+      throws InvalidImageReferenceException, IOException {
+    ObjectMapper yamlObjectMapper = new ObjectMapper(new YAMLFactory());
+    BuildFileSpec buildFile =
+        yamlObjectMapper.readValue(
+            Files.newBufferedReader(buildFilePath, Charsets.UTF_8), BuildFileSpec.class);
```
chanseokoh, line 49: Close the reader with a try-resource block?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -0,0 +1,86 @@
...
+    Path projectRoot = buildFilePath.toAbsolutePath().getParent();
+
+    JibContainerBuilder containerBuilder;
+    if (buildFile.getFrom().isPresent()) {
+      BaseImageSpec from = buildFile.getFrom().get();
+      containerBuilder = Jib.from(buildFile.getFrom().get().getImage());
```
chanseokoh, line 55: Perhaps
```
      containerBuilder = Jib.from(from.getImage());
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.nio.file.Paths;
+import java.time.Instant;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class BuildFilesTest {
+
+  @Rule public TemporaryFolder tmp = new TemporaryFolder();
```
chanseokoh, line 45: final

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
```diff
@@ -0,0 +1,109 @@
...
+    Assert.assertEquals(AbsoluteUnixPath.get("/home"), resolved.getWorkingDirectory());
+    Assert.assertEquals(ImmutableList.of("sh", "script.sh"), resolved.getEntrypoint());
+    Assert.assertEquals(ImmutableList.of("--param", "param"), resolved.getCmd());
+
+    Assert.assertEquals(1, resolved.getLayers().size());
+    FileEntriesLayer resolvedLayer = (FileEntriesLayer) resolved.getLayers().get(0);
+    Assert.assertEquals("scripts", resolvedLayer.getName());
+    Assert.assertEquals(
+        FileEntriesLayer.builder()
+            .setName("scripts")
```
chanseokoh, line 79: Probably unnecessary if we are only getting entries via `getEntries()`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
loosebazooka: oh right, good call.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
loosebazooka: oh right, must've been leftover from when I was trying to compare FileEntriesLayer objects directly

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
loosebazooka: No we have not, but this seems like a good idea. We can add this in a new PR and then start writing new tests using Truth. We can also update old tests slowly to use it.

jib-cli/src/test/resources/buildfiles/projects/allProperties/jib.yaml
loosebazooka: this is fixed

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
```diff
@@ -0,0 +1,108 @@
...
+            .addEntry(
+                projectRoot.resolve("project/script.sh"), AbsoluteUnixPath.get("/home/script.sh"))
+            .build()
+            .getEntries(),
+        resolvedLayer.getEntries());
+    Assert.assertEquals(LayerObject.Type.FILE_ENTRIES, resolvedLayer.getType());
+  }
+
+  @Test
+  public void testToJibContainerBuilder_requiredProperties()
```
mpeddada1, line 88: quick question on this: Is this test only for when a scratch base image is used? And is that why we allow some of these required parameters to be empty or null?

loosebazooka: This test adds in only the required properties and then verifies that the output is the "defaults". When the defaults change, this will fail and we will know that we either have to revert the change or update our documentation to correctly reflect what the defaults are.

mpeddada1: Cool, thanks!

mpeddada1 (approved):

PR #2736, opened 2020-08-31
Clean up and simplify code

chanseokoh:
Mostly simplifying tests.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
@@ -17,6 +17,7 @@
 package com.google.cloud.tools.jib.registry;
 
 import com.google.cloud.tools.jib.Command;
+import com.google.common.collect.Lists;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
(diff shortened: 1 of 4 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -65,18 +66,14 @@ protected void after() {
   public void start() throws IOException, InterruptedException {
     // Runs the Docker registry.
     ArrayList<String> dockerTokens =
-        new ArrayList<>(
-            Arrays.asList(
-                "docker", "run", "--rm", "-d", "-p", port + ":5000", "--name", containerName));
+        Lists.newArrayList(
```
loosebazooka, line 69: Maybe also `ArrayList<String> dockerTokens` -> `List<String> ...` ?

jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -52,13 +52,9 @@
 public class BuildImageMojoIntegrationTest {
 
   @ClassRule
-  public static final LocalRegistry localRegistry1 =
+  public static final LocalRegistry localRegistry =
```
loosebazooka, line 55: It seems like if we only use one registry, then `buildAndRunComplex` doesn't need to accept registry info as parameters?

chanseokoh: Ohhh, that's right. Also no need to accept username and password info.

PR #2739, opened 2020-09-03
Use Gradle Property and Provider to enable lazy evaluation for jib.to.image and jib.to.tags

### LINKED ISSUE #2727: Convert Gradle config parameters (such as jib.to.image and jib.to.tags) to use Property type
I tried using `gradle-git-properties` Gradle plugin to set a git commit as an image tag. Following their [plugin doc](https://github.com/n0mer/gradle-git-properties),
```gradle
plugins {
    id 'com.gorylenko.gradle-git-properties' version '2.2.3'
    id 'com.google.cloud.tools.jib' version '2.5.0'
    id 'java'
}

gitProperties {
      extProperty = 'gitProps' // git properties will be put in a map at project.ext.gitProps
}
// make sure the generateGitProperties task always executes (even when git.properties is not changed)
generateGitProperties.outputs.upToDateWhen { false }

// make sure generateGitProperties task to execute before accessing generated properties
task printGitProperties(dependsOn: 'generateGitProperties') {
    doLast {
        println "git.commit.id.abbrev=" + project.ext.gitProps['git.commit.id.abbrev']
    }
}
```
Running `./gradlew printGitProperties` works as documented.

The doc also shows an example using Groovy GString lazy evaluation.
```gradle
jar {
  manifest {
    attributes(
      // Use GString lazy evaluation to delay until git properties are populated
      'Build-Revision': "${-> project.ext.gitProps['git.commit.id.abbrev']}"

      // OTOH, the following code fails with "Cannot get property 'gitProps' on extra properties extension as it does not exist"
      //'Build-Revision': project.ext.gitProps['git.commit.id.abbrev']
    )
  }
}
```
I confirm `./gradlew jar` works and the `MANIFEST.MF` in the JAR contains the `Build-Revision` attribute.

However, we still can't use GString lazy evaluation with most of the Jib properties. It fails with "Cannot get property 'gitProps' on extra properties extension as it does not exist."

I think it makes sense to convert `jib.to.image` and `jib.to.tags` to `Property<String>` and `SetProperty<String>`. `Property<String>` has a two `.set()` methods, one with signature `.set(String)` and the other with `.set(Provider<String>)`, which I think will allow configuring these values lazily.

mpeddada1:
Fixes #2727

**Usage:**
- If user wants to use lazy evaluation for `jib.to.image` and `jib.to.tags` with String `value` they can do the following:
`jib.to.image = project.provider{value}` or `jib.to.tags = project.provider{[value, "tag2"]}`


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -33,28 +36,27 @@
...
   }
 
   @Input
-  @Nullable
-  @Optional
-  public String getImage() {
-    if (System.getProperty(PropertyNames.TO_IMAGE) != null) {
-      return System.getProperty(PropertyNames.TO_IMAGE);
-    }
+  public Provider<String> getImage() {
```
chanseokoh, line 54: I think we can keep the current return type of `String`. Is there a specific reason for returning `Provider<String>`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -33,28 +36,27 @@
...
+  private final Project project;
 
-  @Nullable private String image;
+  private Property<String> image;
   private Set<String> tags = Collections.emptySet();
   @Nullable private String credHelper;
 
   @Inject
-  public TargetImageParameters(ObjectFactory objectFactory) {
+  public TargetImageParameters(ObjectFactory objectFactory, Project project) {
```
chanseokoh, line 46: No need to accept `objectFactory`, as you can get it by `project.getObjects()`, which is what you are already doing in the new code.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -33,28 +36,27 @@
...
-  @Optional
-  public String getImage() {
-    if (System.getProperty(PropertyNames.TO_IMAGE) != null) {
-      return System.getProperty(PropertyNames.TO_IMAGE);
-    }
+  public Provider<String> getImage() {
     return image;
   }
 
   public void setImage(String image) {
```
chanseokoh, line 58: I think we need an additional method `setImage(Provider<String>)`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -33,28 +36,27 @@
...
   private Set<String> tags = Collections.emptySet();
   @Nullable private String credHelper;
 
   @Inject
-  public TargetImageParameters(ObjectFactory objectFactory) {
+  public TargetImageParameters(ObjectFactory objectFactory, Project project) {
     auth = objectFactory.newInstance(AuthParameters.class, "to.auth");
+    this.project = project;
+    image = project.getObjects().property(String.class);
+    image.set(project.provider(() -> System.getProperty(PropertyNames.TO_IMAGE)));
```
chanseokoh, line 50: I'd like to keep the consistency that we read system properties in getters.

chanseokoh (commented):
Aside from writing unit tests, have you manually tested if this works?

mpeddada1:
> Aside from writing unit tests, have you manually tested if this works?
I'm still manually experimenting with an example `build.gradle` to see if it actually works so this PR is still a work in progress.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
mpeddada1: good call

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -34,42 +36,51 @@
...
 
   @Inject
   public TargetImageParameters(ObjectFactory objectFactory) {
     auth = objectFactory.newInstance(AuthParameters.class, "to.auth");
+    image = objectFactory.property(String.class);
+    tags = objectFactory.setProperty(String.class);
   }
 
   @Input
-  @Nullable
```
chanseokoh, line 47: Probably we need to keep `@Nullable`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -34,42 +36,51 @@
...
+    return image.getOrNull();
   }
 
   public void setImage(String image) {
-    this.image = image;
+    this.image.set(image);
+  }
+
+  public void setImage(Provider<String> image) {
+    if (image != null) {
```
chanseokoh, line 63: I speculate that we won't need this null check.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -34,42 +36,51 @@
...
-  private Set<String> tags = Collections.emptySet();
+  private Property<String> image;
+  private SetProperty<String> tags;
   @Nullable private String credHelper;
 
   @Inject
   public TargetImageParameters(ObjectFactory objectFactory) {
     auth = objectFactory.newInstance(AuthParameters.class, "to.auth");
+    image = objectFactory.property(String.class);
+    tags = objectFactory.setProperty(String.class);
```
chanseokoh, line 47: Maybe it's safer to initialize it to an empty set like we do it here:

https://github.com/GoogleContainerTools/jib/blob/94f8bdafe0f816c548f96392b34660d5574cc5c4/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java#L116

chanseokoh (commented):
Just went over it briefly as a draft.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
mpeddada1: hm removing this check is resulting in an `UnexpectedBuildFailure` with the message, `Cannot set the value of a property using a null provider` in places where the target image value is not set. (For example, https://github.com/GoogleContainerTools/jib/blob/f892bdba90cbe64a151e0196fa0c8e633f529a7f/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2Test.java#L110

chanseokoh: Aha, looks like Gradle prefers this method signature over `String`. If that's the case, I think it should be

```suggestion
    if (image == null) {
      this.image = null;
    } else {
      this.image.set(image);
    }
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -34,42 +36,52 @@
...
+    return tags.get();
   }
 
   public void setTags(Set<String> tags) {
-    this.tags = tags;
+    this.tags.set(tags);
+  }
+
+  public void setTags(Provider<Set<String>> tags) {
+    this.tags.set(tags);
```
chanseokoh, line 84: In light of https://github.com/GoogleContainerTools/jib/pull/2739/files#r485008215, I wonder if we need a null check. Can you run a manual test and see which of the two `setTags()` gets called?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -34,42 +36,52 @@
...
+  }
+
+  public void setImage(Provider<String> image) {
+    if (image != null) {
+      this.image.set(image);
+    }
   }
 
   @Input
-  @Optional
```
chanseokoh, line 61: I think this should be preserved. The user doesn't have to configure this parameter.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -34,42 +36,52 @@
...
   @Inject
   public TargetImageParameters(ObjectFactory objectFactory) {
     auth = objectFactory.newInstance(AuthParameters.class, "to.auth");
+    image = objectFactory.property(String.class);
+    tags = objectFactory.setProperty(String.class).empty();
   }
 
   @Input
   @Nullable
-  @Optional
```
chanseokoh, line 48: I think this should be preserved. The user doesn't have to configure this parameter.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
mpeddada1: Ah, exactly! You're right, when `jib.to.image` is set to `null` then Gradle prefers this method. With the `if (image != null)` checker, we are currently returning the message `Missing target image parameter, perhaps you should add a 'jib.to.image' configuration parameter to your build.gradle...` if `jib.to.image` is not specified, which seems like expected behavior. I could be missing something here but why do we want to make the Property be null instead of not just not setting a value for it?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
mpeddada1: That is a good idea. So unlike the case with the image, setting `jib.to.tags` to null is resulting in GroovyRuntimeException with `Ambiguous method overloading...Cannot resolve which method to invoke for [null] due to overlapping prototypes between` where one `setTags` is not preferred over the other. The issue occurs when _only_ `setTags(Provider<Set<String>>)` is present, which is when setting `jib.to.tags=null` results in the `Cannot set the value of property using null provider` message.

mpeddada1: Passing in a set or provider<set<string>> that _contains_ `null` seems to resulting in NPE so I think we need to handle that?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
chanseokoh: Good question.

From the general API perspective, irrelevant of Jib: let's say a Java developer is using some Java library which has `Person.setEmail(String)`. It's an optional property, and by default `getAdress()` will return `null`. Let's say the developer does
   ```java
   // prints out "no email given"
   System.out.println(person.getEmail() == null ? "no email given" : person.getEmail());
   person.setEmail("foo@example.com");
   person.setEmail(null);
   System.out.println(person.getEmail() == null ? "no email given" : person.getEmail());
   ```
   Then, which would be reasonable for the last `println()` to print: `no email given` or `foo@example.com`? Of course, it may depend on the situation, but in this case, I expect `setEmail(null)` to clear the previous value and make the property "missing"; it is counter-intuitive it becomes a no-op. 

Also note that `jib.to.image` is basically an optional property (as indicated by the `@Optional` annotation). Of course, it is required when pushing to a remote registry (`jib:build`), and it will say 'Missing target image parameter'. However, it isn't required when pushing to a local Docker daemon (`jib:dockerBuild`); in this case, Jib will auto-pick some name for it.

But, before we add the null check, I'd like to verify Gradle always calls `setImage(null)` even when `jib.to.image` is simply missing (not configured) rather than when `jib.to.image` is explicitly set to `null`. Reading https://github.com/GoogleContainerTools/jib/pull/2739/files#r485907745, it looks like this is only when explicitly setting `jib.to.image` to `null`. In that case, I don't think we need the null check actually; I think we can just fix the failing test.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
chanseokoh: Hmm... I think we should be fine as long as nothing breaks when `jib.to.tags` is missing (not configured). People shouldn't explicitly set it to `null` anyway. That said, I now wonder the test failure you mentioned in https://github.com/GoogleContainerTools/jib/pull/2739/files#r485008215 is when you explicitly set the property to `null` rather than when it's missing. If former, I don't think we need the null check there either. 

> Passing in a set or provider<set> that contains null seems to resulting in NPE so I think we need to handle that?

Let's file an issue for that. It basically sounds like an orthogonal issue we don't want to mix in here.

mpeddada1: Right that's a good point. I filed an issue for it in #2760

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
mpeddada1: Thank you so much for the detailed explanation! 

And with regards to the last point about https://github.com/GoogleContainerTools/jib/pull/2739/files#r485907745, the issue only occurs when `jib.to.image` is explicitly set to `null`. I believe the tests that were failing set `jib.to.image` to `System.getProperty("_TARGET_IMAGE") and didn't initialize `_TARGET_IMAGE`? Because System.getProperty(String key) returns `null` if there is no property with that key.

chanseokoh: Yeah, that makes sense. I think we don't actually need a null check; we can just fix the test, e.g., setting `_TARGET_IMAGE` to "ignored."

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2Test.java
```diff
@@ -58,7 +58,7 @@
   private static String verifyTaskSuccess(TestProject project, @Nullable String moduleName) {
     String taskName =
         ":" + (moduleName == null ? "" : moduleName + ":") + JibPlugin.SKAFFOLD_FILES_TASK_V2_NAME;
-    BuildResult buildResult = project.build(taskName, "-q");
+    BuildResult buildResult = project.build(taskName, "-q", "-D_TARGET_IMAGE");
```
chanseokoh, line 61: How about `-D_TARGET_IMAGE=ignored`?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/skaffold/SyncMapTaskTest.java
```diff
@@ -61,6 +61,7 @@ private static SkaffoldSyncMapTemplate generateTemplate(
     List<String> buildParams = new ArrayList<>();
     buildParams.add(taskName);
     buildParams.add("-q");
+    buildParams.add("-D_TARGET_IMAGE");
```
chanseokoh, line 64: ditto

chanseokoh (commented):
Looks good to me. I'll let @loosebazooka approve this.

chanseokoh:
> * If user wants to use lazy evaluation for `jib.to.image` and `jib.to.tags` with String `value` they can do the following:
>   `jib.to.image = project.provider{value}` or `jib.to.tags = project.provider{[value, "tag2"]}`

@loosebazooka so we decided to accept `Provider<String>`, but I guess `Supplier<String>` will similarly work, which will enable syntax like `jib.to.image = () -> "value"`. I just thought using `Provider` was like a convention, but I'm actually curious why we decided to use `Provider` and not something like `Supplier`.

jib-gradle-plugin/src/test/resources/gradle/projects/lazy-evaluation/src/main/java/com/test/HelloWorld.java
```diff
@@ -0,0 +1,36 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.net.URISyntaxException;
+import java.nio.charset.StandardCharsets;
+
+public class HelloWorld {
+
+  public static void main(String[] args) throws URISyntaxException, IOException {
+    try (Reader reader =
```
loosebazooka, line 29: If you're not using the contents of this file to do anything special, just a simple `System.out.println("Hello world")` is probably sufficient.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -370,6 +373,39 @@ public void testJibTaskGroupIsSet() {
         taskName -> Assert.assertEquals(taskName, "Jib", tasks.getByPath(taskName).getGroup()));
   }
 
+  @Test
+  public void testLazyEvalForImageAndTags() {
+    try {
+      testProject.build(JibPlugin.JIB_EXTENSION_NAME);
```
loosebazooka, line 379: I think this technically works, but is confusing. Maybe `JibPlugin.BUILD_IMAGE_TASK_NAME` is more appropriate?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -370,6 +373,39 @@ public void testJibTaskGroupIsSet() {
...
+      Assert.fail("Expect this to fail");
+    } catch (UnexpectedBuildFailure ex) {
+      String output = ex.getBuildResult().getOutput().trim();
+
+      // Regex to parse through image and tag values from build output.
+      String cyanWordRegex = "\\u001B\\[36m(.+?)\\u001B\\[0m";
```
loosebazooka, line 385: I don't know if this is true, but if you use `--jib.console=plain`, it might be easier to read the output, and you can ignore the cyan coloring and stuff?

Also since we already know the image name and tags, could we just use those in the comparison directly instead of extracting them out and then using the Assert at line 404, 405?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
mpeddada1: Ah that makes more sense

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
mpeddada1: Thanks for the suggestion! Unfortunately, `jib.console=plain` doesn't display the progress bar but still keeps the cyan coloring:(

> Also since we already know the image name and tags, could we just use those in the comparison directly instead of extracting them out and then using the Assert at line 404, 405?
Good point, I just applied the change.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2Test.java
mpeddada1: Yup, this is much better.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
chanseokoh: How about passing `--console=plain` to Gradle? (Note it is not `-D...` this time.)

mpeddada1: Hm the cyan coloring still appears when `--console=plain` is passed to Gradle and when `Djib.console=plain` is used.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -370,6 +373,37 @@ public void testJibTaskGroupIsSet() {
         taskName -> Assert.assertEquals(taskName, "Jib", tasks.getByPath(taskName).getGroup()));
   }
 
+  @Test
+  public void testLazyEvalForImageAndTags() {
+    try {
+      testProject.build(JibPlugin.BUILD_IMAGE_TASK_NAME);
+      Assert.fail("Expect this to fail");
+    } catch (UnexpectedBuildFailure ex) {
+      String output = ex.getBuildResult().getOutput().trim();
```
loosebazooka, line 382: Does it makes sense to remove this whole regex and just do

cleanOutput = output.replace("color-stuff","")

Assert.assertThat(cleanOutput, Matchers.contains("Containerizing application....")

mpeddada1: Nice. Although, `assertThat` seems to have been deprecated so I used `assertTrue(matcher.find())` instead?

loosebazooka: Oh yeah, you can use `MatcherAssert` like in: https://github.com/GoogleContainerTools/jib/blob/cc930f213a66c3895e5e7f3890566a3f50acce97/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/skaffold/InitMojoTest.java#L66

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
chanseokoh: Ugh, you're right. I see there are a couple things that are always colored.
```
Containerizing application to Docker daemon as francium25/test...
```
```
Built image to Docker daemon as francium25/test
```
I think we should fix this. Could you file a bug (and also link this issue to get rid of the interim code to handle color)?

chanseokoh:
@loosebazooka can you answer https://github.com/GoogleContainerTools/jib/pull/2739#issuecomment-690727510?

loosebazooka:
> @loosebazooka so we decided to accept Provider<String>, but I guess Supplier<String> will similarly work, which will enable syntax like jib.to.image = () -> "value". I just thought using Provider was like a convention, but I'm actually curious why we decided to use Provider and not something like Supplier.

Well the underlying type is Property which can potentially accept either a Provider or Property, Provider by nature appears to be immutable, so I don't think the base type should not be Provider for user configured objects (but I think this is only enforced by convention or by using final).

As for Supplier, I'm not sure, I can't find any docs on it at the moment and it appears to be an internal object?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
mpeddada1: Yup, that's a good idea. I added a TODO with reference to issue #2764.

loosebazooka (approved):

PR #2763, opened 2020-09-15
Remove PlatformSpec unsupported parameters for now

loosebazooka:
Since jib-core doesn't supported advanced platform spec options yet, remove these and reintroduce them later.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
@@ -36,11 +36,11 @@ import com.fasterxml.jackson.annotation.JsonProperty;
  *   - aes
  * }</pre>
  */
+// TODO: reintroduce platform details when ready
+// TODO: revert https://github.com/GoogleContainerTools/jib/pull/2763
 public class PlatformSpec {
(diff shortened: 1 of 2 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
```diff
@@ -43,41 +39,23 @@
 public class PlatformSpec {
   private final String architecture;
   private final String os;
-  @Nullable private final String osVersion;
-  private final List<String> osFeatures;
-  @Nullable private final String variant;
-  private final List<String> features;
+  // TODO: reintroduce platform details when ready
+  // TODO: revert https://github.com/GoogleContainerTools/jib/pull/2763
```
mpeddada1, line 43: nit:` revert  https://github.com/GoogleContainerTools/jib/pull/2763 once platform spec options are supported or once ${issue} is submitted`?

Maybe we can move the TODO to the javadoc, since it pertains to the entire class.

loosebazooka: good call, will move to top of class, but since it's not really exposed documentation, I left it out of the javadoc.

chanseokoh (approved):

PR #2770, opened 2020-09-18
Use base image as key for builtImagesAndContainerConfigurationPushResults

chanseokoh:
On top of #2769. Review #2769 first.

In the same vein in #2769, updates the map to use base images as the primary key.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -71,17 +71,17 @@ public class StepsRunner {
     private Future<ImagesAndRegistryClient> baseImagesAndRegistryClient = failedFuture();
     private Future<Map<Image, List<Future<PreparedLayer>>>> baseImagesAndLayers = failedFuture();
     @Nullable private List<Future<PreparedLayer>> applicationLayers;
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -550,7 +552,7 @@ private void pushImages() {
               Verify.verifyNotNull(results.baseImagesAndLayerPushResults.get().get(baseImage)));
 
           Future<BlobDescriptor> containerConfigPushResult =
-              results.builtImagesAndContainerConfigurationPushResults.get().get(builtImage);
+              results.baseImagesAndContainerConfigPushResults.get().get(baseImage);
```
loosebazooka, line 555: was this previously wrong?

loosebazooka (approved):

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: No. Previously the key of the map was built images, so `Map.get()` accepted `builtImage`. Now this PR changed the key to base images, so `Map.get()` should accept `baseImage`. The end result remains same.

PR #2772, opened 2020-09-18
Refactor StepsRunner for consistent loop pattern

chanseokoh:
Apply the following consistent pattern over multiple base images:

```java
void scheduleImages() {
  List<Future<R>> results = ...
  for (Image i : images) {
    Future<R> result = scheduleImage(i);
    results.add(result);
  }
  return results;
}

Future<R> scheduleImage(Image i) {
  return executorService.submit(...);
}
```

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/PlatformSpec.java
@@ -18,10 +18,6 @@ package com.google.cloud.tools.jib.cli.buildfile;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.collect.ImmutableList;
-import java.util.List;
-import java.util.Optional;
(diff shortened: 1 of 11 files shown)


loosebazooka (commented):
This seems fine, but it looks like we're just refactoring out some logic into methods? It's kind of hard to see where this is going especially since this isn't triggering some new testing or code sharing? Can you elaborate a little more?

chanseokoh:
Yeah, just extracting inner code of a loop into a method. It's not that this refactoring is a prerequisite of something I am planning. It's just that the code pattern in `StepsRunner` is mixed, so I just wanted to make it consistent as a cleanup. For example, `buildImages()` and `buildImage()` are already using this pattern, while others are not:

https://github.com/GoogleContainerTools/jib/blob/11e21ab51e96a00291472f3c5fab21755ef703c1/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java#L518-L550

The pattern is that any `for` loop in `StepsRunner` is a loop over multiple base images, and the inner loop code is a thread scheduling operation on each image.

loosebazooka (approved):

PR #2773, opened 2020-09-18
Fix to handle integration test failure caused by #2739

mpeddada1:
Specifically, [this](https://github.com/GoogleContainerTools/jib/blob/faff19bd59737e10db4425e47b7e6c9904e83833/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/EmptyProjectIntegrationTest.java#L76) integration test was failing. This PR serves as a follow-up to #2739.

### DIFF
--- a/jib-gradle-plugin/src/integration-test/resources/gradle/projects/empty/build.gradle
+++ b/jib-gradle-plugin/src/integration-test/resources/gradle/projects/empty/build.gradle
@@ -26,5 +26,5 @@ jib {
 
 def additionalTag = System.getProperty("_ADDITIONAL_TAG")
 if (additionalTag != null && !additionalTag.isEmpty()) {
-  jib.to.tags = [System.getProperty("_ADDITIONAL_TAG")] as Set
+  jib.to.tags = [System.getProperty("_ADDITIONAL_TAG")]
 }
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/src/integration-test/resources/gradle/projects/empty/build.gradle
```diff
@@ -26,5 +26,5 @@ jib {
 
 def additionalTag = System.getProperty("_ADDITIONAL_TAG")
 if (additionalTag != null && !additionalTag.isEmpty()) {
-  jib.to.tags = [System.getProperty("_ADDITIONAL_TAG")]
+  jib.to.tags = [System.getProperty("_ADDITIONAL_TAG")] as Set
```
chanseokoh, line 29: Hmm... I think we should keep it backward-compatible. It will be common to use a list. I suspect that adding another `setTags()` stops Gradle from auto-casting in to a `Set`. @loosebazooka maybe we need to add a `setTags(List)` too?

mpeddada1: You're absolutely right. I just realized that this change doesn't really solve the backward-compatibility issue.

chanseokoh (approved):

PR #2779, opened 2020-09-22
Refactor StepsRunner to have step methods accept progressDispatcherFactory

chanseokoh:
The current `StepsRunner` code makes it hard to write unit tests for each method.

Having each step method accept `ProgressEventDispatcher.Factory` is also consistent overall, as some methods (those working on individual images) already accept it like this one:

https://github.com/GoogleContainerTools/jib/blob/664d6615850964915bfc42cb28eb3d7ef1447473/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java#L430-L434

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -47,6 +47,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -115,17 +116,16 @@ public static StepsRunner begin(BuildContext buildContext) {
...
+  // the unfortunate chicken-and-egg situation when using ProgressEventDispatcher. The current
+  // ProgressEventDispatcher model requires allocating the total units of work (i.e., steps)
+  // up front. That is, to instantiate a root ProgressEventDispatcher, we should know ahead how many
+  // steps we will run. However, to run a step, we need a root progress dispatcher. So, we take each
+  // step as a lambda and save them to run later. Then we can count the number of lambdas, create a
+  // root dispatcher with the count, and run the saved lambdas using the dispatcher.
+  private final List<Consumer<ProgressEventDispatcher.Factory>> stepsToRun = new ArrayList<>();
 
   @Nullable private String rootProgressDescription;
-  @Nullable private ProgressEventDispatcher rootProgressDispatcher;
```
mpeddada1, line 128: Do we no longer need to initialize a root progress dispatcher?

mpeddada1 (commented):
What is it about StepsRunner that makes it challenging to write unit tests?

chanseokoh:
Good question, as I haven't documented the motivation in the PR. Suppose that I want to test `pullBaseImages()` by calling it in a unit test. However, `pullBaseImages()` assumes a particular instance state where `rootProgressDispatcher` is initialized.
```java
  private void pullBaseImages() {
```

In the current code, `rootProgressDispatcher` is initialized when calling `StepsRunner.run()`, which is the final method to actually execute all the registered steps. If we change the method signature to
```java
  private void pullBaseImages(ProgressEventDispatcher.Factory childProgressDispatcherFactory) {
```
then I can just pass a mock factory when calling `pullBaseImages()` in a test.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: Right, this is no longer needed, as each method (such as `PullBaseImages()`) accepts a factory from the `StepsRunner.run()` code, rather than the method retrieves a factory from `rootProgressDispatcher` on its own.

loosebazooka (approved):

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
mpeddada1: Thanks!

PR #2780, opened 2020-09-22
Don't schedule multiple threads downloading same base image layers in StepsRunner

### LINKED ISSUE #2750: Do not push duplicate base image layers when building multi-platform images
High priority. Base images of different platforms may share layers. Impacts performance a lot, as base image layers are large.

Also, it's uncertain if pushing the same layer concurrently will cause any trouble.

chanseokoh:
~~On top of #2779. Review #2779 first.~~

Fixes #2750.

Eventually I decided to go with the minimally disruptive approach to resolve the issue.

Inside `obtainBaseImageLayers()` where the method schedules threads for each layer in an `Image`, it now checks first if a thread was previously scheduled for the given layer (by comparing the layer digest). For implementing the check, I instantiate a `Map` of (digest --> layer `Future`) where the key is a layer digest. The method schedules a thread (running `ObtainBaseImageLayerStep`) only when the layer `Future` is missing in the `Map`.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
@@ -68,21 +68,22 @@ public class StepsRunner {
           new IllegalStateException("invalid usage; required step not configured"));
     }
 
-    private Future<ImagesAndRegistryClient> baseImagesAndRegistryClient = failedFuture();
-    private Future<Map<Image, List<Future<PreparedLayer>>>> baseImagesAndLayers = failedFuture();
     @Nullable private List<Future<PreparedLayer>> applicationLayers;
-    private Future<Map<Image, Future<Image>>> baseImagesAndBuiltImages = failedFuture();
     private Future<ManifestTemplate> manifestListOrSingleManifest = failedFuture();
     private Future<RegistryClient> targetRegistryClient = failedFuture();
-    private Future<Map<Image, List<Future<BlobDescriptor>>>> baseImagesAndLayerPushResults =
-        failedFuture();
     private Future<List<Future<BlobDescriptor>>> applicationLayerPushResults = failedFuture();
-    private Future<Map<Image, Future<BlobDescriptor>>> baseImagesAndContainerConfigPushResults =
-        failedFuture();
     private Future<Optional<ManifestAndDigest<ManifestTemplate>>> manifestCheckResult =
         failedFuture();
     private Future<List<Future<BuildResult>>> imagePushResults = failedFuture();
     private Future<BuildResult> buildResult = failedFuture();
+
+    private Future<ImagesAndRegistryClient> baseImagesAndRegistryClient = failedFuture();
+    private Future<Map<Image, List<Future<PreparedLayer>>>> baseImagesAndLayers = failedFuture();
+    private Future<Map<Image, List<Future<BlobDescriptor>>>> baseImagesAndLayerPushResults =
+        failedFuture();
+    private Future<Map<Image, Future<BlobDescriptor>>> baseImagesAndContainerConfigPushResults =
+        failedFuture();
+    private Future<Map<Image, Future<Image>>> baseImagesAndBuiltImages = failedFuture();
   }
 
   /**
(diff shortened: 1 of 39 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/StepsRunnerTest.java
```diff
@@ -0,0 +1,151 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** Tests for {@link StepsRunner}. */
+@RunWith(MockitoJUnitRunner.class)
+public class StepsRunnerTest {
+
+  // ListeningExecutorService is annotated with @DoNotMock, so define a concrete class.
```
loosebazooka, line 49: The instructions seem to indicate maybe a different approach, but I don't know if you already considered that.

> Use TestingExecutors.sameThreadScheduledExecutor, or wrap a real Executor from java.util.concurrent.Executors with MoreExecutors.listeningDecorator

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -306,36 +311,61 @@ private void obtainBaseImagesLayers(
...
                   baseImagesAndLayers.put(baseImage, layers);
                 }
                 return baseImagesAndLayers;
               }
             });
   }
 
-  private List<Future<PreparedLayer>> obtainBaseImageLayers(
+  @VisibleForTesting
+  List<Future<PreparedLayer>> obtainBaseImageLayers(
```
loosebazooka, line 331: I thought that maybe this could have threading issues, but we never call this in parallel right? Is there any value in marking this not threadsafe?

loosebazooka (approved):

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
chanseokoh: Added a comment
```
// This method updates the given "preparedLayersCache" and should not be called concurrently.
```
. (FYI, `@NotThreadSafe` can only be applied to a class.)

The non-blocking method is called only by a single thread task inside `obtainBaseImagesLayers()`.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/StepsRunnerTest.java
chanseokoh: I saw the comment, but I think it cannot apply to what I'm trying to achieve. `TestingExecutors.sameThreadScheduledExecutor` and `MoreExecutors.listeningDecorator` actually executes given tasks, whereas here I want to mock it at a higher level to simulate tasks.

I tried quite a lot of time to find the simplest way to mock `ListeningExecutorService.submit()`, but this code was the shortest I could get so far.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -306,36 +311,62 @@ private void obtainBaseImagesLayers(
...
+
+    try (ProgressEventDispatcher progressDispatcher =
+        progressDispatcherFactory.create(
+            "launching base image layer pullers", baseImage.getLayers().size())) {
+      for (Layer layer : baseImage.getLayers()) {
+        DescriptorDigest digest = layer.getBlobDescriptor().getDigest();
+        Future<PreparedLayer> preparedLayer = preparedLayersCache.get(digest);
+
+        if (preparedLayer != null) {
+          progressDispatcher.dispatchProgress(1);
```
mpeddada1, line 348: Just a thought: we currently check to see if the layer already exists before pushing a new one, but what would we do if we needed to update a layer?/ Can base image layers be updated?

chanseokoh: Not sure I understand your point. To clarify a few things,
- This section of code is to *download* a base image layer.
- In a later stage, base image layers need to be pushed to a target registry (but only those layers missing in the registry).
- A layer is just a block of bytes (blob) identified and addressed by a SHA digest. And every blob is immutable and content-addressable. There's nothing like "updating" a blob or an existing image. If a content of a blob or an "image" is different, it is a new blob or a new image with different addresses.

mpeddada1: I see, thanks for clarifying! This really clears up some of the key concepts I was missing:)

PR #2783, opened 2020-09-23
Fix platform configuration in Gradle plugin

### LINKED ISSUE #2781: jib-gradle-plugin fails when the platform/architecture is not amd64.
**Environment**:

- *Jib version: 2.5.0 
- *Build tool: Gradle
- *OS: OS X

**Description of the issue**: The jib-gradle-plugin is not replacing the given platform config but adding to an existing collection[1]. Jib-core requires exactly one in the collection[2].

Side note: I realize multi-platform builds are not supported yet. I'm just trying to do a single non-amd64 build.

BTW, Thanks for the great tool!

[1] https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/PlatformParametersSpec.java#L45
[2] https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/api/JibContainerBuilder.java#L597

**Expected behavior**:
Jib should download the arm64-based base image and build.

**Steps to reproduce**:

  1. Setup the jib config provided below.
  2. `./gradlew bootJar jibDockerBuild`
  3. Fail

```groovy
jib {
    from {
        image = "amazoncorretto:11"
        platforms {
            platform {
                architecture = 'arm64'
                os = 'linux'
            }
        }
    }
}
```

**Log output**: 
```
> multi-platform image building is not yet supported
```

**Additional Information**: 
The platforms collection has 2 items even though I'm only specifying 1 in the config. I'm not a gradle internals dev, but I'm assuming this gets invoked regardless of if I try to override it in the config or not: https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java#L48

chanseokoh:
Fixes #2781.

We started with the default amd64/linux, and we were always adding new platforms.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -49,17 +49,18 @@ public class BaseImageParameters {
   @Nested
   @Optional
   public ListProperty<PlatformParameters> getPlatforms() {
-    if (platforms.get().isEmpty()) {
-      PlatformParameters amd64Linux = new PlatformParameters();
-      amd64Linux.setArchitecture("amd64");
-      amd64Linux.setOs("linux");
-
-      ListProperty<PlatformParameters> platforms =
-          objectFactory.listProperty(PlatformParameters.class);
-      platforms.add(amd64Linux);
+    if (!platforms.get().isEmpty()) {
       return platforms;
     }
-    return platforms;
+
+    PlatformParameters amd64Linux = new PlatformParameters();
+    amd64Linux.setArchitecture("amd64");
+    amd64Linux.setOs("linux");
+
+    ListProperty<PlatformParameters> defaultPlatforms =
+        objectFactory.listProperty(PlatformParameters.class);
+    defaultPlatforms.add(amd64Linux);
+    return defaultPlatforms;
   }
 
   public void platforms(Action<? super PlatformParametersSpec> action) {
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -41,29 +41,24 @@ public class BaseImageParameters {
   public BaseImageParameters(ObjectFactory objectFactory) {
     this.objectFactory = objectFactory;
     auth = objectFactory.newInstance(AuthParameters.class, "from.auth");
-    platforms = objectFactory.listProperty(PlatformParameters.class).empty();
+    platforms = objectFactory.listProperty(PlatformParameters.class);
     platformParametersSpec =
         objectFactory.newInstance(PlatformParametersSpec.class, objectFactory, platforms);
+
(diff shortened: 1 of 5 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -35,23 +35,32 @@
...
-    platform.setOs("linux");
-    platform.setArchitecture("amd64");
-    platforms.add(platform);
   }
 
   @Nested
   @Optional
   public ListProperty<PlatformParameters> getPlatforms() {
-    return platforms;
+    if (!platforms.get().isEmpty()) {
```
loosebazooka, line 52: It does feel a little weird that a `getter` is mutating things.

loosebazooka: could we just delegate default setting to a lower level?

chanseokoh: Technically it doesn't mutate an internal state, but it's true the class is no longer a POJO or bean. I also didn't like it and wondered other solutions yesterday. Setting the default at other levels could be another way, but it also had some cons.

Today, I think I found a better way than these: start the class with the default value as before, but only reset it when the user specifies `platforms {}`. This behavior is also more or less similar to how it works in Maven; if nothing is defined, Maven assumes amd64/linux, but as soon as the user specifies `<platforms>`, it becomes an empty list.. What do you think?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -38,14 +38,14 @@
   @Inject
   public BaseImageParameters(ObjectFactory objectFactory) {
     auth = objectFactory.newInstance(AuthParameters.class, "from.auth");
-    platforms = objectFactory.listProperty(PlatformParameters.class).empty();
+    platforms = objectFactory.listProperty(PlatformParameters.class);
```
chanseokoh, line 41: `empty()` turns out to be unnecessary.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
loosebazooka: Hrmm, I just realized I don't understand how this affects verifying image compatibility? Do we have code somewhere that verifies and image matches the specified platform?

chanseokoh: #2781 and this PR have nothing to do with verifying image platforms. The issue was that defining
```
jib.from.platforms {
  platform {
    architecture = 'non-amd64'
    os = '...'
  }
}
```
incorrectly resulted in *two* platforms configured: the configured one plus the default amd64/linux. This was because `platform { ...}` just called `ListProperty.add()`.

loosebazooka (approved):

PR #2785, opened 2020-09-24
Add templating to jib-cli

loosebazooka:
Allow templating in cli buildfiles

Expected usage:
```
$ jib build -Pkey=value
```
will replace any instances of `${key}` with `value' in the raw build file. This means the user can technically replace whole sections of the buildfile.

### DIFF
--- a/jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
+++ b/jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
@@ -157,7 +157,8 @@ public class BuildFilesTest {
 
     JibContainerBuilder jibContainerBuilder =
         BuildFiles.toJibContainerBuilder(
-            Paths.get(resource.toURI()), ImmutableMap.of("replace\nthis", "creationTime: 1234"));
+            Paths.get(resource.toURI()),
+            ImmutableMap.of("replace" + System.lineSeparator() + "this", "creationTime: 1234"));
     ContainerBuildPlan resolved = jibContainerBuilder.toContainerBuildPlan();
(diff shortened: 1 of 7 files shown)


jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
```diff
@@ -105,4 +108,58 @@ public void testToJibContainerBuilder_requiredProperties()
...
+  @Test
+  public void testToBuildFileSpec_withTemplating()
+      throws URISyntaxException, InvalidImageReferenceException, IOException {
+    URL resource = Resources.getResource("buildfiles/projects/templating/valid.yaml");
+
+    JibContainerBuilder jibContainerBuilder =
+        BuildFiles.toJibContainerBuilder(
+            Paths.get(resource.toURI()),
+            ImmutableMap.of(
+                "unused", "ignored", // keys that are defined but not used do not throw an error
```
mpeddada1, line 121: nit: 
```suggestion
                "unused", "ignored", // keys that are defined but not used so do not throw an error
```

mpeddada1 (commented):
LGTM. For documentation purposes, what could be a valid use case for templating?

loosebazooka:
Oh right, so lets say you have a buildfile but you have some artifact that is versioned or something

```
$ gradle build
-> outputs myjar-1.0.9.jar
```
and you have a build file that you don't want to keep changing
```
entrypoint: 
  - java
  - -jar
  - ${jar-name}
```
then you can just run the cli with that property defined on the command line
```
$ jib build -Pjar-name=myjar-1.0.9.jar
```
I mean this is just an example, but it can be useful for anything where a value is not fixed in the build file and is determined by some external factor.

mpeddada1 (approved):

PR #2787, opened 2020-09-25
Cli surface proposal

loosebazooka:
Proposal for design for the cli surface. Implementation will follow

Rendered: https://github.com/GoogleContainerTools/jib/blob/cli-surface-proposal/proposals/jib-cli-surface.md

### DIFF
--- a/proposals/jib-cli-surface.md
+++ b/proposals/jib-cli-surface.md
@@ -29,6 +29,8 @@ build        build a container
 
 #### Build Config
 ```
+    --application-cache <directory>        location of the application cache
+    --base-image-cache <directory>         location of the base image cache
 -b, --build-file <file>                    location of the build file (default <context>/jib.yaml)
 -c, --context <dir>                        location of the build context (default .)
     --docker-config <directory>            location of docker configuration
--- a/proposals/jib-cli-surface.md
+++ b/proposals/jib-cli-surface.md
@@ -90,5 +90,6 @@ Information on communicating with the docker daemon is determined using the `DOC
 
 ## Other considerations
 
-- Users have asked to warm up the cache, perhaps we can have a cache build?
+- Users have asked to warm up the cache, perhaps we can have a `cache` command option
 - In the future we should be able to build `jar`s from the command line
+- Default application cache directory is a temporary directory (will lead to slow rebuilds), we should do a better job of defining a default here.
--- a/proposals/jib-cli-surface.md
+++ b/proposals/jib-cli-surface.md
@@ -29,6 +29,7 @@ build        build a container
 
 #### Build Config
 ```
+    --additional-tags <tag1>[,<tag2>,...]  additional tags for target
     --application-cache <directory>        location of the application cache (jib default is temp directory)
     --base-image-cache <directory>         location of the base image cache (jib default is user cache)
 -b, --build-file <file>                    location of the build file (default <context>/jib.yaml)
@@ -36,7 +37,6 @@ build        build a container
     --docker-config <directory>            location of docker configuration
     --name <image-ref>                     image name to bake into tar file, required when using "-t tar://..." 
 -p, --parameter <name>=<value>             templating parameters replace `${name}` with `value` in the build file (repeatable)
-    --tag <tag1>[,<tag2>,...]              additional tags for target
 ```
 
 #### Auth/Security
--- a/proposals/jib-cli-surface.md
+++ b/proposals/jib-cli-surface.md
@@ -1,0 +1,33 @@
+# Jib CLI Surface Specification
+
+Specification for a command structures exposed to users of the jib-cli
+
+## Usage
+`jib command [OPTIONS]`
+
+## Options
+
+### Required
+1. `-t, --target` a target image reference
+
+### Optional
+1. `-b, --build-file` location of the build file (default ./jib.yaml)
+1. `--credHelper<registry>=<credHelper>` the credential helper to use for a registry
+1. `-P<name>=<value>, `, templating parameters replace `${name}` with `value` in the build file
+1. `--tags` additional tags for target (optional)
+1. `-v, --version` print version information and exit
+1. `--verbostiy` change logging verbosity (error, warn, lifecycle (default), info, debug)
+
+## Commands
+1. `build` build to registry
+1. `dockerBuild` build to docker daemon
+1. `buildTar` build to a tar file
+
+## `build`
+
+## `dockerBuilder` only
+
+## `buildTar` only
+
+### Required
+1. `-f` output tar file  
--- a/proposals/jib-cli-surface.md
+++ b/proposals/jib-cli-surface.md
@@ -48,9 +48,8 @@ build        build a container
 
 Credentials can be specified using credential helpers or username + password. The following options are available
 ```
-    --credHelper <credHelper>              credential helper to use for a registry
-    --to-credHelper <credHelper>           credential helper to use for target image registry
-    --from-credHelper <credHelper>         credential helper to use for base image registry
+    --credential-helper <credHelper>       credential helper to use for a registry, a path or name suffix (docker-credential-<suffix>) (repeatable)
+
     --username <username>                  configure a username for authenticating against registries
     --password <password>                  configure a passowrd for authenticating against registries (interactive if <password> is omitted)
     --to-username <username>               configure a username for authenticating on the registry that an image is being built to
@@ -58,22 +57,12 @@ Credentials can be specified using credential helpers or username + password. Th
     --from-username <username>             configure a username for authentication on the registry that a base image is being sourced from
     --from-password <password>             configure a password for authentication on the registry that a base image is being sourced from (interactive if <password> is omitted)
 ```
-Of these, only the following combinations of the above options are allowed
-###### Cred Helpers only
-1. `--credHelper`
-1. `--to-credHelper`
-1. `--from-credHelper`
-1. `--to-credHelper`, `from-credHelper`
-
-###### Username/Password only
+combinations of `username` and `password` flags come with restrictions and can be use only in the following ways:
 1. `--username`, `--password`
 1. `--to-username`, `--to-password`
 1. `--from-username`, `--from-password`
 1. `--to-username`, `--to-password`, `--from-username`, `--from-password`
 
-###### Mixed CredHelper and Username/Password
-1. `--to-credHelper`, `--from-username`, `--from-password`
-1. `--to-username`, ``--to-password`, `--from-credhelper`
 
 #### Info Params
 ```
--- a/proposals/jib-cli-surface.md
+++ b/proposals/jib-cli-surface.md
@@ -8,26 +8,55 @@ Specification for a command structures exposed to users of the jib-cli
 ## Options
 
 ### Required
-1. `-t, --target` a target image reference
+```
+-t  --target <image-ref>  a target image reference
+```
 
 ### Optional
-1. `-b, --build-file` location of the build file (default ./jib.yaml)
-1. `--credHelper<registry>=<credHelper>` the credential helper to use for a registry
-1. `-P<name>=<value>, `, templating parameters replace `${name}` with `value` in the build file
-1. `--tags` additional tags for target (optional)
-1. `-v, --version` print version information and exit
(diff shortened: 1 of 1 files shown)


proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,62 @@
+# Jib CLI Surface Specification
+
+Specification for a command structures exposed to users of the jib-cli
+
+## Usage
+`jib command [OPTIONS]`
+
+## Options
```
mpeddada1, line 8: super nit: maybe switch the ordering of the "Commands" section and "Options" section to match the format in the usage section

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,62 @@
...
+```
+-t  --target <image-ref>  a target image reference
+```
+
+### Optional
+```
+-b  --build-file <file>                    location of the build file (default ./jib.yaml)
+    --credHelper <registry>=<credHelper>   credential helper to use for a registry (repeatable)
+    --docker-config <directory>            location of docker configuration
+    --help                                 print usage
```
mpeddada1, line 20: Are these indentations intended?

loosebazooka: yeah, generally following the format of some command `--help` (like `ls --help`) although I think `java --help` has a slightly different style. Anyway, the indentations are intentional, I can clean it up a little more though

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,60 @@
...
+```
+-b, --build-file <file>                    location of the build file (default <context>/jib.yaml)
+-c, --context <dir>                        location of the build context (default .)
+    --credHelper <registry>=<credHelper>   credential helper to use for a registry (repeatable)
+    --docker-config <directory>            location of docker configuration
+    --help                                 print usage
+-p, --parameter <name>=<value>             templating parameters replace `${name}` with `value` in the build file (repeatable)
+    --tag <tag1>[,<tag2>,...]              additional tags for target
+    --verbosity <level>                    set logging verbosity (error, warn, lifecycle (default), info, debug)
+-v, --version                              print version information and exit
```
mpeddada1, line 34: Is this the general jib-cli version?

mpeddada1, line 34: Is this the general jib-cli version?

loosebazooka: yeah this would be whatever version information we bake into the system.

mpeddada1: nice. we could add this maybe?
```suggestion
-v, --version                              print jib-cli version information and exit 
```

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Options
+
+### Required
+```
+-t  --target <image-ref>  a target image reference
+```
+
+### Optional
+```
```
mpeddada1, line 25: This is just a thought and also very much optional: How about having  a separate section for "general purpose flags" such as `help`, `version` and `verbosity`?

proposals/jib-cli-surface.md
loosebazooka: I think when you run `tool --version`, it's expected that you are expecting the version of the tool, no?

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,60 @@
...
+## Usage
+`jib COMMAND [OPTIONS]`
+
+## Commands
+
+The jib cli exposes the following top level commands, mirroring what is available in the plugins
+```
+build        build to registry
+dockerBuild  build to docker daemon
+buildTar     build to a tar file
```
chanseokoh, line 14: Just food for thoughts:

We have had multiple instances of feedback to be able to "warm up a cache" without pushing or creating a tar. Honestly, I wish the Jib build plugins had a way. And if we support the URI style, I think there's no need to have three commands.

Related, we may considering providing an option to configure the application layer cache (the default is to use a temporary directory) and the base image cache, but I'm not sure how useful they will be in practice.

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,60 @@
...
+### Required
+```
+-t  --target <image-ref>  a target image reference
+```
+
+### Optional
+```
+-b, --build-file <file>                    location of the build file (default <context>/jib.yaml)
+-c, --context <dir>                        location of the build context (default .)
+    --credHelper <registry>=<credHelper>   credential helper to use for a registry (repeatable)
```
chanseokoh, line 28: No simple way to pass a username and password pair if there's no credential helper or a Docker CLI? However, I don't know what's a conventional way to safely pass credentials for a CLI tool.

Anyways, `--allow-insecure` or `--insecure-registries=gcr.io,ghcr.io,...` sounds like a useful option. (Unfortunately, jib-core doesn't support per-registry configuration.)

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,71 @@
...
+-p, --parameter <name>=<value>             templating parameters replace `${name}` with `value` in the build file (repeatable)
+    --tag <tag1>[,<tag2>,...]              additional tags for target
+```
+
+#### Auth/Security
+```
+    --allow-insecure-registries            allow jib to communicate with registries over https
+    --credHelper <registry>=<credHelper>   credential helper to use for a registry (repeatable)
+    --send-credentials-over-http           allow jib to send credentials over http (used in conjunction with --allow-insecure-registries)
+    --to-auth <username> <password>        configure a username and password for authentication on the registry that an image is being built to
```
loosebazooka, line 45: I think maybe I might split this into
```
--to-auth-username <username>
--to-auth-password <password>
```
this lets us take advantage of some password related tools that pico cli adds

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,74 @@
...
+    --docker-config <directory>            location of docker configuration
+    --name <image-ref>                     image name to bake into tar file (only available when using -t tar://...)
+-p, --parameter <name>=<value>             templating parameters replace `${name}` with `value` in the build file (repeatable)
+    --tag <tag1>[,<tag2>,...]              additional tags for target
+```
+
+#### Auth/Security
+```
+    --allow-insecure-registries            allow jib to communicate with registries over https
+    --credHelper <registry>=<credHelper>   credential helper to use for a registry (repeatable)
```
loosebazooka, line 43: maybe this should be --to-credHelper

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,101 @@
...
+#### Auth/Security
+```
+    --allow-insecure-registries            allow jib to communicate with registries over https
+    --send-credentials-over-http           allow jib to send credentials over http (used in conjunction with --allow-insecure-registries)
+```
+##### Credentials
+
+Credentials can be specified using credential helpers or username + password. The following options are available
+```
+    --credHelper <credHelper>              credential helper to use for a registry
```
loosebazooka, line 51: this is all wrong, credential helpers already encode the server url, we do no need to,from,etc for cred helpers

chanseokoh: Some part of the general Docker auth has bothered me a bit. For example, whether using a credential helper or `docker login`, it didn't seem like you can provide different usernames for the base and the target repos if the repositories are on the same registry. That is, say you have two private Docker Hub repos `alice/image` and `bob/image` (i.e., repos in two different accounts), then you have to repeatedly do `docker login -u alice` and `docker login -u bob` if you need to work on both repos. Using a credential helper doesn't help either; FWIW, credentials are scoped solely by the registry.

Maybe it's just rare in practice that one has two different login accounts on the same registry? Anyways, I think having a single credential helper makes sense. Theoretically feasible, but I think it's unlikely that someone has multiple credential helpers set up only to return different accounts for the same registry.

proposals/jib-cli-surface.md
loosebazooka: Added a warm up cache section at the end, presumably that would be a separate command? we'd have to provide a shortcut in jib-core stepsrunner to do it though

proposals/jib-cli-surface.md
loosebazooka: Added some more flags here to cover these

proposals/jib-cli-surface.md
loosebazooka: Yeah, I can see that, we can always add in to/from later. It seems strange for now to do it though.

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,90 @@
...
+                          tar:///aboslute/path.tar
+```
+
+### Optional
+
+
+#### Build Config
+```
+    --application-cache <directory>        location of the application cache
+    --base-image-cache <directory>         location of the base image cache
```
loosebazooka, line 33: define default?

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,90 @@
...
+-b, --build-file <file>                    location of the build file (default <context>/jib.yaml)
+-c, --context <dir>                        location of the build context (default .)
+    --docker-config <directory>            location of docker configuration
+    --name <image-ref>                     image name to bake into tar file (only available when using -t tar://...)
```
loosebazooka, line 37: is this required or optional?

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,90 @@
...
+1. `--username`, `--password`
+1. `--to-username`, `--to-password`
+1. `--from-username`, `--from-password`
+1. `--to-username`, `--to-password`, `--from-username`, `--from-password`
+
+
+#### Info Params
+```
+    --help                                 print usage and exit
+    --stacktrace                           print stacktrace on error (for debugging issues in the jib-cli)
```
loosebazooka, line 70: should be hidden in help?

loosebazooka: also enable http tracing? (default level = config, output = console)

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,90 @@
...
+    --verbosity <level>                    set logging verbosity (error, warn, lifecycle (default), info, debug)
+-v, --version                              print version information and exit 
+```
```
loosebazooka, line 73: --serialize also hidden

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Environment Variables
+
+### Docker daemon specification
+
+Information on communicating with the docker daemon is determined using the `DOCKER_*` environment variables.
+
+## Other considerations
+
+- Users have asked to warm up the cache, perhaps we can have a cache build?
```
loosebazooka, line 87: jib core needs to support this?

proposals/jib-cli-surface.md
loosebazooka: Name is required for building "tar" images (https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java#L104)

proposals/jib-cli-surface.md
loosebazooka: I think there's some work to be done here to define a better default, but the default that jib-core uses is a temporary directory.

loosebazooka: added to "Other Considerations" section

proposals/jib-cli-surface.md
loosebazooka: updated text

proposals/jib-cli-surface.md
```diff
@@ -0,0 +1,96 @@
...
+    --send-credentials-over-http           allow jib to send credentials over http (used in conjunction with --allow-insecure-registries)
+```
+##### Credentials
+
+Credentials can be specified using credential helpers or username + password. The following options are available
+```
+    --credential-helper <credHelper>       credential helper to use for a registry, a path or name suffix (docker-credential-<suffix>) (repeatable)
+
+    --username <username>                  configure a username for authenticating against registries
+    --password <password>                  configure a passowrd for authenticating against registries (interactive if <password> is omitted)
```
mpeddada1, line 54:
```suggestion
    --password <password>                  configure a password for authenticating against registries (interactive if <password> is omitted)
```

mpeddada1 (approved):

PR #2791, opened 2020-09-29
Fixes running select integration tests from intellij

loosebazooka:
The intellij test runner has some issues with tests that are run
if integrationTest depends on test. Remove that dependency and
this fixes the issue.

Also adds some additional logging output and update release scripts

### DIFF
--- a/jib-build-plan/scripts/prepare_release.sh
+++ b/jib-build-plan/scripts/prepare_release.sh
@@ -39,8 +39,8 @@ if [[ $(git status -uno --porcelain) ]]; then
   Die 'There are uncommitted changes.'
 fi
 
-# Runs integration tests.
-./gradlew :jib-build-plan:integrationTest --info --stacktrace
+# Runs checks and integration tests.
+./gradlew :jib-build-plan:check :jib-build-plan:integrationTest --info --stacktrace
 
 # Checks out a new branch for this version release (eg. 1.5.7).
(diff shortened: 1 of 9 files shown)


jib-build-plan/scripts/prepare_release.sh
```diff
@@ -39,8 +39,8 @@ if [[ $(git status -uno --porcelain) ]]; then
   Die 'There are uncommitted changes.'
 fi
 
-# Runs integration tests.
-./gradlew :jib-build-plan:integrationTest --info --stacktrace
+# Runs checks and integration tests.
+./gradlew :jib-build-plan:check :jib-build-plan:integrationTest --info --stacktrace
```
chanseokoh, line 43: Quick question: `kokoro/continuous.sh` runs `build` and `integrationTest`. What's the difference between `build` and `check`?

chanseokoh (approved):

jib-build-plan/scripts/prepare_release.sh
loosebazooka: yeah I should've mentioned, `build`  already runs `check` (but not `integrationTest`, we always have had integration test be explicit)

loosebazooka: `check` runs all tests, checkstyle and format things

PR #2793, opened 2020-09-30
Check configured platforms for single image building

### LINKED ISSUE #2745: Investigate what to do if the platform of a local Docker or tar base image doesn't match the specified <platform>
Basically #2746 (Investigate what to do if the base image platform doesn't match the configured or default platform for single-platform image building, but this is to track the cases of using local Docker and tar images.

Related: #2744 (Investigate if it's feasible to use local Docker or tar base image when building multi-platform images)

### LINKED ISSUE #2746: Investigate what to do if the base image platform doesn't match the configured or default platform for single-platform image building
As of now, if the base image reference is not a manifest list but a manifest, we always take the manifest. That is, configured platforms (as well as the default `amd64/linux` when missing) are ignored and there is no warning even if the actual platform doesn't match the configured one.

However, probably we shouldn't show a mismatch warning at the current state where we cannot distinguish the default configuration from an explicit configuration. The user may intentionally specify a base image ref for a non-`amd64/linux` manifest, and we don't want to force the user to always specify `<platform>`. But at least we can definitely show a warning when the user configured multiple platforms and the base image ref was not a manifest list.

Related: #2745 is a subset of this issue whose scope is limited to a local Docker or tar image.

FTR: https://github.com/GoogleContainerTools/jib/pull/2670#discussion_r466622742

chanseokoh:
Fixes #2746. Fixes #2745.

When the base image is not a manifest list (including a local Docker and tar image), logs a warning in the follow cases.
1. Multiple platforms are configured.
2. The configured platform doesn't match the actual platform, unless the "configured" platform is `amd64/linux`.

About 2): Currently we cannot distinguish between explicitly configured `amd64/linux` and the implicit `amd64/linux` default when omitted, so unfortunately sometimes we need to suppress logging even if the platforms don't match.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PlatformChecker.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PlatformChecker.java
@@ -46,7 +46,8 @@ public class PlatformChecker {
 
     if (platforms.size() != 1) {
       eventHandlers.dispatch(
-          LogEvent.warn("platforms configured, but '" + baseImageName + "' is not a manifest list"));
+          LogEvent.warn(
+              "platforms configured, but '" + baseImageName + "' is not a manifest list"));
     } else {
       Platform platform = platforms.iterator().next();
       if (!platform.getArchitecture().equals(containerConfig.getArchitecture())
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PlatformChecker.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PlatformChecker.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.jib.builder.steps;
 
-import com.google.cloud.tools.jib.api.ImageReference;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.api.buildplan.Platform;
 import com.google.cloud.tools.jib.configuration.BuildContext;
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -379,10 +380,14 @@ private ContainerConfigurationTemplate pullContainerConfigJson(
         return Collections.singletonList(
             JsonToImageTranslator.toImage((V21ManifestTemplate) manifest));
       }
+
+      ContainerConfigurationTemplate containerConfig =
+          Verify.verifyNotNull(manifestsAndConfigs.get(0).getConfig());
+      PlatformChecker.checkManifestPlatform(buildContext, containerConfig);
```
chanseokoh, line 386: When using a cached manifest for a remote base image.

chanseokoh:
Ready for review.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PlatformChecker.java
```diff
@@ -0,0 +1,75 @@
...
+          LogEvent.warn(
+              "platforms configured, but '" + baseImageName + "' is not a manifest list"));
+    } else {
+      Platform platform = platforms.iterator().next();
+      if (!platform.getArchitecture().equals(containerConfig.getArchitecture())
+          || !platform.getOs().equals(containerConfig.getOs())) {
+
+        // Unfortunately, "platforms" has amd64/linux by default even if the user didn't explicitly
+        // configure it. Skip reporting to suppress false alarm.
+        if (!(platform.getArchitecture().equals("amd64") && platform.getOs().equals("linux"))) {
```
mpeddada1, line 58: Just making sure I'm understanding this correctly, so this check needs to be included because we don't have a way to differentiate between implicitly or explicitly configured platforms?

chanseokoh: Yes, when it's a single platform of `amd64/linux`. That is, even if the user doesn't configure anything, `platforms` is a singleton set with `amd64/linux`. If you think the code comment on line 56-57 can be improved, feel free to suggest an edit.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PlatformChecker.java
```diff
@@ -0,0 +1,75 @@
...
+
+/** Provides helper methods to check platforms. */
+public class PlatformChecker {
+
+  /**
+   * Assuming the base image is not a manifest list, checks and warns misconfigured platforms.
+   *
+   * @param buildContext the {@link BuildContext}
+   * @param containerConfig container configuration JSON of the base image
+   */
```
mpeddada1, line 35: Is there a way to check that the build image is in fact not a manifest list in this method, before we proceed to doing the other checks?

chanseokoh: Yeah, I had the same question. I know this is a bit weird. It's mainly the consequence of factoring out some portion of code to reuse that in multiple places under certain conditions. But I think it's also weird to just pass something unnecessary to the method only to check the argument.

mpeddada1: Yup, that's a valid point since it's meant to serve as a helper. As long as we check the condition before calling this method, we should be good.

mpeddada1 (commented):
Looks good to me.

mpeddada1 (approved):

Issue #2796, opened 2020-09-30
Jib CLI Jar Processing Tracking Bug


mpeddada1:
While the jib cli is meant to allow users to containerize any file without having to integrate jib into java build tools (maven and gradle), this particular cli command will be used to **containerize a jar file**.
Specifically:
    - It is aimed at requiring minimal input from users in order to make the jar workable once it is on the image. 
    - It will make inferences about the jar application entrypoint and will optimize the primary jar into either three layers (class 
      files, non *.class files and dependencies) or two layers (jar, dependencies). This will give users the ability to containerize 
      their jar files while still maintaining the original intent of jib in providing granularity.

Rollout Plan: 
- [x] Implement first part of JarProcessor to determine the type of the jar provided.
- [ ] Implement explodedMode for regular jars.
- [ ] Implement packagedMode for regular jars.
- [ ] Implement explodedMode for Spring Boot fat jars.
- [ ] Implement packagedMode for Spring Boot fat jars. 
- [ ] Add other image parameters to JarInfoBuilder.

loosebazooka:
So the more I think about this, the more each of these bugs should be separate and tracked together as a project.

mpeddada1:
That's a nice idea. Closing this tracking bug since [Jib Cli Jar Processing Project](https://github.com/GoogleContainerTools/jib/projects/11) has been created.

PR #2797, opened 2020-09-30
Implement JarProcessor to determine jar type.

mpeddada1:
Part 1 of #2796.

Determine whether the jar is a Spring Boot fat jar or regular jar, given the path to the jar file. Note that this method doesn't take into account, Spring Boot thin jars or other types of fat jars yet. If the jar is not a Spring Boot jar then it is considered a regular jar. 

This implementation will use Truth (https://truth.dev/) for testing.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+   * @param jarPath path to the jar.
+   * @return the jar type.
+   * @throws IOException if I/O error occurs when opening the file.
+   */
+  public static JarType determineJarType(Path jarPath) throws IOException {
+    JarFile jarFile = new JarFile(jarPath.toString());
+    if (jarFile.getEntry("BOOT-INF") != null) {
+      return JarType.SPRINGBOOT;
+    }
+    return JarType.REGULAR;
```
mpeddada1, line 50: May need to add more logic to verify that the jar is in fact a regular jar, otherwise throw an exception saying that non-springboot fat jars are not supported.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.jar.JarFile;
+
+public class JarProcessor {
+
+  /**
+   * Jar Type.
+   *
+   * <ul>
+   *   <li>{@code REGULAR} a regular jar.
+   *   <li>{@code SPRINGBOOT} a springboot fat jar. *
```
chanseokoh, line 30:
```suggestion
   *   <li>{@code SPRING_BOOT} a springboot fat jar.
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Determines whether the jar is a spring-boot or regular jar, given a path to the jar.
+   *
+   * @param jarPath path to the jar.
+   * @return the jar type.
+   * @throws IOException if I/O error occurs when opening the file.
+   */
+  public static JarType determineJarType(Path jarPath) throws IOException {
+    JarFile jarFile = new JarFile(jarPath.toString());
```
chanseokoh, line 46: nit: it doesn't matter at all here, but generally I try to avoid dealing with path as a string.
```suggestion
    JarFile jarFile = new JarFile(jarPath.toFile());
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.common.io.Resources;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.junit.Test;
+
+public class JarProcessorTest {
+
+  private static final String SPRINGBOOT_RESOURCE_DIR = "jar/springboot";
```
chanseokoh, line 31:
```suggestion
  private static final String SPRING_BOOT_RESOURCE_DIR = "jar/springboot";
```
And maybe this can just be `jar/springboot/springbootJar.jar`?

chanseokoh (commented):
Looks good to me.

I'd like to avoid adding a jar of 46.8 MB. Too big for the value we get. I think we can create a small fake, containing only `BOOT-INF`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Jar Type.
+   *
+   * <ul>
+   *   <li>{@code REGULAR} a regular jar.
+   *   <li>{@code SPRING_BOOT} a springboot fat jar.
+   * </ul>
+   */
+  public enum JarType {
+    REGULAR,
```
loosebazooka, line 34: I think `STANDARD` is probably the more standard term?

mpeddada1: Done.

loosebazooka (approved):

PR #2802, opened 2020-10-05
Implement JarInfo for cli jar command to store image parameters.

mpeddada1:
Part of #2796
JarInfo will hold all the image parameters derived from [JarProcessor](https://github.com/GoogleContainerTools/jib/blob/c2081332a1e95ca04c5e46469b435380a2286f27/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java#L23). The implementation is based on [ContainerBuildPlan](https://github.com/GoogleContainerTools/jib/blob/212cc8a372737cce901e01b6ae6b395ff77283de/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java#L33).


mpeddada1:
ContainerBuildPlan essentially does the same thing so don't need this intermediate class right now.

PR #2813, opened 2020-10-09
Implement explode mode for regular jar.

mpeddada1:
This will be used to explode a jar and create three layers on the container for classes, resources and dependencies. 
Part of #2796


chanseokoh:
There's an interesting failure. The test seems non-deterministic, affecting image reproducibility?
```
com.google.cloud.tools.jib.cli.jar.JarProcessorTest > testExplodeMode_standard FAILED
    contents match, but order was wrong
    expected: [/app/explodedJar/META-INF, /app/explodedJar/class5.class, /app/explodedJar/directory1, /app/explodedJar/directory1/class1.class, /app/explodedJar/directory1/class2.class, /app/explodedJar/directory2, /app/explodedJar/directory2/class4.class, /app/explodedJar/directory2/directory3, /app/explodedJar/directory2/directory3/class3.class, /app/explodedJar/directory4]
    but was : [/app/explodedJar/class5.class, /app/explodedJar/directory1, /app/explodedJar/directory1/class1.class, /app/explodedJar/directory1/class2.class, /app/explodedJar/directory2, /app/explodedJar/directory2/class4.class, /app/explodedJar/directory2/directory3, /app/explodedJar/directory2/directory3/class3.class, /app/explodedJar/directory4, /app/explodedJar/META-INF]
        at com.google.cloud.tools.jib.cli.jar.JarProcessorTest.testExplodeMode_standard(JarProcessorTest.java:73)
```

mpeddada1 (commented):
Oh oops, let me try fixing this.

chanseokoh:
The last commit fixes the test, but I think it adversely hides the problem in the prod code. The earlier failure was due to a different order of files in the layer, which is important and affects reproducibility. I'd say it was correct that the test was checking the order of files. We need to identify the cause of the non-deterministic ordering.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -204,7 +204,7 @@ public static JavaContainerBuilder from(TarImage tarImage) {
   private RelativeUnixPath dependenciesDestination = RelativeUnixPath.get("libs");
   private RelativeUnixPath othersDestination = RelativeUnixPath.get("classpath");
   @Nullable private String mainClass;
-  private ModificationTimeProvider modificationTimeProvider =
+  private static ModificationTimeProvider modificationTimeProvider =
```
chanseokoh, line 207: This should remain an instance variable (non-`static`), scoped and limited only to each individual instance. `static` means a global variable shared by every instance of `JavaContainerBuilder`.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -691,7 +700,18 @@ private void addFileToLayer(
...
+   *
+   * @param layerBuilders map that keeps track of all layers, with the key being the layer type and
+   *     value being a {@link FileEntriesLayer.Builder}.
+   * @param layerType the layer type.
+   * @param sourceRoot path to root directory.
+   * @param pathFilter predicate to filter the files by.
+   * @param basePathInContainer base path in container.
+   * @throws IOException if I/O error occurs when opening the directory.
+   */
+  public static void addDirectoryContentsToLayer(
```
chanseokoh, line 714: Should remain an instance method.

And note `JavaContainerBuilder` is a public API. Probably we don't want to expose this internal helper method.

Just asking: did we decide to use `JavaContainerBuilder` instead of `JibContainerBuilder` for some reason? It's fine if `JavaContainerBuilder` works, but just saying I believe it's a bit curated and limiting.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -679,7 +679,16 @@ private JavaContainerBuilder addDirectory(
...
+  /**
+   * Adds a file to a {@link FileEntriesLayer.Builder}.
+   *
+   * @param layerBuilders map that keeps track of all layers, with the key being the layer type and
+   *     value being a {@link FileEntriesLayer.Builder}.
+   * @param layerType the layer type.
+   * @param sourceFile the source file path.
+   * @param pathInContainer path in container.
+   */
+  public static void addFileToLayer(
```
chanseokoh, line 691: Ditto.

We need to think twice before making this part of the public API surface.

chanseokoh:
Hmm... it's not immediately clear how the order could be non-deterministic. A map is usually suspected, but I believe `ImmutableMap` guarantees the given order when iterating.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
mpeddada1: Thanks for pointing this out! No longer using JavaContainerBuilder in JarProcessor.

mpeddada1:
It seems like `stream#sorted()` was displaying different behavior for windows and macos when we were sorting on `Path`. This issue should be fixed now.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +66,88 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+
+  /**
+   * Explode jar and create three layers for classes, resources and dependencies on container.
+   *
+   * @param jarPath path to jar file.
+   * @return list of {@link FileEntriesLayer}.
+   * @throws IOException if I/O error occurs when opening the file.
+   */
+  public static List<FileEntriesLayer> explodeStandardJar(Path jarPath) throws IOException {
+    TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider();
```
loosebazooka, line 78: I think this should be configurable and if not then use a temporary directory.

Potentially we could have a jib local build directory that we can pass in.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +66,88 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+                isResourceFile,
+                APP_ROOT.resolve(RelativeUnixPath.get("explodedJar")))
+            .setName("Resources")
+            .build();
+
+    // Get dependencies from Class-Path in the jar's manifest and create a FileEntriesLayer.Builder
+    // with these dependencies as entries.
+    List<Path> dependencies = new ArrayList<>();
+    JarFile jarFile = new JarFile(jarPath.toFile());
+    String classPath = jarFile.getManifest().getMainAttributes().getValue("Class-Path");
```
loosebazooka, line 110: `"Class-Path"` -> `Attributes.Name.CLASS_PATH`

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +66,88 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+    if (classPath != null) {
+      for (String dependency : Splitter.on(" ").split(classPath)) {
```
loosebazooka, line 112: `" "` -> `"//s+"` for handling any kind of white space
I might also do `classPath.trim()` ?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +66,88 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+        Path depPath = Paths.get(dependency);
+        dependencies.add(depPath);
+      }
+    } else {
+      throw new IllegalStateException("Class path is not specified.");
```
loosebazooka, line 117: Do we have to fail here? Would this code path also handle the shaded jar style jars (which may not have `Class-Path` entries)?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
chanseokoh: May be a good idea. For example, [here](https://stackoverflow.com/questions/3057841/too-long-line-in-manifest-file-while-trying-to-create-jar), I see people saying like

> two spaces immediately following `Class-Path:`, a space at the end of [each] line, and one before each line, and finally an extra new-line at the end did the trick.

although I'm not sure the Java API handles all these whitespaces correctly. Worth checking anyway.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +64,85 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+
+    // Get dependencies from Class-Path in the jar's manifest and create a FileEntriesLayer.Builder
+    // with these dependencies as entries.
+    List<Path> dependencies = new ArrayList<>();
+    JarFile jarFile = new JarFile(jarPath.toFile());
+    String classPath = jarFile.getManifest().getMainAttributes().getValue("Class-Path");
+    if (classPath != null) {
+      for (String dependency : Splitter.on(" ").split(classPath)) {
+        Path depPath = Paths.get(dependency);
+        dependencies.add(depPath);
```
chanseokoh, line 112: super nit:

```suggestion
        dependencies.add(Paths.get(dependency));
```

Just FYI, another way is
```
.splitToList().stream().map(Paths::get).collect(Collectors.toList());
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +64,85 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+    FileEntriesLayer.Builder dependenciesLayerBuilder = FileEntriesLayer.builder();
+    dependencies.forEach(
+        path ->
+            dependenciesLayerBuilder.addEntry(
+                path, APP_ROOT.resolve(RelativeUnixPath.get("dependencies")).resolve(path)));
+    dependenciesLayerBuilder.setName("Dependencies");
+
+    layers.add(classesLayer);
+    layers.add(resourcesLayer);
+    layers.add(dependenciesLayerBuilder.build());
```
chanseokoh, line 126: I believe we customarily put the most frequently changing layers at the bottom (needs to double-check). So perhaps

```suggestion
    layers.add(dependenciesLayerBuilder.build());
    layers.add(resourcesLayer);
    layers.add(classesLayer);
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -18,17 +18,24 @@
...
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Collectors;
 import org.junit.Test;
 
 public class JarProcessorTest {
 
   private static final String SPRING_BOOT_RESOURCE_DIR = "jar/springboot/springboot_sample.jar";
+  private static final String STANDARD_RESOURCE_DIR_WITH_CP = "jar/standard/standardJarWithCp.jar";
```
chanseokoh, line 38:
```suggestion
  private static final String STANDARD_JAR_WITH_CLASS_PATH_MANIFEST = "jar/standard/standardJarWithCp.jar";
```

Also applies to other constants.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -44,4 +51,77 @@ public void testDetermineJarType_standard() throws IOException, URISyntaxExcepti
...
+
+  @Test
+  public void testExplodeMode_standard() throws IOException, URISyntaxException {
+    Path standardJar = Paths.get(Resources.getResource(STANDARD_RESOURCE_DIR_WITH_CP).toURI());
+    List<FileEntriesLayer> layers = JarProcessor.explodeStandardJar(standardJar);
+    FileEntriesLayer classesLayer = layers.get(0);
+    FileEntriesLayer resourcesLayer = layers.get(1);
+    FileEntriesLayer dependenciesLayer = layers.get(2);
+
+    assertThat(layers.size()).isEqualTo(3);
```
chanseokoh, line 63: Assert before attempting `layers.get(2)`. Otherwise, for example, if the code base accidentally changes to produce only two layer, we'll hit an index-out-of-bound error when attempting `.get(2)`.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -44,4 +51,77 @@ public void testDetermineJarType_standard() throws IOException, URISyntaxExcepti
...
+
+    // Validate classes layer.
+    // TODO: Validate order of file paths once
+    // https://github.com/GoogleContainerTools/jib/issues/2821 is fixed.
+    assertThat(classesLayer.getEntries().size()).isEqualTo(10);
```
chanseokoh, line 68: I guess no need to assert the size when using `.containsExactly`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -44,4 +51,77 @@ public void testDetermineJarType_standard() throws IOException, URISyntaxExcepti
...
+            AbsoluteUnixPath.get("/app/explodedJar/directory2"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory2/class4.class"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory2/directory3"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory2/directory3/class3.class"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory4"));
+
+    // Validate resources layer.
+    // TODO: Validate order of file paths once
+    // https://github.com/GoogleContainerTools/jib/issues/2821 is fixed.
+    assertThat(resourcesLayer.getEntries().size()).isEqualTo(9);
```
chanseokoh, line 92: Ditto?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -44,4 +51,77 @@ public void testDetermineJarType_standard() throws IOException, URISyntaxExcepti
...
+            AbsoluteUnixPath.get("/app/explodedJar/directory1"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory1/resource1.txt"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory2"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory2/directory3"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory2/directory3/resource2.sql"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory4"),
+            AbsoluteUnixPath.get("/app/explodedJar/directory4/resource3.txt"));
+
+    // Validate dependencies layer.
+    assertThat(dependenciesLayer.getEntries().size()).isEqualTo(3);
```
chanseokoh, line 113: Ditto?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +64,85 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+   * @throws IOException if I/O error occurs when opening the file.
+   */
+  public static List<FileEntriesLayer> explodeStandardJar(Path jarPath) throws IOException {
+    TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider();
+    Path tempDirectoryPath = tempDirectoryProvider.newDirectory();
+    ZipUtil.unzip(jarPath, tempDirectoryPath);
+
+    List<FileEntriesLayer> layers = new ArrayList<>();
+    Predicate<Path> isClassFile = path -> path.getFileName().toString().endsWith(".class");
+    Predicate<Path> isResourceFile = path -> !path.getFileName().toString().endsWith(".class");
```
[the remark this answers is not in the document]
loosebazooka: Oh this is cool.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: That's a good question. I didn't intend for this code path to handle shaded jars. It mainly focuses on regular jars, which require the class-path to be fully specified in order for it to work. We could try to intelligently determine what the dependencies are in the future implementations, but for now I decided to first check that the Class-Path was present before proceeding to the next steps. What do you think?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
mpeddada1: That's a good idea!

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Pretty neat.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +64,85 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+
+    // Determine class and resource files in the directory containing jar contents and create
+    // FileEntriesLayer for each type of layer (class or resource), while maintaining the
+    // file's original project structure.
+    FileEntriesLayer classesLayer =
+        addDirectoryContentsToLayer(
+                FileEntriesLayer.builder(),
+                tempDirectoryPath,
```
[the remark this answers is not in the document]
mpeddada1: Done:)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
chanseokoh: A regular jar includes one that doesn't have Class-Path. It's possible that many people will try a hello-world app (i.e, no dependencies), so I think we shouldn't fail.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Done:)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Oh great point, added handling for any kind of white space and added extra spaces in the class-path field in the sample jar file  to make sure this is working.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Thanks for pointing this out.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Oh that's a very valid point! Okay I will remove this exception.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +67,97 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+   * @throws IOException if I/O error occurs when opening the jar file or if the directory for the
+   *     temporary directory path provided, doesn't exist.
+   */
+  public static List<FileEntriesLayer> explodeStandardJar(Path jarPath, @Nullable Path tempDirPath)
+      throws IOException {
+    Path localExplodedJarRoot;
+    if (tempDirPath == null) {
+      try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
+        localExplodedJarRoot = tempDirectoryProvider.newDirectory();
+      }
```
chanseokoh, line 86: Once a `Closeable` class is closed, it's permanently shutdown and not usable. So you should put the entire code that needs to use `localExplodedJarRoot` into the try-resource block.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +67,97 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+    // Determine class and resource files in the directory containing jar contents and create
+    // FileEntriesLayer for each type of layer (class or resource), while maintaining the
+    // file's original project structure.
+    FileEntriesLayer classesLayer =
+        addDirectoryContentsToLayer(
+                FileEntriesLayer.builder(),
+                localExplodedJarRoot,
+                isClassFile,
+                APP_ROOT.resolve(RelativeUnixPath.get("explodedJar")))
+            .setName("Classes")
```
chanseokoh, line 105: Minor point, but Jib build plugins use non-capitalized name "classes" (as defined in `JavaContainerBuilder.LayerType`).

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +67,97 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+            .build();
+    FileEntriesLayer resourcesLayer =
+        addDirectoryContentsToLayer(
+                FileEntriesLayer.builder(),
+                localExplodedJarRoot,
+                isResourceFile,
+                APP_ROOT.resolve(RelativeUnixPath.get("explodedJar")))
+            .setName("Resources")
```
chanseokoh, line 113: ditto

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -18,17 +18,25 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
 import org.junit.Test;
 
 public class JarProcessorTest {
 
   private static final String SPRING_BOOT_RESOURCE_DIR = "jar/springboot/springboot_sample.jar";
+  private static final String STANDARD_JAR_WITH_CLASS_PATH_MANIFEST =
+      "jar/standard/standardJarWithClassPath.jar";
   private static final String STANDARD_RESOURCE_DIR = "jar/standard/standardJar.jar";
```
chanseokoh, line 40:
```suggestion
  private static final String SPRING_BOOT_JAR = "jar/springboot/springboot_sample.jar";
  private static final String STANDARD_JAR_WITH_CLASS_PATH_MANIFEST =
      "jar/standard/standardJarWithClassPath.jar";
  private static final String STANDARD_JAR = "jar/standard/standardJar.jar";
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +67,97 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+        jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
+    if (classPath != null) {
+      dependencies =
+          Splitter.onPattern("\\s+")
+              .splitToList(classPath.trim())
+              .stream()
+              .map(Paths::get)
+              .collect(Collectors.toList());
+    } else {
+      dependencies = new ArrayList<>();
```
chanseokoh, line 130: You can remove this `else` branch if you initialize the variable on line 118 like `List<Path> dependencies = new ArrayList<>();`.

How about skipping adding the dependencies layer if it's empty? To be fair, it shouldn't matter, because I believe jib-core will drop any empty layers. But we can be conservative at this level too.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -35,6 +54,29 @@
     SPRING_BOOT;
   }
 
+  /** Represents the different types of layers. */
+  @VisibleForTesting
+  public enum ExplodedModeLayerType {
```
chanseokoh, line 59: I think an enum is unnecessary if this is only to define layer names. It looks okay to me to use the string literals in the code as before.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: That's good to know. Modified this to not add a dependency layer if there are no dependencies specified in `Class-Path`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: I added an enum for layer types, similar to what's being done in `JavaContainerBuilder`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +91,92 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+   * @param jarPath path to jar file.
+   * @param tempDirPath path to temporary jib local directory to use.
+   * @return list of {@link FileEntriesLayer}.
+   * @throws IOException if I/O error occurs when opening the jar file or if the directory for the
+   *     temporary directory path provided doesn't exist.
+   */
+  public static List<FileEntriesLayer> explodeStandardJar(Path jarPath, @Nullable Path tempDirPath)
+      throws IOException {
+    Path localExplodedJarRoot = tempDirPath;
+    try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
```
chanseokoh, line 107: Oh! I realized there's a problem. When `TempDirectoryProvider` is closed, all the contents (i.e., exploded jar) will be cleaned up. That is, nothing left when exiting the method. However, tar creating happens far later in jib-core. I guess this method should just use `tempDirPath` (should not be `@Nullable`), unless the method opts to use a system temp directory. Since the method can accept `tempDirPath`, maybe we should just use it.

And how about `tempDirPath` --> `tempDirectory`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +91,92 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+      String classPath =
+          jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
+      if (classPath != null) {
+        List<Path> dependencies =
+            Splitter.onPattern("\\s+")
+                .splitToList(classPath.trim())
+                .stream()
+                .map(Paths::get)
+                .collect(Collectors.toList());
+        FileEntriesLayer.Builder dependenciesLayerBuilder = FileEntriesLayer.builder();
```
chanseokoh, line 150: Nit: can chain `...builder().setName("dependencies")`.

chanseokoh (commented):
Overall, looks good to me.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +65,88 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+   */
+  public static List<FileEntriesLayer> explodeStandardJar(Path jarPath, Path tempDirPath)
+      throws IOException {
+    Path localExplodedJarRoot = tempDirPath;
+
+    ZipUtil.unzip(jarPath, localExplodedJarRoot);
+
+    List<FileEntriesLayer> layers = new ArrayList<>();
+    Predicate<Path> isClassFile = path -> path.getFileName().toString().endsWith(".class");
+    Predicate<Path> isResourceFile = isClassFile.negate();
```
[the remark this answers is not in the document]
chanseokoh: That's an interesting question. If you remember, at least the Spring Boot launcher requires it. I think it reads the file from classpath. Maybe it's not a great harm to include it ATM and see if someone complains?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -18,30 +18,181 @@
...
     JarType jarType = JarProcessor.determineJarType(standardJar);
     assertThat(jarType).isEqualTo(JarType.STANDARD);
   }
+
+  @Test
+  public void testExplodeMode_standard_withClassPathInManifest()
+      throws IOException, URISyntaxException {
+    Path standardJar =
+        Paths.get(Resources.getResource(STANDARD_JAR_WITH_CLASS_PATH_MANIFEST).toURI());
+    File tempDirectory = Files.createTempDir();
```
chanseokoh, line 65: Any reason not using `@Rule public final TemporaryFolder` as we do in other tests?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +65,88 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+    // file's original project structure.
+    FileEntriesLayer classesLayer =
+        addDirectoryContentsToLayer(
+                FileEntriesLayer.builder(),
+                localExplodedJarRoot,
+                isClassFile,
+                APP_ROOT.resolve(RelativeUnixPath.get("explodedJar")))
+            .setName("classes")
+            .build();
+    FileEntriesLayer resourcesLayer =
```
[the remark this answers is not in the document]
mpeddada1: Great question. For a jar that doesn't contain resources, the resources layer will only contain the manifest.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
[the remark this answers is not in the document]
mpeddada1: Along those lines- I decided to keep it, going off of what we've seen with Spring Boot. I couldn't find a particular reason to intentionally remove it so I left in for the time being? Although, the main cost I see is creating an extra resources layer just for the manifest if we have a jar with only classes.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +65,86 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+      layers.add(dependenciesLayerBuilder.build());
+    }
+
+    layers.add(resourcesLayer);
+    layers.add(classesLayer);
+    return layers;
+  }
+
+  private static FileEntriesLayer.Builder addDirectoryContentsToLayer(
+      FileEntriesLayer.Builder builder,
```
chanseokoh, line 134: Sorry, looking at this again, maybe this doesn't need to accept a builder. It can create it locally. That said, it can accept a layer name instead? For example,
```java
private static FileEntriesLayer createLayerFromDirectoryContents(String layerName, ...) {
  FileEntriesLayer.Builder builder = FileEntriesLayer.builder().setName(layerName);
  new DirectoryWalker(...)...;
  return builder.build();
}
```
WDYT?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -18,30 +18,260 @@
...
     assertThat(jarType).isEqualTo(JarType.STANDARD);
   }
+
+  @Test
+  public void testExplodeMode_standard_emptyJar() throws IOException, URISyntaxException {
+    Path standardJar = Paths.get(Resources.getResource(STANDARD_JAR_EMPTY).toURI());
+    Path destDir = temporaryFolder.newFolder().toPath();
+    List<FileEntriesLayer> layers = JarProcessor.explodeStandardJar(standardJar, destDir);
+
+    // Validate that only two layers are created.
```
chanseokoh, line 70: super nit: I think this comment is obvious enough to be removed.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -49,4 +65,86 @@ public static JarType determineJarType(Path jarPath) throws IOException {
     }
     return JarType.STANDARD;
   }
+
+  /**
+   * Explode jar and create three layers for classes, resources and dependencies on container.
```
chanseokoh, line 70:
```suggestion
   * Explodes jar and create three layers for classes, resources and dependencies on container.
```
and we don't put a period for `@param`, `@return` and such, but honestly, I didn't care.

chanseokoh (commented):
LGTM

chanseokoh:
There a trivial merge conflict too.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
mpeddada1: True, done.

chanseokoh (approved):

PR #2816, opened 2020-10-12
Create Triage Party yaml for jib 

mpeddada1:
Usage instructions: 
* Steps to render the triage-party page are provided in https://github.com/google/triage-party#try-it.
* The only difference is that we will need to run the following command for this yaml file:
```export GITHUB_TOKEN="$(cat $HOME/.github-token)"
docker build --tag=tp .
docker run \
  --rm \
  -e GITHUB_TOKEN \
  -v "$PWD/jib_triage_party.yaml:/app/config/config.yaml" \
  -p 8080:8080 tp```

### DIFF
--- a/jib_triage_party.yaml
+++ b/jib_triage_party.yaml
@@ -17,16 +17,20 @@ settings:
   min_similarity: 0.8
   repos:
     - https://github.com/GoogleContainerTools/jib
+    - https://github.com/GoogleContainerTools/jib-extensions
+    - https://github.com/GoogleCloudPlatform/app-maven-plugin
+    - https://github.com/GoogleCloudPlatform/app-gradle-plugin
 
 collections:
   - id: daily
     name: Daily Triage
     dedup: false
     description: >
-      Triage-oncall zeroes out this queue once a day. The priorities are:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: jib_triage_party.yaml [hunk 1]
```diff
@@ -0,0 +1,754 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+settings:
+  name: jib
+  min_similarity: 0.8
+  repos:
+    - https://github.com/GoogleContainerTools/jib
```
loosebazooka, line 19: The only strange thing I feel about this is that the repos unrelated to jib are included in here. Should this config not reside in github?

loosebazooka (approved):
Lets just get this in so we can get started. We can update the config as we move on.

jib_triage_party.yaml
mpeddada1: hm yeah I feel a little strange about this as well.

mpeddada1:
Moving this file out of github.

PR #2830, opened 2020-10-15
Allow DefaultCredentialRetrievers to accept multiple user defined helpers

loosebazooka:
This will allow it to work with params that can be specified multiple times like `--credential-helper` on the cli


chanseokoh:
Oh, now you said it, I think we should have `--from-credential-helper` and `--to-credential-helper`. Having a single credential helper for every registry is an old way (e.g., `credsStore:` in `~/.docker/config.json`). Now people can configure different helper binaries for different registries with `credHelpers:`.

Apart for that, why do we want to be able to configure multiple cred helpers for a single registry? (`DefaultCredentialRetrievers` is meant for a single registry.)

loosebazooka:
I think if we use `--to` and `--from` we don't need this. So let me go update that and close this PR.

PR #2835, opened 2020-10-16
Proposal for Jib CLI Jar Processing

mpeddada1:
rendered: https://github.com/GoogleContainerTools/jib/blob/cli-jar-proposal/proposals/cli-jar-processing.md

### DIFF
--- a/proposals/cli-jar-processing.md
+++ b/proposals/cli-jar-processing.md
@@ -13,11 +13,12 @@ A standard jar can be containerized in two modes, exploded or packaged.
 ### Exploded Mode
 Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY}`
 The default mode for containerizing a jar. It will explode a jar into the following layers:  
-- Dependencies Layer: Contains dependencies derived from `Class-Path` in jar manifest. Note that this layer will not be created if `Class-Path` is not present in the manifest.
+- Dependencies Layer: Contains dependencies whose versions do not contain `SNAPSHOT`. Note that this layer will not be created if `Class-Path` is not present in the manifest.
+- Snapshot-Dependencies Layer: Contains dependencies whose versions contain `SNAPSHOT`. Note that this layer will not be created if `Class-Path` is not present in the manifest.
 - Resources Layer: Contains resources parsed from jar file. Note that it will also include `MANIFEST.MF`.
 - Classes Layer: Contains classes parsed from jar file. 
 
-**Entrypoint** : `java -cp /app/dependencies/ /app/explodedJar/ ${MAIN_CLASS}`
+**Entrypoint** : `java -cp /app/dependencies/:/app/explodedJar/ ${MAIN_CLASS}`
 
 ### Packaged Mode
 Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY} --mode packaged`.
@@ -39,7 +40,7 @@ The default mode for containerizing a jar. It will explode a jar according to wh
 - Resources Layer: Contains resources parsed from `BOOT-INF/classes/` in the jar and `META-INF/`.
 - Classes Layer: Contains classes parsed from `BOOT-INF/classes/` in the jar.
 
-**Entrypoint** : `java -cp . org.springframework.boot.loader.JarLauncher`
+**Entrypoint** : `java -cp /app org.springframework.boot.loader.JarLauncher`
 
 ### Packaged Mode
 Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY} --mode packaged`
--- a/proposals/cli-jar-processing.md
+++ b/proposals/cli-jar-processing.md
@@ -3,7 +3,7 @@
 Relevant Issue: [#2796](https://github.com/GoogleContainerTools/jib/issues/2796)
 
 # Motivation 
-Allow users to containerize arbitrary jar files without having to integrate Jib into java build tools (maven and gradle).
+Allow users to containerize arbitrary jar files without having to integrate Jib into java build tools (maven and gradle) or use a `.yaml` jib cli buildfile.
 
 # Proposal Changes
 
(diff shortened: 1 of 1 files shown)


proposals/cli-jar-processing.md
```diff
@@ -0,0 +1,48 @@
+# Proposal : Jib CLI Jar Processing 
+
+Relevant Issue: [#2796](https://github.com/GoogleContainerTools/jib/issues/2796)
+
+# Motivation 
+Allow users to containerize arbitrary jar files without having to integrate Jib into java build tools (maven and gradle).
```
loosebazooka, line 6: or create a jib-cli buildfile?

proposals/cli-jar-processing.md
```diff
@@ -0,0 +1,48 @@
...
+Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY}`
+The default mode for containerizing a jar. It will explode a jar into the following layers:  
+- Dependencies Layer: Contains dependencies derived from the `BOOT-INF/libs` directory in the jar.
+- Resources Layer: Contains resources parsed from the jar. Note that it will also include `MANIFEST.MF`.
+- Classes Layer: Contains classes parsed from the jar.
+
+**Entrypoint** : `java -cp . org.springframework.boot.loader.JarLauncher`
+
+### Packaged Mode
+Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY} --mode packaged`
```
loosebazooka, line 43: So I think in packaged mode, a fat jar should just be dropped into the container as is. Though we highly recommend against this, the fat jar does not contain an `original jar` -- it contains classes/resources.

proposals/cli-jar-processing.md
```diff
@@ -0,0 +1,48 @@
...
+- Dependencies:  Contains the dependencies derived from `Class-Path` in jar manifest. Note that this layer will not be created if `Class-Path` is not present in the manifest.
+- Jar Layer: Contains the original jar.
+
+**Entrypoint** : `java -jar ${JAR_NAME}.jar`
+
+## Spring-Boot Fat Jar
+A Spring-Boot Fat Jar can be containerized in two modes, exploded or packaged. 
+
+### Exploded Mode
+Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY}`
```
loosebazooka, line 34: I think we should try to parse a layers.idx if we find one: https://spring.io/blog/2020/08/14/creating-efficient-docker-images-with-spring-boot-2-3

proposals/cli-jar-processing.md
mpeddada1: Ah thank you

proposals/cli-jar-processing.md
mpeddada1: Oh nice.

proposals/cli-jar-processing.md
```diff
@@ -0,0 +1,48 @@
...
+A standard jar can be containerized in two modes, exploded or packaged. 
+
+### Exploded Mode
+Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY}`
+The default mode for containerizing a jar. It will explode a jar into the following layers:  
+- Dependencies Layer: Contains dependencies derived from `Class-Path` in jar manifest. Note that this layer will not be created if `Class-Path` is not present in the manifest.
+- Resources Layer: Contains resources parsed from jar file. Note that it will also include `MANIFEST.MF`.
+- Classes Layer: Contains classes parsed from jar file. 
+
+**Entrypoint** : `java -cp /app/dependencies/ /app/explodedJar/ ${MAIN_CLASS}`
```
chanseokoh, line 20:
```suggestion
**Entrypoint** : `java -cp /app/dependencies/:/app/explodedJar/ ${MAIN_CLASS}`
```
Would be `;` instead of `:` on Windows, but we don't support creating Windows images anyway.

proposals/cli-jar-processing.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Spring-Boot Fat Jar
+A Spring-Boot Fat Jar can be containerized in two modes, exploded or packaged. 
+
+### Exploded Mode
+Achieved by calling `jib jar ${JAR_NAME}.jar --to ${TARGET_REGISTRY}`
+The default mode for containerizing a jar. It will explode a jar according to what is specified in the `layers.idx` file of the jar, if present, or according to following format:
+- Dependencies Layer: For a dependency whose version does not contain `SNAPSHOT`.
+- Spring-Boot-Loader Layer: Contains jar loader classes.
+- Snapshot-Dependencies Layer: For a dependency whose version contains `SNAPSHOT`.
```
chanseokoh, line 38: We can also create a snapshot dep layer in the non-Spring Boot case.

proposals/cli-jar-processing.md
```diff
@@ -0,0 +1,48 @@
...
+- Resources Layer: Contains resources parsed from `BOOT-INF/classes/` in the jar and `META-INF/`.
+- Classes Layer: Contains classes parsed from `BOOT-INF/classes/` in the jar.
+
+**Entrypoint** : `java -cp . org.springframework.boot.loader.JarLauncher`
```
chanseokoh, line 42: Maybe
```suggestion
**Entrypoint** : `java -cp /app org.springframework.boot.loader.JarLauncher`
```
?

proposals/cli-jar-processing.md
mpeddada1: That is an interesting idea. I modified the standard jar description to include a snapshot dep layer. Will create an issue for this.

mpeddada1: Actually now that I'm looking at this again, what is the advantage of having a separate snapshot dependencies layer?

loosebazooka: `SNAPSHOT` layers are generally those that are created from a non-release build of a project (like nightly). They could be built locally from a project (like we do for people testing jib in our [CONTRIBUTING.md](https://github.com/GoogleContainerTools/jib/blob/master/CONTRIBUTING.md#debugging-the-jib-maven-plugin-jib-maven-plugin)), or they could be something like a nightly build from a project between releases.

As a consequence snapshot dependencies are "mutable" in a sense (1.0-SNAPSHOT could be built every day with different results), they can change between builds. If we have frequently changing dependencies in the regular dependencies layer then that layer changes on every build and we have to push all dependencies up.

If we separate out snapshot dependencies, then we can isolate them from the more stable other regular dependencies and only rebuild/repush a snapshot layer.

chanseokoh (approved):

PR #2836, opened 2020-10-16
Credential configuration support

loosebazooka:
Add a helper class to help process credentials with support from DefaultCredentialRetrievers

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Credentials.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Credentials.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.cli.cli2;
+
+import com.google.cloud.tools.jib.api.CredentialRetriever;
+import com.google.cloud.tools.jib.plugins.common.DefaultCredentialRetrievers;
+import java.io.FileNotFoundException;
+import java.util.List;
+
+/**
+ * A helper class to process command line args and generate a list of {@link CredentialRetriever}s.
+ */
+public class Credentials {
+  /**
+   * Get credentials for a target image registry.
+   *
+   * @param buildOptions The command line build options
+   * @param defaultCredentialRetrievers An initialized {@link DefaultCredentialRetrievers} to use
+   * @return a list of credentials for a target image registry
+   * @throws FileNotFoundException when a credential helper file cannot be found
+   */
+  public static List<CredentialRetriever> getToCredentialRetrievers(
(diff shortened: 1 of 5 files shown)


jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/JibCliTest.java
```diff
@@ -387,6 +387,8 @@ public void testParse_passwordWithoutUsername() {
             {{"--username=x", "--password=x", "--to-username=x", "--to-password=x"}},
             {{"--username=x", "--password=x", "--to-credential-helper=x"}},
             {{"--username=x", "--password=x", "--from-credential-helper=x"}},
+            {{"--from-credential-helper=x", "--from-username=x", "--from-password=x"}},
```
loosebazooka, line 390: also just a minor update to fix mixing test case

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Credentials.java
```diff
@@ -0,0 +1,91 @@
...
+import com.google.cloud.tools.jib.plugins.common.DefaultCredentialRetrievers;
+import java.io.FileNotFoundException;
+import java.util.List;
+
+/**
+ * A helper class to process command line args and generate a list of {@link CredentialRetriever}s.
+ */
+public class Credentials {
+  /**
+   * Get credentials for a target image registry.
```
chanseokoh, line 29:
```suggestion
   * Gets credentials for a target image registry.
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CredentialsTest.java
```diff
@@ -0,0 +1,265 @@
...
+      assertThat(captor.getValue()).isEqualTo(Credential.from("abc", "xyz"));
+      Mockito.verify(defaultCredentialRetrievers).asList();
+      Mockito.verifyNoMoreInteractions(defaultCredentialRetrievers);
+    }
+  }
+
+  @RunWith(Parameterized.class)
+  public static class GetFromCredentialRetrieverUsernamePassword {
+    @Rule public MockitoRule mockitoJUnit = MockitoJUnit.rule();
+    @Mock private DefaultCredentialRetrievers defaultCredentialRetrievers;
```
chanseokoh, line 217: All these can be `final`.

chanseokoh (approved):

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CredentialsTest.java
loosebazooka: yeah I keep making this mistake, I should be more careful, maybe there's a rule we can add to something

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CredentialsTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+  private Object paramsToNone() {
+    return new Object[]{
+        new String[]{"--from-credential-helper=ignored"},
+        new String[]{"--from-username=ignored", "--from-password=ignored"},
+    };
+  }
+
+  @Test
+  @Parameters(method = "paramsToNone")
```
mpeddada1, line 54: optional: Not sure how this would look but how about specifying the params directly in the tests instead of calling a method?

loosebazooka: It does appear to work for array types.

chanseokoh: Probably @loosebazooka meant "doesn't."

chanseokoh (approved):

Issue #2839, opened 2020-10-19
jib-cli todo integration tests


loosebazooka:
Create integration tests once the system is fully integrated
- [ ] BuildFiles.createJibContainerBuilder

chanseokoh:
Should this be part of #2570?

loosebazooka:
I was thinking that was feature work related to the cli. This is integration testing.

loosebazooka:
Also I think maybe should be using the `Projects` feature. Let me try something out.

PR #2840, opened 2020-10-19
Add credentials for `from` image in BuildFile

loosebazooka:
This doesn have a test, needs to added as an integration test after the system is integrated together. In general it seems like the code around parsing an image and creating a containerizer/container-builder is a little tough and tends to be handled by integration tests.

follow-up issue: https://github.com/GoogleContainerTools/jib/issues/2839

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
@@ -35,13 +35,13 @@ import com.google.cloud.tools.jib.cli.cli2.JibCli;
 import com.google.cloud.tools.jib.frontend.CredentialRetrieverFactory;
 import com.google.cloud.tools.jib.plugins.common.DefaultCredentialRetrievers;
 import com.google.cloud.tools.jib.plugins.common.logging.ConsoleLogger;
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.stream.Collectors;
 import org.apache.commons.text.StringSubstitutor;
@@ -69,7 +69,7 @@ public class BuildFiles {
    *
    * @param projectRoot the root context directory of this build
    * @param buildFilePath a file containing the build definition
-   * @param buildOptions the build configuration from the command line
+   * @param cliOptions the build configuration from the command line
    * @param logger a logger to inject into various objects that do logging
    * @return a {@link JibContainerBuilder} generated from the contents of {@code buildFilePath}
    * @throws IOException if an I/O error occurs opening the file, or an error occurs while
@@ -77,13 +77,13 @@ public class BuildFiles {
    * @throws InvalidImageReferenceException if the baseImage reference can not be parsed
    */
   public static JibContainerBuilder toJibContainerBuilder(
-      Path projectRoot, Path buildFilePath, JibCli buildOptions, ConsoleLogger logger)
+      Path projectRoot, Path buildFilePath, JibCli cliOptions, ConsoleLogger logger)
       throws InvalidImageReferenceException, IOException {
-    BuildFileSpec buildFile = toBuildFileSpec(buildFilePath, buildOptions.getTemplateParameters());
+    BuildFileSpec buildFile = toBuildFileSpec(buildFilePath, cliOptions.getTemplateParameters());
 
     JibContainerBuilder containerBuilder =
         buildFile.getFrom().isPresent()
-            ? createJibContainerBuilder(buildFile.getFrom().get(), buildOptions, logger)
+            ? createJibContainerBuilder(buildFile.getFrom().get(), cliOptions, logger)
             : Jib.fromScratch();
 
     buildFile.getCreationTime().ifPresent(containerBuilder::setCreationTime);
@@ -104,11 +104,10 @@ public class BuildFiles {
     return containerBuilder;
   }
 
-  @VisibleForTesting
   // TODO: add testing, need to do via intergration tests as there's no good way to extract out that
   //   the base image was populated as the user intended currently.
   static JibContainerBuilder createJibContainerBuilder(
-      BaseImageSpec from, JibCli buildOptions, ConsoleLogger logger)
+      BaseImageSpec from, JibCli cliOptions, ConsoleLogger logger)
       throws InvalidImageReferenceException, FileNotFoundException {
     String baseImageReference = from.getImage();
     if (baseImageReference.startsWith(DOCKER_DAEMON_IMAGE_PREFIX)) {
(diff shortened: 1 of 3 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -98,4 +103,42 @@ public static JibContainerBuilder toJibContainerBuilder(
...
+    Credentials.getFromCredentialRetrievers(buildOptions, defaultCredentialRetrievers)
+        .forEach(registryImage::addCredentialRetriever);
+    JibContainerBuilder containerBuilder = Jib.from(registryImage);
+    if (!from.getPlatforms().isEmpty()) {
+      containerBuilder.setPlatforms(
+          from.getPlatforms()
+              .stream()
+              .map(
+                  platformSpec ->
+                      new Platform(platformSpec.getArchitecture(), platformSpec.getOs()))
```
mpeddada1, line 139: Is it required for platform spec to be fully specified?

mpeddada1: Oh looks like it is.

mpeddada1 (commented):
Looks good to me.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -53,33 +69,22 @@ private static BuildFileSpec toBuildFileSpec(
    *
    * @param projectRoot the root context directory of this build
    * @param buildFilePath a file containing the build definition
-   * @param templateParameters a map of templating variables to apply on the file before parsing
+   * @param buildOptions the build configuration from the command line
```
mpeddada1, line 72: maybe we can add a small `e.g ` at the end to make this more clear?

mpeddada1 (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -98,4 +103,42 @@ public static JibContainerBuilder toJibContainerBuilder(
...
+        .forEach(registryImage::addCredentialRetriever);
+    JibContainerBuilder containerBuilder = Jib.from(registryImage);
+    if (!from.getPlatforms().isEmpty()) {
+      containerBuilder.setPlatforms(
+          from.getPlatforms()
+              .stream()
+              .map(
+                  platformSpec ->
+                      new Platform(platformSpec.getArchitecture(), platformSpec.getOs()))
+              .collect(Collectors.toSet()));
```
chanseokoh, line 140: The iteration order is actually important. Internally we used `LinkedHashSet` everywhere to preserve the order while conveniently removing duplicates (also signifying that duplicates can't exist). However, I see that it can easily make the Jib API user to mess up with the order. Maybe accepting a `Set` wasn't a good idea.

Anyways, the question is, would `Collectors.toSet()` preserve the insertion order?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -98,4 +103,42 @@ public static JibContainerBuilder toJibContainerBuilder(
     }
     return containerBuilder;
   }
+
+  @VisibleForTesting
+  // TODO: add testing, need to do via intergration tests as there's no good way to extract out that
+  //   the base image was populated as the user intended currently.
+  static JibContainerBuilder createJibContainerBuilder(
```
chanseokoh, line 110: I think this can be private without @VisibleForTesting?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -98,4 +103,42 @@ public static JibContainerBuilder toJibContainerBuilder(
     }
     return containerBuilder;
   }
+
+  @VisibleForTesting
+  // TODO: add testing, need to do via intergration tests as there's no good way to extract out that
+  //   the base image was populated as the user intended currently.
```
mpeddada1, line 109: sorry, does this PR need to submitted before we start writing the integration tests?

loosebazooka: yeah, well, really they all do, so we can run full builds using the CLI.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
loosebazooka: oh right.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
loosebazooka: no but I think I can do: `Collectors.toCollection(LinkedHashSet::new)`

loosebazooka: Why does the order matter? for selecting defaults?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFilesTest.java
```diff
@@ -38,13 +40,27 @@
...
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 
 public class BuildFilesTest {
 
   @Rule public final TemporaryFolder tmp = new TemporaryFolder();
+  @Rule public MockitoRule rule = MockitoJUnit.rule();
+
+  @Mock ConsoleLogger consoleLogger;
+  @Mock JibCli jibCli;
```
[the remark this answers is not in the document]
loosebazooka: oof, sorry, it strikes again

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
loosebazooka: I could add en example, but not sure what you're for an example of?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
chanseokoh: jib-core will create a manifest list with the given platform order.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
mpeddada1: ah right.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
mpeddada1: Oops I totally missed that buildOptions was of type `JibCli`.  Adding an example is probably not useful here. Should the name be a more general, like `cliOptions`, if JibCli also includes things like verbosity and console?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
loosebazooka: oh right.

PR #2849, opened 2020-10-20
Fill in directories for files added by filter.

### LINKED ISSUE #2843: [jib-cli] Correctly create parent directories on filter files
https://github.com/GoogleContainerTools/jib/pull/2700#discussion_r471560251

currently for filter files, the parent directory is ignored by nature of the filter. Because of this, the filtered files are not added with their parents populated. This could lead to an issue where the directory permissions are not as expected.

for a directory structure like:
```
dir1/dir2/
  - file1
  - file2
  - ignore1
```
and a file filter like `**/file*`, the layer parser only adds
```
dir1/dir2/file1
dir1/dir2/file2
```

What it should add is:
```
dir1
dir1/dir2
dir1/dir2/file1
dir1/dir2/file2
```
so that it correctly adds the permissions of `dir1` and `dir2` based on the spec from the build file for directories

loosebazooka:
This happens where parent directories are filtered
but need to be filled in for files in those
directories

fixes #2843

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.jib.cli.buildfile;
 import com.google.cloud.tools.jib.api.buildplan.AbsoluteUnixPath;
 import com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer;
 import com.google.cloud.tools.jib.api.buildplan.FileEntry;
+import com.google.cloud.tools.jib.api.buildplan.FilePermissions;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.nio.file.FileSystems;
@@ -26,7 +27,10 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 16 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -96,49 +104,72 @@
...
+                } else if (Files.isRegularFile(path)) {
+                  if (!path.startsWith(src)) {
+                    // if we end up in a situation where the file added is somehow outside of the
+                    // tree then we do not know how to properly handle it at the moment. It could
+                    // be from a link scenario that we do not understand.
+                    throw new IllegalStateException(
+                        src.toString() + " is not a parent of " + path.toString());
+                  }
+                  Path parent = path.getParent();
+                  while (true) {
```
loosebazooka, line 156: Actually these need to be added in reverse order

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/Layers.java
```diff
@@ -71,6 +75,10 @@
           Path src = rawSrc.isAbsolute() ? rawSrc : buildRoot.resolve(rawSrc);
           AbsoluteUnixPath dest = copySpec.getDest();
 
+          if (!Files.isDirectory(src) && !Files.isRegularFile(src)) {
```
chanseokoh, line 78: Is this primarily to guard against symlinks?

loosebazooka: Yeah there's two checks. One for non existence and one for unknown types. Perhaps the messaging should be different

jib-cli/src/test/resources/buildfiles/layers/includesExcludesTest/layers.yaml
```diff
@@ -33,11 +33,20 @@ entries:
         excludes:
           # equivalent to "**/excludedDir/**"
           - "**/excludedDir/"
+  - name: "exclude dir and contents"
+    files:
+      - src: "project"
+        dest: "/target/edac"
+        excludes:
+          # equivalent to "**/excludedDir/**"
```
chanseokoh, line 41:
```suggestion
```

chanseokoh (approved):

PR #2854, opened 2020-10-20
Close JarFile after instantiating

chanseokoh:
Turns out `JarFile` is `Closeable`. My IDE shows a resource leak warning.


mpeddada1 (approved):
Oh I didn't realize that it was closeable. Thanks for sending out this fix!

Issue #2845, opened 2020-10-19
[jib-cli-jar] Compute entrypoint for standard jar. 


mpeddada1:

mpeddada1:
Closed by #2862

Issue #1871, opened 2019-07-28
Jar loading order error causes spring-boot startup to fail


xiaopihai7256:
**Environment**:

- *Jib version: 1.4.0*
- *Build tool: maven 3.3.9* 
- *OS: windows 10 1903*
- *Server OS [use to run Docker imange]: Linux version 5.1.14-1.el7.elrepo.x86_64 *
- Spring-boot: 2.0.6.RELEASE

**Description of the issue**:

Sorry, my English is not good, but I will try to explain the problems I have encountered.

In my project, there are two jars with different names, we call A.jar and B.jar. A rewrites and overwrites a Class with the same name in B. In order to work properly, I introduce A in pom.xml in order higher than B, which works fine in Spring-boot's Fatjar.

But when I use Jib, B always loads before A, causing the coverage of the Class to not take effect, so when the application starts, the JDK prompts "XXXX method not found in XXX.class". finally, application start failed

**Expected behavior**:

A.jar can be loaded earlier than B.jar, just like working in Spring-boot's Fatjar.

**Steps to reproduce**:

  1. add Jib maven plugin to pom.xml
  2. run command: `mvn compile jib:build -Dregistry_url={my server ip}:4000 -Dregistry_username={username} -Dregistry_password={password} -Dimage_tag=v1.2.0-dev-2019072101 -DsendCredentialsOverHttp=true`
  3. try to run docker image, spring-boot startup failed

**`jib-maven-plugin` Configuration**: 
```xml
<plugin>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-maven-plugin</artifactId>
</plugin>
<plugin>
    <groupId>com.google.cloud.tools</groupId>
    <artifactId>jib-maven-plugin</artifactId>
    <version>1.4.0</version>
    <configuration>
        <container>
            <jvmFlags>
                <jvmFlag>-Xms512m</jvmFlag>
            </jvmFlags>
            <ports>
                <port>8080</port>
            </ports>
            <useCurrentTimestamp>true</useCurrentTimestamp>
        </container>
        <from>
            <image>registry.cn-hangzhou.aliyuncs.com/choerodon-tools/javabase:0.7.1</image>
        </from>
        <to>
            <image>${registry_url}/${docker.image.prefix}/${project.artifactId}:${maven.build.timestamp}</image>
            <auth>
                <username>${registry_username}</username>
                <password>${registry_password}</password>
            </auth>
        </to>
[...]

PR #2862, opened 2020-10-24
Compute entrypoint for a standard jar in exploded mode.

mpeddada1:
For #2845
This PR also contains some minor refactorings.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
@@ -89,7 +89,7 @@ public class JarProcessor {
 
     // Get dependencies from Class-Path in the jar's manifest and add a layer each for non-snapshot
     // and snapshot dependencies. If Class-Path is not present in the jar's manifest then skip
-    // adding a dependencies layer.
+    // adding the dependencies layers.
     String classPath = null;
     try (JarFile jarFile = new JarFile(jarPath.toFile())) {
       classPath = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
@@ -125,12 +125,11 @@ public class JarProcessor {
       }
     }
 
-    Predicate<Path> isClassFile = path -> path.getFileName().toString().endsWith(".class");
-    Predicate<Path> isResourceFile = isClassFile.negate();
-
     // Determine class and resource files in the directory containing jar contents and create
     // FileEntriesLayer for each type of layer (classes or resources), while maintaining the
     // file's original project structure.
+    Predicate<Path> isClassFile = path -> path.getFileName().toString().endsWith(".class");
+    Predicate<Path> isResourceFile = isClassFile.negate();
     FileEntriesLayer classesLayer =
         addDirectoryContentsToLayer(
             CLASSES,
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
@@ -22,6 +22,7 @@ import com.google.cloud.tools.jib.api.buildplan.RelativeUnixPath;
 import com.google.cloud.tools.jib.filesystem.DirectoryWalker;
 import com.google.cloud.tools.jib.plugins.common.ZipUtil;
 import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -37,6 +38,10 @@ import java.util.stream.Collectors;
 public class JarProcessor {
 
   private static final AbsoluteUnixPath APP_ROOT = AbsoluteUnixPath.get("/app");
+  private static final String CLASSES = "classes";
+  private static final String RESOURCES = "resources";
+  private static final String DEPENDENCIES = "dependencies";
+  private static final String SNAPSHOT_DEPENDENCIES = "snapshot dependencies";
 
   /**
    * Jar Type.
@@ -89,7 +94,6 @@ public class JarProcessor {
     try (JarFile jarFile = new JarFile(jarPath.toFile())) {
       classPath = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
     }
-
     if (classPath != null) {
       Predicate<String> isSnapshot = name -> name.contains("SNAPSHOT");
       List<String> allDependencies = Splitter.onPattern("\\s+").splitToList(classPath.trim());
@@ -103,7 +107,7 @@ public class JarProcessor {
           allDependencies.stream().filter(isSnapshot).map(Paths::get).collect(Collectors.toList());
       if (!nonSnapshotDependencies.isEmpty()) {
         FileEntriesLayer.Builder nonSnapshotDependenciesLayerBuilder =
-            FileEntriesLayer.builder().setName("dependencies");
+            FileEntriesLayer.builder().setName(DEPENDENCIES);
         nonSnapshotDependencies.forEach(
             path ->
                 nonSnapshotDependenciesLayerBuilder.addEntry(
@@ -112,7 +116,7 @@ public class JarProcessor {
       }
       if (!snapshotDependencies.isEmpty()) {
         FileEntriesLayer.Builder snapshotDependenciesLayerBuilder =
-            FileEntriesLayer.builder().setName("snapshot dependencies");
+            FileEntriesLayer.builder().setName(SNAPSHOT_DEPENDENCIES);
         snapshotDependencies.forEach(
             path ->
                 snapshotDependenciesLayerBuilder.addEntry(
@@ -129,13 +133,13 @@ public class JarProcessor {
     // file's original project structure.
     FileEntriesLayer classesLayer =
         addDirectoryContentsToLayer(
-            "classes",
+            CLASSES,
             localExplodedJarRoot,
             isClassFile,
             APP_ROOT.resolve(RelativeUnixPath.get("explodedJar")));
     FileEntriesLayer resourcesLayer =
         addDirectoryContentsToLayer(
-            "resources",
+            RESOURCES,
             localExplodedJarRoot,
             isResourceFile,
             APP_ROOT.resolve(RelativeUnixPath.get("explodedJar")));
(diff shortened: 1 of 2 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,42 @@ public static JarType determineJarType(Path jarPath) throws IOException {
     return layers;
   }
 
+  /**
+   * Compute the entrypoint for a standard jar in exploded mode.
```
chanseokoh, line 153: I remember this used to be caught by the formatter. Maybe not anymore.

```suggestion
   * Computes the entrypoint for a standard jar in exploded mode.
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorTest.java
```diff
@@ -281,4 +282,46 @@ public void testExplodeMode_standard_withoutClassPathInManifest_containsOnlyClas
             AbsoluteUnixPath.get("/app/explodedJar/class1.class"),
             AbsoluteUnixPath.get("/app/explodedJar/class2.class"));
   }
+
+  @Test
+  public void testExplodeMode_standard_computeEntrypoint_allLayersPresent()
+      throws IOException, URISyntaxException {
+    Path standardJar =
+        Paths.get(Resources.getResource(STANDARD_JAR_WITH_CLASS_PATH_MANIFEST).toURI());
+    Path destDir = temporaryFolder.newFolder().toPath();
```
chanseokoh, line 291: Just so you know, if you don't need a nested directory or multiple directories, you can use the root: `temporaryFolder.getRoot().toPath()`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,42 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+   *
+   * @param jarPath path to jar file
+   * @param tempDirPath path to temporary jib local directory
+   * @param layers list of {@link FileEntriesLayer}
+   * @return list of {@link String} representing entrypoint
+   * @throws IOException if I/O error occurs when opening the jar file or if temporary directory
+   *     provided doesn't exist
+   * @throws IllegalArgumentException if main class is not found in the jar manifest
+   */
+  public static ImmutableList<String> computeEntrypoint_explodedStandard(
```
chanseokoh, line 163: Using `_` is only accepted in test method names. `computeEntrypointForExplodedStandard()`?

And I don't know who will require this method, but you should always double-checking if a method may have a narrower access scope. (Just don't habitually make everything `public` until necessary.)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,42 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+      Path jarPath, Path tempDirPath, List<FileEntriesLayer> layers)
+      throws IOException, IllegalStateException {
```
chanseokoh, line 165: I admit we don't have a good established policy around throwing, handling, and reporting exceptions and errors. We have not done well so far. But at least the situation for Jib CLI is not so grim, as it is not a library but a CLI. So I think throwing a runtime exception to crash for this type of errors is more or less acceptable.

Anyways, what I wanted to say is that, I think it's fine to not specify `IllegalStateException` (or `IllegalArgumentException`). Also in the Javadoc.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,42 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+    Path localExplodedJarRoot = tempDirPath;
+    ZipUtil.unzip(jarPath, localExplodedJarRoot);
+
+    String mainClass = null;
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      mainClass = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.MAIN_CLASS);
+    }
+    if (mainClass == null) {
+      throw new IllegalArgumentException("Main-Class not found in jar's manifest.");
```
chanseokoh, line 174: The CLI will crash with this message. And providing a clear message implying an actionable fix increases the chance of the user understanding and fixing the issue on their own, resulting in less frustrated getting-started experience (which is important to increase the chance of adoption). It also reduces our workload by reducing the likelyhood of multiple user opening the same issue.

```suggestion
      throw new IllegalArgumentException("'Main-Class:' attribute to define an application main class not defined in the input JAR's manifest ('META-INF/MANIFEST.MF` in the JAR).");
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,42 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+    }
+    List<String> classpath = new ArrayList<>();
+    for (FileEntriesLayer layer : layers) {
+      if (layer.getName().equals(DEPENDENCIES) || layer.getName().equals(SNAPSHOT_DEPENDENCIES)) {
+        classpath.add(APP_ROOT.resolve("dependencies").toString());
```
chanseokoh, line 179: At least this should be
```suggestion
        classpath.add(APP_ROOT + "/dependencies/*");
```
Just so you know, there's an issue with the order of loading jars (like #1871 for Jib build plugins), but I think there's nothing we can do about it for a standard jar.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,42 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+        break;
+      }
+    }
+    classpath.add(APP_ROOT.resolve("explodedJar").toString());
+    String classPathString = String.join(":", classpath);
```
chanseokoh, line 184: Considering the order of loading classes, `explodedJar` should come before `dependencies`. That said, how about going simple? It's okay to put a non-existing directory. I think the level of code reduction is worth.

```java
String classpath = APP_ROOT + "/explodedJar:" + APP_ROOT + "/dependencies/*";
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Ah you're right. Better error messaging is certainly helpful especially with something that causes the code to crash. Thanks.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: oops I keep mixing up the naming with python convention

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: is the ordering issue not a problem for a standard jar atm?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -76,7 +81,7 @@ public static JarType determineJarType(Path jarPath) throws IOException {
    * @throws IOException if I/O error occurs when opening the jar file or if temporary directory
    *     provided doesn't exist
    */
-  public static List<FileEntriesLayer> explodeStandardJar(Path jarPath, Path tempDirPath)
+  protected static List<FileEntriesLayer> explodeStandardJar(Path jarPath, Path tempDirPath)
```
chanseokoh, line 84: Perhaps this can be package-private?
```suggestion
  static List<FileEntriesLayer> explodeStandardJar(Path jarPath, Path tempDirPath)
```
And if this needs to be package-private only to be able to call from tests (i.e., it should have been `private` if not for the tests), add `@VisibleForTesting`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,33 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+    Path localExplodedJarRoot = tempDirPath;
+    ZipUtil.unzip(jarPath, localExplodedJarRoot);
+
+    String mainClass = null;
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      mainClass = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.MAIN_CLASS);
+    }
+    if (mainClass == null) {
+      throw new IllegalArgumentException(
+          "`Main-Class:` attribute to define an application main class not defined in the input Jar's manifest (`META-INF/MANIFEST.MF` in the JAR).");
```
chanseokoh, line 173: Sorry, I think this reads better:
```suggestion
          "`Main-Class:` attribute for an application main class not defined in the input Jar's manifest (`META-INF/MANIFEST.MF` in the Jar).");
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,33 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+   * @param tempDirPath path to temporary jib local directory
+   * @param layers list of {@link FileEntriesLayer}
+   * @return list of {@link String} representing entrypoint
+   * @throws IOException if I/O error occurs when opening the jar file or if temporary directory
+   *     provided doesn't exist
+   */
+  protected static ImmutableList<String> computeEntrypointForExplodedStandard(
+      Path jarPath, Path tempDirPath, List<FileEntriesLayer> layers) throws IOException {
+    Path localExplodedJarRoot = tempDirPath;
+    ZipUtil.unzip(jarPath, localExplodedJarRoot);
```
chanseokoh, line 165: Oh, actually, I remember unzipping a jar somewhere else too. Are we unzipping it multiple times? We should unzip only once for efficiency.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -145,6 +149,33 @@ public static JarType determineJarType(Path jarPath) throws IOException {
...
+
+    String mainClass = null;
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      mainClass = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.MAIN_CLASS);
+    }
```
chanseokoh, line 170: Nit: I think this code block is super fast, so it's good to put everything in to the try-resource block.
```suggestion
    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
      String mainClass = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.MAIN_CLASS);
      if (mainClass == null)
      ...
      return ...
    }
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
chanseokoh: Actually, I think there is no need to unzip a jar?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Done.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
mpeddada1: Cleaned this up a bit.

chanseokoh (approved):

Issue #2733, opened 2020-08-27
Classpath ordering in container dependent on host filesystem


cpopp:
**Environment**:

- *Jib version:* 1.6.1
- *Build tool:* maven, 3.5.4
- *OS:* Ubuntu 18.04.5

**Description of the issue**:
The use of /app/libs/* in the classpath argument for java means that ordering of jars on the classpath is dependent on the host system. We had conflicting jars in our application, but it worked fine on our dev servers, however another set of servers had the jars ordered differently on the classpath and the jar conflict resulted in our application not starting.

**Expected behavior**:
We expect the containers to be as portable as possible between environments so we hoped the classpath would be ordered identically between environments which would occur if for example the jars were listed explicitly on the classpath rather than as /app/libs/*

**Steps to reproduce**:
  1. Build image with Google jib
  2. Run on one host and print the classpath
  3. Run on another host with an alternative filesystem and print the classpath
  4. See classpath order differs

**`jib-maven-plugin` Configuration**: 
```xml
<plugin>
    <groupId>com.google.cloud.tools</groupId>
    <artifactId>jib-maven-plugin</artifactId>
    <version>1.6.1</version>
</plugin>
```

**Additional Information**: 

Classpath wildcard iteration on a Linux host is shown here in the openjdk source: https://github.com/openjdk/jdk/blob/6bab0f539fba8fb441697846347597b4a0ade428/src/java.base/share/native/libjli/wildcard.c#L209

Which points to the use of readdir (https://www.man7.org/linux/man-pages/man3/readdir.3.html) which indicates:

> The order in which filenames are read by successive calls to
>        readdir() depends on the filesystem implementation; it is unlikely
>        that the names will be sorted in any fashion.

which is exactly what we ran into when the option for mounting the host filesystem differed.

chanseokoh:
Hi @cpopp,

This is a known issue: #1871 #1907 It's on our radar and something we really wanted to fix early, but we always had other priority issues.

[...]

Issue #2471, opened 2020-05-18
Enable to generate and package a CDS archive


rmannibucau:
Hi,

CDS enables java application to boot way faster (in particular when any scanning/reflection is involved) so it would be neat to integrate it with JIB.
It requires to launch 2-3 commands to generate the archive and then just bundle it and modify the jvm arguments.

https://blog.codefx.org/java/application-class-data-sharing/#Creating-A-JDK-Class-Data-Archive explains it quite well.

Since the classpath must be stable, doing it in jib seems the most reliable.

Romain

chanseokoh:
Never heard of CDS, but this does seem worth looking into it at some point. Thanks for the feedback! Does CDS only work with `-jar my.jar` or can work with the classpath launch style (`-cp ... com.example.MyMain`)?

rmannibucau:
Wrote a quick post about it: https://rmannibucau.metawerx.net/post/java-class-data-sharing-docker-startup

Long story short, CDS is supported by any java app but classpath beginning must be stable and match the archive so jib can't use lib/*.jar for example.

I wrapped jib in a custom main and gain is ~30% on java 11 (i'm using zulu) on my app (CDI) so definitively worth enabling :).

chanseokoh:
Thanks for the info. From which Java version can we use this feature?

> jib can't use lib/*.jar for example.

@GoogleContainerTools/java-tools-build I remember the discussion that we can't get rid of the classpath wildcard `libs/*` until Java 9+ because of the max argument length limit or the max command line length (particularly short on Windows). However, looks like [modern Linux kernels support up to 2MB](https://stackoverflow.com/a/41762581/1701388) and seems like this is [not an issue in practice](https://stackoverflow.com/a/11243586/1701388). (We don't have to think about Windows.)

rmannibucau:
@chanseokoh think it was around java 10 (not sure commands are 100% the same in the first versions). These commands work well on java 11.
BTW I would expect this feature to be off by default since it requires to execute docker commands so somebody enabling it would do it intentionally and classpath would fit the command line anyway I guess?

chanseokoh:
FTR:
https://cloud.google.com/run/docs/tips/java#appcds
https://github.com/saturnism/jvm-helloworld-by-example/tree/master/helloworld-springboot-tomcat-shaded

saturnism:
I took Jib for a spin to see how to do this mechanically (w/o automation). The basics premises are:
[...]

Issue #2228, opened 2020-01-14
docker run failed


SandNight:
**Environment**:
- *Jib version:* 1.8.0
- *Build tool:* Maven  3.6.1
- *OS:* Mac 10.15.2

**Description of the issue**:

run jib-maven-plugin:1.8.0:dockerBuild in idea
the image build successful
but when I run this image, failed

Maybe the JDK configuration is wrong
What should I do

**Expected behavior**:
run success

**Steps to reproduce**:

  1. dockerBuild
  2. run image

**`jib-maven-plugin` Configuration**: 
```xml
<plugin>
                <groupId>com.google.cloud.tools</groupId>
                <artifactId>jib-maven-plugin</artifactId>
                <version>1.8.0</version>
                <configuration>
                    <to>
                        <image>harbor.test.com/${project.parent.artifactId}/${project.artifactId}:${env}-${project.version}</image>
                    </to>
                    <from>
                        <image>openjdk:alpine</image>
                    </from>
                    <container>
                        <mainClass>com.test.auth.server.AuthServerApplication</mainClass>
                        <jvmFlags>
                            <jvmFlag>-Xms512m</jvmFlag>
                        </jvmFlags>
                        <ports>
                            <port>9082</port>
                        </ports>
                        <environment>
                            <TZ>Asia/Shanghai</TZ>
                        </environment>
                        <volumes>/tmp</volumes>
                        <useCurrentTimestamp>true</useCurrentTimestamp>
                        <labels>
                            <maintainer>jade</maintainer>
                        </labels>
                    </container>
                    <allowInsecureRegistries>true</allowInsecureRegistries>
                </configuration>
            </plugin>
```

**Log output**: 

[...]

Issue #1907, opened 2019-08-15
Fail to use JIB with JAXWS dependencies


raizoor:
**Environment**:

- *Jib version: 1.4*
- *Build tool:* Maven 3.6 -->
- *OS: Linux Server*

**Description of the issue**:
Team,

Today, i have this problem:
- An team need building API that need use JAXWS dependencies, to handle SOAP requests. My build tests (unit and integration) has ok, but when goes to deploy, we have that stack:

> *************************** APPLICATION FAILED TO START *************************** Description: An attempt was made to call a method that does not exist. The attempt was made from the following location: io.undertow.websockets.jsr.Bootstrap.handleDeployment(Bootstrap.java:84) The following method did not exist: javax.servlet.ServletContext.getClassLoader()Ljava/lang/ClassLoader; The method's class, javax.servlet.ServletContext, is available from the following locations: jar:file:/app/libs/geronimo-servlet_2.5_spec-1.2.jar!/javax/servlet/ServletContext.class jar:file:/app/libs/javax.servlet-api-4.0.1.jar!/javax/servlet/ServletContext.class It was loaded from the following location: file:/app/libs/geronimo-servlet_2.5_spec-1.2.jar 

As I search, this error occours when spring detect two classes to do same thing, but in this case, he uses the bad servlet =/
If eu use mvn clean install, i have success.
Exists an option where can i set which class I want use? Or do you know how can i resolve this?
Now i'm using undertow, but I already test with tomcat, but not success.

raizoor:
[tempPom.txt](https://github.com/GoogleContainerTools/jib/files/3507340/tempPom.txt)

chanseokoh:
I assume `mvn spring-boot:run` runs OK, and this problem happens only when running the image built by Jib.

Can you do `mvn dependency:tree` and see how `geronimo-servlet_2.5_spec-1.2.jar` is pulled in? Perhaps just upload the output result.

raizoor:
Hi @chanseokoh ,

Yap, if I don't use jib, it work.
I have already run mvn tree, and Geronimo is pulled by `<groupId>org.apache.cxf</groupId>
            <artifactId>cxf-rt-transports-http-netty-server</artifactId>
`
I tried to put scope = provided, but it broke my startup API =/

chanseokoh:
[...]

PR #2866, opened 2020-10-26
Add option to expand wildcard in entrypoint runtime classpath

### LINKED ISSUE #1871: Jar loading order error causes spring-boot startup to fail

chanseokoh:
Gives a viable solution to #1871 (#1907, #2228, #2733).

Also enables the AppCDS support (#2471).

This PR adds the option `jib.container.expandClasspathDependencies` that expands the wildcard `.../libs/*` in the entrypoint JVM classpath. It's off by default. The feature to use an `@argument` file will be added later for a permanent fix for the issue later. See https://github.com/GoogleContainerTools/jib/pull/2866#discussion_r512790086 for details.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
@@ -35,10 +35,6 @@ public class JibSystemProperties {
 
   @VisibleForTesting public static final String SKIP_EXISTING_IMAGES = "jib.skipExistingImages";
 
-  @VisibleForTesting
-  static final String USE_WILDCARD_FOR_DEPENDENCY_CLASSPATH =
-      "jib.useWildcardForDependencyClasspath";
-
   /**
    * Gets the HTTP connection/read timeouts for registry interactions in milliseconds. This is
    * defined by the {@code jib.httpTimeout} system property. The default value is 20000 if the
@@ -125,17 +121,6 @@ public class JibSystemProperties {
    */
   public static boolean skipExistingImages() {
     return Boolean.getBoolean(SKIP_EXISTING_IMAGES);
-  }
-
-  /**
-   * Gets whether or not to use a wildcard or enumerate dependency JARs in the entrypoint runtime
-   * classpath.
-   *
-   * @return {@code true} if Jib should use a wildcard for dependency JARs; {@code false} to
-   *     enumerate all the JARs
-   */
-  public static boolean useWildcardForDependencyClasspath() {
-    return Boolean.getBoolean(USE_WILDCARD_FOR_DEPENDENCY_CLASSPATH);
   }
 
   private static void checkNumericSystemProperty(String property, Range<Integer> validRange) {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
@@ -24,6 +24,7 @@ import com.google.common.collect.Range;
 public class JibSystemProperties {
 
   public static final String UPSTREAM_CLIENT = "_JIB_UPSTREAM_CLIENT";
+  private static final String DISABLE_USER_AGENT = "_JIB_DISABLE_USER_AGENT";
 
   @VisibleForTesting public static final String HTTP_TIMEOUT = "jib.httpTimeout";
 
@@ -32,9 +33,11 @@ public class JibSystemProperties {
   public static final String SEND_CREDENTIALS_OVER_HTTP = "sendCredentialsOverHttp";
   public static final String SERIALIZE = "jib.serialize";
 
-  private static final String DISABLE_USER_AGENT = "_JIB_DISABLE_USER_AGENT";
-
   @VisibleForTesting public static final String SKIP_EXISTING_IMAGES = "jib.skipExistingImages";
+
+  @VisibleForTesting
+  static final String USE_WILDCARD_FOR_DEPENDENCY_CLASSPATH =
+      "jib.useWildcardForDependencyClasspath";
 
   /**
    * Gets the HTTP connection/read timeouts for registry interactions in milliseconds. This is
@@ -122,6 +125,17 @@ public class JibSystemProperties {
    */
   public static boolean skipExistingImages() {
     return Boolean.getBoolean(SKIP_EXISTING_IMAGES);
+  }
+
+  /**
+   * Gets whether or not to use a wildcard or enumerate dependency JARs in the entrypoint runtime
+   * classpath.
+   *
+   * @return {@code true} if Jib should use a wildcard for dependency JARs; {@code false} to
+   *     enumerate all the JARs
+   */
+  public static boolean useWildcardForDependencyClasspath() {
+    return Boolean.getBoolean(USE_WILDCARD_FOR_DEPENDENCY_CLASSPATH);
   }
 
   private static void checkNumericSystemProperty(String property, Range<Integer> validRange) {
--- a/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
+++ b/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
@@ -351,6 +351,15 @@ public class MavenProjectProperties implements ProjectProperties {
   }
 
   @Override
+  public List<Path> getDependencies() {
+    return project
+        .getArtifacts()
+        .stream()
+        .map(artifact -> artifact.getFile().toPath())
+        .collect(Collectors.toList());
+  }
+
+  @Override
   public void waitForLoggingThread() {
     singleThreadedExecutor.shutDownAndAwaitTermination(LOGGING_THREAD_SHUTDOWN_TIMEOUT);
   }
--- a/jib-core/CHANGELOG.md
+++ b/jib-core/CHANGELOG.md
@@ -9,7 +9,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
-- Fixed `NullPointerException` when pulling an OCI base image whose manifest does not have `mediaType` information. ([#2819](https://github.com/GoogleContainerTools/jib/pull/2819))
+- Fixed `NullPointerException` when pulling an OCI base image whose manifest does not have `mediaType` information. ([#2819](https://github.com/GoogleContainerTools/jib/issues/2819))
 - Fixed build failure when using a Docker daemon base image (`docker://...`) that has duplicate layers. ([#2829](https://github.com/GoogleContainerTools/jib/issues/2829))
 
 ## 0.16.0
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java
@@ -24,8 +24,11 @@ import com.google.cloud.tools.jib.configuration.ContainerConfiguration;
 import com.google.cloud.tools.jib.image.Image;
 import com.google.cloud.tools.jib.image.LayerPropertyNotFoundException;
 import com.google.cloud.tools.jib.image.json.HistoryEntry;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.UnmodifiableIterator;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.List;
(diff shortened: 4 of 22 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/global/JibSystemProperties.java
```diff
@@ -32,10 +33,12 @@
   public static final String SEND_CREDENTIALS_OVER_HTTP = "sendCredentialsOverHttp";
   public static final String SERIALIZE = "jib.serialize";
 
-  private static final String DISABLE_USER_AGENT = "_JIB_DISABLE_USER_AGENT";
-
   @VisibleForTesting public static final String SKIP_EXISTING_IMAGES = "jib.skipExistingImages";
 
+  @VisibleForTesting
+  static final String USE_WILDCARD_FOR_DEPENDENCY_CLASSPATH =
+      "jib.useWildcardForDependencyClasspath";
```
loosebazooka, line 40: I realize this is a draft, but I'm not super fond of this name. I sometimes think `preserveClasspathOrder` is a decent name, but i'm not sure.

chanseokoh: Sounds good. And this made me realize that `JibSystemProperties` is for jib-core. The property should be in jib-plugins-common.

chanseokoh: I think it's natural to negate the name, such as `noClasspathOrderPreserving`. The `PropertyNames` class only defines property names, and `System.getBoolean()` doesn't have a way to return `true` as default (which I think makes sense to avoid confusion).

loosebazooka: Hrmm.. maybe in that case, your naming makes more sense. `useClasspathWildcard` is similar but shorter?

Also any idea if we would include this as a config option rather than system property?

loosebazooka: Actually maybe the full name is necessary

chanseokoh: The idea was to have the flag as a temporary escape hatch that will go away soon once we don't get any report from the user. But I think I was too naive. Previous I speculate that we don't have to worry about the argument length limit (https://github.com/GoogleContainerTools/jib/issues/2733#issuecomment-683015187, https://github.com/GoogleContainerTools/jib/issues/2471#issuecomment-630229008). However, after searching "argument list too long" on the Internet today, I completely lost my previous confidence. Now I think the length limit is real and this change is not safe. For example: https://github.com/gradle/gradle/issues/5754.

So it would be safer to have this as an opt-in feature. But that's really really unfortunate, because issues from a different library order are very cryptic and difficult to identify the cause. I really wish Java 8 supported the `@`-argument file feature.

Anyways, if we do this, now I think we should have it as a config option (and additionally a system property). But the question as to whether we expand it by default or not remains.

But ideally, for Java 9+, I think we should really use the `@`-argument file feature; we can always preserve the classpath order while never hitting the system argument length limit. This is going to be a lot bigger change.

In sum, perhaps
- Java 9+: always use a `@`-argument file. (`preserveClasspathOrder` is irrelevant.)
- Java <=8: expand or not based on `preserveClasspathOrder`. Not decisive on if it should be true by default.

loosebazooka: I think false by default works (as it is the current behavior) for java8-

chanseokoh:
Now out of draft.

Renamed and relocated the flag to `jib.container.expandClasspathDependencies` (affects the `entrypoint` container config). Made the default to false (keeping the current behavior to use `.../libs/*`.

I think this feature can go in independently of using the `@argument` file.

Need to update the public usage doc after we release 2.7.0.

chanseokoh:
Update the title.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
```diff
@@ -350,6 +350,15 @@ public JibContainerBuilder createJibContainerBuilder(
     return new DirectoryWalker(Paths.get(project.getBuild().getOutputDirectory())).walk().asList();
   }
 
+  @Override
+  public List<Path> getDependencies() {
+    return project
+        .getArtifacts()
+        .stream()
+        .map(artifact -> artifact.getFile().toPath())
```
mpeddada1, line 358: `GradleProjectProperties` also checks if the dependency is a jar whereas this does not. Is this intentional?

chanseokoh: Good point. I am not sure if all the artifacts from `project.getArtifacts()` will always be `.jar` files, but I think usually they will be `.jar` files for typical Java projects.

I've checked the `MavenProjectProperties` code gain, and I can tell at least that we always put all the files from `project.getArtifacts()` into `/app/libs`. (All of them are added as dependencies and only them.)

https://github.com/GoogleContainerTools/jib/blob/a720aa1fb9d86c3a7115544ee4e3d59602361e12/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java#L291-L299

([`classifyDependencies(...)`](https://github.com/GoogleContainerTools/jib/blob/a720aa1fb9d86c3a7115544ee4e3d59602361e12/jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java#L334-L344) classifies `project.getArtifacts()` into project, snapshot, and third-party dependencies.)

So, doing this should match the actual dependencies we put.

mpeddada1 (commented):
I'm guessing that since the default is now set to current behavior, length restriction for the classpath shouldn't be as big of a problem? Should we still document this as a warning when the user chooses to enable classpath expansion.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MavenProjectProperties.java
chanseokoh: And for Gradle, checking `.jar` is a bit different to obtain dependencies than we current do in `GradleProjectProperties.createJibContainerBuilder()`; it doesn't strictly follow the logic there. However, I think this should work.

chanseokoh:
I think we should shorten the classpath, because we show the entrypoint by default and it will always be extremely long (lie, super super long) whenever the user turns on the option.

chanseokoh:
About the warning, I think you're talking about a potential failure due to line length limit? That's a good idea. I'll document that in `CHANGELOG`. We should also document it when we update our usage doc.

mpeddada1 (approved):

chanseokoh:
Manual testing works correctly. Merging now.

geek-soft:
Great work 👍  The feature solved my issue in quite a few Spring apps.

Is there any plan to release it?

chanseokoh:
We plan to make a release very soon. Will let you know.

chanseokoh:
@geek-soft Jib 2.7.0 has been released with this feature! Let us know if you hit any issues.

PR #2869, opened 2020-10-28
Implement class to generate JibContainerBuilder from contents of jar file

### LINKED ISSUE #2864: [jib-cli-jar] Read a jar and convert it into a JibContainerBuilder
The JibContainerBuilder will then be used to generate a container.

mpeddada1:
Fixes #2864

This will only populate the JibContainerBuilder with the layers and entrypoint derived from the processing the jar. It currently doesn't take into account an option of providing a base image through the cli so it builds the image from scratch.


chanseokoh:
> doesn't take into account an option of providing a base image through the cli

We should pick a reasonable default (currently Distroless as the Jib build plugins do) if a base image is not given.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverter.java
```diff
@@ -0,0 +1,52 @@
...
+   * @return JibContainerBuilder
+   * @throws IOException if I/O error occurs when opening the jar file or if temporary directory
+   *     provided doesn't exist
+   * @throws InvalidImageReferenceException if the base image used is invalid
+   */
+  public static JibContainerBuilder toJibContainerBuilder(Path jarPath, Path tempDirPath)
+      throws IOException, InvalidImageReferenceException {
+
+    // Use distroless as the base image.
+    JibContainerBuilder containerBuilder = Jib.from("gcr.io/distroless/java");
```
chanseokoh, line 45: Can you file an issue to use `gcr.io/distroless/java:11` when the application is Java 11+?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverter.java
```diff
@@ -0,0 +1,52 @@
...
+   *
+   * @param jarPath path to the jar file
+   * @param tempDirPath path to a temporary directory which will be used store the exploded jar's
+   *     contents
+   * @return JibContainerBuilder
+   * @throws IOException if I/O error occurs when opening the jar file or if temporary directory
+   *     provided doesn't exist
+   * @throws InvalidImageReferenceException if the base image used is invalid
+   */
+  public static JibContainerBuilder toJibContainerBuilder(Path jarPath, Path tempDirPath)
```
chanseokoh, line 41: Not always, `to<AnotherClass>()` is conventional for an _instance_ of a class to transform to or provide an instance of another class. And in this case, the method name can be made more clear. Or, maybe, this is a very short code block, and creating a converter class is an overkill? But honestly, I don't grasp the overall design honestly, I am not sure.

And, perhaps this can be package-private? Let's start with limited scope. It's easy to make it `public`, but it's difficult to find methods with unnecessarily larger access.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverterTest.java
```diff
@@ -0,0 +1,111 @@
...
+public class JarToJibContainerBuilderConverterTest {
+  private static final String SIMPLE_STANDARD_JAR = "jar/standard/basicStandardJar.jar";
+
+  @Rule public final TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test
+  public void testToJibContainerBuilder_basicInfo()
+      throws IOException, URISyntaxException, InvalidImageReferenceException {
+    Path standardJar = Paths.get(Resources.getResource(SIMPLE_STANDARD_JAR).toURI());
+    Path destDir = temporaryFolder.newFolder().toPath();
```
chanseokoh, line 51: super nit
```suggestion
    Path destDir = temporaryFolder.getRoot().toPath();
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverterTest.java
```diff
@@ -0,0 +1,111 @@
...
+    assertThat(buildPlan.getLabels()).isEmpty();
+    assertThat(buildPlan.getVolumes()).isEmpty();
+    assertThat(buildPlan.getExposedPorts()).isEmpty();
+    assertThat(buildPlan.getUser()).isNull();
+    assertThat(buildPlan.getWorkingDirectory()).isNull();
+    assertThat(buildPlan.getEntrypoint())
+        .isEqualTo(
+            ImmutableList.of("java", "-cp", "/app/explodedJar:/app/dependencies/*", "HelloWorld"));
+    assertThat(
+            ((FileEntriesLayer) buildPlan.getLayers().get(0))
```
chanseokoh, line 70: Why is the assertion for `.get(0)` looks different than other layers below?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverter.java
mpeddada1: I agree, the naming was a little unclear. I modified the class name to be a bit more consistent with `BuildFiles` so that the method will be called in this format -> `JarFiles.toJibContainerBuilder.` But let me know if this looks okay? The method is very small now, but the plan is to eventually pass in `JibCli` options as another parameter, which will be used to set those extra image parameters (such as environment, etc.). 

Ah since this method is going to be accessed in the `com...jib.cli.cli2` package when we are containerizing the application after the jar subcommand is called, I decided to make it public. Do you think this change is a little premature? Or in other words, is it better to increase the scope only when we actually use it elsewhere.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverterTest.java
mpeddada1: oops, my bad

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverter.java
mpeddada1: Good call. Created #2871.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverter.java
chanseokoh: Hmm... what if this methods goes into `JarProcessor`? Do you have a specific design principle in mind that you wanted to separate it out?

And if it is certain that this will be called from a different package, we can leave it as `public`. I just wasn't sure and thought the possibility this may be called in the same package.

mpeddada1: That is an interesting idea. I could be thinking about this too early but I was slightly reluctant on making `JarProcessor` do too many things at once. Additionally, since we're also planning on adding more modes in the future, this would result in the class become pretty long - so I decided to have one class that does just the processing of the jar and another one that does the generating of `JibContainerBuilder` (and incorporating cli params in the future)
I totally understand that the class may seem a little small right now but once we have more modes + jar types, this class would probably need to do the following:
1. Get the mode from JibCli options (default would be exploded mode).
2. Call `JarProcessor.determineJarType` to determine the jar type. 
3. Have some logic as to which `JarProcessor` processing method to call (either explodeJar or packageJar) based on the mode and the type of jar.
4. [In the future] Incorporate optional cli parameters.

chanseokoh: That makes sense. I just felt that the name `JarFiles` is too generic and easily misunderstood as a general JAR file utility, where this is really specific to instantiate and configure `JibContainerBuilder`. I really wanted to avoid `JarFiles` and tried to think of a better name, and then got to the idea of putting it into `JarProcessor`. However, even after thinking about the name for a long time, I can't really suggest something good. Otherwise, LGTM.

chanseokoh: And I guess the name will really depend on what this class will do going forward and what methods it will have at the end.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
```diff
@@ -0,0 +1,108 @@
...
+    assertThat(buildPlan.getBaseImage()).isEqualTo("gcr.io/distroless/java");
+    assertThat(buildPlan.getPlatforms()).isEqualTo(ImmutableSet.of(new Platform("amd64", "linux")));
+    assertThat(buildPlan.getCreationTime()).isEqualTo(Instant.EPOCH);
+    assertThat(buildPlan.getFormat()).isEqualTo(ImageFormat.Docker);
+    assertThat(buildPlan.getEnvironment()).isEmpty();
+    assertThat(buildPlan.getLabels()).isEmpty();
+    assertThat(buildPlan.getVolumes()).isEmpty();
+    assertThat(buildPlan.getExposedPorts()).isEmpty();
+    assertThat(buildPlan.getUser()).isNull();
+    assertThat(buildPlan.getWorkingDirectory()).isNull();
```
chanseokoh, line 62: Food for thought: these are certainly interesting things, and I guess the user should be allowed to configure them. Then we get back to the same question we had for the usual Jib build command. Originally, everything was through command-line options and we didn't have a "build file." This makes me think, maybe, the `jib jar` subcommand should also support a "build file." Then the `layers` field should probably be interpreted in a different way. I don't know.

chanseokoh (approved):
Hope we find a better name than `JarFiles` eventually.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
mpeddada1: Yeah, that's definitely an interesting option to consider. The advantage of this would be that the user would be allowed to configure these params without having write a really long command. On the other hand (maybe it is too early to tell) could this potentially blur the line between the `jib build` command and the `jib jar` command or cause some confusion? For example, if I need the build file anyways, I might be better off using the `jib build` command instead? --> I could be missing some viable use cases here. 

This also makes me think: Since our general idea for the jar command is to require minimal user input, is it worth picking some optimal defaults (like workingDirectory or user,  for example)?

chanseokoh: > if I need the build file anyways, I might be better off using the jib build command instead?

`jib jar` still does way more things, like optimally creating multi layers (especially having a more tailored logic for Spring Boot), setting an entrypoint with right classpath, etc. And I envision that it will provide a nice getting-started experience and cover simple use cases without having to write a build file. At the same time, it can also support a build file, so that people who need advanced and complex configurations can consider that as an option for convenience.

And definitely `jib jar` is designed to be Java and JAR-specific, so it makes sense to set optimal defaults for many configurations. I actually think we aim for that. We can be highly opinionated on that. Actually, we are already so by creating layers as we think is better.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarToJibContainerBuilderConverter.java
mpeddada1: That is true, `JarFiles` is not the best name. How does `JarToJibContainerBuilderPopulator` sound?

chanseokoh: If the class is for configuring `JibContainerBuilder` (and only for it), I think `JarProcessor.computeEntrypointForExplodedStandard()` should be in this class. I'm a bit confused with the boundary between this class and `JarProcessor`. Maybe after moving `computeEntrypointForExplodedStandard()`, rename `JarProcessor` to `LayerPopulator` (the class seems only specific to create layers for different modes) and rename this class to `JarProcessor`? This class seems to be more general than "LayerPopulator", where the latter is at a more lower level specific to creating layers used by this class.

mpeddada1: As discussed offline, since the methods to compute the entrypoint also depend on the modes of processing the jar, will keep these methods as they are and rename `JarProcessor` to `JarModeProcessor` (since this class focuses on processing the jar based on particular modes).

PR #2872, opened 2020-10-29
Rename JarProcessor method and adjust method access

chanseokoh:
- `determineJarType` would have been `private` if it were not for tests.
- `createExplodedModeLayersForStandardJar()`: from the API perspective, what the method does is creating layers. Unzipping a jar is an implementation detail that the caller doesn't need to know about.

What do you think?


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessor.java
```diff
@@ -68,16 +70,16 @@ public static JarType determineJarType(Path jarPath) throws IOException {
   }
 
   /**
-   * Explodes jar and create three layers for classes, resources and dependencies on container.
+   * Creates three layers for classes, resources, and dependencies on container for a standard jar.
```
mpeddada1, line 73: Sorry, this was my fault but how about 
```suggestion
   * Creates layers for dependencies, snapshot dependencies, resources and class on container for a standard jar.
```

mpeddada1 (approved):
Probably need to incorporate changes from master but LGTM. Thank you for making the change!

PR #2876, opened 2020-10-30
Fix broken Jib CLI logging

chanseokoh:
Jib CLI logging was misbehaving, ~~swallowing messages and~~ unnecessarily outputting blank lines:
```
$ jib --target=docker://foo build
Base image 'ubuntu' does not use a specific image digest - build may not be reproducible

The base image requires auth. Trying again for ubuntu...

Using credentials from Docker config (/home/chanseok/.docker/config.json) for ubuntu


Using base image with digest: sha256:fff16eea1a8ae92867721d90c59a75652ea66d29c05294e6e2f898704bdb8cf1


Container entrypoint set to [sh, hello.sh]

Executing tasks:
```

We should be adding message consumers to `ConsoleLoggerBuilder` only when the corresponding log levels should be consumed.

Output after the fix:
```
$ jib --target=docker://foo build
Base image 'ubuntu' does not use a specific image digest - build may not be reproducible
The base image requires auth. Trying again for ubuntu...
Using credentials from Docker config (/home/chanseok/.docker/config.json) for ubuntu
Using base image with digest: sha256:fff16eea1a8ae92867721d90c59a75652ea66d29c05294e6e2f898704bdb8cf1

Container entrypoint set to [sh, hello.sh]
Executing tasks:
```

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/logging/CliLogger.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/logging/CliLogger.java
@@ -48,7 +48,7 @@ public class CliLogger {
         isRichConsole(consoleOutput) && verbosity.atLeast(Verbosity.lifecycle);
     ConsoleLoggerBuilder builder =
         enableRichProgress
-            ? ConsoleLoggerBuilder.rich(executor, true)
+            ? ConsoleLoggerBuilder.rich(executor, false)
             : ConsoleLoggerBuilder.plain(executor);
     if (verbosity.atLeast(Verbosity.error)) {
       builder.error(stderr::println);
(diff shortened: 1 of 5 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/logging/CliLogger.java
```diff
@@ -22,36 +22,54 @@
...
-        isRichConsole
-            ? ConsoleLoggerBuilder.rich(executor, true)
-            : ConsoleLoggerBuilder.plain(executor).progress(cliLogger::lifecycle);
-    builder.error(cliLogger::error);
-    builder.warn(cliLogger::warn);
-    builder.lifecycle(cliLogger::lifecycle);
-    builder.info(cliLogger::info);
-    builder.debug(cliLogger::debug);
+        enableRichProgress
+            ? ConsoleLoggerBuilder.rich(executor, false)
```
chanseokoh, line 51: Flipping `enableTwoCursorUpJump` doesn't actually change output and is not required to fix the issue. However, since there's no need to employ the two-cursor-up-jump trick (was only necessary for Maven), I'm turning it off.

loosebazooka (approved):

PR #2877, opened 2020-10-30
Give better error message on Jib CLI

chanseokoh:
Sometimes, `ex.getMessage()` doesn't give information enough to understand that an exception occurred. For example, file-not-found I/O exception:
```
$ jib --target=docker://foo build
./jib.yaml
$
```

Also, specifying a directory for `-b` doesn't give a good error message.
```
$ jib --target=docker://foo -b /tmp build
java.io.IOException: Is a directory
 at [Source: (org.apache.commons.text.io.StringSubstitutorReader); line: 1, column: 1]
$
```

I decided to give special treatment for checking a build file, as I think first time users will likely get into it a lot. (@mpeddada1 and I both got into this and had to spend a lot of time figuring out what happened.)

I spent some time to write a test for logging in `Build`, but it required me to considerably refactor the class. It didn't seem worth only to be able to write tests, so eventually I reverted the tests.

Lastly, I think it's better to add the prefixes for ERROR and WARN logs to make them clear.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Build.java
```diff
@@ -40,18 +43,29 @@ public Integer call() {
     try {
       ConsoleLogger logger =
           CliLogger.newLogger(globalOptions.getVerbosity(), globalOptions.getConsoleOutput());
+
+      Path buildFile = globalOptions.getBuildFile();
+      if (!Files.isReadable(buildFile)) {
+        logger.log(Level.ERROR, "Cannot open Build File YAML: " + buildFile);
```
mpeddada1, line 49: Maybe we can make the error message a bit more specific - "The Build File YAML either does not exist or cannot be opened for reading" since `isReadable` checks for both.

mpeddada1 (approved):

PR #2878, opened 2020-11-02
Add jar cli sub-command to the jib cli

### LINKED ISSUE #2865: [jib-cli-jar] Build cli jar command.
Convert cli options to containerizer.

mpeddada1:
Fixes #2865.
This sub-command would be called in the following form:
```
jib --target=docker://jib-cli-image jar /path/to/my-jar.jar
```
This PR doesn't include unit tests (as I'm not sure if it is worth refactoring the class as mentioned in #2877 for the error logging test) but the command was tested manually. We still need to add integration testing to check that the command is working as expected.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -0,0 +1,88 @@
...
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "jar", showAtFileInUsageHelp = true, description = "Containerize a jar")
+public class Jar implements Callable<Integer> {
+  @CommandLine.ParentCommand
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  protected JibCli globalOptions;
+
+  @CommandLine.Option(
+      names = {"--jar"},
```
chanseokoh, line 40: I think we should accept this is an ordinary argument, not through an option, e.g., `jib ... jar my.jar`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -0,0 +1,88 @@
...
+      JibContainerBuilder containerBuilder =
+          JarFiles.toJibContainerBuilder(getJarFile(), Paths.get("build-artifacts"));
+      containerBuilder.containerize(containerizer);
+      MoreFiles.deleteDirectoryContents(
+          Paths.get("build-artifacts"), RecursiveDeleteOption.ALLOW_INSECURE);
+    } catch (Exception ex) {
+      if (globalOptions.isStacktrace()) {
+        ex.printStackTrace();
+      }
+      System.err.println(ex.getMessage());
```
chanseokoh, line 83: Should also print the exception type. Check `Build.java`. That said, can we explore a way to share code?

chanseokoh: Or maybe it's not much worth to share this short block. In any case, let's print the exception type.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
mpeddada1: Good catch, thanks.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
mpeddada1: Yeah I was trying to see if there was anything more that I could share but that required me to add some logic as to which sub-command we're using since the steps to containerize are a bit different for both. I'll make the exception more consistent with Build.java though.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -0,0 +1,85 @@
...
+  @CommandLine.Parameters(description = "The path to the jar file (ex: path/to/my-jar.jar)")
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  private Path jarFile;
+
+  /**
+   * Returns a user configured Path to a jar file.
+   *
+   * @return a path to a jar file
+   */
+  public Path getJarFile() {
```
chanseokoh, line 48: Is this method is required?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -0,0 +1,85 @@
...
+      if (Files.isDirectory(jarFile)) {
+        logger.log(
+            LogEvent.Level.ERROR,
+            "The file path provided is for a directory. Please provide a path to a jar file: "
+                + jarFile);
+        return 1;
+      }
+      Containerizer containerizer = Containerizers.from(globalOptions, logger);
+      JibContainerBuilder containerBuilder =
+          JarFiles.toJibContainerBuilder(getJarFile(), Paths.get("build-artifacts"));
```
chanseokoh, line 72: We need to sort out what we should provide for a temporary staging directory. We may eventually go for a dedicated build output directory, but for now, I think we can provide a system temp directory. Let's talk offline.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -40,22 +38,15 @@
...
-  public Path getJarFile() {
-    return jarFile;
-  }
+  private final TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider();
 
   @Override
   public Integer call() {
     globalOptions.validate();
-    Path jarFile = getJarFile();
     try {
```
chanseokoh, line 46:
```suggestion
    try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
```
and remove the field and the `finally` block.

mpeddada1: oh this is much better

chanseokoh (approved):

PR #2885, opened 2020-11-05
Add integration testing for jib cli jar command. 

### LINKED ISSUE #2880: [jib-cli-jar] Add integration testing for jib cli jar command.

mpeddada1:
Fixes #2880

Initially, we weren't terminating the [executor service](https://github.com/GoogleContainerTools/jib/blob/9cbad71ccb00c4d4ef473027d5866605a22730b9/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/logging/SingleThreadedExecutor.java) which is used by CliLogger. This was causing some unexpected behavior in tests that were testing error logging.


jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -0,0 +1,94 @@
...
+  private final PrintStream originalOut = System.out;
+  private final PrintStream originalErr = System.err;
+  private final ByteArrayOutputStream out = new ByteArrayOutputStream();
+  private final ByteArrayOutputStream err = new ByteArrayOutputStream();
+
+  @Before
+  public void setUp() {
+    out.reset();
+    err.reset();
+    System.setOut(new PrintStream(out));
```
chanseokoh, line 48: This is a creative trick, but I feel it isn't really safe to replace stdout and stderr. For example, the unit framework depends on it, and I think we will not be able to see test failures.

I have been thinking of two options for verifying command output: run as a separate process, or refactor the code. I didn't really like refactoring code only to be able to check log messages. Eventually I think we will reach the point where we really need to check log messages. And we will eventually find a way to run Jib CLI as a standalone process. Maybe for now, if we can't easily find a way to check log messages, I'm fine punting it.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Test
+  public void testErrorLogging_directoryGiven() throws URISyntaxException {
+    Path jarFile = Paths.get(Resources.getResource("emptyDir").toURI());
+    Integer actual =
+        new CommandLine(new JibCli())
+            .execute("--target", "docker://jib-cli-image", "jar", jarFile.toString());
+
+    assertThat(actual).isEqualTo(FAILURE_CODE);
```
chanseokoh, line 76: Let's just go for
```suggestion
    assertThat(exitCode).isEqualTo(1);
```

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
mpeddada1: Ah yeah this doesn't seem too great. Maybe we should create an issue to make the logger use `PrintWriter` instead of printing to `System.err` or `System.out` - https://picocli.info/#_testing_your_application.

chanseokoh: Oh, nice documentation. I think `setOut` and `setErr` will just work without us having to change the logger. Have you tried that?

mpeddada1: I agree haha. And sorry, are you referring to this approach?:
```
MyApp app = new MyApp();
CommandLine cmd = new CommandLine(app);
StringWriter sw = new StringWriter();
cmd.setOut(new PrintWriter(sw));
```

chanseokoh: Yes. I feel like this will work.

mpeddada1: When I had attempted this approach the output and the error messages could not be set using `setOut/Err` and were getting printed onto the console. I think it's because we are currently directly printing to `System.out/err` in the logger instead of writing to the picocli writer? `This assumes that the application uses the PrintWriter provided by CommandLine.getOut or CommandLine.getErr`

chanseokoh: Ah, you're right. Let's change the logger. I think it'll be pretty straightforward.

mpeddada1: Sounds good! Added #2886.

chanseokoh: :+1: Let's fix #2886 first before this PR.

mpeddada1: Done.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -0,0 +1,76 @@
...
+        .isEqualTo("[ERROR] The file path provided does not exist: unknown.jar\n");
+  }
+
+  @Test
+  public void testErrorLogging_directoryGiven() throws URISyntaxException {
+    CommandLine jibCli = new CommandLine(new JibCli());
+    StringWriter sw = new StringWriter();
+    jibCli.setErr(new PrintWriter(sw));
+
+    Path jarFile = Paths.get(Resources.getResource("emptyDir").toURI());
```
chanseokoh, line 53: I don't think we have "emptyDir" in our test runtime classpath. Maybe just `Paths.get("/")`?

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.nio.file.Paths;
+import org.junit.Test;
+import picocli.CommandLine;
+
+public class JarCommandTest {
+
+  @Test
+  public void testErrorLogging_fileDoesNotExist() {
+    CommandLine jibCli = new CommandLine(new JibCli());
+    StringWriter sw = new StringWriter();
```
chanseokoh, line 37: The Google style doesn't allow 'sw'. I think we can go for
```suggestion
    StringWriter stringWriter = new StringWriter();
```

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Test
+  public void testJar_toDocker() throws IOException, InterruptedException, URISyntaxException {
+    Path jarFile = Paths.get(Resources.getResource("simpleJar.jar").toURI());
+    Integer actual =
+        new CommandLine(new JibCli())
+            .execute("--target", "docker://jib-cli-image", "jar", jarFile.toString());
+    String output = new Command("docker", "run", "--rm", "jib-cli-image").run();
+
+    assertThat(actual).isEqualTo(0);
```
chanseokoh, line 73: exitCode

chanseokoh (approved):

PR #2887, opened 2020-11-06
Modify CliLogger to use PrintWriter

### LINKED ISSUE #2886: [jib-cli] Modify CliLogger to use PrintWriter
We are currently writing logging information directly to System.out and System.err. Modify CliLogger to write to `PrintWriter` using `getOut/Err`. This is to make our tests both cleaner and safer (as it allows us to avoid replacing `stdout` and `stderr` for the sake of testing). Implementation tips are provided in the following doc: https://picocli.info/#_testing_your_application

mpeddada1:
Fixes #2886


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -41,13 +42,20 @@
   @SuppressWarnings("NullAway.Init") // initialized by picocli
   private Path jarFile;
 
+  private final SingleThreadedExecutor executor = new SingleThreadedExecutor();
```
chanseokoh, line 45: I don't think this needs to be a field. Just instantiate as a local variable in `call()`. I know you'll going to use it in a `finally` block, so you can create it before `try {`.

jib-cli/src/integration-test/JarCommandTest.java
```diff
@@ -0,0 +1 @@
+
```
chanseokoh, line 1: Remove the empty file.

mpeddada1: oops, sorry about that

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -23,13 +23,17 @@
...
+import com.google.cloud.tools.jib.plugins.common.logging.SingleThreadedExecutor;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.Callable;
 import picocli.CommandLine;
+import picocli.CommandLine.Model.CommandSpec;
 
 @CommandLine.Command(name = "jar", showAtFileInUsageHelp = true, description = "Containerize a jar")
 public class Jar implements Callable<Integer> {
+  @CommandLine.Spec CommandSpec spec = CommandSpec.create();
```
chanseokoh, line 35: Probably

```suggestion
  @CommandLine.Spec protected CommandSpec spec = CommandSpec.create();
```

I guess it can't be `private`?

chanseokoh (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
mpeddada1: great catch. On second thought, I think it should be private since it's only accessed in this class.

chanseokoh: Appu put `globalOptions` as `protected`, so probably there's a reason for that. But I'm not sure about `CommandSpec`. Just do some actual `jib jar` run to confirm you see correct output.

mpeddada1: Ah good point, running ` jib --target=docker://jib-cli-image jar unknown.jar` returned the expected output : `[ERROR] The file path provided does not exist: unknown.jar`. I think it could be okay?

chanseokoh (approved):
Yeah, looks good.

Issue #2846, opened 2020-10-19
[jib-cli-jar] Implement 'packaged' mode for standard jar.


mpeddada1:
Option to containerize a jar as it is.

mpeddada1:
Closed with #2888

PR #2888, opened 2020-11-09
Implement packaged mode for standard jar.

mpeddada1:
Implementation of #2846 
Adds cli option to set the processing mode, implements packaged mode and adds integration testing. "Exploded" mode is set as the default.
Requires #2885.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -43,6 +44,14 @@
   @SuppressWarnings("NullAway.Init") // initialized by picocli
   private Path jarFile;
 
+  @CommandLine.Option(
+      names = "--mode",
+      defaultValue = "exploded",
+      paramLabel = "<mode>",
+      description = "The processing mode")
```
chanseokoh, line 51: I think this will be what the user will see on the command-line with the help usage. Should include what values are available.

That said, take a look at the `--vebosity` and `--console` options.

```java
  @Option(
      names = "--verbosity",
      paramLabel = "<level>",
      defaultValue = "lifecycle",
      description =
          "set logging verbosity, candidates: ${COMPLETION-CANDIDATES}, default: ${DEFAULT-VALUE}")
  private Verbosity verbosity;

  @Option(
      names = "--console",
      paramLabel = "<type>",
      defaultValue = "auto",
      description =
          "set console output type, candidates: ${COMPLETION-CANDIDATES}, default: ${DEFAULT-VALUE}")
  private ConsoleOutput consoleOutput;
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -191,4 +208,47 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
...
+              .collect(Collectors.toList());
+      List<Path> snapshotDependencies =
+          allDependencies.stream().filter(isSnapshot).map(Paths::get).collect(Collectors.toList());
+      if (!nonSnapshotDependencies.isEmpty()) {
+        FileEntriesLayer.Builder nonSnapshotDependenciesLayerBuilder =
+            FileEntriesLayer.builder().setName(DEPENDENCIES);
+        nonSnapshotDependencies.forEach(
+            path ->
+                nonSnapshotDependenciesLayerBuilder.addEntry(
+                    path, APP_ROOT.resolve(RelativeUnixPath.get("dependencies")).resolve(path)));
```
chanseokoh, line 239: I think missed this. `jib jar` will fail if running from a different directory (e.g., `jib ... jar ../my.jar`). Perhaps the first argument to `addEntry` should be `jarPath.resolve(path)`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -86,47 +87,13 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
-            FileEntriesLayer.builder().setName(SNAPSHOT_DEPENDENCIES);
-        snapshotDependencies.forEach(
-            path ->
-                snapshotDependenciesLayerBuilder.addEntry(
-                    path, APP_ROOT.resolve(RelativeUnixPath.get("dependencies")).resolve(path)));
-        layers.add(snapshotDependenciesLayerBuilder.build());
-      }
-    }
+    // Add dependencies layers.
+    layers.addAll(getDependenciesLayers(jarPath));
```
chanseokoh, line 93: In this case, I think it's fine to have
```suggestion
    List<FileEntriesLayer> layers = getDependenciesLayers(jarPath);
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -173,6 +168,28 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+   * @throws IOException if I/O error occurs when opening the jar file
+   */
+  static ImmutableList<String> computeEntrypointForPackagedStandard(Path jarPath)
+      throws IOException {
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      String mainClass =
+          jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.MAIN_CLASS);
+      if (mainClass == null) {
+        throw new IllegalArgumentException(
+            "`Main-Class:` attribute for an application main class not defined in the input Jar's "
```
chanseokoh, line 185: Not that I care, but just FYI, this [doc](https://docs.oracle.com/javase/tutorial/deployment/jar/basicsindex.html) stylizes the archive as "JAR." It stands for "Java™ Archive (JAR)" files.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: Ah great catch. To make sure that I'm understanding this correctly, would it need to be jarRoot.resolve(path) where jarRoot is the directory where the jar is located?

chanseokoh: I actually tried this. The command works at `~/A` but fails at `~`.

```
~/A$ pwd
/home/chanseok/A
~/A$ ls C.jar guava-20.0.jar 
C.jar  guava-20.0.jar
~/A$ jib --target=docker://foo jar C.jar
[WARN] Base image 'gcr.io/distroless/java' does not use a specific image digest - build may not be reproducible
Using base image with digest: sha256:90596f62a8ec0701d00845b85dfccf7f53b5e0bb3c49e275d7f19ea6c3e4565d

Container entrypoint set to [java, -cp, /app/explodedJar:/app/dependencies/*, C]
Executing tasks:

~/A$ cd ..
~$ jib --target=docker://foo jar A/C.jar
[WARN] Base image 'gcr.io/distroless/java' does not use a specific image digest - build may not be reproducible
Using base image with digest: sha256:90596f62a8ec0701d00845b85dfccf7f53b5e0bb3c49e275d7f19ea6c3e4565d
Executing tasks:
[==================            ] 60.0% complete
> scheduling building manifests

java.nio.file.NoSuchFileException: guava-20.0.jar
```

And yeah, perhaps the source path should be something like `jarPath.getParent().resolve(path)`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: Oh nice, I will follow the convention in that case.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -151,6 +116,32 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+    // Add dependencies layers.
+    List<FileEntriesLayer> layers = getDependenciesLayers(jarPath);
+
+    // Add layer for jar.
+    FileEntriesLayer jarLayer =
+        FileEntriesLayer.builder()
+            .setName(JAR)
+            .addEntry(
+                jarPath,
+                APP_ROOT.resolve(RelativeUnixPath.get("jar")).resolve(jarPath.getFileName()))
```
chanseokoh, line 138: Oh, I remember we decided to do `java -jar`. In that case, we should retain the directory structure for the dependencies given in `Class-Path:` in `MANIFEST.MF`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -44,6 +44,14 @@
   @SuppressWarnings("NullAway.Init") // initialized by picocli
   private Path jarFile;
 
+  @CommandLine.Option(
+      names = "--mode",
+      defaultValue = "exploded",
+      paramLabel = "<mode>",
+      description = "The jar processing mode, candidates: packaged, default: exploded")
```
chanseokoh, line 51: Why not creating a dedicated type and have the descriptions like other options?
```
"set console output type, candidates: ${COMPLETION-CANDIDATES}, default: ${DEFAULT-VALUE}")
```
I think it will take care of validation for us too.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
chanseokoh: Actually, I do see the dependency directory structure is retained under `/app/dependencies`. But for the `java -jar` to work, the easiest solution is to put the main JAR under `/app/dependencies`. But then, the directory name `/app/dependencies` doesn't make much sense. Maybe we should place the main JAR and other dependency JARs (while retaining the directory structure) at `/app`?

That said, it's worth having a test that has `Class-Path:`, e.g., `Class-Path: foo.jar nested/bar.jar`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -191,4 +204,50 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
...
+      List<Path> snapshotDependencies =
+          allDependencies.stream().filter(isSnapshot).map(Paths::get).collect(Collectors.toList());
+      Path jarParent = jarPath.getParent();
+      if (!nonSnapshotDependencies.isEmpty()) {
+        FileEntriesLayer.Builder nonSnapshotDependenciesLayerBuilder =
+            FileEntriesLayer.builder().setName(DEPENDENCIES);
+        nonSnapshotDependencies.forEach(
+            path ->
+                nonSnapshotDependenciesLayerBuilder.addEntry(
+                    jarParent.resolve(path),
```
chanseokoh, line 236: I just tried my `C.jar` example, but it throws an NPE. In this case, `jarPath` is simply `C.jar` (not an absolute path) and `jarPath.getParent()` returns null.

```
~/A$ jib --target=docker://foo --stacktrace jar C.jar
java.lang.NullPointerException
        at com.google.cloud.tools.jib.cli.jar.JarModeProcessor.lambda$getDependenciesLayers$6(JarModeProcessor.java:236)
        at java.util.ArrayList.forEach(ArrayList.java:1257)
        at com.google.cloud.tools.jib.cli.jar.JarModeProcessor.getDependenciesLayers(JarModeProcessor.java:233)
```

Maybe it can be
```java
      Path jarParent = jarPath.getParent() == null ? Paths.get("") : jarPath.getParent();
```

mpeddada1: I ran into the same issue, you're right. Okay I will fix this and add some tests for jars that use dependencies.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -57,20 +57,52 @@ public void testErrorLogging_directoryGiven() throws URISyntaxException {
...
+            .execute(
+                "--target", "docker://jib-cli-image", "jar", jarFile.toString(), "--mode=packaged");
+    String output = new Command("docker", "run", "--rm", "jib-cli-image").run();
+
+    assertThat(exitCode).isEqualTo(0);
+    assertThat(output).isEqualTo("HelloWorld");
+  }
+
+  @Test
+  public void testJar_unknownMode_toDocker() throws URISyntaxException {
```
chanseokoh, line 93:
```suggestion
  public void testJar_unknownMode() throws URISyntaxException {
```

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -57,20 +57,52 @@ public void testErrorLogging_directoryGiven() throws URISyntaxException {
...
+    CommandLine jibCli = new CommandLine(new JibCli());
+    StringWriter stringWriter = new StringWriter();
+    jibCli.setErr(new PrintWriter(stringWriter));
+
+    Path jarFile = Paths.get(Resources.getResource("jarTest/jarWithCp.jar").toURI());
```
chanseokoh, line 98: Perhaps we don't even need to pass a jar file?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/ProcessingMode.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.tools.jib.cli.jar;
+
+public enum ProcessingMode {
+  packaged,
```
chanseokoh, line 20: Let's have exploded (default mode) come first.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
```diff
@@ -33,20 +33,28 @@
...
     JibContainerBuilder containerBuilder = Jib.from("gcr.io/distroless/java");
 
-    List<FileEntriesLayer> layers =
-        JarModeProcessor.createExplodedModeLayersForStandardJar(jarPath, tempDirPath);
-    List<String> entrypoint = JarModeProcessor.computeEntrypointForExplodedStandard(jarPath);
+    List<FileEntriesLayer> layers;
+    List<String> entrypoint;
+    if (mode.equals(ProcessingMode.packaged)) {
+      layers = JarModeProcessor.createPackagedModeLayersForStandardJar(jarPath);
+      entrypoint = JarModeProcessor.computeEntrypointForPackagedStandard(jarPath);
```
chanseokoh, line 53: Maybe we can be consistent with the names. Given that the class name is already `JarModeProcessor`,

```suggestion
      layers = JarModeProcessor.createLayersForPackagedStandard(jarPath);
      entrypoint = JarModeProcessor.computeEntrypointForPackagedStandard(jarPath);
```
And unfortunately, renaming a method means renaming test methods too.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -191,4 +202,49 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
...
+    String classPath = null;
+
+    // Get dependencies from Class-Path in the jar's manifest and add a layer each for non-snapshot
+    // and snapshot dependencies. If Class-Path is not present in the jar's manifest then skip
+    // adding the dependencies layers.
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      classPath = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
+    }
+    if (classPath != null) {
+      Predicate<String> isSnapshot = name -> name.contains("SNAPSHOT");
```
chanseokoh, line 218: Now that we have refactored out this logic to this method, short-circuiting is possible, and I think it helps readability. And now I think it should be fine to expand the `try` block until the end. For example,

```suggestion
    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
      String classPath = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
      if (classPath == null) {
        return new ArrayList<>();
      }

      List<FileEntriesLayer> layers = new ArrayList<>();
      ...
    }
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -191,4 +202,49 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
...
+          allDependencies
+              .stream()
+              .filter(isSnapshot.negate())
+              .map(Paths::get)
+              .collect(Collectors.toList());
+      List<Path> snapshotDependencies =
+          allDependencies.stream().filter(isSnapshot).map(Paths::get).collect(Collectors.toList());
+      Path jarParent = jarPath.getParent() == null ? Paths.get("") : jarPath.getParent();
+      if (!nonSnapshotDependencies.isEmpty()) {
+        FileEntriesLayer.Builder nonSnapshotDependenciesLayerBuilder =
```
chanseokoh, line 230: super nit: I think in this local scope, it's fine to name it `nonSnapshotLayer`. I think it'll help reduce cognitive load when reading the code block.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -191,4 +202,49 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
...
+            FileEntriesLayer.builder().setName(DEPENDENCIES);
+        nonSnapshotDependencies.forEach(
+            path ->
+                nonSnapshotDependenciesLayerBuilder.addEntry(
+                    jarParent.resolve(path), pathOnContainer.resolve(path)));
+        layers.add(nonSnapshotDependenciesLayerBuilder.build());
+      }
+      if (!snapshotDependencies.isEmpty()) {
+        FileEntriesLayer.Builder snapshotDependenciesLayerBuilder =
```
chanseokoh, line 239: Likewise, `snapshotLayer`.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -326,6 +326,125 @@ public void testExplodedMode_standard_computeEntrypoint_noMainClass() throws URI
...
+
+    FileEntriesLayer jarLayer = layers.get(0);
+    assertThat(jarLayer.getName()).isEqualTo("jar");
+    assertThat(
+            jarLayer
+                .getEntries()
+                .stream()
+                .map(FileEntry::getExtractionPath)
+                .collect(Collectors.toList())
+                .get(0))
```
chanseokoh, line 349: `jarLayer().getEntries().get(0).getExtractionPath()`?

And before this, also assert that the size of `jarLayer().getEntries()` is 1.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -326,6 +326,125 @@ public void testExplodedMode_standard_computeEntrypoint_noMainClass() throws URI
...
+        Paths.get(Resources.getResource(STANDARD_JAR_WITHOUT_CLASS_PATH_MANIFEST).toURI());
+    List<FileEntriesLayer> layers =
+        JarModeProcessor.createPackagedModeLayersForStandardJar(standardJar);
+
+    assertThat(layers.size()).isEqualTo(1);
+
+    FileEntriesLayer jarLayer = layers.get(0);
+
+    assertThat(jarLayer.getName()).isEqualTo("jar");
+    assertThat(
```
chanseokoh, line 366: ditto

But this test is identical to the one above. May be remove this test?

chanseokoh:
The Guava JAR is a bit large. How about just creating a JAR containing only one class and have the main JAR reference that?

chanseokoh:
And it's opaque how the new test JAR uses its dependencies. I think adding assertions for the value of `Class-Path:` (and possibly the relative locations of dep JARs too if that helps) will help.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -86,47 +87,12 @@ static JarType determineJarType(Path jarPath) throws IOException {
    */
   static List<FileEntriesLayer> createExplodedModeLayersForStandardJar(
       Path jarPath, Path tempDirPath) throws IOException {
+    // Add dependencies layers.
+    List<FileEntriesLayer> layers =
+        getDependenciesLayers(jarPath, APP_ROOT.resolve(RelativeUnixPath.get("dependencies")));
```
chanseokoh, line 92:
```suggestion
        getDependenciesLayers(jarPath, APP_ROOT.resolve("dependencies"));
```

mpeddada1:
This PR is ready for review.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -163,14 +223,51 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+      if (mainClass == null) {
+        throw new IllegalArgumentException(
+            "`Main-Class:` attribute for an application main class not defined in the input JAR's "
+                + "manifest (`META-INF/MANIFEST.MF` in the JAR).");
+      }
+      return ImmutableList.of("java", "-jar", APP_ROOT + "/" + jarPath.getFileName().toString());
+    }
+  }
+
+  private static ImmutableMap<String, List<Path>> getSnapshotAndNonSnapshotDependencies(
```
chanseokoh, line 255: I'm not a big fan of using a map for the purpose of returning more than one value. How about creating a private inner class? For this purpose, I think we don't even need getters and setters.

```java
private static class LocalDependencies {
  private final List<Path> snapshots;
  private final List<Path> nonSnapshots;

  private LocalDependencies(List<Path> snapshots, List<Path> nonSnapshots) {
    this.snapshots = snapshots;
    this.nonSnapshots = nonSnapshots;
  }
}
```

But is it not possible for this method to return built layers? (I remember the original code worked in that way, and I liked it.) In that way, I think we can remove duplicate code between the exploded mode and the packaged mode?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
```diff
@@ -106,4 +107,41 @@ public void testToJibContainerBuilder_basicInfo()
...
+    assertThat(buildPlan.getFormat()).isEqualTo(ImageFormat.Docker);
+    assertThat(buildPlan.getEnvironment()).isEmpty();
+    assertThat(buildPlan.getLabels()).isEmpty();
+    assertThat(buildPlan.getVolumes()).isEmpty();
+    assertThat(buildPlan.getExposedPorts()).isEmpty();
+    assertThat(buildPlan.getUser()).isNull();
+    assertThat(buildPlan.getWorkingDirectory()).isNull();
+    assertThat(buildPlan.getEntrypoint())
+        .isEqualTo(ImmutableList.of("java", "-jar", "/app/basicStandardJar.jar"));
+    assertThat(((FileEntriesLayer) buildPlan.getLayers().get(0)).getEntries())
```
chanseokoh, line 132: Assert that `buildPlan.getLayers()` is of size 2 before this.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -295,7 +294,21 @@ public void testCreateExplodedModeLayersForStandardJar_dependencyDoesNotExist()
   }
 
   @Test
-  public void testExplodeMode_standard_computeEntrypoint_allLayersPresent()
+  public void testComputeEntrypointForExplodedStandard_noMainClass() throws URISyntaxException {
```
chanseokoh, line 297: Probably no one throws `URISyntaxException`. Try removing it.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -83,58 +84,16 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
-                    snapshotDependenciesLayerBuilder,
-                    jarParent.resolve(path),
-                    APP_ROOT.resolve("dependencies").resolve(path.getFileName())));
-        layers.add(snapshotDependenciesLayerBuilder.build());
-      }
-    }
+  static List<FileEntriesLayer> createLayersForExplodedStandard(Path jarPath, Path tempDirPath)
+      throws IOException {
+    // Add dependencies layers.
+    List<FileEntriesLayer> layers = getDependenciesLayers(jarPath, "exploded");
```
chanseokoh, line 90: We have the enum class defined in this Java package (`c.g.c.t.jib.cli.jar.ProcessingMode`). Why not use it?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -163,14 +145,92 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+      throws IOException {
+    List<FileEntriesLayer> layers = new ArrayList<>();
+
+    // Get dependencies from Class-Path in the jar's manifest and add a layer each for non-snapshot
+    // and snapshot dependencies. If Class-Path is not present in the jar's manifest then skip
+    // adding the dependencies layers.
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      String classPath =
+          jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
+      if (classPath != null) {
```
chanseokoh, line 187: I think short-circuiting will help readability.
```java
      if (classPath == null) {  // should flip the comparison
        return new ArrayList<>();
      }
      List<FileEntriesLayer> layers = new ArrayList<>();
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -163,14 +145,92 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+        Path jarParent = jarPath.getParent() == null ? Paths.get("") : jarPath.getParent();
+        Predicate<String> isSnapshot = name -> name.contains("SNAPSHOT");
+        List<String> allDependencies = Splitter.onPattern("\\s+").splitToList(classPath.trim());
+        List<Path> nonSnapshotDependencies =
```
chanseokoh, line 191: nit: let's just call these `nonShapshots`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -163,14 +145,92 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+                .map(Paths::get)
+                .collect(Collectors.toList());
+        List<Path> snapshotDependencies =
+            allDependencies
+                .stream()
+                .filter(isSnapshot)
+                .map(Paths::get)
+                .collect(Collectors.toList());
+        if (!nonSnapshotDependencies.isEmpty()) {
+          FileEntriesLayer.Builder nonSnapshotDependenciesLayerBuilder =
```
chanseokoh, line 204: nit: `nonSnapshotLayer`

chanseokoh (commented):
LGTM

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: Oh yes, good idea!

chanseokoh (approved):

PR #2890, opened 2020-11-10
Fix ordering of options in Jib Cli

mpeddada1:
Use mixins to share options between subcommands.
Usage:
`jib jar --target=<image-target> <my-jar>.jar`
or
`jib --target=<image-target> jar <my-jar>.jar`
This PR makes a few changes:
- Introduces class that contains all the common cli options.
- Adds the common cli options to the subcommands with the help of the `@Mixin` annotation. 
- Moves options that are specific to the build subcommand to `Build.java`(specifically, `--context`, `--build-file` and `--parameter`)
- Creates `CredentialsForJarTest` and `CredentialsForBuildTest`, which will be the same as `CredentialsTest` but for the subcommands.
- Creates `ContainerizersForJarTest` and `ContainerizersForBuildTest` instead of `ContainerizersTest`.
- Creates `BuildTest` and `JarTest` instead of `JibCliTest`.


chanseokoh:
Great to find this feature. I guess we need to add this option to almost all parameters (e.g., `--verbosity`, `--credential-helper`, etc).

loosebazooka:
Oh this is interesting, so options can go anywhere?

mpeddada1 (commented):
Yeah, it can be specified on the top-level command or anywhere on the subcommand. For example, `jib jar <my-jar>.jar --target=<target>` should also work.

chanseokoh:
I think there is a different implication than using a mixins. For example, I think every new command will inherit the same set of options, which we may not want in the future. And it still allows specifying options before sub-commands. But I think this works as a good solution now.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/JibCli.java
```diff
@@ -83,7 +85,8 @@
       required = true,
       paramLabel = "<target-image>",
       description =
-          "The destination image reference or jib style url,%nexamples:%n gcr.io/project/image,%n registry://image-ref,%n docker://image,%n tar://path")
+          "The destination image reference or jib style url,%nexamples:%n gcr.io/project/image,%n registry://image-ref,%n docker://image,%n tar://path",
+      scope = CommandLine.ScopeType.INHERIT)
   @SuppressWarnings("NullAway.Init") // initialized by picocli
   private String targetImage;
```
mpeddada1, line 92: Should we also move `context`, `build-file` and `parameter` to Build.java?

chanseokoh: Good question. By `parameters`, you mean the three hidden parameters? I think they should be shared. And for `context` and `build-file`, it makes sense to move `Build.java`, since we are not using it in `Jar.java` at the moment.

mpeddada1: Sounds good! Oh sorry by parameter, I meant the `templating parameter`.

chanseokoh: I don't think I'll come through all the options. Just make sure to move options only specific to the `build` command (like those three) to `Build.java`.

BTW, the doc says that `INHERIT` cannot be used in `@ArgGroup`, which we have in some places. But I guess it's irrelevant to us?

mpeddada1: Ah good point. Applying the `INHERIT` feature for the `@ArgGroup`  still seems to be a work in progress. Currently, using `INHERIT` on the options in the arg group allows those options to be inherited by all the subcommands but it doesn't let the other arg group attributes to be inherited (for example, validation for exclusivity). This might be something that we need when we are handling credentials for the `jar` and `build` subcommands. Hm it is probably better to use Mixins for now as support for `INHERIT` is still a little new.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Build.java
```diff
@@ -38,24 +43,73 @@
 
   @CommandLine.Spec private CommandSpec spec = CommandSpec.create();
 
+  @CommandLine.Mixin
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  CommonCliOptions commonCliOptions;
```
chanseokoh, line 48: Can this be `private`? Or `@VisibleForTesting` if required?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Build.java
```diff
@@ -38,24 +43,73 @@
...
+
   @CommandLine.ParentCommand
   @SuppressWarnings("NullAway.Init") // initialized by picocli
   protected JibCli globalOptions;
```
chanseokoh, line 52: Maybe we can remove this field? `JibCli` doesn't have any options anymore, right?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Build.java
```diff
@@ -38,24 +43,73 @@
...
+
+  public Map<String, String> getTemplateParameters() {
+    Verify.verifyNotNull(templateParameters);
+    return templateParameters;
+  }
+
   @Override
   public Integer call() {
-    globalOptions.validate();
+    validate();
```
chanseokoh, line 102: `commonCliOptions.validate()`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Build.java
```diff
@@ -86,4 +139,14 @@ public Integer call() {
     }
     return 0;
   }
+
+  /** Validates parameters defined in this class that could not be done declaratively. */
+  public void validate() {
```
chanseokoh, line 144: No need to duplicate the method if we call `commonCliOptions.validate()` above?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -41,6 +43,10 @@
   @SuppressWarnings("NullAway.Init") // initialized by picocli
   protected JibCli globalOptions;
 
+  @CommandLine.Mixin
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  CommonCliOptions commonCliOptions;
```
chanseokoh, line 48: `private` or `@VisibleForTesting`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -97,4 +103,14 @@ public Integer call() {
     }
     return 0;
   }
+
+  /** Validates parameters defined in this class that could not be done declaratively. */
+  public void validate() {
```
chanseokoh, line 108: ditto?

chanseokoh:
@mpeddada1 I was trying to see if some part of the original code works for the tests and decided to simplify it myself: e1d7115

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Build.java
mpeddada1: Good catch.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Build.java
mpeddada1: Oh yes, thanks for pointing it out.

mpeddada1:
@chanseokoh Thank you so much for cleaning it up! Looks so much better.

chanseokoh (commented):
LGTM

chanseokoh (approved):

PR #2893, opened 2020-11-13
Modify standard jar exploded mode to not maintain original directory structure when adding dependencies to container.

mpeddada1:
Add dependency directly to container, without maintaining original directory structure. For example, if the dependency path is originally `nested/dependency.jar`, the path on the container will be `app/dependendencies/dependency.jar`. This is for the dependencies to be recognized when the `-cp` is specified with a wildcard, in the form: `java -cp /app/explodedJar:/app/dependencies/* <MainClass>`.


jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -64,11 +68,20 @@ public void testErrorLogging_directoryGiven() throws URISyntaxException {
...
   public void testJar_toDocker() throws IOException, InterruptedException, URISyntaxException {
-    Path jarFile = Paths.get(Resources.getResource("simpleJar.jar").toURI());
+    Path jarPath = Paths.get(Resources.getResource("jarTest/jarWithCp.jar").toURI());
     Integer exitCode =
         new CommandLine(new JibCli())
-            .execute("--target", "docker://jib-cli-image", "jar", jarFile.toString());
+            .execute("--target", "docker://jib-cli-image", "jar", jarPath.toString());
     String output = new Command("docker", "run", "--rm", "jib-cli-image").run();
+    String classPath = null;
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
```
chanseokoh, line 77: It's okay to put everything into the `try` block, especially in tests.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -64,11 +68,20 @@ public void testErrorLogging_directoryGiven() throws URISyntaxException {
...
+      classPath = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
+    }
+
+    // Validate classpath before checking output
+    assertThat(classPath).isNotNull();
+    List<String> dependencies = Splitter.onPattern("\\s+").splitToList(classPath.trim());
```
chanseokoh, line 83: No need for splitting. I think it can be `assertThat(classPath).isEqualTo("dependency1.jar;directory/dependency2.jar")`, even without checking not null.

And just double-check that the main in `simplJar` does use classes in both `dependency1.jar` and `dependency2.jar`.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
```diff
@@ -67,7 +67,8 @@ public void testToJibContainerBuilder_basicInfo()
         .isEqualTo(
             FileEntriesLayer.builder()
                 .addEntry(
-                    Paths.get("dependency1"), AbsoluteUnixPath.get("/app/dependencies/dependency1"))
+                    standardJar.getParent().resolve(Paths.get("dependency1")),
```
chanseokoh, line 70:
```suggestion
                    standardJar.getParent().resolve("dependency1"),
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -123,7 +127,10 @@ static JarType determineJarType(Path jarPath) throws IOException {
         snapshotDependencies.forEach(
             path ->
                 snapshotDependenciesLayerBuilder.addEntry(
-                    path, APP_ROOT.resolve(RelativeUnixPath.get("dependencies")).resolve(path)));
+                    jarParent.resolve(path),
+                    APP_ROOT
+                        .resolve(RelativeUnixPath.get("dependencies"))
```
chanseokoh, line 132:
```suggestion
                        .resolve("dependencies")
```
and the one above tool.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -108,13 +108,17 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
           allDependencies.stream().filter(isSnapshot).map(Paths::get).collect(Collectors.toList());
+      Path jarParent = jarPath.getParent() == null ? Paths.get("") : jarPath.getParent();
       if (!nonSnapshotDependencies.isEmpty()) {
         FileEntriesLayer.Builder nonSnapshotDependenciesLayerBuilder =
             FileEntriesLayer.builder().setName(DEPENDENCIES);
         nonSnapshotDependencies.forEach(
             path ->
                 nonSnapshotDependenciesLayerBuilder.addEntry(
-                    path, APP_ROOT.resolve(RelativeUnixPath.get("dependencies")).resolve(path)));
+                    jarParent.resolve(path),
```
chanseokoh, line 118: I just realized one thing we are missing. Remember [this output](https://github.com/GoogleContainerTools/jib/pull/2888#discussion_r520255662)? A lot of people will try Jib CLI on a standard JAR without having dependency JARs at all or in right relative locations, and they will just see `java.nio.file.NoSuchFileException: some.jar`. And they may falsely think Jib CLI is bugged by failing to load its dependency. I think it's very important to check the existence of dependency JARs locally before we delegate to jib-core, and throw an exception (`IllegalStateException`?) with a detailed message that some dependency JARs of the user app don't exist locally. Doesn't have to be in this PR, but we should not forget this.

chanseokoh (commented):
LGTM, only some minor comments.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: That is a great point! And I agree, checking this earlier + returning a meaningful error message are both super important.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
mpeddada1: Done.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -173,6 +174,21 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+      FileEntriesLayer.Builder layerbuilder,
+      Path depPath,
+      Path jarParent,
+      AbsoluteUnixPath pathOnContainer) {
+    File depFile = new File(jarParent.resolve(depPath).toString());
+    if (!depFile.exists()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Dependency: %s needs to be in the same parent directory as the JAR.",
+              depPath.toString()));
```
chanseokoh, line 187: I think this will cause confusion when a dependency is nested. It reads like such a dependency should be in the same parent directory. How about something like

"Dependency required by the JAR (as specified in 'Class-Path:' in the JAR manifest) doesn't exist: <full path of jarParent.resolve(depPath)>"

? And then we can also pass the full path instead of two argument (`depPath` and `jarParent`) to the method. WDYT?

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -64,13 +66,36 @@ public void testErrorLogging_directoryGiven() throws URISyntaxException {
...
-            .execute("--target", "docker://jib-cli-image", "jar", jarFile.toString());
+            .execute("--target", "docker://jib-cli-image", "jar", jarPath.toString());
     String output = new Command("docker", "run", "--rm", "jib-cli-image").run();
+    String classPath = null;
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      classPath = jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
 
-    assertThat(exitCode).isEqualTo(0);
-    assertThat(output).isEqualTo("Hello World");
+      assertThat(classPath).contains("dependency1.jar directory/dependency2.jar");
```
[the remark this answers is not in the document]
mpeddada1: Oops, my bad

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -173,6 +174,21 @@ static JarType determineJarType(Path jarPath) throws IOException {
     }
   }
 
+  private static void addDependency(
+      FileEntriesLayer.Builder layerbuilder,
+      Path depPath,
+      Path jarParent,
+      AbsoluteUnixPath pathOnContainer) {
+    File depFile = new File(jarParent.resolve(depPath).toString());
+    if (!depFile.exists()) {
```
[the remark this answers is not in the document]
mpeddada1: Done.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: Oh yeah, that can be really confusing. Passing in the full dependency path is a good idea.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -173,6 +174,21 @@ static JarType determineJarType(Path jarPath) throws IOException {
     }
   }
 
+  private static void addDependency(
+      FileEntriesLayer.Builder layerbuilder,
+      Path depPath,
+      Path jarParent,
+      AbsoluteUnixPath pathOnContainer) {
```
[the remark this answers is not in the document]
mpeddada1: Hm I was planning on reusing this helper in #2888, which requires a different path on the container. I could directly compute it in here and refactor this method in the next PR if it doesn't make sense in the current context though?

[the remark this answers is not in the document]
chanseokoh: Ah, right. That's OK then.

chanseokoh (approved):
LGTM. Just double-check if all of the following files are required before merging:

- test/resources/jar/standard/dependency1
- test/resources/jar/standard/dependency2
- test/resources/jar/standard/dependency3-SNAPSHOT-1.jar
- test/resources/jar/standard/directory/dependency4
- test/resources/jar/standard/singleDepJar.jar
- integration-test/resources/jarTest/dependency1.jar
- integration-test/resources/jarTest/directory/dependency2.jar
- integration-test/resources/jarTest/jarWithCp.jar
- integration-test/resources/noDependencyJar.jar
- integration-test/resources/simpleJar.jar

PR #2910, opened 2020-12-02
Migrate away from Docker Hub in integration tests

chanseokoh:
Due to the new Docker Hub rate limit, we are starting to see errors like

```
com.google.cloud.tools.jib.registry.BlobPullerIntegrationTest > classMethod FAILED
    java.lang.RuntimeException: Command 'docker run --rm -d -p 5000:5000 --name registry-a8087b15-fa3c-401c-bcbe-09f2a5e52dc9 registry:2' failed: Unable to find image 'registry:2' locally
    docker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
    See 'docker run --help'.
```

This PR uses `mirror.gcr.io` when feasible.

Note, normally, one shouldn't directly reference `mirror.gcr.io`, as the mirror is a pull-through cache and doesn't guarantee the existence of images. However, in practice I don't think `busybox` and `registry:2` will ever fail on `mirror.gcr.io`.

The PR also does a lot of cleanups and improves performance by switching to use a local registry.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/BlobPusherIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/BlobPusherIntegrationTest.java
@@ -31,14 +31,12 @@ import org.junit.Test;
 /** Integration tests for {@link BlobPusher}. */
 public class BlobPusherIntegrationTest {
 
-  @ClassRule public static LocalRegistry localRegistry = new LocalRegistry(5000);
+  @ClassRule public static final LocalRegistry localRegistry = new LocalRegistry(5000);
 
   private final FailoverHttpClient httpClient = new FailoverHttpClient(true, false, ignored -> {});
 
   @Test
-  public void testPush()
-      throws DigestException, IOException, RegistryException, InterruptedException {
-    localRegistry.pullAndPushToLocal("busybox", "busybox");
+  public void testPush() throws DigestException, IOException, RegistryException {
     Blob testBlob = Blobs.from("crepecake");
     // Known digest for 'crepecake'
     DescriptorDigest testBlobDigest =
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPullerIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPullerIntegrationTest.java
@@ -52,7 +52,6 @@ public class ManifestPullerIntegrationTest {
     RegistryClient registryClient =
         RegistryClient.factory(EventHandlers.NONE, "localhost:5000", "busybox", httpClient)
             .newRegistryClient();
-    registryClient.doPullBearerAuth();
 
     V21ManifestTemplate manifestTemplate =
         registryClient.pullManifest("latest", V21ManifestTemplate.class).getManifest();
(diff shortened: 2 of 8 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPullerIntegrationTest.java
```diff
@@ -74,38 +75,32 @@ public void testPull_v22() throws IOException, RegistryException {
...
 
-    // Make sure we can't cast a v22ManifestTemplate to v22ManifestListTemplate in ManifestPuller
-    try {
-      registryClient.pullManifest(KNOWN_MANIFEST_LIST_SHA, V22ManifestTemplate.class);
-      Assert.fail();
-    } catch (ClassCastException ex) {
-      // pass
-    }
+    // Generic call to ":latest" pulls a manifest list
+    ManifestTemplate manifestList2 = registryClient.pullManifest("latest").getManifest();
```
mpeddada1, line 88: Perhaps a more descriptive name- for example, `manifestListLatest`.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/BlobPusherIntegrationTest.java
```diff
@@ -31,14 +31,12 @@
 /** Integration tests for {@link BlobPusher}. */
 public class BlobPusherIntegrationTest {
 
-  @ClassRule public static LocalRegistry localRegistry = new LocalRegistry(5000);
+  @ClassRule public static final LocalRegistry localRegistry = new LocalRegistry(5000);
```
mpeddada1, line 34: Is this still needed?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/ManifestPusherIntegrationTest.java
```diff
@@ -30,19 +30,13 @@
...
 /** Integration tests for {@link ManifestPusher}. */
 public class ManifestPusherIntegrationTest {
 
-  @ClassRule public static LocalRegistry localRegistry = new LocalRegistry(5000);
-
-  @BeforeClass
-  public static void setUp() throws IOException, InterruptedException {
-    localRegistry.pullAndPushToLocal("busybox", "busybox");
-  }
+  @ClassRule public static final LocalRegistry localRegistry = new LocalRegistry(5000);
```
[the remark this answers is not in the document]
chanseokoh: Tests in this class push a blob to this registry. We could use our GCP integration test project, but I think in this case we should use an isolated registry that starts from a clean state.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/BlobPusherIntegrationTest.java
chanseokoh: The test in this class pushes a blob to this registry (line 47). We could use our GCP integration test project, but I think in this case we should use an isolated registry that starts from a clean state.

mpeddada1: oops I see.

mpeddada1 (commented):
LGTM

mpeddada1 (approved):

PR #2912, opened 2020-12-03
Improve integration tests

chanseokoh:
1. Each test took a lot of time because it needed to download the base image over and over because we set a temporary cache directory. The reason was to check consecutive build times in one test. No need to start from a clean state in every test. This will significantly boost the speed of this class.
1. Removed `testBuildToDockerDaemon` as it's a strict subset of `testBuildToDockerDaemon_multipleTags`. Normally we should have both if they were unit tests, but since these are integration tests, I think it's fine to have only the larger test for performance.
1. Using `openjdk:8-jre-slim` instead of `openjdk:8-jre-alpine` as a base image in `testBuildToDockerRegistry_dockerHubBaseImage`, as it causes the following error locally. (Maybe the error depends on a certain platform?)
```
com.google.cloud.tools.jib.api.ContainerizerIntegrationTest > testBuildToDockerRegistry_dockerHubBaseImage FAILED
    java.lang.RuntimeException: Command 'docker run --rm localhost:5000/testimage:testtag' failed: Error loading shared library libjli.so: No such file or directory (needed by /usr/bin/java)
    Error relocating /usr/bin/java: JLI_Launch: symbol not found
        at com.google.cloud.tools.jib.Command.run(Command.java:68)
        at com.google.cloud.tools.jib.Command.run(Command.java:45)
        at com.google.cloud.tools.jib.api.ContainerizerIntegrationTest.testBuildToDockerRegistry_dockerHubBaseImage(ContainerizerIntegrationTest.java:282)
```


chanseokoh:
Ah, wait. There seems to be a flaky issue with progress checking.

chanseokoh:
Now ready for review.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -149,35 +149,43 @@ private static void assertLayerSize(int expected, String imageReference)
...
+  private ProgressChecker progressChecker = new ProgressChecker();
 
   @Test
   public void testSteps_forBuildToDockerRegistry()
       throws IOException, InterruptedException, ExecutionException, RegistryException,
-          CacheDirectoryCreationException {
+          CacheDirectoryCreationException, InvalidImageReferenceException {
+    System.setProperty("jib.alwaysCacheBaseImage", "true");
+    String imageReference = "localhost:5000/testimage:testtag";
+    Path cacheDirectory = temporaryFolder.newFolder().toPath();
```
mpeddada1, line 160: Do we want to enable caching for the other tests as well?

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -283,54 +288,38 @@ public void testBuildToDockerRegistry_dockerHubBaseImage()
...
-        ImageReference.of(null, "testdocker", null),
-        Collections.emptyList());
+        Containerizer.to(DockerDaemonImage.named("testdocker")),
+        Arrays.asList("testtag2", "testtag3"));
 
     progressChecker.checkCompletion();
 
-    assertDockerInspect("testdocker");
     assertLayerSize(7, "testdocker");
+    assertDockerInspect("testdocker");
```
mpeddada1, line 302: I know these methods were not introduced in this PR but since we are using a helper for the assert statements, could we make the name a bit more detailed? Something along the lines of `assertThatExpectedLayerSizeReturned` or `assertThatExpectedDockerInspectOutputReturned`.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -271,9 +276,9 @@ public void testSteps_forBuildToDockerRegistry_skipExistingDigest()
   public void testBuildToDockerRegistry_dockerHubBaseImage()
       throws InvalidImageReferenceException, IOException, InterruptedException, ExecutionException,
           RegistryException, CacheDirectoryCreationException {
-    buildRegistryImage(
-        ImageReference.parse("openjdk:8-jre-alpine"),
-        ImageReference.of("localhost:5000", "testimage", "testtag"),
+    buildImage(
+        ImageReference.parse("openjdk:8-jre-slim"),
```
mpeddada1, line 280: Nice, I'm guessing this fixes the compatibility issue and also results in a smaller image:)

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
mpeddada1: Or make `line 157` and `line 160` part of `setUp()`

chanseokoh: Caching is always enabled. It's just that this particular test purposely gives a new empty directory to not reuse any existing cache. (The test times consecutive builds to ensure the second build runs faster due to caching from the first run.)

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
chanseokoh: I wish I could, but `assertDockerInspect` checks a lot of image metadata, including labels, ports, environment variables, and history.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -383,10 +345,7 @@ private JibContainer buildImage(
             .setLabels(ImmutableMap.of("key1", "value1", "key2", "value2"))
             .setFileEntriesLayers(fakeLayerConfigurations);
 
-    Path cacheDirectory = temporaryFolder.newFolder().toPath();
     containerizer
```
loosebazooka, line 348: so now we're just using the system default cache?

loosebazooka (approved):

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
chanseokoh: Yeah, the usual place for base images. I think that's fine. We've been not setting cache directories in other integration tests.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
mpeddada1: Oh gotcha, that makes a lot of sense. Thanks!

PR #2906, opened 2020-12-01
Convert jib.*.auth.username and jib.*.auth.password to properties

dubinsky:
Fixes https://github.com/GoogleContainerTools/jib/issues/2905

The following are now properties and not plain strings:
- from.auth.username
- from.auth.password
- from.image
- to.auth.username
- to.auth.password

PR #2915, opened 2020-12-03
Have consistency in instantiating Gradle parameter classes

chanseokoh:
While reviewing #2906, I was perplexed as it seemed working even though `objectFactory.newInstance()` does not pass an `objectFactory` as a constructor argument. Turns out [some classes are auto-injected](https://docs.gradle.org/current/javadoc/org/gradle/api/model/ObjectFactory.html#newInstance-java.lang.Class-java.lang.Object...-):

> In addition to those parameters provided as an argument to this method, the following services are also available for injection:
>
> * ObjectFactory.

To avoid potential confusion in the future, I'm making the usage consistent.


mpeddada1 (approved):
Pretty cool.

PR #2917, opened 2020-12-03
Minor cleanups on Jib CLI

chanseokoh:
I don't know if it's better to create `CommandSpec` directly or not. Trying the original way.


mpeddada1 (approved):
LGTM

Issue #2847, opened 2020-10-19
[jib-cli-jar] Implement 'exploded' mode for spring-boot fat jar. 


mpeddada1:

mpeddada1:
Closed with #2918 and #2926.

PR #2918, opened 2020-12-04
Implement explode mode for Spring Boot Fat Jar

mpeddada1:
For #2847 
Implements explode mode for spring boot fat jar to create layers as specified by [`BOOT-INF/layers.idx` ](https://spring.io/blog/2020/08/14/creating-efficient-docker-images-with-spring-boot-2-3)(if present), otherwise creates dependencies, spring-boot-loader, snapshot dependencies, resources and classes layers as determined by the [standard spring boot executable jar format.](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html)

Notes on testing:
- Integration tests implemented in #2926


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -131,6 +138,81 @@ static JarType determineJarType(Path jarPath) throws IOException {
     return layers;
   }
 
+  /**
+   * Creates layers as specified in BOOT-INF/layers.idx (if present) or for dependencies,
+   * spring-boot-loader, snapshot dependencies, resource and classes for a spring boot fat jar.
+   *
+   * @param jarPath path to jar file
+   * @param localExplodedJarRoot path to temporary jib local directory
```
chanseokoh, line 146: ditto

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -131,6 +138,81 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+      ZipUtil.unzip(jarPath, localExplodedJarRoot);
+      if (layerIndex != null) {
+        return createLayersForLayeredSpringBootJar(localExplodedJarRoot);
+      }
+
+      ArrayList<FileEntriesLayer> layers = new ArrayList<>();
+
+      // Non-snapshot layer
+      Predicate<Path> isInBootInfLib =
+          path -> path.getParent().startsWith(localExplodedJarRoot.resolve("BOOT-INF/lib"));
```
chanseokoh, line 165: Although `BOOT-INF/lib` works, for non-test code, we do `.resolve("BOOT-INF").resolve("lib")` to avoid using a certain path separator.

And why do we need `getParent()` if we are checking `startsWith()`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -131,6 +138,81 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+              "spring-boot-loader", localExplodedJarRoot, isLoader, APP_ROOT);
+
+      // Classes layer.
+      Predicate<Path> isClass = path -> path.getFileName().toString().endsWith(".class");
+      Predicate<Path> isInBootInfClasses =
+          path ->
+              path.getParent()
+                  .startsWith(
+                      localExplodedJarRoot.resolve(
+                          localExplodedJarRoot.resolve("BOOT-INF/classes")));
```
chanseokoh, line 192: This doesn't seem right. It's surprising this double-resolve actually worked.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -131,6 +138,81 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+      Predicate<Path> finalPredicateClasses = isInBootInfClasses.and(isClass);
+      FileEntriesLayer classesLayer =
+          addDirectoryContentsToLayer(
+              CLASSES, localExplodedJarRoot, finalPredicateClasses, APP_ROOT);
+
+      // Resources layer.
+      Predicate<Path> isInMetaInf =
+          path -> path.getParent().startsWith(localExplodedJarRoot.resolve("META-INF"));
+      Predicate<Path> finalPredicateResources =
+          isInBootInfClasses.or(isInMetaInf).and(isClass.negate());
```
chanseokoh, line 202: I think the following is more accurate:
```
isInMetaInf.or(isInBootInfClasses.and(isClass.negate()))
```
And I thought we should also include resources files at the JAR content root? Is it not the case?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -233,6 +325,66 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+   * Creates layers as specified by the layers.idx file (located in the BOOT-INF/ directory of the
+   * JAR).
+   *
+   * @param localExplodedJarRoot Path to temporary directory
+   * @return list of {@link FileEntriesLayer}
+   * @throws IOException if temporary directory provided doesn't exist
+   */
+  private static List<FileEntriesLayer> createLayersForLayeredSpringBootJar(
+      Path localExplodedJarRoot) throws IOException {
+    Path layerIndexPath = localExplodedJarRoot.resolve(Paths.get("BOOT-INF/layers.idx"));
```
chanseokoh, line 338: `localExplodedJarRoot.resolve("BOOT-INF")...`. No need for `Paths.get()`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -131,6 +138,81 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+      Path jarPath, Path localExplodedJarRoot) throws IOException {
+
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      ZipEntry layerIndex = jarFile.getEntry("BOOT-INF/layers.idx");
+      ZipUtil.unzip(jarPath, localExplodedJarRoot);
+      if (layerIndex != null) {
+        return createLayersForLayeredSpringBootJar(localExplodedJarRoot);
+      }
+
+      ArrayList<FileEntriesLayer> layers = new ArrayList<>();
```
chanseokoh, line 161: Let's move this down to line 207.

chanseokoh (commented):
I'll continue reviewing this tomorrow.

chanseokoh:
Along with my comments above, how about this? https://github.com/GoogleContainerTools/jib/compare/cli-spring-boot-expl..cli-spring-boot-expl-2?expand=1

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -417,4 +423,318 @@ public void testComputeEntrypointForPackagedStandard_withMainClass()
...
+                .collect(Collectors.toList()))
+        .containsExactly(
+            AbsoluteUnixPath.get("/app/BOOT-INF"),
+            AbsoluteUnixPath.get("/app/BOOT-INF/classes"),
+            AbsoluteUnixPath.get("/app/BOOT-INF/classes/class1.class"),
+            AbsoluteUnixPath.get("/app/BOOT-INF/classes/classDirectory/class2.class"),
+            AbsoluteUnixPath.get("/app/BOOT-INF/classes/classDirectory"),
+            AbsoluteUnixPath.get("/app/BOOT-INF/lib"),
+            AbsoluteUnixPath.get("/app/META-INF"),
+            AbsoluteUnixPath.get("/app/META-INF/MANIFEST.MF"),
```
chanseokoh, line 519: Interesting. I guess `layer.idx` in this test doesn't have a resource layer for `META-INF`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -417,4 +423,318 @@ public void testComputeEntrypointForPackagedStandard_withMainClass()
     assertThat(actualEntrypoint)
         .isEqualTo(ImmutableList.of("java", "-jar", "/app/standardJarWithClassPath.jar"));
   }
+
+  @Test
+  public void testCreateLayersForExplodedLayeredSpringBoot()
+      throws IOException, URISyntaxException {
```
chanseokoh, line 429: I think it's worth adding the contents of `layer.idx` as a code comment.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -417,4 +423,318 @@ public void testComputeEntrypointForPackagedStandard_withMainClass()
...
+            AbsoluteUnixPath.get("/app/org/springframework"),
+            AbsoluteUnixPath.get("/app/org/springframework/boot"),
+            AbsoluteUnixPath.get("/app/org/springframework/boot/loader"),
+            AbsoluteUnixPath.get("/app/org/springframework/boot/loader/data"));
+  }
+
+  @Test
+  public void testCreateLayersForExplodedLayeredSpringBoot_singleEmptyLayerListed()
+      throws IOException, URISyntaxException {
+    // Springboot JAR with "snapshot-dependencies" layer listed as empty in BOOT-INF/layers.idx
```
chanseokoh, line 530: Content of layers.idx as a comment?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
```diff
@@ -146,4 +151,222 @@ public void testToJibContainerBuilder_packagedStandard_basicInfo()
...
+    assertThat(buildPlan.getEnvironment()).isEmpty();
+    assertThat(buildPlan.getLabels()).isEmpty();
+    assertThat(buildPlan.getVolumes()).isEmpty();
+    assertThat(buildPlan.getExposedPorts()).isEmpty();
+    assertThat(buildPlan.getUser()).isNull();
+    assertThat(buildPlan.getWorkingDirectory()).isNull();
+    assertThat(buildPlan.getEntrypoint())
+        .isEqualTo(
+            ImmutableList.of("java", "-cp", "/app", "org.springframework.boot.loader.JarLauncher"));
+    assertThat(buildPlan.getLayers().size()).isEqualTo(4);
```
chanseokoh, line 177: Can we also check the name of each layer?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
```diff
@@ -146,4 +151,222 @@ public void testToJibContainerBuilder_packagedStandard_basicInfo()
...
+    assertThatExpectedEntriesPresentInSnapshotLayer_SpringBoot(
+        ((FileEntriesLayer) buildPlan.getLayers().get(2)).getEntries(), destDir);
+    assertThat(((FileEntriesLayer) buildPlan.getLayers().get(3)).getEntries())
+        .containsExactlyElementsIn(
+            FileEntriesLayer.builder()
+                .addEntry(destDir.resolve("META-INF/"), AbsoluteUnixPath.get("/app/META-INF/"))
+                .addEntry(
+                    destDir.resolve("META-INF/MANIFEST.MF"),
+                    AbsoluteUnixPath.get("/app/META-INF/MANIFEST.MF"))
+                .addEntry(destDir.resolve("BOOT-INF/"), AbsoluteUnixPath.get("/app/BOOT-INF/"))
```
chanseokoh, line 191: I also commented this below, but we should try to exclude directories that don't belong to this layer.

loosebazooka:
Nothing specific about this, but about the layout in general. JarModeProcesssor seems to be doing too much work. I think I would imagine it more like:

```
interface JarProcessor
 |_ StandardExplodedProcessor
 |_ StandardPackagedProcessor
 |_ SpringBootPackagedProcessor
 |_ ...
```

Where each of those classes implements the JarProcessor interface? We can talk about this in the design meeting?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: Sorry, would that mean resources files outside the `BOOT-INF/classes/` and `META-INF/` directories?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -262,4 +414,18 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
             });
     return builder.build();
   }
+
+  private static Optional<Predicate<Path>> computePredicateForLayeredSpringBoot(
+      List<String> layerContents, Path localExplodedJarRoot) {
+    List<Predicate<Path>> allPredicates = new ArrayList<>();
+    for (String pathName : layerContents) {
+      if (pathName.endsWith("/")) {
```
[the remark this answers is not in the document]
mpeddada1: That is a good point and something I considered while writing this condition. But documentation on the [format for layers.idx](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html#executable-jar-war-index-files-layers) says that "a directory name ends with /, a file name does not" which is why decided to leave it in. (Sorry, the comment was pretty misleading)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
chanseokoh: No, it means `isInMetaInf || (isInBootInfClasses && !isClass)`. OTOH, your original condition is `(isInBootInfClasses || isInMetaInf) && !isClass)`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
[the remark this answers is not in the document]
mpeddada1: I can modify it if we don't want to rely too much on the documentation.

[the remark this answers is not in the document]
chanseokoh: Ah, it's documented. I think we can rely on it. I thought it's not true based on your sample in the code comment:
```
      // - application:
      //   - BOOT-INF/classes
``` 
`BOOT-INF/classes` doesn't end with `/` and I thought this was an actual `layer.idx` content you extracted from a sample JAR. Fix the misleading code comment.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
chanseokoh: That said, I'm curious if a spring jar can have resources files at root? Perhaps not?

chanseokoh: (OTOH, a regular jar has resource files for the application at root.)

chanseokoh: I wonder if it can't be like this for a spring jar:

1. resources under `BOOT-INF/classes/` --> resources for the app
1. resources under root (excluding `BOOT-INF/`) --> resources for the spring launcher? (Not accessible from the app runtime but they are for configuring Spring Boot?)
1. resources under `META-INF/` --> for the app or the launcher? Or both?

mpeddada1: hm afaik,[ spring boot jars are structured in this way:](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html#executable-jar-nested-jars) 
``` 
+-META-INF
 |  +-MANIFEST.MF
 +-org
 |  +-springframework
 |     +-boot
 |        +-loader
 |           +-<spring boot loader classes>
 +-BOOT-INF
    +-classes
    |  +-mycompany
    |     +-project
    |        +-YourClasses.class
    +-lib
       +-dependency1.jar
       +-dependency2.jar
```  
The application classes and resources go into `BOOT-INF/classes/`. `META-INF/` can also contain some files such as `pom.properties`, `pom.xml` and `build-info.properties`.  But outside of those three directories (META/, BOOT-INF/, org/) I don't see the possibility of any other resources existing - at least going by the standard structure of the spring boot JAR.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -233,6 +325,66 @@ static JarType determineJarType(Path jarPath) throws IOException {
...
+   *
+   * @param localExplodedJarRoot Path to temporary directory
+   * @return list of {@link FileEntriesLayer}
+   * @throws IOException if temporary directory provided doesn't exist
+   */
+  private static List<FileEntriesLayer> createLayersForLayeredSpringBootJar(
+      Path localExplodedJarRoot) throws IOException {
+    Path layerIndexPath = localExplodedJarRoot.resolve(Paths.get("BOOT-INF/layers.idx"));
+    Pattern layerNamePattern = Pattern.compile("-\\s(.*):");
+    Pattern fileNamePattern = Pattern.compile("\\s\\s-\\s(.*)");
```
chanseokoh, line 340: The [Spring doc](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html#executable-jar-war-index-files-layers) makes it very explicit that they are either dash-space or space-space-dash-space. I think we can use ` ` (space) instead of `\\s`. Also the doc says paths and layer names are quoted. So we can have it like `Pattern.compile("- \"([^\"]*)\":")` (untested) without later removing quotes?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -42,7 +50,7 @@
   private static final String CLASSES = "classes";
   private static final String RESOURCES = "resources";
   private static final String DEPENDENCIES = "dependencies";
-  private static final String SNAPSHOT_DEPENDENCIES = "snapshot dependencies";
+  private static final String SNAPSHOT_DEPENDENCIES = "snapshot-dependencies";
```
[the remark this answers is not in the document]
mpeddada1: I initially changed it to be consistent with the format in layers.idx. But you're right --I reverted the change.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -79,20 +87,19 @@ static JarType determineJarType(Path jarPath) throws IOException {
    * a standard jar.
    *
    * @param jarPath path to jar file
-   * @param tempDirPath path to temporary jib local directory
+   * @param localExplodedJarRoot path to temporary jib local directory
```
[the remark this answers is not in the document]
mpeddada1: Right, reverted this change.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: Oops, my bad.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
mpeddada1: The layer.idx has an "application" layer which contains both classes and resources.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
mpeddada1: Good idea.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
mpeddada1: Good point. Applied this change.

mpeddada1:
> Along with my comments above, how about this? https://github.com/GoogleContainerTools/jib/compare/cli-spring-boot-expl..cli-spring-boot-expl-2?expand=1

Thank you for the suggested edits @chanseokoh! I've applied them to this PR.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
mpeddada1: Done.

mpeddada1:
This PR is ready for review.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -251,15 +392,36 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
       AbsoluteUnixPath basePathInContainer)
       throws IOException {
     FileEntriesLayer.Builder builder = FileEntriesLayer.builder().setName(layerName);
+    HashSet<Path> addedPaths = new HashSet<>();
     new DirectoryWalker(sourceRoot)
         .filterRoot()
-        .filter(path -> Files.isDirectory(path) || pathFilter.test(path))
+        .filter(path -> pathFilter.test(path))
```
chanseokoh, line 398: I vaguely remember I added `Files.isDirectory(path)` to not miss adding empty directories. (It's not because we wanted to add all the directories leading to `/`. Maybe missing empty directories doesn't matter in practice almost all the time, especially for `jib jar`. I wonder if we can just get rid of `Files.isDirectory(path)` and the new `while` loop below? As long as it works, we should be good?

mpeddada1: Good idea - Tried adding the path directly without adding any empty directories and it seemed to work! Will go with this for now.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -95,16 +105,21 @@ static JarType determineJarType(Path jarPath) throws IOException {
     Path localExplodedJarRoot = tempDirPath;
     ZipUtil.unzip(jarPath, localExplodedJarRoot);
     Predicate<Path> isClassFile = path -> path.getFileName().toString().endsWith(".class");
-    Predicate<Path> isResourceFile = isClassFile.negate();
+    Predicate<Path> isFile = path -> path.toFile().isFile();
+    Predicate<Path> isResourceFile = isFile.and(isClassFile.negate());
```
chanseokoh, line 109: Nit: this should also work:
```java
    Predicate<Path> isFile = Files::isRegularFile;
```
or
```java
    Predicate<Path> isResourceFile = isClassFile.negate().and(Files::isRegularFile);
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -253,13 +396,24 @@ private static FileEntriesLayer addDirectoryContentsToLayer(
...
-                  basePathInContainer.resolve(sourceRoot.relativize(path));
-              builder.addEntry(path, pathOnContainer);
+              builder.addEntry(path, basePathInContainer.resolve(sourceRoot.relativize(path)));
             });
     return builder.build();
   }
+
+  private static Predicate<Path> isInListedDirectoryOrIsSameFile(
+      List<String> layerContents, Path localExplodedJarRoot) {
+    Predicate<Path> predicate = Predicates.alwaysFalse();
```
chanseokoh, line 409: Start this with checking a file and remove the AND operation at the caller site?
```java
Predicate<Path> predicate = Files::isRegularFile;
```

mpeddada1: hm this method needs to check if the path is in **one** of the directories listed or if the path is the same as **one** of the files listed, which requires us to use the `or` condition. Since checking for a file requires `and` (we always want to check or this), starting with this predicate is a little challenging. I added it at the end of the method, once all the predicates were checked, but please let me know if I'm missing something here.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
mpeddada1: thank you!

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
chanseokoh: Ah, you're right. Good catch.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -129,14 +113,8 @@ public void testCreateLayersForExplodedStandard_withClassPathInManifest()
...
     assertThat(snapshotLayer.getName()).isEqualTo("snapshot dependencies");
-    assertThat(
-            snapshotLayer
-                .getEntries()
-                .stream()
-                .map(FileEntry::getExtractionPath)
-                .collect(Collectors.toList()))
-        .isEqualTo(
-            ImmutableList.of(AbsoluteUnixPath.get("/app/dependencies/dependency3-SNAPSHOT-1.jar")));
+    assertThat(snapshotLayer.getEntries().get(0).getExtractionPath())
```
chanseokoh, line 116: Likewise, need to assert the size of `getEntries()`.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessorTest.java
```diff
@@ -252,16 +210,8 @@ public void testCreateLayersForExplodedStandard_withoutClassPathInManifest_conta
...
-        resourcesLayer
-            .getEntries()
-            .stream()
-            .map(FileEntry::getExtractionPath)
-            .collect(Collectors.toList());
-    assertThat(actualResourcesPath)
-        .containsExactly(
-            AbsoluteUnixPath.get("/app/explodedJar/META-INF"),
-            AbsoluteUnixPath.get("/app/explodedJar/META-INF/MANIFEST.MF"));
+    assertThat(resourcesLayer.getEntries().get(0).getExtractionPath())
```
chanseokoh, line 213: ditto

chanseokoh (commented):
LGTM. Only minor comments.

chanseokoh (approved):

PR #2926, opened 2020-12-07
Add integration tests for explode mode of spring boot jar

mpeddada1:
Requires #2918 to be reviewed first.


jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -142,6 +142,39 @@ public void testNoDependencyJar_packagedMode_toDocker()
     }
   }
 
+  @Test
+  public void testSpringbootLayeredJar_explodedMode_toDocker()
+      throws IOException, InterruptedException, URISyntaxException {
+    Path jarPath =
+        Paths.get(Resources.getResource("jarTest/springboot/springboot_layered.jar").toURI());
```
chanseokoh, line 149: How are these spring boot jars constructed? They don't seem like an actual jar built from a Spring Boot project. (Usually it runs a server listening on port 8080).

chanseokoh (commented):
@loosebazooka what do you think of adding Spring Boot jars (~15MB) to the repo? We've been creating a fake, but for this kind of integration tests, I cannot think of a good alternative.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
mpeddada1: Great question! To create the jars, I downloaded the demo project from https://start.spring.io/ and ran `./mvnw package`. Although, I just realized that I'm currently asserting on System.out instead of actually testing if the desired output shows up on localhost:8080 after having run `SpringApplication.run()` -- which is probably not a sign of a good test for a spring boot application. I'll correct this.

chanseokoh: Hmm... yeah. It looked weird that a Spring Boot app prints `Hello World` and quits. Do the tests actually pass?

We have similar Spring Boot integration tests, For example, take a look at

https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SpringBootProjectIntegrationTest.java

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -154,4 +167,71 @@ public void testJar_unknownMode() {
         .contains(
             "Invalid value for option '--mode': expected one of [exploded, packaged] (case-sensitive) but was 'unknown'");
   }
+
+  @Test
+  public void testSpringBootLayeredJar_explodedMode() throws IOException, InterruptedException {
+    springBootProjectLayered.build("clean", "bootJar");
+    Path jarParentPath =
+        springBootProjectLayered.getProjectRoot().toAbsolutePath().resolve("build").resolve("libs");
```
chanseokoh, line 175: Hmm... I'm curious. Does it not work without `toAbsolutePath()`?

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -154,4 +167,71 @@ public void testJar_unknownMode() {
...
+    String output =
+        new Command("docker", "run", "--rm", "--detach", "-p8080:8080", "spring-boot-jar-layered")
+            .run();
+
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      assertThat(jarFile.getEntry("BOOT-INF/layers.idx")).isNotNull();
+      assertThat(getContent(new URL("http://localhost:8080"))).isEqualTo("Hello world");
+      assertThat(exitCode).isEqualTo(0);
+
+      new Command("docker", "stop", output.trim()).run();
```
chanseokoh, line 190: This will not run if one of the assertions above fail or some code throws an exception. Put it in `finally`, or `@After public tearDown()` as done in other tests.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -154,4 +167,71 @@ public void testJar_unknownMode() {
...
+        new CommandLine(new JibCli())
+            .execute("jar", "--target", "docker://spring-boot-jar-layered", jarPath.toString());
+    String output =
+        new Command("docker", "run", "--rm", "--detach", "-p8080:8080", "spring-boot-jar-layered")
+            .run();
+
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      assertThat(jarFile.getEntry("BOOT-INF/layers.idx")).isNotNull();
+      assertThat(getContent(new URL("http://localhost:8080"))).isEqualTo("Hello world");
+      assertThat(exitCode).isEqualTo(0);
```
chanseokoh, line 188: Nit: I'd put this before `docker run` so that you don't run `docker run` if `jib jar` fails for some reason.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
```diff
@@ -154,4 +167,71 @@ public void testJar_unknownMode() {
...
+      assertThat(jarFile.getEntry("BOOT-INF/layers.idx")).isNull();
+      assertThat(getContent(new URL("http://localhost:8080"))).isEqualTo("Hello world");
+      assertThat(exitCode).isEqualTo(0);
+
+      new Command("docker", "stop", output.trim()).run();
+    }
+  }
+
+  @Nullable
+  static String getContent(URL url) throws InterruptedException {
```
chanseokoh, line 221: `private`

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/TestProject.java
```diff
@@ -0,0 +1,92 @@
...
+              if (Files.isDirectory(path)) {
+                Files.createDirectory(destPath);
+              } else {
+                Files.copy(path, destPath);
+              }
+            });
+  }
+
+  private final String testProjectName;
+  private String gradleVersion = GradleVersion.version("5.6.4").getVersion();
```
chanseokoh, line 56: `final`

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/TestProject.java
```diff
@@ -0,0 +1,92 @@
...
+import java.net.URISyntaxException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import org.gradle.testkit.runner.BuildResult;
+import org.gradle.testkit.runner.GradleRunner;
+import org.gradle.util.GradleVersion;
+import org.junit.rules.TemporaryFolder;
+
+public class TestProject extends TemporaryFolder implements Closeable {
```
chanseokoh, line 32: I wish we had a way to share test utility classes across all projects. But let's not try to resolve this in this PR.

@loosebazooka

chanseokoh (commented):
Oh, we already had a util class that makes it easy to build from source. Nice!

LGTM in generally. Only minor comments.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/cli2/JarCommandTest.java
mpeddada1: Ah good catch. `toAbsolutePath()` doesn't make a difference

chanseokoh (approved):

PR #2928, opened 2020-12-08
merge overlapping tests

loosebazooka:
We had some overlap in tests for both `jar` and `build` paths, merged those into one. They can be common at the commonCliOptions level


mpeddada1 (commented):
LGTM

mpeddada1 (approved):

PR #2929, opened 2020-12-08
Add new Utility to determine cache location

loosebazooka:
There are two cache locations in the cli:
1. The base image cache (`--base-image-cache), by default shared with any other tools using the default values from jib core. this is intended, base images can be large, and sharing a single cache would be good. CLI options `--base-image-cache`
2. The project cache (`--project-cache`, previously `--application-cache`), is the cache directory for any application layers but can also be used for any other build artifacts that need to be stored as part of the build. Caching uses paths to project files, so we want to this to be consistent across builds for temporary files projects (like expanded jars)
   - the default for this is `<user-cache>/jib/cli/project/<hash>`, the `hash` is calculated by obtaining and normalizing the absolute path to a project.
   - the user maybe configure this to their own liking, we chose not to put in in the project root because of unknown interactions with user configuration for creating layers.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Caches.java
```diff
@@ -0,0 +1,81 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Charsets;
+import java.nio.file.Path;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Optional;
+import javax.annotation.Nullable;
+
+/** A class to determine cache locations for any cli commands. */
+public class Caches {
```
mpeddada1, line 29: Perhaps, `CacheLocation`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CachesTest.java
```diff
@@ -0,0 +1,81 @@
...
+  @Rule public TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Test
+  public void testCaches_defaults() {
+    CommonCliOptions commonCliOptions = new CommonCliOptions();
+    Path buildContext = Paths.get("some-context");
+    Caches caches = Caches.from(commonCliOptions, buildContext);
+
+    Truth8.assertThat(caches.getBaseImageCache()).isEmpty();
+    Truth8.assertThat(caches.getProjectCache())
```
mpeddada1, line 44: very nit: don't need the Truth8 at the beginning. Same below.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CachesTest.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Test
+  public void testHashPath_sameFileDifferentPaths() throws IOException {
+    temporaryFolder.newFolder("ignored");
+    Path path1 = temporaryFolder.getRoot().toPath().resolve("ignored").resolve("..");
+    Path path2 = temporaryFolder.getRoot().toPath();
+
+    Assert.assertNotEquals(path1, path2); // the general equality should not hold true
+    Assert.assertTrue(Files.isSameFile(path1, path2)); // path equality holds
```
mpeddada1, line 72: `assertThat(path1).isNotEqualTo(path2)` 

`assertThat(Files.isSameFine(path1, path2)).isTrue()`

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Caches.java
```diff
@@ -0,0 +1,81 @@
...
+        commonCliOptions.getBaseImageCache().orElse(null),
+        commonCliOptions.getProjectCache().orElse(defaultProjectCache));
+  }
+
+  @VisibleForTesting
+  static String hashPath(Path path) {
+    try {
+      byte[] hashedBytes =
+          MessageDigest.getInstance("SHA-256")
+              .digest(path.toAbsolutePath().normalize().toString().getBytes(Charsets.UTF_8));
```
mpeddada1, line 57: For my understanding, why is `.normalize()` needed here?

loosebazooka: it's to remove any `.` and `..` from the path

loosebazooka: since the hashing occurs on the string representation of the path, we want `/home/./potato/../tomato == `/home/tomato`, since they are in essence the same path.

chanseokoh: AFAIK, I remember `toAboslutePath()` simply adds `/` at the front. Can we somehow guarantee that `path` is a real absolute path? Maybe we should use `File.getCanonicalPath()` (e.g., as in [`ZipUtil`](https://github.com/GoogleContainerTools/jib/blob/master/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java#L50))? `Path` is purely conceptual, while `File.getCanonicalPath()` does actual computation against a filesystem.

loosebazooka: Yeah initially I didn't want the file system to process anything, but you're probably right. Changed to canonical, but we can revisit later, if problem occur?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectoriesTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    assertThat(cacheDirectories.getBaseImageCache()).hasValue(Paths.get("test-base-image-cache"));
+    assertThat(cacheDirectories.getProjectCache()).isEqualTo(Paths.get("test-project-cache"));
+  }
+
+  @Test
+  public void testHashPath_sameFileDifferentPaths() throws IOException {
+    temporaryFolder.newFolder("ignored");
+    Path path1 = temporaryFolder.getRoot().toPath().resolve("ignored").resolve("..");
+    Path path2 = temporaryFolder.getRoot().toPath();
```
mpeddada1, line 73: Sorry missed this last time, we could make the naming a little more specific for these two variables?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
```diff
@@ -0,0 +1,90 @@
...
+  @Nullable private final Path baseImageCache;
+  private final Path projectCache;
+
+  /**
+   * Create a caches helper for cli cache locations.
+   *
+   * @param commonCliOptions cli options
+   * @param contextRoot the context root, if a single file, use the parent directory, this context
+   *     root must exist
+   * @return an instance of Caches with cli specific cache locations
```
mpeddada1, line 41:
```suggestion
   * @return an instance of CacheDirectories with cli specific cache locations
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
```diff
@@ -0,0 +1,90 @@
...
+
+  @Nullable private final Path baseImageCache;
+  private final Path projectCache;
+
+  /**
+   * Create a caches helper for cli cache locations.
+   *
+   * @param commonCliOptions cli options
+   * @param contextRoot the context root, if a single file, use the parent directory, this context
+   *     root must exist
```
mpeddada1, line 40: nit: line 40 needs to be moved up

mpeddada1 (approved):

Issue #2931, opened 2020-12-08
jib-core: incompatible with jackson 2.12.0 (Java 8 date/time type `java.time.Instant` not supported by default)


iamdanfox:
**Environment**:

- *Jib version:* jib-core 0.16.0
- *Build tool:* Gradle 6.7.1
- *OS:* Mac, Linux

**Description of the issue**:

I'm writing a gradle plugin leveraging jib-core 0.16.0, and while it works fine on most projects, I've found that if gradle's plugin classpath happens to contain a newer version of jackon (specifically 2.12.0), then `Cache#retrieve` throws:

```
org.gradle.internal.UncheckedException: java.util.concurrent.ExecutionException: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type `java.time.Instant` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: java.util.ArrayList[0]->com.google.cloud.tools.jib.cache.LayerEntriesSelector$LayerEntryTemplate["sourceModificationTime"])
	at org.gradle.internal.UncheckedException.throwAsUncheckedException(UncheckedException.java:67)
	at org.gradle.internal.UncheckedException.throwAsUncheckedException(UncheckedException.java:41)
	at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:107)
	at org.gradle.api.internal.project.taskfactory.StandardTaskAction.doExecute(StandardTaskAction.java:58)
	at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:51)
	at org.gradle.api.internal.project.taskfactory.StandardTaskAction.execute(StandardTaskAction.java:29)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.run(ExecuteActionsTaskExecuter.java:555)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
	...
	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:56)
	at java.base/java.lang.Thread.run(Thread.java:832)
[...]

PR #2908, opened 2020-12-02
Bump jackson dependencies from 2.11.3 to 2.12.0

laurentgo:
[Jackson Release notes](https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.12)

Adds jsr310 datatype module which is now required for serializing `java.util.time.Instant`.

Bumps [jackson-databind](https://github.com/FasterXML/jackson-databind) from 2.11.3 to 2.12.0.
- [Release notes](https://github.com/FasterXML/jackson-databind/releases)
- [Commits](https://github.com/FasterXML/jackson-databind/compare/jackson-databind-2.11.3...jackson-databind-2.12.0)

Bumps [jackson-dataformat-yaml](https://github.com/FasterXML/jackson-dataformats-text) from 2.11.3 to 2.12.0.
- [Release notes](https://github.com/FasterXML/jackson-dataformats-text/releases)
- [Commits](https://github.com/FasterXML/jackson-dataformats-text/compare/jackson-dataformats-text-2.11.3...jackson-dataformats-text-2.12.0)

---

Also fixes #2931.

PR #2932, opened 2020-12-08
Jackson 2.12.0 compatibility (register jsr310 JavaTime module explicitly)

iamdanfox:
**Before this PR**

I've filed this issue: https://github.com/GoogleContainerTools/jib/issues/2931, summary below:

When I tried to use `jib-core` with jackson 2.12.0 on the classpath, `com.google.cloud.tools.jib.cache.Cache` stops working because `LayerEntryTemplate` has an `Instant` field and jackson will no longer serialize/deserialize this type [out of the box](https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.12.0/src/main/java/com/fasterxml/jackson/databind/util/BeanUtil.java#L301-L304), and we get this message:

> Java 8 date/time type `java.time.Instant` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling

**After this PR**

Users of jib-core should find functionality works fine with jackson 2.12 on the classpath.

**Possible downsides**

Imposing the jackson 2.11.x -> 2.12 upgrade on downstream consumers might be a bit disruptive, because it might surface subtle behaviour changes in their own jackson usage.

**Alternative options**

- we could add the jsr310 jar (to become _compatible_ with jackson 2.12) but still only compile against jackson 2.11.3
- we could change `LayerEntryTemplate` to use a epoch milli long instead of an Instant

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,8 +66,9 @@ subprojects {
 
     COMMONS_COMPRESS: 'org.apache.commons:commons-compress:1.20',
     COMMONS_TEXT: 'org.apache.commons:commons-text:1.9',
-    JACKSON_DATABIND: 'com.fasterxml.jackson.core:jackson-databind:2.11.3',
-    JACKSON_DATAFORMAT_YAML: 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.11.3',
+    JACKSON_DATABIND: 'com.fasterxml.jackson.core:jackson-databind:2.12.0',
+    JACKSON_DATAFORMAT_YAML: 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:2.12.0',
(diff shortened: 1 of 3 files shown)


chanseokoh:
Thanks for digging in and putting up a PR. Another user also filed a PR (#2908) and suggested the same fix. However, I noticed an [incompatible behavior](https://github.com/GoogleContainerTools/jib/pull/2908#issuecomment-739016113) after upgrading, so I was wondering if there's a way to keep it compatible. Any thoughts?

iamdanfox:
Ah ok let's close out and continue the discussion on https://github.com/GoogleContainerTools/jib/pull/2908 - I'm pretty sure we could write something to deserialize those old `{"epochSecond":1538581712,"nano":416152000}` values, although it might be a bit gross.

PR #2937, opened 2020-12-11
Add GitHub Actions for Jib CLI release

chanseokoh:
We need to upload an archive containing all the JAR files, but I think this is a good starting point.

### DIFF
--- a/.github/workflows/jib-cli-release.yml
+++ b/.github/workflows/jib-cli-release.yml
@@ -15,22 +15,7 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
 
-      #- name: Create commits
-      #  run: |
-      #    git checkout -b my-test
-      #    git push origin my-test
-      #    git config user.email "chanseok@google.com"
-      #    git config user.name "Chanseok Oh"
-      #    date +%s > report.txt
-      #    git add -A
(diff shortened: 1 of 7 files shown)


loosebazooka:
how does a release get triggered?

chanseokoh:
It's manual. On the GitHub Actions page, there's a "Run workflow" button when selecting the "Release Jib CLI release" workflow.

image

loosebazooka:
oh cool

.github/workflows/jib-cli-release.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: Build Jib CLI
+        run: |
+          # TODO: run integration test? (Requries auth with GCP.)
+          ./gradlew clean build jib-cli:instDist
+
+      - name: Run Gradle release
+        run: |
+          git checkout -b cli_release_v${{ github.event.inputs.release_version }}
+          git config user.email "chanseok@google.com"
```
loosebazooka, line 26: I wonder if this can be configured to be the user triggering the event?

loosebazooka (approved):

.github/workflows/jib-cli-release.yml
chanseokoh: Yeah, I need to figure this out eventually.

PR #2938, opened 2020-12-11
Use cacheDirectories in cli commands

loosebazooka:
Integrates the new CacheDirectories util into the cli commands

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
@@ -19,7 +19,9 @@ package com.google.cloud.tools.jib.cli.cli2;
 import com.google.cloud.tools.jib.filesystem.XdgDirectories;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
+import com.google.common.base.Preconditions;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -35,12 +37,15 @@ public class CacheDirectories {
   /**
    * Create a caches helper for cli cache locations.
    *
-   * @param commonCliOptions cli options
-   * @param contextRoot the context root, if a single file, use the parent directory, this context
+   * @param commonCliOptions cli options for user configured cache directories
+   * @param contextRoot the context root, use the parent directory of single files, this context
    *     root must exist
    * @return an instance of CacheDirectories with cli specific cache locations
    */
   public static CacheDirectories from(CommonCliOptions commonCliOptions, Path contextRoot) {
+    Preconditions.checkArgument(
+        Files.isDirectory(contextRoot),
+        "contextRoot must be a directory, but " + contextRoot.toString() + " is not.");
     return new CacheDirectories(
         commonCliOptions.getBaseImageCache().orElse(null),
         commonCliOptions
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
@@ -16,13 +16,13 @@
 
 package com.google.cloud.tools.jib.cli.cli2;
 
-import com.google.cloud.tools.jib.filesystem.XdgDirectories;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Charsets;
 import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Optional;
@@ -30,6 +30,8 @@ import javax.annotation.Nullable;
 
 /** A class to determine cache locations for any cli commands. */
 public class CacheDirectories {
+
+  private static final String APPLICATION_LAYER_CACHE_DIR = "application-layers";
 
   @Nullable private final Path baseImageCache;
   private final Path projectCache;
@@ -51,8 +53,8 @@ public class CacheDirectories {
         commonCliOptions
             .getProjectCache()
             .orElse(
-                XdgDirectories.getCacheHome()
-                    .resolve("cli")
+                Paths.get(System.getProperty("java.io.tmpdir"))
+                    .resolve("jib-cli-cache")
                     .resolve("projects")
                     .resolve(getProjectCacheDirectoryFromProject(contextRoot))));
   }
@@ -91,5 +93,9 @@ public class CacheDirectories {
 
   public Path getProjectCache() {
     return projectCache;
+  }
+
+  public Path getApplicationLayersCache() {
+    return projectCache.resolve(APPLICATION_LAYER_CACHE_DIR);
   }
 }
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
@@ -85,7 +85,7 @@ public class Jar implements Callable<Integer> {
       }
 
       CacheDirectories cacheDirectories =
-          CacheDirectories.from(commonCliOptions, jarFile.getParent());
+          CacheDirectories.from(commonCliOptions, jarFile.toAbsolutePath().getParent());
       Containerizer containerizer = Containerizers.from(commonCliOptions, logger, cacheDirectories);
 
       JibContainerBuilder containerBuilder =
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
@@ -84,7 +84,9 @@ public class Jar implements Callable<Integer> {
         return 1;
       }
 
-      Containerizer containerizer = Containerizers.from(commonCliOptions, logger);
+      CacheDirectories cacheDirectories =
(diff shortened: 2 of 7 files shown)


jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectoriesTest.java
```diff
@@ -46,26 +47,38 @@ public void testCacheDirectories_defaults() {
...
+    Path ignoredContext = temporaryFolder.newFolder("ignored").toPath();
+    CacheDirectories cacheDirectories = CacheDirectories.from(commonCliOptions, ignoredContext);
 
     assertThat(cacheDirectories.getBaseImageCache()).hasValue(Paths.get("test-base-image-cache"));
     assertThat(cacheDirectories.getProjectCache()).isEqualTo(Paths.get("test-project-cache"));
   }
 
+  @Test
+  public void testCacheDirectories_failIfContextIsNotDirectory() throws IOException {
+    Path badContext = temporaryFolder.newFile("a.file").toPath();
```
chanseokoh, line 70: Not a big deal, but `temporaryFolder.newFile().toPath()` will work too.

And use the new exception assertion style in jib-cli (using a lambda).

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -84,7 +84,9 @@ public Integer call() {
         return 1;
       }
 
-      Containerizer containerizer = Containerizers.from(commonCliOptions, logger);
+      CacheDirectories cacheDirectories =
+          CacheDirectories.from(commonCliOptions, jarFile.getParent());
```
chanseokoh, line 88: Technically I think the `jarFile` path (not the parent) can still work. The difference is whether JARs in the same directory will use the same hash directory and share the app cache. Is there a specific reason to use a parent directory?

One thing matters is when that `jarFile` doesn't have a parent directory, which will be a common case, e.g., when doing `jib jar app.jar`. We learned that `jarFile.getParent() == null` in this case. So I think the argument check in `from()` will fail. 

And I just got this question if we can't just use a fixed app cache directory (without hashed sub-directories) shared by every Jib CLI build. But maybe it's good to have a hashed sub-directory per app to avoid generating too many directories in the cache root?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
```diff
@@ -35,12 +37,15 @@
...
-   * @param contextRoot the context root, if a single file, use the parent directory, this context
+   * @param commonCliOptions cli options for user configured cache directories
+   * @param contextRoot the context root, use the parent directory of single files, this context
    *     root must exist
    * @return an instance of CacheDirectories with cli specific cache locations
    */
   public static CacheDirectories from(CommonCliOptions commonCliOptions, Path contextRoot) {
+    Preconditions.checkArgument(
+        Files.isDirectory(contextRoot),
+        "contextRoot must be a directory, but " + contextRoot.toString() + " is not.");
```
chanseokoh, line 48: Just to confirm again: even if `Path` is a relative path (for example, `.`), `path.toFile().getCanonicalPath()` is always a full absolute path based on the current working directory?

chanseokoh (commented):
So we took care of reusing an app cache.

But I don't remember how we will take care of the issue that we unzip jars into `/tmp` every time, which results in a different absolute path for each file in a jar.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectories.java
loosebazooka: canonical path will return a the same path for the same file -- no matter what the currently relative pathing is and whatever the working directory is

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
loosebazooka: > We learned that jarFile.getParent() == null in this case. 

Interesting, this is probably just because the parent path element doesn't exist, but shouldn't preclude us from getting a full path and then determining the parent?

chanseokoh: If the original path value (e.g., `app.jar`) can be resolved with `.getCanonicalPath()` (e.g., the file exists in the current working directory), then yeah, I think `.getParent()` on the resolved path should work. But calling `.getCanonicalPath()` after `.getParent()` will probably fail. In any case, this code will fail as it will not pass the `Preconditions.checkArgument()`.

loosebazooka: Right, perhaps the jar is fine then. I was initially worried that the jar would be different each time (like includes version number or something), and then we'll fail to reuse the cache.

The other option is to use a common application cache and another separate directory for "other" project files?

chanseokoh: > the jar would be different each time (like includes version number or something), and then we'll fail to reuse the cache.

Oh, that's a very good point.

dstiers11 (approved):

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectoriesTest.java
```diff
@@ -34,36 +34,52 @@
...
+    Path badContext = temporaryFolder.newFile().toPath();
+    CommonCliOptions commonCliOptions =
+        CommandLine.populateCommand(new CommonCliOptions(), "-t", "ignored");
+
+    IllegalArgumentException iae =
+        Assert.assertThrows(
+            IllegalArgumentException.class,
+            () -> CacheDirectories.from(commonCliOptions, badContext));
+    assertThat(iae.getMessage())
+        .isEqualTo("contextRoot must be a directory, but " + badContext.toString() + " is not.");
```
mpeddada1, line 82: nit: `assertThat().hasMessageThat().isEqualTo()`

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/cli2/CacheDirectoriesTest.java
```diff
@@ -34,36 +34,52 @@
...
+        .isEqualTo(Paths.get("test-project-cache").resolve("application-layers"));
+  }
+
+  @Test
+  public void testCacheDirectories_failIfContextIsNotDirectory() throws IOException {
+    Path badContext = temporaryFolder.newFile().toPath();
+    CommonCliOptions commonCliOptions =
+        CommandLine.populateCommand(new CommonCliOptions(), "-t", "ignored");
+
+    IllegalArgumentException iae =
```
mpeddada1, line 77: maybe `exception` instead of an abbreviation?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/cli2/Jar.java
```diff
@@ -84,7 +84,9 @@ public Integer call() {
         return 1;
       }
 
-      Containerizer containerizer = Containerizers.from(commonCliOptions, logger);
+      CacheDirectories cacheDirectories =
+          CacheDirectories.from(commonCliOptions, jarFile.toAbsolutePath().getParent());
```
chanseokoh, line 88: I don't think this will work. @mpeddada1 can you check what `jarFile.toAbsolutePath()` evaluates to when running `jib jar app.jar` (i.e., `jarFile` is simply `app.jar`).

mpeddada1: Yup. Calling `jarFile.toAbsolutePath()` with `jib jar sample.jar` results in `/Users/<username>/<path to jib folder>/sample.jar`. So it resolves it for the current directory.

chanseokoh: Oh, I was wrong for the whole time about `Path.toAbsolutePath()`. I realized its [Javadoc](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#toAbsolutePath()) says it actually resolves the path against a file system.

> Otherwise, this method resolves the path in an implementation dependent manner, typically by resolving the path against a file system default directory. 

> `/Users/<username>/<path to jib folder>/sample.jar`

You mean `.../<parent path of sample.jar>/sample.jar` in the end, right? For example, if the current directory is `/foo`, and if I run `foo $ jib bar/sample.jar`, then `jarFile.toAbsoluatePath()`, will correctly resolve to `/foo/bar/sample.jar` (as long as the file exists), right?

mpeddada1: > You mean `.../<parent path of sample.jar>/sample.jar` in the end, right?
Yes.

> You mean `.../<parent path of sample.jar>/sample.jar` in the end, right? For example, if the current directory is `/foo`, and if I run `foo $ jib bar/sample.jar`, then `jarFile.toAbsoluatePath()`, will correctly resolve to `/foo/bar/sample.jar` (as long as the file exists), right?

Or it would be `/Users/username/foo/bar/sample.jar` for unix and something like `C:\Users\username\foo\bar\sample.jar` for windows?

mpeddada1: ^ well depending on the filesystem default directory according to the doc

chanseokoh: Okay, you assumed the current working directory was `/Users/username/foo`. In my comment above, the working directory I assumed is `/foo`. So in my case, it will resolve to `/foo/bar/sample.jar`.

In any case, looks like `jarFile.toAbsoluatePath()` does the right thing that we want, so I guess no issue here.

chanseokoh (approved):

PR #2947, opened 2020-12-16
Add GitHub workflow to prepare Gradle release

chanseokoh:
Migrating `jib-gradle-plugin/prepare_release.sh` to a GitHub Actions workflow. Triggered manually.

Compared to the previous script, the workflow also automatically does the following:
- Creates a PR to bump the version.
- Creates a draft GitHub release with a template content.


.github/workflows/jib-gradle-prepare-release.yml
```diff
@@ -0,0 +1,63 @@
+name: Prepare Jib Gradle release
+on:
+  workflow_dispatch:
+    inputs:
+      release_version:
+        description: new release version
+        required: true
+        default: (for example, 0.1.0)
```
loosebazooka, line 8: does this default have some behavior that someone can't accidentally just say "okay" to "(for example, 0.1.0)" ?

loosebazooka (approved):

.github/workflows/jib-gradle-prepare-release.yml
chanseokoh: I can't find anything to that effect. But at least the code checks the version string and exits if not in the right format.
```

PR #2948, opened 2020-12-16
 Consolidate prepare release scripts and migrate to GitHub Actions

chanseokoh:
I was duplicating `jib-gradle-prepare-release.yml`, and it became too many, as we have 7 release targets (`build-plan`, `core`, `maven`, `gradle`, `extension-common`, `maven-extension`, and `gradle-extension`). I think it's better to consolidate them into a single GitHub Actions workflow.

Note, I should have renamed `jib-gradle-prepare-release.yml` to `jib-prepare-release.yml`, but that loses GitHub diff comparison. I'll rename the file in a separate PR.


.github/workflows/jib-gradle-prepare-release.yml
```diff
@@ -1,38 +1,61 @@
...
       - name: Check out code
         uses: actions/checkout@v2
 
-      - name: Build project
+      - name: Check input
         run: |
+          echo '* input project: "${{ github.event.inputs.project }}"'
+
+          case ${{ github.event.inputs.project }} in
+            build-plan|core|maven|gradle|extension-common|maven-extension|gradle-extension) ;;
```
loosebazooka, line 27: so we can't validate this in the inputs section above?

.github/workflows/jib-gradle-prepare-release.yml
```diff
@@ -41,24 +64,38 @@ jobs:
...
           # Provided by Actions. No need to create a token.
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          source_branch: gradle_release_v${{ github.event.inputs.release_version }}
-          pr_title: "Gradle release v${{ github.event.inputs.release_version }}"
+          source_branch: ${{ github.event.inputs.project }}-release-v${{ github.event.inputs.release_version }}
+          pr_title: "${{ github.event.inputs.project }} release v${{ github.event.inputs.release_version }}"
           pr_body: "To be merged after the release is complete."
 
-      - name: Draft GitHub release
+      - name: Draft Maven/Gradle GitHub release
```
loosebazooka, line 71: Just to clarify, we don't yet draft releases for the extension projects?

loosebazooka (approved):

.github/workflows/jib-gradle-prepare-release.yml
chanseokoh: I was not able to find any mentions about such a feature. [This article](https://levelup.gitconnected.com/how-to-manually-trigger-a-github-actions-workflow-4712542f1960) says

> Validation for these inputs is currently limited to simple presence checking, and the only input type available is a free text field (strings only).

Also found [this user question](https://github.community/t/can-workflow-dispatch-input-be-option-list/127338) asked on Aug 13 about a feature for the options-list UI.

.github/workflows/jib-gradle-prepare-release.yml
chanseokoh: Yeah, we never created GitHub releases for these. (We could have.)

Issue #2930, opened 2020-12-08
Deprecation warning for property jib.container.expandClasspathDependencies


juergenzimmermann:
**Environment**:

- *Jib version:* 2.7.0
- *Build tool:* Gradle 6.8-rc-1
- *OS:* Windows 10 Enterprise 2004

**Description of the issue**:
When I invoke `gradle jibDockerBuild` I get this deprecation warning:
```
Property 'jib.container.expandClasspathDependencies' @Input properties with primitive type 'boolean' cannot be @Optional. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0. See https://docs.gradle.org/6.8-rc-1/userguide/more_about_tasks.html#sec:up_to_date_checks for more details.
```

**Expected behavior**:
No warning.

**Steps to reproduce**:
Just a simple gradle project with the config below.

**`jib-gradle-plugin` Configuration**:
```kotlin
jib {
    from {
        image = "gcr.io/distroless/java-debian10:11"
    }
    to {
        image = project.name
    }
}
```

**Log output**: see above

loosebazooka:
this should be fixed

chanseokoh:
@juergenzimmermann 2.7.1 is released with the fix.

juergenzimmermann:
@chanseokoh Thank you!

PR #2952, opened 2020-12-21
Upgrade build to 6.0.1

loosebazooka:
- minimum to run jib is still 5.1

There are some eclipse related configurations that maybe we don't need anymore, but I don't have a good way to test eclipse right now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
-['jib-core', 'jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
+['jib-core', 'jib-gradle-plugin', 'jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
   project(projectName) {
     apply plugin: 'java-library'
   }
--- a/build.gradle
+++ b/build.gradle
@@ -22,7 +22,7 @@ idea.project.settings {
 import net.ltgt.gradle.errorprone.CheckSeverity
 
 // `java-library` must be applied before `java`
-// gradle-java-plugin (in jib-gradle-plugin) auto applies java-library, so ensure that happens first
+// java-gradle-plugin (in jib-gradle-plugin) auto applies java-library, so ensure that happens first
 ['jib-core', 'jib-gradle-plugin', 'jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
   project(projectName) {
     apply plugin: 'java-library'
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,8 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
+// `java-library` must be applied before `java`
+// gradle-java-plugin (in jib-gradle-plugin) auto applies java-library, so ensure that happens first
 ['jib-core', 'jib-gradle-plugin', 'jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
   project(projectName) {
     apply plugin: 'java-library'
--- a/jib-gradle-plugin/build.gradle
+++ b/jib-gradle-plugin/build.gradle
@@ -72,6 +72,7 @@ pluginBundle {
 }
 
 gradlePlugin {
+  testSourceSets sourceSets.integrationTest, sourceSets.test
   plugins {
     jibPlugin {
       id = 'com.google.cloud.tools.jib'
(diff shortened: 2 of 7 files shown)


build.gradle
```diff
@@ -21,7 +21,7 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
-['jib-core', 'jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
+['jib-core', 'jib-gradle-plugin', 'jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
```
chanseokoh, line 24: Why do we want to apply `java-library` for `jib-gradle-plugin`? It's not a Java library. And it's not in parity with `jib-maven-plugin` either.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -112,7 +112,6 @@ public void setExtraClasspath(List<String> classpath) {
   }
 
   @Input
-  @Optional
   public boolean getExpandClasspathDependencies() {
```
chanseokoh, line 115: I guess this fixes #2930. And I was wondering if we should fix it by converting it to a `Property` [as we did for `allowInsecureRegistries`](https://github.com/GoogleContainerTools/jib/blob/cc990bec2d5813e86a39a16da655f601c7d5604b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java#L95).

jib-gradle-plugin/build.gradle
```diff
@@ -72,6 +72,7 @@ pluginBundle {
 }
 
 gradlePlugin {
+  testSourceSets sourceSets.integrationTest, sourceSets.test
```
chanseokoh, line 75: Oh, we missed this? And I'm curious what the effect of adding this is?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
loosebazooka: Perhaps, but I think it's beyond the scope of this PR. (this just fixes the issue that a boolean can't be optional)

build.gradle
loosebazooka: ah, my comment is missing when I rewrote the PR. The java-gradle-plugin now auto applies the java-library plugin, so we must do this, I'll add in the comment.

jib-gradle-plugin/build.gradle
loosebazooka: For some reason, in gradle 6.0+, without this, the gradleTestKit isn't applied to integration tests (so anything where we exectute gradle would probably not work anymore).

build.gradle
chanseokoh: If `java-gradle-plugin` auto-applies `java-library`, then doesn't it mean we don't have to apply it manually here? (That is, no need to include `jib-gradle-plugin` in this line?

And the Travis "pr" build failed and this seems related. Interestingly, however, the "push" build didn't fail.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
chanseokoh: Okay, I think this is a simple fix that works.

build.gradle
loosebazooka: Right the reason is we need to apply `Java-library` before `java`

loosebazooka: or we get some error where the something is already configured that the `java-library` plugin wants to modify.

build.gradle
```diff
@@ -21,7 +21,9 @@ idea.project.settings {
 
 import net.ltgt.gradle.errorprone.CheckSeverity
 
-['jib-core', 'jib-gradle-plugin-extension-api', 'jib-maven-plugin-extension-api'].each { projectName ->
+// `java-library` must be applied before `java`
+// gradle-java-plugin (in jib-gradle-plugin) auto applies java-library, so ensure that happens first
```
chanseokoh, line 25:
```suggestion
// java-gradle-plugin (in jib-gradle-plugin) auto applies java-library, so ensure that happens first
```

chanseokoh (approved):
LGTM. The Travis "pr" build will probably fail though.

loosebazooka:
maybe it's merging funny, will try with a rebase

loosebazooka:
I don't think we should be applying the gradle-java-plugin in the cli. I'm not even sure we should use the gradle testkit, perhaps we should just exec gradlew directly

chanseokoh (approved):
According to the commit comment of 180ac64, is it a temporary fix you are going to revert soon?

loosebazooka:
I don't intend on reverting it, I want to clean up that test to use a direct execution of gradle

PR #2956, opened 2020-12-22
Hide and disbale Jib CLI jar sub-command

chanseokoh:
I was not able to find an option in picocli to disable a command. Maybe `hidden` is good enough. But I also added a guard to prevent it from running. (picocli still does argument validation though.)


loosebazooka:
interesting

loosebazooka (approved):

PR #2961, opened 2020-12-23
Fix Gradle config for Eclipse IDE

chanseokoh:
- `eclipse.classpath.plusConfigurations`: https://stackoverflow.com/a/36651485/1701388
- `setCanBeResolved(true)`: followed https://stackoverflow.com/a/55232236/1701388 although I don't know what this implies.
- Also excluded some test project resources from source in jib-cli

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -212,6 +212,7 @@ subprojects {
 
   configurations {
     integrationTestImplementation.extendsFrom testImplementation
+    integrationTestImplementation.setCanBeResolved(true)
     integrationTestRuntime.extendsFrom testRuntime
   }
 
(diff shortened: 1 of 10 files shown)


loosebazooka:
This seems kinda similar to https://github.com/GoogleContainerTools/jib/pull/2958/files, are the integration tests marked as "test" modules in this mechanism?

chanseokoh:
Yeah, `src/integration-test/{java,resources}` are still marked as "test" in my Eclipse after this change.

This PR resolves different issues than what #2958 used to do; it's unrelated to marking them "test." Without `eclipse.classpath.plusConfigurations`, Eclipse cannot resolve classes defined in `integrationTestImplementation`.

loosebazooka (approved):

PR #2963, opened 2020-12-23
Set commit author to the workflow triggering user for Jib CLI release

chanseokoh:


.github/workflows/jib-cli-release.yml
```diff
@@ -53,7 +52,8 @@ jobs:
           source_branch: cli-release-v${{ github.event.inputs.release_version }}
           pr_title: "CLI release v${{ github.event.inputs.release_version }}"
           pr_body: "To be merged after the release is complete."
-    
+          pr_label: "kokoro:run"
```
loosebazooka, line 55: why this? is it because the github user no longer auto triggers a kokoro run?

loosebazooka (approved):

.github/workflows/jib-cli-release.yml
loosebazooka: I thought github emails associated with our accounts would still trigger kokoro?

chanseokoh: I don't know why, but `<github user>@users.noreply.github.com` as a PR author (as opposed to `@google.com`) doesn't trigger Kokoro (at least in my case). I also tried manually associating the email to my GitHub account on the settings page (the settings email page doesn't display the email in the list), but of course, it complains that the email is already in use. Maybe this should be configured on the Kokoro side?

PR #2965, opened 2020-12-28
Refactoring JarModeProcessor 

### LINKED ISSUE #2934: [jib-cli-jar] Restructure JarModeProcessor
Details of design are present in the comment below: 

Nothing specific about this, but about the layout in general. JarModeProcesssor seems to be doing too much work. I think I would imagine it more like:

```
interface JarProcessor
 |_ StandardExplodedProcessor
 |_ StandardPackagedProcessor
 |_ SpringBootPackagedProcessor
 |_ ...
```

Where each of those classes implements the JarProcessor interface? We can talk about this in the design meeting?

_Originally posted by @loosebazooka in https://github.com/GoogleContainerTools/jib/issues/2918#issuecomment-741126761_

mpeddada1:
Fixes #2934 
Modified JarModeProcessor to be an interface which would be implemented by `StandardExplodedModeProcessor`, `StandardPackagedModeProcessor`, `SpringBootExplodedModeProcessor` and `SpringBootPackagedModeProcessor`.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarModeProcessor.java
```diff
@@ -16,431 +16,28 @@
...
-        FileEntriesLayer.builder()
-            .setName(JAR)
-            .addEntry(jarPath, APP_ROOT.resolve(jarPath.getFileName()))
-            .build();
-    layers.add(jarLayer);
-
-    return layers;
-  }
+/** Interface to create layers and compute entrypoint from jar file contents. */
+public interface JarModeProcessor {
```
mpeddada1, line 25: Unfortunately, I had to modify `createLayers()` and `computeEntrypoint()` to no longer accept parameters for the sake of consistency. In the case of `createLayers()`, while explode mode requires a temporary directory, packaged mode doesn't. On the other hand, `computeEntrypoint()` for spring-boot exploded **doesn't** require the `jarPath` but the same for spring-boot packaged **does**. The `jarPath` and `temporaryDirectory` are now set using setters.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -95,13 +100,33 @@ public Integer call() {
         return 1;
       }
 
+      JarProcessorHelper.JarType jarType = JarProcessorHelper.determineJarType(jarFile);
```
loosebazooka, line 103: I like the direction of this. But instead of JarProcessorHelper returning a type, perhaps it could return a JarProcessor instance?

something where the code might end up looking like:
```
JarProcessor processor = JarProcessors.for(jarFile, temporaryDirectoryProvider)
containerBuilder = JarFiles.toJibContainerBuilder(processor)
...
```

mpeddada1: Ah that's an interesting idea. Would that mean moving all this logic to JarProcessorHelper and making it return a JarModeProcessor instance?

loosebazooka: I think so, you may not need the "type" enum anymore?

mpeddada1: I like the idea of moving all the logic out of `Jar`. I'm just not sure about having it in `JarModeProcessorHelper` since it is supposed to contain only redundant code for the JarModeProcessor? That being said, it was probably a mistake to have it contain `determineJarType` at all. How does creating a separate class for this sound?

loosebazooka: Yeah sure.

These are just my ideas, but I think something along these lines could work?
`JarProcessors` -> has static method for JarProcessor creation `from`
Remove `Mode` from most names since we're not exposing that anymore
`JarModeProcessor` -- rename --> `JarProcessor`
`StandardPackagedModeProcessor` -> `StandardPackagedJarProcessor`... etc?

`JarProcessorHelper` is kind of a strange name because it doesn't truly say what it is doing (which is the trap that we often fall into when writing helpers), perhaps it could be called `JarLayers` or something?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessorHelper.java
```diff
@@ -0,0 +1,156 @@
...
+                    mode.equals(ProcessingMode.packaged)
+                        ? APP_ROOT.resolve(path)
+                        : APP_ROOT.resolve(DEPENDENCIES).resolve(path.getFileName())));
+        layers.add(snapshotLayer.build());
+      }
+      return layers;
+    }
+  }
+
+  static FileEntriesLayer addDirectoryContentsToLayer(
```
loosebazooka, line 127: could this be called `getDirectoryContentsAsLayer` or something?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
mpeddada1: Yes that's a good idea. `JarProcessors` may sometimes be mistaken for `JarProcessor`. I'm still struggling to find a fitting name but what do you think of `JarProcessorSelector`?

> `JarProcessorHelper` is kind of a strange name because it doesn't truly say what it is doing (which is the trap that we often fall into when writing helpers), perhaps it could be called `JarLayers` or something?
`JarLayers` is an accurate name, thank you! It describes the intent of the class pretty well.

loosebazooka: The name `JarProcessors` is a pretty common pattern in java, `Path` and `Paths` and we also use it all over the place. I'm not too concerned with any confusion there.

mpeddada1: Ok that's reassuring, thanks!

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -95,13 +97,12 @@ public Integer call() {
         return 1;
       }
 
+      JarProcessor processor =
+          JarProcessors.from(jarFile, tempDirectoryProvider.newDirectory(), mode);
```
loosebazooka, line 101: Can we only instantiate a new temporary directory when required?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -0,0 +1,76 @@
...
+   * @param temporaryDirectory path to temporary directory
+   * @param mode processing mode
+   * @return JarProcessor
+   * @throws IOException if I/O error occurs when opening the jar file
+   */
+  public static JarProcessor from(Path jarPath, Path temporaryDirectory, ProcessingMode mode)
+      throws IOException {
+    String jarType = determineJarType(jarPath);
+    if (jarType.equals(SPRING_BOOT) && mode.equals(ProcessingMode.packaged)) {
+      SpringBootPackagedProcessor processor = new SpringBootPackagedProcessor();
```
loosebazooka, line 41: Can these JarProcessor types can just accept the necessary inputs at the constructor level? `setJarPath` or `setTemporaryDirectory` seems to be only called during instantiation?

Then the types become more `immutable`.

loosebazooka (commented):
Looks goods, some minor stuff I should've caught earlier.

dstiers11 (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
mpeddada1: Right, great point!

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorsTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+
+/** Tests for {@link JarProcessors}. */
+@RunWith(MockitoJUnitRunner.class)
+public class JarProcessorsTest {
+
+  private static final String SPRING_BOOT = "jar/spring-boot/springboot_sample.jar";
+  private static final String STANDARD = "jar/standard/emptyStandardJar.jar";
+
+  @Mock private static TempDirectoryProvider mockTemporaryDirectoryProvider;
```
loosebazooka, line 39: I would check whether this mock is called or not? We only expect it to create a new directory for 2 of the 4 types?
And I would expect no interactions on the other 2?

loosebazooka (approved):
Just a few test checks.

Issue #2871, opened 2020-10-29
[jib-cli-jar] Intelligently determine whether to use AdoptOpenJDK base image for java8 or java 11 based on version of application. 


mpeddada1:
The Jib CLI Jar will start using an AdoptOpenJDK base image instead of distroless.

mpeddada1:
Closed with #3075

PR #2972, opened 2021-01-06
Add option to specify base image for cli jar command

### LINKED ISSUE #2923: [jib-cli-jar] Add option to specify base image in jib cli jar command
The jib cli jar subcommand currently only uses the distroless base image (the default) but we should add a `--base-image` parameter to allow users to specify a base image of their choice.

mpeddada1:
Closes #2923
This PR adds the `--from` option to the jib cli jar subcommand. It will be used to specify the desired base image.

### DIFF
--- a/jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
+++ b/jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
@@ -247,10 +247,8 @@ public class JarCommandTest {
         new CommandLine(new JibCli())
             .execute(
                 "jar",
-                "--target",
-                "docker://cli-gcr-base",
-                "--from",
-                "gcr.io/google-appengine/openjdk:8",
+                "--target=docker://cli-gcr-base",
+                "--from=gcr.io/google-appengine/openjdk:8",
                 jarPath.toString());
     assertThat(exitCode).isEqualTo(0);
     String output = new Command("docker", "run", "--rm", "cli-gcr-base").run();
--- a/jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
+++ b/jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
@@ -239,6 +239,24 @@ public class JarCommandTest {
     assertThat(getContent(new URL("http://localhost:8080"))).isEqualTo("Hello world");
   }
 
+  @Test
+  public void testJar_baseImageSpecified()
+      throws IOException, URISyntaxException, InterruptedException {
+    Path jarPath = Paths.get(Resources.getResource("jarTest/standard/noDependencyJar.jar").toURI());
+    Integer exitCode =
+        new CommandLine(new JibCli())
+            .execute(
+                "jar",
+                "--target",
+                "docker://cli-gcr-base",
+                "--from",
+                "gcr.io/google-appengine/openjdk:8",
+                jarPath.toString());
+    assertThat(exitCode).isEqualTo(0);
+    String output = new Command("docker", "run", "--rm", "cli-gcr-base").run();
+    assertThat(output).isEqualTo("Hello World");
+  }
+
   @Nullable
   private static String getContent(URL url) throws InterruptedException {
     for (int i = 0; i < 40; i++) {
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
@@ -36,11 +36,11 @@ import java.nio.file.Paths;
 import java.util.List;
 import java.util.stream.Collectors;
 
-/** Creates jib container builder. */
+/** Helper class for creating JibContainerBuilders from JibCli specifications. */
 public class ContainerBuilders {
 
   /**
-   * Generates a {@link JibContainerBuilder} depending on the base image specified.
+   * Creates a {@link JibContainerBuilder} depending on the base image specified.
    *
    * @param baseImageReference base image
    * @param platforms platforms for multi-platform support in build command.
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
@@ -33,8 +33,7 @@ import com.google.cloud.tools.jib.plugins.common.DefaultCredentialRetrievers;
 import com.google.cloud.tools.jib.plugins.common.logging.ConsoleLogger;
 import java.io.FileNotFoundException;
 import java.nio.file.Paths;
-import java.util.List;
-import java.util.stream.Collectors;
+import java.util.Set;
 
 /** Helper class for creating JibContainerBuilders from JibCli specifications. */
 public class ContainerBuilders {
@@ -42,8 +41,8 @@ public class ContainerBuilders {
   /**
    * Creates a {@link JibContainerBuilder} depending on the base image specified.
    *
-   * @param baseImageReference base image
-   * @param platforms platforms for multi-platform support in build command.
+   * @param baseImageReference base image reference
+   * @param platforms platforms for multi-platform support in build command
    * @param commonCliOptions common cli options
    * @param logger console logger
    * @return a {@link JibContainerBuilder}
@@ -52,7 +51,7 @@ public class ContainerBuilders {
    */
   public static JibContainerBuilder create(
       String baseImageReference,
-      List<Platform> platforms,
+      Set<Platform> platforms,
       CommonCliOptions commonCliOptions,
       ConsoleLogger logger)
       throws InvalidImageReferenceException, FileNotFoundException {
@@ -76,7 +75,7 @@ public class ContainerBuilders {
         .forEach(registryImage::addCredentialRetriever);
     JibContainerBuilder containerBuilder = Jib.from(registryImage);
     if (!platforms.isEmpty()) {
-      containerBuilder.setPlatforms(platforms.stream().collect(Collectors.toSet()));
+      containerBuilder.setPlatforms(platforms);
     }
     return containerBuilder;
   }
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2021 Google LLC.
+ *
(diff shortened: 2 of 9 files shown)


jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
```diff
@@ -217,4 +237,34 @@ public void testToJibContainerBuilder_packagedSpringBoot_basicInfo()
                 .build()
                 .getEntries());
   }
+
+  @Test
+  public void testToJibContainerBuilder_dockerBaseImage()
+      throws IOException, InvalidImageReferenceException {
```
mpeddada1, line 243: Currently, I can't think of a good way to test for the tar base image without accessing the build context (similar to what's being done in PluginConfigurationProcessorTest)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
```diff
@@ -30,20 +46,55 @@
...
     List<FileEntriesLayer> layers = processor.createLayers();
     List<String> entrypoint = processor.computeEntrypoint();
     containerBuilder.setEntrypoint(entrypoint).setFileEntriesLayers(layers);
     return containerBuilder;
   }
+
+  private static JibContainerBuilder createJibContainerBuilder(
+      String from, CommonCliOptions commonCliOptions, ConsoleLogger logger)
+      throws InvalidImageReferenceException, FileNotFoundException {
+    String baseImageReference = from;
```
chanseokoh, line 79: It's unfortunate that the code is duplicated here and in `BuildFiles`. @loosebazooka thoughts?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
chanseokoh: Ah, you mean `ContainerBuildPlan.getBaseImage()` is just an image reference without `docker://` or `tar://`, so it's not possible to check if `JibContainerBuilder` was correctly configured, right?

One way (which I don't really like) is to create an [indirection class to expose a package-protected method](https://github.com/GoogleContainerTools/jib/blob/master/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/api/JibContainerBuilderTestHelper.java), which is what `PluginConfigurationProcessorTest` did. The helper is under the `c.g.c.t.jib.api` package, so it can access `toBuildContext()`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
loosebazooka: Yeah I think the two methods can be merged into one with the signature:

```
(String image, @Nullable List<Platform> platforms, CommonCliOptions commonCliOptions, ConsoleLogger logger)
```

We could put it in a new file `ContainerBuilders.java` for now next to `Containerizers` (in `com.google.cloud.tools.jib.cli`). We can move it around later if it makes more sense?

mpeddada1: Sounds like a good idea. I'll make this change in this PR since it doesn't look like a major refactoring?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarFilesTest.java
mpeddada1: Ah yes, exactly. Thank you! Modified the tests with this solution.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
loosebazooka: sgtm

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -111,40 +102,20 @@ public static JibContainerBuilder toJibContainerBuilder(
...
-                imageReference,
-                logEvent -> logger.log(logEvent.getLevel(), logEvent.getMessage())));
-    Credentials.getFromCredentialRetrievers(commonCliOptions, defaultCredentialRetrievers)
-        .forEach(registryImage::addCredentialRetriever);
-    JibContainerBuilder containerBuilder = Jib.from(registryImage);
-    if (!from.getPlatforms().isEmpty()) {
-      containerBuilder.setPlatforms(
-          from.getPlatforms()
+    List<Platform> platforms = new ArrayList<>();
+    if (!baseImageSpec.getPlatforms().isEmpty()) {
```
chanseokoh, line 109: This `if (...)` probably isn't necessary. For an empty list, stream will return an empty collection. So just `List<Platform> platforms = baseImageSpec.getPlatforms().stream().map().collect()`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
```diff
@@ -0,0 +1,83 @@
...
+    DefaultCredentialRetrievers defaultCredentialRetrievers =
+        DefaultCredentialRetrievers.init(
+            CredentialRetrieverFactory.forImage(
+                imageReference,
+                logEvent -> logger.log(logEvent.getLevel(), logEvent.getMessage())));
+    Credentials.getFromCredentialRetrievers(commonCliOptions, defaultCredentialRetrievers)
+        .forEach(registryImage::addCredentialRetriever);
+    JibContainerBuilder containerBuilder = Jib.from(registryImage);
+    if (!platforms.isEmpty()) {
+      containerBuilder.setPlatforms(platforms.stream().collect(Collectors.toSet()));
```
chanseokoh, line 79: If you make `create()` accept `Set<Platform>`, it won't need `stream().collect()`.

BTW, [this link](https://stackoverflow.com/q/27611896/1701388) says `toSet()` doesn't preserve the order. The order of platforms is important, so that's why the original code used `Collectors.toCollection(LinkedHashSet::new))`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.stream.Collectors;
+
+/** Helper class for creating JibContainerBuilders from JibCli specifications. */
+public class ContainerBuilders {
+
+  /**
+   * Creates a {@link JibContainerBuilder} depending on the base image specified.
+   *
+   * @param baseImageReference base image
+   * @param platforms platforms for multi-platform support in build command.
```
chanseokoh, line 46: nit:

```suggestion
   * @param platforms platforms for multi-platform support in build command
```

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
```diff
@@ -239,6 +239,24 @@ public void testSpringBootJar_packagedMode() throws IOException, InterruptedExce
...
+      throws IOException, URISyntaxException, InterruptedException {
+    Path jarPath = Paths.get(Resources.getResource("jarTest/standard/noDependencyJar.jar").toURI());
+    Integer exitCode =
+        new CommandLine(new JibCli())
+            .execute(
+                "jar",
+                "--target",
+                "docker://cli-gcr-base",
+                "--from",
+                "gcr.io/google-appengine/openjdk:8",
```
chanseokoh, line 253: nit: how about this?

```suggestion
                "jar",
                "--target=docker://cli-gcr-base",
                "--from=gcr.io/google-appengine/openjdk:8",
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -111,40 +102,20 @@ public static JibContainerBuilder toJibContainerBuilder(
...
+      platforms =
+          baseImageSpec
+              .getPlatforms()
               .stream()
               .map(
                   platformSpec ->
                       new Platform(platformSpec.getArchitecture(), platformSpec.getOs()))
-              .collect(Collectors.toCollection(LinkedHashSet::new)));
+              .collect(Collectors.toList());
```
chanseokoh, line 117: `collect(toCollection(LinkedHashSet::new))`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/ContainerBuildersTest.java
```diff
@@ -0,0 +1,111 @@
...
+  @Mock private CommonCliOptions mockCommonCliOptions;
+
+  @Mock private ConsoleLogger mockLogger;
+
+  @Test
+  public void testCreate_dockerBaseImage()
+      throws IOException, InvalidImageReferenceException, CacheDirectoryCreationException {
+    JibContainerBuilder containerBuilder =
+        ContainerBuilders.create(
+            "docker://docker-image-ref", Collections.EMPTY_SET, mockCommonCliOptions, mockLogger);
```
chanseokoh, line 54: Sorry, missed this. Let's use `Collections.emptyList()` and `.emptySet()`. https://stackoverflow.com/q/14870819/1701388

chanseokoh (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
```diff
@@ -30,16 +35,27 @@
...
+  public static JibContainerBuilder toJibContainerBuilder(
+      JarProcessor processor,
+      Jar jarOptions,
+      CommonCliOptions commonCliOptions,
+      ConsoleLogger logger)
       throws IOException, InvalidImageReferenceException {
 
-    // Use distroless as the base image.
-    JibContainerBuilder containerBuilder = Jib.from("gcr.io/distroless/java");
+    // Use distroless as the default base image.
```
loosebazooka, line 53: I think here we should try to check if the jar is java8 or java11 and chose the right distroless? (perhaps in a followup PR)

loosebazooka (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
mpeddada1: Yup, good point - we have an issue for this #2871. Will make this change in the next PR.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java
mpeddada1: Ah my bad, thank you!

PR #2975, opened 2021-01-07
Add more options to cli jar command 

mpeddada1:
Options added:
- `--jvm-flags`
- `--exposed-ports`
- `--volumes`
- `--environment-variables`
- `--labels`
- `-u`, `--user`
- `--image-format`
- `--program-args`


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -67,11 +78,71 @@
...
+      paramLabel = "<base-image>",
       description = "The base image to use.")
   @SuppressWarnings("NullAway.Init") // initialized by picocli
   private String from;
 
+  @CommandLine.Option(
+      names = "--jvm-flags",
+      paramLabel = "<jvm-flags>",
+      split = ",",
+      description = "JVM arguments.")
```
chanseokoh, line 90: I think it's worth adding a comma-separated usage example. `-Dmy.property=value,-Xshare:off`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -67,11 +78,71 @@
...
+  private List<String> jvmFlags = Collections.emptyList();
+
+  @CommandLine.Option(
+      names = "--expose",
+      paramLabel = "<exposed-ports>",
+      split = ",",
+      description = "Ports to expose on container.")
```
chanseokoh, line 97: Example like `5000,7/dup`. I believe my example works but needs to confirm.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -67,11 +78,71 @@
...
+  private List<String> exposedPorts = Collections.emptyList();
+
+  @CommandLine.Option(
+      names = "--volumes",
+      paramLabel = "<volumes>",
+      split = ",",
+      description = "Directories on container to hold extra volumes.")
```
chanseokoh, line 104: Example, but I forgot the exact format required.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/SpringBootExplodedProcessor.java
```diff
@@ -117,12 +117,14 @@ public SpringBootExplodedProcessor(Path jarPath, Path tempDirectoryPath) {
...
 
   @Override
-  public ImmutableList<String> computeEntrypoint() {
-    return ImmutableList.of(
-        "java",
-        "-cp",
-        JarLayers.APP_ROOT.toString(),
-        "org.springframework.boot.loader.JarLauncher");
+  public ImmutableList<String> computeEntrypoint(List<String> jvmFlags) {
+    List<String> entrypoint = new ArrayList<>(4 + jvmFlags.size());
```
chanseokoh, line 121: nit: how about `ImmutableList.builder()`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
```diff
@@ -58,8 +58,17 @@ public static JibContainerBuilder toJibContainerBuilder(
...
     List<FileEntriesLayer> layers = processor.createLayers();
-    List<String> entrypoint = processor.computeEntrypoint();
+    List<String> entrypoint = processor.computeEntrypoint(jarOptions.getJvmFlags());
+
     containerBuilder.setEntrypoint(entrypoint).setFileEntriesLayers(layers);
+    containerBuilder.setExposedPorts(jarOptions.getExposedPorts());
+    containerBuilder.setVolumes(jarOptions.getVolumes());
+    containerBuilder.setEnvironment(jarOptions.getEnvironment());
+    containerBuilder.setLabels(jarOptions.getLabels());
+    containerBuilder.setProgramArguments(jarOptions.getProgramArguments());
```
chanseokoh, line 68: nit: chain these? `.setExposedPorts(...).setVolumes(...).setEnvironment(...)`

chanseokoh:
I think we should support `--entrypiont` too. But this may complicate the code, so maybe in a separate PR?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
mpeddada1: Good idea!

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
mpeddada1: Thank you! I confirmed that your example works.

mpeddada1:
Yup, I agree. I can add this option in a separate PR.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -67,11 +78,72 @@
...
+          "Directories on container to hold extra volumes,  example: --volumes=/var/log,/var/log2.")
+  private List<String> volumes = Collections.emptyList();
+
+  @CommandLine.Option(
+      names = "--environment-variables",
+      paramLabel = "<environment-variables>",
+      split = ",",
+      description =
+          "Environment variables to write into container, example: --environment-variables env1=env_value1,env2=env_value2.")
+  private Map<String, String> environment = new LinkedHashMap<>();
```
chanseokoh, line 114: Perhaps for consistency, `Collections.emptyMap()`. If `Collections.emptyList()` worked for others, `emptyMap()` should work. It's certain that no one is modifying the initially assigned collection, because `Collections.empty???()` are immutable. (They throw an exception if someone tries to modify it.)

chanseokoh (approved):
LGTM, pending a minor comment.

chanseokoh (approved):

PR #2981, opened 2021-01-08
Update issue templates (new mechanism)

loosebazooka:


chanseokoh (approved):
Should we remove `ISSUE_TEMPLATE.md`?

loosebazooka:
I think so? But I'm not sure I fully understand what is going to happen when I submit this.

PR #2988, opened 2021-01-11
Build Failure if JAR requires Java >11 

mpeddada1:
Make `gcr.io/distroless/java:11` base image the default and throw `IncompatibleBaseImageJavaVersionException` if the provided jar needs to be run with java 11+.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -34,10 +42,17 @@
...
+   * @throws IncompatibleBaseImageJavaVersionException if the base image java version cannot support
+   *     this build
    */
   public static JarProcessor from(
       Path jarPath, TempDirectoryProvider temporaryDirectoryProvider, ProcessingMode mode)
-      throws IOException {
+      throws IOException, IncompatibleBaseImageJavaVersionException {
+    Integer jarVersion = getVersion(jarPath);
+    if (jarVersion > 11) {
+      throw new IncompatibleBaseImageJavaVersionException(11, jarVersion);
```
chanseokoh, line 53: In build plugins, we catch this exception at a higher stack and provide a detailed error message with an action item. Maybe we can throw an `IllegalStateException`? The message should report the app version, what the problem is, and how to fix.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -65,4 +80,40 @@ private static String determineJarType(Path jarPath) throws IOException {
...
+
+  /**
+   * Determines the java version of JAR. Derives the version from the first .class file it finds in
+   * the JAR.
+   *
+   * @param jarPath path to the jar
+   * @return java version
+   * @throws IOException if I/O exception thrown when opening the jar file
+   */
+  private static Integer getVersion(Path jarPath) throws IOException {
```
chanseokoh, line 92: Maybe `getJavaMajorVersion`?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -65,4 +80,40 @@ private static String determineJarType(Path jarPath) throws IOException {
...
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      Enumeration<JarEntry> jarEntries = jarFile.entries();
+      while (jarEntries.hasMoreElements()) {
+        String jarEntry = jarEntries.nextElement().toString();
+        if (jarEntry.endsWith(".class")) {
+          URLClassLoader loader = new URLClassLoader(new URL[] {jarPath.toUri().toURL()});
+          try (InputStream classFileInfo = loader.getResourceAsStream(jarEntry)) {
+            DataInputStream data = new DataInputStream(classFileInfo);
```
chanseokoh, line 100: Can we do this?

```suggestion
          try (InputStream classFileInfo = loader.getResourceAsStream(jarEntry);
              DataInputStream data = new DataInputStream(classFileInfo)) {
```
And why is it called classFile*Info*? Isn't it really a class file?

Or, in most cases, I believe this is OK:
```java
try (InputStream classFile = new DataInputStream(loader.getResourceAsStream(...))) {
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -65,4 +80,40 @@ private static String determineJarType(Path jarPath) throws IOException {
...
+
+            // Check magic number
+            if (data.readInt() != 0xCAFEBABE) {
+              throw new IOException("Invalid class file format.");
```
chanseokoh, line 104: I guess this should work with the Java 9 [`module-info.class`](https://stackoverflow.com/q/57358750/1701388), but it's worth testing.

I found a [simple example](https://github.com/tfij/Java-9-modules---the-simplest-example), where running `compile.sh` generates `build/pl.tfij.java9modules.app/module-info.class` and running `package.sh` generates `mods/*.jar`. I'd also be interesting to see if `jib jar pl.tfij.java9modules.app@1.0.jar` works fine.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
```diff
@@ -55,7 +55,7 @@ public static JibContainerBuilder toJibContainerBuilder(
         jarOptions.getFrom().isPresent()
             ? ContainerBuilders.create(
                 jarOptions.getFrom().get(), Collections.emptySet(), commonCliOptions, logger)
-            : Jib.from("gcr.io/distroless/java");
+            : Jib.from("gcr.io/distroless/java:11");
```
loosebazooka, line 58: Just curious: is this the debian10 one?

loosebazooka:
title should be `> 11` instead of `11+` ?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
chanseokoh: No. Now you said it, I think we should use `gcr.io/distroless/java-debian10:11` at least for Java 11.

Currently, the build plugins use `gcr.io/distroless/java:11` in the hope that the default entrypoint `gcr.io/distroless/java:11` would eventually roll over to Debian 10. However, I don't think it'll happen, so for Jib CLI, we can start with `gcr.io/distroless/java-debian10:11`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
mpeddada1: Oh nice. yes, it is worth double-checking. I tested it with the example you suggested above and can confirm that the version calculated is indeed 9.
A snippet of the test:
```
  Path jarPath = Paths.get(Resources.getResource(JAVA_9_JAR).toURI());
  Integer version = JarProcessors.getVersion(jarPath);
  assertThat(version).isEqualTo(9);
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
mpeddada1: Thanks for bringing this up. I'll change the default to `gcr.io/distroless/java-debian10:11 `

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
chanseokoh: Oh, I should have been clearer. Can you test this section of code on the compiled `module-info.class`? If you use a jar, it's possible that the code retrieved the version from a different `.class` file. I'm trying to see if the `module-info.class` uses the same magic number and the structure.

Also, have you tested `jib jar pl.tfij.java9modules.app@1.0.jar`?

chanseokoh: BTW, it doesn't have to be compiled with Java 9. If you use Java 11, it should report 11.

mpeddada1: Ahh I see, sorry. I tested this again and there's something interesting happening here. This method returns 9 for `module-info.class` but 11 (which is what I used while running `compile.sh` and `package.sh`) for `pl.tfij.java9modules.app.ModuleApp.class`. I need to look into this more but I'm wondering if `module-info.class` is always in java 9?

> Also, have you tested `jib jar pl.tfij.java9modules.app@1.0.jar`?

And running this results in a ` java.lang.ClassNotFoundException: pl.tfij.java9modules.greetings.Greeting`. In terms of how our code is set up, since that jar's manifest doesn't specify a Class-path, there isn't a way for us to determine what the dependency is.

chanseokoh: Ah, that makes sense. I think we should skip `module-info.class`. Not sure why, but it looks like the number is always Java 9.

> In terms of how our code is set up, since that jar's manifest doesn't specify a `Class-path`, there isn't a way for us to determine what the dependency is.

Aha. I see that `run.sh` doesn't run the jar with `java -jar` but with `java --module-path ...` (which corresponds to `java --class-path` in Java <=8 in some sense). I tried to do a quick search, but I'm not sure how one can build a runnable jar with modules that require dependency jars. We should revisit the Java 9+ module at some point. For this PR, let's just focus on determining the correct Java version.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -65,4 +80,40 @@ private static String determineJarType(Path jarPath) throws IOException {
...
+   * @param jarPath path to the jar
+   * @return java version
+   * @throws IOException if I/O exception thrown when opening the jar file
+   */
+  private static Integer getJavaMajorVersion(Path jarPath) throws IOException {
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      Enumeration<JarEntry> jarEntries = jarFile.entries();
+      while (jarEntries.hasMoreElements()) {
+        String jarEntry = jarEntries.nextElement().toString();
+        if (jarEntry.endsWith(".class") && !jarEntry.equals("module-info.class")) {
```
chanseokoh, line 97: Is `jarEntry` a filename without path? If not, maybe it should be `!jarEntry.endsWith("module-info.class")`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -38,6 +44,15 @@
   public static JarProcessor from(
       Path jarPath, TempDirectoryProvider temporaryDirectoryProvider, ProcessingMode mode)
       throws IOException {
+    Integer jarJavaVersion = getJavaMajorVersion(jarPath);
+    if (jarJavaVersion > 11) {
+      throw new IllegalStateException(
+          "This application is in java "
+              + jarJavaVersion
+              + ". It is incompatible with the default base image which only supports versions up to java 11. "
+              + "Please consider specifying a base image of your choice.");
```
chanseokoh, line 53: How about this?

The input JAR (jarPath) is compiled with Java x, but the default base image only supports versions up to Java 11. Specify a custom base image with `--base-image` (if this is the right option).

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
mpeddada1: good catch, thanks!

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -65,4 +84,41 @@ private static String determineJarType(Path jarPath) throws IOException {
...
+   *
+   * @param jarPath path to the jar
+   * @return java version
+   * @throws IOException if I/O exception thrown when opening the jar file
+   */
+  @VisibleForTesting
+  static Integer getJavaMajorVersion(Path jarPath) throws IOException {
+    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
+      List<String> jarEntries =
+          jarFile.stream().map(JarEntry::toString).collect(Collectors.toList());
```
mpeddada1, line 100: I realized that using .nextElement() results in a `NoSuchElementException` if there are no more elements - this would cause an issue if an empty jar was passed in.

chanseokoh: But you were checking `hasMoreElements()`. It will never fail.

I liked the previous code with `Enumeration`. Although performance-wise, I think there's zero impact, but it allowed short-circuiting. (Now `stream().collect()` needs to process all elements.)

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -17,20 +17,23 @@
...
 import java.util.jar.JarFile;
+import java.util.stream.Collectors;
 
 /** Class to create a {@link JarProcessor} instance depending on jar type and processing mode. */
 public class JarProcessors {
   private static String SPRING_BOOT = "spring-boot";
   private static String STANDARD = "standard";
-  private static Integer VERSION_NOT_FOUND = 0;
+
+  @VisibleForTesting static Integer VERSION_NOT_FOUND = 0;
```
chanseokoh, line 36: I'm fine just using the number 0 in the test. Actually, the test can verify the literal (`assertThat(version).isEqualTo(0)`).

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorsTest.java
```diff
@@ -74,4 +76,25 @@ public void testFrom_springBootExploded() throws IOException, URISyntaxException
...
+    Path jarPath = Paths.get(Resources.getResource(JAVA_14_JAR).toURI());
+    IllegalStateException exception =
+        assertThrows(
+            IllegalStateException.class,
+            () ->
+                JarProcessors.from(
+                    jarPath, mockTemporaryDirectoryProvider, ProcessingMode.exploded));
+    assertThat(exception)
+        .hasMessageThat()
+        .startsWith("The input JAR (" + jarPath + ") is compiled with Java " + 14);
```
chanseokoh, line 91:
```suggestion
        .startsWith("The input JAR (" + jarPath + ") is compiled with Java 14");
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorsTest.java
```diff
@@ -74,4 +76,25 @@ public void testFrom_springBootExploded() throws IOException, URISyntaxException
...
+  }
+
+  @Test
+  public void testGetVersion_versionNotFound() throws URISyntaxException, IOException {
```
chanseokoh, line 95:
```suggestion
  public void testGetJavaMajorVersion_versionNotFound() throws URISyntaxException, IOException {
```

chanseokoh (commented):
Looks good to me, pending some minor comments.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
mpeddada1: Oh no, my bad. For some reason, I had an extra `.nextElement()` in the while loop when I was testing this locally. Sorry about the confusion! Reverting this change.

chanseokoh (approved):

PR #2999, opened 2021-01-13
Add registry mirror config to jib-core API and Jib global config 

chanseokoh:
Adds the property `registryMirrors` to the global Jib config. Also see #3002 for the proposed public doc change.

```json
{
  "disableUpdateCheck": false,
  "registryMirrors": [
    {
      "registry": "registry-1.docker.io",
      "mirrors": ["mirror.gcr.io", "localhost:5000"]
    },
    {
      "registry": "quay.io",
      "mirrors": ["private-mirror.test.com"]
    }
  ]
}
```

Also adds `Containerizer.withRegistryMirrors(...)` to the jib-core API.


chanseokoh:
This one's ready for review, but #2996 should go in first.

chanseokoh:
This is ready for review. I'm curious what you feel about storing and returning a map and handling errors in this way.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildContext.java
```diff
@@ -498,6 +520,15 @@ public boolean getAlwaysCacheBaseImage() {
     return layerConfigurations;
   }
 
+  /**
+   * Gets the registry mirrors.
+   *
+   * @return the registry mirrors
+   */
+  public ImmutableMap<String, ImmutableList<String>> getRegistryMirrors() {
```
mpeddada1, line 528: How about using `ImmutableListMultimap` instead of  `ImmutableMap<String, ImmutableList<String>>` here?

chanseokoh: Nice, I used it in a few places, and looks better.

loosebazooka:
description or bug?

chanseokoh:
Sorry, updated the doc. Just realized I forget updating the jib-core `CHANGELOG`. Will update soon.

chanseokoh:
> I forget updating the jib-core `CHANGELOG`. ~Will update soon.~

Nah, I will update `CHANGELOG` once the registry mirror feature is fully working.

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -303,8 +309,25 @@ public Containerizer setAlwaysCacheBaseImage(boolean alwaysCacheBaseImage) {
...
+    return this;
+  }
+
   Set<String> getAdditionalTags() {
-    return additionalTags;
+    return ImmutableSet.copyOf(additionalTags);
+  }
+
+  ListMultimap<String, String> getRegistryMirrors() {
+    return ImmutableListMultimap.copyOf(registryMirrors);
```
mpeddada1, line 330:
```suggestion
    return registryMirrors.build();
```

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -303,8 +309,25 @@ public Containerizer setAlwaysCacheBaseImage(boolean alwaysCacheBaseImage) {
...
 
+  /**
+   * Sets mirrors for a base image registry. Jib will try its mirrors in the given order before
+   * finally trying the registry.
+   *
+   * @param registry base image registry for which mirrors are configured
+   * @param mirrors a list of mirrors, where each element is in the form of {@code host[:port]}
+   * @return this
+   */
+  public Containerizer withRegistryMirrors(String registry, List<String> mirrors) {
```
loosebazooka, line 320: Naming wise, isn't this more "addRegistryMirrors" rather than "withRegistryMirrors"?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
```diff
@@ -81,21 +83,24 @@ public void setTargetImage(String targetImage) {
...
     // Asserts required @Input parameters are not null.
     Preconditions.checkNotNull(jibExtension);
     TaskCommon.disableHttpLogging();
     TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider();
 
     GradleProjectProperties projectProperties =
         GradleProjectProperties.getForProject(getProject(), getLogger(), tempDirectoryProvider);
-    Future<Optional<String>> updateCheckFuture =
-        TaskCommon.newUpdateChecker(projectProperties, GlobalConfig.readConfig(), getLogger());
+    Future<Optional<String>> updateCheckFuture = Futures.immediateFuture(Optional.empty());
```
loosebazooka, line 99: Should this change be rebased in?

loosebazooka (commented):
Maybe I got here late, but generally I think we use the style of abstract/interface for method inputs and return types? So ListMultimap is the type we expose and ImmutableListMultimap is an impl detail?

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
```diff
@@ -136,6 +141,7 @@ public static Containerizer to(TarImage tarImage) {
   private String toolName = DEFAULT_TOOL_NAME;
   @Nullable private String toolVersion = DEFAULT_TOOL_VERSION;
   private boolean alwaysCacheBaseImage = false;
+  private ListMultimap<String, String> registryMirrors = ArrayListMultimap.create();
```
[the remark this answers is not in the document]
mpeddada1: I believe this suggestion doesn't apply anymore.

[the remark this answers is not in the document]
chanseokoh: It's possible to use `ImmutableListMultimap` but I realized it's awkward with our public API. `Containerize` is a mutable object that keeps updated whenever calling `withAdditionalTag()` or `withRegistryMirrors()`. For the same reason, we have
```java
private final Set<String> additionalTags = new HashSet<>();
```

jib-core/src/main/java/com/google/cloud/tools/jib/api/Containerizer.java
chanseokoh: Ah, makes sense. "add" was ambiguous originally when not using `MultiMap`, but I realized `ListMultiMap.putAll()` is additive than replacing.

jib-core/src/test/java/com/google/cloud/tools/jib/configuration/BuildContextTest.java
```diff
@@ -83,7 +84,7 @@ public void testBuilder() throws Exception {
     List<String> expectedEntrypoint = Arrays.asList("some", "entrypoint");
     List<String> expectedProgramArguments = Arrays.asList("arg1", "arg2");
     Map<String, String> expectedEnvironment = ImmutableMap.of("key", "value");
-    ImmutableSet<Port> expectedExposedPorts = ImmutableSet.of(Port.tcp(1000), Port.tcp(2000));
+    Set<Port> expectedExposedPorts = ImmutableSet.of(Port.tcp(1000), Port.tcp(2000));
```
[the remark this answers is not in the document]
chanseokoh: I think this is fine. Sometimes you do want to have `ArrayList<String> var = new ArrayList<>()`, but often you do `List<String> var = ...` for collections locally.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildImageTask.java
chanseokoh: I did this is a wrong way. Fixed.

loosebazooka (approved):

glasser:
With the Gradle plugin is there any way to specify mirrors in the actual code or do we have to do this global config file?

PR #3003, opened 2021-01-14
Add new release checklist to release process for plugins only

loosebazooka:
This will only create a release checklist for the plugin releases.


.github/workflows/prepare-release.yml
```diff
@@ -85,6 +87,21 @@ jobs:
             ### Major Changes
 
             See [CHANGELOG.md](https://github.com/GoogleContainerTools/jib/blob/master/jib-${{ github.event.inputs.project }}-plugin/CHANGELOG.md) for more details.
+      
+      - uses: JasonEtco/create-an-issue@v2
+        if: ${{ github.event.inputs.project == 'maven' || github.event.inputs.project == 'gradle' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PROJECT_NAME: jib-${{ github.event.inputs.project }}-plugin
```
chanseokoh, line 95: This seems unnecesary.

chanseokoh (approved):

.github/workflows/prepare-release.yml
loosebazooka: Ah right, an artifact from before. I'll remove

PR #3008, opened 2021-01-19
Preserve modification time after unzipping file

### LINKED ISSUE #2913: Unpacking jar in jib cli jar code resulting in modification of sourceModificationTime
Need to keep this in mind before handling #2879.
Unzipping a JAR on the command line (by calling `unzip jar`, for example) maintains the original timestamp, however, unzipping the JAR in our code results in the current timestamp. This occurs even when the JAR is unpacked into the the same directory as the one that `unzip jar` unpacks to. 

The `sourcemodificationTime` is needed to determine whether the source file has been changed or not, but if it is changing every time we unpack a file, caching will be of little use.

mpeddada1:
Closes #2913 
Note that this PR will set the modification times of the directories that are not entries of the zip or tar file to a constant value (epoch plus 1) instead of letting them have a modification time of the current time.


chanseokoh (commented):
LGTM. While we're at it, how about fixing `TarExtractor` too?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ZipUtilTest.java
```diff
@@ -66,6 +70,28 @@ public void testZipSlipVulnerability_unix() throws URISyntaxException {
...
+  @Test
+  public void testUnzip_modificationTimePreserved() throws URISyntaxException, IOException {
+    Path destination = tempFolder.getRoot().toPath();
+    Path archive =
+        Paths.get(Resources.getResource("plugins-common/test-archives/test.zip").toURI());
+    try (ZipFile zipFile = new ZipFile(archive.toString())) {
+      FileTime sourceModificationTime1 = zipFile.getEntry("file1.txt").getLastModifiedTime();
+      FileTime sourceModificationTime2 = zipFile.getEntry("my-zip/file2.txt").getLastModifiedTime();
+      FileTime sourceModificationTime3 =
+          zipFile.getEntry("my-zip/some/sub/folder/file3.txt").getLastModifiedTime();
```
loosebazooka, line 82: I think, in this case, we should just have these values as constants that we know. Manually look in the zip and grab the values, and use those here. That will help us determine if some other part of the machinery breaks (like getLastModifiedTime).

chanseokoh: `ZipFile.getEntry().getLastModifiedTime()` and `Files.getLastModifiedTime()` are standard JDK. We can trust them, so I think it's okay?

loosebazooka: Right, I don't actually expect those to break. But in a case that we have access to the constant values (as it is a premade zip in the repository?), I feel like it's a more robust test style?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java
```diff
@@ -63,6 +63,7 @@ public static void unzip(Path archive, Path destination) throws IOException {
             ByteStreams.copy(zipIn, out);
           }
         }
+        Files.setLastModifiedTime(entryPath, entry.getLastModifiedTime());
```
chanseokoh, line 66: Maybe we can even set the time when calling `createDirectories()` or `newOutputStream()` through `FileAttribute`? If that's possible, we can reduce two IO operations into one. Actually, it is necessary if `createDirectories()` creates multiple directories?

mpeddada1: I agree with you. Unfortunately, setting the `"basic:lastModifiedTime"` attribute when calling `createDirectories()` results in an [`UnsupportedOperationException`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html) which occurs when an attribute can't be set atomically while creating a directory so I'm not sure if it is possible.

chanseokoh: Hmm... maybe we need to to roll our own version of `Files.createDirectories()` that incrementally creates subdirectories when they don't exists and sets the file timestamp. Otherwise, we can still end up creating directories with the current timestamp. Thoughts?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -72,6 +74,18 @@ public static void extract(Path source, Path destination) throws IOException {
...
       }
+
+      // Preserve source modification timestamps of files. If the entry is a symbolic link then set
+      // it's modification time to that of the target file.
+      for (TarArchiveEntry entry : modificationTimeMap.keySet()) {
+        FileTime sourceModificationTime = modificationTimeMap.get(entry);
+        if (entry.isSymbolicLink()) {
+          Path targetPath = destination.resolve(entry.getName()).toRealPath();
+          Path targetRelativePath = destination.relativize(targetPath);
+          sourceModificationTime = modificationTimeMap.get(targetRelativePath);
```
chanseokoh, line 85: I don't have confidence if this computation guarantees reconstructing the original path. (Path computation is difficult and depends on many factors. Honestly I don't know what `toRealPath()` does honestly, its Javadoc also says "the precise definition of this method is implementation dependent.") I also don't really like creating a map structure for this. Can't we just set timestamps inside the loop above?

mpeddada1: > I don't have confidence if this computation guarantees reconstructing the original path. (Path computation is difficult and depends on many factors. Honestly I don't know what `toRealPath()` does honestly, its Javadoc also says "the precise definition of this method is implementation dependent.")

Ah thank you for catching this! I looked into this a bit more and I think `Files.readSymbolicLink` is what I need to get the target of the symbolic link.

> I also don't really like creating a map structure for this. Can't we just set timestamps inside the loop above?

Calling set timestamp inside the for loop for the symlink was initially resulting in a `FileNotFoundException`. This was because we were calling `setLastModifiedTime` *before* the target file was created. The call to `setLastModifiedTime` on a symlink worked fine if the target file was created before the symlink but not the other way around. This is why I had to place the code at the end of the method (after all the files has been created).

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -41,21 +44,20 @@
    */
   public static void extract(Path source, Path destination) throws IOException {
     String canonicalDestination = destination.toFile().getCanonicalPath();
-
     try (InputStream in = new BufferedInputStream(Files.newInputStream(source));
         TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(in)) {
-
+      Map<TarArchiveEntry, FileTime> modificationTimeMap = new HashMap();
```
chanseokoh, line 49: This gives me a warning.

```suggestion
      Map<TarArchiveEntry, FileTime> modificationTimeMap = new HashMap<>();
```

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
chanseokoh, line 85: `modificationTimeMap.get(targetRelativePath)` doesn't make sense. The key should be of type `TarArchiveEntry` but here it's `Path`. How can this even work?

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -79,4 +89,64 @@ public void testExtract_symlinks() throws URISyntaxException, IOException {
...
+  public void testExtract_symlinks_modificationTimePreserved()
+      throws URISyntaxException, IOException {
+    Path source = Paths.get(Resources.getResource("core/symlinks.tar").toURI());
+    Path destination = temporaryFolder.getRoot().toPath();
+
+    Map<String, FileTime> originalModificationTimeMap = createModificationTimeMap(source);
+
+    TarExtractor.extract(source, destination);
+
+    // Validate that the symlink's modification time is set to the target's.
```
chanseokoh, line 120: This seems weird. The timestamp of a symlink should be different from that of its target.

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -79,4 +89,64 @@ public void testExtract_symlinks() throws URISyntaxException, IOException {
...
+  }
+
+  private static Map<String, FileTime> createModificationTimeMap(Path source) throws IOException {
+    Map<String, FileTime> modificationTimeMap = new HashMap();
+    try (InputStream in = new BufferedInputStream(Files.newInputStream(source));
+        TarArchiveInputStream tarArchiveInputStream = new TarArchiveInputStream(in)) {
+      for (TarArchiveEntry entry = tarArchiveInputStream.getNextTarEntry();
+          entry != null;
+          entry = tarArchiveInputStream.getNextTarEntry()) {
+        modificationTimeMap.put(entry.getName(), FileTime.from(entry.getModTime().toInstant()));
```
chanseokoh, line 147: If `entry` is a symlink, which timestamp does `FileTime.from()` return: the symlink or its target?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
chanseokoh: > to get the target of the symbolic link.

We shouldn't be concerned about the target? Are you trying to modify the target? Also, the target may not actually exist in an archive (i.e., a broken symlink). This shouldn't cause a problem.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -72,6 +76,18 @@ public static void extract(Path source, Path destination) throws IOException {
           }
         }
       }
+      preserveModificationTimes(destination, entries);
```
chanseokoh, line 79: Can we move this out of the `try` block?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -72,6 +76,18 @@ public static void extract(Path source, Path destination) throws IOException {
...
+    }
+  }
+
+  private static void preserveModificationTimes(Path destination, List<TarArchiveEntry> entries)
+      throws IOException {
+    new DirectoryWalker(destination)
+        .filter(path -> Files.isDirectory(path))
+        .walk(path -> Files.setLastModifiedTime(path, FileTime.from(Instant.ofEpochSecond(1L))));
```
chanseokoh, line 87: As this is an unconventional behavior, I think we should apply this only when enabled. `TarExtractor` and `ZipUtil` are deemed general and isolated.

And maybe we can create an instance of `FileTime.fromMillis(1)` (I think it's equivalent) at the start of the method and avoid re-creating the instance everytime.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ZipUtilTest.java
```diff
@@ -34,6 +39,7 @@
 public class ZipUtilTest {
 
   @Rule public final TemporaryFolder tempFolder = new TemporaryFolder();
+  public final FixedClock clock = new FixedClock();
```
chanseokoh, line 42: I don't think this is used? And not sure if `FixedCLock` from `com.google.api.client.testing.http` is a public API.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ZipUtil.java
```diff
@@ -63,7 +69,19 @@ public static void unzip(Path archive, Path destination) throws IOException {
...
```
chanseokoh, line 72: This line should be removed, right?

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -79,4 +83,53 @@ public void testExtract_symlinks() throws URISyntaxException, IOException {
...
+        .isEqualTo(FileTime.from(Instant.parse("2019-08-01T16:13:09Z")));
+    assertThat(Files.getLastModifiedTime(destination.resolve("file B")))
+        .isEqualTo(FileTime.from(Instant.parse("2019-08-01T16:12:00Z")));
+    assertThat(Files.getLastModifiedTime(destination.resolve("folder")))
+        .isEqualTo(FileTime.from(Instant.parse("2019-08-01T16:12:33Z")));
+    assertThat(Files.getLastModifiedTime(destination.resolve("folder").resolve("nested folder")))
+        .isEqualTo(FileTime.from(Instant.parse("2019-08-01T16:13:30Z")));
+    assertThat(
+            Files.getLastModifiedTime(
+                destination.resolve("folder").resolve("nested folder").resolve("file C")))
```
chanseokoh, line 104: Just FYI, no need to change: in *tests*, we're tolerant in writing `destination.resolve("folder/nested folder/file C"`) if it can simplify code or improve readability.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/ZipUtilTest.java
mpeddada1: Ah yes sorry, cleaned this up

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
mpeddada1: That is a very good point. Done.

And a quick note as per our discussion offline: It looks like `DirectoryWalker.walk()` doesn't follow symbolic links. It calls `Files.walk(path)` which doesn't follow symbolic links (by default) unless `FileVisitOption.FOLLOW_LINKS` is passed as an argument.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -73,5 +89,34 @@ public static void extract(Path source, Path destination) throws IOException {
...
+   * @param entries list of entries in tar file
+   * @param enableReproducibleTimestamps whether or not reproducible timestamps should be used
+   * @throws IOException when I/O error occurs
+   */
+  private static void preserveModificationTimes(
+      Path destination, List<TarArchiveEntry> entries, boolean enableReproducibleTimestamps)
+      throws IOException {
+    if (!Files.exists(destination)) {
+      return;
+    }
```
chanseokoh, line 110: This can never be true. We already extracted files into `destination`.

And the check we talked about was only for when `enableReproducibleTimestamps` is true. That is, if it's enabled, we should proceed only when either 1) `destination` doesn't exist; or 2) `destination` exists but it's empty. (Or, I'm fine with the contract where `destination` must not exist when `enableReproducibleTimestamps` is true, if that's simpler.) In any case, this precondition should be documented in the Javadoc of `extract()`. That said, the check should be at the beginning of `extract()` so that it fails fast.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -73,5 +89,34 @@ public static void extract(Path source, Path destination) throws IOException {
...
+    if (enableReproducibleTimestamps) {
+      FileTime epochPlusOne = FileTime.fromMillis(1000L);
+      new DirectoryWalker(destination)
```
chanseokoh, line 113: How does it work with a symbolic link to a directory?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
chanseokoh: If you are still confused or there exists something unclear, don't hesitate to ask to clarify.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
chanseokoh: Ah, this is already answered, sorry.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -40,22 +44,42 @@
...
+   * @param destination the output directory
+   * @param enableReproducibleTimestamps whether or not reproducible timestamps should be used
+   * @throws IOException if extraction fails
+   * @throws IllegalStateException when reproducible timestamps are enabled but the target root used
+   *     for extracting the tar contents is not empty
+   */
+  public static void extract(Path source, Path destination, boolean enableReproducibleTimestamps)
+      throws IOException {
+    boolean isEmptyDestination =
+        Files.isDirectory(destination) && destination.toFile().list().length == 0;
```
chanseokoh, line 63: Let's do this computation only when necessary. That is, doing this inside the `if` statement. Perhaps
```java
if (enableReproducibleTimestamps && Files.isDirectory(destination) && destination.toFile().list.length() != 0) {
  throw new ...
}
```
(`isDirectory()` implies `exists()`.)

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -40,22 +44,42 @@
...
+    if (enableReproducibleTimestamps && Files.exists(destination) && !isEmptyDestination) {
+      throw new IllegalStateException(
+          "Reproducible timestamps can only be enabled when the target root is an empty directory.");
```
chanseokoh, line 66: the target root does not exist or is an empty directory

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -73,5 +97,31 @@ public static void extract(Path source, Path destination) throws IOException {
         }
       }
     }
+    preserveModificationTimes(destination, entries, enableReproducibleTimestamps);
+  }
+
+  /**
+   * Preserve modification time of files and directories in a tar file. If a directory is not an
+   * entry in the tar file and reproducible timestamps are enabled then it's modification timestamp
```
chanseokoh, line 105:
```suggestion
   * entry in the tar file and reproducible timestamps are enabled then its modification timestamp
```

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -73,5 +97,31 @@ public static void extract(Path source, Path destination) throws IOException {
...
+   * @param enableReproducibleTimestamps whether or not reproducible timestamps should be used
+   * @throws IOException when I/O error occurs
+   */
+  private static void preserveModificationTimes(
+      Path destination, List<TarArchiveEntry> entries, boolean enableReproducibleTimestamps)
+      throws IOException {
+    if (enableReproducibleTimestamps) {
+      FileTime epochPlusOne = FileTime.fromMillis(1000L);
+      new DirectoryWalker(destination)
+          .filter(path -> Files.isDirectory(path))
```
chanseokoh, line 119: nit
```suggestion
          .filter(Files::isDirectory)
```

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -79,4 +84,52 @@ public void testExtract_symlinks() throws URISyntaxException, IOException {
...
+
+  @Test
+  public void testExtract_reproducibleTimestampsEnabled_invalidDestination() throws IOException {
+    Path destination = temporaryFolder.getRoot().toPath();
+    Files.createDirectory(destination.resolve("directory"));
+
+    assertThrows(
+        IllegalStateException.class,
+        () -> TarExtractor.extract(Paths.get("ignore"), destination, true));
+  }
```
chanseokoh, line 134: Let's assert the exception message too, since anyone can throw the very common `IllegalStateException`.

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -79,4 +84,52 @@ public void testExtract_symlinks() throws URISyntaxException, IOException {
...
+    assertThat(Files.getLastModifiedTime(destination.resolve("level-1/level-2")))
+        .isEqualTo(FileTime.fromMillis(1000L));
+    assertThat(Files.getLastModifiedTime(destination.resolve("level-1/level-2/level-3")))
+        .isEqualTo(FileTime.fromMillis(1000L));
+    assertThat(Files.getLastModifiedTime(destination.resolve("level-1/level-2/level-3/file.txt")))
+        .isEqualTo(FileTime.from(Instant.parse("2021-01-29T21:10:02Z")));
+  }
+
+  @Test
+  public void testExtract_reproducibleTimestampsEnabled_invalidDestination() throws IOException {
```
chanseokoh, line 127:
```suggestion
  public void testExtract_reproducibleTimestampsEnabled_destinationNotEmpty() throws IOException {
```

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -79,4 +84,52 @@ public void testExtract_symlinks() throws URISyntaxException, IOException {
...
+    Path destination = temporaryFolder.getRoot().toPath();
+    Files.createDirectory(destination.resolve("directory"));
```
chanseokoh, line 129: FYI, `temporaryFolder.newFile()` should also work.

chanseokoh (commented):
LGTM pending some comments

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
mpeddada1: Ah this is much better, thanks

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
mpeddada1: Good point! Since it's a common exception that makes sense. I tend to err on the side of caution when it comes to asserting the full message of the exception since it makes the test brittle so I tested just the relevant part.

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
mpeddada1: nice

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarExtractor.java
```diff
@@ -73,5 +97,31 @@ public static void extract(Path source, Path destination) throws IOException {
...
+      Path destination, List<TarArchiveEntry> entries, boolean enableReproducibleTimestamps)
+      throws IOException {
+    if (enableReproducibleTimestamps) {
+      FileTime epochPlusOne = FileTime.fromMillis(1000L);
+      new DirectoryWalker(destination)
+          .filter(Files::isDirectory)
+          .walk(path -> Files.setLastModifiedTime(path, epochPlusOne));
+    }
+    for (TarArchiveEntry entry : entries) {
+      Files.setLastModifiedTime(
```
chanseokoh, line 123: Oh, shouldn't we check `entry` is not a symlink? I remember setting a timestamp throws an error if it's a symlink?

mpeddada1: Ah it used to throw an error because calling `Files.setModifiedTime` inside the for loop on the symlink would try to change the timestamp of the target file which didn't exist/wasn't created at the time of the call.

mpeddada1: But hmm now that I think about it, could there be a possibility of the symlink's timestamp being set for the target?

chanseokoh: Oh, so it doesn't throw an error but changes the target. Adding a check seems right.

I think it's worth mentioning in the Javadoc that the timestamp of symlinks will be current even if `enableReproducibleTimestamps` is true.

chanseokoh (approved):
LGTM. Just update the Javadoc about symlink timestamps.

PR #3009, opened 2021-01-21
Add option to specify entrypoint 

mpeddada1:
Adds `--entrypoint`. #2975 needs to be merged first.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
```diff
@@ -58,8 +58,22 @@ public static JibContainerBuilder toJibContainerBuilder(
             : Jib.from("gcr.io/distroless/java-debian10:11");
 
     List<FileEntriesLayer> layers = processor.createLayers();
-    List<String> entrypoint = processor.computeEntrypoint();
-    containerBuilder.setEntrypoint(entrypoint).setFileEntriesLayers(layers);
+    List<String> entrypoint =
```
loosebazooka, line 61: I think there might be some value in adding a warning when both entrypoint and jvmflags are set.

mpeddada1: We are currently logging a warning in `Jar#call()` but moving it to `JarFiles` seems like a good idea. It will make testing easier.

loosebazooka: oh, I missed that earlier, I feel like your original location is better.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -67,11 +78,80 @@
...
+      names = "--program-args",
+      paramLabel = "<program-arguments>",
+      split = ",",
+      description = "Program arguments for container entrypoint.")
+  private List<String> programArguments = Collections.emptyList();
+
+  @CommandLine.Option(
+      names = "--entrypoint",
+      paramLabel = "<entrypoint>",
+      split = "\\s+",
```
chanseokoh, line 150: Q: does picoCLI automatically takes care of escaping for the split characters? For example, for `--program-args`, is  there a way to include `,` in an argument?

mpeddada1: When using split regex with picocli I think the user still needs to take care of escaping. A backlash needs to be escaped with another backlash otherwise it won't compile. In the case of `--program-args`, there is a way to include `,` in an argument. For example `--program-args=\"Hello, World\", arg2` will result in `["Hello,World", arg2]`. Does this answer your question?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarFiles.java
mpeddada1: Ah I see.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
chanseokoh: Basically yes. As long as there is a way to pass one of the delimiters, we should be good.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -175,6 +183,9 @@ public Integer call() {
                 + jarFile);
         return 1;
       }
+      if (!entrypoint.isEmpty() && !jvmFlags.isEmpty()) {
+        logger.log(LogEvent.Level.WARN, "--jvm-flag is ignored when --entrypoint is specified");
```
chanseokoh, line 187:
```suggestion
        logger.log(LogEvent.Level.WARN, "--jvm-flags is ignored when --entrypoint is specified");
```

chanseokoh (approved):

PR #3010, opened 2021-01-21
Improve progress reporting in PullBaseImageStep

chanseokoh:
`PullBaseImageStep` has got complicated over time, and the progress reporting was broken. I am trying to improve the situation with this PR. It's still not perfect, but I think it's better.

- At function call boundaries, I think it's often better to pass a dispatcher factory. The callee will have the freedom to decide how to sub-divide the given progress node.
- I think it's more robust to do `try (... = factory.create("description", size)) { ... }` at the beginning of a function as early as possible.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -95,8 +95,8 @@ class PullBaseImageStep implements Callable<ImagesAndRegistryClient> {
           LayerCountMismatchException, BadContainerConfigurationFormatException,
           CacheCorruptedException, CredentialRetrievalException {
     EventHandlers eventHandlers = buildContext.getEventHandlers();
-    try (ProgressEventDispatcher progressEventDispatcher =
-            progressEventDispatcherFactory.create("pulling base image manifest", 2);
+    try (ProgressEventDispatcher progressDispatcher =
+            progressEventDispatcherFactory.create("pulling base image manifest", 4);
         TimerEventDispatcher ignored1 = new TimerEventDispatcher(eventHandlers, DESCRIPTION)) {
 
       // Skip this step if this is a scratch image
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -95,8 +95,8 @@ public ImagesAndRegistryClient call()
           LayerCountMismatchException, BadContainerConfigurationFormatException,
           CacheCorruptedException, CredentialRetrievalException {
     EventHandlers eventHandlers = buildContext.getEventHandlers();
-    try (ProgressEventDispatcher progressEventDispatcher =
-            progressEventDispatcherFactory.create("pulling base image manifest", 2);
+    try (ProgressEventDispatcher progressDispatcher =
+            progressDispatcherFactory.create("pulling base image manifest", 3);
```
chanseokoh, line 99: `progressDispatcher.newChildProducer()` (which decrements one progress unit) can be called up to 3 times inside this method.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -211,65 +215,80 @@ public ImagesAndRegistryClient call()
...
-
-    List<ManifestAndConfigTemplate> manifestsAndConfigs = new ArrayList<>();
-    ImmutableList.Builder<Image> images = ImmutableList.builder();
-    // If a manifest list, search for the manifests matching the given platforms.
-    for (Platform platform : buildContext.getContainerConfiguration().getPlatforms()) {
-      String message = "Searching for architecture=%s, os=%s in the base image manifest list";
+    try (ProgressEventDispatcher progressDispatcher1 =
+        progressDispatcherFactory.create("pulling base image manifest and container config", 2)) {
+      ManifestAndDigest<?> manifestAndDigest =
+          registryClient.pullManifest(baseImageConfig.getImageQualifier());
```
chanseokoh, line 228: The structure is now like this:

```
try ( dispatcher1 with 2 allocation units ) {
   // The first allocation unit is for pulling a manifest.
   dispatcher1.dispatchProgress(1) (decrements 1 unit and dispatch) after pullManifest()

   // The second unit is for either
   //   1) pulling one container config (case 2 below); or
   //   2) pulling (manifest x config config) of n platforms (case 3 below)
   childProgressFactory = dispatcher1.newChildProducer() (decrements 1 unit)

   if case 1: v21manifest --> just return
   if case 2: v22manifest --> childProgressFactory passed to pullContainerConfigJson()
   if case 3: manifest list
       try ( dispatcher2 with 2 * platforms.size() created from childProgressFactory) {
          for each platform {
            dispatcher2.dispatchProgress(1) after pullManifest()
            dispatcher2.newChildProducer() passed to pullContainerConfigJson()
          }
       }
}

File: jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java [hunk 3]
```diff
@@ -138,15 +138,16 @@ public ImagesAndRegistryClient call()
...
-      // some risk. https://github.com/GoogleContainerTools/jib/pull/2200#discussion_r359069026
-      // contains some related discussions.
-      RegistryClient noAuthRegistryClient =
-          buildContext.newBaseImageRegistryClientFactory().newRegistryClient();
       try {
+        // First, try with no credentials. This works with public GCR images (but not Docker Hub).
+        // TODO: investigate if we should just pass credentials up front. However, this involves
+        // some risk. https://github.com/GoogleContainerTools/jib/pull/2200#discussion_r359069026
+        // contains some related discussions.
+        RegistryClient noAuthRegistryClient =
```
loosebazooka, line 146: does this need to be in the try?

loosebazooka (approved):

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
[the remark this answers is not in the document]
chanseokoh: No, but I found this doesn't need to be outside the try.

PR #3011, opened 2021-01-21
Add registry mirror support in Jib build plugins

chanseokoh:
Supports only public mirrors such as `mirror.gcr.io`. (If a mirror requires credentials, it won't work.)

I went with the simple approach to try the mirrors when pulling a base image manifest in `PullBaseImageStep`. The class sets up a `RegistryClient` and returns it, so we just try pulling from mirrors and the actual registry until success. (So, this assumes that if a mirror returns a manifest with 200, all subsequent requests to pull blobs will succeed, which is not an unreasonable assumption.)

Tested with the following config:

```json
{
  "disableUpdateCheck":false,
  "registryMirrors": [
    {
      "registry": "registry-1.docker.io",
      "mirrors": ["localhost:5000", "mirror.gcr.io"]
    }
  ]
}
```

In my test, the first mirror `localhost:5000` always fails (connection error), but it doesn't fail the entire build. It tries `mirror.gcr.io` next. Then, if `mirror.gcr.io` has cached the base image, Jib nevers contact Docker Hub. If `mirror.gcr.io` returns 404, it correctly falls back to Docker Hub.


chanseokoh:
Added tests and ready for review. But #3010 needs to go first.

Will update CHANGELOG in the meantime.

jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildContext.java
```diff
@@ -532,7 +532,21 @@ public boolean getAlwaysCacheBaseImage() {
    * @return a new {@link RegistryClient.Factory}
    */
   public RegistryClient.Factory newBaseImageRegistryClientFactory() {
-    return newRegistryClientFactory(baseImageConfiguration);
+    return newBaseImageRegistryClientFactory(baseImageConfiguration.getImageRegistry());
```
loosebazooka, line 535: should this javadoc now indicate that the serverUrl is derived from the baseimage configuration?

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -409,4 +418,95 @@ public void testGetCachedBaseImages_v22ManifestListCached_onlyPlatforms()
...
+        pullBaseImageStep.tryMirrors(buildContext, progressDispatcherFactory);
+    Assert.assertTrue(result.isPresent());
+    Assert.assertEquals(gcrRegistryClientFactory.newRegistryClient(), result.get().registryClient);
+
+    Mockito.verify(eventHandlers)
+        .dispatch(LogEvent.info("trying mirror quay.io for the base image"));
+    Mockito.verify(eventHandlers)
+        .dispatch(LogEvent.info("trying mirror gcr.io for the base image"));
+    Mockito.verify(eventHandlers)
+        .dispatch(LogEvent.debug("failed to get manifest from mirror quay.io: not found"));
```
loosebazooka, line 488: if this is the expected log output ordering, I think it's a little confusing?

I would expect it to be
```
trying quay..
failed to get from quay...
trying gcr..
```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -62,8 +67,10 @@
 public class PullBaseImageStepTest {
```
loosebazooka, line 67: should there be a test for all registry mirrors failing and last resort is to use the original registry url?

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
chanseokoh: Used `Mockito.InOrder()` and rearranged the order.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -138,6 +140,13 @@ public ImagesAndRegistryClient call()
         }
       }
 
+      Optional<ImagesAndRegistryClient> mirrorPull =
+          tryMirrors(buildContext, progressDispatcher.newChildProducer());
+      if (mirrorPull.isPresent()) {
```
loosebazooka, line 145: could this message go in tryMirrors? with the name of the mirror that was successful?

loosebazooka (approved):

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
chanseokoh: Done. Also checking the message in one of the tests.

Issue #3018, opened 2021-01-28
[jib-cli] build command not finding buildfile


loosebazooka:
The jib cli is unable to locate the buildfile, even though jib.yaml is in the default context root. Only works if --buildfile is explicitly specified.

run
```
jib build --target docker://something
```
with files
```
jib.yaml
script.sh
```
Throws a null pointer exception

but works with 
```
jib build --target docker://something --buildfile ./jib.yaml
```

PR #3019, opened 2021-01-28
Use correctly processed options in build command

loosebazooka:
fixes https://github.com/GoogleContainerTools/jib/issues/3018

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
@@ -78,7 +78,8 @@ public class Build implements Callable<Integer> {
    *
    * @return a path to a buildfile
    */
-  public Path getBuildFile() {
+  @VisibleForTesting
+  Path getBuildFile() {
     if (buildFileUnprocessed == null) {
       return contextRoot.resolve("jib.yaml");
     }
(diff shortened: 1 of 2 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
```diff
@@ -71,21 +72,17 @@
...
-
   /**
    * Returns a user configured Path to a buildfile and if none is configured returns jib.yaml in
-   * {@link #getContextRoot()}.
+   * {@link #contextRoot}.
    *
-   * @return a path to a bulidfile
+   * @return a path to a buildfile
    */
   public Path getBuildFile() {
```
chanseokoh, line 81: Doesn't seem like it needs to be `public`. `@VisibleForTesting`?

chanseokoh (approved):
Interesting. It used to work, but I guess it broke at some point during refactoring.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
loosebazooka: oh right.

loosebazooka:
> Interesting. It used to work, but I guess it broke at some point during refactoring.

Yeah probably because I never added any good integration tests :\

PR #3022, opened 2021-01-29
Add Jib extensions info to build plugin docs

### LINKED ISSUE #2986: Support dependencies inclusion/exclusion into layer
This is a feature request. 

Sometimes it's needed to exclude certain dependencies from resulting build. Example - when AspectJ is used to weave certain libraries, it produces result in classes and jar of that dependency should not be included in the build. 

To have exclusions with maven-shade-plugin for example, we are using `excludes`, which are supported by that plugin. With jib same result is possible to receive by using `provided` scope, however that have unintended effect that all transitive dependencies become provided as well and we have to add those dependencies as well and change their scope to runtime. This requires adding sometimes quite a lot of dependencies, which increases pom.xml somewhat.

Suggestion is to support some way of selecting which dependencies should not go in the resulting build. Here's an example of excluding from maven-shade-plugin:
```

                    <artifactSet>
                        <excludes> <!-- Following artifacts are weaved by aspectj and are already in the fat jar -->
                            <exclude>com.typesafe.akka:akka-actor_${scala.tools.version}</exclude>
                            <exclude>io.kamon:kamon-akka-${akka.major.version}_${scala.tools.version}</exclude>
                            <exclude>org.aspectj:aspectjtools</exclude>
                        </excludes>
                    </artifactSet>
```

It's just an example how it could look like. Jib could have a different configuration. There could be inclusion rule as well. Maybe it would be even convenient to define list of layers, each with its own inclusions/exclusions, for situation when user wants to separate dependencies into different layers based on how often they're changed (f.e. if project has a library shared between services, which changes often and framework libs, which do not).

chanseokoh:
Closes #2986.


docs/faq.md
```diff
@@ -537,6 +539,15 @@ $ docker manifest inspect openjdk:8
 }
 ```
 
+### I want to exclude files from layers, have more fine-grained control over layers, change file ownership, etc.
+
+See ["Jib build plugins don't have the feature that I need"](#jib-build-plugins-dont-have-the-feature-that-i-need).
+
+### Jib build plugins don't have the feature that I need.
+
+The Jib plugins have an extension framework that enables anyone to easily extend and tailor the Jib behavior to their liking. We have already written a few useful extensions, for example, that enables [fine-grained layer control](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-layer-filter-extension-gradle) (such as including deleting files and moving files into new layers), and [supports Quarkus](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-quarkus-extension-gradle). You can also write your own extension. Check out the [jib-extensions](https://github.com/GoogleContainerTools/jib-extensions) repository.
```
loosebazooka, line 548:
```suggestion
The Jib build plugins have an extension framework that enables anyone to easily extend Jib's behavior to their needs. We maintain select [first-party](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party) plugins for popular use cases like [fine-grained layer control](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-layer-filter-extension-gradle) and [Quarkus support](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-quarkus-extension-gradle), but anyone can write and publish an extension. Check out the [jib-extensions](https://github.com/GoogleContainerTools/jib-extensions) repository for more information.
```

jib-gradle-plugin/README.md
```diff
@@ -469,6 +470,12 @@ jib {
 }
 ```
 
+
+### Jib Extensions
+
+The Jib plugins have an extension framework that enables anyone to easily extend and tailor the Jib behavior to their liking. We have already written a few useful extensions, for example, that enables [fine-grained layer control](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-layer-filter-extension-gradle) (such as including deleting files and moving files into new layers), and [supports Quarkus](https://github.com/GoogleContainerTools/jib-extensions/tree/master/first-party/jib-quarkus-extension-gradle). You can also write your own extension. Check out the [jib-extensions](https://github.com/GoogleContainerTools/jib-extensions) repository.
```
loosebazooka, line 476: Not sure fond of the idea of repeating what's in the faq here. Maybe we can redirect to it?

chanseokoh: The point of #2986 is that some features like fine-grained layer control are worth mentioning in the main plugin doc so that it can be searched and found easily. (The FAQ was an afterthought. I think the main entry for first time users are the main doc, not FAQ.) I still want to mention a few specific extensions, and then it basically becomes the entire FAQ description, which is actually fairly short.

loosebazooka: sounds good to me.

loosebazooka (approved):

Issue #1778, opened 2019-06-17
Make used Gradle SourceSet and Configuration configurable


remmeier:
The Jib Gradle plugin is hard-coded to make use of the "main" classpath and "runtime" configuration. It should be possible to specify those two values in the JibExtension to allow customizations:

- add further libraries to classpath for Docker only.
- remove libraries from the classpath as they are already available in a base image.
- change the source set because the projects main "purpose" is not being a docker image, but something else and have docker-related sources of the project in a non-main source set. We have that pattern quite frequently when we build plugins for other system that are then installed as initContainer. So the main artifact of the project would be the "plugin", whereas the initContainer is more a side-thing related to deployment. Currently we are forced to make setup two projects rather than just having two source sets.

Many/most other Gradle plugins allow to do the same thing to maintain flexiblity of what gets touched by plugin. And effort for allowing that is minimal.

The ticket is related to https://github.com/GoogleContainerTools/jib/issues/894, but as requested created a ticket on its own. In most cases the changes requested here should be simpler and suffice for developers to allow doing customizations. The other ticket then would allow even more customization at expensve of additional complexity.

loosebazooka:
> add further libraries to the classpath for Docker only
Can you explain this point a little more? It seems like you can either use the `runtime` scope or the `extraDirectories` option.

> remove libraries from the classpath as they are already available in a base image.
This can be handled by the `compileOnly` or `provided` mechanisms

> change the source set because the projects main "purpose" is not being a docker image, but something else and have docker-related sources of the project in a non-main source set. We have that pattern quite frequently when we build plugins for other system that are then installed as initContainer. So the main artifact of the project would be the "plugin", whereas the initContainer is more a side-thing related to deployment. Currently we are forced to make setup two projects rather than just having two source sets.

I'm a little confused about this point. In reality it sounds like two projects IS what you want. They are separate artifacts and that are somewhat related.

remmeier:
> I'm a little confused about this point. In reality it sounds like two projects IS what you want. 
> They are separate artifacts and that are somewhat related.

[...]

PR #3034, opened 2021-02-04
Configurable gradle configurations

cromefire:
Fixes parts of #1778 

This introduces a new configuration option `configurationName` that just sets the name of the gradle configuration that should be used.


cromefire:
In practice this looks like:
```kotlin
jib {
    configurationName.set("docker")
}
```

cromefire:
@googlebot I signed it!

cromefire:
This is basically working, but I'm not (yet) enough into the mocking stuff to get the tests fixed up again

cromefire:
Okay got all of them fixed.

chanseokoh:
Thanks for the PR! I think overall this looks good but haven't looked into all the details or done tests. This week we're focused on another high priority work, so please be patient.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -221,6 +227,17 @@ public String getContainerizingMode() {
...
+  @Optional
+  public Property<String> getConfigurationName() {
+    return configurationName;
+  }
+
+  public String readConfigurationName() {
+    String property = System.getProperty(PropertyNames.CONFIGURATION_NAME);
+    return property != null ? property : configurationName.get();
+  }
+
```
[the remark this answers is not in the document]
cromefire: Because the getter (and the get and the read function have a different semantic meaning) has a special meaning in kotlin, which is used for the gradle kotlin DSL (which is a official, fully supported option besides the old groovy DSL). For this property I changed it from the current model of bare getters and setters to expose the `Property` itself (This would be good for all options, but because it's not backwards compatible it seems like a good idea to only introduce it for new properties). This is a best practice, as it allows for lazily setting the value of the option for example (this also applies to the groovy DSL).

The getter for the property has to be named like a standard getter for kotlin to allow the conversation to a synthetic property like this:
```kotlin
jib {
    configurationName.set("value")
}
```
If the getter is name differently than the standard getter this happens:
```kotlin
jib {
    alternativeGetter().set("value")
}
```
which is probably not desirable.

One alternative that might be possible is the usage of a `public final` property, I will investigate that and report back on both solutions' compatibility with both the groovy and the kotlin DSL, but in either case the getter seems not a good idea in that case.

One additional option that's independent from the two above is to replace all instances of the read function with a `getConfigurationName().get()` but that is missing the default value default (which might not be needed though, I'll investigate, whether it can ever return `null`).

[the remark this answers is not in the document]
cromefire: One note about the exposing of `Property`: The recent API additions of feels also seem to heavily favor this style ([look here](https://docs.gradle.org/current/userguide/toolchains.html#sec:consuming) and switch to the kotlin DSL and notice the `languageVersion.set()`)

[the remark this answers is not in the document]
mpeddada1: I appreciate your detailed explanation. I can now see why you are trying to expose the Property using the getter (since that seems to be the convention for lazy configuration). The part that is a little confusing is the call to `readConfigurationName()` whenever `configurationName` is referenced. I'm leaning more towards your third option, where we use `getConfigurationName().get()` instead of `readConfigurationName()`  which also aligns with the examples in https://docs.gradle.org/current/userguide/implementing_gradle_plugins.html. But yeah, we need to double-check if it is possible for it to return `null`.

chanseokoh:
Should we also update `JibPlugin`?

https://github.com/GoogleContainerTools/jib/blob/532a86bd55b576cc30916be3e48979f110e5402d/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java#L211-L217

cromefire:
Yes, I think so, will have a look

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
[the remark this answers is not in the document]
chanseokoh: It'd be really nice if we could expose `Property<String>` in general, but there's a catch: I believe the reason that we factor in reading system properties in the getters marked with `@Input` is to correctly make Gradle check UP-TO-DATE. That is, Jib doesn't only read input values from `build.gradle` but also from other external source. For example, let's say you ran Jib once and a subsequent Gradle build will skip running the Jib task if there's no change to `build.gradle`. However, if you change some option on the command-line (e.g., with `-Djib.configurationName=...`), we want Gradle to re-run Jib even if nothing has changed in `build.gradle`.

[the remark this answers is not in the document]
mpeddada1: [To correct my comment above] After discussing this a bit and given @chanseokoh's comment, the last point in my comment about checking for the `null` value is not really valid.
But as mentioned above, due to the uniqueness of our situation and how system properties take precedence over values from `build.gradle` in jib, it still seems to be the case that we need to have this logic about system properties in the getter.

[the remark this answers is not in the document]
cromefire: Yeah right I misread the code, it's not a null-check it's a system.property override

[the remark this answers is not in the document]
cromefire: So the `@Input` should be moved down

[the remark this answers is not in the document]
chanseokoh: > So the `@Input` should be moved down

I don't think you can.
```
   > Warning: Type 'JibExtension': non-property method 'readConfigurationName()' should not be annotated with: @Input, @Optional.
```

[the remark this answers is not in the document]
cromefire: The problem I see otherwise is that changes of the property won't be reflected by the up-to-date status. Maybe we should stop the system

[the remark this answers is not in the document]
chanseokoh: It should have been nice if we were able to add `@Input` to `readConfigurationName()`, but looks like `java-gradle-plugin` doesn't allow it probably because of the weird naming. (OTOH, certainly we can add `@Input` to `getConfigurationName()`.) And in order for Gradle to properly work with the UP-TO-DATE check in the present of our support for system properties, I think the best course of action is to keep `@Input` on `getConfigurationName()` while the method should factor in the system property. I did confirm that the UP-TO-DATE status is correctly reflected based on the return value of a getter marked with `@Input`. We can still hold and accept a value as `Property`.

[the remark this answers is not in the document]
cromefire: That does break compatibility with the kotlin DSL though and that's the dilemma, so it seems like it's either proper, property based configuration, with maybe defaults based on a system property or a system property override.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -221,6 +227,17 @@ public String getContainerizingMode() {
     return property != null ? property : containerizingMode.get();
   }
 
+  @Input
+  @Optional
+  public Property<String> getConfigurationName() {
+    return configurationName;
```
loosebazooka, line 233: our of curiosity, can we just set the property to the system property value here?
```
public Property<String> getConfigurationName() {
  String property = System.getProperty(PropertyNames.CONFIGURATION_NAME);
  if (property != null) {
    configurationName.set(property);
  }
  return configurationName;
}
```

while this still looks a little odd, if it is possible, it matches more closely our previous patterns and it removes the ambiguous `GetconfigurationName/readConfigurationName` situation.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -118,6 +120,15 @@ public JibExtension(Project project) {
...
     containerizingMode = objectFactory.property(String.class);
+    configurationName =
+        objectFactory
+            .property(String.class)
+            .convention(
+                project.provider(
+                    () ->
+                        System.getProperty(
+                            PropertyNames.CONFIGURATION_NAME,
+                            JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME)));
```
chanseokoh, line 131: I don't think this works. The system property should take precedence over the value set in `build.gradle`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
cromefire: I've realized it via a default value now, that sounds like a good idea anyway because it gives the user all the control. So if you want the override behavior you can just set the value to `System.getProperty("some.property", "default")`. The downside is of course that this is different from the rest of the API, creating an inconsistency in some cases.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
cromefire: See my reasoning above, the only workaround possible is probably the one above, but that seems really unidiomatic. I can implement that if you insist, but I'd prefer not to. On the other hand so is a lot of the plugin currently, so that might be something worth considering for a major release of the plugin.

cromefire:
> Should we also update `JibPlugin`?
> 
> https://github.com/GoogleContainerTools/jib/blob/532a86bd55b576cc30916be3e48979f110e5402d/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java#L211-L217

That breaks the build (seems to exclude the own code) and it works without it, so I guess not.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
chanseokoh: We try to think from the user perspective, and I think what is more important than trying to have the code idiomatic is that
1. The user should be able to conveniently override Jib configurations on the command-line (i.e., via system properties).
2. Jib should not build an incorrect image. (When some Jib config was changed on the command-line, Jib should rebuild a new image with the new config.)
I am fine with whatever best and simplest way to achieve these. But what's really important is that these should be satisfied.

cromefire: I'd be in favor of a more standard gradle approach (this is actually the first plugin where I've seen predefined overrides via system properties) that gives to use the ability to opt-in into overrides using his/her technology of choice.

As the case with all gradle options a user may opt in any case to use system properties or environment variables:
```kotlin
jib {
    configurationName.set(System.property("jib.configurationName", "someCustomDefault"))
    // or for example
    configurationName.set(System.getenv("CONFIGURATION_NAME"))
    // or even
    configurationName.set(if (file("abc.txt").exists()) "option1" else "option2")
}
```
This approach is different from what maven is doing I think, but the idiomatic pattern with gradle. Your second point is satisfied right away and the user can opt into the first point by choosing the technology used.

I can see though, why you would't want to change this in a minor new feature (and create inconsistencies in the API).

chanseokoh: > The user should be able to conveniently override Jib configurations on the command-line

I do understand what you've been saying, but please allow me to re-iterate: being able to conveniently override or set values *on the command-line* without ever touching `build.gradle` or even any file on the target repo is very important. For example, it's even possible, that solely on the command-line, you can run Jib on any Gradle project solely where Jib isn't configured. For example, [this Tekton Jib pipeline task](https://github.com/tektoncd/catalog/blob/8dd769f8e8c6e0f359ee708d5c7f7cb24c885fee/task/jib-gradle/0.2/jib-gradle.yaml#L49-L70) can take any Java source repo as input and containerize the app. The user doesn't even have to know about Jib. 

And there are many users setting system properties only on the command-line (we've seen many cases particularly on CI/CD env), and supporting this is really critical for Jib. (Some examples [here](https://stackoverflow.com/a/64967072/1701388), [here](https://stackoverflow.com/a/61352644/1701388), [here](https://stackoverflow.com/q/59254492/1701388),[here]( https://stackoverflow.com/a/62127210/1701388), and [here]( https://stackoverflow.com/q/57850321/1701388).) And we've learned that, convenience and usability matters a lot for end users.

cromefire: That would still be possible via defaults.

Also all your samples refer to maven and tekton already writes gradle configuration anyway (as said gradle has a different model, so if you need to apply the plugin, you'll have to modify the config anyway, as is the case with any other built-in functionally with gradle.

chanseokoh: Often, the person building an image is not the person developing an app, and the former person doesn't have control over an app being containerized. (BTW, I don't agree that "Tekton already writes gradle configuration." It's just creates a temp file `init-script.gradle` somewhere (happens to be under `$HOME` but could have been under `/tmp/init-script.gradle`); it doesn't touch the source repo at all.) But even so, Tekton is just one example. For example, Cloud Run Button is a tool that you can add to any public GitHub repos, which then will build the source repo. It prefers using Jib if Jib is already configured on the repo. It first [checks if the Jib plugin is applied](https://github.com/GoogleCloudPlatform/cloud-run-button/blob/ff276c6f2ec7ff1b80710426bdc3d8eaafe3e2d6/cmd/cloudshell_open/jib.go#L45-L46) on the source repo by actually trying to run some Jib command and checking the result, and if the Jib plugin is found on the source repo, it [runs Jib while overriding values](https://github.com/GoogleCloudPlatform/cloud-run-button/blob/ff276c6f2ec7ff1b80710426bdc3d8eaafe3e2d6/cmd/cloudshell_open/jib.go#L27-L28) configured in source repo. Cloud Run Button doesn't yet support Gradle, but Gradle part would work in the exact same manner. If you didn't think Tekton doesn't touch the source repo (which I don't really agree), I am certain you'll agree that this Cloud Run Button for Gradle (if existed) would truly not modify the source repo.

And while these SO examples are not gradle, as I said, from the perspective of a person building an image or working on the CI/CD side or the build platform won't see the difference (as in the Cloud Run Button example). For them, they should just be able to run Jib (whether using Gradle or Maven or anything else; convention doesn't matter) while conclusively overriding parameters already configured in the project. That is, even if the project were Gradle and not Maven, some of the users on the SO (which may not be the dev people but ops) would use the same way to override Jib config values.

> That would still be possible via defaults.

Not sure if I follow this. If you are talking about your last commit change where you set the property convention to a system property value, I believe that doesn't work as we intend. For example, often people define `jib.to.image` in `build.gradle`, which will simply override the default value coming from the system property. But then, the person building an image may want to change the image push destination without touching `builld.gradle` by giving `-Djib.to.image` on the command-line. I see that, with your last commit, the system property will be used if the corresponding value isn't set in `build.gradle`, but you can't override the value in `build.gradle` with `-Djib.to.image` on the command-line, right?

So, with your last commit change, I don't really see an actual migration path for the Tekton or Cloud Run Button examples if Jib drops the current support for system properties.

cromefire: > BTW, I don't agree that "Tekton already writes gradle configuration." It just creates one temp file init-script.gradle somewhere (happens to be under $HOME but could have been under /tmp/init-script.gradle

Well `init-script.gradle` _is_ gradle configuration. Something like that would also be needed for cloud run. This would simply result in something like this (if you want to be sure that you want to override it):
```bash
echo "initscript {
  repositories { maven { url 'https://plugins.gradle.org/m2' } }
  dependencies { classpath 'gradle.plugin.com.google.cloud.tools:jib-gradle-plugin:+' }
}
rootProject {
  afterEvaluate {
    if (!project.plugins.hasPlugin('com.google.cloud.tools.jib')) {
      project.apply plugin: com.google.cloud.tools.jib.gradle.JibPlugin
    }
    jib {
      configurationName = "something"
    }
  }
}" > $HOME/init-script.gradle
```

> I see that, with your last commit, the system property would be used only if the corresponding value isn't set in build.gradle. That is, you can't override the value in build.gradle with -Djib.to.image on the command-line, right?

That is correct and as you really want that specifically, I'll just go an implement it. I'm not sure that discussing this further would bring up any new results.

> So, with your last commit change, I don't really see an actual migration path for the Tekton or Cloud Run Button examples if Jib drops the current support for system properties that can override Jib values set in a project.

See above

chanseokoh: > See above

Sure, technically it's does not look impossible, but again, it is very important for the user to be able to *conveniently* override Jib config values on the command-line. You will agree that the possible workaround is too cumbersome and brittle. That is simply not acceptable in practice, and I am sure that we will face with a huge backlash. I can bring up more examples (such as Skaffold and Cloud Code for IntelliJ), but seems like I don't need to. It is important for us to keep the current support to be successful.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -298,16 +311,18 @@ public JibContainerBuilder createJibContainerBuilder(
...
-    runtimeClasspath.forEach(
-        file -> {
-          if (file.exists()
-              && file.isFile()
-              && file.getName().toLowerCase(Locale.US).endsWith(".jar")) {
-            dependencies.add(file.toPath());
-          }
-        });
+    runtimeClasspath
+        .getAsFileTree()
```
chanseokoh, line 317: `getAsFileTree()`: what does this do and why do we want to add this? Can you guarantee this does not change the current behavior? The order of the elements in the return value is important (that's why we have the code comment right above), but the [Javadoc](https://docs.gradle.org/current/javadoc/org/gradle/api/file/FileTree.html) of `FileTree` says

> The order of the files in a `FileTree` is not stable, even on a single computer.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
chanseokoh, line 131: Need to revert this change: 
```suggestion
    configurationName =
        objectFactory
            .property(String.class)
            .convention(JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME);
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
```diff
@@ -215,6 +215,10 @@ public void apply(Project project) {
                   .getSourceSets()
                   .getByName(SourceSet.MAIN_SOURCE_SET_NAME);
           jibDependencies.add(mainSourceSet.getRuntimeClasspath());
+          jibDependencies.add(
+              projectAfterEvaluation
```
chanseokoh, line 219: Quick question, as I don't have a lot of Gradle expertise: would it be better if line 212 used `projectAfterEvaluation` instead of `project` (e.g., at least for consistency and eliminating future questions like this)?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
cromefire: Well then we have to find something better for mocking, because implementation of configuration ist changing quite often

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
cromefire: From a technical perspective ist irrelevant, the `Project` instance changes while evaluation

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
cromefire: And it's huge, I'm not into mocking a lot so I don't know how this would be done.

chanseokoh: Hmm... thanks for trying. I'll see what I can do. This can take some time though.

chanseokoh: Almost there. Only some minor last things blocking me.

chanseokoh: Heads-up: #3091 I may need to add or remove one or two, but it's working.

chanseokoh: #3091 is merged, Sync with HEAD, and you should be able to fix the tests.

cromefire: Integrated

cromefire:
Not sure what's up with the code coverage

chanseokoh (approved):
LGTM. Thanks for your contribution! I think we can merge this when ready.

cromefire:
Will also submit a quick followup, to get the gradle version up to date, it's basically a 2 line change now with the new testing

chanseokoh:
Merging now. Thanks again for your contribution!

@mpeddada1 we need to add the config param to the plugin doc and also update CHANGELOG.

cromefire:
Maybe it's because of your usage of `@Inject`?

chanseokoh:
I think `@Inject` is necessary, and I am not sure if the constructor is related to the error.
```
> Could not create task ':myTask'.
   > Could not create task of type 'SimpleTask'.
      > The constructor for type SimpleTask should be annotated with @Inject.
```
The error is due to calling `contents.set(property)` when the `contents` is already [finalized](https://docs.gradle.org/current/userguide/lazy_configuration.html#unmodifiable_property). Maybe the code like this is not really safe? :disappointed:

cromefire:
Maybe one should check for `property != null && !property.equals(contents.get())`, that should avoid that problem

cromefire:
All in all it is a work around and not really how it was intended to be used so yeah that's the tradeoff...

chanseokoh:
> Maybe one should check for `property != null && !property.equals(contents.get())`, that should avoid that problem

Oh, that did the trick! I now see it was applying the system property multiple times. Thanks! I should update the code.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java
```diff
@@ -221,6 +227,22 @@ public String getContainerizingMode() {
...
+   * Returns the configurationName property while setting it to the value of the system property if
+   * present.
+   *
+   * @return The configurationName property
+   */
+  @Input
+  @Optional
+  public Property<String> getConfigurationName() {
+    String property = System.getProperty(PropertyNames.CONFIGURATION_NAME);
+    if (property != null) {
```
chanseokoh, line 240: FTR, this can be problematic and should be fixed. https://github.com/GoogleContainerTools/jib/pull/3034#issuecomment-792844429

PR #3035, opened 2021-02-05
Modify jib cli jar command to preserve source file path

### LINKED ISSUE #2879: [jib-cli-jar] Enable caching for jib cli jar
This is would require creating a build artifacts directory for the command so that the jar file wouldn't need to be unzipped every time it is used.

mpeddada1:
Modifies the jar command to use a fixed directory `exploded-jar` which is located under `<user-cache>/jib-cli-cache/projects/<hash>` (where `<user-cache>` is normally `/tmp` on Linux if undefined) when unzipping. This is to preserve the sourceFile layer entry which is used to calculate the "selector" for cached layer tarballs. 

Closes #2879


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/SpringBootExplodedProcessor.java
```diff
@@ -41,20 +41,20 @@
 public class SpringBootExplodedProcessor implements JarProcessor {
 
   @Nullable private final Path jarPath;
-  @Nullable private final Path tempDirectoryPath;
+  @Nullable private final Path targetDirectoryPath;
 
-  public SpringBootExplodedProcessor(Path jarPath, Path tempDirectoryPath) {
+  public SpringBootExplodedProcessor(Path jarPath, Path targetDirectoryPath) {
```
chanseokoh, line 46: It's difficult to understand what `targetDirectoryPath` means. `targetExplodedJarRoot`? And maybe we can remove `localExplodedJarRoot` below.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/SpringBootExplodedProcessor.java
```diff
@@ -41,20 +41,20 @@
...
     this.jarPath = jarPath;
-    this.tempDirectoryPath = tempDirectoryPath;
+    this.targetDirectoryPath = targetDirectoryPath;
   }
 
   @Override
   public List<FileEntriesLayer> createLayers() throws IOException {
-    if (tempDirectoryPath == null || jarPath == null) {
+    if (targetDirectoryPath == null || jarPath == null) {
```
chanseokoh, line 53: `targetDirectoryPath` can never be `null`. Do you remember why we had this null check?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/SpringBootExplodedProcessor.java
```diff
@@ -41,20 +41,20 @@
 public class SpringBootExplodedProcessor implements JarProcessor {
 
   @Nullable private final Path jarPath;
-  @Nullable private final Path tempDirectoryPath;
+  @Nullable private final Path targetDirectoryPath;
```
chanseokoh, line 44: Remove `@Nullable` for `jarPath` and `targetDirectoryPath`, and double-check that indeed they cannot be null. Do you remember why they started with `@Nullable`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorsTest.java
```diff
@@ -39,41 +43,53 @@
...
     Path jarPath = Paths.get(Resources.getResource(STANDARD).toURI());
+    Path destination = temporaryFolder.getRoot().toPath();
+    when(mockCacheDirectories.getExplodedJarCache()).thenReturn(destination);
+
     JarProcessor processor =
-        JarProcessors.from(jarPath, mockTemporaryDirectoryProvider, ProcessingMode.exploded);
-    Mockito.verify(mockTemporaryDirectoryProvider).newDirectory();
+        JarProcessors.from(jarPath, mockCacheDirectories, ProcessingMode.exploded);
+
+    verify(mockCacheDirectories).getExplodedJarCache();
```
chanseokoh, line 59: Maybe we can check if `Files.exists(destination.resolve("some-file"))`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorsTest.java
```diff
@@ -39,41 +43,53 @@
...
 
   @Test
   public void testFrom_standardPackaged() throws IOException, URISyntaxException {
     Path jarPath = Paths.get(Resources.getResource(STANDARD).toURI());
     JarProcessor processor =
-        JarProcessors.from(jarPath, mockTemporaryDirectoryProvider, ProcessingMode.packaged);
-    Mockito.verifyNoInteractions(mockTemporaryDirectoryProvider);
+        JarProcessors.from(jarPath, mockCacheDirectories, ProcessingMode.packaged);
+
+    verifyNoInteractions(mockCacheDirectories);
```
chanseokoh, line 69: Perhaps verify the exploded jar directory is empty.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorsTest.java
```diff
@@ -39,41 +43,53 @@
...
 
-  @Mock private TempDirectoryProvider mockTemporaryDirectoryProvider;
+  @Mock private CacheDirectories mockCacheDirectories;
+
+  @Rule public final TemporaryFolder temporaryFolder = new TemporaryFolder();
 
   @Test
   public void testFrom_standardExploded() throws IOException, URISyntaxException {
     Path jarPath = Paths.get(Resources.getResource(STANDARD).toURI());
+    Path destination = temporaryFolder.getRoot().toPath();
```
chanseokoh, line 53: Path explodedJarRoot ?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/SpringBootExplodedProcessor.java
mpeddada1: This is a great catch, thank you! Yes `targetDirectoryPath` and `jarPath` cannot be null.`@Nullable` was added when I was refactoring `"JarModeProcessor"` into `StandardExplodedProcessor`, `SpringBootExplodedProcessor`, etc. I was initially using setters to set the `tempDirectoryPath` and `jarPath` since not all variations of `createLayers` and `computeEntrypoint` use **both** these parameters. The design changed to allow these values be passed in at a constructor level but I forgot to remove the `@Nullable`.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/jar/JarProcessorsTest.java
mpeddada1: Hm I'm not sure if this is possible in this unit test?`JarProcessors` is used to handle all the logic behind creating a `JarProcessor` instance. It doesn't do the unzipping of the jar or creation of layers.

chanseokoh: Ah, I see.

chanseokoh (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -59,11 +61,22 @@ public static JarProcessor from(
...
+      return new SpringBootExplodedProcessor(jarPath, explodedJarRoot);
     } else if (jarType.equals(STANDARD) && mode.equals(ProcessingMode.packaged)) {
       return new StandardPackagedProcessor(jarPath);
     } else {
-      return new StandardExplodedProcessor(jarPath, temporaryDirectoryProvider.newDirectory());
+      Path explodedJarRoot = cacheDirectories.getExplodedJarDirectory();
+      // Clear the exploded-jar directory first
+      if (Files.exists(explodedJarRoot)) {
+        MoreFiles.deleteRecursively(explodedJarRoot, RecursiveDeleteOption.ALLOW_INSECURE);
+      }
```
chanseokoh, line 78: I wasn't sure who should empty the directory, e.g., if `StandardExplodedProcessor` should clear it instead of `JarProcessors`. I think either makes sense, but if we want the caller to clear it, we should check the precondition inside `StandardExplodedProcessor` to verify that the directory does not exist or is empty.

Otherwise, LGTM, but I'll take a closer look at Monday again.

mpeddada1: Thanks for pointing this out! I moved the code block to `SpringBootExplodedProcessor` and `StandardExplodedProcessor`  to keep the logic in `JarProcessors` simple.

chanseokoh (approved):

chanseokoh:
Can you also manually verify that consecutively running `jib jar` on some fixed JAR path builds an image with the same digest?

mpeddada1:
@chanseokoh I have confirmed that consecutively running `jib jar` on the same jar creates an image with the same digest.

PR #3050, opened 2021-02-11
[jib-cli-jar] Add option to specify creation time of container

mpeddada1:
Required a bit of refactoring to use the `Instants` class which was initially implemented for the build command.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
@@ -279,8 +279,9 @@ public class Jar implements Callable<Integer> {
    * @return an optional creation time
    */
   public Optional<Instant> getCreationTime() {
-    return (creationTime == null)
-        ? Optional.empty()
-        : Optional.of(Instants.fromMillisOrIso8601(creationTime, "creationTime"));
+    if (creationTime != null) {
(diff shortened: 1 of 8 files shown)


chanseokoh (commented):
Can we add a test that checks the default time of the built image (or something to that effect) when the option was not given?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -150,6 +151,14 @@
           "Entrypoint for container. Overrides the default entrypoint, example: --entrypoint='custom entrypoint'")
   private List<String> entrypoint = Collections.emptyList();
 
+  @CommandLine.Option(
+      names = "--creation-time",
+      paramLabel = "<creation-time>",
+      description =
```
chanseokoh, line 157: Another option is to provide the default (e.g., 0?). That way, maybe `creationTime` can never be null?

Either way has pros and cons though.

mpeddada1: I did think about adding a default but I assumed this would be a little redundant since we already have a default value for this field in jib?

chanseokoh: Either is fine. At least we can document the default value in the option description.

chanseokoh (approved):

PR #3075, opened 2021-02-24
Use AdoptOpenJDK image as default base image in Jib CLI Jar

mpeddada1:

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
@@ -198,9 +198,8 @@ public class Jar implements Callable<Integer> {
       CacheDirectories cacheDirectories =
           CacheDirectories.from(commonCliOptions, jarFile.toAbsolutePath().getParent());
       JarProcessor processor = JarProcessors.from(jarFile, cacheDirectories, mode);
-      Integer jarJavaVersion = JarProcessors.getJavaMajorVersion(jarFile);
       JibContainerBuilder containerBuilder =
-          JarFiles.toJibContainerBuilder(processor, this, commonCliOptions, logger, jarJavaVersion);
+          JarFiles.toJibContainerBuilder(processor, this, commonCliOptions, logger);
       Containerizer containerizer = Containerizers.from(commonCliOptions, logger, cacheDirectories);
       containerBuilder.containerize(containerizer);
     } catch (Exception ex) {
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
@@ -198,8 +198,9 @@ public class Jar implements Callable<Integer> {
       CacheDirectories cacheDirectories =
           CacheDirectories.from(commonCliOptions, jarFile.toAbsolutePath().getParent());
       JarProcessor processor = JarProcessors.from(jarFile, cacheDirectories, mode);
+      Integer jarJavaVersion = JarProcessors.getJavaMajorVersion(jarFile);
       JibContainerBuilder containerBuilder =
-          JarFiles.toJibContainerBuilder(processor, this, commonCliOptions, logger);
+          JarFiles.toJibContainerBuilder(processor, this, commonCliOptions, logger, jarJavaVersion);
       Containerizer containerizer = Containerizers.from(commonCliOptions, logger, cacheDirectories);
       containerBuilder.containerize(containerizer);
     } catch (Exception ex) {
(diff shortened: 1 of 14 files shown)


chanseokoh (commented):
The plan is to check the Java version and choose between Java 8 and 11, right?

chanseokoh (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -90,8 +89,7 @@ private static String determineJarType(Path jarPath) throws IOException {
    * @return java version
    * @throws IOException if I/O exception thrown when opening the jar file
    */
-  @VisibleForTesting
-  static Integer getJavaMajorVersion(Path jarPath) throws IOException {
+  public static Integer getJavaMajorVersion(Path jarPath) throws IOException {
     try (JarFile jarFile = new JarFile(jarPath.toFile())) {
       Enumeration<JarEntry> jarEntries = jarFile.entries();
       while (jarEntries.hasMoreElements()) {
```
chanseokoh, line 95: Unrelated, we need to close the loader below.
```java
          try (URLClassLoader loader = new URLClassLoader(new URL[] {jarPath.toUri().toURL()});
              DataInputStream classFile =
                  new DataInputStream(loader.getResourceAsStream(jarEntry))) {
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -198,8 +198,9 @@ public Integer call() {
       CacheDirectories cacheDirectories =
           CacheDirectories.from(commonCliOptions, jarFile.toAbsolutePath().getParent());
       JarProcessor processor = JarProcessors.from(jarFile, cacheDirectories, mode);
+      Integer jarJavaVersion = JarProcessors.getJavaMajorVersion(jarFile);
```
chanseokoh, line 201: Instead of making `getJavaMajorVersion()` public and calling it twice (although I have no idea if there will be noticeable performance impact when "opening a jar" twice with `new JarFile()` in this case), how about adding a field `JarProcessor.javaVersion` and a getter for it? (Maybe the current `getJavaMajorVersion()` should be renamed to, say, `{read|determine}JavaMajorVersion` if we do so.) That way, we can avoid running the code twice. I see `JarProcessor` is already passed into `toJibContainerBuilder()`, so it just needs to call the getter.

mpeddada1: Ah I missed this. Thanks for pointing this out - adding a getter to `JarProcessor` is a good idea. Since interfaces don't allow instance fields, I had to add a parameter to the constructors of the `JarProcessor` instances which resulted in a couple of extra changes to the test.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/SpringBootPackagedProcessor.java
```diff
@@ -25,9 +25,17 @@
 public class SpringBootPackagedProcessor implements JarProcessor {
```
chanseokoh, line 25: Looks like the access can be limited to the package.

```suggestion
class SpringBootPackagedProcessor implements JarProcessor {
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/SpringBootPackagedProcessor.java
```diff
@@ -25,9 +25,17 @@
...
+  private final Integer jarJavaVersion;
 
-  public SpringBootPackagedProcessor(Path jarPath) {
+  /**
+   * Constructor for {@link SpringBootPackagedProcessor}.
+   *
+   * @param jarPath path to jar file
+   * @param jarJavaVersion jar java version
+   */
+  public SpringBootPackagedProcessor(Path jarPath, Integer jarJavaVersion) {
```
chanseokoh, line 36:
```suggestion
  SpringBootPackagedProcessor(Path jarPath, Integer jarJavaVersion) {
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
chanseokoh: While we're at it, how about fixing this too?

chanseokoh (commented):
LGTM, only minor comments.

In many cases, the previous approach to call some methods multiple times is better for readability and compact code in general. It's just that this was a special case involving opening a jar. Although it can still be the case that opening and inspecting is really quick and all of this is meaningless. But yeah, I think the current pattern doesn't look bad.

chanseokoh (approved):

PR #3089, opened 2021-02-26
Fix creation of release PR in release workflow. 

mpeddada1:
To be able to trigger unit tests from the release PR.

### DIFF
--- a/.github/workflows/jib-cli-release.yml
+++ b/.github/workflows/jib-cli-release.yml
@@ -52,6 +52,7 @@ jobs:
           source_branch: cli-release-v${{ github.event.inputs.release_version }}
           pr_title: "CLI release v${{ github.event.inputs.release_version }}"
           pr_body: "To be merged after the release is complete."
+          pr_label: "PR: Merge After Release"
 
       - name: Draft GitHub release
         uses: actions/create-release@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/jib-cli-release.yml
```diff
@@ -46,8 +46,8 @@ jobs:
       - name: Create pull request
         uses: repo-sync/pull-request@v2
         with:
-          # Provided by Actions. No need to create a token.
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          # Use a personal access token instead of the one created by Github Actions.
```
chanseokoh, line 49: Something like

```
# Use a personal token to file a PR as a non-bot author to trigger other workflows (e.g., unit tests):
# <link to the official doc about why using a personal token>;
```

.github/workflows/jib-cli-release.yml
```diff
@@ -46,8 +46,8 @@ jobs:
...
+          github_token: ${{ secrets.GA_RELEASE_PR_PERSONAL_TOKEN }}
           source_branch: cli-release-v${{ github.event.inputs.release_version }}
           pr_title: "CLI release v${{ github.event.inputs.release_version }}"
           pr_body: "To be merged after the release is complete."
```
chanseokoh, line 53: Remove the kokoro label below.

chanseokoh (approved):

.github/workflows/jib-cli-release.yml
chanseokoh: Oh, we should also add `PR: Merge After Release` too.

mpeddada1: oops nice catch

PR #3093, opened 2021-02-27
Feature/set environment variable for cred helper

### LINKED ISSUE #2814: Set environment variables for credHelper
I want to use the "ecr-login" credHelper for both `from` and `to`, but I need to use different environment variables (`AWS_PROFILE` in my case), because the IAM is different in each case. If it is possible somehow, I would like to know, and if not, I would like to request this feature (just like e.g. `environment` can be set for `dockerClient`).

As a workaround, I will now try to run the credHelper manually and pass in username/password values via system properties (`-Djib.to.auth.username=...`)

vinod-tahelyani:
Fixes #2814

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -28,12 +28,16 @@ import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Nested;
 import org.gradle.api.tasks.Optional;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /** Object in {@link JibExtension} that configures the base image. */
 public class BaseImageParameters {
 
   private final AuthParameters auth;
   private Property<String> image;
   @Nullable private String credHelper;
+  @Nullable private Map<String, String> environment;
   private final PlatformParametersSpec platformParametersSpec;
   private final ListProperty<PlatformParameters> platforms;
 
@@ -48,6 +52,7 @@ public class BaseImageParameters {
     amd64Linux.setArchitecture("amd64");
     amd64Linux.setOs("linux");
     platforms.add(amd64Linux);
+    environment = new HashMap<>();
   }
 
   @Nested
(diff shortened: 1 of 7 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleRawConfiguration.java
```diff
@@ -225,4 +221,22 @@ public Path getImageJsonOutputPath() {
   public List<? extends PlatformConfiguration> getPlatforms() {
     return jibExtension.getFrom().getPlatforms().get();
   }
+
+  @Override
+  public Map<String, String> getFromEnvironment() {
```
chanseokoh, line 226: getFromCredHelperEnvironment

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -93,6 +98,17 @@ public void setCredHelper(String credHelper) {
...
+  @Input
+  @Nullable
+  @Optional
+  public Map<String, String> getEnvironment() {
+    return environment;
+  }
+
+  public void setEnvironment(Map<String, String> environment) {
+    this.environment = environment;
+  }
```
chanseokoh, line 110: The way we want is to set this through `jib.from.credHelper.environment` as I showed in https://github.com/GoogleContainerTools/jib/issues/2814#issuecomment-783645262, not through `jib.from.environment` as you did here.

An important thing is to maintain backward compatibility. That is, all of these should work:
```
jib.from {
    image = ...
    credHelper = 'ecr-login`
}
```
```
jib.from {
    image = ...
    credHelper.helper = 'ecr-login'
}
```
```
jib.from {
    image = ...
    credHelper {
        helper = 'ecr-login'
        environment = ...
    }
}
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TargetImageParameters.java
```diff
@@ -112,6 +117,17 @@ public void setCredHelper(String credHelper) {
...
+  @Input
+  @Nullable
+  @Optional
+  public Map<String, String> getEnvironment() {
+    return environment;
+  }
+
+  public void setEnvironment(Map<String, String> environment) {
+    this.environment = environment;
+  }
```
chanseokoh, line 129: ditto

chanseokoh (commented):
Thanks for your contribution!

I have some high-level comments. And be sure to run `./gradlew googleJavaFormat` (or simply, `./gradlew goJF`) to auto-format the code.

chanseokoh:
@vinod-tahelyani just checking in, do you still plan to complete this PR?

chanseokoh:
Closing (https://github.com/GoogleContainerTools/jib/issues/2814#issuecomment-871456317).

PR #3095, opened 2021-03-01
[jib-cli-jar] Add clearer error message for when the class file format is invalid. 

mpeddada1:


chanseokoh (approved):
LGTM. Hopefully this doesn't break any test?

mpeddada1:
It doesn't break any existing tests but it might be a good idea to add tests for this.

chanseokoh (approved):
LGTM. Just curious, how did you hit this error? Why was the class invalid? Will the user easily encounter this?

mpeddada1:
> LGTM. Just curious, how did you hit this error? Why was the class invalid? Will the user easily encounter this?

Good question. In order to hit this error, the class file would have to not start with `0xCAFEBABE` (which is the standard java class file format). In this test, the class file in question, `class1.class`, started with some invalid value (`invalid class` or `0 1 2 3` for example). It is a little hard to say how often the user would run into this problem. The one scenario I can think of is that they have a corrupted class file but are unaware of it - in this case it might be better to have an informative message when determining the major version fails? But open to hearing your thoughts on this.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/jar/JarProcessors.java
```diff
@@ -116,6 +118,11 @@ static Integer determineJavaMajorVersion(Path jarPath) throws IOException {
             int majorVersion = classFile.readUnsignedShort();
             int javaVersion = (majorVersion - 45) + 1;
             return javaVersion;
+          } catch (EOFException ex) {
+            throw new IllegalArgumentException(
```
mpeddada1, line 122: Testing with a jar file containing a fake empty .class file  resulted in an EOFException with no error message. This change is to provide more clarity on why this exception is being thrown.

chanseokoh: Awesome. It's more likely that there's an empty `.class` file than a corrupted file.

chanseokoh (approved):

PR #3099, opened 2021-03-02
Upgrad gradle to 6.8.2

cromefire:
Just a simple upgrade and a one line change to keep the tests running


chanseokoh:
Thanks for doing this.

However, we intentionally builds our artifacts using an old Java version (Java 8) and Gradle version. By forcing an old Gradle version, we can avoid accidentally using new API available only in new Gradle versions. (Is there's a way to enforce an old Gradle API version for compilation while using a newer Gradle?)

cromefire:
I'll check I guess could just depend on an old gradle API

PR #3102, opened 2021-03-02
Update CHANGELOG for jib-cli 0.2.0

mpeddada1:

### DIFF
--- a/jib-cli/CHANGELOG.md
+++ b/jib-cli/CHANGELOG.md
@@ -12,6 +12,6 @@ All notable changes to this project will be documented in this file.
 ## 0.2.0
 
 ### Added
-- Added the `jar` command which can be used to containerize a JAR. Currently offering two modes: `exploded`, where the contents of the JAR are put into optimized layers on the container, and `packaged`, where the JAR is containerized as is.  ([#11](https://github.com/GoogleContainerTools/jib/projects/11))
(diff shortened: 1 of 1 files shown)


jib-cli/CHANGELOG.md
```diff
@@ -9,3 +9,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+## 0.2.0
+
+### Added
+- Added the `jar` command which can be used to containerize a JAR. Currently offering two modes: `exploded`, where the contents of the JAR are put into optimized layers on the container, and `packaged`, where the JAR is containerized as is.([#11](https://github.com/GoogleContainerTools/jib/projects/11))
```
chanseokoh, line 15: How about showcasing usage?

```suggestion
- Added the `jar` command which can be used to containerize a JAR with `$ jar jar --target ... my-app.jar`. Currently offering two modes: `exploded`, where the contents of the JAR are put into optimized layers on the container, and `packaged`, where the JAR is containerized as is. ([#11](https://github.com/GoogleContainerTools/jib/projects/11))
```
And I think we should not promote `packaged` as much as possible.

mpeddada1: Showing usage adds a nice touch.  

> And I think we should not promote `packaged` as much as possible.

Makes sense. Changed this a bit to only specify the default mode.

chanseokoh (approved):

chanseokoh (approved):

PR #3103, opened 2021-03-03
Fix typos in usage text for Jar command

mpeddada1:

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
@@ -91,14 +91,14 @@ public class Jar implements Callable<Integer> {
 
   @CommandLine.Option(
       names = "--expose",
-      paramLabel = "<exposed-ports>",
+      paramLabel = "<port>",
       split = ",",
       description = "Ports to expose on container, example: --expose=5000,7/udp.")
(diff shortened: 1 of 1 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -80,37 +80,37 @@
 
   @CommandLine.Option(
       names = "--jvm-flags",
-      paramLabel = "<jvm-flags>",
+      paramLabel = "<jvm-flag>",
```
chanseokoh, line 83: Isn't this only for textual message in the help text? Previously, I thought showing
```
--jvm-flags <jvm-flags> JVM arguments, ...
```
made sense, because you'll provide a list of flags. Why should it be in a singular form?

mpeddada1: I initially thought so too. But it shows up as  `--jvm-flags=<jvm-flags>[,<jvm-flags>...]  JVM arguments, example: --jvm-flags=-Dmy.property=value,-Xshare:off` which seems a little confusing. For additional-tags, we do  `--additional-tags=<tag>[,<tag>...] Additional tags for target image`

chanseokoh (approved):

PR #3112, opened 2021-03-04
Remove out-of-scope tests in `GradleProjectProperties`

chanseokoh:
I think existing tests already check if calling `setAppRoot()` correctly takes effect. No point in having multiple tests using different app root values (`/app`, `/my/app`, and something else).

If this looks good, I'll apply a similar change to `MavenProjectProperties`

### DIFF
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
@@ -301,7 +301,7 @@ public class GradleProjectPropertiesTest {
   }
 
   @Test
-  public void testCreateContainerBuilder_extractionPaths()
+  public void testCreateContainerBuilder_correctExtractionPaths()
       throws InvalidImageReferenceException, CacheDirectoryCreationException {
     ContainerBuilderLayers layers = new ContainerBuilderLayers(setupBuildContext());
 
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
@@ -392,25 +392,37 @@ public class GradleProjectPropertiesTest {
 
   @Test
   public void testCreateContainerBuilder_noErrorIfWebInfClassesDoesNotExist()
-      throws IOException, InvalidImageReferenceException, CacheDirectoryCreationException {
+      throws IOException, InvalidImageReferenceException {
     temporaryFolder.newFolder("WEB-INF", "lib");
     setUpWarProject(temporaryFolder.getRoot().toPath());
-    setupBuildContext(); // should pass
+
+    assertThat(
+            gradleProjectProperties.createJibContainerBuilder(
+                JavaContainerBuilder.from("ignored"), ContainerizingMode.EXPLODED))
+        .isNotNull();
   }
 
   @Test
   public void testCreateContainerBuilder_noErrorIfWebInfLibDoesNotExist()
-      throws IOException, InvalidImageReferenceException, CacheDirectoryCreationException {
+      throws IOException, InvalidImageReferenceException {
     temporaryFolder.newFolder("WEB-INF", "classes");
     setUpWarProject(temporaryFolder.getRoot().toPath());
-    setupBuildContext(); // should pass
+
+    assertThat(
+            gradleProjectProperties.createJibContainerBuilder(
+                JavaContainerBuilder.from("ignored"), ContainerizingMode.EXPLODED))
+        .isNotNull();
   }
 
   @Test
   public void testCreateContainerBuilder_noErrorIfWebInfDoesNotExist()
-      throws IOException, InvalidImageReferenceException, CacheDirectoryCreationException {
+      throws IOException, InvalidImageReferenceException {
     setUpWarProject(temporaryFolder.getRoot().toPath());
-    setupBuildContext(); // should pass
+
+    assertThat(
+            gradleProjectProperties.createJibContainerBuilder(
+                JavaContainerBuilder.from("ignored"), ContainerizingMode.EXPLODED))
+        .isNotNull();
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -420,63 +390,27 @@ public void testCreateContainerBuilder_war()
             "/my/app/WEB-INF/classes/package/Other.class");
   }
 
-  @Test
-  public void testCreateContainerBuilder_defaultWebAppRoot()
-      throws URISyntaxException, IOException, InvalidImageReferenceException,
```
chanseokoh, line 425: This test wasn't even testing if the extraction paths start with `DEFAULT_WEB_APP_ROOT`. The test is identical to the one right above.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -330,41 +328,13 @@ public void testCreateContainerBuilder_nonDefaultAppRoot()
         .containsExactly("/my/app/classes/HelloWorld.class", "/my/app/classes/some.class");
   }
 
-  @Test
-  public void testCreateContainerBuilder_defaultAppRoot()
```
chanseokoh, line 334: identical test right above

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -238,8 +238,7 @@ public void testGetMajorJavaVersion_jvm11() {
   @Test
   public void testCreateContainerBuilder_correctFiles()
```
mpeddada1, line 239:
```suggestion
  public void testCreateContainerBuilder_correctSourceFiles()
```

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -302,10 +301,9 @@ public void testCreateContainerBuilder_noClassesFiles()
   }
 
   @Test
-  public void testCreateContainerBuilder_nonDefaultAppRoot()
+  public void testCreateContainerBuilder_extractionPaths()
```
mpeddada1, line 304:
```suggestion
  public void testCreateContainerBuilder_correctExtractionPaths()
```

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesTest.java
```diff
@@ -420,63 +390,27 @@ public void testCreateContainerBuilder_war()
...
-            unzipTarget.resolve("WEB-INF/classes/package/Other.class"));
-  }
-
   @Test
   public void testCreateContainerBuilder_noErrorIfWebInfClassesDoesNotExist()
       throws IOException, InvalidImageReferenceException, CacheDirectoryCreationException {
     temporaryFolder.newFolder("WEB-INF", "lib");
     setUpWarProject(temporaryFolder.getRoot().toPath());
-    setupBuildContext("/anything"); // should pass
+    setupBuildContext(); // should pass
```
mpeddada1, line 398: Ah sorry, missed this earlier. The behavior seems a little vague here. Adding a few tests on what setUpBuildContext results in might help?

chanseokoh: There are tests to check the layers. These tests are like follow-ups for successful tests. For example, this one ensures "noErrorIfWebInfClassesDoesNotExist." Or, do you have any ideas?

mpeddada1: Ah I see. Hm ideally, we want to make the test to be as explicit as possible and make it so that it's easier to understand the behavior of the method without us having to look through the code. In this test, for example,  I know that calling `setupBuildContext` doesn't throw an error but there isn't a way for me to tell if we are creating the correct layers (unless I take a look at the code itself)? It looks like that other tests use wars that have both `Web-INF/Classes` and `Web-INF/Lib` so in this test, maybe asserting on the layers would help in spotting the difference between `War with both Web-INF/Classes and Web-INF/Lib` and `War without Web-INF/Classes`. But let me know what you think.

chanseokoh: Here's my perspective: basically, a unit test should test one thing.

The other test verifies if we are creating the correct layers. The test data happens to have both `WEB-INF/classes` and `WEB-INF/lib`. If we want to test correct layer creation with a slightly different test data, I think we should create a new unit test for it (which may be meaningful to see the difference of the layer outcome as you said, but I think isn't really necessary given the coverage of the other test).

Now, this test is to ensure that calling `MavenProjectProperties.createJibContainerBuilder()` does not throw an exception if there are no `.class` files. (OTOH, for non-WAR, it does throw an exception.) I agree `setupBuildContext()` hides some stuff, but since we're using `setupBuildContext()` in all other test to reduce boiler plate for calling `createJibContainerBuilder()`, it's following the same pattern. If we add layer-checking code into this test, I think not only it's testing two things but also distracts the reader from noticing what this test really verifies. We could add a comment, but often that's unnecessary with the test name `testCreateContainerBuilder_noErrorIfWebInfClassesDoesNotExist()`.

mpeddada1: It just feels a little strange that we're just calling `setupBuildContext` in our test without having some sort of an assert block. I totally agree with you that test name does a good job of explaining what the test is supposed to do, but at the moment there doesn't seem to be a difference between the test method and the code it is testing? 

And to your point about a test, testing one thing -> You're absolutely right that a test method should only test one behavior. But for this test,  the behavior we want to focus on is: "what happens when Web-inf/classes is missing". When we assert on the classes layer, the idea that no error was thrown becomes apparent when we check the layer being created. We would solve the issue of under-asserting instead of testing a second behavior?

chanseokoh: Yeah, you do have a point. Now asserting that it's successfully returning a built `JibContainerBuilder`.

mpeddada1 (approved):

PR #3118, opened 2021-03-05
Update FAQ for Jib CLI 

mpeddada1:


docs/faq.md
```diff
@@ -55,6 +60,16 @@ Check out [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-
 
 Also see [rules_docker](https://github.com/bazelbuild/rules_docker) for a similar existing container image build tool for the [Bazel build system](https://github.com/bazelbuild/bazel). The tool can build images for languages such as Python, NodeJS, Java, Scala, Groovy, C, Go, Rust, and D.
 
+### My build process doesn't let me integrate with the Jib plugins
+
+The [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli) can be useful for users with complex build workflows that make it hard to integrate the Jib Maven or Gradle plugins. It is a standalone application that is powered by [Jib-Core](#https://github.com/GoogleContainerTools/jib/tree/master/jib-core) and offers two commands:
+
+[Build](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli#build-command): Which builds from the file system content with the help of a yaml build file.
```
chanseokoh, line 67: Bullet-points would look better? And I think they can be more concise.

```
* [Build](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli#build-command): Builds images from the filesystem content.

* [Jar](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli#jar-command): Examines your JAR and builds an optimized image or containerizes the JAR as-is.

```

docs/faq.md
```diff
@@ -55,6 +60,16 @@ Check out [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-
 
 Also see [rules_docker](https://github.com/bazelbuild/rules_docker) for a similar existing container image build tool for the [Bazel build system](https://github.com/bazelbuild/bazel). The tool can build images for languages such as Python, NodeJS, Java, Scala, Groovy, C, Go, Rust, and D.
 
+### My build process doesn't let me integrate with the Jib plugins
```
chanseokoh, line 63: How about this?

```suggestion
### My build process doesn't let me integrate with the Jib Maven or Gradle plugin
```

docs/faq.md
```diff
@@ -55,6 +60,16 @@ Check out [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-
...
+
+The [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli) can be useful for users with complex build workflows that make it hard to integrate the Jib Maven or Gradle plugins. It is a standalone application that is powered by [Jib-Core](#https://github.com/GoogleContainerTools/jib/tree/master/jib-core) and offers two commands:
```
chanseokoh, line 65:
```suggestion
The [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli) can be useful for users with complex build workflows that make it hard to integrate the Jib Maven or Gradle plugins. It is a standalone application that is powered by [Jib Core](https://github.com/GoogleContainerTools/jib/tree/master/jib-core) and offers two commands:
```

docs/faq.md
```diff
@@ -756,3 +771,48 @@ On examining the container structure with [Dive](https://github.com/wagoodman/di
...
 Solution: The user installed the file in a different location.
+
+## Jib CLI 
+ 
+### How does the `jar` command support Standard JARs?
+
+A standard JAR can be containerized by the `jar` command in two modes, exploded or packaged. 
+
+#### Exploded Mode (Recommended)
+Achieved by calling `jib jar --target ${TARGET_REGISTRY} ${JAR_NAME}.jar`
```
chanseokoh, line 782: I think we can make the contents less verbose overall? Using less words. For example, instead of "It is the default mode for containerizing a JAR", just "The default mode."

Also check out https://github.com/GoogleContainerTools/jib/blob/02c9b5ce3e40c80f93ee900e4903c6d2cdc71cee/docs/faq.md#how-are-jib-applications-layered I think the contents have too much details? I think we don't need to mention `MANIFEST.MF`.

For example,

Dependencies Layer: other dependency JARs (if exist)
Snapshot-Dependencies Layer: snapshot dependencies (if exist)

docs/faq.md
```diff
@@ -756,3 +771,48 @@ On examining the container structure with [Dive](https://github.com/wagoodman/di
 The user had used Jib's ability to install extra files into the image ([Maven](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#adding-arbitrary-files-to-the-image), [Gradle](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#adding-arbitrary-files-to-the-image)) to install a library file by placing it in `src/main/jib/lib/libfoo.so`. This would normally cause the `libfoo.so` to be installed in the image as `/lib/libfoo.so`. But `/lib` and `/lib64` in the user's base image were symbolic links. Jib does not follow such symbolic links when creating the image. And at container initialization time, Docker treats these symlinks as a file, and thus the symbolic link was replaced with `/lib` as a new directory. As a result, none of the system shared libraries were resolved and dynamically-linked programs failed.
 
 Solution: The user installed the file in a different location.
+
+## Jib CLI 
+ 
+### How does the `jar` command support Standard JARs?
```
chanseokoh, line 777: The FAQ seems more like the technicalities of how a JAR is containerized?
```
### How does the `jar` command containerize Standard JARs?
```

And I think it's worth mentioning that we support both a thin JAR (where other dependencies JARs are specified in `MANIFEST.MF`) and a fat JAR. For thin JARs, adding this link may help: https://docs.oracle.com/javase/tutorial/deployment/jar/downman.html

chanseokoh:
When changing titles, don't forget to update the ToC and link refs.

docs/faq.md
mpeddada1: Much better!

docs/faq.md
mpeddada1: You're right. I cleaned it up a bit to make it less verbose.

docs/faq.md
```diff
@@ -756,3 +771,54 @@ On examining the container structure with [Dive](https://github.com/wagoodman/di
...
+#### Exploded Mode (Recommended)
+Achieved by calling `jib jar --target ${TARGET_REGISTRY} ${JAR_NAME}.jar`
+
+The default mode for containerizing a JAR. It will respect `layers.idx` (if present) or create layers in the following format:
+
+- Other Dependencies Layer
+- Spring-Boot-Loader Layer
+- Snapshot-Dependencies Layer
+- Resources Layer
+- Classes Layer
```
chanseokoh, line 815: For consistency with the [other FAQ](https://github.com/GoogleContainerTools/jib/blob/02c9b5ce3e40c80f93ee900e4903c6d2cdc71cee/docs/faq.md#how-are-jib-applications-layered), how about reversing the layer order?

docs/faq.md
```diff
@@ -756,3 +771,54 @@ On examining the container structure with [Dive](https://github.com/wagoodman/di
...
+**Entrypoint** : `java -jar ${JAR_NAME}.jar`
+
+### How does the `jar` command support Spring Boot JARs?
+The `jar` command currently supports containerization of Spring Boot Fat JARs.
+A Spring-Boot Fat JAR can be containerized in two modes, exploded or packaged. 
+
+#### Exploded Mode (Recommended)
+Achieved by calling `jib jar --target ${TARGET_REGISTRY} ${JAR_NAME}.jar`
+
+The default mode for containerizing a JAR. It will respect `layers.idx` (if present) or create layers in the following format:
```
chanseokoh, line 809: "`layers.idx` in the JAR" and "create optimized layers"?

Let's add link for `layers.idx`. One of the two below? Either looks good.

https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html#executable-jar-war-index-files-layers
https://spring.io/blog/2020/08/14/creating-efficient-docker-images-with-spring-boot-2-3

docs/faq.md
```diff
@@ -756,3 +771,54 @@ On examining the container structure with [Dive](https://github.com/wagoodman/di
...
+## Jib CLI 
+ 
+### How does the `jar` command support Standard JARs?
+
+A standard JAR can be containerized by the `jar` command in two modes, exploded or packaged. 
+
+#### Exploded Mode (Recommended)
+Achieved by calling `jib jar --target ${TARGET_REGISTRY} ${JAR_NAME}.jar`
+
+The default mode for containerizing a JAR. It will open up the JAR into the following layers:
```
chanseokoh, line 784: Something like this? The intention is to naturally drop "optimized building."

```suggestion
The default mode for containerizing a JAR. It will open up the JAR and optimally places files into the following layers:  
```

docs/faq.md
mpeddada1: You're right. I cleaned it up a bit to make it less verbose.

docs/faq.md
chanseokoh: Oh, actually, I think technically a fat JAR will work, but I realized we need to do more work to better support it. (I believe embedded dependencies do not go into the dependencies layer.) Supporting a fat JAR will be an important feature that we should eventually implement.

For now, I think we should document this limitation (i.e., inefficiency). Basically, I think we can say that we support  thin JARs (with the link in the comment above), and then mention the current limitation and that we are working on better fat JAR support. (Also this FAQ should probably have a link to the other FAQ for the Spring Boot fat jar case.) Thoughts?

docs/faq.md
```diff
@@ -756,3 +771,54 @@ On examining the container structure with [Dive](https://github.com/wagoodman/di
...
+
+It will result in the following layers on the container:
+- Dependencies Layer
+- Jar Layer
+
+**Entrypoint** : `java -jar ${JAR_NAME}.jar`
+
+### How does the `jar` command support Spring Boot JARs?
+The `jar` command currently supports containerization of Spring Boot Fat JARs.
+A Spring-Boot Fat JAR can be containerized in two modes, exploded or packaged.
```
chanseokoh, line 804: "fat" is actually an informal term.

```suggestion
The `jar` command currently supports containerization of Spring Boot fat JARs. A Spring-Boot fat JAR can be containerized in two modes, exploded or packaged. 
```

docs/faq.md
chanseokoh: BTW, is the order correct? Why is the spring boot loader in between the two dep layers?

docs/faq.md
```diff
@@ -55,6 +60,16 @@ Check out [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-
 
 Also see [rules_docker](https://github.com/bazelbuild/rules_docker) for a similar existing container image build tool for the [Bazel build system](https://github.com/bazelbuild/bazel). The tool can build images for languages such as Python, NodeJS, Java, Scala, Groovy, C, Go, Rust, and D.
 
+### My build process doesn't let me integrate with the Jib Maven or Gradle plugin
+
+The [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli) can be useful for users with complex build workflows that make it hard to integrate the Jib Maven or Gradle plugins. It is a standalone application that is powered by [Jib Core](https://github.com/GoogleContainerTools/jib/tree/master/jib-core) and offers two commands:
```
chanseokoh, line 65:
```suggestion
The [Jib CLI](https://github.com/GoogleContainerTools/jib/tree/master/jib-cli) can be useful for users with complex build workflows that make it hard to integrate the Jib Maven or Gradle plugin. It is a standalone application that is powered by [Jib Core](https://github.com/GoogleContainerTools/jib/tree/master/jib-core) and offers two commands:
```

docs/faq.md
mpeddada1: Ah this is how we ordered the layers for the non-layered JAR - the purpose was to make it similar to the default order defined by Spring Boot (when `layers.idx` is specified) https://spring.io/blog/2020/08/14/creating-efficient-docker-images-with-spring-boot-2-3

mpeddada1: Hm for reversing the ordering: this is how the layers are currently ordered in the cli so I wonder if it is better to leave it as it is to represent that?

chanseokoh: Well then just reverse the order in the other FAQ. In either way, we can achieve consistency.

docs/faq.md
```diff
@@ -168,12 +183,13 @@ Jib packages your Java application into the following paths on the image:
 Jib makes use of [layering](https://containers.gitbook.io/build-containers-the-hard-way/#layers) to allow for fast rebuilds - it will only rebuild the layers containing files that changed since the previous build and will reuse cached layers containing files that didn't change. Jib organizes files in a way that groups frequently changing files separately from large, rarely changing files. For example, `SNAPSHOT` dependencies are placed in a separate layer from other dependencies, so that a frequently changing `SNAPSHOT` will not force the entire dependency layer to rebuild itself.
 
 Jib applications are split into the following layers:
-* Classes
-* Resources
-* Project dependencies
-* Snapshot dependencies
-* All other dependencies
+
 * Each extra directory (`jib.extraDirectories` in Gradle, `<extraDirectories>` in Maven) builds to its own layer
```
chanseokoh, line 187: Sorry, actually we were not precise about the order before. With the reversed order, the extra directory layers should come after classes (below classes in the new order).

docs/faq.md
```diff
@@ -756,3 +772,59 @@ On examining the container structure with [Dive](https://github.com/wagoodman/di
...
 
 Solution: The user installed the file in a different location.
+
+## Jib CLI 
+ 
+### How does the `jar` command support Standard JARs?
+The Jib CLI supports both [thin JARs](https://docs.oracle.com/javase/tutorial/deployment/jar/downman.html) (where dependencies are specified in the JAR's manifest) and fat JARs.
+
+However, the current limitation of using a fat JAR is that the embedded dependencies will not be placed into the designated dependencies layers.
+They will instead be placed into the classes or resources layer. We hope to have better support for fat JARs in the future.
```
chanseokoh, line 782: How about something like this?

```suggestion
However, the current limitation of using a fat JAR is that the embedded dependencies will not be placed into the designated dependencies layers. They will instead be placed into the classes or resources layer. Therefore, for efficiency, we recommend against containerizing fat JARs (Spring Boot fat JARs are an [exception](and link to the FAQ below?)) if you can prepare thin JARs. We hope to have better support for fat JARs in the future.
```

And just saying, if you do want to render the second sentence as a new paragraph, you should put an empty line between before it. If starting a new paragraph was not your intention, I suggest not putting the second sentence in the next line to avoid confusion.

chanseokoh (approved):
LGTM, only one minor comment.

And please test the new ToC links after merging. (I made a mistake with the Jib config FAQ before, if you remember :smile:)

docs/faq.md
chanseokoh: > "`layers.idx` in the JAR" and "create optimized layers"?

Oh, and how about this?

docs/faq.md
chanseokoh: I think you missed the part "Therefore, for efficiency, we recommend against containerizing fat JARs (Spring Boot fat JARs are an [exception](and link to the FAQ below?)) if you can prepare thin JARs."

mpeddada1: great catch, just updated this.

mpeddada1:
> And please test the new ToC links after merging. (I made a mistake with the Jib config FAQ before, if you remember 😄)

Thanks for the heads up! Verified the links to make sure they were okay:)

chanseokoh (approved):

PR #3123, opened 2021-03-09
Deprecate Distroless methods in jib-core

chanseokoh:

PR #3124, opened 2021-03-09
Change the default base images from Distroless in Jib build plugins

chanseokoh:
~~_NOTE: this PR is to merge the changes into the feature branch `feature-branch-default-base-images`._ We are going to maintain the feature branch and regularly sync it with the main branch for a while.~~

The changes are on top of #3123.

1. For non-WAR projects, the new base images are `adoptopenjdk:8-jre` and `adoptopenjdk:11-jre`.
2. For WAR projects, the new base image is [`jetty`](https://hub.docker.com/_/jetty).
3. For WAR projects, we no longer inherits the entrypoint from the base image. The new default entrypoint is explicitly set to `java -jar /user/local/jetty/start.jar`. The idea is from https://github.com/GoogleContainerTools/jib/issues/1270#issuecomment-791525157.
4. Also cleans up a few places where Distroless is referenced.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
@@ -90,8 +90,8 @@ public class JavaContainerBuilder {
    *
    * @return a new {@link JavaContainerBuilder}
    * @see <a href="https://github.com/GoogleContainerTools/distroless">The distroless repository</a>
-   * @deprecated Use other {@code from()} methods with the image reference {@code
-   *     gcr.io/distroless/java}. However, using Distroless Java is no longer recommended.
+   * @deprecated Use {@code from()} with the image reference {@code gcr.io/distroless/java}.
+   *     However, using Distroless Java is no longer recommended.
    */
   @Deprecated
   public static JavaContainerBuilder fromDistroless() {
@@ -123,8 +123,8 @@ public class JavaContainerBuilder {
    *
    * @return a new {@link JavaContainerBuilder}
    * @see <a href="https://github.com/GoogleContainerTools/distroless">The distroless repository</a>
-   * @deprecated Use other {@code from()} methods with the image reference {@code
-   *     gcr.io/distroless/java/jetty} and change the app root by calling {@code
+   * @deprecated Use {@code from()} with the image reference {@code gcr.io/distroless/java/jetty}
+   *     and change the app root by calling {@code
    *     JavaContainerBuilder.setAppRoot("/jetty/webapps/ROOT")}. However, using Distroless Java is
    *     no longer recommended.
    */
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
@@ -90,7 +90,10 @@ public class JavaContainerBuilder {
    *
    * @return a new {@link JavaContainerBuilder}
    * @see <a href="https://github.com/GoogleContainerTools/distroless">The distroless repository</a>
+   * @deprecated Use other {@code from()} methods with the image reference {@code
+   *     gcr.io/distroless/java}. However, using Distroless Java is no longer recommended.
    */
+  @Deprecated
   public static JavaContainerBuilder fromDistroless() {
     try {
       return from(RegistryImage.named("gcr.io/distroless/java"));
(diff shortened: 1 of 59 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -588,7 +589,7 @@ static JavaContainerBuilder getJavaContainerBuilderWithBaseImage(
                 "mainClass, extraClasspath, jvmFlags, and expandClasspathDependencies "
                     + "are ignored for WAR projects"));
       }
-      return null;
+      return Arrays.asList("java", "-jar", "/usr/local/jetty/start.jar");
```
mpeddada1, line 592: Let's update the Javadoc to mention this

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/IncompatibleBaseImageJavaVersionException.java
```diff
@@ -19,7 +19,7 @@
 /**
  * Exception when the Java version in the base image is incompatible with the Java version of the
  * application to be containerized. For example, when the project is Java 11 but the base image is
- * {@code gcr.io/distroless/java:8}.
+ * {@code adoptopenjdk:8-jre-slim}.
```
mpeddada1, line 22:
```suggestion
 * {@code adoptopenjdk:8-jre}.
```

chanseokoh: nice catch

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
```diff
@@ -124,7 +125,8 @@ public static JavaContainerBuilder fromDistroless() {
    * @see <a href="https://github.com/GoogleContainerTools/distroless">The distroless repository</a>
    * @deprecated Use {@code from()} with the image reference {@code gcr.io/distroless/java/jetty}
    *     and change the app root by calling {@code
-   *     JavaContainerBuilder.setAppRoot("/jetty/webapps/ROOT")}.
+   *     JavaContainerBuilder.setAppRoot("/jetty/webapps/ROOT")}. However, using Distroless Java is
+   *     no longer recommended.
```
mpeddada1, line 129: Does this need to be?:
```suggestion
   *     JavaContainerBuilder.setAppRoot("var/lib/jetty/webapps/ROOT")}. However, using Distroless Java is
   *     no longer recommended.
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
mpeddada1: And just to make sure I'm understanding this correctly, what does this new entrypoint do exactly?

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
chanseokoh: Sharp eye. The answer is no. When using Distroless, the app root should be `/jetty/webapps`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh: It's the command to start the Jetty server. Jetty will start running and serving a deployed WAR app.

jib-core/src/main/java/com/google/cloud/tools/jib/api/JavaContainerBuilder.java
mpeddada1: Oh right, oops. I missed the part where distroless was mentioned haha

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
mpeddada1: Cool! Just out of curiosity - why was this entrypoint null before?

mpeddada1: Oh it was inherited from the base image

mpeddada1 (approved):

chanseokoh:
Adding the `kokoro:run` did nothing. Did an internal search, and it looks like we have to manually trigger Kokoro builds. Just did that.

chanseokoh:
Need to update CHANGELOG. When released, a lot of doc updates too.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
chanseokoh, line 592: FTR: it was wrong to return this when a custom base image is specified. The issue is fixed later in #3185.

Issue #3127, opened 2021-03-11
Intermittent AccessDeniedException while building the image


thoweber:
**Environment**:

- *Jib version:* jib-maven-plugin 2.8.0 
- *Build tool:* Apache Maven 3.6.2 on OpenJDK 11
- *OS:* Windows 10

**Description of the issue**:
The image build is intermittently failing while copying files in the jib cache. This issue is related to https://github.com/GoogleContainerTools/jib/issues/1763. The failure is caused by the virus scanner locking the file, but since the build is done in a work environment, there is not much I can do about it. The virus scanning software used is McAfee Endpoint Security 10.6.
Moreover the friendly exception message introduced with https://github.com/GoogleContainerTools/jib/pull/1887 is not shown. This is probably due to https://github.com/GoogleContainerTools/jib/blob/52d867fac945b05edbc0e977812dedfc1e7a03da/jib-core/src/main/java/com/google/cloud/tools/jib/cache/Retry.java#L135-L140 eventually rethrowing the exception instead of returning `false`.

**Expected behavior**:
The image build should not fail due to a virus scanner temporarily blocking access to a cache file. I suggest to increase the waiting period and/or the number of retries. Ideally this would be configurable by the user.
If the image build fails, the IOException with the friendly message from https://github.com/GoogleContainerTools/jib/blob/52d867fac945b05edbc0e977812dedfc1e7a03da/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java#L134-L139 should be thrown, instead of the raw AccessDeniedException.

```
[ERROR] Failed to execute goal com.google.cloud.tools:jib-maven-plugin:2.8.0:dockerBuild (default-cli) on project module: C:\GIT\project\module\target\jib-cache\tmp\9796658
997542756861 -> C:\GIT\project\module\target\jib-cache\layers\da49230cceb485be22432aa073ee7baf8dd3d868a13bc7eacb39688edf4d1e22 -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.google.cloud.tools:jib-maven-plugin:2.8.0:dockerBuild (default-cli) on project module: C:\GIT\project\module\target\jib-cache\tmp\9796658997542756861 -> C:\GIT\project\module\target\jib-cache\layers\da49230cceb485be22432aa073ee7baf8dd3d868a13bc7eacb39688edf4d1e22
[...]

PR #3133, opened 2021-03-12
Increase retry attempts and fix the issue of exception being consumed

chanseokoh:
Helps with #3127.

- `.maximumRetries(10)`, increased from 5. The delay is 15ms, so we only increase the total time by 75ms.
- As reported, the helpful error message is not shown when the operation fails due to repeated `FilesSstemException`s.
- Improves the error message. If the user still hits the issue, it's possible that the anti-virus software is impacting the performance a lot. The link provides not only a workaround but the potential to increase performance.

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageWriterTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageWriterTest.java
@@ -345,16 +345,14 @@ public class CacheStorageWriterTest {
     Exception exception =
         assertThrows(
             IOException.class,
-            () -> CacheStorageWriter.moveIfDoesNotExist(Paths.get("/foo"), Paths.get("/bar")));
+            () -> CacheStorageWriter.moveIfDoesNotExist(Paths.get("foo"), Paths.get("bar")));
     assertThat(exception)
         .hasMessageThat()
-        .isEqualTo(
-            "unable to move: /foo to /bar; such failures are often caused by interference from "
-                + "antivirus (https://github.com/GoogleContainerTools/jib/issues/3127#issuecomment-796838294), "
-                + "or rarely if the operation is not supported by the file system (for example: "
-                + "special non-local file system)");
+        .contains(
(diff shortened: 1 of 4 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageWriterTest.java
```diff
@@ -336,6 +340,23 @@ public void testWriteLocalConfig() throws IOException, URISyntaxException, Diges
     Assert.assertEquals("wasm", savedContainerConfig.getArchitecture());
   }
 
+  @Test
+  public void testMoveIfDoesNotExist_exceptionAfterFailure() {
+    Exception exception =
+        assertThrows(
+            IOException.class,
+            () -> CacheStorageWriter.moveIfDoesNotExist(Paths.get("/foo"), Paths.get("/bar")));
```
mpeddada1, line 348:
```suggestion
            () -> CacheStorageWriter.moveIfDoesNotExist(Paths.get("foo"), Paths.get("bar")));
```

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageWriterTest.java
```diff
@@ -336,6 +340,23 @@ public void testWriteLocalConfig() throws IOException, URISyntaxException, Diges
...
+    assertThat(exception)
+        .hasMessageThat()
+        .isEqualTo(
+            "unable to move: /foo to /bar; such failures are often caused by interference from "
+                + "antivirus (https://github.com/GoogleContainerTools/jib/issues/3127#issuecomment-796838294), "
+                + "or rarely if the operation is not supported by the file system (for example: "
+                + "special non-local file system)");
```
mpeddada1, line 355: sometimes asserting on the whole message can make the test brittle. How about we assert on just the essential part? : `unable to move: /foo to /bar; such failures are often caused by interference from antivirus`

mpeddada1 (approved):

PR #3157, opened 2021-03-19
[Jib CLI] fix concurrency issue with --http-trace

### LINKED ISSUE #3156: [Jib CLI] --http-trace sometimes doesn't work
The behavior is very unpredictable. That makes me suspect a concurrency issue. And it seems to always work with `--serialize`.

chanseokoh:
Fixes #3156.

I am not able to reproduce the issue with this fix.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
@@ -34,7 +34,9 @@ import picocli.CommandLine;
 public class JibCli {
 
   static Logger configureHttpLogging(Level level) {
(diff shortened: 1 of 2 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
```diff
@@ -33,6 +34,7 @@
 public class JibCli {
 
   static Logger configureHttpLogging(Level level) {
+    new ApacheHttpTransport(); // fixes https://github.com/GoogleContainerTools/jib/issues/3156
```
mpeddada1, line 37: Let's add a small comment to detail what this does? Something like, "to initialize a static logger field.."

mpeddada1 (approved):

PR #3159, opened 2021-03-20
#3158 - [Jib core] Tar archives with same contents are not reproducible

### LINKED ISSUE #3158: [Jib core] Tar archives with same contents are not reproducible
**Environment**:

- *Jib version:*  v0.18.0-core
- *Build tool:* Gradle 6.01
- *OS:* macOS 10.15.7

**Description of the issue**:
I create a tar file from a registry image.
`Jib.from(RegistryImage.named(image)).containerize(Containerizer.to(TarImage.at(outputPath).named(image)));`

This is being executed as part of a bazel genrule, with the tar file as the output.  We discovered that if this is rerun, the tar file created is different when using the same inputs, run on the same machine but at a later time.

**Expected behavior**:
We expect the tar file to be identical if the inputs have not changed.

**Steps to reproduce**:

  1.  Create tar file using Jib as above
  2. Rename the tar file output.tar
  3. Wait 5 seconds, and create another tar file using Jib as above, rename the tar output2.tar
  4. shassum -a 256 output.tar
  5. shassum -a 256 output2.tar

**Additional Information**: 
The issue is due to TarStreamBuilder creating blobs using new TarArchiveEntry(name)
By default this sets the mod time of the blob to be current time, thus making the resulting Tar image not reproducible.

`this.modTime = new Date().getTime() / MILLIS_PER_SECOND;`

#3159 3159

davidtron:
Fixes #3158.

Sets the mod time on TarArchiveEntry to DEFAULT_MODIFICATION_TIME (Epoch second 1) in order to create reproducible Tar images. Associated unit test creates 2 tars output streams from the same input 2 seconds apart to show the failure if we don't explicitly set the TarArchiveEntry mod time

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageTarball.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageTarball.java
@@ -31,6 +31,7 @@ import com.google.common.collect.ImmutableSet;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
+import java.time.Instant;
 import java.util.Collections;
 
 /** Translates an {@link Image} to a tarball that can be loaded into Docker. */
@@ -48,6 +49,7 @@ public class ImageTarball {
   private final Image image;
   private final ImageReference imageReference;
   private final ImmutableSet<String> allTargetImageTags;
+  private final Instant creationTime = Instant.EPOCH;
 
   /**
    * Instantiate with an {@link Image}.
@@ -88,7 +90,8 @@ public class ImageTarball {
       DescriptorDigest digest = layer.getBlobDescriptor().getDigest();
       long size = layer.getBlobDescriptor().getSize();
 
-      tarStreamBuilder.addBlobEntry(layer.getBlob(), size, "blobs/sha256/" + digest.getHash());
+      tarStreamBuilder.addBlobEntry(
+          layer.getBlob(), size, "blobs/sha256/" + digest.getHash(), creationTime);
       manifest.addLayer(size, digest);
     }
 
@@ -99,21 +102,26 @@ public class ImageTarball {
     manifest.setContainerConfiguration(configDescriptor.getSize(), configDescriptor.getDigest());
     tarStreamBuilder.addByteEntry(
         JsonTemplateMapper.toByteArray(containerConfiguration),
-        "blobs/sha256/" + configDescriptor.getDigest().getHash());
+        "blobs/sha256/" + configDescriptor.getDigest().getHash(),
+        creationTime);
 
     // Adds the manifest to the tarball
     BlobDescriptor manifestDescriptor = Digests.computeDigest(manifest);
     tarStreamBuilder.addByteEntry(
         JsonTemplateMapper.toByteArray(manifest),
-        "blobs/sha256/" + manifestDescriptor.getDigest().getHash());
+        "blobs/sha256/" + manifestDescriptor.getDigest().getHash(),
+        creationTime);
 
     // Adds the oci-layout and index.json
     tarStreamBuilder.addByteEntry(
-        "{\"imageLayoutVersion\": \"1.0.0\"}".getBytes(StandardCharsets.UTF_8), "oci-layout");
+        "{\"imageLayoutVersion\": \"1.0.0\"}".getBytes(StandardCharsets.UTF_8),
+        "oci-layout",
+        creationTime);
     OciIndexTemplate index = new OciIndexTemplate();
     // TODO: figure out how to tag with allTargetImageTags
     index.addManifest(manifestDescriptor, imageReference.toStringWithQualifier());
-    tarStreamBuilder.addByteEntry(JsonTemplateMapper.toByteArray(index), "index.json");
+    tarStreamBuilder.addByteEntry(
+        JsonTemplateMapper.toByteArray(index), "index.json", creationTime);
 
     tarStreamBuilder.writeAsTarArchiveTo(out);
   }
@@ -127,7 +135,7 @@ public class ImageTarball {
       String layerName = layer.getBlobDescriptor().getDigest().getHash() + LAYER_FILE_EXTENSION;
 
       tarStreamBuilder.addBlobEntry(
-          layer.getBlob(), layer.getBlobDescriptor().getSize(), layerName);
+          layer.getBlob(), layer.getBlobDescriptor().getSize(), layerName, creationTime);
       manifestTemplate.addLayerFile(layerName);
     }
 
@@ -136,7 +144,8 @@ public class ImageTarball {
         new ImageToJsonTranslator(image).getContainerConfiguration();
     tarStreamBuilder.addByteEntry(
         JsonTemplateMapper.toByteArray(containerConfiguration),
-        CONTAINER_CONFIGURATION_JSON_FILE_NAME);
+        CONTAINER_CONFIGURATION_JSON_FILE_NAME,
+        creationTime);
 
     // Adds the manifest to tarball.
     for (String tag : allTargetImageTags) {
@@ -144,7 +153,8 @@ public class ImageTarball {
     }
     tarStreamBuilder.addByteEntry(
         JsonTemplateMapper.toByteArray(Collections.singletonList(manifestTemplate)),
-        MANIFEST_JSON_FILE_NAME);
+        MANIFEST_JSON_FILE_NAME,
+        creationTime);
 
     tarStreamBuilder.writeAsTarArchiveTo(out);
   }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageTarball.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageTarball.java
@@ -49,7 +49,7 @@ public class ImageTarball {
   private final Image image;
   private final ImageReference imageReference;
   private final ImmutableSet<String> allTargetImageTags;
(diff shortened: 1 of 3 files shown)


loosebazooka:
I think this might be working as intended. Where are you encountering this as an issue that needs to be fixed? ReproducibleLayerBuilder adds in the behavior of reproducibility.

davidtron:
The problem occurs when I create a Tar file using Jib
`Jib.from(RegistryImage.named(image)).containerize(Containerizer.to(TarImage.at(outputPath).named(image)));`
I've put the steps to reproduce the issue #3158 and an example unit test that will fail without the proposed change to set the tar modtime to a fixed time.

My question is - should Jib produce a reproducible tar file when the input files are the same but the command has been run at a different time?

loosebazooka:
Ah okay, I see. 

I think this needs to be solved in ImageTarBall's calls to TarStreamBuilder (#addByteEntry and #addBlobEntry). Those methods should be setting the timestamp you want on the entries.

@chanseokoh does that make sense?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -89,6 +91,7 @@ public void addByteEntry(byte[] contents, String name) {
   public void addBlobEntry(Blob blob, long size, String name) {
     TarArchiveEntry entry = new TarArchiveEntry(name);
     entry.setSize(size);
+    entry.setModTime(DEFAULT_MODIFICATION_TIME.getEpochSecond());
```
chanseokoh, line 94: I realized that this isn't enough. The `addByteEntry()` above also needs to set the timestamp. Both `addBlobEntry()` and `addByteEntry()` are used in `ImageTarball` to create a tar image stream.

However, we have layers of abstraction in our codebase with a hierarchy, so it doesn't feel right to make `TarStreamBuilder` to get into the business of deciding the timestamp of tar entries. This is a general class to build a tar stream. What I think we should do is to change the signature of the method to `addBlobEntry(Blob blob, long size, String name, Instant modTime)`, and have `ImageTarball` pass in the time.

chanseokoh (commented):
Thanks a lot for your contribution! I think we can accept the general direction to reset timestamps. Please check out the comment below.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
chanseokoh: > I think this needs to be solved in ImageTarBall's calls to TarStreamBuilder (#addByteEntry and #addBlobEntry). Those methods should be allowing a timestamp parameter.

Oh, just realized @loosebazooka already suggested moving this to `ImageTarball` 😂

chanseokoh:
Note for self: CHANGELOG

chanseokoh:
@davidtron just checking in, do you plan to continue the work?

davidtron:
@chanseokoh yes just updated the code to pull up to ImageTarball as suggested

jib-core/src/main/java/com/google/cloud/tools/jib/image/ImageTarball.java
```diff
@@ -48,6 +49,7 @@
   private final Image image;
   private final ImageReference imageReference;
   private final ImmutableSet<String> allTargetImageTags;
+  private final Instant creationTime = Instant.EPOCH;
```
chanseokoh, line 52:
```suggestion
  private static final Instant TAR_ENTRY_MODIFICATION_TIME = Instant.EPOCH;
```
and move the field up.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -71,10 +72,12 @@ public void addTarArchiveEntry(TarArchiveEntry entry) {
    *
    * @param contents the bytes to add to the tarball
    * @param name the name of the entry (i.e. filename)
+   * @param modTime the time the entry is created
```
chanseokoh, line 75:
```suggestion
   * @param modificationTime the modification time of the entry
```

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -71,10 +72,12 @@ public void addTarArchiveEntry(TarArchiveEntry entry) {
...
    */
-  public void addByteEntry(byte[] contents, String name) {
+  public void addByteEntry(byte[] contents, String name, Instant modTime) {
```
chanseokoh, line 77:
```suggestion
  public void addByteEntry(byte[] contents, String name, Instant modificationTime) {
```

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarStreamBuilderTest.java
```diff
@@ -46,6 +47,7 @@
   private byte[] fileAContents;
   private byte[] fileBContents;
   private TarStreamBuilder testTarStreamBuilder = new TarStreamBuilder();
+  private final Instant creationTime = Instant.EPOCH;
```
chanseokoh, line 50:
```suggestion
```

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarStreamBuilderTest.java
```diff
@@ -96,10 +98,11 @@ public void testToBlob_stringsAndTarArchiveEntriesWithCompression() throws IOExc
...
   @Test
   public void testToBlob_multiByte() throws IOException {
-    testTarStreamBuilder.addByteEntry("日本語".getBytes(StandardCharsets.UTF_8), "test");
-    testTarStreamBuilder.addByteEntry("asdf".getBytes(StandardCharsets.UTF_8), "crepecake");
+    testTarStreamBuilder.addByteEntry("日本語".getBytes(StandardCharsets.UTF_8), "test", creationTime);
+    testTarStreamBuilder.addByteEntry(
+        "asdf".getBytes(StandardCharsets.UTF_8), "crepecake", creationTime);
     testTarStreamBuilder.addBlobEntry(
-        Blobs.from("jib"), "jib".getBytes(StandardCharsets.UTF_8).length, "jib");
+        Blobs.from("jib"), "jib".getBytes(StandardCharsets.UTF_8).length, "jib", creationTime);
```
chanseokoh, line 105: In tests, just do
```suggestion
        Blobs.from("jib"), "jib".getBytes(StandardCharsets.UTF_8).length, "jib", Instant.EPOCH);
```

chanseokoh (commented):
Thanks!

Can you add some tests? I think its' not difficult to do so in `TarStreamBuilderTest`, as I see other tests open a tar and go over `TarArchiveEntry`?

davidtron:
Just updating the test, I found that when we add a TarEntry to TarStreamBuilder we don't manipulate the time, instead rely on the fact that ReproducibleLayerBuilder which delegates to UniqueTarArchiveEntries to manage the mod time.

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -71,10 +72,12 @@ public void addTarArchiveEntry(TarArchiveEntry entry) {
    *
    * @param contents the bytes to add to the tarball
    * @param name the name of the entry (i.e. filename)
+   * @param modificationTime the modification time of the entry
    */
-  public void addByteEntry(byte[] contents, String name) {
+  public void addByteEntry(byte[] contents, String name, Instant modificationTime) {
     TarArchiveEntry entry = new TarArchiveEntry(name);
     entry.setSize(contents.length);
+    entry.setModTime(modificationTime.getEpochSecond());
```
chanseokoh, line 80: Shouldn't this be `toEpochMillis()`? I guess writing a test will confirm it.

chanseokoh (commented):
Again, thanks for your contribution!

`TarStreamBuilder` is completely independent of `ReproducibleLayerBuilder`, so no need to bring that into your thinking. It's just that `addTarArchiveEntry(TarArchiveEntry)` accepts a fully-built tar entry as-is (we should not manipulate the entry at all). OTOH, the other add methods internally create a new tar entry with the given properties, so it's only natural to set those properties.

So I'm not exactly sure how you're writing tests, but you can test all of them to check if they all work in the intended way. If `addTarArchiveEntry(TarArchiveEntry)` adds an entry whose modTime is set to 2021-04-09, then you can just check if the result tar archive has the very entry with the same modTime?

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
davidtron: Hi @chanseokoh the tests are in the unit test `TarStreamBuilderTest`
You are quite right it should be passed in using `toEpochMillis` because the implementation of TarArchiveEntry divides whats passed in by 1000.  I've updated the unit test to set an explicit time rather than the Epoch to catch that, and show that the expected time is `MODIFICATION_TIME.truncatedTo(SECONDS)`

chanseokoh: I don't see the changes. Forgot to push a new commit?

chanseokoh (approved):
Thanks for your contribution!

I actually wanted to have separate and dedicated unit tests, but given the back-and-forth, I can take care of it after merging this.

Issue #3161, opened 2021-03-22
[Jib CLI] Automatic update checks


mpeddada1:
Implement automatic update checker for the Jib CLI.  This feature will be used to check if the latest version is available and show a message to user, letting them know that they can update to the latest version if they haven't already done so.  This is an opt-out feature.  

Note that the update checker already exists for the Jib Maven and Gradle plugins. #2193 for reference.

mpeddada1:
Closed with #3165

PR #3165, opened 2021-03-24
Implement update checker for Jib CLI

mpeddada1:
For #3161


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
```diff
@@ -106,9 +109,15 @@ public Integer call() {
...
             spec.commandLine().getErr(),
             executor);
+    Future<Optional<String>> updateCheckFuture = Futures.immediateFuture(Optional.empty());
     try {
       JibCli.configureHttpLogging(commonCliOptions.getHttpTrace().toJulLevel());
-
+      GlobalConfig globalConfig = GlobalConfig.readConfig();
+      updateCheckFuture =
+          JibCli.newUpdateChecker(
+              globalConfig,
```
chanseokoh, line 118:
```suggestion
              GlobalConfig.readConfig(),
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
```diff
@@ -67,6 +79,45 @@ static void logTerminatingException(
             + "\u001B[0m");
   }
 
+  static Future<Optional<String>> newUpdateChecker(
+      GlobalConfig globalConfig, Verbosity verbosity, Consumer<LogEvent> log) {
+    if (!verbosity.equals(Verbosity.info) || globalConfig.isDisableUpdateCheck()) {
```
chanseokoh, line 84: I think when the verbosity >= info.

chanseokoh (commented):
There's an issue with the update message that `UpdateChecker.performUpdateCheck()` generates. It says "... Update your build configuration to use the latest features and fixes!"

And I see that Jib build plugins and Jib CLI will write to the same `lastUpdateCheck` file, but I think that's fine.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
```diff
@@ -67,6 +79,45 @@ static void logTerminatingException(
...
+      return Futures.immediateFuture(Optional.empty());
+    }
+    ExecutorService executorService = Executors.newSingleThreadExecutor();
+    try {
+      return UpdateChecker.checkForUpdate(
+          executorService,
+          VERSION_URL,
+          JibCli.class.getPackage().getImplementationTitle(),
+          JibCli.class.getPackage().getImplementationVersion(),
```
chanseokoh, line 93: So `.getPackage()` works only when the `JibCli` is packaged in a `.jar` file, which is true now. But I fear it may not work when we generate a native executable. Should we use `VersionInfo.TOOL_NAME` and `VersionInfo.getSimpleVersion()`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/JibCliTest.java
```diff
@@ -19,21 +19,37 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
 
+@RunWith(MockitoJUnitRunner.class)
 public class JibCliTest {
 
+  @Rule public final TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  @Mock GlobalConfig globalConfig;
+  @Mock ConsoleLogger logger;
```
chanseokoh, line 51: private

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -122,11 +122,13 @@
...
         return Optional.of(
-            "A new version of Jib ("
+            "A new version of "
+                + toolName
+                + " ("
                 + version.latest
                 + ") is available (currently using "
                 + currentVersion
-                + "). Update your build configuration to use the latest features and fixes!");
+                + "). Upgrade to use the latest features and fixes!");
```
chanseokoh, line 131: So I think we want to customize the update message. For Jib CLI, it'll really help if it can directly present the GitHub Release for the new version.

chanseokoh:
Hmm... I always thought there should be a way to refactor `UpdateChecker` and the code around it to be simpler and more general. I feel the current code is a bit convoluted.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
mpeddada1: Good thinking

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/JibCliTest.java
```diff
@@ -68,4 +82,12 @@ public void testLogTerminatingException_stackTrace() {
         .log(LogEvent.Level.ERROR, "\u001B[31;1mjava.io.IOException: test error message\u001B[0m");
     verifyNoMoreInteractions(logger);
   }
+
+  @Test
+  public void testNewUpdateChecker_noUpdateCheck() throws ExecutionException, InterruptedException {
+    when(globalConfig.isDisableUpdateCheck()).thenReturn(true);
+    Future<Optional<String>> updateChecker =
+        JibCli.newUpdateChecker(globalConfig, Verbosity.info, ignored -> {});
+    assertThat(updateChecker.get()).isEqualTo(Optional.empty());
```
chanseokoh, line 91:
```suggestion
    assertThat(updateChecker.get()).isEmpty();
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -121,12 +121,22 @@
...
+        String updateMessage =
+            "A new version of "
+                + toolName
+                + " ("
                 + version.latest
                 + ") is available (currently using "
                 + currentVersion
-                + "). Update your build configuration to use the latest features and fixes!");
+                + "). ";
+        if (toolName.equals("jib-cli")) {
```
chanseokoh, line 132: Hmm... although it works, I think this is too hacky to accept. Maybe we can make this method return only the latest version string (i.e., `version.latest`), and have the caller construct the message (maybe in `finishUpdateChecker`)? Or, any other thoughts?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
mpeddada1: That does make sense. I modified the message for the jib-cli to include a link to the release page.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -121,12 +121,22 @@
...
+          return Optional.of(
+              updateMessage
+                  + "Download the latest jib-cli version from https://github.com/GoogleContainerTools/jib/releases.");
```
chanseokoh, line 135: And since we have the `version.latest`, I think we can directly link the Release with the target version?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
mpeddada1: You're absolutely right. I refactored the code to make this method return `version.latest` and constructed the message in `finishUpdateCheck`.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
mpeddada1: Good point

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
```diff
@@ -37,6 +48,9 @@
     subcommands = {Build.class, Jar.class})
 public class JibCli {
 
+  private static final String VERSION_URL = "https://storage.googleapis.com/jib-versions/jib-cli";
+  private static final String TOOL_VERSION = VersionInfo.getVersionSimple();
```
chanseokoh, line 52: I find no need to declare it as a constant field.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/JibCli.java
```diff
@@ -67,6 +81,49 @@ static void logTerminatingException(
...
+                      + ") is available (currently using "
+                      + TOOL_VERSION
+                      + "). Download the latest jib-cli version from "
+                      + ProjectInfo.GITHUB_URL
+                      + "/releases/tag/v"
+                      + latestVersion
+                      + "-cli"
+                      + "\n"
+                      + ProjectInfo.GITHUB_URL
+                      + "/blob/master/jib-cli/CHANGELOG.md\u001B[0m"
```
chanseokoh, line 120: I think we should have `\n` after this. Maybe `\n\n` is better?

It's a bit hard to parse this and check if this is correctly formed. Maybe `String.format` can improve readability? And I think we can just embed "Jib CLI" without `VersionInfo.TOOL_NAME`.

```java
String message = "A new version of Jib CLI (%d) is available (currently using %d). Download the latest Jib CLI from %s/release/tag/v%d-cli\n"
    + "%s/blob/master/jib-cli/CHANGELOG.md\u001B[0m\n\n"
    + ...
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/JibCliTest.java
```diff
@@ -17,23 +17,42 @@
...
 import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
 
+@RunWith(MockitoJUnitRunner.class)
 public class JibCliTest {
 
+  @Rule public final TemporaryFolder temporaryFolder = new TemporaryFolder();
```
chanseokoh, line 51: Why do we need this?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/TaskCommonTest.java
```diff
@@ -71,7 +82,26 @@ public void testGetBootWarTask_bootWarProject() {
     project.getPlugins().apply(SpringBootPlugin.class);
 
     TaskProvider<Task> bootWarTask = TaskCommon.getBootWarTaskProvider(project);
-    Assert.assertNotNull(bootWarTask);
-    Assert.assertNotNull(bootWarTask instanceof BootWar);
+    assertThat(bootWarTask).isNotNull();
+    assertThat(bootWarTask instanceof BootWar).isNotNull();
```
chanseokoh, line 86: Doesn't it have `assertThat(bootWarTask).isInstanceOf(...)`?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
```diff
@@ -140,9 +134,9 @@
...
   /**
-   * Returns a message indicating Jib should be upgraded if the check succeeded and the current
-   * version is outdated, or returns {@code Optional.empty()} if the check was interrupted or did
-   * not determine that a later version was available.
+   * Returns the latest Jib version available if the check succeeded and the current version is
+   * outdated, or returns {@code Optional.empty()} if the check was interrupted or did not determine
+   * that a later version was available.
    *
    * @param updateMessageFuture the {@link Future} returned by {@link UpdateChecker#checkForUpdate}
    * @return the {@link Optional} message to upgrade Jib if a later version was found, else {@code
```
chanseokoh, line 142:
```suggestion
   * @return the latest version, if found
```

chanseokoh:
And did you say you'll update CHANGELOG?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/JibCliTest.java
mpeddada1: Good catch

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/TaskCommonTest.java
```diff
@@ -60,8 +71,8 @@ public void testGetWarTask_normalWarProject() {
     project.getPlugins().apply(WarPlugin.class);
 
     TaskProvider<Task> warTask = TaskCommon.getWarTaskProvider(project);
-    Assert.assertNotNull(warTask);
-    Assert.assertNotNull(warTask instanceof War);
```
chanseokoh, line 64: Oh, looks like the old check was plainly wrong.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/TaskCommon.java
```diff
@@ -67,19 +68,27 @@
     }
   }
 
+  @VisibleForTesting
```
chanseokoh, line 71: The method is required to be package-private, not just for testing.
```suggestion
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/MojoCommon.java
```diff
@@ -76,18 +76,21 @@ static void finishUpdateChecker(
...
-                  LogEvent.lifecycle(
-                      "Please see https://github.com/GoogleContainerTools/jib/blob/master/docs/privacy.md for info on disabling this update check."));
-              projectProperties.log(LogEvent.lifecycle(""));
+            latestVersion -> {
+              String changelogUrl =
+                  ProjectInfo.GITHUB_URL + "/blob/master/jib-maven-plugin/CHANGELOG.md";
+              String privacyUrl = ProjectInfo.GITHUB_URL + "/blob/master/docs/privacy.md";
+              String message =
+                  String.format(
+                      "\n\u001B[33mA new version of %s (%s) is available (currently using %s). Update your"
```
chanseokoh, line 85: Sorry, missed this. For Maven, we intentionally use multiple `.log()` instead of `\n` when we need to start a new line. Not a big deal, but this is because Maven prepends the verbosity info on every line, for example,
```
[INFO] Containerizing application to francium25/test...
```

chanseokoh (commented):
LGTM. Only minor comments.

PR #3171, opened 2021-03-29
Add inclusion/exclusion filtering configuration for ExtraDirectories in Maven and Gradle

### LINKED ISSUE #2564: Allow to add single file to a container (with setting rights)
**Environment**:

- *Jib version:* 2.4.0
- *Build tool:* maven:3.6.3-amazoncorretto-11
- *OS:* Amazon Linux 2

**Description of the issue**:

It would be useful if it was possible to add single file to a container (with setting rights). I am currently migrating from one system to the other and for some time I need to keep things compatible between both of them. Currently it is not possible (without workaround) to move the following Dockerfile command to jib-maven-plugin (it's possible only for directories):

COPY ./ssh/id_rsa /home/myuser/.ssh/id_rsa

**Steps to reproduce**:

    <extraDirectories>
        <paths>
            <path>
                <from>./ssh/id_rsa</from>
                <into>/home/myuser/.ssh/id_rsa</into>
             </path>
        </paths>
    </extraDirectories>

Failed to execute goal com.google.cloud.tools:jib-maven-plugin:2.4.0:build (default-cli) on project config: /home/builder/src/config/ssh/id_rsa is not a directory -> [Help 1]

### LINKED ISSUE #3164: Enhancement: Have an <extraFiles> option to add individual files instead of a whole directory
**Environment**:

- *Jib version:*2.8.0
- *Build tool:* mvn
- *OS:* Windows

**Description of the issue**:
If this does not exist already, could there be an enhancement to add individual files into an image layer, rather than selecting all the files in a directory.

**Expected behavior**:
```
<extraFiles>
   <extraFile><from>../folder1/context1.xml</from><to>/app/contexts</to></extrafile>
<extraFiles>
```

**Additional Information**: If this is not available, is there an expected version where this might be added?

chanseokoh:
Adds new "includes" and "excludes" config parameters in Maven and Gradle.

Previously, `RawConfiguration` returned a map (for `<from>` --> `<into>`), but since now it requires two more fields, we should return a structure.

Toward #2564 (and #3164).

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -37,6 +37,7 @@ import com.google.cloud.tools.jib.api.buildplan.ModificationTimeProvider;
 import com.google.cloud.tools.jib.api.buildplan.Platform;
 import com.google.cloud.tools.jib.frontend.CredentialRetrieverFactory;
 import com.google.cloud.tools.jib.global.JibSystemProperties;
+import com.google.cloud.tools.jib.plugins.common.RawConfiguration.ExtraDirectoriesConfiguration;
 import com.google.cloud.tools.jib.plugins.common.RawConfiguration.PlatformConfiguration;
 import com.google.cloud.tools.jib.plugins.common.globalconfig.GlobalConfig;
 import com.google.cloud.tools.jib.plugins.extension.JibPluginExtensionException;
@@ -58,7 +59,6 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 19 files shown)


chanseokoh:
Ready for review. This only adds new configuration options.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -429,15 +429,12 @@ static JibContainerBuilder processCommonConfiguration(
         getCreationTime(rawConfiguration.getCreationTime(), projectProperties));
 
     // Adds all the extra files.
-    for (Map.Entry<Path, AbsoluteUnixPath> entry :
-        rawConfiguration.getExtraDirectories().entrySet()) {
-      Path sourceDirectory = entry.getKey();
-      AbsoluteUnixPath targetDirectory = entry.getValue();
-      if (Files.exists(sourceDirectory)) {
+    for (ExtraDirectoriesConfiguration extraDirectory : rawConfiguration.getExtraDirectories()) {
+      if (Files.exists(extraDirectory.getFrom())) {
```
mpeddada1, line 433: nit: we can store `extraDirectory.getFrom()` in a variable?

mpeddada1 (approved):

PR #3173, opened 2021-03-31
<includes> and <excludes> filtering feature for extraDirectories

### LINKED ISSUE #2564: Allow to add single file to a container (with setting rights)

chanseokoh:
On top of #3171. Review it first.

Fixes #2564 (and #3164).

The idea is to allow glob patterns that works similar to how [`maven-resources-plugin`](https://maven.apache.org/plugins/maven-resources-plugin/examples/include-exclude.html) works. This makes adding a single file easy as well.

```xml
         <extraDirectories>
            <paths>
              <path>
                <from>src/main/resources/static</from>
                <into>/app/static</into>
                <includes>foo.txt</includes> <!-- only foo.txt will be copied -->
              </path>
              <path>
                <from>src/main/resources/pictures</from>
                <into>/app/pictures</into>
                <!-- only copy these, but exclude some. -->
                <includes>**/*.bmp,**/*.gif</includes>
                <excludes>
                  <exclude>**/*test*.*</exclude>
                  <exclude>**/*.thumbnail</exclude>
                </excludes>
              </path>
            </paths>
          </extraDirectories>
```

In a follow-up PR, I'll add integration tests for Maven and Gradle and update CHANGELOG.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
@@ -35,6 +35,7 @@ import java.time.Instant;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 
@@ -63,47 +64,60 @@ public class JavaContainerBuilderHelper {
       throws IOException {
     FileEntriesLayer.Builder builder =
         FileEntriesLayer.builder().setName(LayerType.EXTRA_FILES.getName());
-    Map<PathMatcher, FilePermissions> pathMatchers = new LinkedHashMap<>();
+    Map<PathMatcher, FilePermissions> permissionsPathMatchers = new LinkedHashMap<>();
     for (Map.Entry<String, FilePermissions> entry : extraDirectoryPermissions.entrySet()) {
-      pathMatchers.put(
+      permissionsPathMatchers.put(
           FileSystems.getDefault().getPathMatcher("glob:" + entry.getKey()), entry.getValue());
     }
 
     DirectoryWalker walker = new DirectoryWalker(sourceDirectory).filterRoot();
+    // add exclusion filters
     excludes
         .stream()
         .map(pattern -> FileSystems.getDefault().getPathMatcher("glob:" + pattern))
         .forEach(pathMatcher -> walker.filter(path -> !pathMatcher.matches(path)));
+    // add an inclusion filter
     includes
         .stream()
         .map(pattern -> FileSystems.getDefault().getPathMatcher("glob:" + pattern))
         .map(pathMatcher -> (Predicate<Path>) (path -> pathMatcher.matches(path)))
         .reduce((matches1, matches2) -> matches1.or(matches2))
         .ifPresent(walker::filter);
+    // walk the source tree and add layer entries
     walker.walk(
         localPath -> {
           AbsoluteUnixPath pathOnContainer =
               targetDirectory.resolve(sourceDirectory.relativize(localPath));
           Instant modificationTime = modificationTimeProvider.get(localPath, pathOnContainer);
-          FilePermissions permissions = extraDirectoryPermissions.get(pathOnContainer.toString());
-          if (permissions == null) {
-            // Check for matching globs
-            Path containerPath = Paths.get(pathOnContainer.toString());
-            for (Map.Entry<PathMatcher, FilePermissions> entry : pathMatchers.entrySet()) {
-              if (entry.getKey().matches(containerPath)) {
-                builder.addEntry(localPath, pathOnContainer, entry.getValue(), modificationTime);
-                return;
-              }
-            }
-
-            // Add with default permissions
-            builder.addEntry(localPath, pathOnContainer, modificationTime);
+          Optional<FilePermissions> permissions =
+              determinePermissions(
+                  pathOnContainer, extraDirectoryPermissions, permissionsPathMatchers);
+          if (permissions.isPresent()) {
+            builder.addEntry(localPath, pathOnContainer, permissions.get(), modificationTime);
           } else {
-            // Add with explicit permissions
-            builder.addEntry(localPath, pathOnContainer, permissions, modificationTime);
+            builder.addEntry(localPath, pathOnContainer, modificationTime);
           }
         });
     return builder.build();
+  }
+
+  private static Optional<FilePermissions> determinePermissions(
+      AbsoluteUnixPath path,
+      Map<String, FilePermissions> extraDirectoryPermissions,
+      Map<PathMatcher, FilePermissions> permissionsPathMatchers) {
+    // The check is only for optimization. (`permissionsPathMatchers` is constructed from the map.)
+    FilePermissions permissions = extraDirectoryPermissions.get(path.toString());
+    if (permissions != null) {
+      return Optional.of(permissions);
+    }
+
+    // Check for matching globs
+    for (Map.Entry<PathMatcher, FilePermissions> entry : permissionsPathMatchers.entrySet()) {
+      if (entry.getKey().matches(Paths.get(path.toString()))) {
+        return Optional.of(entry.getValue());
+      }
+    }
+    return Optional.empty();
   }
 
   /**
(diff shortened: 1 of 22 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -64,33 +69,40 @@ public static FileEntriesLayer extraDirectoryLayerConfiguration(
...
               }
-            });
+            }
+
+            // Add with default permissions
+            builder.addEntry(localPath, pathOnContainer, modificationTime);
+          } else {
+            // Add with explicit permissions
+            builder.addEntry(localPath, pathOnContainer, permissions, modificationTime);
+          }
```
mpeddada1, line 104:
```suggestion
          } 
            // Add with explicit permissions
            builder.addEntry(localPath, pathOnContainer, permissions, modificationTime);
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -64,33 +69,40 @@ public static FileEntriesLayer extraDirectoryLayerConfiguration(
...
+              targetDirectory.resolve(sourceDirectory.relativize(localPath));
+          Instant modificationTime = modificationTimeProvider.get(localPath, pathOnContainer);
+          FilePermissions permissions = extraDirectoryPermissions.get(pathOnContainer.toString());
+          if (permissions == null) {
+            // Check for matching globs
+            Path containerPath = Paths.get(pathOnContainer.toString());
+            for (Map.Entry<PathMatcher, FilePermissions> entry : pathMatchers.entrySet()) {
+              if (entry.getKey().matches(containerPath)) {
+                builder.addEntry(localPath, pathOnContainer, entry.getValue(), modificationTime);
+                return;
```
mpeddada1, line 95: there seem to be too many levels of nesting here, can we put this in a helper method?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -64,33 +69,40 @@ public static FileEntriesLayer extraDirectoryLayerConfiguration(
...
+        .map(pattern -> FileSystems.getDefault().getPathMatcher("glob:" + pattern))
+        .forEach(pathMatcher -> walker.filter(path -> !pathMatcher.matches(path)));
+    includes
+        .stream()
+        .map(pattern -> FileSystems.getDefault().getPathMatcher("glob:" + pattern))
+        .map(pathMatcher -> (Predicate<Path>) (path -> pathMatcher.matches(path)))
+        .reduce((matches1, matches2) -> matches1.or(matches2))
+        .ifPresent(walker::filter);
+    walker.walk(
+        localPath -> {
```
mpeddada1, line 84: Hm this method is starting to get a little long, can we have line 83 - 105 in a separate method?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
```diff
@@ -64,33 +69,40 @@ public static FileEntriesLayer extraDirectoryLayerConfiguration(
...
+    excludes
+        .stream()
+        .map(pattern -> FileSystems.getDefault().getPathMatcher("glob:" + pattern))
+        .forEach(pathMatcher -> walker.filter(path -> !pathMatcher.matches(path)));
+    includes
+        .stream()
+        .map(pattern -> FileSystems.getDefault().getPathMatcher("glob:" + pattern))
+        .map(pathMatcher -> (Predicate<Path>) (path -> pathMatcher.matches(path)))
+        .reduce((matches1, matches2) -> matches1.or(matches2))
+        .ifPresent(walker::filter);
```
mpeddada1, line 82: I'm probably missing something here, why is the `reduce` step needed here?

chanseokoh: So before `reduce`, we have a list of `Predicate`s where each predicate tests if a given path matches its `<include>` glob pattern (i.e., testing if the file should be included).

Then `reduce` converts a list to a single `Predicate` that is contributed by "or"ing all the predicates. So, this predicate tests if a given path matches any of the `<include>` patterns. And `walker::filter()` accepts a single predicate. Note calling `walker::filter` repeatedly for each `Predicate` before `reduce()` makes them connected with "and", not "or".

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
chanseokoh: The closure makes use of a lot of external variables, and mechanically refactoring out the closure results in a long list of arguments.
```java
  private static void addLayerEntry(
      Path sourceDirectory,
      AbsoluteUnixPath targetDirectory,
      Map<String, FilePermissions> extraDirectoryPermissions,
      ModificationTimeProvider modificationTimeProvider,
      Map<PathMatcher, FilePermissions> permissionPathMatchers,
      FileEntriesLayer.Builder layerBuilder,
      Path localPath) {
```
I realized most of the complication is due to determining permissions and calling `addEntry` based on that. So I refactored out the permissions part.

I also added a few more code comments in this method.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
chanseokoh: Doesn't matter in the refactored method, but to be clear, `else` is needed unless you add `return` in the "true" branch.

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelperTest.java
```diff
@@ -92,17 +80,91 @@ public void testExtraDirectoryLayerConfiguration() throws URISyntaxException, IO
...
+    Path extraFilesDirectory = Paths.get(Resources.getResource("core/layer").toURI());
+    FileEntriesLayer layerConfiguration =
+        JavaContainerBuilderHelper.extraDirectoryLayerConfiguration(
+            extraFilesDirectory,
+            AbsoluteUnixPath.get("/"),
+            Arrays.asList("**/*"),
+            Arrays.asList("**/*"),
+            Collections.emptyMap(),
+            (ignored1, ignored2) -> Instant.EPOCH);
+    assertThat(layerConfiguration.getEntries()).isEmpty();
```
mpeddada1, line 150: ah so excludes takes precedence?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
mpeddada1: Looks good. Putting the permissions logic in a helper helped a lot

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelper.java
mpeddada1: Makes sense. Thanks for the explanation!

mpeddada1 (approved):

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/JavaContainerBuilderHelperTest.java
chanseokoh: Yeah, if something is excluded, then that's absolute and final. I just need to double-check the behavior with the maven-resource-plugin.

chanseokoh: Just verified that the behavior is the same as the Maven resources plugin.

PR #3176, opened 2021-03-31
Add CHANGELOG and integration tests for extra directories filtering 

chanseokoh:
For #2564 and #3164.

### DIFF
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -342,7 +342,8 @@ public class SingleProjectIntegrationTest {
     String output =
         new Command("docker", "run", "--rm", "--entrypoint=ls", targetImage, "-1R", "/extras")
             .run();
-    // No "bar" or "*.txt" files. Specifically,
+
+    // No "bar" or "*.txt" files. Only copies the following:
     //   /extras/cat.json
     //   /extras/foo
     //   /extras/sub/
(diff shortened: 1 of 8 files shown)


jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
```diff
@@ -97,32 +94,29 @@ private static void assertLayerSize(int expected, String imageReference)
...
+    assertThat(dockerInspect)
+        .contains(
             "            \"Volumes\": {\n"
                 + "                \"/var/log\": {},\n"
                 + "                \"/var/log2\": {}\n"
-                + "            },"));
-    MatcherAssert.assertThat(
-        dockerInspect,
-        CoreMatchers.containsString(
+                + "            },");
```
mpeddada1, line 102: optional: could we use regex matching to only test the values that are relevant?

mpeddada1 (commented):
Could we add a test for maven as well?

jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
chanseokoh: I guess that involves a non-trivial amount of work, so maybe another time.

chanseokoh:
Added a Maven integration test too.

I also realized a minor issue with the filtering. Will fix it in a follow-up PR.

jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildDockerMojoIntegrationTest.java
```diff
@@ -99,12 +101,28 @@ public void testExecute_simple()
...
 
+  @Test
+  public void testExecute_simple_extraDirectoriesFiltering()
+      throws DigestException, IOException, InterruptedException, VerificationException {
+    String targetImage = "simpleimage:maven" + System.nanoTime();
+    buildToDockerDaemon(simpleTestProject, targetImage, "pom-extra-dirs-filtering.xml");
+    String output =
+        new Command("docker", "run", "--rm", "--entrypoint=ls", targetImage, "-1R", "/extras")
+            .run();
+    // No "bar" or "*.txt" files. Specifically,
```
mpeddada1, line 118: how about something like this? 
```suggestion

    // No "bar" or "*.txt" files. Only copies the following:
```

mpeddada1 (approved):

PR #3181, opened 2021-04-01
Fix extraDirectories filtering to match relative path

chanseokoh:
Most of the time, you'll specify a relative path in `<includes>` and `<excludes>`, which is the case in all of our tests.

Turns out we need to relativize local path to match against the glob matchers. The reason the tests seemed to work was that we always started with `**/`, which works for absolute path. That is, previously, `includes = ['foo']` didn't work.

---

Previous PR: #3173
Issues: #2564 #3164


mpeddada1 (approved):
LGTM

PR #3182, opened 2021-04-02
Update plugins docs for Jib 3.0

chanseokoh:
Is there anything else we missed?

PR #3189, opened 2021-04-06
Add and update various docs (FAQ, CHANGELOGs, READMEs, ...) for Jib plugins 3.0 release

chanseokoh:
Switching the base image from Distroless is a big part of Jib plugins 3.0. There's a new doc `default_base_image.md`.

Note we have another doc PR #3182.


docs/faq.md
```diff
@@ -756,9 +744,7 @@ There are some common reasons why containers fail on launch.
 
 #### My shell script won't run
  
-The default base image used by Jib, ([`distoless/java`](https://github.com/GoogleContainerTools/distroless/tree/master/java)), does not include a shell, and thus shell scripts won't launch.
-
-Solution: use a different base image with a shell.
+The default base image used by the old Jib Maven and Gradle plugins did not have a shell, and thus shell scripts won't launch. See [Where is bash?](#where-is-bash) for more details.
```
mpeddada1, line 747:
```suggestion
The default base image used by the Jib Maven and Gradle plugins prior to 3.0 did not have a shell, and thus shell scripts won't launch. See [Where is bash?](#where-is-bash) for more details.
```

jib-gradle-plugin/README.md
```diff
@@ -508,11 +508,11 @@ The Jib build plugins have an extension framework that enables anyone to easily
 
 ### WAR Projects
 
-Jib also containerizes WAR projects. If the Gradle project uses the [WAR Plugin](https://docs.gradle.org/current/userguide/war_plugin.html), Jib will by default use the [distroless Jetty](https://github.com/GoogleContainerTools/distroless/tree/master/java/jetty) as a base image to deploy the project WAR. No extra configuration is necessary other than using the WAR Plugin to make Jib build WAR images.
+Jib also containerizes WAR projects. If the Gradle project uses the [WAR Plugin](https://docs.gradle.org/current/userguide/war_plugin.html), Jib will by default use the [jetty](https://hub.docker.com/_/jetty) as a base image to deploy the project WAR. No extra configuration is necessary other than using the WAR Plugin to make Jib build WAR images.
```
mpeddada1, line 511:
```suggestion
Jib also containerizes WAR projects. If the Gradle project uses the [WAR Plugin](https://docs.gradle.org/current/userguide/war_plugin.html), Jib will by default use [jetty](https://hub.docker.com/_/jetty) as a base image to deploy the project WAR. No extra configuration is necessary other than using the WAR Plugin to make Jib build WAR images.
```

docs/default_base_image.md
```diff
@@ -0,0 +1,13 @@
...
+
+Note that this change matters only when you are not specifying any base image via `jib.from.image` (Gradle) or `<from><image>` (Maven), which we have recommended against consistently. For strong reproducibility, you should pin to a specific base image using a digest (or at least a tag). And we want to be clear that Jib's default choice for AdoptOpenJDK does not imply any endorsement to it; you should do your due diligence when making a choice for a base image.
```
mpeddada1, line 7: Hm can we elaborate a bit more on why we recommend against specifying the base image?  It could cause some confusion because we also mention later on that the change doesn't imply endorsement for the new default and users have a choice to pick another base image.

docs/default_base_image.md
```diff
@@ -0,0 +1,13 @@
+# Default Base Images in Jib
```
mpeddada1, line 1: Let's add an entry in the FAQ for this doc?

README.md
```diff
@@ -56,7 +56,7 @@ The [examples](examples) directory includes the following examples (and more).
 
 ## How Jib Works
```
mpeddada1, line 57: Do any of the jib-gradle, jib-maven or jib-core examples mention distroless?

docs/default_base_image.md
chanseokoh: Oh, maybe the opening sentence made you misunderstand. It's that we have consistently recommended specifying a base image (to pin to a specific image), not _against_. I'll try rewording it.

README.md
chanseokoh: Did a `git grep -i distroless`, and no examples explicitly set distroless (although it's likely that many of them just don't set `from.image` at all).

mpeddada1: Nice!

docs/default_base_image.md
mpeddada1: Oh! That makes sense! Sorry, totally my bad. I think the double negatives may have confused me.

chanseokoh:
Just realized we need to clear document how exactly the user can migrate. It's very important. I made significant updates to the docs. That said, I need to test these out before we can finalize the release.

docs/default_base_image.md
```diff
@@ -0,0 +1,56 @@
...
+   }
+   ```
+   ```xml
+     <configuration>
+       <from><image>gcr.io/distroless/java:11</image></from> <!-- or ":8" -->
+     </configuration>
+   ```
+   However, even when you decide to keep using Distroless, at least we strongly recommend `gcr.io/distroless/java-debian10:11`, because, as of Apr 2021, `gcr.io/distroless/java:{8,11}` is based on Debian 9 that reached end-of-life. (Note `gcr.io/distroless/java-debian10` doesn't have `:8`.)
+
+* WAR proejcts
```
mpeddada1, line 33:
```suggestion
* WAR projects
```

docs/faq.md
```diff
@@ -463,8 +457,7 @@ jib.to.image = 'gcr.io/my-gcp-project/my-app:' + System.nanoTime()
 A Dockerfile that performs a Jib-like build is shown below:
 
 ```Dockerfile
-# Jib uses distroless java as the default base image
-FROM gcr.io/distroless/java:latest
+FROM adoptopenjdk:11-jre
```
mpeddada1, line 460: let's keep the comment about the default image

docs/faq.md
```diff
@@ -756,9 +750,7 @@ There are some common reasons why containers fail on launch.
 
 #### My shell script won't run
  
-The default base image used by Jib, ([`distoless/java`](https://github.com/GoogleContainerTools/distroless/tree/master/java)), does not include a shell, and thus shell scripts won't launch.
-
-Solution: use a different base image with a shell.
+The default base image used by the Jib Maven and Gradle plugins prior to 3.0 did not have a shell, and thus shell scripts won't launch. See [Where is bash?](#where-is-bash) for more details.
```
mpeddada1, line 753: maybe something like this?
```suggestion
Jib Maven and Gradle plugins prior to 3.0 used Distroless Java as the default base image, which does not have a shell. See [Where is bash?](#where-is-bash) for more details.
```

mpeddada1: Actually, do we need this FAQ if we have the "Where is bash?" one

mpeddada1 (commented):
LGTM, one minor comment

docs/faq.md
chanseokoh: We can take it out maybe a year later. It'll take a lot of time before the majority of people upgrade to 3.0, so I think it's possible that they'll hit this issue.

mpeddada1 (approved):

PR #3198, opened 2021-04-08
Document options to avoid Docker Hub rate limit

chanseokoh:
I realized this is important to document.

I'll also expand the FAQ to cover the options not explained in this PR.

### DIFF
--- a/docs/default_base_image.md
+++ b/docs/default_base_image.md
@@ -23,7 +23,7 @@ Docker Hub enforces download rate limit. Because Jib pulls base images from Dock
         ]
       }
 ```
-Note that, even after Jib fully cached a base image, Jib still connects to Docker Hub to check the imaage every time it runs (unless you pin to a specific base image using a SHA digest). This is to check if the cached base image is up-to-date.
+Note that, even after Jib fully cached a base image, Jib still connects to Docker Hub to check the image every time it runs (unless you pin to a specific base image using a SHA digest). This is to check if the cached base image is up-to-date.
 
 Some options:
 * Configure a registry mirror.
--- a/docs/default_base_image.md
+++ b/docs/default_base_image.md
@@ -8,6 +8,32 @@ For WAR projects, the default is the official [`jetty`](https://hub.docker.com/_
 
 Note that Jib's default choice for AdoptOpenJDK and Jetty does not imply any endorsement to them. In fact, for strong reproducibility (which also results in better performance and efficiency), we always recommend configuring [`jib.from.image`](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#from-closure) (Gradle) or [`<from><image>`](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#from-object) (Maven) to pin to a specific base image using a digest (or at least a tag). And while doing so, you should do your due diligence to figure out which base image will work best for you.
 
+### Docker Hub Download Rate Limit
+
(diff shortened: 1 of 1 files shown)


docs/default_base_image.md
```diff
@@ -8,6 +8,32 @@ For WAR projects, the default is the official [`jetty`](https://hub.docker.com/_
...
+        "errors": [
+          {
+            "code": "TOOMANYREQUESTS",
+            "message": "You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"
+          }
+        ]
+      }
+```
+Note that, even after Jib fully cached a base image, Jib still connects to Docker Hub to check the imaage every time it runs (unless you pin to a specific base image using a SHA digest). This is to check if the cached base image is up-to-date.
+
```
mpeddada1, line 27:
```suggestion
Note that, even after Jib fully cached a base image, Jib still connects to Docker Hub to check the image every time it runs (unless you pin to a specific base image using a SHA digest). This is to check if the cached base image is up-to-date.

```

docs/default_base_image.md
```diff
@@ -8,6 +8,32 @@ For WAR projects, the default is the official [`jetty`](https://hub.docker.com/_
...
+Some options:
+* Configure a registry mirror.
```
mpeddada1, line 29: let's add a link to the faq here

docs/default_base_image.md
```diff
@@ -8,6 +8,32 @@ For WAR projects, the default is the official [`jetty`](https://hub.docker.com/_
...
+* Prevent Jib from accessing Docker Hub (after Jib cached a base image locally).
+  * Pin to a specific base image using a SHA digest (for example, `jib.from.image=adoptopenjdk:11-jre@sha256:...`).
+  * Do offline building.
```
mpeddada1, line 32:
```suggestion
  * Do offline building using `--offline`.
```

docs/default_base_image.md
```diff
@@ -8,6 +8,32 @@ For WAR projects, the default is the official [`jetty`](https://hub.docker.com/_
...
+* Retry with increasing backoffs.
+
+See this [FAQ](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#i-am-hitting-docker-hub-rate-limits-how-can-i-configure-registry-mirrors) for more details about the options.
```
mpeddada1, line 35: This FAQ only talks about registry mirrors. We may not needed if we mention it above?

chanseokoh:
I will expand the FAQ entry (as I said in the PR description). I think it's worth covering all the options in both places but don't have to duplicate the contents. We can briefly mention the options here and cover others in the FAQ.

docs/default_base_image.md
chanseokoh: See https://github.com/GoogleContainerTools/jib/pull/3198#issuecomment-816715782

docs/default_base_image.md
chanseokoh: I intentionally don't want to mention `--offline` without explaining more details, as the user can easily run into a problem. That is, see https://github.com/GoogleContainerTools/jib/pull/3198#issuecomment-816715782

docs/default_base_image.md
mpeddada1: Oops my bad, I didn't see the comment above.

docs/default_base_image.md
mpeddada1: Okay, that's a valid point.

mpeddada1 (approved):

Issue #3204, opened 2021-04-12
Changing base jetty image for WAR: running image fails


aanno:
**Environment**:

- *Jib version:* 3.0.0
- *Build tool:*  mvn 3.6.3
- *OS:* Linux (Fedora 33)

**Description of the issue**:
Changing base jetty image for WAR: running image fails

**Expected behavior**:
Changing base jetty image for WAR: running image should run WAR on jetty

**Steps to reproduce**:

**`jib-maven-plugin` Configuration**: 
```xml
			<plugin>
				<groupId>com.google.cloud.tools</groupId>
				<artifactId>jib-maven-plugin</artifactId>
				<version>3.0.0</version>
				<configuration>
					<from>
						<image>jetty:11.0.2-jre11</image>
					</from>
					<to>
						<image>matching-server-war-jetty</image>
					</to>
				</configuration>
				<executions>
					<execution>
						<phase>package</phase>
						<goals>
							<goal>build</goal>
						</goals>
					</execution>
				</executions>
			</plugin>
```

When using the maven configuration above:
```bash
$ podman run --rm -p 8080:8080 matching-server-war-jetty:latest 2>&1 | tee log
/docker-entrypoint.sh: 98: /docker-entrypoint.sh: cannot create /var/lib/jetty/jetty.start: Permission denied
```
When using the maven configuration above _without the **from** part_, everything works as expected.

chanseokoh:
Hi @aanno,

I know what's happening. The way we make the default `jetty` image (when you are not specifying `<from>`) works is to set the entrypoint to `java -jar /usr/local/jetty/start.jar`. 

(from #3124)

> For WAR projects, we no longer inherits the entrypoint from the base image. The new default entrypoint is explicitly set to `java -jar /usr/local/jetty/start.jar`. The idea is from https://github.com/GoogleContainerTools/jib/issues/1270#issuecomment-791525157.

So I expect you should be able to resolve the issue by specifying
```xml
<configuration>
  ...
  <container>
    <entrypoint>java,-jar,/usr/local/jetty/start.jar</entrypoint>
  </container>
</configuration>
```

---

Maybe we can help this situation by setting this entrypoint automatically if the base image is from the official `jetty` repo and no entrypoint is given? Or, maybe we can just document how to configure this [here](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin#war-projects)?

chanseokoh:
[...]

PR #3205, opened 2021-04-12
Document WAR containerization with jetty

chanseokoh:
Helps with #3204.


chanseokoh:
Ah, wait. I am running into some problem with this example.

mpeddada1 (approved):

PR #3225, opened 2021-04-30
WIP: Initial attempt at fixing final field access

Sineaggi:
This is a tentative solution to an issue found when using `native-image` to build binaries that depend on jib.

Currently during deser jackson attempts to set private final fields like `mediaType` or `schemaVersion`.
```
	... 8 more
Caused by: java.lang.IllegalAccessException: Cannot set final field: com.google.cloud.tools.jib.image.json.V22ManifestListTemplate.mediaType. Enable by specifying "allowWrite" for this field in the reflection configuration.
	at java.lang.reflect.Field.set(Field.java:780)
	at com.fasterxml.jackson.databind.deser.impl.FieldProperty.deserializeAndSet(FieldProperty.java:150)
```

This fix assumes that setting private final fields reflectively is something to be avoided, and adds a `@JsonConstructor` annotation to help jackson out.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
@@ -16,12 +16,15 @@
 
 package com.google.cloud.tools.jib.image.json;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.cloud.tools.jib.json.JsonTemplate;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
```diff
@@ -72,6 +75,16 @@
   private final int schemaVersion = SCHEMA_VERSION;
   private final String mediaType = MANIFEST_MEDIA_TYPE;
 
+  @JsonCreator
+  public V22ManifestListTemplate(@JsonProperty("mediaType") String mediaType, @JsonProperty("schemaVersion") Integer schemaVersion, @JsonProperty("manifests") List<ManifestDescriptorTemplate> manifests) {
+    Preconditions.checkArgument(Objects.equals(mediaType, MANIFEST_MEDIA_TYPE), "mediaType " + mediaType + " does not equal " + MANIFEST_MEDIA_TYPE);
+    Preconditions.checkArgument(Objects.equals(schemaVersion, SCHEMA_VERSION), "schemaVersion " + schemaVersion + " does not equal " + SCHEMA_VERSION);
+    this.manifests = manifests;
+  }
```
chanseokoh, line 83: I guess the behavior is different in that, previously, when deserializing a JSON that stores a value different from `SCHEMA_VERSION`, the code does not fail and the Java object can always force the value `SCHEMA_VERSION` defined in this Java code. OTOH, I think the new code will crash at runtime. (However, I admit that it's extremely unlikely that a serialized JSON has different constants than what we expect.)

chanseokoh:
Thanks a lot for trying native image for Jib! I guess you're compiling your application that depends on the jib-core library, right? If possible, keep us posted about your progress. We are very interested in this. We tried native image for Jib CLI a while ago, but we had limited success. It only works with a serial build (`--serialize`) and crashes at runtime with the dreaded SegFault. (I'm suspecting a GraalVM bug).

---

I can see the changes in this PR can be a way to side-step the annoying failure when generating a native image, and I do think this approach does have some values considering the difficulty and frictions with the native image generation.

At the same time, we don't really need changes in the Java code to enable native image. As the error says, what you need to do is enable "allowWrite" for the field: https://github.com/oracle/graal/issues/2994 For example,
```diff
   "name":"com.example.JacksonIssueDemo$JsonTemplate",
   "allDeclaredFields":true,
   "allDeclaredMethods":true,
-  "allDeclaredConstructors":true
+  "allDeclaredConstructors":true,
+  "fields":[
+    {"name":"finalValue", "allowWrite": true}
+  ]
 },
```

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/V22ManifestListTemplate.java
Sineaggi: I'm not super in love with this either, we could always just log if the mediaType or schemaVersion aren't what we expect.

PR #3226, opened 2021-04-30
[Jib CLI WAR] Implement exploded mode for standard WAR

mpeddada1:
Major Changes:
- Implemented StandardWarExplodedProcessor
- Renamed JarProcessor to ArtifactProcessor (is used by both jar processors and war processors)
- Renamed JarProcessors to ArtifactProcessors. It will eventually have two methods, `fromJar`(already implemented) and `fromWar`, which will contain logic to select mode processors for JARs and WARs respectively. 
- Moved common code from JarLayers to ArtifactLayers (to be used by both jar and war processors)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessor.java
```diff
@@ -14,15 +14,15 @@
...
+package com.google.cloud.tools.jib.cli;
 
 import com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer;
 import com.google.common.collect.ImmutableList;
 import java.io.IOException;
 import java.util.List;
 
-/** Interface to create layers and compute entrypoint from jar file contents. */
-public interface JarProcessor {
+/** Interface to create layers and compute entrypoint from JAR or WAR file contents. */
```
chanseokoh, line 24: Update the Javadocs of the methods too. They say only "jar."

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactLayers.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.cloud.tools.jib.api.buildplan.AbsoluteUnixPath;
+import com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer;
+import com.google.cloud.tools.jib.filesystem.DirectoryWalker;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.function.Predicate;
+
+public class ArtifactLayers {
+
+  public static final String CLASSES = "classes";
```
chanseokoh, line 28: Add a comment?
```
// Used for layer names.
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/StandardWarExplodedProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+  private final Integer warJavaVersion;
+  private final AbsoluteUnixPath appRoot;
+
+  /**
+   * Constructor for {@link StandardWarExplodedProcessor}.
+   *
+   * @param warPath path to WAR file
+   * @param targetExplodedWarRoot path to exploded-war root
+   * @param warJavaVersion war java version
+   * @param appRoot the absolute path of the app on the container
```
chanseokoh, line 47: We don't provide the `--appRoot` option yet, right?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/StandardWarExplodedProcessor.java
```diff
@@ -0,0 +1,132 @@
...
+    Predicate<Path> isClass = path -> path.getFileName().toString().endsWith(".class");
+    Predicate<Path> isInWebInfClasses =
+        path -> path.startsWith(targetExplodedWarRoot.resolve("WEB-INF").resolve("classes"));
+    Predicate<Path> classesPredicate = isInWebInfClasses.and(isClass);
+    FileEntriesLayer classesLayer =
+        ArtifactLayers.getDirectoryContentsAsLayer(
+            ArtifactLayers.CLASSES, targetExplodedWarRoot, classesPredicate, appRoot);
+
+    // Resources layer.
+    Predicate<Path> resourcesPredicate = isInWebInfLib.or(isClass).negate();
```
chanseokoh, line 97: For consistency, maybe this?
```suggestion
    Predicate<Path> resourcesPredicate = isInWebInfLib.or(isClass).negate().and(isFile);
```
Or the other way around: maybe we consistently apply `isFile.and(...)` in the arguments to `getDirectoryContentsAsLayer` in all cases.

chanseokoh (commented):
Looks good to me.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessor.java
mpeddada1: Good catch.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/StandardWarExplodedProcessor.java
mpeddada1: Ah good question. Nope, we don't provide the `--appRoot` option yet, it will be added when `War.class` is implemented. This class will only accept appRoot as an input but the idea is to set it in `ArtifactProcessors`.  If a base image is specified then the user is expected to provide the `appRoot` (otherwise we throw an exception). If a base image is not specified then we create `StandardWarExplodedProcessor` with the default jetty app root.

chanseokoh: Sounds like a good plan.

> If a base image is specified then the user is expected to provide the `appRoot` (otherwise we throw an exception).

Maybe throw an exception only when it's not the `jetty` on Docker Hub. I imagine often you set a base image only to use a tag or digest for `jetty`, which was the case in #3204.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/StandardWarExplodedProcessor.java
```diff
@@ -0,0 +1,135 @@
...
+
+    ZipUtil.unzip(warPath, targetExplodedWarRoot, true);
+    Predicate<Path> isFile = Files::isRegularFile;
+    Predicate<Path> isInWebInfLib =
+        path -> path.startsWith(targetExplodedWarRoot.resolve("WEB-INF").resolve("lib"));
+    Predicate<Path> isSnapshot = path -> path.getFileName().toString().contains("SNAPSHOT");
+
+    // Non-snapshot layer
+    Predicate<Path> isInWebInfLibAndIsNotSnapshot = isInWebInfLib.and(isSnapshot.negate());
+    Predicate<Path> nonSnapshotPredicate = isInWebInfLibAndIsNotSnapshot;
```
chanseokoh, line 72: At this point, I think we should just remove the variable `nonSnapshotPredicate`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/StandardWarExplodedProcessor.java
```diff
@@ -0,0 +1,135 @@
...
+    FileEntriesLayer nonSnapshotLayer =
+        ArtifactLayers.getDirectoryContentsAsLayer(
+            ArtifactLayers.DEPENDENCIES,
+            targetExplodedWarRoot,
+            isFile.and(nonSnapshotPredicate),
+            appRoot);
+
+    // Snapshot layer
+    Predicate<Path> isInWebInfLibAndIsSnapshot = isInWebInfLib.and(isSnapshot);
+    Predicate<Path> snapshotPredicate = isInWebInfLibAndIsSnapshot;
```
chanseokoh, line 82: ditto

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/StandardWarExplodedProcessor.java
mpeddada1: Yes, that is a great point! Will have to keep this in mind when implementing `ArtifactProcessors`

chanseokoh (approved):

PR #3237, opened 2021-05-14
Fix good first issue filter in readme

NaitYoussef:
Current filter on good first issue show no open issues the filter uses "kind/good first issue" but "good first issue" is used to create issues


NaitYoussef:
@googlebot I signed it!

chanseokoh:
@NaitYoussef niceeee, thanks for fixing this!

Issue #3094, opened 2021-03-01
Convert Gradle config parameters (such as jib.container.labels) to use lazy evaluation


remy-tiitre:
Just like there was an issue to make the jib.to.tags and jib.to.auth to be lazyly evaluated I would need to have the container.labels do the same. Even the reason is the same. I use com.gorylenko.gradle-git-properties Gradle plugin and would like to reuse its properties to add labels to my images. Something like that:
```
jib {
    container {
        labels = [
            'org.label-schema.vcs-url': project.ext.gitProps['git.remote.origin.url'],
            'org.label-schema.vcs-branch': project.ext.gitProps['git.branch'],
            'org.label-schema.vcs-ref': project.ext.gitProps['git.commit.id.abbrev']
        ]
    }
}
```
In our case I would make that configuration global and add this snipped to our custom gradle wrapper init script.

I did try something like this:
```
jib
    container {
        labels = [
            ...
            'org.label-schema.vcs-ref': "${-> project.ext.gitProps['git.commit.id.abbrev']}"
        ]
    }
}
```
But this errored out with this message:
`> class org.codehaus.groovy.runtime.GStringImpl cannot be cast to class java.lang.String (org.codehaus.groovy.runtime.GStringImpl is in unnamed module of loader org.gradle.internal.classloader.VisitableURLClassLoader @47c62251; java.lang.String is in module java.base of loader 'bootstrap')`

chanseokoh:
Thanks for your feedback! Given the use case, I think making labels (among other parameters) lazily configurable will be very useful.

We also welcome contributions.

Technical notes:
- Going forward, we decided to expose `Property<?>` in [this particular way](https://github.com/GoogleContainerTools/jib/pull/3034#discussion_r581438469) for new parameters. And perhaps this is not possible with existing parameters for compatibility reasons. (Could be revamped in a major version change.)
- Previously, we simply [added a setter for `Property<?>`](https://github.com/GoogleContainerTools/jib/pull/2906/files) for existing parameters.

chanseokoh:
As an immediate workaround, I wonder if setting `jib.container.labels` in a `project.afterEvalaute` block can work for the `com.gorylenko.gradle-git-properties` plugin, like [in this way](https://stackoverflow.com/a/59631660/1701388).

NaitYoussef:
Hello would like to help with this issue

chanseokoh:
@NaitYoussef thanks for your interest!

[...]

PR #3242, opened 2021-05-17
Use Gradle Property and Provider to enable lazy evaluation for jib.container.labels (#3094)

NaitYoussef:
Use Gradle Property and Provider to enable lazy evaluation for jib.container.labels (#3094)

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
@@ -24,6 +24,9 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
+import javax.inject.Inject;
+import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.MapProperty;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Optional;
 
@@ -43,12 +46,17 @@ public class ContainerParameters {
   private ImageFormat format = ImageFormat.Docker;
   private List<String> ports = Collections.emptyList();
   private List<String> volumes = Collections.emptyList();
-  private Map<String, String> labels = Collections.emptyMap();
+  private MapProperty<String, String> labels;
   private String appRoot = "";
   @Nullable private String user;
   @Nullable private String workingDirectory;
   private String filesModificationTime = "EPOCH_PLUS_SECOND";
   private String creationTime = "EPOCH";
+
+  @Inject
+  public ContainerParameters(ObjectFactory objectFactory) {
+    labels = objectFactory.mapProperty(String.class, String.class).empty();
+  }
 
   @Input
   @Nullable
@@ -195,16 +203,16 @@ public class ContainerParameters {
 
   @Input
   @Optional
-  public Map<String, String> getLabels() {
-    if (System.getProperty(PropertyNames.CONTAINER_LABELS) != null) {
-      return ConfigurationPropertyValidator.parseMapProperty(
-          System.getProperty(PropertyNames.CONTAINER_LABELS));
+  public MapProperty<String, String> getLabels() {
+    String labelsProperty = System.getProperty(PropertyNames.CONTAINER_LABELS);
+    if (labelsProperty != null) {
+      Map<String, String> parsedLabels =
+          ConfigurationPropertyValidator.parseMapProperty(labelsProperty);
+      if (!parsedLabels.equals(labels.get())) {
+        labels.set(parsedLabels);
+      }
     }
     return labels;
-  }
-
-  public void setLabels(Map<String, String> labels) {
-    this.labels = labels;
   }
 
   @Input
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleRawConfigurationTest.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleRawConfigurationTest.java
@@ -21,14 +21,22 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Sets;
 import java.nio.file.Paths;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import org.gradle.api.provider.MapProperty;
 import org.junit.Assert;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
 import org.mockito.Mockito;
+import org.mockito.junit.MockitoJUnitRunner;
 
 /** Test for {@link GradleRawConfiguration}. */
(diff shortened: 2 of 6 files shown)


chanseokoh:
Thanks! There's a formatting violation that's causing unit-tests failures. You can auto-format the code with `./gradlew googleJavaFormat`.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -200,11 +209,15 @@ public void setVolumes(List<String> volumes) {
...
+    return labels.get();
   }
 
   public void setLabels(Map<String, String> labels) {
-    this.labels = labels;
+    this.labels.set(labels);
+  }
+
+  public void setLabels(Provider<Map<String, String>> labels) {
+    this.labels.set(labels);
```
chanseokoh, line 220: We should do this in the new way (e.g., like [this](https://github.com/GoogleContainerTools/jib/pull/3034/files#diff-ec1f44085899ade105be7cc3c4febe441db808c45bd067be6827bd19c2f52958R238)), as explained in #3094. That is, no more setters, and `getLabels()` exposes the `MapProperty`. (This assumes the existing usage of directly assigning a map will continue to work, which I think will be the case.)

NaitYoussef: Thank you @chanseokoh, ill check it out

NaitYoussef: I can refactor jib.to.image and jib.to.tags as well to use the new way

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -195,18 +203,16 @@ public void setVolumes(List<String> volumes) {
...
-  public Map<String, String> getLabels() {
-    if (System.getProperty(PropertyNames.CONTAINER_LABELS) != null) {
-      return ConfigurationPropertyValidator.parseMapProperty(
-          System.getProperty(PropertyNames.CONTAINER_LABELS));
+  public MapProperty<String, String> getLabels() {
+    String labelsProperty = System.getProperty(PropertyNames.CONTAINER_LABELS);
+    if (labelsProperty != null) {
+      Map<String, String> parsedLabels =
+          ConfigurationPropertyValidator.parseMapProperty(labelsProperty);
+      labels.set(parsedLabels);
```
chanseokoh, line 211: Sorry, this was missing in #3094. (I updated the issue to reflect this information.) We should call `.set()` only when the values differ. Using `.equals()` will do.
```suggestion
      if (!parsedLabels.equals(labels.get()) {
        labels.set(parsedLabels);
      }
```

chanseokoh (commented):
Thanks again!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
chanseokoh: Thanks for asking! Since we think this new technique isn't 100% backward-compatible, we want to have this change only for a single config for the next release.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
```diff
@@ -156,7 +155,6 @@ public void testContainer() {
           container.setMainClass("mainClass");
           container.setArgs(Arrays.asList("arg1", "arg2", "arg3"));
           container.setPorts(Arrays.asList("1000", "2000-2010", "3000"));
-          container.setLabels(ImmutableMap.of("label1", "value1", "label2", "value2"));
```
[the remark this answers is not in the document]
NaitYoussef: I can add it, but I think then we will be testing the MapProperty behaviour and not our code

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
```diff
@@ -140,7 +140,6 @@ public void testContainer() {
     assertThat(testJibExtension.getContainer().getArgs()).isNull();
     assertThat(testJibExtension.getContainer().getFormat()).isSameInstanceAs(ImageFormat.Docker);
     assertThat(testJibExtension.getContainer().getPorts()).isEmpty();
-    assertThat(testJibExtension.getContainer().getLabels()).isEmpty();
```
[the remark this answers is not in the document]
NaitYoussef: OK

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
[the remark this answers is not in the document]
chanseokoh: Hmm... good point. Let's remove this and the `assertThat()` below.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
NaitYoussef: I did it this way intentionally for performance reasons, thought doing an affectation even when both maps are equal is more efficient that comparing them and do affectation only when they're different,  what do you think

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleRawConfigurationTest.java
```diff
@@ -23,6 +23,7 @@
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import org.gradle.api.internal.provider.DefaultMapProperty;
```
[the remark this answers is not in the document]
NaitYoussef: I didn't pay attention to that thanks

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
chanseokoh: Sorry, I should have been clearer. This is not about performance but to avoid a potential error: https://github.com/GoogleContainerTools/jib/pull/3034#issuecomment-792838731

NaitYoussef: Ah Ok my bad

chanseokoh (approved):
Awesome, thanks for your contribution! Merging this now.

NaitYoussef:
Thank you @chanseokoh  for your guidance. Looking forward to help with other issues !

PR #3248, opened 2021-05-24
Implement class to create JibContainerBuilder from the contents of a WAR file

mpeddada1:
Implement class to create JibContainerBuilder from an ArtifactProcessor.
Note that this class doesn't include war command-line options yet.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
@@ -32,7 +32,7 @@ public class WarFiles {
    *
    * @param processor artifact processor
    * @return JibContainerBuilder
-   * @throws IOException if I/O error occurs when opening the jar file or if temporary directory
+   * @throws IOException if I/O error occurs when opening the war file or if temporary directory
    *     provided doesn't exist
    * @throws InvalidImageReferenceException if the base image reference is invalid
    */
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2021 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
 package com.google.cloud.tools.jib.cli.war;
 
 import com.google.cloud.tools.jib.api.InvalidImageReferenceException;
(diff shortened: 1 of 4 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -0,0 +1,50 @@
...
+  public static JibContainerBuilder toJibContainerBuilder(ArtifactProcessor processor)
+      throws IOException, InvalidImageReferenceException {
+    JibContainerBuilder containerBuilder = Jib.from("jetty");
+    List<FileEntriesLayer> layers = processor.createLayers();
+
+    // JVM Flags are ignored
+    List<String> entrypoint = processor.computeEntrypoint(ImmutableList.of());
+
+    containerBuilder.setEntrypoint(entrypoint).setFileEntriesLayers(layers);
+    return containerBuilder;
```
chanseokoh, line 48: Is this going to be expanded a lot? (I guess yes.) Otherwise, this could be
```java
return Jib.from("jetty")
    .setEntrypoint(processor....)
    .setFileEntriesLayers(processor.createLayers());
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
```diff
@@ -0,0 +1,75 @@
...
+      throws IOException, InvalidImageReferenceException {
+    FileEntriesLayer layer =
+        FileEntriesLayer.builder()
+            .setName("classes")
+            .addEntry(
+                Paths.get("path/to/tempDirectory/class1.class"),
+                AbsoluteUnixPath.get("/my/app/class1.class"))
+            .build();
+    when(mockStandardWarExplodedProcessor.createLayers()).thenReturn(Arrays.asList(layer));
+    when(mockStandardWarExplodedProcessor.computeEntrypoint(ArgumentMatchers.anyList()))
```
chanseokoh, line 54:
```suggestion
    when(mockStandardWarExplodedProcessor.computeEntrypoint(anyList()))
```

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
```diff
@@ -0,0 +1,75 @@
...
+        .thenReturn(ImmutableList.of("java", "-jar", "/usr/local/jetty/start.jar"));
```
chanseokoh, line 55: Maybe just for consistency, `Arrays.asList()` instead of `ImmutableList.of()`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    JibContainerBuilder containerBuilder =
+        WarFiles.toJibContainerBuilder(mockStandardWarExplodedProcessor);
+    ContainerBuildPlan buildPlan = containerBuilder.toContainerBuildPlan();
+
+    assertThat(buildPlan.getBaseImage()).isEqualTo("jetty");
+    assertThat(buildPlan.getEntrypoint())
+        .isEqualTo(ImmutableList.of("java", "-jar", "/usr/local/jetty/start.jar"));
```
chanseokoh, line 63: Why not `.containersExactly(...).inOrder()`?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
```diff
@@ -0,0 +1,75 @@
...
+    assertThat(buildPlan.getLayers().size()).isEqualTo(1);
```
chanseokoh, line 64: Doesn't it have `assertThat(...).isSize(1)` or something like that?

chanseokoh (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Thanks for the suggestion, I cleaned this up a little. Yeah this method will be expanded to take the user-specified base image and entrypoint into account. One example of the logic to be included: If the user specifies a base image then we use the entrypoint specified through the cli (which will take precedence) otherwise we inherit it from the base image.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
mpeddada1: Nice!

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
mpeddada1: Good point, I also applied this change to `JarFilesTest`.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
mpeddada1: Done

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
mpeddada1: Ah `computeEntrypoint` returns an `ImmutableList` so changing this to `Arrays.asList()` results in an error.

PR #3249, opened 2021-05-24
[Refactoring] Create class for common cli options that will be shared between the Jar and War command

mpeddada1:
Review #3248 first.

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
@@ -31,7 +31,7 @@ import java.util.stream.Collectors;
 import picocli.CommandLine;
 
 /** Common command line options shared between jar and war command. */
-public class CommonArtifactCommandOptions {
+public class CommonContainerConfigCliOptions {
 
   @CommandLine.Option(
       names = "--from",
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
@@ -20,10 +20,6 @@ import com.google.cloud.tools.jib.api.Containerizer;
 import com.google.cloud.tools.jib.api.JibContainer;
 import com.google.cloud.tools.jib.api.JibContainerBuilder;
 import com.google.cloud.tools.jib.api.LogEvent;
-import com.google.cloud.tools.jib.api.Ports;
-import com.google.cloud.tools.jib.api.buildplan.AbsoluteUnixPath;
-import com.google.cloud.tools.jib.api.buildplan.ImageFormat;
-import com.google.cloud.tools.jib.api.buildplan.Port;
 import com.google.cloud.tools.jib.cli.jar.ArtifactProcessors;
 import com.google.cloud.tools.jib.cli.jar.JarFiles;
 import com.google.cloud.tools.jib.cli.jar.ProcessingMode;
@@ -32,21 +28,16 @@ import com.google.cloud.tools.jib.plugins.common.globalconfig.GlobalConfig;
 import com.google.cloud.tools.jib.plugins.common.logging.ConsoleLogger;
 import com.google.cloud.tools.jib.plugins.common.logging.SingleThreadedExecutor;
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimaps;
 import com.google.common.util.concurrent.Futures;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.Duration;
-import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Future;
-import java.util.stream.Collectors;
 import picocli.CommandLine;
 import picocli.CommandLine.Model.CommandSpec;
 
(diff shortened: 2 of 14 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
```diff
@@ -195,15 +114,17 @@ public Integer call() {
...
 
       CacheDirectories cacheDirectories =
           CacheDirectories.from(commonCliOptions, jarFile.toAbsolutePath().getParent());
-      ArtifactProcessor processor = ArtifactProcessors.from(jarFile, cacheDirectories, this);
+      ArtifactProcessor processor =
+          ArtifactProcessors.from(jarFile, cacheDirectories, this, commonContainerConfigCliOptions);
       JibContainerBuilder containerBuilder =
-          JarFiles.toJibContainerBuilder(processor, this, commonCliOptions, logger);
+          JarFiles.toJibContainerBuilder(
+              processor, this, commonCliOptions, commonContainerConfigCliOptions, logger);
```
chanseokoh, line 127: Just got this idea: both `commonCliOptions` and `commonContainerConfigCliOptions` are a field of `this` (the `Jar` class). Maybe `Jar` can have getters for that, and then you only need to pass `this`?

But not sure if this is a bad or good idea.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -0,0 +1,50 @@
...
+   *     provided doesn't exist
+   * @throws InvalidImageReferenceException if the base image reference is invalid
+   */
+  public static JibContainerBuilder toJibContainerBuilder(ArtifactProcessor processor)
+      throws IOException, InvalidImageReferenceException {
+    JibContainerBuilder containerBuilder = Jib.from("jetty");
+    List<FileEntriesLayer> layers = processor.createLayers();
+
+    // JVM Flags are ignored
+    List<String> entrypoint = processor.computeEntrypoint(ImmutableList.of());
```
chanseokoh, line 45: Wait, why doesn't `ArtifactProccesor.computeEntrypoint()` accept `--program-args`? How is `--program-args` being used currently?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
```diff
@@ -0,0 +1,179 @@
...
+      names = {"--image-format"},
+      defaultValue = "Docker",
+      paramLabel = "<image-format>",
+      description =
+          "Format of container, candidates: ${COMPLETION-CANDIDATES}, default: ${DEFAULT-VALUE}.")
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  private ImageFormat format;
+
+  @CommandLine.Option(
+      names = "--program-args",
```
chanseokoh, line 91: Maybe this shouldn't be included, which has a similar nature to `--jvm-args` in that it's a factor in computing a final entrypoint?

chanseokoh (commented):
Technically `--mode packaged` is doable (but I guess this is a low priority that I don't know we will work on).

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
mpeddada1: Ah that is an interesting idea and would certainly help us avoid passing in another parameter. But when `War.class` is implemented, it would also have to include those same getters so I don't know if we want duplicate that code?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: The program arguments are currently set using `containerBuilder.setProgramArguments()`. But now that I think about it, maybe `computeEntrypoint` in `ArtifactProcessor` shouldn't take in `jvm-flags` as an input. It isn't used in the `War` case at the moment and can cause some confusion in the future.

mpeddada1: I was thinking we could add `program-args` and `jvm-flags` as parameters to the constructors of the jar processors? to avoid having unused params.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
mpeddada1: Oh good catch.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
chanseokoh: Ah, I forgot that `programArguments` corresponds to `CMD` in `Dockerfile`, which is different from `ENTRYPOINT`. So maybe it's okay that `computeEntrypoint` doesn't accept `programArguments`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
chanseokoh: Ah, I forgot that `programArguments` corresponds to `CMD` in `Dockerfile`, which is different from `ENTRYPOINT`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Jar.java
chanseokoh: I think getters will be 1-liner. But then, I just realized `War` and `Jar` should probably implement a common interface if we want others to call those getters. Maybe let's just go with the current approach?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Ah right. Out of curiosity - since program arguments are added at the end of the entrypoint anyway, both could work right? (i.e. using `setProgramArguments` and making `computeEntrypoint()` accept `programArguments`)

chanseokoh: I think using `setProgramArguments` is right. Making `computeEntrypoint()` accept it doesn't make sense to me, since `ENTRYPOINT` and `CMD` are two different properties. It's just that I got confused earlier.

mpeddada1:
> Technically `--mode packaged` is doable (but I guess this is a low priority that I don't know we will work on).

Agreed. Packaged mode is certainly doable but for now we will be limiting our scope to implementing what is currently supported by the Jib plugins (i.e. exploded mode).

chanseokoh (approved):

PR #3285, opened 2021-06-08
Implement War command 

mpeddada1:
Implements the `war` command with integration testing.
Few things to note:
- The war command includes the `--app-root` flag to allow the app root of the container to be set. This is especially useful when a custom base image is specified.
- If a custom base image is used then the entrypoint will be inherited from the base image. However, if the user specifies an entrypoint then the user-specified entrypoint will take precedence.


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessor.java
```diff
@@ -42,5 +42,7 @@
    */
   ImmutableList<String> computeEntrypoint(List<String> jvmFlags) throws IOException;
 
-  Integer getJavaVersion();
+  default Integer getJavaVersion() {
```
chanseokoh, line 45: Do we have to do this? We can have classes implement this explicitly.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessors.java
```diff
@@ -14,33 +14,39 @@
...
 import java.net.URLClassLoader;
 import java.nio.file.Path;
 import java.util.Enumeration;
+import java.util.Optional;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 
 /**
- * Class to create a {@link ArtifactProcessor} instance depending on jar type and processing mode.
+ * Class to create a {@link com.google.cloud.tools.jib.cli.ArtifactProcessor} instance depending on
```
chanseokoh, line 38: Why the fully qualified name?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
```diff
@@ -0,0 +1,139 @@
...
+  @VisibleForTesting
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  CommonContainerConfigCliOptions commonContainerConfigCliOptions;
+
+  @CommandLine.Parameters(description = "The path to the war file (ex: path/to/my-war.war)")
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  private Path warFile;
+
+  @CommandLine.Option(
+      names = "--app-root",
```
chanseokoh, line 65: I think this option should be common to all the commands (`build`, `jar`, and `war`).

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
```diff
@@ -0,0 +1,139 @@
...
+      paramLabel = "<app root>",
+      description = "The app root on the container")
+  @SuppressWarnings("NullAway.Init") // initialized by picocli
+  private Path appRoot;
+
+  @Override
+  public Integer call() {
```
chanseokoh, line 72: Hmm... just asking, can there be a way to remove duplicate code between war and jar?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -19,30 +19,75 @@
...
+      ConsoleLogger logger)
       throws IOException, InvalidImageReferenceException {
-    JibContainerBuilder containerBuilder = Jib.from("jetty");
+    JibContainerBuilder containerBuilder;
+    List<FileEntriesLayer> layers;
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    if (baseImage.isPresent()) {
+      containerBuilder =
+          ContainerBuilders.create(
+              commonContainerConfigCliOptions.getFrom().get(),
```
chanseokoh, line 60: `baseImage.get()`

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -19,30 +19,75 @@
...
+              Collections.emptySet(),
+              commonCliOptions,
+              logger);
+    } else {
+      containerBuilder = Jib.from("jetty");
```
chanseokoh, line 65: This makes me wonder why the other branch can't juse be `Jib.from(baseImage.get())` or why this branch doesn't use `ContainerBuilders.create("jetty", ...)`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -19,30 +19,75 @@
...
-    List<String> entrypoint = processor.computeEntrypoint(ImmutableList.of());
+    List<String> entrypoint;
+    if (baseImage.isPresent() && !baseImage.get().startsWith("jetty")) {
+      List<String> entrypointFromCli = commonContainerConfigCliOptions.getEntrypoint();
+      entrypoint =
+          entrypointFromCli.isEmpty()
+              ? null // inherit from base image
+              : entrypointFromCli;
+    } else {
+      entrypoint = processor.computeEntrypoint(ImmutableList.of());
```
chanseokoh, line 76: Aha, maybe this is explicitly telling jib-core to set an empty entrypoint. I remember the jib-core API distinguishes between null (i.e., inherit) and an empty list.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -19,30 +19,75 @@
...
+              logger);
+    } else {
+      containerBuilder = Jib.from("jetty");
+    }
 
-    // JVM Flags are ignored
-    List<String> entrypoint = processor.computeEntrypoint(ImmutableList.of());
+    List<String> entrypoint;
+    if (baseImage.isPresent() && !baseImage.get().startsWith("jetty")) {
+      List<String> entrypointFromCli = commonContainerConfigCliOptions.getEntrypoint();
```
chanseokoh, line 70: If an entrypoint is given, we should always honor it even if it's `jetty`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Aha that would make sense^ I just realized what the problem was. The entrypoint was correctly being set to `null` but I was setting `cmd` to an empty list instead of setting it to `null`. This is why it wasn't getting inherited from the base image.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
mpeddada1: Yes that is true. This is indeed a general purpose flag but I would also say that it is an absolute necessity for the `war` command. Since it is more of an enhancement for the other commands and this PR is already pretty large, I was thinking of doing this in a different PR. Although, I'm not sure how much of a difference adding `--app-root` would make to the `build` command since the creator of the buildfile can use the `dest` variable to copy files to a specific destination on the container?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
mpeddada1: That is a great question. This has proven to be a little tricky. The code here looks very similar to the one the jar command but the problem is that the individual methods take in different parameters (such as Jar cli options or War cli options). At the moment, the jar cli options (like `mode` or `jvm flags`) aren't accessible to the war command. The one section that I thought was common between the two was this:
```
ArtifactProcessor processor = ArtifactProcessors.fromWar(warFile, cacheDirectories, this, 
 commonContainerConfigCliOptions);
JibContainerBuilder containerBuilder =. WarFiles.toJibContainerBuilder(
              processor, commonCliOptions, commonContainerConfigCliOptions, logger);
Containerizer containerizer = Containerizers.from(commonCliOptions, logger, cacheDirectories);
// Enable registry mirrors
Multimaps.asMap(globalConfig.getRegistryMirrors()).forEach(containerizer::addRegistryMirrors);
JibContainer jibContainer = containerBuilder.containerize(containerizer);
```
I started creating a helper class for this to serve as an `ArtifactToContainerConverter` but soon realized that in order to make this work for both cases, this class would need access to both `jar` and `war` cli options.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: `ContainerBuilders.create("jetty")` can be used in place of `Jib.from` but the difference is that the former does some extra credential work to get credentials for a base image registry while the latter doesn't. I assumed that this may not be needed if we were to use the default? This way we don't have to log the following messaging every time:

```
[WARN] The credential helper (docker-credential-desktop) has nothing for server URL: registry-1.docker.io
Got output:
credentials not found in native keychain
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
chanseokoh: Oh, good point. The option doesn't make sense to the `build` command but only for `jar` and `war`. We don't have to do this in this PR.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
chanseokoh: Gotcha.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
chanseokoh: Hmm... I am aware of the ugliness of the warning (which the user can resolve by configuring `~/.docker/config.json` in a more modern way). However, I think we should still configure auth for public Docker Hub images. Logging in increases the rate limit.

mpeddada1: Oh that is a really good point. Modified this to use `ContainerBuilders.create("jetty")`. instead of `Jib.from()`

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Done.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Thanks, done.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessors.java
mpeddada1: My bad, thanks.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -17,32 +17,86 @@
...
+    List<FileEntriesLayer> layers;
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    if (baseImage.isPresent()) {
+      containerBuilder =
+          ContainerBuilders.create(
+              baseImage.get(), Collections.emptySet(), commonCliOptions, logger);
+    } else {
+      containerBuilder =
+          ContainerBuilders.create("jetty", Collections.emptySet(), commonCliOptions, logger);
+    }
```
chanseokoh, line 64:
```suggestion
    String baseImage = commonContainerConfigCliOptions.getFrom().orElse("jetty");
    JibContainerBuilder containerBuilder =
          ContainerBuilders.create(baseImage, Collections.emptySet(), commonCliOptions, logger);
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -17,32 +17,86 @@
...
 
-    // JVM Flags are ignored
-    List<String> entrypoint = processor.computeEntrypoint(ImmutableList.of());
+    List<String> programArguments =
+        commonContainerConfigCliOptions.getProgramArguments().isEmpty()
+            ? null
+            : commonContainerConfigCliOptions.getProgramArguments();
```
chanseokoh, line 69:
```suggestion
    List<String> programArguments =
        commonContainerConfigCliOptions.getProgramArguments().orElse(null);
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -17,32 +17,86 @@
...
+    layers = processor.createLayers();
+    containerBuilder
+        .setEntrypoint(computeEntrypoint(commonContainerConfigCliOptions, processor))
+        .setProgramArguments(programArguments)
+        .setFileEntriesLayers(layers)
```
chanseokoh, line 74:
```suggestion
        .setFileEntriesLayers(processor.createLayers())
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -17,32 +17,86 @@
...
+      throws IOException {
+    List<String> entrypoint = commonContainerConfigCliOptions.getEntrypoint();
+    if (!entrypoint.isEmpty()) {
+      return entrypoint;
+    }
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    Boolean isDefaultBaseImage =
+        !baseImage.isPresent() || (baseImage.isPresent() && baseImage.get().startsWith("jetty"));
+    if (isDefaultBaseImage) {
+      return processor.computeEntrypoint(ImmutableList.of());
```
chanseokoh, line 98: Hmm... I cannot say I like this. I may not remember all the context exactly, but I feel the entire logic of this function should reside inside `processor.computeEntrypoint()`. The `processor` instance should be created with (or given at some point potentially including now) all the necessary information to be able to do its job to compute and give us an entyrpoint. Your thoughts?

jib-cli/src/test/java/com/google/cloud/tools/jib/ArtifactProcessorsTest.java
```diff
@@ -174,4 +185,50 @@ public void testDetermineJavaMajorVersion_emptyClassFile() throws URISyntaxExcep
...
     assertThat(exception).hasMessageThat().startsWith("Reached end of class file (class1.class)");
   }
+
+  @Test
+  public void testFromWar_noJettyBaseImageAndNoAppRoot() {
+    when(mockCommonContainerConfigCliOptions.getFrom()).thenReturn(Optional.of("base-image"));
+
+    IllegalStateException exception =
+        assertThrows(
+            IllegalStateException.class,
```
chanseokoh, line 195: `IllegalArgumentException` sounds better.

jib-cli/src/test/java/com/google/cloud/tools/jib/ArtifactProcessorsTest.java
```diff
@@ -174,4 +185,50 @@ public void testDetermineJavaMajorVersion_emptyClassFile() throws URISyntaxExcep
...
+            mockCacheDirectories,
+            mockWarCommand,
+            mockCommonContainerConfigCliOptions);
+
+    assertThat(processor).isInstanceOf(StandardWarExplodedProcessor.class);
+  }
+
+  @Test
+  public void testFromWar_jettyBaseImageSpecified_success() {
+    when(mockCommonContainerConfigCliOptions.getFrom()).thenReturn(Optional.of("jetty"));
```
chanseokoh, line 225: How about "jetty:tag" to make it more interesting?

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/WarTest.java
```diff
@@ -0,0 +1,584 @@
...
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import org.apache.commons.lang3.ArrayUtils;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import picocli.CommandLine;
+import picocli.CommandLine.MissingParameterException;
+
+@RunWith(JUnitParamsRunner.class)
+public class WarTest {
```
chanseokoh, line 43: Can you point out which parts have been changed and are worth reviewing in this test?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Ah I agree that it is not ideal to compute entrypoint in two places. It is definitely possible to have this logic in computeEntrypoint in the `processor` but I'm worried that this will blur the line between `WarFiles` and the processors. I would think the processor was more for opening up the artifact and creating layers whereas `WarFiles` was for accepting cli arguments. But I'm open to the option of adding this logic to `processor` if it is making the code confusing to read.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/WarTest.java
mpeddada1: Absolutely. `testParse_defaults()` and `testParse_appRoot()` are the only two methods that have changed.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Nice, thank you.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Done.

jib-cli/src/test/java/com/google/cloud/tools/jib/ArtifactProcessorsTest.java
mpeddada1: Good idea.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessors.java
```diff
@@ -80,6 +85,36 @@ public static ArtifactProcessor from(
...
+    boolean isCustomJetty =
+        baseImage.isPresent()
+            && (baseImage.get().equals("jetty") || baseImage.get().startsWith("jetty:"));
+    boolean isDefaultBaseImage = !baseImage.isPresent() || isCustomJetty;
+    if (!isDefaultBaseImage && !warOptions.getAppRoot().isPresent()) {
+      throw new IllegalArgumentException(
+          "Please set the app root of the container with `--app-root` when specifying a base image that is not jetty.");
+    }
+    AbsoluteUnixPath appRootPath =
+        appRoot.isPresent() ? appRoot.get() : AbsoluteUnixPath.get(DEFAULT_JETTY_APP_ROOT);
```
chanseokoh, line 117: nit: `appRoot.orElse(AbsoluteUnixPath.get(...))`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessors.java
```diff
@@ -80,6 +85,36 @@ public static ArtifactProcessor from(
...
+      CacheDirectories cacheDirectories,
+      War warOptions,
+      CommonContainerConfigCliOptions commonContainerConfigCliOptions) {
+    Optional<AbsoluteUnixPath> appRoot = warOptions.getAppRoot();
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    boolean isCustomJetty =
+        baseImage.isPresent()
+            && (baseImage.get().equals("jetty") || baseImage.get().startsWith("jetty:"));
+    boolean isDefaultBaseImage = !baseImage.isPresent() || isCustomJetty;
+    if (!isDefaultBaseImage && !warOptions.getAppRoot().isPresent()) {
```
chanseokoh, line 112:
```suggestion
    if (!isDefaultBaseImage && !appRoot.isPresent()) {
```

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessors.java
```diff
@@ -80,6 +85,36 @@ public static ArtifactProcessor from(
...
+   * @return ArtifactProcessor
+   */
+  public static ArtifactProcessor fromWar(
+      Path warPath,
+      CacheDirectories cacheDirectories,
+      War warOptions,
+      CommonContainerConfigCliOptions commonContainerConfigCliOptions) {
+    Optional<AbsoluteUnixPath> appRoot = warOptions.getAppRoot();
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    Boolean isDefaultBaseImage =
```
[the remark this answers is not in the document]
mpeddada1: That is a great point. Used `ImageReference.parse` before checking if the base image is `jetty`

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ArtifactProcessors.java
```diff
@@ -80,6 +87,38 @@ public static ArtifactProcessor from(
...
+      Path warPath,
+      CacheDirectories cacheDirectories,
+      War warOptions,
+      CommonContainerConfigCliOptions commonContainerConfigCliOptions)
+      throws InvalidImageReferenceException {
+    Optional<AbsoluteUnixPath> appRoot = warOptions.getAppRoot();
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    boolean isCustomJetty =
+        baseImage.isPresent()
+            && ImageReference.parse(baseImage.get()).getRepository().contains("jetty");
```
chanseokoh, line 110: It should be `.equals()` rather than `.contains()`. And isn't it that `getRepository()` returns `library/jetty` for Docker Hub images? And perhaps check `getRegistry()` too just in case?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
chanseokoh: Ah, I see. That makes sense. In that perspective, how about not calling `processor.computeEntrypoint()` at all? I see what `processor.computeEntrypoint()` doesn't have anything to do with processing the contents of a WAR. Also, it doesn't make sense that `processor.computeEntrypoint()` blindly returns an entrypoint that only works for `jetty`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -17,32 +17,75 @@
...
+  private static List<String> computeEntrypoint(
+      CommonContainerConfigCliOptions commonContainerConfigCliOptions, ArtifactProcessor processor)
+      throws IOException {
+    List<String> entrypoint = commonContainerConfigCliOptions.getEntrypoint();
+    if (!entrypoint.isEmpty()) {
+      return entrypoint;
+    }
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    boolean isDefaultBaseImage =
+        !baseImage.isPresent() || (baseImage.isPresent() && baseImage.get().startsWith("jetty"));
```
chanseokoh, line 85: Need to update this too.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: That is a good point. `computeEntrypoint()` did look into the contents of a JAR but it doesn't serve the same purpose for WAR files. Now that I think about it, it may be better to have a separate `JarProcessor` and a `WarProcessor` instead of `ArtifactProcessor` since the only method that is implemented in the WAR case is `createLayers`? This would improve the readability by getting rid of the unused methods from `StandardWarExplodedProcessor`

chanseokoh (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -16,33 +16,83 @@
...
+      return entrypoint;
+    }
+    Optional<String> baseImage = commonContainerConfigCliOptions.getFrom();
+    if (isJetty(baseImage)) {
+      return ImmutableList.of("java", "-jar", "/usr/local/jetty/start.jar");
+    }
+    return null;
+  }
+
+  private static Boolean isJetty(Optional<String> baseImage) throws InvalidImageReferenceException {
```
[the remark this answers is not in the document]
mpeddada1: Yeah, it is unfortunate that this has proven to be cumbersome especially with the front-end CLI classes. I will merge this PR but I'm considering sending out another PR to refactor this a bit.
This is more of a design choice but if we make `--jvm-flags`, `--mode` and `--app-root` available to both commands then we could be able resolve the problem described in this [comment](https://github.com/GoogleContainerTools/jib/pull/3285#discussion_r649173334).

PR #3292, opened 2021-06-09
Doc updates about runtime classpath and main class

chanseokoh:


jib-maven-plugin/README.md
```diff
@@ -285,7 +286,7 @@ Property | Type | Default | Description
...
 `extraClasspath` | list | *None* | Additional paths in the container to prepend to the computed Java classpath.
-`expandClasspathDependencies` | boolean | `false` | When set to true, does not use a wildcard (for example, `/app/lib/*`) for dependency JARs in the default Java runtime classpath but instead enumerates the JARs. Has the effect of preserving the classpath loading order as defined by the Maven project.
+`expandClasspathDependencies` | boolean | `false` | <ul><li>Java 8 *or* Jib < 3.1: When set to true, does not use a wildcard (for example, `/app/lib/*`) for dependency JARs in the default Java runtime classpath but instead enumerates the JARs. Has the effect of preserving the classpath loading order as defined by the Maven project.</li><li>Java >= 9 *and* Jib >= 3.1: Jib always enumerates the dependencies JARs. (Therefore, the option has no effect.) This is achieved by [creating and using a JVM argument file](#custom-container-entrypoint) for the `--class-path` JVM argument.</li></ul>
```
mpeddada1, line 289: What do you think of this?
```suggestion
`expandClasspathDependencies` | boolean | `false` | <ul><li>Java 8 and >=9 before Jib 3.1: When set to true, does not use a wildcard (for example, `/app/lib/*`) for dependency JARs in the default Java runtime classpath but instead enumerates the JARs. Has the effect of preserving the classpath loading order as defined by the Maven project.</li><li>Java >= 9 *and* Jib >= 3.1: This option has no effect. Jib *always* enumerates the dependencies JARs.  This is achieved by [creating and using a JVM argument file](#custom-container-entrypoint) for the `--class-path` JVM argument.</li></ul>
```

mpeddada1 (approved):
LGTM, only one minor comment.

jib-maven-plugin/README.md
chanseokoh: I like the suggestion for the second bullet point.

For the first one, I think `Java 8 and >=9 before Jib 3.1` is ambiguous, in that it could be read like `(Java 8 and >= 9) before Jib 3.1`. In any case, I'll merge this to now and apply the suggestion in a later PR. Many doc changes (including Stack Overflow) depend on this. I'll make sure to update this.

chanseokoh: Will be done by #3293.

PR #3305, opened 2021-06-11
Update docs for Jib Core 0.19.0

mpeddada1:


mpeddada1:
@chanseokoh need a small review on this.

chanseokoh (approved):

chanseokoh:
Oh, and just so you know, note that we deprecated the distroless-related methods, but it doesn't mean we removed them. The methods are there during the deprecation period and there are no breaking changes. It's just that they are marked deprecated potentially with warnings at runtime until they will be decommissioned in a major release bump if it were GA.

mpeddada1:
ah, thanks for clarifying!

PR #3310, opened 2021-06-18
Remove duplicate code from Jib CLI

mpeddada1:

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
@@ -183,13 +183,12 @@ public class CommonContainerConfigCliOptions {
    * Returns {@code true} if the user-specified base image is jetty or if a custom base image is not
    * specified.
    *
-   * @param baseImage the base image
    * @return a boolean
    * @throws InvalidImageReferenceException if image reference is invalid
    */
-  public static Boolean isJetty(Optional<String> baseImage) throws InvalidImageReferenceException {
-    if (baseImage.isPresent()) {
-      ImageReference baseImageReference = ImageReference.parse(baseImage.get());
+  public Boolean isJetty() throws InvalidImageReferenceException {
+    if (from != null) {
+      ImageReference baseImageReference = ImageReference.parse(from);
       return baseImageReference.getRegistry().equals("registry-1.docker.io")
           && baseImageReference.getRepository().equals("library/jetty");
     }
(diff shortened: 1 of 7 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
```diff
@@ -176,4 +178,21 @@
...
+
+  /**
+   * Returns {@code true} if the user-specified base image is jetty or if a custom base image is not
+   * specified.
+   *
+   * @param baseImage the base image
+   * @return a boolean
+   * @throws InvalidImageReferenceException if image reference is invalid
+   */
+  public static Boolean isJetty(Optional<String> baseImage) throws InvalidImageReferenceException {
```
chanseokoh, line 190: What if we make this non-static and examine the `from` field of the class?

mpeddada1: Great catch! That makes more sense.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java
```diff
@@ -176,4 +178,20 @@
...
   }
+
+  /**
+   * Returns {@code true} if the user-specified base image is jetty or if a custom base image is not
+   * specified.
+   *
+   * @return a boolean
+   * @throws InvalidImageReferenceException if image reference is invalid
+   */
+  public Boolean isJetty() throws InvalidImageReferenceException {
```
chanseokoh, line 189: Sorry, missed this. I think the name should be `isJettyBaseImage`. Otherwise, it's confusing.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
```diff
@@ -144,29 +143,9 @@ public void testToJibContainerBuilder_nonJettyBaseImageSpecifiedAndNoEntrypoint(
             mockLogger);
     ContainerBuildPlan buildPlan = containerBuilder.toContainerBuildPlan();
 
-    assertThat(buildPlan.getBaseImage()).isEqualTo("base-image");
     assertThat(buildPlan.getEntrypoint()).isNull();
```
chanseokoh, line 146: Hmm... does it make sense that the entrypoint is null when `isJetty()` from the mock is returning true?

chanseokoh (commented):
Basically LGTM

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/war/WarFilesTest.java
mpeddada1: I added an assert statement to clarify this. Let me know what you think?

chanseokoh: Oh, my bad. I mistakenly thought the new addition on line 65 was for `setUp()` to affect every test.

chanseokoh (approved):

PR #3336, opened 2021-07-07
Add jar size suffix to target dependency paths on image when dependency jars have same artifact ID and version

### LINKED ISSUE #3331: Classpath file (/app/jib-classpath-file) doesn't match libs folder if libraries have same artifactId/version
**Environment**:

- *Jib version:* 3.1+ (3.1.0 and 3.1.1)
- *Build tool:* maven 3.6.3
- *OS:* MacOS 10.16

**Description of the issue**:

In our project, we have two dependencies with the `groupId` as only difference:

```xml
        <dependency>
            <groupId>com.name.app1</groupId>
            <artifactId>rest-client</artifactId>
            <version>2.0.0</version>
        </dependency>
        <dependency>
            <groupId>com.name.app2</groupId>
            <artifactId>rest-client</artifactId>
            <version>2.0.0</version>
        </dependency>
```

Both are, in this case, in the same version (which never happened before). 

If I run the application, I end up with `ClassNotFoundException`. 

If I dig a little in the application, I can see the following results:

```shell
> $ docker run -it --entrypoint=/busybox/sh --rm my-custom-image:1.2.3
~ $ ls /app/libs/ | grep rest-client
rest-client-2.0.0-187729.jar
rest-client-2.0.0-54224.jar
~ $ cat /app/jib-classpath-file | sed 's/:/\n/g' | grep rest-client
/app/libs/rest-client-2.0.0.jar
/app/libs/rest-client-2.0.0.jar
```

To sumarize, `jib` is working well at lib level to add a suffix at the end of the `jar` in the `/app/libs` folder to avoid a filename clash at filesystem level. 

But, if we look at the `/app/jib-classpath-file` to represent the classpath for our app, the hash is not integrated and none of the dependency can be resolved by our app leading to a `ClassNotFoundException`. 

The expected behavoir would be to launch the app with the correct classpath, so to generate correctly the `/app/jib-classpath-file`. 

If you need more information, let me know 😇.

chanseokoh:
Fixes #3331.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -61,6 +61,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -620,13 +621,27 @@ public class PluginConfigurationProcessor {
 
     if (projectProperties.getMajorJavaVersion() >= 9
         || rawConfiguration.getExpandClasspathDependencies()) {
-      List<String> dependencies =
-          projectProperties
-              .getDependencies()
+      List<Path> jars = projectProperties.getDependencies();
+      List<String> duplicates =
+          jars.stream()
+              .map(path -> path.getFileName().toString())
+              .collect(Collectors.groupingBy(filename -> filename, Collectors.counting()))
+              .entrySet()
               .stream()
-              .map(path -> appRoot.resolve("libs").resolve(path.getFileName()).toString())
+              .filter(entry -> entry.getValue() > 1)
+              .map(Map.Entry::getKey)
               .collect(Collectors.toList());
-      classpath.addAll(dependencies);
+
+      for (Path jar : jars) {
+        // Handle duplicates by appending filesize to the end of the file. This renaming logic
+        // must be in sync with the code that does the same in the other place. See
+        // https://github.com/GoogleContainerTools/jib/issues/3331
+        String jarName = jar.getFileName().toString();
+        if (duplicates.contains(jarName)) {
+          jarName = jarName.replaceFirst("\\.jar$", "-" + Files.size(jar)) + ".jar";
+        }
+        classpath.add(appRoot.resolve("libs").resolve(jarName).toString());
+      }
     } else {
       classpath.add(appRoot.resolve("libs/*").toString());
     }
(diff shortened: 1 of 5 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -620,13 +621,27 @@ static JavaContainerBuilder getJavaContainerBuilderWithBaseImage(
...
-      classpath.addAll(dependencies);
+
+      for (Path jar : jars) {
+        // Handle duplicates by appending filesize to the end of the file. This renaming logic
+        // must be in sync with the code that does the same in the other place. See
+        // https://github.com/GoogleContainerTools/jib/issues/3331
+        String jarName = jar.getFileName().toString();
+        if (duplicates.contains(jarName)) {
+          jarName = jarName.replaceFirst("\\.jar$", "-" + Files.size(jar)) + ".jar";
+        }
```
mpeddada1, line 642: This situation is pretty rare but what if we end up with jars that are of the same size? How about we append incremented integers to jar name? Such as `jar-name-1.jar` and `jar-name-2.jar`.

chanseokoh: I remember the team had a discussion on this (although might not be really deeply). There probably were multiple reasons that we converged on this method. I think one of the main reasons was that, you can identify which dependency is which by looking at the classpath string. And I think it's not just rare; I don't think anyone will ever run into the situation where two dependencies have the same artifact ID, version, and the JAR file size. I think the chance in practice is really zero.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -620,13 +621,27 @@ static JavaContainerBuilder getJavaContainerBuilderWithBaseImage(
...
         || rawConfiguration.getExpandClasspathDependencies()) {
-      List<String> dependencies =
-          projectProperties
-              .getDependencies()
+      List<Path> jars = projectProperties.getDependencies();
+      List<String> duplicates =
+          jars.stream()
+              .map(path -> path.getFileName().toString())
+              .collect(Collectors.groupingBy(filename -> filename, Collectors.counting()))
+              .entrySet()
```
mpeddada1, line 629: nit: Maybe we could simplify the logic by put this under a separate variable (`occurrencesMap`)?

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
mpeddada1: I see. You're right, the probability of this happening is very small.

mpeddada1 (approved):

mpeddada1 (approved):

PR #3341, opened 2021-07-08
Update Jib versions post-3.1.2 release

chanseokoh:


docs/faq.md
```diff
@@ -332,7 +332,7 @@ In Maven, you can use the `maven-resources-plugin` to copy files to your extra d
   ...
   <plugin>
     <artifact>maven-resources-plugin</artifact>
-    <version>3.1.1</version>
+    <version>3.2.0</version>
```
chanseokoh, line 335: I realized there doesn't exist 3.1.1 for `maven-resources-plugin`. I think it was a wrong update during the post release process in the past.

mpeddada1: That is a great catch!

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-jar-containerization.xml
```diff
@@ -37,7 +37,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-jar-plugin</artifactId>
-        <version>3.1.1</version>
+        <version>3.2.0</version>
```
mpeddada1, line 40: `maven-jar-plugin`'s latest version is 3.2.0. I'm assuming we're intentionally updating it here?

mpeddada1 (approved):

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-jar-containerization.xml
chanseokoh: Yeah, `maven-jar-plugin` does have 3.1.1, but I think we can always use the latest.

Issue #1409, opened 2019-01-14
Support retries to push to a registry


rmannibucau:
**Description of the issue**:

On laggy registries - which means you often get a 504 timeout issue - it is not possible to push images with jib whereas it works with docker with a retry (it retries not yet uploaded layers).

**Expected behavior**:

A retry would happen on layers (default to 3 maybe?).

chanseokoh:
#691 is related, although that isn't about dealing with 504.

briandealwis:
We should also investigate using [chunked uploads](https://docs.docker.com/registry/spec/#pushing-an-image) since some layers may be rather huge.  The only worrisome aspect is that it's unclear what should be our minimum chunk size:

> There is no enforcement on layer chunk splits other than that the server must receive them in order. The server may enforce a minimum chunk size. If the server cannot accept the chunk, a 416 _Requested Range Not Satisfiable_ response will be returned and will include a Range header indicating the current status:

It appears that chunks must be provided in sequence.  The documentation describing the 416 result says:
> Out of order chunk: the range of the next chunk must start immediately after the “last valid range” from the previous response

rmannibucau:
up? on openshift we easily get SSL issues (indeed without any modification a relaunch works) when pushing concurrently layers so a proper retry would help on client side.

chanseokoh:
One form retry is implemented in #3351. There can be other ways of retrying, but #3351 should already be very helpful. Closing the issue.

PR #3351, opened 2021-07-21
basic retry for http client

rmannibucau:
Follow up of https://github.com/GoogleContainerTools/jib/issues/1409

It enables http client retry when the blob content is re-readable.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
@@ -38,7 +38,5 @@ public interface Blob {
    * @return {@code true} if {@link #writeTo(OutputStream)} can be called multiple times, {@code
    *     false} otherwise.
    */
-  default boolean isRetryable() {
-    return false;
-  }
+  boolean isRetryable();
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
@@ -32,7 +32,8 @@ public interface Blob {
   BlobDescriptor writeTo(OutputStream outputStream) throws IOException;
 
   /**
-   * Enables to notify to the HTTP client if the underlying request can be retried.
+   * Enables to notify if the underlying request can be retried
+   * (useful in the context of a retryable HTTP request for ex).
    * @return {@code true} if {@link #writeTo(OutputStream)} can be called multiple times, {@code false} otherwise.
    */
   default boolean isRetryable() {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
@@ -32,9 +32,11 @@ public interface Blob {
   BlobDescriptor writeTo(OutputStream outputStream) throws IOException;
 
   /**
-   * Enables to notify if the underlying request can be retried
-   * (useful in the context of a retryable HTTP request for ex).
-   * @return {@code true} if {@link #writeTo(OutputStream)} can be called multiple times, {@code false} otherwise.
+   * Enables to notify if the underlying request can be retried (useful in the context of a
+   * retryable HTTP request for ex).
+   *
+   * @return {@code true} if {@link #writeTo(OutputStream)} can be called multiple times, {@code
+   *     false} otherwise.
    */
   default boolean isRetryable() {
     return false;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
@@ -30,4 +30,12 @@ public interface Blob {
    * @throws IOException if writing the BLOB fails
    */
   BlobDescriptor writeTo(OutputStream outputStream) throws IOException;
+
+  /**
+   * Enables to notify to the HTTP client if the underlying request can be retried.
+   * @return {@code true} if {@link #writeTo(OutputStream)} can be called multiple times, {@code false} otherwise.
+   */
+  default boolean isRetryable() {
+    return false;
+  }
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
@@ -25,9 +25,11 @@ import java.io.OutputStream;
 class WritableContentsBlob implements Blob {
 
   private final WritableContents writableContents;
+  private final boolean retryable;
 
-  WritableContentsBlob(WritableContents writableContents) {
+  WritableContentsBlob(WritableContents writableContents, boolean retryable) {
     this.writableContents = writableContents;
+    this.retryable = retryable;
   }
 
   @Override
@@ -38,6 +40,6 @@ class WritableContentsBlob implements Blob {
   // in general it is since the underlying data is accessed in the lambda
   @Override
   public boolean isRetryable() {
-    return true;
+    return retryable;
   }
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
@@ -37,7 +37,6 @@ class WritableContentsBlob implements Blob {
     return Digests.computeDigest(writableContents, outputStream);
   }
 
-  // in general it is since the underlying data is accessed in the lambda
   @Override
   public boolean isRetryable() {
     return retryable;
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
@@ -34,4 +34,10 @@ class WritableContentsBlob implements Blob {
   public BlobDescriptor writeTo(OutputStream outputStream) throws IOException {
     return Digests.computeDigest(writableContents, outputStream);
   }
+
+  // in general it is since the underlying data is accessed in the lambda
+  @Override
+  public boolean isRetryable() {
+    return true;
+  }
 }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
@@ -59,7 +59,7 @@ public class BlobHttpContent implements HttpContent {
 
   @Override
   public boolean retrySupported() {
-    return false;
+    return blob.isRetryable();
   }
 
   @Override
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
@@ -17,12 +17,15 @@
 package com.google.cloud.tools.jib.http;
 
 import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpBackOffIOExceptionHandler;
 import com.google.api.client.http.HttpHeaders;
 import com.google.api.client.http.HttpMethods;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpResponseException;
 import com.google.api.client.http.HttpTransport;
 import com.google.api.client.http.apache.v2.ApacheHttpTransport;
+import com.google.api.client.util.BackOff;
+import com.google.api.client.util.ExponentialBackOff;
 import com.google.api.client.util.SslUtils;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.common.annotations.VisibleForTesting;
@@ -312,6 +315,19 @@ public class FailoverHttpClient {
         httpTransport
             .createRequestFactory()
             .buildRequest(httpMethod, new GenericUrl(url), request.getHttpContent())
+            .setIOExceptionHandler(new HttpBackOffIOExceptionHandler(new ExponentialBackOff()) {
+              @Override
+              public boolean handleIOException(final HttpRequest request, final boolean supportsRetry) throws IOException {
+                final boolean result = super.handleIOException(request, supportsRetry);
+                final String reqString = request.getRequestMethod() + " " + request.getUrl();
+                if (result) { // google-http-client does not log that properly so let's compensate it
+                  logger.accept(LogEvent.warn(reqString + " failed and will be retried"));
+                } else {
+                  logger.accept(LogEvent.warn(reqString + " failed and will NOT be retried"));
+                }
+                return result;
+              }
+            })
             .setUseRawRedirectUrls(true)
             .setHeaders(requestHeaders);
     if (request.getHttpTimeout() != null) {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
@@ -342,7 +342,7 @@ public class FailoverHttpClient {
                         enableRetries && super.handleIOException(request, supportsRetry);
                     String requestUrl = request.getRequestMethod() + " " + request.getUrl();
                     if (result) { // google-http-client does not log that properly so let's
-                                  // compensate it
+                      // compensate it
                       logger.accept(LogEvent.warn(requestUrl + " failed and will be retried"));
                     } else {
                       logger.accept(LogEvent.warn(requestUrl + " failed and will NOT be retried"));
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
@@ -123,6 +123,7 @@ public class FailoverHttpClient {
     }
   }
 
+  private final boolean enableRetries;
   private final boolean enableHttpAndInsecureFailover;
   private final boolean sendAuthorizationOverHttp;
   private final Consumer<LogEvent> logger;
@@ -146,6 +147,7 @@ public class FailoverHttpClient {
       boolean sendAuthorizationOverHttp,
       Consumer<LogEvent> logger) {
     this(
+        true,
         enableHttpAndInsecureFailover,
         sendAuthorizationOverHttp,
         logger,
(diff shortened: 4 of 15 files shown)


meltsufin (commented):
@rmannibucau Thanks for the PR, but it looks like some unit tests are broken. Do they pass for you locally?

jib-core/src/main/java/com/google/cloud/tools/jib/http/BlobHttpContent.java
```diff
@@ -59,7 +59,7 @@ public String getType() {
 
   @Override
   public boolean retrySupported() {
-    return false;
+    return blob.isRetryable();
```
chanseokoh, line 62: Continuing the discussion, you'll see that now this package `c.g.c.t.jib.http` is the abstraction layer where we can start talking about things around HTTP.

jib-core/src/test/java/com/google/cloud/tools/jib/http/FailoverHttpClientTest.java
```diff
@@ -25,15 +28,21 @@
 import com.google.api.client.http.HttpTransport;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.blob.Blobs;
+import com.sun.net.httpserver.HttpHandler;
+import com.sun.net.httpserver.HttpServer;
```
chanseokoh, line 32: I do know this will never be an issue in practice at all, but we don't really want to introduce `com.sum.*` following general guidelines.

https://www.oracle.com/java/technologies/faq-sun-packages.html
https://rules.sonarsource.com/java/RSPEC-1191

Maybe take a look at `WithServerFailoverHttpClientTest.java` for an alternative.

chanseokoh (commented):
Thanks for your contribution!

jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
```diff
@@ -34,4 +34,10 @@
   public BlobDescriptor writeTo(OutputStream outputStream) throws IOException {
     return Digests.computeDigest(writableContents, outputStream);
   }
+
+  // in general it is since the underlying data is accessed in the lambda
+  @Override
+  public boolean isRetryable() {
+    return true;
```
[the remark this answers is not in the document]
rmannibucau: Fact is they are retryable in job-core and I agree - as mentionned in the comment - they are not generally.
It means that if we want this PR to work, we must have it true or rewrite all wrutable contents blob of jib-core to not use writable contents but something retryable which can be a RetryableWritableContents as a cheap solution or more specific new Blob impls.

Retry means = can be reread (but reread was weird, no?)

jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
```diff
@@ -30,4 +30,12 @@
    * @throws IOException if writing the BLOB fails
    */
   BlobDescriptor writeTo(OutputStream outputStream) throws IOException;
+
+  /**
+   * Enables to notify to the HTTP client if the underlying request can be retried.
+   * @return {@code true} if {@link #writeTo(OutputStream)} can be called multiple times, {@code false} otherwise.
+   */
+  default boolean isRetryable() {
```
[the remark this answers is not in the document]
rmannibucau: I guess this respects it even if the comment is focused on http since it is the main usage today. The return part is very explicit about the fact it is about calling N > 1 times writeTo, no?

jib-core/src/test/java/com/google/cloud/tools/jib/http/FailoverHttpClientTest.java
rmannibucau: @chanseokoh this is not an issue, http server is exported and "public" even if the package does not make it that explicit: https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/jdk.httpserver/share/classes/module-info.java. I saw WithServerFailoverHttpClientTest but thought that since HttpServer is there and fits better the test it was more relevant than a socket layer solution which is less readable IMHO.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -312,6 +315,19 @@ private Response call(String httpMethod, URL url, Request request, HttpTransport
...
             .buildRequest(httpMethod, new GenericUrl(url), request.getHttpContent())
+            .setIOExceptionHandler(new HttpBackOffIOExceptionHandler(new ExponentialBackOff()) {
+              @Override
+              public boolean handleIOException(final HttpRequest request, final boolean supportsRetry) throws IOException {
+                final boolean result = super.handleIOException(request, supportsRetry);
+                final String reqString = request.getRequestMethod() + " " + request.getUrl();
+                if (result) { // google-http-client does not log that properly so let's compensate it
+                  logger.accept(LogEvent.warn(reqString + " failed and will be retried"));
+                } else {
+                  logger.accept(LogEvent.warn(reqString + " failed and will NOT be retried"));
```
chanseokoh, line 326: I think the `else` branch can be debug (or removed). 
```suggestion
                  logger.accept(LogEvent.warn(reqString + " failed due to IOException and will be retried"));
                } else {
                  logger.accept(LogEvent.debug(reqString + " failed due to IOException and will NOT be retried"));
```

jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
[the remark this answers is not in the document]
chanseokoh: I think this should be false. It's risky in that, in the long run it makes it too easy for maintainers to eventually shoot themselves in the foot.

And in jib-core, there is a case where calling `writeTo()` [initiates a network call](https://github.com/GoogleContainerTools/jib/blob/a5c31ef3c94ff70aa7db23a793d376a984948eea/jib-core/src/main/java/com/google/cloud/tools/jib/registry/RegistryClient.java#L494-L506); that is, calling `writeTo()` multiple times means making HTTP requests multiple times, and even though we may get the same content from a server each time, I don't really feel comfortable calling this "retryable."

Nonetheless, it is incorrect to call `WritableContents` is retryable.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -312,6 +315,19 @@ private Response call(String httpMethod, URL url, Request request, HttpTransport
         httpTransport
             .createRequestFactory()
             .buildRequest(httpMethod, new GenericUrl(url), request.getHttpContent())
+            .setIOExceptionHandler(new HttpBackOffIOExceptionHandler(new ExponentialBackOff()) {
```
chanseokoh, line 318: What's the maximum number of retries (and the maximum time) until this eventually gives up? We don't want for Jib to run for too long when it's not recoverable (e.g., server is down for a day).

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -312,6 +315,19 @@ private Response call(String httpMethod, URL url, Request request, HttpTransport
...
+              @Override
+              public boolean handleIOException(final HttpRequest request, final boolean supportsRetry) throws IOException {
+                final boolean result = super.handleIOException(request, supportsRetry);
+                final String reqString = request.getRequestMethod() + " " + request.getUrl();
```
chanseokoh, line 322: nit per Google-style:
```suggestion
              public boolean handleIOException(HttpRequest request, boolean supportsRetry) throws IOException {
                boolean result = super.handleIOException(request, supportsRetry);
                String requestUrl = request.getRequestMethod() + " " + request.getUrl();
```

jib-core/src/test/java/com/google/cloud/tools/jib/http/FailoverHttpClientTest.java
```diff
@@ -416,6 +425,55 @@ public void testFollowFailoverHistory_portsDifferent() throws IOException {
...
+          exch.sendResponseHeaders(200, 0);
+          exch.close();
+        });
+
+        // here is the failure - no response sent (IOException for the client)
+        ex.close();
+        return;
+      }
+      // make the test fail
+      ex.sendResponseHeaders(123, 0);
```
chanseokoh, line 450: I don't think this works as you intended. 123 actually causes IOException. OTOH, 400 works.

There are some minor style errors, and given above, I'd have the code to look like this: 
```java
import static com.google.common.truth.Truth.assertThat;

public void testRetries() throws IOException {
    HttpServer server = HttpServer.create(new InetSocketAddress(0), 1);
    AtomicBoolean failed = new AtomicBoolean();
    server
        .createContext("/")
        .setHandler(
            exchange ->
                exchange.sendResponseHeaders(failed.compareAndSet(false, true) ? 123 : 200, 0));
    try {
      server.start();
      int port = server.getAddress().getPort();
      List<LogEvent> events = new ArrayList<>();
      int returnCode =
          new FailoverHttpClient(true, true, events::add)
              .get(new URL("http://localhost:" + port), Request.builder().build())
              .getStatusCode();
      assertThat(returnCode).isEqualTo(200);
      assertThat(failed.get()).isTrue();
      assertThat(events)
          .containsExactly(
              LogEvent.warn("GET http://localhost:" + port + " failed and will be retried"));
    } finally {
      server.stop(0);
    }
}
```

jib-core/src/test/java/com/google/cloud/tools/jib/http/FailoverHttpClientTest.java
chanseokoh: I am aware of that and the package is exported in the module but still wanted to exclude it. Can you guarantee that, especially for Java 8, every JDK implementation (e.g., IBM JVM) provides this package? Even the Javadoc you linked still says "JDK-specific HTTP server API."

However, practically this will cause no issue to anyone, so I don't want to make your life too difficult. I can accept this.

chanseokoh:
Also, there are test failures, so please take a look at them too.

rmannibucau:
Just a quick message to say I am AFK and will look when back - until someone beats me at it ;).
Only point I am not sure is writable blob which should then be split to work (Blobs.of(content, retryable)? or just ensure network call is done lazily) otherwise it will not reach the goal.

chanseokoh:
Thanks for the update. Feel free to take your time.

> Only point I am not sure is writable blob which should then be split to work (Blobs.of(content, retryable)?

I think adding another `Blobs.from()` is a reasonable option?
```java
  public static Blob from(WritableContents writable, boolean retryable) {
    return new WritableContentsBlob(writable, retryable);
  }
```

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
rmannibucau: In sec (quoted from the javadoc of the exponential backoff)

```
   request#     retry_interval     randomized_interval
  
   1             0.5                [0.25,   0.75]
   2             0.75               [0.375,  1.125]
   3             1.125              [0.562,  1.687]
   4             1.687              [0.8435, 2.53]
   5             2.53               [1.265,  3.795]
   6             3.795              [1.897,  5.692]
   7             5.692              [2.846,  8.538]
   8             8.538              [4.269, 12.807]
   9            12.807              [6.403, 19.210]
   10           19.210              BackOff.STOP
```

Guess another PR can change the retry logic to not use exp backoff but these defaults should tolerate the small hips I saw and just require configuration propagation work so can be done after once the strategies are discussed/decided.

jib-core/src/main/java/com/google/cloud/tools/jib/blob/Blob.java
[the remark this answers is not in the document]
chanseokoh: Please remove `default`.

jib-core/src/main/java/com/google/cloud/tools/jib/blob/WritableContentsBlob.java
```diff
@@ -25,13 +25,21 @@
...
     this.writableContents = writableContents;
+    this.retryable = retryable;
   }
 
   @Override
   public BlobDescriptor writeTo(OutputStream outputStream) throws IOException {
     return Digests.computeDigest(writableContents, outputStream);
   }
+
+  // in general it is since the underlying data is accessed in the lambda
```
chanseokoh, line 40: Remove.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -312,6 +333,23 @@ private Response call(String httpMethod, URL url, Request request, HttpTransport
...
+                  public boolean handleIOException(HttpRequest request, boolean supportsRetry)
+                      throws IOException {
+                    boolean result =
+                        enableRetries && super.handleIOException(request, supportsRetry);
+                    String requestUrl = request.getRequestMethod() + " " + request.getUrl();
+                    if (result) { // google-http-client does not log that properly so let's
+                                  // compensate it
+                      logger.accept(LogEvent.warn(requestUrl + " failed and will be retried"));
+                    } else {
+                      logger.accept(LogEvent.warn(requestUrl + " failed and will NOT be retried"));
```
chanseokoh, line 348: LogEvent.debug

rmannibucau: Can we keep info? debug means it will never be seen whereas it must be seen to report the server has issues. Happy to work on a config to change the level later if it bother users but strongly think the default should be visible without any log tuning.

chanseokoh:
Oops, I tried to resolve the merge conflict but it broke the code.
https://github.com/GoogleContainerTools/jib/pull/3351/commits/f34b525cc4327bc86f88bee57267cf58b71b7d44#diff-5b8da1dfe36a07daee79c63d41fdbecb8c864182144f2ca7449e90b6ed2b7914R66

It should be 
```diff
-entry, entry.isFile() ? Blobs.from(entry.getPath()) : Blobs.from(ignored -> {}), true);
+entry, entry.isFile() ? Blobs.from(entry.getPath()) : Blobs.from(ignored -> {}, true));
```

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
chanseokoh: The user will see the IOException anyway, right? (BTW, as I mentioned in my earlier comment, I'm only taking about the `else` branch.) I think that is enough and seeing "will NOT be retried" every time on a failure, particularly if `supportsRetry` is false seems redundant. And generally speaking, we tend to be a bit sparing for WARN if it is not something actionable.

However, if this is never logged when for a no-error case (i.e., everything is working successfully), I can accept that. Can you confirm?

rmannibucau: Yes, it is only logged when there was an error and both cases are important (why wasnt it retried and why is it slow or replay basically).

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
chanseokoh: FTR: https://github.com/GoogleContainerTools/jib/pull/3351#discussion_r683571069

jib-core/src/test/java/com/google/cloud/tools/jib/http/FailoverHttpClientTest.java
chanseokoh: I will take care of this myself in a follow-up PR.

chanseokoh:
I think this LGTM. Thanks for the contribution! Will take another look in the afternoon. There are a few things that I want to clean up, but I can take care of this myself later.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -176,6 +176,6 @@ public Blob build() throws IOException {
       tarStreamBuilder.addTarArchiveEntry(entry);
     }
 
-    return Blobs.from(tarStreamBuilder::writeAsTarArchiveTo);
+    return Blobs.from(tarStreamBuilder::writeAsTarArchiveTo, true);
```
chanseokoh, line 179: Just noticed that generally speaking `TarStreamBuilder::writeAsTarArchiveTo` is not retryable, as non-retryable `Blob`s [can be added](https://github.com/GoogleContainerTools/jib/blob/1fd6b62c90bb96a53fdc719265c4224157810361/jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java#L92) to it.
```suggestion
    return Blobs.from(tarStreamBuilder::writeAsTarArchiveTo, false);
```

jib-core/src/main/java/com/google/cloud/tools/jib/tar/TarStreamBuilder.java
```diff
@@ -77,7 +78,7 @@ public void addByteEntry(byte[] contents, String name, Instant modificationTime)
     TarArchiveEntry entry = new TarArchiveEntry(name);
     entry.setSize(contents.length);
     entry.setModTime(modificationTime.toEpochMilli());
-    archiveMap.put(entry, Blobs.from(outputStream -> outputStream.write(contents)));
+    archiveMap.put(entry, Blobs.from(outputStream -> outputStream.write(contents), false));
```
chanseokoh, line 81: I think true. (`contents` isn't totally immutable as it's a reference to an array, but this is implied and assumed in the existing code.)
```suggestion
    archiveMap.put(entry, Blobs.from(outputStream -> outputStream.write(contents), true));
```

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
rmannibucau: It does reopen the stream  so should be retryable no?
Dont have it in mind but think it must be retryable somehow or we will get troubles later.

chanseokoh: The method `writeAsTarArchiveTo` is retryable if and only if all of the added blobs are retryable, which isn't guaranteed (even if it may be the case that no non-retryable blobs are currently added in our codebase). So I think `false` is the right thing to do.

We won't get into a trouble later because of this. The only place this is used is when [assembling an application layer](https://github.com/GoogleContainerTools/jib/blob/1fd6b62c90bb96a53fdc719265c4224157810361/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java#L108-L109), and we do cache and return a new blob from a cache file using `Blobs.from()`. That blob is retryable, so your retry logic will certainly work for pushing application layers. There is no need to insist `true` here only for the sake of making the logic work. `false` is conceptually correct, so let's do that.

chanseokoh (approved):
Awesome! I think this will significantly reduce user friction. Something we always wanted to do. Thanks again!

chanseokoh:
Follow-up: #3390

chanseokoh:
@rmannibucau you can test the the retry feature with 3.1.4. I had to tweak it a bit due to #3422.

Thanks again for you contribution!

PR #3353, opened 2021-07-22
Reset user and group for automatically added parent directories while building reproducible layer tars

### LINKED ISSUE #3352: When building tar layers, user and group for automatically added parent directories are not explicitly reset
This was discovered and analyzed in #3342.

We recursively add all parent directories of a file entry when assembling a tar for a layer, but we are not explicitly setting the user and group. The latest `commons-compress` 1.21 had a behavior change in that it picks up the user and group of a local directory, while previously they were set to `0:0` by default.

Regardless of whether `commons-compress` reverts the behavior, I think it's better for us to always reset these properties.

chanseokoh:
Fixes #3352.

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
@@ -16,12 +16,13 @@
 
 package com.google.cloud.tools.jib.api;
 
+import static com.google.common.truth.Truth.assertThat;
+
 import com.google.cloud.tools.jib.api.buildplan.AbsoluteUnixPath;
 import com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer;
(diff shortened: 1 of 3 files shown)


mpeddada1:
Interesting find! I think proactively keeping user id and group id constant is a good idea for reproducibility. It shouldn't result in a change in existing behavior since they previously used to be `0:0` by default. Could be helpful to document this here: https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#please-tell-me-more-about-reproducibility.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ReproducibleImageTest.java
```diff
@@ -16,12 +16,13 @@
 
 package com.google.cloud.tools.jib.api;
 
+import static com.google.common.truth.Truth.assertThat;
```
mpeddada1, line 19: Love the refactoring changes here.

mpeddada1 (approved):

Issue #3364, opened 2021-08-02
Support username and password tokens from dockercfg auths


tpokki:
**Environment**:

- *Jib version:* 3.1.2 
- *Build tool:* maven 3.6.3
- *OS:* macOS

**Description of the issue**:
Jib is missing support for `username` and `password` fields in dockercfg auths block.
Example
```
{
  "auths": {
      "registry.example.com": {
           "username": "someuser",
           "password": "somepw"
       }
}
```

**Expected behavior**:
Jib to pick up username and password fields from dockercfg (and not just `credHelper` and `auth`). 

**Steps to reproduce**:

  1. Add username and password for your registry into your config.json as described above
  2. Build
  3. Jib does not pick up username and password

**Additional Information**: 

I have implemented the necessary changes to add support in here:
https://github.com/tpokki/jib/commit/479a5dfcebfe1d451bd43d00b7e7adddbc16080f

If you approve, I'll create PR. 

Thanks,
Toni

chanseokoh:
I think this is great. Create a PR (with a test), and we'll review it. Thanks a lot!

chanseokoh:
Fixed by #3365.

chanseokoh:
@tpokki we released Jib plugins 3.1.4 with the fix.

chanseokoh:
Thanks for your contribution!

PR #3365, opened 2021-08-02
Added support for username and password tokens in dockercfg auth

tpokki:
Support username and password tokens from dockercfg auths #3364


chanseokoh (approved):
Awesome! Thank you for your great contribution!

PR #3366, opened 2021-08-02
fix sonar bug to call Optional#isPresent() before accessing value.

zhumin8:
Fix sonar bug for fixit: https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTIQcB_fbtb802TR&open=AXrlUTIQcB_fbtb802TR 

This change should be complaint with sonar rules, but is it worth the change or should we just leave it alone? It does make one less call to `buildFile.getLayers()`, so that could be a benefit. 

Note: there are a bunch of similar sonar bugs.


chanseokoh:
There isn't an practical issue currently (most things in Jib are immutable, and `getLayers()` is a simple getter), but for the sake of silencing SonarCloud, I think it's worth updating the code.

chanseokoh (approved):

PR #3367, opened 2021-08-02
Fix try-with-resources issue from sonarcloud

mpeddada1:
Fixing try-with-resources issue from [sonarcloud](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTfRcB_fbtb802Xj&open=AXrlUTfRcB_fbtb802Xj): "Use try-with-resources or close this "FileOutputStream" in a "finally" clause."
I'm hoping moving outputStream into the try block should help with this.


chanseokoh:
I don't think this makes any difference.

And I think we should suppress this sonarcloud advise for this use case. We want the stream open when passing it to the constructor.

mpeddada1:
You're right. I may have opened this PR too soon. Since this was an intentional change, I will close this PR and suppress the sonarcloud warning.

PR #3368, opened 2021-08-02
fix sonar bug: add private constructor to hide the implicit public one.

zhumin8:
fix for sonar bug: https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTIQcB_fbtb802TP&open=AXrlUTIQcB_fbtb802TP

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
@@ -42,7 +42,7 @@ import org.apache.commons.text.io.StringSubstitutorReader;
 public class BuildFiles {
 
   private BuildFiles() {
-    throw new IllegalStateException("Static utility class");
+    throw new IllegalStateException("Utility class, should not be instantiated.");
   }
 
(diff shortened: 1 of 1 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/buildfile/BuildFiles.java
```diff
@@ -41,6 +41,10 @@
 /** Class to convert BuildFiles to build container representations. */
 public class BuildFiles {
 
+  private BuildFiles() {
+    throw new IllegalStateException("Utility class, should not be instantiated.");
+  }
```
chanseokoh, line 46: Unless SonarCloud insists, I think we should just do
```
private BuildFiles() {}
```

zhumin8: I think SonarCloud will be happy as long as we have a private constructor.  I just made the change.

* and my mistake, this is a "code smell" rather than a "bug" in sonar's dashboard.

chanseokoh (approved):

PR #3369, opened 2021-08-02
Add more documentation for when InterruptedException is thrown in UpdateChecker

mpeddada1:
<del>Fixes this sonarcloud [issue](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&open=AXrlUTRncB_fbtb802U6&resolved=false&severities=MAJOR&types=BUG): "Either re-interrupt this method or rethrow the "InterruptedException" that can be caught here."</del>

UPDATE : As described [here](https://github.com/GoogleContainerTools/jib/pull/3369#issuecomment-891340043) the update checker is a special case where we intend to silently consume any kind of error. Added a comment explaining this instead of restoring interrupted status.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
@@ -145,11 +145,9 @@ public class UpdateChecker {
     if (updateMessageFuture.isDone()) {
       try {
         return updateMessageFuture.get();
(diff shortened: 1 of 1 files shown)


chanseokoh:
The principle makes sense, and it may not matter if we restore the interrupted status. However, this update check is a very special task where we intend any kinds of errors are silently consumed, I feel it's safer to ignore it and stop there.

chanseokoh:
But it'd be good to comment this explicitly, so that someone doesn't re-attempt to "fix" this in the future.

chanseokoh (approved):

PR #3374, opened 2021-08-03
Remove VisibleForTesting annotation for code used in production

mpeddada1:
In order to be compliant with sonarcloud policy where class members annotated with `@VisibleForTesting` shouldn't be accessed in production code. 
For example:
- https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTJ2cB_fbtb802Tp&open=AXrlUTJ2cB_fbtb802Tp

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
@@ -61,7 +61,6 @@ public interface BuildableManifestTemplate extends ManifestTemplate {
     @SuppressWarnings("unused")
     private ContentDescriptorTemplate() {}
 
-    @VisibleForTesting
     public long getSize() {
       return size;
     }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
@@ -61,6 +61,7 @@ public interface BuildableManifestTemplate extends ManifestTemplate {
     @SuppressWarnings("unused")
     private ContentDescriptorTemplate() {}
 
+    @VisibleForTesting
     public long getSize() {
       return size;
     }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
@@ -61,7 +61,6 @@ public interface BuildableManifestTemplate extends ManifestTemplate {
     @SuppressWarnings("unused")
     private ContentDescriptorTemplate() {}
 
-    @VisibleForTesting
     public long getSize() {
       return size;
     }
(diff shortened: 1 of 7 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
```diff
@@ -61,7 +61,6 @@
     @SuppressWarnings("unused")
     private ContentDescriptorTemplate() {}
 
-    @VisibleForTesting
```
chanseokoh, line 64: For this particular one, maybe SonarCloud is confused. Making this package-private doesn't break any non-test code. It's `public` only because some tests need to access it. Maybe we should file a bug against SonarCloud?

However, I am fine with just removing the annotation if it's more of an unnecessary chore to suppress this false positive. It's a JSON template class, and it's totally OK to have getters as public. We just didn't make it `public` at the beginning based on the principle of the least privilege.

mpeddada1: I see. As you mentioned, the annotation is only used for the purpose of the class member being accessible to tests. Since this is a public getter, I think it is still worth removing the annotation as it is not needed for that purpose anymore?

chanseokoh: > the annotation is only used for the purpose of the class member being accessible to tests. ... as it is not needed for that purpose

The annotation is used for that purpose here. If there were no tests, it could be package-private instead of `public`. The only reason we make it `public` is that there is a test class wanting to use it.

However, as I said, I am fine with assuming that conceptually this class can naturally have public getters even though no prod code currently needs `public` access.

chanseokoh: > Having both `public` and `@VisibleForTesting` seems a little redundant, unless I'm missing something here.

This is what is always done when you expose a non-public method to a test that needs it. That is, adding `public` and `@VisibleForTesting` for a non-public method.

mpeddada1: Oh that makes a lot of sense now. In that case, I think you are right that this particular one should be marked as a false positive.

chanseokoh: I'm actually OK with removing it. Your call.

mpeddada1: Ah okay. I had a hard time making a decision on it because I've seen some places where we specify the productionVisibility for the code being annotated (like https://developer.android.com/reference/androidx/annotation/VisibleForTesting) which seems like a good practice. However, since that option is not available for guava I'll just remove it for now to silence sonar and avoid confusion in the future.

chanseokoh (approved):

jib-core/src/main/java/com/google/cloud/tools/jib/image/json/BuildableManifestTemplate.java
chanseokoh: BTW, it is unfortunate that SonarCloud cannot catch this. Sort of a bug there.

mpeddada1: You're right, this is indeed a limitation of the tool. I added a comment on the sonarcloud cloud issue explaining this but I'll also look into see if I can file a bug with them.

PR #3378, opened 2021-08-03
Re-interrupt threads when InterruptedException is caught

mpeddada1:
There are a few more spots in the code where we catch an InterruptedException(causing the thread state to be cleared) but don't restore the state. Re-interrupting the thread causes the interrupted state to be restored. 

- https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTJ2cB_fbtb802To&open=AXrlUTJ2cB_fbtb802To
- https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTKfcB_fbtb802T5&open=AXrlUTKfcB_fbtb802T5
- https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTLLcB_fbtb802UL&open=AXrlUTLLcB_fbtb802UL
- https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTfRcB_fbtb802Xg&open=AXrlUTfRcB_fbtb802Xg
-https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTSXcB_fbtb802VA&open=AXrlUTSXcB_fbtb802VA

### DIFF
--- a/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
+++ b/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
@@ -16,18 +16,23 @@
 
 package com.google.cloud.tools.jib.cli;
 
+import com.google.cloud.tools.jib.api.CacheDirectoryCreationException;
 import com.google.cloud.tools.jib.api.Containerizer;
+import com.google.cloud.tools.jib.api.InvalidImageReferenceException;
 import com.google.cloud.tools.jib.api.JibContainer;
 import com.google.cloud.tools.jib.api.JibContainerBuilder;
 import com.google.cloud.tools.jib.api.LogEvent;
+import com.google.cloud.tools.jib.api.RegistryException;
 import com.google.cloud.tools.jib.cli.buildfile.BuildFiles;
 import com.google.cloud.tools.jib.cli.logging.CliLogger;
 import com.google.cloud.tools.jib.plugins.common.globalconfig.GlobalConfig;
+import com.google.cloud.tools.jib.plugins.common.globalconfig.InvalidGlobalConfigException;
 import com.google.cloud.tools.jib.plugins.common.logging.ConsoleLogger;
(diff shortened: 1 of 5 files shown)


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/Build.java
```diff
@@ -144,6 +144,7 @@ public Integer call() {
       JibCli.writeImageJson(commonCliOptions.getImageJsonPath(), jibContainer);
     } catch (Exception ex) {
       JibCli.logTerminatingException(logger, ex, commonCliOptions.isStacktrace());
+      Thread.currentThread().interrupt();
```
chanseokoh, line 147: I think we should interrupt only when catching `InterruptedException`, not for all exceptions.

mpeddada1: That's a good point. I'll fix this

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
```diff
@@ -117,9 +123,18 @@ public Integer call() {
 
       JibContainer jibContainer = containerBuilder.containerize(containerizer);
       JibCli.writeImageJson(commonCliOptions.getImageJsonPath(), jibContainer);
-    } catch (Exception ex) {
+    } catch (IOException
```
chanseokoh, line 126: Probably, instead of enumerating all these, you can catch `InterruptedException` first and then catch `Exception` later. But this works too. (Need to confirm if this works as intended.)

chanseokoh (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/War.java
mpeddada1: Nice, thanks

PR #3380, opened 2021-08-03
fix for sonar: change of constant name to uppercase.

zhumin8:
Fixes [this](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrt9KgTcqoYycIcfN2P&open=AXrt9KgTcqoYycIcfN2P) sonar code smell.


chanseokoh:
I vote for not changing it. It's a public API surface of the jib-core library (the package name contains `api`), so doing this means a breaking change. Ideally we should go through the deprecation process. (Although jib-core has reached 1.0, it has been out for a very long time and considered stable.)

zhumin8:
Ah, sorry I missed this is a public API. We definitely don't want to do anything breaking here.

PR #3384, opened 2021-08-04
change to try-with-resources as FileLock implements AutoCloseable.

zhumin8:
Fixes code smell in sonar: [Change this "try" to a try-with-resources.
](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTfRcB_fbtb802Xh&open=AXrlUTfRcB_fbtb802Xh)

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
@@ -62,15 +62,8 @@ public class LockFile implements Closeable {
 
     Files.createDirectories(lockFile.getParent());
     FileOutputStream outputStream = new FileOutputStream(lockFile.toFile());
-    FileLock fileLock = null;
-    try {
-      fileLock = outputStream.getChannel().lock();
+    try (FileLock fileLock = outputStream.getChannel().lock(); ) {
       return new LockFile(lockFile, fileLock, outputStream);
-
-    } finally {
-      if (fileLock == null) {
-        outputStream.close();
-      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
```diff
@@ -62,15 +62,8 @@ public static LockFile lock(Path lockFile) throws IOException {
 
     Files.createDirectories(lockFile.getParent());
     FileOutputStream outputStream = new FileOutputStream(lockFile.toFile());
-    FileLock fileLock = null;
-    try {
-      fileLock = outputStream.getChannel().lock();
+    try (FileLock fileLock = outputStream.getChannel().lock()) {
       return new LockFile(lockFile, fileLock, outputStream);
```
suztomo, line 66: To me, this change seems to make a difference on whether the `fileLock` is closed or not

- Prev: fileLock object is not close upon returning LockFile instance.
- Now: fileLock object is closed upon returning LockFile instance.

chanseokoh: @suztomo is right. This static `lock()` method is a "constructor": to instantiate and returns a working `LockFile` object. It is the caller that should close the instance.

mpeddada1: It is interesting that this issue showed up [again](https://github.com/GoogleContainerTools/jib/pull/3367). Let's mark this as a false positive?

chanseokoh: There are actually two `AutoCloseable` classes. In #3367, it was the `OutputStream`. Here, it's `FileLock`.

mpeddada1: oh good eye! I missed this.

zhumin8: Ah, I missed the other related PR. Will make a note in sonar and close this one too.

PR #3392, opened 2021-08-06
Avoid using wildcards in the return type. 

mpeddada1:
Resolves sonarcloud [item](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTY4cB_fbtb802WJ&open=AXrlUTY4cB_fbtb802WJ): "Remove usage of generic wildcard type"

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslatorTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslatorTest.java
@@ -110,7 +110,7 @@ public class JsonToImageTranslatorTest {
 
   @Test
   public void testPortMapToList() throws BadContainerConfigurationFormatException {
-    ImmutableSortedMap<String, Map<?, ?>> input =
+    ImmutableSortedMap<String, Map<String, String>> input =
         ImmutableSortedMap.of(
             "1000",
             ImmutableMap.of(),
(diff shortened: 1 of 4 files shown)


jib-core/src/test/java/com/google/cloud/tools/jib/image/json/JsonToImageTranslatorTest.java
```diff
@@ -148,12 +148,12 @@ public void testVolumeMapToList() throws BadContainerConfigurationFormatExceptio
...
     Assert.assertEquals(expected, JsonToImageTranslator.volumeMapToSet(input));
 
-    ImmutableList<Map<String, Map<?, ?>>> badInputs =
+    ImmutableList<Map<String, Map<String, String>>> badInputs =
         ImmutableList.of(
             ImmutableMap.of("var/job-result-data", ImmutableMap.of()),
             ImmutableMap.of("log", ImmutableMap.of()),
             ImmutableMap.of("C:/udp", ImmutableMap.of()));
-    for (Map<String, Map<?, ?>> badInput : badInputs) {
+    for (Map<String, Map<String, String>> badInput : badInputs) {
```
chanseokoh, line 156: Maybe we can keep the original code for the cases like in these tests. The sonarcloud is only talking about a wildcard in a return type of a method. I think if the type is obvious in a small local scope, I think it reduces the boilerplate and cognitive load (as long as sonarcloud doesn't complain).

mpeddada1: That's a good observation. Unfortunately, `volumeMapToSet` takes in `containerConfigurationTemplate.getContainerVolumes` as an input in line 201 of `JsonToImageTranslator` so it requires the parameter type to be Map<String, Map<String, String>> . The code doesn't compile when we give it an input that's of type `Map<String, Map<?,?>>`

chanseokoh (approved):

PR #3394, opened 2021-08-06
Fix sonar code smell: use log

zhumin8:
Fix sonar issues:
- [Replace this use of System.out or System.err by a logger.](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTQvcB_fbtb802U2&open=AXrlUTQvcB_fbtb802U2)
- [Use the built-in formatting to construct this argument.](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTfhcB_fbtb802Xm&open=AXrlUTfhcB_fbtb802Xm) X2

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
@@ -23,7 +23,6 @@ import java.nio.file.Paths;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 
 /**
@@ -133,7 +132,7 @@ public class XdgDirectories {
       }
       Path localAppData = Paths.get(localAppDataEnv);
       if (!Files.exists(localAppData)) {
-        LOGGER.log(Level.WARNING, "{} does not exist", localAppData);
+        LOGGER.warning(() -> localAppData + " does not exist");
         return xdgPath.resolve(windowsSubDirectory);
       }
       return localAppData.resolve(windowsSubDirectory);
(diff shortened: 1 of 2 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/XdgDirectories.java
```diff
@@ -132,7 +133,7 @@ private static Path getOsSpecificDirectory(
       }
       Path localAppData = Paths.get(localAppDataEnv);
       if (!Files.exists(localAppData)) {
-        LOGGER.warning(localAppData + " does not exist");
+        LOGGER.log(Level.WARNING, "{} does not exist", localAppData);
```
chanseokoh, line 136: Are you sure `{}` works? The [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html) of JUL (aka `java.util.logging`) `Logger` says

> Typically, formatters use java.text.MessageFormat style formatting to format parameters, so for example a format string `"{0} {1}"` would format two parameters as strings.

Reading the `MessageFormat` [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/text/MessageFormat.html), I think it should contain a number?

Also, doesn't `LOGGER.warning()` have an override that accepts message arguments?

zhumin8: > Are you sure {} works?

Thanks for pointing out! I was mistaken with log4j formats. And you are right,`{0}` is the way to go.
> Also, doesn't LOGGER.warning() have an override that accepts message arguments?

Ah, perfect. I think this is what I needed. `warning(Supplier<String> msgSupplier)` this accepts a supplier and will [only to be constructed if the logging level is such](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Logger.html#warning-java.util.function.Supplier-).

chanseokoh (approved):

PR #3400, opened 2021-08-10
Remove use of wildcards from return types

mpeddada1:
Removes the use of wildcards from the following classes:
- [`RawConfiguration`](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTSscB_fbtb802VE&open=AXrlUTSscB_fbtb802VE)
- [`ContainerBuildPlan`](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrt9KgqcqoYycIcfN2S&open=AXrt9KgqcqoYycIcfN2S)
- [`TimerEvent`](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrlUTfGcB_fbtb802Xf&open=AXrlUTfGcB_fbtb802Xf)

### DIFF
--- a/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
+++ b/jib-build-plan/src/main/java/com/google/cloud/tools/jib/api/buildplan/ContainerBuildPlan.java
@@ -479,7 +479,7 @@ public class ContainerBuildPlan {
     return cmd == null ? null : new ArrayList<>(cmd);
   }
 
-  public List<? extends LayerObject> getLayers() {
+  public List<LayerObject> getLayers() {
     return new ArrayList<>(layers);
   }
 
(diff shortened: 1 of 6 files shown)


chanseokoh:
I think some of these are not safe. For example, particularly for the public API that concerns `LayerObject`. `List<LayerObject>` and `List<? extends LayerObject>` are two different types, although I'm not certain if it's a safe change for a return type. I'd like to hold off.

And I also don't like casting. It adds warnings.

mpeddada1:
That's valid. We should be careful when changing the return type of a public method. The sonar suggestion to remove the wildcard is helpful when implementing a new method but it can be a risky move to apply it to an existing method.  
As far as the casting is concerned, this was safest way avoid using wildcards but I also wasn't sure if the tradeoff would be worth it. It is also not great in terms of readability.

I will hold off on making this change.

PR #3413, opened 2021-08-12
Sonar security: replaced regex.

zhumin8:
Remove use of regex in response to [this](https://sonarcloud.io/project/security_hotspots?id=GoogleContainerTools_jib&hotspots=AXrt9KAtcqoYycIcfN2F) sonar security hotspot review:

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -506,7 +506,8 @@ public class PluginConfigurationProcessor {
     }
 
     // Verify Java version is compatible
-    String prefixRemoved = baseImageConfig.replaceFirst(".*://", "");
+    String[] splits = baseImageConfig.split("://");
+    String prefixRemoved = splits[splits.length - 1];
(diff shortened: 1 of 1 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -506,7 +506,8 @@ static JavaContainerBuilder getJavaContainerBuilderWithBaseImage(
     }
 
     // Verify Java version is compatible
-    String prefixRemoved = baseImageConfig.replaceFirst(".*://", "");
+    String[] splits = baseImageConfig.split("://");
+    String prefixRemoved = splits[splits.length - 1];
```
suztomo, line 510: In general, Guava's Splitter gives more expected behavior.
https://github.com/google/guava/wiki/StringsExplained#splitter

suztomo (approved):
Guava or not, it looks good to me.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
zhumin8: Thanks for the link!  Updated to Guava Splitter.

PR #3423, opened 2021-08-13
 fix to keep only one method invocation in try/catch testing for exception

zhumin8:
Fix sonar issue: [Refactor the body of this try/catch to have only one invocation possibly throwing a runtime exception.](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AXrt9KfgcqoYycIcfN2M&open=AXrt9KfgcqoYycIcfN2M).

### DIFF
--- a/jib-build-plan/src/test/java/com/google/cloud/tools/jib/api/buildplan/AbsoluteUnixPathTest.java
+++ b/jib-build-plan/src/test/java/com/google/cloud/tools/jib/api/buildplan/AbsoluteUnixPathTest.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.jib.api.buildplan;
 
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.junit.Assert;
 import org.junit.Assume;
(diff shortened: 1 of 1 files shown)


jib-build-plan/src/test/java/com/google/cloud/tools/jib/api/buildplan/AbsoluteUnixPathTest.java
```diff
@@ -75,13 +76,14 @@ public void testResolve_relativeUnixPath() {
 
   @Test
   public void testResolve_Path_notRelative() {
+    AbsoluteUnixPath absoluteUnixPath = AbsoluteUnixPath.get("/");
+    Path path = Paths.get("/not/relative");
     try {
-      AbsoluteUnixPath.get("/").resolve(Paths.get("/not/relative"));
+      absoluteUnixPath.resolve(path);
```
mpeddada1, line 82: How about we replace the use of the try-catch here with the following?:
```
IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> absoluteUnixPath.resolve(path));
assertThat(exception).hasMessageThat().isEqualTo("Cannot resolve against absolute Path: " + path, ex.getMessage());
```

zhumin8: Ah, I like that `assertThrows` avoids the try catch block. Let me update with it.

mpeddada1 (approved):

PR #3424, opened 2021-08-13
Network retry on IO errors only after insecure failover is finalized

### LINKED ISSUE #3422: Network retry on IOException (new in Jib 3.1.3) significantly slows down insecure registry failover (allowInsecureRegistries=true)
Problem identified in https://github.com/GoogleContainerTools/jib/issues/3409#issuecomment-898677291.

We implemented retrying in #3351 (and a clean-up #3390) but overlook the issue when `allowInsecureRegistries=true`, where insecure connection failover depends on `SSLException` (a subclass of `IOException`). Consequently, HTTP failover can happen only after exhausting retrying.
```
Base image 'amazoncorretto:16' does not use a specific image digest - build may not be reproducible
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will NOT be retried
Cannot verify server at https://localhost:5000/v2/. Attempting again with no TLS verification.
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will be retried
GET https://localhost:5000/v2/ failed and will NOT be retried
Failed to connect to https://localhost:5000/v2/ over HTTPS. Attempting again with HTTP.
```

chanseokoh:
Fixes #3422.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
@@ -133,7 +133,7 @@ public class FailoverHttpClient {
 
   private final Deque<HttpTransport> transportsCreated = new ArrayDeque<>();
   private final Deque<Response> responsesCreated = new ArrayDeque<>();
-  private final boolean turnOffRetry;
+  private final boolean enableRetry;
 
   /**
    * Create a new FailoverHttpclient.
@@ -146,7 +146,7 @@ public class FailoverHttpClient {
       boolean enableHttpAndInsecureFailover,
       boolean sendAuthorizationOverHttp,
       Consumer<LogEvent> logger) {
-    this(enableHttpAndInsecureFailover, sendAuthorizationOverHttp, logger, false);
+    this(enableHttpAndInsecureFailover, sendAuthorizationOverHttp, logger, true);
   }
 
   @VisibleForTesting
@@ -154,14 +154,14 @@ public class FailoverHttpClient {
       boolean enableHttpAndInsecureFailover,
       boolean sendAuthorizationOverHttp,
       Consumer<LogEvent> logger,
-      boolean turnOffRetry) {
+      boolean enableRetry) {
     this(
         enableHttpAndInsecureFailover,
         sendAuthorizationOverHttp,
         logger,
         FailoverHttpClient::getSecureHttpTransport,
         FailoverHttpClient::getInsecureHttpTransport,
-        turnOffRetry);
+        enableRetry);
   }
 
   @VisibleForTesting
@@ -171,13 +171,13 @@ public class FailoverHttpClient {
       Consumer<LogEvent> logger,
       Supplier<HttpTransport> secureHttpTransportFactory,
       Supplier<HttpTransport> insecureHttpTransportFactory,
-      boolean turnOffRetry) {
+      boolean enableRetry) {
     this.enableHttpAndInsecureFailover = enableHttpAndInsecureFailover;
     this.sendAuthorizationOverHttp = sendAuthorizationOverHttp;
     this.logger = logger;
     this.secureHttpTransportFactory = secureHttpTransportFactory;
     this.insecureHttpTransportFactory = insecureHttpTransportFactory;
-    this.turnOffRetry = turnOffRetry;
+    this.enableRetry = enableRetry;
   }
 
   /**
@@ -335,7 +335,7 @@ public class FailoverHttpClient {
             .buildRequest(httpMethod, new GenericUrl(url), request.getHttpContent())
             .setUseRawRedirectUrls(true)
             .setHeaders(requestHeaders);
-    if (!turnOffRetry && retryOnIoException) {
+    if (enableRetry && retryOnIoException) {
       httpRequest.setIOExceptionHandler(createBackOffRetryHandler());
     }
     if (request.getHttpTimeout() != null) {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
@@ -315,12 +315,14 @@ public class FailoverHttpClient {
     }
   }
 
+  // TODO: remove retryOnIoException and turn on/off retry based on whether it's an SSLException or
+  // not: https://github.com/GoogleContainerTools/jib/issues/3422
   private Response call(
       String httpMethod,
       URL url,
       Request request,
       HttpTransport httpTransport,
-      boolean retryOnIoException)
+      boolean retryOnIoException) // See https://github.com/GoogleContainerTools/jib/issues/3424
       throws IOException {
     boolean clearAuthorization = !isHttpsProtocol(url) && !sendAuthorizationOverHttp;
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
@@ -128,12 +128,12 @@ public class FailoverHttpClient {
   private final Consumer<LogEvent> logger;
   private final Supplier<HttpTransport> secureHttpTransportFactory;
   private final Supplier<HttpTransport> insecureHttpTransportFactory;
-  private final boolean retryOnIoException;
 
   private final ConcurrentHashMap<String, Failover> failoverHistory = new ConcurrentHashMap<>();
 
   private final Deque<HttpTransport> transportsCreated = new ArrayDeque<>();
   private final Deque<Response> responsesCreated = new ArrayDeque<>();
+  private final boolean turnOffRetry;
 
   /**
    * Create a new FailoverHttpclient.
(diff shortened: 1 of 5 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -128,12 +128,12 @@ private static HttpTransport getInsecureHttpTransport() {
   private final Consumer<LogEvent> logger;
   private final Supplier<HttpTransport> secureHttpTransportFactory;
   private final Supplier<HttpTransport> insecureHttpTransportFactory;
-  private final boolean retryOnIoException;
 
   private final ConcurrentHashMap<String, Failover> failoverHistory = new ConcurrentHashMap<>();
 
   private final Deque<HttpTransport> transportsCreated = new ArrayDeque<>();
   private final Deque<Response> responsesCreated = new ArrayDeque<>();
+  private final boolean turnOffRetry;
```
chanseokoh, line 136: This is now like a global and ultimate turn-off switch (only there for making some tests work).

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -252,7 +252,7 @@ public Response put(URL url, Request request) throws IOException {
   public Response call(String httpMethod, URL url, Request request) throws IOException {
     if (!isHttpsProtocol(url)) {
       if (enableHttpAndInsecureFailover) { // HTTP requested. We only care if HTTP is enabled.
-        return call(httpMethod, url, request, getHttpTransport(true));
+        return call(httpMethod, url, request, getHttpTransport(true), true);
```
chanseokoh, line 255: This is a terminal call to connect to plain HTTP (i.e., no insecure failover necessary in response to `SSLException`) , so we can enable retrying (`true` for the last argument).

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -263,7 +263,7 @@ public Response call(String httpMethod, URL url, Request request) throws IOExcep
     }
 
     try {
-      return call(httpMethod, url, request, getHttpTransport(true));
+      return call(httpMethod, url, request, getHttpTransport(true), !enableHttpAndInsecureFailover);
```
chanseokoh, line 266: No history for this server, so we enable retrying only when failover is disabled (i.e., if HTTPS fails, then it means failure). OTOH, if failover is enabled, we don't want to retry so that we can quickly failover to test insecure HTTPS.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -272,13 +272,13 @@ public Response call(String httpMethod, URL url, Request request) throws IOExcep
 
       try {
         logInsecureHttpsFailover(url);
-        Response response = call(httpMethod, url, request, getHttpTransport(false));
+        Response response = call(httpMethod, url, request, getHttpTransport(false), false);
```
chanseokoh, line 275: Not a terminal call, we may have to failover to plain HTTP once more, so disable retrying.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -272,13 +272,13 @@ public Response call(String httpMethod, URL url, Request request) throws IOExcep
...
         failoverHistory.put(url.getHost() + ":" + url.getPort(), Failover.INSECURE_HTTPS);
         return response;
 
       } catch (SSLException ignored) { // This is usually when the server is plain-HTTP.
         logHttpFailover(url);
-        Response response = call(httpMethod, toHttp(url), request, getHttpTransport(true));
+        Response response = call(httpMethod, toHttp(url), request, getHttpTransport(true), true);
```
chanseokoh, line 281: Terminal plain-HTTP call. Retrying can be safely enabled.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -293,7 +293,7 @@ public Response call(String httpMethod, URL url, Request request) throws IOExcep
         // port 443) and we could not connect to 443. It's worth trying port 80.
         if (enableHttpAndInsecureFailover && isHttpsProtocol(url) && url.getPort() == -1) {
           logHttpFailover(url);
-          Response response = call(httpMethod, toHttp(url), request, getHttpTransport(true));
+          Response response = call(httpMethod, toHttp(url), request, getHttpTransport(true), true);
```
chanseokoh, line 296: Terminal plain-HTTP call. Retrying can be safely enabled.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
```diff
@@ -307,15 +307,20 @@ public Response call(String httpMethod, URL url, Request request) throws IOExcep
     Preconditions.checkArgument(isHttpsProtocol(url));
     switch (failoverHistory.getOrDefault(url.getHost() + ":" + url.getPort(), Failover.NONE)) {
       case HTTP:
-        return Optional.of(call(httpMethod, toHttp(url), request, getHttpTransport(true)));
+        return Optional.of(call(httpMethod, toHttp(url), request, getHttpTransport(true), true));
       case INSECURE_HTTPS:
-        return Optional.of(call(httpMethod, url, request, getHttpTransport(false)));
+        return Optional.of(call(httpMethod, url, request, getHttpTransport(false), true));
```
chanseokoh, line 312: If in the history, then the protocol is already fixed for the server. (That is, we've gone through failover before and the result is cached.) These are terminal calls, so enable retrying.

jib-core/src/test/java/com/google/cloud/tools/jib/http/WithServerFailoverHttpClientTest.java
```diff
@@ -70,7 +70,7 @@ public void testGet()
   public void testSecureConnectionOnInsecureHttpsServer()
       throws IOException, InterruptedException, GeneralSecurityException, URISyntaxException {
     FailoverHttpClient secureHttpClient =
-        new FailoverHttpClient(false /*secure*/, false, logger, false);
+        new FailoverHttpClient(false /*secure*/, false, logger, true);
```
chanseokoh, line 73: This one particularly needs turning off retrying. (The test still works without turning off, but it takes a lot of time before retrying finally gives up.)

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
mpeddada1: qq: I could be missing some context here. What if we make this `turnOnRetry` instead?

chanseokoh: So my thinking for `turnOff` and some context: if we didn't have tests, and we wouldn't have this field. This PR also introduces another variable (`retryOnIoException`) in the method below, so it could a bit be confusing if we have both "turn on retry" and "retry on something", but maybe that's only me. But more relevantly, I thought `turnOffRetry` sort of implies that retry is on by default (which is what we want and will always be the case if it were not for tests), and if you don't want retrying for some reason, then you have to explicitly turn it off. But if people with a fresh eye without the context may think this is more confusing, I can just flip it.

mpeddada1: I see! Your logic of having the retry on by default makes sense but I usually find it a little hard to parse negative booleans because the meaning of `true` or `false` for such variables gets inverted. Or perhaps we could name this `enableRetry`? `enableRetry` gives it more global sense and `retryOnIOException` is specific to a particular case.

chanseokoh: Sounds good. Done.

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
mpeddada1: A comment mentioning this could be helpful for future readers. WDYT?

mpeddada1 (approved):

jib-core/src/main/java/com/google/cloud/tools/jib/http/FailoverHttpClient.java
chanseokoh: Good idea. Since the argument is to be removed eventually, and also for brevity, added a link to this PR in one of the comments.

Issue #2751, opened 2020-09-03
Investigate what to have as image build output (jib-image.json, jib-image.digest, and jib-image.id) for multi-platform image building


chanseokoh:
Should we write multiple `jib-image.digest` and `jib-image.id` files for each manifest? What about the file names? Should we also write a digest of a manifest list (potentially with a different name)? Or, should we write only a single digest for a manifest list?

Does it make sense change the JSON format of `jib-image.json`? Should we keep backward compatibility?

On the technical side, we need to figure out the right return type of `PushImageStep` after sorting out these questions:

https://github.com/GoogleContainerTools/jib/blob/94f8bdafe0f816c548f96392b34660d5574cc5c4/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PushImageStep.java#L40-L46

tom-haines:
@chanseokoh if relevant, whilst docker exporter `does not currently support exporting manifest lists` it does allow to output tar bundles, but only supports one platform at a time.  

Example buildx commands re `amd64` first, and `arm64` at end:

```bash
docker buildx build --platform linux/amd64 -o type=docker,dest=/tmp/helloworld/amd64.tar .
tar xf -C amd64 amd64.tar
tree .
```

output:
```text
├── Dockerfile
├── amd64
│   ├── blobs
│   │   └── sha256
│   │       ├── 16ec32c2132b43494832a05f2b02f7a822479f8250c173d0ab27b3de78b2f058
│   │       ├── 1b307460a9df24f41e954a977e17a389633f6cde89fd75efa93b4259a69f8966
│   │       ├── 34f1332d72bf8b6edcc1a4d11f3310cb84c0a69b6f9374dad77253ad07c4f308
│   │       └── f90645af5f5b704bdd16861777f8ac9c50a405a397b7cb32d6cf34b6f4b928ee
│   ├── index.json
│   ├── manifest.json
│   └── oci-layout
└── amd64.tar
```

```bash
$ cat amd64/manifest.json | jq -r
```

output:
```json
[
  {
    "Config": "blobs/sha256/34f1332d72bf8b6edcc1a4d11f3310cb84c0a69b6f9374dad77253ad07c4f308",
    "RepoTags": null,
    "Layers": [
      "blobs/sha256/16ec32c2132b43494832a05f2b02f7a822479f8250c173d0ab27b3de78b2f058",
[...]

PR #3442, opened 2021-08-23
Improve digest discovery portion of docs

tom-haines:
Improvement to documentation regarding digest discovery during [#2751 discussion](https://github.com/GoogleContainerTools/jib/issues/2751).

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -558,33 +558,20 @@ $ docker manifest inspect openjdk:8
 ```
 
 You can inspect the digest (`@sha256:...`) of a manifest image with the `--verbose` option.
-```
-$ docker manifest inspect --verbose openjdk:8
-[
-	{
-		"Ref": "docker.io/library/openjdk:8@sha256:74a0711547137cd3787c0b84266a7ba4cb814b98b8950cf95bc2e526bfc05e47",
-    ...
-  }
-]    
-   ...
-   // This confirms that openjdk:8 points to a manifest list.
-   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
-   "manifests": [
-      {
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -557,6 +557,35 @@ $ docker manifest inspect openjdk:8
...
+         ...
+         "digest": "sha256:1fbd49e3fc5e53154fa93cad15f211112d899a6b0c5dc1e8661d6eb6c18b30a6",
+         "platform": {
+            "architecture": "arm64",
+            "os": "linux",
+            "variant": "v8"
+         }
+      }
+   ]
+}
```
mpeddada1, line 585: Perhaps we can remove this section since we've already mentioned this above.

mpeddada1 (commented):
Thank you for making our docs more comprehensive!

docs/faq.md
tom-haines: Good suggestion thanks @mpeddada1 .  I've updated to make relevance of output clearer  (eee6882), and removed duplication.

docs/faq.md
```diff
@@ -557,6 +557,22 @@ $ docker manifest inspect openjdk:8
 }
 ```
 
+You can inspect the digest (`@sha256:...`) of a manifest image with the `--verbose` option.
```
chanseokoh, line 560: I don't think this is necessary. Without `--verbose`, you can already see the digest of each image in the list already as seen above.

mpeddada1:
@tom-haines Apologies for making you do the edits. But after seeing @chanseokoh's comment, I agree that including the `--verbose` flag may be a little redundant since info about the digest is already present above. Closing this PR but appreciate your attempt.

PR #3476, opened 2021-10-04
Upgrade maven-core and maven-compat to 3.8.3 and update unit test setup

mpeddada1:
More details in #3473 
Maven-core 3.8.3 upgraded `plexus-sec-dispatcher` to 2.0 which introduced two constructors for `DefaultSecDispatcher`. This causes the current test setup to break.


suztomo (approved):

mpeddada1:
Thank you!

PR #3496, opened 2021-11-09
Refactor tests

chanseokoh:
Only refactoring. No behavioral changes.

### DIFF
--- a/jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
+++ b/jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
@@ -29,13 +29,13 @@ import com.google.cloud.tools.jib.api.buildplan.ImageFormat;
 import com.google.common.collect.ImmutableList;
 import java.nio.file.Paths;
 import java.time.Instant;
-import java.util.Arrays;
-import java.util.Collection;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
 
 /** Tests for {@link BuildFileSpec}. */
+@RunWith(JUnitParamsRunner.class)
 public class BuildFileSpecTest {
 
   private static final ObjectMapper mapper = new ObjectMapper(new YAMLFactory());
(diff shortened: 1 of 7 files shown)


jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
```diff
@@ -169,146 +169,112 @@ public void testBuildFileSpec_nullCollections() throws JsonProcessingException {
...
-
-    @Test
-    public void testBuildFileSpec_emptyOkay() throws JsonProcessingException {
-      String data = "apiVersion: v1alpha1\n" + "kind: BuildFile\n" + fieldName + ": []";
-
-      assertThat(mapper.readValue(data, BuildFileSpec.class)).isNotNull();
-    }
-
-    @Test
-    public void testBuildFileSpec_nullOkay() throws JsonProcessingException {
```
chanseokoh, line 215: This method is merged into the top-level method of the same name.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
```diff
@@ -169,146 +169,112 @@ public void testBuildFileSpec_nullCollections() throws JsonProcessingException {
...
-      JsonProcessingException exception =
-          assertThrows(
-              JsonProcessingException.class, () -> mapper.readValue(data, BuildFileSpec.class));
-      assertThat(exception)
-          .hasMessageThat()
-          .contains("Property '" + fieldName + "' cannot be an empty string");
-    }
-
-    @Test
-    public void testBuildFileSpec_nullOkay() throws JsonProcessingException {
```
chanseokoh, line 245: ditto

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
```diff
@@ -169,146 +169,112 @@ public void testBuildFileSpec_nullCollections() throws JsonProcessingException {
...
-
-    @Test
-    public void testBuildFileSpec_emptyOkay() throws JsonProcessingException {
-      String data = "apiVersion: v1alpha1\n" + "kind: BuildFile\n" + fieldName + ": {}";
-
-      assertThat(mapper.readValue(data, BuildFileSpec.class)).isNotNull();
-    }
-
-    @Test
-    public void testBuildFileSpec_nullOkay() throws JsonProcessingException {
```
chanseokoh, line 308: ditto

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
```diff
@@ -93,7 +93,7 @@ public void testBuildFileSpec_full() throws JsonProcessingException {
   public void testBuildFileSpec_apiVersionRequired() {
     String data = "kind: BuildFile\n";
 
-    JsonProcessingException exception =
+    Exception exception =
```
mpeddada1, line 96: Can we leave these as `JsonProcessingException`?

mpeddada1 (approved):
Thank you for doing this! The tests look so nice now.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
```diff
@@ -169,146 +169,112 @@ public void testBuildFileSpec_nullCollections() throws JsonProcessingException {
...
-
-    @Test
-    public void testBuildFileSpec_nullOkay() throws JsonProcessingException {
-      String data = "apiVersion: v1alpha1\n" + "kind: BuildFile\n" + fieldName + ": null";
-
-      assertThat(mapper.readValue(data, BuildFileSpec.class)).isNotNull();
-    }
+  @Test
+  @Parameters(value = {"volumes", "exposedPorts", "entrypoint", "cmd"})
+  public void testBuildFileSpec_noNullEntries(String fieldName) {
```
mpeddada1, line 174: Oops sorry I missed this. Perhaps we can combine this test with the one below. Similar to what sonarcloud [suggested](https://sonarcloud.io/project/issues?id=GoogleContainerTools_jib&issues=AX0Fd6hL2M0IIkbbKH6o&open=AX0Fd6hL2M0IIkbbKH6o&pullRequest=3496)

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
chanseokoh: What's the reason for that? We are only checking the exception message. In some sense, it's similar to `List list = ...` instead of `ArrayList list = ...`.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
chanseokoh: They are testing different things and their error messages are different. I think merging them into one is too extreme. The current granularity seems good.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
mpeddada1: This might be more of a personal choice but sometimes it can be helpful to be specific about the exception being returned. Sometimes our method can return a different exception, causing the tests to pass even when it is not supposed to. However, in this case it might be okay.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
mpeddada1: Sounds good.

jib-cli/src/test/java/com/google/cloud/tools/jib/cli/buildfile/BuildFileSpecTest.java
chanseokoh: Oh, that's not a concern. We explicitly define the expected exception as the first argument for `assertThrows`. It's only that we just use the type `Exception` to hold the returned instance. The instance will always be `JsonProcessingException`.
```java
    Exception exception =
        assertThrows(
            JsonProcessingException.class, () -> mapper.readValue(data, BuildFileSpec.class));
```

mpeddada1: Ah yes, you're right. In that case, using a generic type shouldn't be problem.

PR #3500, opened 2021-11-11
Add enabling container registry FAQ answer

Mukamik:


docs/faq.md
```diff
@@ -609,6 +609,8 @@ Depending on registry implementations, it is also possible that the registry act
 
 If the registry returns `401 Unauthorized` or `"code":"UNAUTHORIZED"`, it is often due to credential misconfiguration. Examples:
 
+* The Container Registry API is not yet enabled for your project.
```
chanseokoh, line 612: Let's make this the second bullet point (before "`$HOME/.docker/config.json` may also contain ..."), since the first one is the most important information for most users. Also, I suggest

"For Google Cloud Registry (GCR), the Container Registry API is ..."

And

"You Enable the API" --> "You enable the API"

chanseokoh (approved):
LGTM, only minor comments.

docs/faq.md
```diff
@@ -609,6 +609,8 @@ Depending on registry implementations, it is also possible that the registry act
...
+   - You Enable the API from [Cloud Console](https://console.cloud.google.com/flows/enableapi?apiid=containerregistry.googleapis.com&_ga=2.101495560.861748223.1636407745-651361740.1631828346) or with the following [Cloud SDK](https://cloud.google.com/sdk/docs) command: `gcloud services enable containerregistry.googleapis.com`
```
suztomo, line 613: Do you need the _ga parameter?

suztomo:
The URL has _ga parameter

docs/faq.md
chanseokoh: Probably not. Good catch.

docs/faq.md
Mukamik: Good feedback, thanks.
Updated the md, resolving this conversation.

chanseokoh (approved):

Issue #2275, opened 2020-02-10
Jib should maintain hard-links and symlinks


briandealwis:
The hardlinking supported by the Spotify docker-client can [result in large image-size savings](https://twitter.com/jvanzyl/status/1226511828110467072) for situations using lots of hard-links (https://github.com/jvanzyl/dockerfile-maven/issues/2).  Jib should [track hard-links](https://github.com/jvanzyl/docker-client/blob/8da77cf5952f1819e8afeda0cebd2b61bac858c7/src/main/java/com/spotify/docker/client/CompressedDirectory.java#L302-L309).  Jib should also preserve symlinks (perhaps only if they reference within the layer?).

jvanzyl:
Great savings in the time to produce the image as well, give far less material is transferred.

TendoPW:
Hello, I would like to work on this issue. I am new to Jib and now I am confused about where should I modify in the code to support hard-link. I tried to read the code but I haven't found it. Could you please give me some help?

briandealwis:
The relevant starting point is `com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer#addEntryRecursive()`.

dfelliott:
It is trivial to add support for symlinks to `com.google.cloud.tools.jib.image.ReproducibleLayerBuilder#build`.

Just use the standard Java `Files.isSymbolicLink` API on the `sourceFile` and if it is a symlink create a `TarArchiveEntry` for a symlink using the result of `Files.readSymbolicLink`.  If not a symlink, the existing code for creating a `TarArchiveEntry` is correct.

There's a little bit more to it if the source filesystem isn't UNIX or if the target of the link is not a relative path.  Still, it would be nice if ordinary symlinks with relative targets "just worked" and it's a pretty trivial change here to make that happen.  One might argue if the path is absolute then presumably it's already intended for the container so nothing special has to be done with it.

There is one major catch though: the current behavior is to follow symlinks, so changing will break any builds that rely on the follow behavior.

meltsufin:
@dfelliott Thanks for the idea. However, this FR is not currently prioritized.

chanseokoh:
We gave some thought on this in the past, and we think we want [a new dedicated configuration and API in jib-core](https://github.com/GoogleContainerTools/jib/pull/3506#issuecomment-978402386).

Issue #1576, opened 2019-03-24
extraDirectory doesn't work with destination as symbolic link


prashantpro:
**Description of the issue**:
I have a maven project structure and have used the default jib location of extraDirectory.

src/main/jib/
Added my desired file i.e. 'server.xml' that needed to be placed under '/config' within the base image.
Thus, the structure now looked like this: (As jib requires recreating the structure under 'jib' directory)

src/main/jib/config/server.xml

The destination directory '/config' is actually a symbolic link in the base image.
After jib creates the image, on inspecting it can be seen that the symbolic link has now got replaced with an actual directory. Which breaks the build for me.

**Expected behavior**:

Symbolic link directories are typical in Java EE server images where these links are provided for deployment or configuration file placement.
Here the 'config' directory is the case for openliberty server.

Would expect this structure to be preserved if possible. Or provide some means of achieving the goal of putting files under images which use a symbolic link as a destination folder.

**Environment**: 
Maven 3.5
Java 8
OS - MacOS

**`jib-maven-plugin` Configuration**:
```xml
<plugin>
                <groupId>com.google.cloud.tools</groupId>
                <artifactId>jib-maven-plugin</artifactId>
                <version>1.0.2</version>
                <configuration>
                    <from>
                        <image>openliberty/open-liberty:javaee8-java8-openj9</image>
                    </from>
                    <to>
                        <image>demo-project:1.0</image>
                    </to>
                    <container>
                        <appRoot>/apps/</appRoot>                        
                    </container>
                </configuration>
            </plugin>
```

**Additional Information**: 
I inspected the image by logging into it with bash (running container). And I can see that the symbolic link 'config' within the server, now got changed to an actual directory.

Let me know if you need more details? This actually should be easily reproducible without requiring any Java or server technology, it's just a matter of how JIB copies the structure on the base image.

loosebazooka:
Hey @prashantpro. Thanks for bringing this up.

The workaround for this is to just directly reference the target that the link is referencing. I image that would be annoying if you're swapping servers regularly but might be acceptable for a simple case.

We currently have this low priority on our schedule, and given how jib works might consider not solving this. However if we get enough attention we can take another look. So please thumbs up this issue if it is seriously affecting you.

prashantpro:
Low priority is fine, I don't think this is a deal breaker.
And yes, that workaround suggested is what I had ended up doing to move ahead.
I brought this up as an issue in case it was something simple to solve that the team could have closed.

[...]

PR #3506, opened 2021-11-16
Adding symbolic link support.

seanleblanc:
Related to: https://github.com/GoogleContainerTools/jib/issues/2275

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
@@ -25,6 +25,7 @@ import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -33,6 +34,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarConstants;
 
 /**
  * Builds a reproducible layer {@link Blob} from files. The reproducibility is implemented by strips
@@ -147,12 +149,21 @@ public class ReproducibleLayerBuilder {
 
     // Adds all the layer entries as tar entries.
     for (FileEntry layerEntry : layerEntries) {
-      // Adds the entries to uniqueTarArchiveEntries, which makes sure all entries are unique and
-      // adds parent directories for each extraction path.
-      TarArchiveEntry entry =
-          new TarArchiveEntry(
-              layerEntry.getSourceFile(), layerEntry.getExtractionPath().toString());
 
+      TarArchiveEntry entry;
+      if (Files.isSymbolicLink(layerEntry.getSourceFile())) {
+        entry =
+                new TarArchiveEntry(
+                        layerEntry.getExtractionPath().toString(), TarConstants.LF_SYMLINK );
+        Path targetPath = Files.readSymbolicLink(layerEntry.getSourceFile());
+        entry.setLinkName(targetPath.toString());
+      } else {
+        // Adds the entries to uniqueTarArchiveEntries, which makes sure all entries are unique and
+        // adds parent directories for each extraction path.
+        entry =
+                new TarArchiveEntry(
+                        layerEntry.getSourceFile(), layerEntry.getExtractionPath().toString());
+      }
       // Sets the entry's permissions by masking out the permission bits from the entry's mode (the
       // lowest 9 bits) then using a bitwise OR to set them to the layerEntry's permissions.
       entry.setMode((entry.getMode() & ~0777) | layerEntry.getPermissions().getPermissionBits());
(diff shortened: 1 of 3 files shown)


chanseokoh:
Thank you for taking a stab at this, and sorry for the late follow-up. I've mulled over this for a while. Although this is a possible implementation for the support, I actually envisioned a different approach in the past: to have a [dedicated configuration in the same way that the Bazel `docker_rules` takes](https://github.com/bazelbuild/rules_docker/blob/master/docs/container.md#container_layer-symlinks). Rationale:
- This is a breaking change, and there may be people relying on the current behavior. (However, I'm not saying we cannot make a breaking change.)
- The approach in this PR is universal and absolute, and there's no way to opt out of this behavior. Some people may not want to create symlinks but copy their referenced files.
- It affects source files and directories for all layers, while I think it's safer to restrict it to "extraDirectories."
- On Windows, it can be cumbersome to create symlinks (hence cumbersome to create an image that has symlinks). Even on Linux, I think creating symlinks through configuration and explicit jib-core API is more convenient and robust.
- Bazel `docker_rules` might have additional reasons to take such an approach that I haven't thought about.
- Configuration and explicit API eliminate ambiguity: otherwise, people may need to try and test symlinks themselves to verify if Jib handles them in the way they want.
- It can be a bit nuanced what to do depending on whether a symlink refers to relative path or absolute path?

So, a configuration like the following at the top level is what have been thinking about:
```xml
<links>
  <link>
    <path>/path/to/link</path>
    <target>/path/to/target</target>
    <!-- and potentially type: soft or hard, but perhaps only soft? -->
  </link>
</links>
```
Unfortunately, I anticipate implementing this will require much more efforts, perhaps extending the public API of `jib-core`.

chanseokoh:
Just in case, this idea is not something I capriciously came up with just now: https://github.com/GoogleContainerTools/jib/issues/1576#issuecomment-503810403

jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
```diff
@@ -38,15 +38,42 @@
...
 
 /** Tests for {@link ReproducibleLayerBuilder}. */
 public class ReproducibleLayerBuilderTest {
 
+  @After
+  public void cleanUp() throws IOException, URISyntaxException {
+    removeLinks(getLinks());
+  }
+
+  private List<Path> getLinks() throws URISyntaxException {
```
FirstLast94397, line 58: ?

jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
```diff
@@ -67,6 +94,26 @@ private static void verifyNextTarArchiveEntry(
...
+      String expectedLinkName)
+      throws IOException {
+    TarArchiveEntry header = tarArchiveInputStream.getNextTarEntry();
+    assertThat(header.getName()).isEqualTo(expectedExtractionPath);
+    assertThat(header.getLinkName()).isEqualTo(expectedLinkName);
+    assertThat(header.isSymbolicLink()).isTrue();
+  }
+
   /**
    * Verifies that the next {@link TarArchiveEntry} in the {@link TarArchiveInputStream} is a
```
FirstLast94397, line 118: I want see**
```suggestion
   * @param tarArchiveInputStream the {@link TarArchiveInputStream} to read from
   * @param expectedExtractionPath the expected extraction path of the next entry
   * @param expectedLinkName the expected link name of the next entry
   * @throws IOException if an I/O exception occurs
   */
  private static void verifyNextTarArchiveEntryIsLink(
      TarArchiveInputStream tarArchiveInputStream,
      String expectedExtractionPath,
      String expectedLinkName)
      throws IOException {
    TarArchiveEntry header = tarArchiveInputStream.getNextTarEntry();
    assertThat(header.getName()).isEqualTo(expectedExtractionPath);
    assertThat(header.getLinkName()).isEqualTo(expectedLinkName);
    assertThat(header.isSymbolicLink()).isTrue();
  }

   /**
    * Verifies that the next {@link TarArchiveEntry} in the {@link - [ ] # TarArchiveInputStream} is a**__[]- - - - - ()__**
```
**

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -25,6 +25,7 @@
 import com.google.common.collect.ImmutableList;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
```
FirstLast94397, line 28:
```suggestion
import java.nio.file.Files;1

```

FirstLast94397 (approved):

StefanGoldmann:
Hello chanseokoh,

granted
- the approach as proposed is breaking compatibility
- an explicit API would be nice but imposes a lot more work

As a remedy, wouldn't it be easily possible to keep backwards compatibilty utilizing a flag in the layer builder, say 'retainSymlinks'?
Defaulting this to 'false' would leave everything as is, and if its true this particular layer builder could behave as proposed by seanleblanc.

What do you think?

Regards, Stefan

StefanGoldmann:
[chanseokoh](https://github.com/chanseokoh)

Hello again, I have extended seanleblanc approach and extended it such that the user needs to explicitly
enable retaining of symbolic links. That way, everything should stay backwards compatible.

I am totally new on github and don't really know how to proceed now. Any advice appreciated.

Anyway, if you like, give it a look. Tests are pending yet ... 
https://github.com/StefanGoldmann/jib-symlinks/commit/3218679f2d165447d6d7debb033e8830ca82458c

putsev:
@StefanGoldmann Could you create another Pull Request with your changes, please?

StefanGoldmann:
I'm on vacation, but will do in 2 weeks.

Regards, Stefan

Andrew Putsev ***@***.***> schrieb am Mi., 7. Feb. 2024,
11:11:

StefanGoldmann:
@putsev Sorry for the late response, pull request now created (https://github.com/GoogleContainerTools/jib/pull/4233)

PR #3512, opened 2021-11-24
Mention that JAVA_TOOL_OPTIONS only support a max of 1024 chars

nastra:
This is to save people from debugging issues with their application and only realizing later why not everything in `JAVA_TOOL_OPTIONS` is being picked up. 
See also http://hg.openjdk.java.net/jdk7/jdk7/hotspot/file/9b0ca45cd756/src/share/vm/runtime/arguments.cpp#l2772 for the size limitation in the JVM code itself.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -214,7 +214,9 @@ spec:
     - name: JAVA_TOOL_OPTIONS
       value: <JVM flags>
 ```
-Note that the `JAVA_TOOL_OPTIONS` environment varible only supports a max length of **1024** characters. Anything longer than this will be cut off by the JVM.
+Note that many JVMs may only support a max length of **1024** characters for the `JAVA_TOOL_OPTIONS` environment variable, and anything longer than this may be cut off by the JVM.
+
+For Java 9+, often you may want to use [`JDK_JAVA_OPTIONS`](https://stackoverflow.com/questions/52986487/what-is-the-difference-between-jdk-java-options-and-java-tool-options-when-using) instead of `JAVA_TOOL_OPTIONS`.
 
 #### Other Environment Variables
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -214,6 +214,7 @@ spec:
     - name: JAVA_TOOL_OPTIONS
       value: <JVM flags>
 ```
+Note that the `JAVA_TOOL_OPTIONS` environment varible only supports a max length of **1024** characters. Anything longer than this will be cut off by the JVM.
```
chanseokoh, line 217:
```suggestion
Note that many JVMs may only support a max length of **1024** characters for the `JAVA_TOOL_OPTIONS` environment variable, and anything longer than this may be cut off by the JVM.

For Java 9+, often you may want to use [`JDK_JAVA_OPTIONS`](https://stackoverflow.com/questions/52986487/what-is-the-difference-between-jdk-java-options-and-java-tool-options-when-using) instead of `JAVA_TOOL_OPTIONS`.
```

chanseokoh:
Thanks for updating the doc! (I was procrastinating to update this section. :sweat_smile:) The [CLA check](https://github.com/GoogleContainerTools/jib/runs/4314598132) is failing. Check the details there, and please sign the CLA so that we can accept your contributino!

chanseokoh:
Oh, never mind. Looks like you already signed the CLA. Thanks a lot! Enabling auto-merge.

docs/faq.md
nastra: thanks for your suggestion. I actually checked newer JDK versions and those don't seem to have that limitation. Still I think it's worth mentioning this limit

chanseokoh (approved):

PR #3522, opened 2021-12-06
Fix incorrect parsing of comma-separate Jib property values

### LINKED ISSUE #2224: Properties with commas in the value
**Environment**: 

- *Jib version:*  1.7.0
- *Build tool:*  Maven
- *OS:* Windows

**Description of the issue**:

Cannot specify a property with comma in the value. Looking at the code it looks very simple, it would be good if there was a way to be able to specify comma as a value by, somehow, escaping it.

**Expected behavior**:

<jvmFlags>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005</jvmFlags>

The above will split the jvmFlags on each comma. It would be nice if one can escape the commas and pass the whole value as is.

**Steps to reproduce**:

  1. Specify <jvmFlags>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005</jvmFlags>
  2. After build the entry point jvm flags will be split into ['-agentlib:jdwp=transport=dt_socket', 'server=y', 'suspend=n', 'address=*:5005']

chanseokoh:
Fixes #2224.

The only character we interpret in a special way is a comma (to separate tokens when parsing a list or a map). There should be a way to put a comma is to escape it, and that's the only character we need escaping for. However, the previous implementation didn't drop the slash used for escaping.

### DIFF
--- a/jib-gradle-plugin/CHANGELOG.md
+++ b/jib-gradle-plugin/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 - Fixed `ClassCastException` when using non-`String` value (for example, [`Provider`](https://docs.gradle.org/current/javadoc/org/gradle/api/provider/Provider.html)) for `Main-Class` manifest attribute of the `jar` task. ([#3396](https://github.com/GoogleContainerTools/jib/issues/3396))
+- Fixed incorrect parsing with comman escaping when providing Jib list or map property values on the command-line. ([#2224](https://github.com/GoogleContainerTools/jib/issues/2224))
 
 ## 3.1.4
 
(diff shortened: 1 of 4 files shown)


suztomo:
Which test case represents the problem in this comment https://github.com/GoogleContainerTools/jib/issues/2224#issuecomment-986211953 ?

chanseokoh:
[`testParseListProperty()`](https://github.com/GoogleContainerTools/jib/pull/3522/files#diff-713879f0e5536cc400b9203ffe4a71a56395d0018f65d300ad36bc729c564925R143) and [`testParseMapProperty()`](https://github.com/GoogleContainerTools/jib/pull/3522/files#diff-713879f0e5536cc400b9203ffe4a71a56395d0018f65d300ad36bc729c564925R154). So the test cases in those methods have been updated. Any changes to other methods are Truth migration.

> Escaping is tough to understand for me. It would be helpful if you can add a comment in the test case such as

I just added another test case of `-Xmx2g,-agentlib:jdwp=transport=dt_socket\,server=y\,address=*:5005`.

suztomo (approved):
Nice. Thank you.

PR #3505, opened 2021-11-14
Adding support for symbolic links.

seanleblanc:
Related: https://github.com/GoogleContainerTools/jib/issues/2275

PR #3526, opened 2021-12-11
Expose jib.from.platforms as property.

### LINKED ISSUE #2742: <platform> configuration should be configurable with system properties
For example, `-Djib.from.platforms=... list of platforms ...`.

Note that `architecture` and `os` are required sub-properties. Need to think about the format. Also consider #2471.

ls-urs-keller:
Feature request: fixes #2742

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -56,14 +56,14 @@ public class BaseImageParameters {
   @Nested
   @Optional
   public ListProperty<PlatformParameters> getPlatforms() {
-    String platformsProperty = System.getProperty(PropertyNames.FROM_PLATFORMS);
-    if (platformsProperty != null) {
-      List<PlatformParameters> parsedPlatforms =
-          ConfigurationPropertyValidator.parseListProperty(platformsProperty).stream()
-              .map(PlatformParameters::ofString)
+    String property = System.getProperty(PropertyNames.FROM_PLATFORMS);
+    if (property != null) {
+      List<PlatformParameters> parsed =
+          ConfigurationPropertyValidator.parseListProperty(property).stream()
+              .map(PlatformParameters::of)
               .collect(Collectors.toList());
-      if (!parsedPlatforms.equals(platforms.get())) {
-        platforms.set(parsedPlatforms);
+      if (!parsed.equals(platforms.get())) {
+        platforms.set(parsed);
       }
     }
     return platforms;
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
@@ -16,7 +16,10 @@
 
 package com.google.cloud.tools.jib.gradle;
 
+import com.google.cloud.tools.jib.plugins.common.ConfigurationPropertyValidator;
 import com.google.cloud.tools.jib.plugins.common.PropertyNames;
+import java.util.List;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 import org.gradle.api.Action;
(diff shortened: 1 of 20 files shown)


zhumin8:
Thanks for opening the PR !
The CLA check is failing. Please check the details [here](https://github.com/GoogleContainerTools/jib/pull/3505/checks?check_run_id=4205485364), and sign the CLA so that we can review and accept your contribution!

ls-urs-keller:
> Thanks for opening the PR ! The CLA check is failing. Please check the details
My mistake https://cla.developers.google.com/clas used my default Google account, I didn't pay attention to it, should be OK now.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -146,6 +147,16 @@ public PermissionConfiguration() {}
...
     }
+
+    private static PlatformParameters ofString(String osArchitecture) {
+      if (osArchitecture == null || !osArchitecture.matches("[^/]+/[^/]+")) {
+        throw new IllegalArgumentException("osArchitecture must be of form os/architecture");
+      }
+      PlatformParameters platformParameters = new PlatformParameters();
+      platformParameters.os = osArchitecture.split("/")[0].trim();
+      platformParameters.architecture = osArchitecture.split("/")[1].trim();
+      return platformParameters;
```
chanseokoh, line 158: We use NullAway. The null check is unnecessary. And I don't like allowing whitespace and using `.trim()`. Let's just say `[^/ ]+/[^/ ]+`. And then, I think we can make this simpler:

```java
      Matcher matcher = Pattern.compile("([^/ ]+)/([^/ ]+)").matcher(osArchitecture);
      if (!matcher.matches()) {
        throw ...
      }
      PlatformParameters platformParameters = new PlatformParameters();
      platformParameters.os = matcher.group(1);
      platformParameters.architecture = matcher.group(2);
      return platformParameters;
```

And let's move this method to the top of the class (sorry, our repo style):
```java

public static class PlatformParameters implements PlatformConfiguration {

  private static PlatformParameters ofString(String osArchitecture) { ...
```

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -354,6 +362,12 @@ protected void checkJibVersion() throws MojoExecutionException {
    * @return the specified platforms
    */
   List<PlatformParameters> getPlatforms() {
+    String property = getProperty(PropertyNames.FROM_PLATFORMS);
+    if (property != null) {
+      return Stream.of(property.split(","))
```
chanseokoh, line 367: For consistency, let's use `ConfigurationPropertyValidator.parseListProperty(property).map(...)...`.

chanseokoh (commented):
Could you add some tests in `JibPluginConfigurationTest`? Pretty trivial to do so.

And, any chance you're interested in enabling this for Gradle? Not that I'm demanding, because it will do require your time.

ls-urs-keller:
> Awesome! Thanks for your contribution.
> 
> Could you add some tests in `JibPluginConfigurationTest`? Pretty trivial to do so if you look at the existing tests.
> 
> And, any chance you're interested in enabling this for Gradle? Not that I'm demanding, because it will do require your time.

I added it although I only tested it through unit tests.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -53,6 +56,16 @@ public BaseImageParameters(ObjectFactory objectFactory) {
   @Nested
   @Optional
   public ListProperty<PlatformParameters> getPlatforms() {
+    String property = System.getProperty(PropertyNames.FROM_PLATFORMS);
+    if (property != null) {
+      List<PlatformParameters> parsed =
+          ConfigurationPropertyValidator.parseListProperty(property).stream()
+              .map(PlatformParameters::of)
+              .collect(Collectors.toList());
+      if (!parsed.equals(platforms.get())) {
```
chanseokoh, line 65: `.equals()` wasn't working because `PlatformParameters` didn't define `.equals()` and `.hashCode()`. I fixed it.

chanseokoh (approved):
Thanks for your time! I've fixed an issue, updated docs, and refactored test code. Will merge this after checking a few things.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
ls-urs-keller: Thank you, missed it.

Issue #3542, opened 2021-12-31
extraDirectories property should validate if the supplied path exists


tommysitu:
**Environment**:

- *Jib version:* 3.1.4
- *Build tool:* Maven 3.6.2

**Description of the issue**:
```
<configuration>
    <extraDirectories>
         <paths>
             <path>missing/folder</path>
          </paths>
     </extraDirectories>
</configuration>
```
Using the `extraDirectories` config above to copy additional files doesn't fail the build even the path is invalid or not a folder. This behaviour is not good for catching bugs early if the additional files are crucial for the docker container. 

**Expected behaviour**:
Should be similar to the `docker build` command which would throw the following exception if a `Dockerfile` has a command to copy a non-existent folder: 

```
 => ERROR [stage-1 2/2] COPY missing/folder /app                                                                                                                                     0.0s
------
 > [stage-1 2/2] COPY missing/folder /app:
------
failed to compute cache key: "/missing/folder" not found: not found

```

meltsufin:
Thank you for submitting this enhancement request.

chanseokoh:
The reason we don't check the existence is due to the default implicit directory `src/main/jib`, which doesn't exist often. Perhaps we should fail only for non-default directories.

tommysitu:
Hi @chanseokoh I've raised a PR. Could you approve the build? so that we can see which tests are failing. Thanks!

chanseokoh:
@tommysitu I just approved it.

tommysitu:
Thanks for reviewing the PR @chanseokoh. Shall I close this issue?

cebrailinanc:
i got  error on gradle
Execution failed for task ':jib'.
> extraDirectories.paths contain "from" directory that doesn't exist locally: /aass/test

my command:  `./gradlew jib ... -Djib.extraDirectories.paths=truststore,/aass/test`

chanseokoh:
@cebrailinanc just create the directory before running Gradle, like `mkdir /aass/test`.

cebrailinanc:
I tried but it didn't work. Because we need to open a folder within the image.

chanseokoh:
Oh, I see what you were thinking. It must be that you thought the syntax of the command-line parameter would be
```
-Djib.extraDirectories.paths=<from_1>,<to_1>,<from_2>,<to_2>,...
```
[...]

PR #3550, opened 2022-01-13
Throw exception if extraDirectories config contains invalid path

tommysitu:
Fix this issue: https://github.com/GoogleContainerTools/jib/issues/3542

A couple of tests that depend on absolute path may fail, I will look into that later


chanseokoh:
I think we can use `FileNotFoundException` instead of creating an new exception.

Also, this shouldn't fail for the default Jib directory `src/main/jib` that often doesn't exist.

tommysitu:
@chanseokoh I thought the reason we want to have a specific exception is for setting error message with details about the error config value which could vary between maven or gradle build, like this one: https://github.com/GoogleContainerTools/jib/pull/3550/files#diff-580c1f111d812c84cb7bfd7d1bf4adaf8e07fce0b02daf944a36e459b25158e3R169-R171

chanseokoh:
Oh, now you said it, I agree being able to print a specific error message is very helpful in this case.

Let's change the exception name though, from `InvalidExtraDirectoryException` to `ExtraDirectoryNotFoundException`. We use "invalid" to mean something else. (That is, this is when a directory doesn't exist, while the Jib extra directory configuration itself is valid.)

tommysitu:
cool. I will update that.

tommysitu:
> Also, this shouldn't fail for the default Jib directory src/main/jib that often doesn't exist.

so if a user has files in this folder, jib will automatically add them to the docker image, without the user to specify `extraDirectories` config. In this case, yeah it makes sense for us to skip it, because we want o validated what the customized `extraDirectories` config value only.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -180,6 +181,9 @@ public void buildDocker()
       throw new GradleException(
           HelpfulSuggestions.forInvalidImageReference(ex.getInvalidReference()), ex);
 
+    } catch (ExtraDirectoryNotFoundException ex) {
+      throw new GradleException(
+          "extraDirectories.paths contain path that is not found: " + ex.getPathValue(), ex);
```
chanseokoh, line 186: How about this?

```suggestion
          "extraDirectories.paths contain \"from\" directory that doesn't exist locally: " + ex.getPathValue(), ex);
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ExtraDirectoryNotFoundException.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2018 Google LLC.
```
chanseokoh, line 2: 2022

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessorTest.java
```diff
@@ -259,13 +260,23 @@ public void testPluginConfigurationProcessor_extraDirectory()
     assertThat(fooEntry.get().getPermissions().toOctalString()).isEqualTo("123");
   }
 
+  @Test
+  public void testPluginConfigurationProcessor__errorOnExtraDirectoryPathNotFound() throws URISyntaxException, NumberFormatException {
+    Path extraDirectory = Paths.get(Resources.getResource("core/layer").toURI()).resolve("xyz");
+    Mockito.<List<?>>when(rawConfiguration.getExtraDirectories())
+        .thenReturn(Arrays.asList(new TestExtraDirectoriesConfiguration(extraDirectory)));
+
+    ExtraDirectoryNotFoundException exception = assertThrows(ExtraDirectoryNotFoundException.class, this::processCommonConfiguration);
```
chanseokoh, line 269: nit:
```suggestion
    Exception exception = assertThrows(ExtraDirectoryNotFoundException.class, this::processCommonConfiguration);
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/ExtraDirectoryNotFoundException.java
```diff
@@ -0,0 +1,34 @@
...
+public class ExtraDirectoryNotFoundException extends Exception {
+
+  private final String path;
+
+  public ExtraDirectoryNotFoundException(String message, String path) {
+    super(message);
+    this.path = path;
+  }
+
+  public String getPathValue() {
```
chanseokoh, line 31:
```suggestion
  public String getPath() {
```

chanseokoh (commented):
LGTM

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
tommysitu: yeah this is much better

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -92,6 +92,7 @@
 
   private static final String JIB_CLASSPATH_FILE = "jib-classpath-file";
   private static final String JIB_MAIN_CLASS_FILE = "jib-main-class-file";
+  private static final String DEFAULT_JIB_DIR = "src/main/jib";
```
chanseokoh, line 95: Not that this isn't going to work on Windows, but we just prefer this form:

```suggestion
  private static final Path DEFAULT_JIB_DIR = Paths.get("src").resolve("main").resolve("jib");
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -446,6 +447,9 @@ static JibContainerBuilder processCommonConfiguration(
                 extraDirectory.getExcludesList(),
                 rawConfiguration.getExtraDirectoryPermissions(),
                 modificationTimeProvider));
+      } else if (!from.endsWith(DEFAULT_JIB_DIR)){
+        // Throw exception if user-provided extraDirectory path is not found (excluding default jib directory)
```
chanseokoh, line 451: nit: unnecessary
```suggestion
```

chanseokoh (commented):
Could you also update `jib-maven-plugin/CHANGELOG` and `jib-gradle-plugin/CHANGELOG`? Thanks!

chanseokoh (approved):
Thanks for your contribution. Will merge it soon.

tommysitu:
unit test still failing because of some dummy folder missing in the test projects, do you want me to fix that?

chanseokoh:
Hmm... yeah, please try to fix them. You can run unit tests locally with `./gradle check`.

jib-maven-plugin/src/test/resources/maven/projects/multi/complex-service/pom.xml
```diff
@@ -22,7 +22,6 @@
             <paths>
               <path>src/main/jib1</path> <!-- should resolve to this projectRoot -->
               <path>${project.basedir}/src/main/jib2</path> <!-- provided as absolute within project -->
-              <path>/some/random/absolute/path/jib3</path> <!-- provided as absolute outside of project -->
```
tommysitu, line 25: can't use absolute path in the test anymore, because it won't exist in the build server

tommysitu:
yeah will do. how about the sonaCloud code analysis failures? Shall I fix that too?

chanseokoh:
Let's ignore SonarCloud, but builds are still failing due to formatting. Sorry, I should have mentioned this earlier, but `./gradlew dev` checks all the things (`check`, `javadoc`, and `googleJavaFormat`), so run that locally to make sure there's no error. It may auto-format your code, so you may need to commit the changes too.

tommysitu:
my bad, should have run all checks locally first. underestimated how rigorous these checks are 😅

tommysitu:
All passing locally with ` ./gradlew clean build jacocoTestReport`. Could you try one last time please? @chanseokoh 🤞

chanseokoh:
Thanks for your contribution.

PR #3575, opened 2022-02-10
Add support for configuration of credential helper with environment variables

### LINKED ISSUE #2814: Set environment variables for credHelper

tobiade:
Fixes #2814


jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
```diff
@@ -131,7 +153,9 @@ private Credential retrieve(List<String> credentialHelperCommand)
       throws IOException, CredentialHelperUnhandledServerUrlException,
           CredentialHelperNotFoundException {
     try {
-      Process process = processBuilderFactory.apply(credentialHelperCommand).start();
+      ProcessBuilder processBuilder = processBuilderFactory.apply(credentialHelperCommand);
+      processBuilder.environment().putAll(environment);
```
tobiade, line 157: One thing I'm curious about here is if a user sets an environment variable for the cred helper from the command line and also sets the same environment variable in their build file. The environment variable set in the build file will take precedence - is that okay? My initial thought is no...

chanseokoh: > environment variable for the cred helper from the command line

You're talking about the environment variables seen from the JVM process (i.e., what you get with `processBuilder.environment()`, which is same as `System.getenv()`), right? I think it is right that the variables configured through Jib parameters overwrite the environment inherited from the JVM.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelperTest.java
```diff
@@ -95,6 +100,32 @@ public void testRetrieve()
...
+            systemProperties,
+            processBuilderFactory,
+            credHelperEnvironment);
+    Credential credential = credentialHelper.retrieve();
+    Assert.assertEquals("myusername", credential.getUsername());
+    Assert.assertEquals("mysecret", credential.getPassword());
+
+    Mockito.verify(processBuilderFactory).apply(command);
+    Mockito.verify(processBuilderEnvironment).putAll(credHelperEnvironment);
+    Assert.assertEquals(1, processBuilderEnvironment.size());
```
chanseokoh, line 126: Perhaps we can check also if the environment has `ENV_VARIABLE` and its value?

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-cred-helper.xml
```diff
@@ -0,0 +1,55 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>
+                <configuration>
+                    <release>11</release>
```
chanseokoh, line 33: We don't run integration tests on GitHub Actions but on a different CI. And we use Java 8 there.

```
<source>1.8</source>
<target>1.8</target>
```

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-cred-helper.xml
```diff
@@ -0,0 +1,55 @@
...
+
+            <plugin>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>jib-maven-plugin</artifactId>
+                <version>${jib-maven-plugin.version}</version>
+                <configuration>
+                    <to>
+                        <image>${_TARGET_IMAGE}</image>
+                        <credHelper>
+                            <helper>gcr</helper>
```
chanseokoh, line 45: Looks like we don't have an integration test using `<credHelper>gcr</credHelper>`, which should continue to work. Could you add that too? Thanks!

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
```diff
@@ -57,6 +57,12 @@
     List<String> getExcludesList();
   }
 
+  interface CredHelperConfiguration {
```
chanseokoh, line 60:
```suggestion
  static interface CredHelperConfiguration {
```

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
```diff
@@ -65,9 +71,9 @@
 
   AuthProperty getToAuth();
 
-  Optional<String> getFromCredHelper();
+  CredHelperConfiguration getFromCredHelperConfig();
```
chanseokoh, line 74: Nit: `getFromCredHelper` is good enough.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
```diff
@@ -57,6 +57,12 @@
     List<String> getExcludesList();
   }
 
+  interface CredHelperConfiguration {
+    Optional<String> getHelperName();
```
chanseokoh, line 61: I think `getHelper()` is good enough, if that doesn't cause a conflict with something.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -180,9 +226,14 @@ public FromConfiguration() {
 
     @Nullable @Parameter private String image;
     @Parameter private List<String> tags = Collections.emptyList();
-    @Nullable @Parameter private String credHelper;
+    @Parameter private CredHelperParameters credHelper;
```
chanseokoh, line 229: I think we can just do
```suggestion
    @Parameter private CredHelperParameters credHelper = new CredHelperParameters();
```
and remove the constructor you added.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -180,9 +226,14 @@ public FromConfiguration() {
...
     @Parameter private ToAuthConfiguration auth = new ToAuthConfiguration();
 
+    /** Constructor for defaults. */
+    public ToConfiguration() {
```
chanseokoh, line 233: Can be removed by the above.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -161,17 +163,61 @@ private static PlatformParameters of(String osArchitecture) {
     }
   }
 
+  /** Configuration for {@code [from|to].credHelper} parameter. */
+  public static class CredHelperParameters implements CredHelperConfiguration {
+    @Nullable @Parameter private String helper;
+    @Parameter private Map<String, String> environment;
```
chanseokoh, line 169: ditto
```suggestion
    @Parameter private Map<String, String> environment = new HashMap<>();
```
and remove the constructor.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
```diff
@@ -25,26 +25,23 @@
...
 import org.gradle.api.provider.ListProperty;
-import org.gradle.api.provider.Property;
-import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.Nested;
 import org.gradle.api.tasks.Optional;
 
 /** Object in {@link JibExtension} that configures the base image. */
-public class BaseImageParameters {
+public class BaseImageParameters extends ImageParameters {
```
chanseokoh, line 33: We are not really a fan of inheritance for this kind of usage. I'd usually add `ImageParameters` as a field. But then, the parameter structure doesn't match the actual Jib configuration. (That is, `AuthParameters` should be a field as we have `image.auth`.) I think it's still fine to add the cred helper parameters as a field. What do you think?

chanseokoh:
> I've triggered test runs and our team will review.

@zhumin8 just FYI, for external contributions, we need to add the `kokoro:run` label to trigger Kokoro.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
```diff
@@ -73,19 +76,38 @@
...
+  }
+
+  /**
+   * Constructs a new {@link DockerCredentialHelper}.
+   *
+   * @param serverUrl the server URL to pass into the credential helper
+   * @param credentialHelper the path to the credential helper executable
+   * @param environment environment variables used in configuring the credential helper
+   */
+  public DockerCredentialHelper(
```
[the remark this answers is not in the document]
tobiade: Not strictly - I added it to avoid refactoring code where the previous constructor was used, specifically the places where we didn't have access to any environment variables that could be used to configure the cred helper.

Doesn't seem to be a lot of places where the previous constructor was used though - can refactor as needed.

jib-core/src/test/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelperTest.java
tobiade: Ah, yes

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
tobiade: Agreed - however, it'll conflict with the `getHelper()` method required by Gradle. I took a similar approach to the one in `PlatformParameters`. I'm not  a Gradle expert but I think it might struggle to properly deal with `Optional<String>` inputs given it's not `Serializable` ( I think)

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
tobiade: Hmm I think I get what you mean. I was trying to reduce the code duplication between `BaseImageParameters` and `TargetImageParameters`. I guess making `ImageParameters` a field would mean Jib configuration might have to change? e.g to configure auth we might need to do something like `from.image.auth` as opposed to `from.auth`?

If that's the case, I'm happy to revert and keep as it was before while making `CredHelperParameters` fields in both
`BaseImageParameters` and `TargetImageParameters`.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
tobiade: Got it

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BaseImageParameters.java
chanseokoh: > I guess making `ImageParameters` a field would mean Jib configuration might have to change? e.g to configure auth we might need to do something like `from.image.auth` as opposed to `from.auth`?

I think technically it is not impossible to have `ImageParameters` as a field while retaining the current Jib configuration structure. However, I don't really like doing that. I'd rather have the code structure in sync with Jib configuration.

So yeah, let's keep `CredHelperParameters` as a field.

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/RawConfiguration.java
tobiade: I believe the `CredHelperConfiguration` interface is implicitly static - why does it need to be explicitly declared as static?

chanseokoh: Ah, you're totally right. If you want, you're welcome to delete unnecessary `static` in other nested interfaces.

zhumin8:
@chanseokoh Thanks for the reminder.

tobiade:
Thanks @chanseokoh. I've addressed comments in latest commit.

chanseokoh (commented):
Thanks! I'll take another look later.

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
```diff
@@ -161,11 +163,50 @@ private static PlatformParameters of(String osArchitecture) {
...
+    public Optional<String> getHelperName() {
+      return Optional.ofNullable(helper);
+    }
+
+    @Override
+    public Map<String, String> getEnvironment() {
+      return environment;
+    }
+
+    public void setHelper(@Nullable String helper) {
```
chanseokoh, line 181: The `helper` argument is nullable here but it's not in `set()`. Should these be consistent?

chanseokoh (approved):
Overall, looks go to me. Just a few minor comments.

Also, could you update `jib-maven-plugin/CHANGELOG.md` and `jib-gradle-plugin/CHANGELOG.md`? (BTW, don't update other docs about the new configuration parameters in this PR, since we update them after making a release.)

chanseokoh:
And you can ignore the SonarCloud failure.

jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerCredentialHelper.java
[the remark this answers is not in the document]
tobiade: Ah, I see. Yes the `DockerCredentialHelper(String serverUrl, Path credentialHelper, Map<String, String> environment)` constructor was being used in non-test classes. It's the constructor being used in `CredentialRetrieverFactory forImage(ImageReference imageReference, Consumer<LogEvent> logger, Map<String, String> environment)`. Np - I'll add it back in

jib-maven-plugin/src/main/java/com/google/cloud/tools/jib/maven/JibPluginConfiguration.java
tobiade: Will change - they should be

jib-gradle-plugin/CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ All notable changes to this project will be documented in this file.
 
 - [`jib.from.platforms`](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#from-closure) parameter for multi-architecture image building can now be configured through Maven and system properties (for example, `-Djib.from.platforms=linux/amd64,linux/arm64` on the command-line). ([#2742](https://github.com/GoogleContainerTools/jib/pull/2742))
 - For retrieving credentials, Jib additionally looks for `$XDG_RUNTIME_DIR/containers/auth.json`, `$XDG_CONFIG_HOME/containers/auth.json`, and `$HOME/.config/containers/auth.json`. ([#3524](https://github.com/GoogleContainerTools/jib/issues/3524))
+- Environment variables can now be used in configuring credential helpers. ([#2814](https://github.com/GoogleContainerTools/jib/issues/2814))
+  ```gradle
+   jib.to {
+      image = 'myimage'
+      credHelper {
+          helper = 'ecr-login'
```
tobiade, line 15: I'm not too familiar with GCP so I've used an AWS example here. Please let me know if a GCP example is preferable.

tobiade:
Thanks @chanseokoh. Updated as discussed.

chanseokoh (approved):
Awesome! Thanks for your contribution.

PR #3593, opened 2022-03-04
fix format config on buildSrc kts srcipt

BiLuoHen:
in buildSrc srcipt: Cannot access class 'com.google.cloud.tools.jib.api.buildplan.ImageFormat'

### DIFF
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
@@ -222,6 +222,7 @@ public class JibExtensionTest {
         container -> {
           container.setFormat("OCI");
         });
+    ContainerParameters container = testJibExtension.getContainer();
     assertThat(container.getFormat()).isSameInstanceAs(ImageFormat.OCI);
   }
 
--- a/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
+++ b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
@@ -212,6 +212,8 @@ public class JibExtensionTest {
     assertThat(container.getArgs()).containsExactly("arg1", "arg2", "arg3").inOrder();
     assertThat(container.getPorts()).containsExactly("1000", "2000-2010", "3000").inOrder();
     assertThat(container.getFormat()).isSameInstanceAs(ImageFormat.OCI);
+    container.setFormat("Docker");
+    assertThat(container.getFormat()).isSameInstanceAs(ImageFormat.Docker);
     assertThat(container.getAppRoot()).isEqualTo("some invalid appRoot value");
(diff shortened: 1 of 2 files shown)


jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
```diff
@@ -212,6 +212,8 @@ public void testContainer() {
     assertThat(container.getArgs()).containsExactly("arg1", "arg2", "arg3").inOrder();
     assertThat(container.getPorts()).containsExactly("1000", "2000-2010", "3000").inOrder();
     assertThat(container.getFormat()).isSameInstanceAs(ImageFormat.OCI);
+    container.setFormat("Docker");
```
mpeddada1, line 215: Let's move this line up to line 202. We want all the setup code to be in one place.

mpeddada1 (commented):
Thank you for your contribution! Had a minor comment.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
BiLuoHen: if we move the line to 202, the set will override the original `setFormat(ImageFormat.OCI)`, it will make the line 214 wrong, and not cover the code `setFormat(ImageFormat format)`, so I write the `setFormat("Docker")` here.

BiLuoHen: and a little confuse: why kokro-macos always failed, first time I thought it is no test code, but after added test code , it still failed. Thanks

mpeddada1: You're right! Sorry about that. In that case, it would be much better to have this new test be in a test method of its own. It is better to have smaller and more focused tests. We can name it `testSetFormat()`.

mpeddada1: Unfortunately we're experiencing some failures in our macos job which we're currently trying to address. I don't think it has anything to do with your PR.

BiLuoHen: > You're right! Sorry about that. In that case, it would be much better to have this new test be in a test method of its own. It is better to have smaller and more focused tests. We can name it `testSetFormat()`.

Got it, I will done it soon.

mpeddada1 (approved):
Thank you @BiLuoHen! LGTM.

BiLuoHen:
> Thank you @BiLuoHen! LGTM.

sorry for requested a review again, just dismiss that, i click a wrong button

elefeint:
We will merge this in after CI is fixed.

elefeint:
[jib-gradle-plugin v3.2.1](https://github.com/GoogleContainerTools/jib/releases/tag/v3.2.1-gradle) released.

PR #3598, opened 2022-03-07
Set up docker machine with VirtualBox for migration to big-sur

mpeddada1:

### DIFF
--- a/kokoro/docker_setup.sh
+++ b/kokoro/docker_setup.sh
@@ -8,9 +8,6 @@ echo $DOCKER_IP
 docker-machine ssh default "echo '{ \"insecure-registries\":[\"$DOCKER_IP:5000\", \"$DOCKER_IP:6000\", \"$DOCKER_IP:8080\"] }' | sudo tee /etc/docker/daemon.json "
 docker-machine ssh default "echo 'DOCKER_OPTS=\"--config-file=/etc/docker/daemon.json\"' | sudo tee -a /var/lib/boot2docker/profile "
 docker-machine ssh default "mkdir /home/docker/auth; docker run --entrypoint htpasswd httpd:2 -Bbn testuser testpassword > /home/docker/auth/htpasswd"
-docker-machine ssh default "cat /home/docker/auth/htpasswd"
-docker-machine ssh default "pwd"
-docker-machine ssh default "echo $HOME"
 docker-machine ssh default "sudo /etc/init.d/docker restart"
 docker-machine inspect default
 docker-machine env default


elefeint:
Still fails on macos, right? Can't find the docker registry?

elefeint (approved):
Makes sense to try. There might be some firewall setting somewhere...

chanseokoh:
It may not be a transient error. See https://docs.docker.com/config/daemon/ipv6/ and b/203732253. But I am just guessing here.

mpeddada1:
Aha thanks for the context! Asked about it internally and this seems like a likely cause. Will keep this PR open for now and wait for updates.

mpeddada1:
We still seem to be running into the following error:
```
com.google.cloud.tools.jib.api.ContainerizerIntegrationTest > testSteps_forBuildToDockerRegistry_skipExistingDigest FAILED
    java.util.concurrent.ExecutionException: org.apache.http.conn.HttpHostConnectException: Connect to localhost:5000 [localhost/0:0:0:0:0:0:0:1, localhost/127.0.0.1] failed: Connection refused (Connection refused)

        Caused by:
        org.apache.http.conn.HttpHostConnectException: Connect to localhost:5000 [localhost/0:0:0:0:0:0:0:1, localhost/127.0.0.1] failed: Connection refused (Connection refused)

            Caused by:
            java.net.ConnectException: Connection refused (Connection refused)
```

mpeddada1:
Replacing `localhost` with direct use of ip results in the following:

```
com.google.cloud.tools.jib.api.ContainerizerIntegrationTest > testSteps_forBuildToDockerRegistry FAILED
    java.lang.RuntimeException: Command 'docker pull x.x.x.x:5000/testimage:testtag' failed: Error response from daemon: Get https://x.x.x.x:5000/v2/: http: server gave HTTP response to HTTPS client
        at com.google.cloud.tools.jib.Command.run(Command.java:79)
        at com.google.cloud.tools.jib.Command.run(Command.java:52)
        at com.google.cloud.tools.jib.registry.LocalRegistry.pull(LocalRegistry.java:119)
        at com.google.cloud.tools.jib.api.ContainerizerIntegrationTest.testSteps_forBuildToDockerRegistry(ContainerizerIntegrationTest.java:189)

```

mpeddada1:
Using the ip directly by doing `export DOCKER_IP="$(docker-machine ip default)"` seemed to have fixed the failure in ManifestPusherIntegrationTest. The error message from [above](https://github.com/GoogleContainerTools/jib/pull/3598#issuecomment-1144177276)  still persists for other tests like ContainerizerIntegrationTest, JibIntegrationTest, etc.

kokoro/presubmit.sh
```diff
@@ -5,21 +5,27 @@ set -o xtrace
 
 gcloud components install docker-credential-gcr
 
+# Docker service does not run by default in Big Sur but can be started with the following commands.
+if [ "${KOKORO_JOB_CLUSTER}" = "MACOS_EXTERNAL" ]; then
+docker-machine ls
+docker-machine start default
+docker-machine ssh default "echo '{ \"insecure-registries\":[\"192.168.99.104:5000\"] }' | sudo tee /etc/docker/daemon.json "
```
chanseokoh, line 12: In the end, this should probably be
```suggestion
docker-machine ssh default "echo '{ \"insecure-registries\":[\"$DOCKER_IP:5000\"] }' | sudo tee /etc/docker/daemon.json "
```
and come after the `export DOCKER_IP` below.

Also, 5000 is the usual port for a local registry, but I remember we also use 6000. I wonder if we can omit the port part or have a way to specify any ports.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -91,7 +93,7 @@ public void start() throws IOException, InterruptedException {
               "-e",
               "REGISTRY_AUTH_HTPASSWD_PATH=/auth/htpasswd"));
     }
-    dockerTokens.add("registry:2");
+    dockerTokens.add("registry.ipv6.docker.com/library/registry:2");
```
chanseokoh, line 96: I guess this change shouldn't be necessary?

mpeddada1: hmm you might be right! Let me try removing this and see what happens.

kokoro/presubmit.sh
mpeddada1: Good point! Referencing the env variable is definitely better in the long run. I think eventually this fix will be applied in the image itself.

mpeddada1:
Tests passed on presubmit with https://github.com/GoogleContainerTools/jib/pull/3598/commits/aa76231e38dcc9762148b548e7522bae25c39437 but failed on continuous build.

mpeddada1:
The Jib CLI tests seem to be failing with the following error on the continuous build:

```
com.google.cloud.tools.jib.cli.JarCommandTest > testStandardJar_explodedMode_toDocker STANDARD_OUT
    [WARN] Base image 'eclipse-temurin:8-jre' does not use a specific image digest - build may not be reproducible
    Getting manifest for base image eclipse-temurin:8-jre...
    Building resources layer...
    Building classes layer...
    The base image requires auth. Trying again for eclipse-temurin:8-jre...
    Using base image with digest: sha256:347d785ee9160e197db44b3b79b4d87e728854b68189c3ca5930c911d96c3f54

    Container entrypoint set to [java, -cp, /app/explodedJar:/app/dependencies/*, HelloWorld]
    Container program arguments set to []
    Loading to Docker daemon...

com.google.cloud.tools.jib.cli.JarCommandTest > testStandardJar_explodedMode_toDocker FAILED
    java.lang.RuntimeException: Command 'docker run --rm 192.168.99.104:5000/exploded-jar --network=host' failed: 
        at com.google.cloud.tools.jib.Command.run(Command.java:79)
        at com.google.cloud.tools.jib.Command.run(Command.java:52)
        at com.google.cloud.tools.jib.cli.JarCommandTest.testStandardJar_explodedMode_toDocker(JarCommandTest.java:99)
```

mpeddada1:
Another update:
`testJar_baseImageSpecified` where we use `gcr.io/google-appengine/openjdk:8' passes but the other Jib CLI tests which use `eclipse-temurin:8-jre` by default are failing when `docker run` is executed

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
```diff
@@ -87,16 +89,29 @@ public void testStandardJar_explodedMode_toDocker()
...
-    String output = new Command("docker", "run", "--rm", "exploded-jar").run();
-    try (JarFile jarFile = new JarFile(jarPath.toFile())) {
-      String classPath =
-          jarFile.getManifest().getMainAttributes().getValue(Attributes.Name.CLASS_PATH);
-
-      assertThat(classPath).isEqualTo("dependency1.jar directory/dependency2.jar");
+            .execute(
+                "jar",
+                "--target",
+                "docker://" + dockerHost + ":5000/exploded-jar",
```
chanseokoh, line 95: Hmm... I have a doubt. Since it's directly pushing to a Docker daemon, I think the name shouldn't matter. Also, `docker run` isn't supposed to make a network connection, since the image should already exist in the daemon cache. But I'm not sure what's going on.

chanseokoh: Hmm... on second thought, it doesn't make sense, because the image including the base image layers is already built and pushed to a Docker daemon by Jib.

kokoro/continuous.sh
```diff
@@ -16,17 +34,6 @@ docker kill $(docker ps --all --quiet) || true
...
-if [ "${KOKORO_JOB_CLUSTER}" = "MACOS_EXTERNAL" ]; then
-  osascript -e 'quit app "Docker"'
-  open -a Docker
-  while ! docker info > /dev/null 2>&1; do sleep 1; done
-fi
-
 cd github/jib
 
-./gradlew clean build integrationTest --info --stacktrace
+./gradlew clean build :jib-cli:integrationTest --info --stacktrace
```
elefeint, line 39: We probably don't mean to just run `jib-cli` here?
AFAIK presubmit is `jib-core` only, but continuous should be everything.

kokoro/presubmit.sh
```diff
@@ -5,21 +5,28 @@ set -o xtrace
...
+docker-machine ls
+docker-machine start default
+export DOCKER_IP="$(docker-machine ip default)"
+echo $DOCKER_IP
+docker-machine ssh default "echo '{ \"insecure-registries\":[\"$DOCKER_IP:5000\", \"$DOCKER_IP:8080\"] }' | sudo tee /etc/docker/daemon.json "
+docker-machine ssh default "echo 'DOCKER_OPTS=\"--config-file=/etc/docker/daemon.json\"' | sudo tee -a /var/lib/boot2docker/profile "
+docker-machine ssh default "sudo /etc/init.d/docker restart"
+docker-machine env default
+eval "$(docker-machine env default)"
+
```
elefeint, line 19: Optional: might want to extract all this into a helper shell script. Or let's do it whenever we have to change anything about it, since it shows up twice.

elefeint (changes requested):
Looks good, let's just continue testing plugins on continuous.

elefeint (changes requested):
Ah. You'll probably have to do `--privileged` and `-network=host` for the plugin tests, too.

mpeddada1:
Ah it also looks like the temurin changes in the gradle tests were missing. Will try running the continuous build again.

mpeddada1:
It looks like `docker login` is still resulting in: `Command 'docker login 192.168.99.104:5000 -u testuser --password-stdin' failed: Error response from daemon: login attempt to http://192.168.99.104:5000/v2/ failed with status: 401 Unauthorized`

mpeddada1:
> Not sure if this is relevant, but `LocalRegistry` programmatically [creates an `htpasswd`](https://github.com/GoogleContainerTools/jib/blob/78f069b6dcb2ce1edbb43dc909006d7bf905fe85/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java#L76-L93) file. That file is for setting up an auth of a local registry running inside the `registry:2` container though.

Yes, this is definitely relevant! The kokoro team suggested that one of the common problems with `docker login` failing is with the htpasswd not being accessible. However, the fact that this file is programmatically created should make it accessible to the VM (unless I'm thinking about this incorrectly). I'm doing a quick experiment with manually creating one through the `docker-machine ssh` command and referencing it in the code to make sure we have our assumptions covered. @chanseokoh do you have any additional insights into why `docker login` doesn't work in some cases?

chanseokoh:
So one way to verify this quickly would be to copy the programmatically generated `htpasswd` file into the Linux VM yourself. For example,
```
docker-machine scp -r tempFolder docker@default:/tmp/auth
```
this will ssh into the host `default` with the user `docker` and do `cp -r` (`scp` = `ssh + cp`). (This assumes `/tmp/auth` doesn't exist; otherwise, you may end up having `/tmp/auth/tempFolder`.) Then for mounting the directory inside a Docker container inside the Linux VM, you would do
```
docker run ... -v /tmp/auth:/auth ...
```

chanseokoh:
Or, you can continue to try with the manually created `htpasswd` file. You just need to make sure that our code always uses the `testuser:testpassword` and the volume mount should be like `-v /hom/docker/auth:/auth`. This may be simpler to test.

mpeddada1:
This explanation adds so much clarity! Thank you!

> Or, you can continue to try with the manually created `htpasswd` file. You just need to make sure that our code always uses the `testuser:testpassword` and the volume mount should be like `-v /home/docker/auth:/auth`. This may be simpler to test.

I'll give this another shot because it looks like I previously made a small typo in the `docker run -v ` command. Will also look into option 1 if this ends up not working.

mpeddada1:
An update: Recent run of  `SingleProjectIntegrationTest.testBuild_simple` at [0ba258c](https://github.com/GoogleContainerTools/jib/pull/3598/commits/0ba258c96bb5394a603bfa896fc1a2022efe1022) was successfully! Looks like the access to the `htpasswd` file was indeed the issue. Will test the rest of the tests now.

mpeddada1:
All tests are passing as of cb9fdff3d740f352a9b346a4ef6c1290015f93f5! Thanks for the great advice @chanseokoh!

@elefeint Thank you for your patience with this! I'm working on a couple of clean ups at the moment. Will re-request a review once this is done.

elefeint (approved):
Awesome!

PR #3612, opened 2022-03-29
Use jackson-bom and bump jackson-databind to 2.13.2.2.

zhumin8:
- Bump jackson-databind to the latest 2.13.2.2.
- Switch and use jackson-bom to manage jackson dependency versions.


elefeint (approved):
Thank you!

PR #3661, opened 2022-05-10
upgrade from javax to jakarta namespace

elefeint:
The integration test `WarCommandTest.testWar_jetty()` intentionally uses the latest Jetty docker image. (in the future, if this causes many issues, we can likely delete that test, as there is another test pinning version).

This PR updates the sample itself to use the new Jakarta EE namespace, and updates the base images of remaining WAR tests to be on a version compatible with Jakarta EE 9.


mpeddada1 (approved):
Great fix!

Issue #3671, opened 2022-05-14
Improve error message when environment map contains null values


oliver-brm:
**Environment**:

- *Jib version:* 3.2.1
- *Build tool:* Gradle 7.4.1
- *OS:* N/A

**Description of the issue**:

Jib's checks if any of the keys and values set in the `container.environment` configuration is `null`, which is super helpful. However, the error message when an `container.environment` value is `null` currently reads only like this: "_environment map contains null values_". For someone not so deep into Jib, this could be not enough of a hint to the actual problem.

**Expected behavior**:
The error message should contain the offending keys, for example like so: "_environment map contains null values for key(s): FOO, BAR, BAZ_"

**Steps to reproduce**:

  1. Use an environment map with a value set to `null`
  2. Run jib (e.g. `./gradlew jib`)

I can offer a solution via PR.

mpeddada1:
Fixed with #3672

mpeddada1:
Thank you for your contribution! `jib-maven-plugin` and `jib-gradle-plugin` 3.3.0, `jib-core` 0.22.0 and `jib-cli` 0.11.0 have been released with this fix.

PR #3672, opened 2022-05-14
Improve error message when environment map contains null values (#3671)

oliver-brm:
This PR improves jib's error message when a value in the environment map (`container.environment`) is `null`.


oliver-brm:
I ran the `goJF` Gradle task, which re-formatted `ContainerConfiguration.java`. It should be fine now.

elefeint (approved):
Thank you!

PR #3693, opened 2022-06-21
PR to test jib cli tests on CI

mpeddada1:
Running with debugger logs to get more information on jib cli tests


elefeint:
The issue is that new temurin-based images don't work with older docker. See #3694 for workaround (dropping down to older ubuntu images for temurin).

PR #3694, opened 2022-06-21
Use temurin builds based on older version of ubuntu

elefeint:
Latest versions of temurin are incompatible with older versions of docker ([source](https://github.com/adoptium/containers/issues/215#issuecomment-1142046045)).

I don't love this fix because it means we are not testing the default Jib CLI behavior, but it will unblock the build.


mpeddada1:
Nice investigation! This is a good temporary fix. We can also mention this to the kokoro team for https://github.com/GoogleContainerTools/jib/pull/3598

PR #3707, opened 2022-07-07
fix: error when base image doesn't support target platforms

### LINKED ISSUE #3563: multiplatform build should fail when base image not available for a specified platform
**Environment**: 

- *Jib version:* 3.2.0
- *Build tool:*  Gradle
- *OS:* Linux

**Description of the issue**:

Build silently ignores platforms that have been defined but the base image doesn't support.

**Expected behavior**:

Build should fail when a platform isn't available. `docker buildx` does so, i.e.
`error: failed to solve: amazoncorretto:17-alpine-jdk: no match for platform in manifest sha256:07d678945d7c61706739add66331f7bd4bd5e290aa6e0c7cbefc9379ac86b225: not found`

**Steps to reproduce**:

  1. `./gradlew jib` with configuration like below

```groovy
jib {
  to {
    image = "someimage"
  }
  from {
    // Any image that doesn't support arm64
    image = "amazoncorretto:17-alpine-jdk"
    platforms {
      platform {
        architecture = "amd64"
        os = "linux"
      }
      platform {
        architecture = "arm64"
        os = "linux"
      }
    }
  }
}
```

In our project, we added the platforms block expecting arm64 but found it to only support amd64

` docker manifest inspect -v public.ecr.aws/aws-otel-test/aws-otel-java-spark:66141a5e34a68c1954f535ac5105457514917d5a`

After some thunking, finally realized that the base image also needed to be updated for arm64

After adding to our base image

https://github.com/aws-observability/aws-otel-java-instrumentation/commit/891a9f603414b0d02028ee21a4ce392015f352c8

the image now has arm64 support as well

`docker manifest inspect -v public.ecr.aws/aws-otel-test/aws-otel-java-spark:891a9f603414b0d02028ee21a4ce392015f352c8`

Failing fast, like buildx, will make it easier / quicker to find such an issue.

gsquared94:
Fixes #3563: Converts warnings to exceptions.

- [X] Create a new issue at https://github.com/GoogleContainerTools/jib/issues/new/choose.
- [X] Verify that integration tests and unit tests are passing after the change.
- [X] Address all checkstyle issues. Refer to the [style guide](https://github.com/GoogleContainerTools/jib/blob/0ed7dca36864b6b19ff61629fc578018041fa15f/STYLE_GUIDE.md#style-guide).

### DIFF
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PlatformCheckerTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PlatformCheckerTest.java
@@ -16,14 +16,16 @@
 
 package com.google.cloud.tools.jib.builder.steps;
 
+import static com.google.common.truth.Truth.assertThat;
+import static org.junit.Assert.assertThrows;
+
 import com.google.cloud.tools.jib.api.ImageReference;
-import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.api.buildplan.Platform;
 import com.google.cloud.tools.jib.configuration.BuildContext;
 import com.google.cloud.tools.jib.configuration.ContainerConfiguration;
 import com.google.cloud.tools.jib.configuration.ImageConfiguration;
-import com.google.cloud.tools.jib.event.EventHandlers;
 import com.google.cloud.tools.jib.image.json.ContainerConfigurationTemplate;
+import com.google.cloud.tools.jib.image.json.PlatformNotFoundInBaseImageException;
 import com.google.common.collect.ImmutableSet;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -40,13 +42,11 @@ public class PlatformCheckerTest {
 
   @Mock private BuildContext buildContext;
   @Mock private ContainerConfiguration containerConfig;
-  @Mock private EventHandlers eventHandlers;
 
   @Before
   public void setUp() {
     Mockito.when(buildContext.getBaseImageConfiguration())
         .thenReturn(ImageConfiguration.builder(ImageReference.scratch()).build());
-    Mockito.when(buildContext.getEventHandlers()).thenReturn(eventHandlers);
     Mockito.when(buildContext.getContainerConfiguration()).thenReturn(containerConfig);
   }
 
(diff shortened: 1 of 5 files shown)


elefeint:
Thank you; we'll review next week.

mpeddada1 (commented):
Thank you for the fix @gsquared94! The macos failure is related to some kokoro changes we recently made. Rebasing with main should help address this.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PlatformCheckerTest.java
```diff
@@ -58,39 +58,41 @@ public void testCheckManifestPlatform_mismatch() {
...
-                "the configured platform (configured arch/configured OS) doesn't match the "
-                    + "platform (actual arch/actual OS) of the base image (scratch)"));
+    Exception ex =
+        assertThrows(
+            PlatformNotFoundInBaseImageException.class,
+            () -> PlatformChecker.checkManifestPlatform(buildContext, containerConfigJson));
+    assertEquals(
+        "the configured platform (configured arch/configured OS) doesn't match the "
+            + "platform (actual arch/actual OS) of the base image (scratch)",
+        ex.getMessage());
```
mpeddada1, line 68: Perhaps we can use Truth here to assert on the message? Same below.

`import static com.google.common.truth.Truth.assertThat`
and `assertThat().hasMessageThat().isEqualTo()`

mpeddada1 (commented):
Almost there! I had one minor comment remaining. Could we also rebase this PR with main (or "master")? It will pick up the recent kokoro macos changes which should help make the check green.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PlatformCheckerTest.java
gsquared94: Done!

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PlatformCheckerTest.java
```diff
@@ -58,39 +58,43 @@ public void testCheckManifestPlatform_mismatch() {
...
+        .hasMessageThat()
+        .isEqualTo(
+            "the configured platform (configured arch/configured OS) doesn't match the "
+                + "platform (actual arch/actual OS) of the base image (scratch)");
   }
 
   @Test
-  public void testCheckManifestPlatform_noWarningIfDefaultAmd64Linux() {
+  public void testCheckManifestPlatform_noWarningIfDefaultAmd64Linux()
+      throws PlatformNotFoundInBaseImageException {
```
mpeddada1, line 74: Oops sorry I missed this earlier. This test no longer performs an assertion so just invoking the method in the test may not give us many insights into what is expected from the method's behavior. How about we replace this test with the following?: `testCheckManifestPlatform_arm64Linux` where the architecture doesn't match but the os matches. This will help us test the `&&` in the following condition: `!(platform.getArchitecture().equals("amd64") && platform.getOs().equals("linux"))`

gsquared94: > just invoking the method in the test may not give us many insights into what is expected from the method's behavior

On the contrary I think that the passing test implies the expectation that _it should not throw any exception_. Also this [yaqs](https://yaqs.corp.google.com/eng/q/5403980279709696#a6384374680518656) answer recommends that.

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PlatformCheckerTest.java
```diff
@@ -58,39 +58,43 @@ public void testCheckManifestPlatform_mismatch() {
...
+    assertThat(ex)
+        .hasMessageThat()
+        .isEqualTo(
+            "the configured platform (configured arch/configured OS) doesn't match the "
+                + "platform (actual arch/actual OS) of the base image (scratch)");
   }
 
   @Test
-  public void testCheckManifestPlatform_noWarningIfDefaultAmd64Linux() {
+  public void testCheckManifestPlatform_noWarningIfDefaultAmd64Linux()
```
mpeddada1, line 73:
```suggestion
  public void testCheckManifestPlatform_noExceptionIfDefaultAmd64Linux()
```

mpeddada1 (approved):

elefeint (approved):

mpeddada1:
LGTM! Thanks again @gsquared94!

PR #3709, opened 2022-07-19
Use Gradle Property and Provider for creationTime and filesModificationTime

### LINKED ISSUE #3708: Enable lazy evaluation of jib.container.creationTime and jib.container.filesModificationTime
Similar to what #2727 did for image name and tags, it would be great to be able to have lazy evaluation for `creationTime` and `filesModificationTime`, e.g. to set times to the latest commit time provided by `gradle-git-properties` (see also #2317).

Currently, this is not easily possible, since the value in the plugin config is eagerly bound before the git properties are set. I.e. this does not work:

```
jib {
    container {
        creationTime = project.provider { project.ext.git['git.commit.time'] }
    }
}
tasks.jib.dependsOn generateGitProperties
tasks.jibBuildTar.dependsOn generateGitProperties
tasks.jibDockerBuild.dependsOn generateGitProperties
```

creckord:
Fixes #3708

Usage:

To use lazy evaluation for `jib.container.creationTime` or `jib.container.filesModificationTime`, the value can be provided as an ISO 8601 date-time string like this:

```
jib {
    container {
        creationTime = project.provider { value }
        filesModificationTime = project.provider { value }
    }
}
```

### DIFF
--- a/jib-gradle-plugin/README.md
+++ b/jib-gradle-plugin/README.md
@@ -36,6 +36,7 @@ For information about the project, see the [Jib project README](../README.md).
     * [Using Docker Credential Helpers](#using-docker-credential-helpers)
     * [Using Specific Credentials](#using-specific-credentials)
   * [Custom Container Entrypoint](#custom-container-entrypoint)
+  * [Reproducible Build Timestamps](#reproducible-build-timestamps)
   * [Jib Extensions](#jib-extensions)
   * [WAR Projects](#war-projects)
   * [Skaffold Integration](#skaffold-integration)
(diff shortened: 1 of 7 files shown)


mpeddada1:
Thank you @creckord! Please sign the CLA as described in the comment [above](https://github.com/GoogleContainerTools/jib/pull/3709#issuecomment-1189107096) so that we can review your contribution!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -266,11 +270,15 @@ public String getFilesModificationTime() {
     if (System.getProperty(PropertyNames.CONTAINER_FILES_MODIFICATION_TIME) != null) {
       return System.getProperty(PropertyNames.CONTAINER_FILES_MODIFICATION_TIME);
     }
-    return filesModificationTime;
+    return filesModificationTime.getOrNull();
```
elefeint, line 273: `getOrElse` with the default value of `EPOCH_PLUS_SECOND`, perhaps? Realistically, it should not be null, unless the user sets `null` explicitly, but easier to be safe.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
```diff
@@ -279,10 +287,14 @@ public String getCreationTime() {
     if (System.getProperty(PropertyNames.CONTAINER_CREATION_TIME) != null) {
       return System.getProperty(PropertyNames.CONTAINER_CREATION_TIME);
     }
-    return creationTime;
+    return creationTime.getOrNull();
```
elefeint, line 290: `getOrElse` with the default value of `EPOCH`, perhaps?

elefeint (changes requested):
Could you add unit tests for ` setCreationTime()` / `setFileModificationTime` variations -- checking for nulls, checking that the default value is kept if not overridden. This will help with Sonar's line coverage requirement, too.

Also, could you update `jib-gradle-plugin/README.md` to reflect that the two properties can now accept providers?

emmileaf (commented):
Regarding the setters (and possibly addressing the code coverage issue): would it make sense to remove them instead and have the getters set and expose the two `Property<String>`s, in a way that’s consistent with the approach from `jib.container.label`'s lazy eval ([#3242](https://github.com/GoogleContainerTools/jib/pull/3242))? (See also [#3242-comment](https://github.com/GoogleContainerTools/jib/pull/3242#discussion_r633841799) and [#3094-comment](https://github.com/GoogleContainerTools/jib/issues/3094#issuecomment-841279320))

mpeddada1:
> Regarding the setters (and possibly addressing the code coverage issue): would it make sense to remove them instead and have the getters set and expose the two `Property<String>`s, in a way that’s consistent with the approach from `jib.container.label`'s lazy eval ([#3242](https://github.com/GoogleContainerTools/jib/pull/3242))? (See also [#3242-comment](https://github.com/GoogleContainerTools/jib/pull/3242#discussion_r633841799) and [#3094-comment](https://github.com/GoogleContainerTools/jib/issues/3094#issuecomment-841279320))

+1. Now that we are using `Property`, we should be able to do away with setters. Also for reference: [Lazy Properties section in official Gradle documentation ](https://docs.gradle.org/current/userguide/lazy_configuration.html#lazy_properties) which mentions `Note that Gradle Groovy DSL generates setter methods for each Property-typed property in a task implementation`

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
creckord: This is now handled by `convention("EPOCH_PLUS_SECOND")` and `convention("EPOCH")` on the (now final) Property objects

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ContainerParameters.java
creckord: This is now handled by `convention("EPOCH_PLUS_SECOND")` and `convention("EPOCH")` on the (now final) Property objects

creckord:
I've changed the getters to return the properties directly following [this](https://github.com/GoogleContainerTools/jib/blob/026a263b22b7265ddb5323ac74351c99486cad21/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibExtension.java#L234-L242) pattern, and removed the setters. I also added some more tests to check proper handling of defaults, nulls and finalized properties.

Note however, that this changes the result of something like this in the Gradle Groovy DSL:
```
String prop = project.extensions.jib.container.creationTime
```
Previously, this would have returned the actual property value. But since we return the property object now, Groovy's implicit type conversion with `toString()` leads to something like `property(class java.lang.String, map(provider(?)))`

The build file would need to be updated to use
```
String prop = project.extensions.jib.container.creationTime.get()
```

creckord:
> Also, since the presubmit checks will require approval to re-run after new commits are pushed, feel free to drop a comment (if there is still work in-progress) when the latest commits are review-ready, and I'll re-trigger them.

Thanks. I should have everything in now.

jib-gradle-plugin/README.md
```diff
@@ -572,6 +573,21 @@ Therefore, *for example*, the following commands will be able to launch your app
 - (Java 9+) `java -cp @/app/jib-classpath-file @/app/jib-main-class-file`
 - (with shell) `java -cp $( cat /app/jib-classpath-file ) $( cat /app/jib-main-class-file )`
 
+### Reproducible Build Timestamps
+
+To ensure that a Jib build is reproducible, Jib sets the image creation time to the Unix epoch (00:00:00, January 1st, 1970 in UTC) and all file modification times to one second past the epoch by default. The the [Jib FAQ](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#why-is-my-image-created-48-years-ago) for more details on reproducible builds.
```
emmileaf, line 578:
```suggestion
To ensure that a Jib build is reproducible, Jib sets the image creation time to the Unix epoch (00:00:00, January 1st, 1970 in UTC) and all file modification times to one second past the epoch by default. See the [Jib FAQ](https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#why-is-my-image-created-48-years-ago) for more details on reproducible builds.
```

emmileaf (approved):
Small typo nitpick, otherwise LGTM!

jib-gradle-plugin/README.md
creckord: Oops, fixed

elefeint (approved):
Thank you! We've discussed the changes to the user experience, and decided that those are worth it for the improvement.

PR #3716, opened 2022-07-29
chore: Set tempFolder permissions for local registry test

emmileaf:
Set explicit permissions to `tempFolder` created for local registry integration tests to `755`.

- Currently the default permissions set for `tempFolder`  is `700`, but it seems like docker running on macOS are sometimes unable to access this.

Possibly related: https://github.com/GoogleContainerTools/jib/pull/3598


elefeint (approved):
Macos failure was a flake; rerunning fixed it.

Issue #2732, opened 2020-08-27
`containerizingMode = "packaged"` fails when jar is not already present


mar-kolya:
**Environment**:

- *Jib version:* 2.5.0
- *Build tool:* gradle
- *OS:* macos/linux

**Description of the issue**:

When `containerizingMode = "packaged"` is enabled jib tasks fail with `Obtaining project build output files failed`.

**Expected behavior**:

Build should work. :)

**Steps to reproduce**:

  1. Set `containerizingMode = "packaged"` 
  2. build container with gradle
  3. Observe failure

**Additional Information**: 

Jib tasks should depend on `jar` task if `containerizingMode = "packaged"` is used  otherwise there is no jar unless some other task builds it. Adding `dependsOn('jar')` works around the issue.

loosebazooka:
That's strange, we already do depend on the 'jar' task. Do you have a complicated build? Can you share your build.gradle with us?
In the failure config can you run `./gradlew jib --dry-run` and share that output as well?

mar-kolya:
Unfortunately I do not have a reproducer available that I can share at the moment. We do have somewhat complex setup with multiple subproject and 'centralized' config in 'allprojects' section on the top level.

As far as I can see `jib` depends on `classes`, but not on `jar` - even with `containerizingMode = "packaged"`

loosebazooka:
I cannot reproduce this locally. Jib with containerizingMode="packaged" automatically adds the jar task as a dependency in my tests.

You will have to provide us with a breaking example to move forward. I only suspect that it has something to do with your build config. (perhaps jar is disabled?) Did you try `./gradlew jib --dry-run` ?

mar-kolya:
So default java-gradle project from idea with this gradle config:

```
plugins {
    id 'java'
    id 'com.google.cloud.tools.jib' version '2.5.0'
}

group 'org.example'
version '1.0-SNAPSHOT'

repositories {
    mavenCentral()
}

dependencies {
    testCompile group: 'junit', name: 'junit', version: '4.12'
}

afterEvaluate {
    jib {
        from {
            image = "base-image"
        }
        containerizingMode = "packaged"
        container {
            appRoot = "/usr/local/app"
        }
    }
}
```
reproduces the issue.

Dry-run output:

```
$./gradlew jib --dry-run
:compileJava SKIPPED
:processResources SKIPPED
:classes SKIPPED
:jib SKIPPED
```

[...]

PR #3737, opened 2022-08-22
Use Gradle Property and Provider to enable lazy evaluation for jib.extraDirectories parameters

emmileaf:
**In this PR:** 

Enable lazy evaluation in Gradle for various `jib.extraDirectories` parameters (ref: discussions in #2732).

`jib.extraDirectories.permissions`: `Map<String, String>` => `MapProperty<String, String>`
`jib.extraDirectories.paths.path.from`: `Path` => `Property<Path>`
`jib.extraDirectories.paths.path.into`: `String` => `Property<String>`

Lazy evaluation (via a Provider) for configuring the `jib.extraDirectories.paths` parameter directly is also enabled. Previously a `Provider` was able to be passed, but gets treated as a non-provider object and resolves file paths too early.

Both use cases below are supported in this PR:

```
extraDirectories {
   paths = project.provider { [value1, value2] }
   permissions = project.provider { value }
}
```

```
extraDirectories {
   paths {
     path { 
       from = project.provider { value }
       into = project.provider { value }
       includes = project.provider { value }
       excludes = project.provider { value }
    }
  }
  permissions = project.provider { value }
}
```


emmileaf:
Hmm I’m not sure why codecov is marking the overloaded `setPaths(Provider<>)` and `setFrom(Provider<>)` as uncovered, despite the tests added in `JibPluginTest.java`. Will look into this!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
```diff
@@ -106,15 +125,15 @@ public void setPaths(Object paths) {
...
    */
   @Input
-  public Map<String, String> getPermissions() {
+  public MapProperty<String, String> getPermissions() {
     String property = System.getProperty(PropertyNames.EXTRA_DIRECTORIES_PERMISSIONS);
     if (property != null) {
-      return ConfigurationPropertyValidator.parseMapProperty(property);
+      Map<String, String> parsedPermissions =
+          ConfigurationPropertyValidator.parseMapProperty(property);
+      if (!parsedPermissions.equals(permissions.get())) {
```
elefeint, line 133: Does this check even matter? There is no harm in setting permissions to an equivalent object.

jib-gradle-plugin/src/test/resources/gradle/projects/lazy-evaluation/src/main/updated-custom-extra-dir/updated
```diff
@@ -0,0 +1 @@
+updated
```
elefeint, line 1: This content is not used for anything? If so, let's have it say "unused" instead of "updated", so people don't search for meaning when looking into this test.

elefeint (changes requested):
This looks great to me. 

Could you update documentation in `jib-gradle-plugin/README.md`?

jib-gradle-plugin/src/test/resources/gradle/projects/lazy-evaluation/src/main/updated-custom-extra-dir/updated
emmileaf: Yeah, this file was added to commit its parent directory, so that it is an existing directory for the test project used by `JibPluginTest`. The file name and contents are trivial, I will change this to something less confusing.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
emmileaf: True! I saw that both PR #3242 and #3034 both included this check in the getter, and my understanding was that we wanted to avoid setting a property multiple times here?

emmileaf:
@elefeint Thank you for the review! I’ve updated the readme and made a note in the changelog.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
```diff
@@ -98,6 +101,22 @@ public void setPaths(Object paths) {
             .collect(Collectors.toList()));
   }
 
+  /**
+   * Sets paths, for lazy evaluation where {@code paths} is a {@link Provider} of a suitable object.
+   *
+   * @param paths provider of paths to set
+   * @see #setPaths(Object)
+   */
+  public void setPaths(Provider<Object> paths) {
```
mpeddada1, line 110: Ah I could be oversimplifying this in my mind but the Provider API should allow us to avoid setters all together. Similar to what is being done in https://github.com/GoogleContainerTools/jib/pull/3242, would it be possible to just modify the getter to return a `Property` and remove the setters?

mpeddada1 (commented):
Thank you for the really detailed PR description! This is a pretty complex lazy evaluation use case to implement!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
emmileaf: Yeah I was pondering about this as well! 

With eliminating `setPermissions` this was able to work pretty straightforwardly, but here (in `setPaths` and `setFrom`) the existing logic to convert object to file paths is more complicated than just value assignment, and I couldn’t find a way for the Gradle-generated setters to handle it. 

With `setInto`, it also looked like a larger change to have the getters return `Property` since it would also change up the shared `ExtraDirectoriesConfiguration` interface. Not sure what the best approach here would be?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
```diff
@@ -98,6 +101,22 @@ public void setPaths(Object paths) {
...
+    // Convert Provider<Object> to Provider<List<ExtraDirectoryParameters>>
+    this.paths.set(
+        paths.map(
+            obj ->
+                project.files(paths).getFiles().stream()
+                    .map(file -> new ExtraDirectoryParameters(objects, project, file.toPath(), "/"))
```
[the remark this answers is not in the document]
emmileaf: Updated, and thank you for this suggestion! Also caught a hidden typo (`project.files(paths)` should have been `project.files(obj)`)when moving this out to a helper method, so yes that was definitely too many layers to parse through.

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -368,6 +368,28 @@ public void testLazyEvalForLabels() {
...
+    assertThat(checkExtraDirectories.getOutput()).contains("[/updated:755]");
+    assertThat(checkExtraDirectories.getOutput()).contains("updated-custom-extra-dir");
+  }
+
+  @Test
+  public void testLazyEvalForExtraDirectories_IndividualPaths() {
+    BuildResult checkExtraDirectories =
+        testProject.build(
+            "check-extra-directories", "-b=build-extra-dirs.gradle", "-Djib.console=plain");
+    assertThat(checkExtraDirectories.getOutput()).contains("updated-custom-extra-dir");
```
mpeddada1, line 384: How about we do this instead?
```suggestion
    assertThat(checkExtraDirectories.getOutput()).contains("extraDirectories (from): [/updated-custom-into-dir]");
```

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
```diff
@@ -368,6 +368,28 @@ public void testLazyEvalForLabels() {
...
+  @Test
+  public void testLazyEvalForExtraDirectories() {
+    BuildResult checkExtraDirectories =
+        testProject.build("check-extra-directories", "-Djib.console=plain");
+    assertThat(checkExtraDirectories.getOutput()).contains("[/updated:755]");
+    assertThat(checkExtraDirectories.getOutput()).contains("updated-custom-extra-dir");
+  }
+
+  @Test
+  public void testLazyEvalForExtraDirectories_IndividualPaths() {
```
mpeddada1, line 380: super nit:
```suggestion
  public void testLazyEvalForExtraDirectories_individualPaths() {
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
```diff
@@ -92,10 +96,28 @@ public List<ExtraDirectoryParameters> getPaths() {
...
+  public void setPaths(Provider<Object> paths) {
+    this.paths.set(paths.map(this::transformObjectToParametersList));
+  }
+
+  /**
+   * Helper method to convert {@code Object} to {@code List<ExtraDirectoryParameters>} in {@code
+   * setFrom}.
+   */
+  @Nonnull
+  private List<ExtraDirectoryParameters> transformObjectToParametersList(Object obj) {
```
mpeddada1, line 117: Since the parameter of this method already gives us an indication of what we are converting, how about we do the following?

```suggestion
  private List<ExtraDirectoryParameters> convertToExtraDirectoryParametersList(Object obj) {
```

mpeddada1 (commented):
This is looking great! Added a couple of final minor comments.

jib-gradle-plugin/src/test/resources/gradle/projects/lazy-evaluation/src/main/updated-custom-extra-dir/unused
```diff
@@ -0,0 +1 @@
+unused
```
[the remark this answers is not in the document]
emmileaf: We have similar existing test files [here](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin/src/test/resources/gradle/projects/simple/src/main/custom-extra-dir), should I just rename to something like `foo` to align with those?

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
emmileaf: The blocker here was that the path will be resolved relative to the [TestProject](https://github.com/GoogleContainerTools/jib/blob/0e731e241609684c0658337115e4370b1a93caae/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java#L67) temp folder. Relative path to test project resources is used because there is currently a check (#3550) that the directory must actually exist, hence the unused file commit above). 

This output filepath will therefore look something like `/<path-to-tmp>/junit<####>/junit<####>/src/main/updated-custom-extra-dir` which becomes hard to assert against. Not sure if there is a more elegant workaround I am missing here!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
emmileaf: Makes sense, the consideration was definitely backward compatibility here. Will keep the guidelines in mind for new parameters in the future.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
emmileaf: Sounds good!

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java
mpeddada1: Ah I see! This can be tricky especially if the root directory of the `TestProject` keeps changing. We have a couple of examples in our tests that reference the `TestProject` and use `.resolve` to concatenate on only the relevant child directories which might be helpful: https://github.com/GoogleContainerTools/jib/blob/3e694e7c7fcb3f56c333359ada9fb40450cf771c/jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/WarCommandTest.java#L84. 

Additionally, unrelated but in order to reference resource files under `src/test/resources`, this command tends comes in handy!: `Paths.get(Resources.getResource(directory).toURI())`. Here is another [example](https://github.com/GoogleContainerTools/jib/blob/5e9a65f48a1631f32afbbe1bf5b574f38bb52d9c/jib-core/src/test/java/com/google/cloud/tools/jib/api/JavaContainerBuilderTest.java#L39) in our test code.

emmileaf: Updated - thank you for the pointers here!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoriesParameters.java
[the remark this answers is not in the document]
mpeddada1: Awesome! Thank you. The method is super easy to understand.

mpeddada1 (approved):
Excellent! Looks good on my end.

elefeint (approved):
Very nice!

PR #3742, opened 2022-08-24
Avoid running executable to verify if docker is correctly installed and clarify documentation for `jib.dockerClient.executable`

mpeddada1:
Currently, in order to verify if the docker path provided (either default or specified by the user) is installed correctly, we actually run the executable. The side effect of this is that any arbitrary executable can be executed by Jib. 

This PR avoids the need to run the executable by using `Files.isExecutable` which verifies if the path provided exists and the JVM has the correct permissions to execute the path.


elefeint (approved):
Thank you!

PR #3744, opened 2022-08-25
Check if file exists when executable path is passed in through jib dockerClient.executable

mpeddada1:
Currently, in order to verify if the docker path provided (either default or specified by the user) is installed correctly, we actually run the executable. The side effect of this is that any arbitrary executable can be executed by Jib.

This PR avoids the need to run the executable by using Files.isExecutable which verifies if the path provided exists and the JVM has the correct permissions to execute the path. However in the case of verifying whether docker is on PATH, running the `docker` command seems unavoidable at the moment. Reasoning in https://github.com/GoogleContainerTools/jib/pull/3744#discussion_r955307325

Verified with `./gradlew jib-gradle-plugin:integrationTest`

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
@@ -40,6 +40,8 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.PosixFileAttributes;
+import java.nio.file.attribute.PosixFilePermission;
 import java.security.DigestException;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
```diff
@@ -103,30 +105,27 @@ public List<DescriptorDigest> getDiffIds() throws DigestException {
...
-   * Checks if Docker is installed on the user's system and accessible by running the default {@code
-   * docker} command.
+   * Checks if Docker is installed on the user's system by running the `docker` command.
    *
    * @return {@code true} if Docker is installed on the user's system and accessible
    */
   public static boolean isDefaultDockerInstalled() {
-    return isDockerInstalled(DEFAULT_DOCKER_CLIENT);
+    try {
+      new ProcessBuilder(DEFAULT_DOCKER_CLIENT.toString()).start();
```
mpeddada1, line 112: Checking for existence of a file for the situation when the docker is installed on PATH has been a little tricky since there can be several places where docker could be install on a system (this also differs from OS to OS). For checking whether the path passed through `jib.dockerClient.executable` is installed correctly, however, we no longer run executable itself.

elefeint (approved):

mpeddada1:
Thank you for the review @elefeint! Changes also succeeded on continuous build. The integration tests cover both scenarios: when default docker is installed and when docker executable is provided.  Will be merging this PR.

Issue #3733, opened 2022-08-18
Multi-module maven project with parallel builds fail with UNAUTHORIZED.


tommyulfsparre:
**Environment**:

- *Jib version: 3.2.1* 
- *Build tool: Maven - 3.8.5* 
- *OS: Linux Ubuntu/MacOS* 

**Description of the issue**:

Building a multi-module Maven project with parallelized builds fails with `UNAUTHORIZED` when each container uses the same base image. The build needs to run with an empty cache.  

Maven command: 
`mvn clean verify -Djava.util.logging.config.file=logging.properties -T 1C`

We enabled jib debug logs and could see that the request to fetch the layer didn’t contain an authorization token for one or more of the concurrent build steps.   

We were able to reproduce this issue locally and are fairly certain that this happens when the cache is empty and then gets partially populated (`manifests_configs.json` exists but not all the layers). This led us to this section: https://github.com/GoogleContainerTools/jib/blob/master/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L134-L140 that links to:   https://github.com/GoogleContainerTools/jib/issues/2220 which describes the issue we are observing. Removing this branch makes the parallel build pass. 

We are fairly certain that this is the same issue as seen in: https://github.com/GoogleContainerTools/jib/issues/2007#issuecomment-1007103136

**Expected behavior**:

Multiple parallel builds should correctly use the same base image (either cached or not) and build the proper tar artifacts.

**Steps to reproduce**:

  1. Configure a multi-module maven project (4 submodules were used for reproducing this issue)
  2. Configura all those module’s pom.xml to use jib and also use the same base image.
  3. Run maven: `mvn clean verify -Djava.util.logging.config.file=logging.properties -DskipTests -T 1C`
  4. It might take more than one try to reproduce this issue (we injected an artificial sleep in the StepRunner to force this race to happen).

**`jib-maven-plugin` Configuration**: 
```xml
<plugin>
    <groupId>com.google.cloud.tools</groupId>
    <artifactId>jib-maven-plugin</artifactId>
    <configuration>
        <from>
            <image>gcr.io/images/baseimage:2022.03-2@sha256:<sha256></image>
            <credHelper>gcloud</credHelper>
        </from>
            <to>
                <image>gcr.io/${project}/${artifactId}:${version}</image>
            </to>
    </configuration>
</plugin>
```

**Log output**: 
```
[...]

Issue #2220, opened 2020-01-07
Possible 401 Unauthorized when base image is referred to by a digest and manifest is cached


chanseokoh:
Observed this in [#2215](https://github.com/GoogleContainerTools/jib/issues/2215#issuecomment-571772325).

> If the Jib cache is empty, Docker Hub returns 404 (and fail) in the first run. Interestingly, subsequent runs return 401. Worth understanding why.

When a base image is referred to by a digest, we first look into the manifest cache. If the manifest is in the cache, we return it along with no (i.e., null) base image credentials.

```java
    if (buildContext.isOffline() || imageReference.isTagDigest()) {
      Optional<Image> image = getCachedBaseImage();
      if (image.isPresent()) {
        return new ImageAndAuthorization(image.get(), null);
      }
```

However, if the base image cache is missing some layers (e.g., pulling base image layers failed for some reason after caching the manifest), Jib will go out and try to download the missing layers. But it may fail because we didn't return base image credentials above.

emmileaf:
Closing as fixed by layers check added in #3767.

Issue #2007, opened 2019-09-16
Jib should avoid parallel image downloads for same image


briandealwis:
Although jib is thread-safe, it should be smarter.

Jib doesn't currently lock the base image cache when downloading a base image, but instead downloads into a temporary directory and then attempts to moves the downloaded image into place.  There is no locking to block other threads.  So a Maven project with _N_ modules that build images using the same base image (like `gcr.io/distroless/java`) may result in _N_ simultaneous pulls of the same image.  Maybe we should provide a component to centralize downloading images?

_Originally posted by @briandealwis in https://github.com/GoogleContainerTools/jib/issues/1904#issuecomment-521352655_

raizoor:
@briandealwis ,

In version >= 1.5.0 JIB verifys if the base image exists on target registry and not download, doesn't it? If it's correct, the centralized folder to base images is only necessary when you don't use a registry.

TadCordle:
That is true, and even when you don't target a registry, Jib still caches base images locally to skip unnecessary pulls. So I think this issue is really only a problem for clean multi-module builds.

chanseokoh:
And this only matters when you enable the Maven parallel builds (`-T <thread counts>`) or run multiple Jib processes simultaneously on the same host. (And only when the base image layers have never been cached.) Does not affect correctness even in those cases.

mpeddada1:
@chanseokoh assuming this can be closed after #2780?

chanseokoh:
@mpeddada1 this issue is a different one. #2780 was only in the context of multi-arch support.

nhoughto:
I've been trying to incorporate Jib into our build process and one problem i'm encountering is that Jib doesn't seem to try and detect and de-dupe inflight build work. So if N requests to build the same new image come in at the same time, it will do the work N times (because the base/application cache is a cache miss).

Obviously the cache directories exist to solve for avoiding work when the work has already been done, but in the scenario where the work hasn't been done yet but many requests to do the same thing are being processed concurrently there is potential for detecting and optimising this scenario I think?

My scenario is a monorepo with N images with the base/first few layers with only the last layer differing per image, so a request to build all of them at once means jib builds the base/shared layers N times rather than once, which makes things take a lot longer than when doing the same thing under docker as it de-dupes work.

[...]

Issue #3740, opened 2022-08-24
BuildStepsExecutionException when running jibDockerBuild 


bwgjoseph:
**Environment**:

- *Jib version:* 3.2.1 (jib-gradle)
- *Build tool:* gradle 7.5
- *OS:* Windows 10 x64
- *JDK:* temurin-jdk-17.0.4+8

**Description of the issue**:

When running the following command `./gradlew jibDockerBuild`, it throws the following error

```logs
BuildStepsExecutionException: unable to move: %APPDATA%\Local\Google\Jib\Cache\tmp\12345 to %APPDATA%\Local\Google\Jib\Cache\layers\54321; such failures are often caused by interference from antivirus (#3127), or rarely if the operation is not supported by the file system (for example: special non-local file system)
```

**Expected behavior**:

Running the command should build a local image using the local Docker Daemon

**Steps to reproduce**:

  1. run `./gradlew jibDockerBuild`

**`jib-gradle-plugin` Configuration**: 
```groovy
jib {
  from {
    image = 'eclipse-temurin:17.0.4_8-jdk'
  }
  to {
    image = 'my-image/built-with-jib'
    tags = ['tag2', 'latest']
  }
  container {
    jvmFlags = ['-Xms512m', '-Xdebug']
    ports = ['8080']
  }
}
```

**Log output**: 

**Additional Information**: 

The error message points me to this https://github.com/GoogleContainerTools/jib/issues/1635#issuecomment-484199610 where it did mentioned about anti-virus affecting it. My research also brings me to #3127 and #1438, similarly, my machine is installed with `mcafee endpoint security 10.7` but I have no control over it.

I have tried to run with the following command `./gradlew jibDockerBuild -Djib.applicationCache=D:\cache -Djib.seralize=true` based on some suggestion I saw over the different issues, but it does not seem to work as well.

**Note**: I should probably also mention that it used to work when I last ran it a few months back. I did a trace, but I don't see any major upgrade in between this period for `gradle/jib` but I'm not sure about anti-virus though.

I am unable to provide a full log as I'm working on an air-gapped environment. Let me know if anything else is required, and I will try to provide

Thanks

chanseokoh:
[...]

PR #3767, opened 2022-09-06
Check for existence of layers in cache before returning cached base image

emmileaf:
**In this PR:**

Implementation of a proposed solution in #3733’s discussions, to check for presence of layer files in cache in addition to the manifest json before returning the cached base image(s). This would also address the unauthorized error described in #2220 and also reported in #2007’s comments.

**Details:**
- Adds `allLayersCached()` method to `Cache` and `CacheStorageReader` that checks for existence of cached layers for all layers described in the image manifest.
- Updates `PullBaseImageStep.getCachedBaseImages()` to invoke this check, and returns no cached images if layers are missing.

**Two approaches for layer check:**
- [`755dcfc`](https://github.com/GoogleContainerTools/jib/commit/755dcfc54990b546117d497b95d0e128513a095d) is a simpler check which verifies that the corresponding layer directory exists in the cache (but not that its contents are valid)
- [`b231d79`](https://github.com/GoogleContainerTools/jib/commit/b231d794c9e717c7f0cfa4feb89ea5d441869400) checks layers by making a call to retrieve the layer, though the retrieved layer is unused and the later `ObtainBaseImageLayerStep` would retrieve it again 

Open to suggestions on what the better approach might be here - the second option does redundant work, but is a more rigorous validation.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -458,8 +458,8 @@ class PullBaseImageStep implements Callable<ImagesAndRegistryClient> {
       Verify.verify(manifestsAndConfigs.size() == 1);
       ManifestTemplate manifest = manifestsAndConfigs.get(0).getManifest();
 
-      if (!baseImageLayersCache.verifyCachedLayers(Verify.verifyNotNull(manifest))) {
-        // Not all layers present in cache
+      // Verify all layers described in manifest are present in cache
+      if (!baseImageLayersCache.allLayersCached(Verify.verifyNotNull(manifest))) {
         return Collections.emptyList();
       }
 
@@ -492,8 +492,8 @@ class PullBaseImageStep implements Callable<ImagesAndRegistryClient> {
       }
 
       ManifestTemplate manifest = Verify.verifyNotNull(manifestAndConfigFound.get().getManifest());
-      if (!baseImageLayersCache.verifyCachedLayers(manifest)) {
-        // Not all layers present in cache
+      // Verify all layers described in manifest are present in cache
+      if (!baseImageLayersCache.allLayersCached(manifest)) {
         return Collections.emptyList();
       }
 
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
@@ -445,8 +445,8 @@ class PullBaseImageStep implements Callable<ImagesAndRegistryClient> {
           LayerCountMismatchException, UnlistedPlatformInManifestListException,
           PlatformNotFoundInBaseImageException {
     ImageReference baseImage = buildContext.getBaseImageConfiguration().getImage();
-    Optional<ImageMetadataTemplate> metadata =
-        buildContext.getBaseImageLayersCache().retrieveMetadata(baseImage);
+    Cache baseImageLayersCache = buildContext.getBaseImageLayersCache();
+    Optional<ImageMetadataTemplate> metadata = baseImageLayersCache.retrieveMetadata(baseImage);
     if (!metadata.isPresent()) {
       return Collections.emptyList();
     }
@@ -457,6 +457,12 @@ class PullBaseImageStep implements Callable<ImagesAndRegistryClient> {
     if (manifestList == null) {
       Verify.verify(manifestsAndConfigs.size() == 1);
       ManifestTemplate manifest = manifestsAndConfigs.get(0).getManifest();
+
+      if (!baseImageLayersCache.verifyCachedLayers(Verify.verifyNotNull(manifest))) {
+        // Not all layers present in cache
+        return Collections.emptyList();
+      }
+
       if (manifest instanceof V21ManifestTemplate) {
         return Collections.singletonList(
             JsonToImageTranslator.toImage((V21ManifestTemplate) manifest));
@@ -486,6 +492,11 @@ class PullBaseImageStep implements Callable<ImagesAndRegistryClient> {
       }
 
       ManifestTemplate manifest = Verify.verifyNotNull(manifestAndConfigFound.get().getManifest());
+      if (!baseImageLayersCache.verifyCachedLayers(manifest)) {
+        // Not all layers present in cache
+        return Collections.emptyList();
+      }
+
       ContainerConfigurationTemplate containerConfig =
           Verify.verifyNotNull(manifestAndConfigFound.get().getConfig());
       images.add(
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
@@ -41,6 +41,36 @@ import java.util.stream.Stream;
 /** Reads from the default cache storage engine. */
 class CacheStorageReader {
 
+  boolean allImageLayersExist(ManifestTemplate manifestTemplate) {
+
+    if (manifestTemplate instanceof V21ManifestTemplate) {
+      for (DescriptorDigest layerDigest :
+          ((V21ManifestTemplate) manifestTemplate).getLayerDigests()) {
+        Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
+        if (!Files.exists(layerDirectory)) {
+          return false;
+        }
+      }
+      return true;
+
+    } else if (manifestTemplate instanceof BuildableManifestTemplate) {
+      for (BuildableManifestTemplate.ContentDescriptorTemplate layerTemplate :
+          ((BuildableManifestTemplate) manifestTemplate).getLayers()) {
+        DescriptorDigest layerDigest = layerTemplate.getDigest();
+        if (layerDigest == null) {
+          return false;
+        }
+        Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
+        if (!Files.exists(layerDirectory)) {
+          return false;
+        }
+      }
+      return true;
+    } else {
+      throw new IllegalArgumentException("Unknown manifest type: " + manifestTemplate);
+    }
+  }
+
   @VisibleForTesting
   static void verifyImageMetadata(ImageMetadataTemplate metadata, Path metadataCacheDirectory)
       throws CacheCorruptedException {
@@ -142,6 +172,11 @@ class CacheStorageReader {
               .setLayerDiffId(cacheStorageFiles.getDigestFromFilename(layerFile))
               .build());
     }
+  }
+
+  boolean verify(DescriptorDigest layerDigest) {
+    Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
+    return Files.exists(layerDirectory);
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
@@ -174,11 +174,6 @@ class CacheStorageReader {
     }
   }
 
-  boolean verify(DescriptorDigest layerDigest) {
-    Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
-    return Files.exists(layerDirectory);
-  }
-
   /**
    * Retrieves the {@link CachedLayer} for the local base image layer with the given diff ID.
    *
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
@@ -41,36 +41,6 @@ import java.util.stream.Stream;
 /** Reads from the default cache storage engine. */
 class CacheStorageReader {
 
-  boolean allImageLayersExist(ManifestTemplate manifestTemplate) {
-
-    if (manifestTemplate instanceof V21ManifestTemplate) {
-      for (DescriptorDigest layerDigest :
-          ((V21ManifestTemplate) manifestTemplate).getLayerDigests()) {
-        Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
-        if (!Files.exists(layerDirectory)) {
-          return false;
-        }
-      }
-      return true;
-
-    } else if (manifestTemplate instanceof BuildableManifestTemplate) {
-      for (BuildableManifestTemplate.ContentDescriptorTemplate layerTemplate :
-          ((BuildableManifestTemplate) manifestTemplate).getLayers()) {
-        DescriptorDigest layerDigest = layerTemplate.getDigest();
-        if (layerDigest == null) {
-          return false;
-        }
-        Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
-        if (!Files.exists(layerDirectory)) {
-          return false;
-        }
-      }
-      return true;
-    } else {
-      throw new IllegalArgumentException("Unknown manifest type: " + manifestTemplate);
-    }
-  }
-
   @VisibleForTesting
   static void verifyImageMetadata(ImageMetadataTemplate metadata, Path metadataCacheDirectory)
       throws CacheCorruptedException {
@@ -109,6 +79,42 @@ class CacheStorageReader {
 
   CacheStorageReader(CacheStorageFiles cacheStorageFiles) {
     this.cacheStorageFiles = cacheStorageFiles;
+  }
+
+  /**
+   * Returns {@code true} if all image layers described in a manifest have a corresponding file
+   * entry in the cache.
+   *
+   * @param manifest the image manifest
+   * @return a boolean
+   */
+  boolean allLayersCached(ManifestTemplate manifest) {
+
+    if (manifest instanceof V21ManifestTemplate) {
+      for (DescriptorDigest layerDigest : ((V21ManifestTemplate) manifest).getLayerDigests()) {
+        Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
+        if (!Files.exists(layerDirectory)) {
+          return false;
+        }
+      }
+      return true;
+
+    } else if (manifest instanceof BuildableManifestTemplate) {
+      for (BuildableManifestTemplate.ContentDescriptorTemplate layerTemplate :
+          ((BuildableManifestTemplate) manifest).getLayers()) {
+        DescriptorDigest layerDigest = layerTemplate.getDigest();
+        if (layerDigest == null) {
+          return false;
+        }
+        Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
+        if (!Files.exists(layerDirectory)) {
+          return false;
+        }
+      }
+      return true;
+    } else {
+      throw new IllegalArgumentException("Unknown manifest type: " + manifest);
+    }
   }
 
   /**
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
@@ -105,8 +105,18 @@ class CacheStorageReader {
     }
 
     for (DescriptorDigest layerDigest : layerDigests) {
-      Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
-      if (!Files.exists(layerDirectory)) {
+      //      Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
+      //      if (!Files.exists(layerDirectory)) {
+      //        return false;
+      //      }
+
+      Optional<CachedLayer> optionalCachedLayer;
+      try {
+        optionalCachedLayer = retrieve(layerDigest);
+      } catch (Exception ex) {
+        return false;
+      }
+      if (!optionalCachedLayer.isPresent()) {
         return false;
       }
     }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
@@ -90,31 +90,27 @@ class CacheStorageReader {
    */
   boolean allLayersCached(ManifestTemplate manifest) {
 
-    if (manifest instanceof V21ManifestTemplate) {
-      for (DescriptorDigest layerDigest : ((V21ManifestTemplate) manifest).getLayerDigests()) {
-        Path layerDirectory = cacheStorageFiles.getLayerDirectory(layerDigest);
-        if (!Files.exists(layerDirectory)) {
-          return false;
-        }
-      }
-      return true;
+    List<DescriptorDigest> layerDigests;
 
+    if (manifest instanceof V21ManifestTemplate) {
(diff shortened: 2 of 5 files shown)


mpeddada1:
@emmileaf Took a first look at the changes! To make sure that I'm understanding the two approaches correctly:
- Approach 1 checks for the existence of {cacheDirectory}/layers/{layerDigest} where the layer digests are read from the manifest. 
- Approach 2 actually tries to grab the cached layer given the digests and if this act results in some sort of an IO issue then the layer is deemed to not be present in the cached?

Both approaches are valid but I'm a little concerned about the performance overhead approach 2 might add especially since the change will result in the cached layer being retrieved twice. Approach 1 might be a sufficient check. In terms of readability it also makes it more explicit that we are checking for the presence of the layer. Catching exceptions also works but there is usually a risk involved with it if we're not as specific as possible since methods can throw an exception for a number of reasons and it is usually recommended to let exceptions propagate to the caller or be handled by surrounding frameworks (testing frameworks, for example). Lmk what you think.

emmileaf:
@mpeddada1 Thanks for the review! Your understanding of the two approaches is spot on. 

I also agree with the concerns here around approach 2’s redundant work overhead and relying on exceptions. The downside of approach 1's check I was trying to work around is that, [`CacheStorageWriter.writeCompressed()`](https://github.com/GoogleContainerTools/jib/blob/724a4ad41a9d3b936886925776954c08fc05184c/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java#L219) creates the `{cacheDirectory}/layers/{layerDigest}` directory first, so it is possible for this check to succeed but the layer partially written (or not valid for retrieval in some other way). Not sure how unlikely that race condition would be though.

Maybe there is some way to make this check more rigorous without fully retrieving the layer? I'll revert the last commit to switch back to approach 1, and see if I can make any improvements from there.

mpeddada1:
> Just realized this statement isn't true and I had misunderstood this cache write logic earlier - it looks like a temp folder is created for writing the layer first, and then the contents get [moved](https://github.com/GoogleContainerTools/jib/blob/724a4ad41a9d3b936886925776954c08fc05184c/jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageWriter.java#L234-L235) into `{cacheDirectory}/layers/{layerDigest}` (ref: [#879 (comment)](https://github.com/GoogleContainerTools/jib/pull/879#issuecomment-419214205)). This addresses the concern with partial layer directories, and I'll go ahead with the switch to approach 1 in this PR. Apologies for the confusion!

Ah `This changes the write to only produce a layer directory at the intended location in full. This way, cache reads never see an incomplete layer directory.` That is reassuring, thank you for verifying this!

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
```diff
@@ -81,6 +81,38 @@ static void verifyImageMetadata(ImageMetadataTemplate metadata, Path metadataCac
...
   }
 
+  /**
+   * Returns {@code true} if all image layers described in a manifest have a corresponding file
+   * entry in the cache.
+   *
+   * @param manifest the image manifest
+   * @return a boolean
+   */
+  boolean allLayersCached(ManifestTemplate manifest) {
```
mpeddada1, line 91: Would it possible to move this method directly to PullBaseImageStep? Since that is the only class where this is used and it could help us avoid having to go through an extra layer of classes to reach this method. Or does this method require any variables that are specific to `CacheStorageReader`?

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
```diff
@@ -81,6 +81,38 @@ static void verifyImageMetadata(ImageMetadataTemplate metadata, Path metadataCac
...
+      layerDigests = ((V21ManifestTemplate) manifest).getLayerDigests();
+    } else if (manifest instanceof BuildableManifestTemplate) {
+      layerDigests =
+          ((BuildableManifestTemplate) manifest)
+              .getLayers().stream()
+                  .map(BuildableManifestTemplate.ContentDescriptorTemplate::getDigest)
+                  .collect(Collectors.toList());
+    } else {
+      throw new IllegalArgumentException("Unknown manifest type: " + manifest);
+    }
```
mpeddada1, line 105: I wonder if this would break current behavior for those doing multi-platform image building? For context https://github.com/GoogleContainerTools/jib/pull/2730, we also do caching for manifest lists in addition to single manifests.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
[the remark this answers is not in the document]
emmileaf: Sound good!

emmileaf: Yeah, I was thinking about this too - this method goes through a few layers because it depends on `CacheStorageReader`’s `cacheStorageFiles` variable [here](https://github.com/GoogleContainerTools/jib/pull/3767/files/ac8a2c7f45bb227e9f35f373510356c524c0ba7c#diff-faacb2e3a1c5067517059fc35d48a2de6de8483ee3d126f79c5f89cbc7e6e780R108).

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
emmileaf: That’s a good callout, and please let me know if I'm misunderstanding anything here! 

Right now the added logic in `PullBaseImageStep.getCachedBaseImages()` never passes anything of `V22ManifestList` type into `areAllLayersCached()` calls explicitly. In the case of manifest lists, this check is [made individually](https://github.com/GoogleContainerTools/jib/pull/3767/files/ac8a2c7f45bb227e9f35f373510356c524c0ba7c#diff-220c5d50240f696a44f029f651cf4ada15dc2e82baa15085668f2bb7fbaf7cdeR495-R499) when looping over the platform-specific manifests, and returns an overall cache miss if any of the platform-specific manifests has incomplete layers. 


But, do you think there is value here to have `CacheStorageReader.areAllLayersCached()` itself handle the manifest list type, rather than rely on the code calling it? Perhaps instead of throwing an exception here, it can also just return false, and leave the rest of the behavior to existing logic?

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -297,6 +300,25 @@ public void testGetCachedBaseImages_emptyCache()
...
+          BadContainerConfigurationFormatException, UnlistedPlatformInManifestListException {
+    ImageReference imageReference = ImageReference.parse("cat");
+    Mockito.when(buildContext.getBaseImageConfiguration())
+        .thenReturn(ImageConfiguration.builder(imageReference).build());
+    ManifestTemplate manifest = Mockito.mock(ManifestTemplate.class);
+    ImageMetadataTemplate imageMetadata =
+        new ImageMetadataTemplate(
+            null, Arrays.asList(new ManifestAndConfigTemplate(manifest, null)));
+
+    Mockito.when(cache.retrieveMetadata(imageReference)).thenReturn(Optional.of(imageMetadata));
```
mpeddada1, line 316:
```suggestion
Mockito.when(cache.retrieveMetadata(imageReference)).thenReturn(Optional.of(imageMetadata));
```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -297,6 +300,25 @@ public void testGetCachedBaseImages_emptyCache()
...
+    Mockito.when(cache.allLayersCached(manifest)).thenReturn(false);
+
+    Assert.assertEquals(Arrays.asList(), pullBaseImageStep.getCachedBaseImages());
+  }
```
mpeddada1, line 320: Let's use `Truth` for newer assertions. It's a little easier to read since the expected is always on the right side and provides better error messaging. We'll have to add the following import statement:
```
import static com.google.common.truth.Truth.assertThat;
```

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
```diff
@@ -297,6 +300,25 @@ public void testGetCachedBaseImages_emptyCache()
     Assert.assertEquals(Arrays.asList(), pullBaseImageStep.getCachedBaseImages());
   }
 
+  @Test
+  public void testGetCachedBaseImages_cachedWithoutAllLayers()
```
mpeddada1, line 304: What do you think of this? 

```suggestion
  public void testGetCachedBaseImages_partiallyCached_emptyListReturned()
```

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageReaderTest.java
```diff
@@ -638,4 +638,52 @@ public void testVerifyImageMetadata_validOciImageIndex() throws CacheCorruptedEx
     CacheStorageReader.verifyImageMetadata(metadata, Paths.get("/cache/dir"));
     // should pass without CacheCorruptedException
   }
+
+  @Test
+  public void testAllLayersCached_v21SingleManifest()
+      throws IOException, CacheCorruptedException, DigestException, URISyntaxException {
+
```
mpeddada1, line 645:
```suggestion
```

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageReaderTest.java
```diff
@@ -638,4 +638,52 @@ public void testVerifyImageMetadata_validOciImageIndex() throws CacheCorruptedEx
...
+    setupCachedMetadataV21(cacheDirectory);
```
mpeddada1, line 646: Minor suggestion: If possible, let's format the test into three blocks: arrange, act and assert, with a space between each of these blocks. The `arrange` block will take care of all setup, the `act` block will call the method we're testing and the `assert` block will do all the verification.

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageReaderTest.java
```diff
@@ -638,4 +638,52 @@ public void testVerifyImageMetadata_validOciImageIndex() throws CacheCorruptedEx
...
+    ImageMetadataTemplate metadata =
+        cacheStorageReader.retrieveMetadata(ImageReference.of("test", "image", "tag")).get();
+
+    V21ManifestTemplate manifest =
+        (V21ManifestTemplate) metadata.getManifestsAndConfigs().get(0).getManifest();
+    Assert.assertEquals(2, manifest.getLayerDigests().size());
```
mpeddada1, line 652: Same as above, let's use Truth here

mpeddada1 (commented):
Thanks for the thorough explanations. We're getting close!

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
```diff
@@ -81,6 +81,38 @@ static void verifyImageMetadata(ImageMetadataTemplate metadata, Path metadataCac
...
   }
 
+  /**
+   * Returns {@code true} if all image layers described in a manifest have a corresponding file
+   * entry in the cache.
+   *
+   * @param manifest the image manifest
+   * @return a boolean
+   */
+  boolean areAllLayersCached(ManifestTemplate manifest) {
```
elefeint, line 91: @mpeddada1 For multi-module parallel builds, should this method or the part that deals with the filesystem be `synchronized`? I wonder what will happen if two different threads have a race condition and are both in this method at the same time.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -486,6 +492,11 @@ List<Image> getCachedBaseImages()
       }
 
       ManifestTemplate manifest = Verify.verifyNotNull(manifestAndConfigFound.get().getManifest());
+      // Verify all layers described in manifest are present in cache
```
elefeint, line 495: (optional) there was quite a bit of code duplication in this method even in the past between the section that deals with general and platform-based manifest processing. Could there be an opportunity for refactoring common logic out into a helper method?

elefeint (changes requested):
Broadly LGTM with my limited knowledge. A couple of random questions inline.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -457,6 +457,12 @@ List<Image> getCachedBaseImages()
     if (manifestList == null) {
       Verify.verify(manifestsAndConfigs.size() == 1);
       ManifestTemplate manifest = manifestsAndConfigs.get(0).getManifest();
+
+      // Verify all layers described in manifest are present in cache
+      if (!baseImageLayersCache.allLayersCached(Verify.verifyNotNull(manifest))) {
+        return Collections.emptyList();
+      }
```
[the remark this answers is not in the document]
emmileaf: Moved this to a line above, since this condition is applied again to the manifest in `toImage()` a few lines down

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -486,6 +492,11 @@ List<Image> getCachedBaseImages()
       }
 
       ManifestTemplate manifest = Verify.verifyNotNull(manifestAndConfigFound.get().getManifest());
+      // Verify all layers described in manifest are present in cache
+      if (!baseImageLayersCache.allLayersCached(manifest)) {
+        return Collections.emptyList();
+      }
+
```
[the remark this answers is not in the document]
emmileaf: It gets messier here to combine them I think? (Since grabbing the manifest is conditional on `manifestAndConfigFound.isPresent()`, plus it also needs to perform the `verifyNotNull` check)

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
emmileaf: Updated

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
emmileaf: Makes sense! Updated

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageReaderTest.java
emmileaf: Updated

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
emmileaf: Ah you’re right - I missed that we could just add `cacheStorageFiles` as a field and move this logic to `Cache`. 

I went ahead and tried to make this change, but noticed that the `Cache` class is set up in a way that many of its methods are wrappers around calls to either `CacheStorageReader` or `CacheStorageWriter`. Looking at the existing test suites, `CacheStorageReaderTest` is also more straightforward to add to than `CacheTest` for unit testing the new check. 

I am tempted to leave this logic in `CacheStorageReader` just to stay consistent with the existing setup here - lmk what you think!

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
emmileaf: Sounds good!

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStepTest.java
emmileaf: Updated

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageReaderTest.java
emmileaf: Updated

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageReaderTest.java
emmileaf: Thanks for the suggestion, this makes a lot of sense! Tried to follow this idea though with the `act` and `assert` block more or less combined, since for a few of the tests had asserts both before and after the action.

jib-core/src/main/java/com/google/cloud/tools/jib/cache/CacheStorageReader.java
chanseokoh: Jib has two approaches for synchronization in different parts: 1) using atomic move (renaming) and 2) explicit locking with a lock file. (I was never a big fan of the first approach despite its simplicity, which had issues like https://github.com/GoogleContainerTools/jib/issues/3740#issuecomment-1225967476 and #2318. That's why we started using the second approach for stuff like caching the metadata). Here, I believe layer directories are set up via atomic move, although you'll want to double-check. So whenever you start to see a layer directory, then it is there, and you're good to go. (There may be concurrent atomic moves to set up the directory, but they will silently fail once the directory is already set up by someone else.)

Note that even if we assume that it needed synchronization, `synchronized` would help only with a single Jib process. The cache should work correctly even if there are multiple JVMs running Jib.

mpeddada1: @elefeint great point, you bring up something that I didn't think of before in terms of sufficient handling of concurrency. 

@chanseokoh thanks for adding more context to this! You're right, @emmileaf discovered that [we do maintain atomicity of writes](https://github.com/GoogleContainerTools/jib/pull/3767#issuecomment-1242003718) when populating a cache. But similar to what you've mentioned, [there are also places that warn of insufficient synchronization](https://effectivejava.googleplex.com/third-edition#page=342).  Thinking out loud/two things that I'm curious about:

1) Noticed a [proposal document](https://github.com/GoogleContainerTools/jib/blob/0ed7dca36864b6b19ff61629fc578018041fa15f/proposals/archives/cache_v2.md#locking) which mentions that using FileLocks can result in a `OverlappingFileLockException` during concurrent runs of Jib within the same JVM. Could this be a potential risk?
2) Would synchronizing writes and reads be a major overhaul?

chanseokoh: Sorry, I wasn't saying there's any place of insufficient synchronization. It's the opposite; with the atomic move, there's no concurrency issue. (It's only that a Jib run may fail when an atomic move is blocked or not supported for any reasons. Even so, there's no data corruption.)

The proposal is an old one, and it may be that the author did not have a full understanding of `OverlappingFileLockException`. You just shouldn't do concurrent access within the same JVM, which is what [Jib is currently handling correctly](https://github.com/GoogleContainerTools/jib/blob/8004ec6f5ac21783b5121b66c4d132b95118d2c8/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java#L55-L57).

mpeddada1: OH sorry I totally misread the comment lol. 

> It's only that a Jib run may fail when an atomic move is blocked or not supported for any reasons. Even so, there's no data corruption.

Got it, that makes sense. Hence, the instance where we had to improve our error messaging to cater for filesystem specific issues with atomic moves (e.g. antivirus).

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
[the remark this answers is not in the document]
mpeddada1: Ah I see! thanks for trying this out. Hm thinking of a way in which we could remove the number of if statements since they are all testing similar things but just at different levels of granularity. Is it possible to put them in a helper method with a descriptive name?

jib-core/src/test/java/com/google/cloud/tools/jib/cache/CacheStorageReaderTest.java
mpeddada1: How about we store these calls in variables? For example:

```
cacheAfterFirstLayerDirectory = Files.createDirectories(cacheStorageFiles.getLayerDirectory(firstLayerDigest));
areAllLayersCachedAfterFirstLayer = cacheStorageReader.areAllLayersCached(manifest)
cacheAfterSecondLayerDirectory = ...
areAllLayersCachedAfterSecondLayer= ..

//Assert block
```

This is just an example so you can pick a name you think works better.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
emmileaf: Ah this is a great suggestion - will refactor this and see if I can make the methods cleaner to read.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
```diff
@@ -456,19 +456,13 @@ List<Image> getCachedBaseImages()
...
-      if (manifest instanceof V21ManifestTemplate) {
-        return Collections.singletonList(
-            JsonToImageTranslator.toImage((V21ManifestTemplate) manifest));
+      ManifestAndConfigTemplate manifestAndConfig = manifestsAndConfigs.get(0);
+      Optional<Image> cachedImage = getBaseImageIfAllLayersCached(manifestAndConfig, true);
+      if (!cachedImage.isPresent()) {
+        return Collections.emptyList();
+      } else {
+        return Collections.singletonList(cachedImage.get());
       }
```
mpeddada1, line 465: super nit: we can do away with the `else` here and return `Collections.singletonList(...)` outside the `if`.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java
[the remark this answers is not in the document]
mpeddada1: Excellent refactoring!

mpeddada1 (approved):
LGTM. This fix required in-depth knowledge into Jib's caching mechanism and you really hit it out of the park! Also made great refactorings along the way.

emmileaf:
@mpeddada1 @elefeint @chanseokoh Thank you so much for all the help here with this PR!

elefeint (approved):
Nice refactoring!

emmileaf:
Thank you! Will update changelog and merge.

herder:
Hi @emmileaf @elefeint - thank you for this fix! 

We'd love to get this available for quite a few of our builds that fail from this, so I was wondering if there is a release with this fix planned soon?

emmileaf:
Hi @herder - we currently have some work in progress on compatibility in our release build environment, but can plan for a release later this week including this change.

emmileaf:
@herder `jib-core` 0.23.0, `jib-maven-plugin` 3.3.1, and `jib-gradle-plugin` 3.3.1 have been released with this fix.

PR #3769, opened 2022-09-07
chore: refactor gradle integration test to cater for both local and kokoro release environment

mpeddada1:
Recent changes in the Kokoro release environment temporarily patch in a jvm argument (`-Djavax.net.ssl.trustStore=/var/cache/proxy.crt.jks`) causing the container build by Jib to be modified when run. Since Jib gradle integration tests try to assert on the container configuration values (including jvm args) of the generated Jib container, this causes the tests to fail with the following assertion error:
```
com.google.cloud.tools.jib.gradle.SingleProjectIntegrationTest > testBuildTar_simple FAILED
    value of:
        run()
    diff (-expected +actual):
        @@ -6,4 +6,5 @@
         cat
         1970-01-01T00:00:01Z
         1970-01-01T00:00:01Z
        +-Djavax.net.ssl.trustStore=/var/cache/proxy.crt.jks
     
        at com.google.cloud.tools.jib.gradle.SingleProjectIntegrationTest.testBuildTar_simple(SingleProjectIntegrationTest.java:499)
```

This refactoring makes the affected integration tests environment-agnostic by splitting the expected output by local or kokoro-release based on the environment the tests are running in.

### DIFF
--- a/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
+++ b/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SingleProjectIntegrationTest.java
@@ -62,10 +62,6 @@ public class SingleProjectIntegrationTest {
   private final String dockerHost =
       System.getenv("DOCKER_IP") != null ? System.getenv("DOCKER_IP") : "localhost";
 
-  private static final Boolean IS_KOKORO_RELEASE =
-      System.getenv("KOKORO_JOB_NAME") != null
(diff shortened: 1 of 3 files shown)


elefeint (commented):
This will work, but I wonder if it's not easier to just remove anything that matches "javax.net.ssl.trustStore" from output list before comparing.

mpeddada1:
Yes, that is a better idea! This way we won't have to keep track of two versions of these tests.

elefeint (approved):
Nice reformatting, too -- much easier to read.

mpeddada1:
Thank you!

PR #3772, opened 2022-09-08
test: Modify resource file for flaking integration test that compares two build times

emmileaf:
Not sure this is the cause for `testBuild_tarBase` test flakes we have been seeing in CI, but here is a guess from troubleshooting:

The errors occur when the second build takes longer than the first, despite caching mechanisms that should lead to a shorter build time.

`testExecute_simple`, the other integration test that compares two build times and has been passing successfully, has [an additional file update block](https://github.com/GoogleContainerTools/jib/blob/7ba06b4eedbeafd2e3e21666af54dcdf48f1bbff/jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java#L382-L395) so that it does not push the same hello world image to target local registry as other tests in the same class. This PR copies over this additional change, since perhaps this is applicable to `testBuild_tarBase` as well, resulting in unwanted caching behavior in the first build?

We could try merging this fix if the flaking issue resurfaces again.


elefeint (approved):
Seems worth a try.

PR #3782, opened 2022-09-19
Revert "fix: Check for existence of layers in cache before returning cached base image"

emmileaf:
Reverts GoogleContainerTools/jib#3767 - something strange is happening in continuous CI checks on all three platforms, need to look into whether they are related to this change


chanseokoh:
Seems like the Gradle plugin portal is unstable.

emmileaf:
> Seems like the Gradle plugin portal is unstable.

Ah you’re right! The issue seems to be from unavailability of https://plugins.gradle.org/m2/org/apache/maven/maven-model/3.8.1/maven-model-3.8.1.pom, which looks related to the discussion in https://github.com/gradle/gradle/issues/15406.

PR #3857, opened 2022-11-01
Simplify release steps

elefeint:
For scaffold update, I suggested they use renovate-bot the last time I was releasing.
For milestones, we don't currently use Github's milestones in Jib.
For communication, most users rely on automated methods like dependabot.

I also updated CONTRIBUTING file to not require a manual update at every release.

The only thing I'd like to remove that I can't is updating the "examples" -- it is best if the samples use the correct jib version, but there is no parent relationship. Maybe having dependabot update those ourselves would work?


emmileaf (approved):
Thank you Elena!

On the note about updating examples with dependabot - we could try adding a config for package ecosystems `maven` and `gradle` with the  `/examples` directory, but I don't know if dependabot would handle version updates for a package from within the package itself’s source repo? 🤔

elefeint:
No idea, which is why I did not do it :) 
Exercise for some time when we have time to look into it.

PR #3886, opened 2022-12-05
ci: test suite and script changes for compatibility with Kokoro ubuntu migration

### LINKED ISSUE #3928: Make build script and test suite changes required for Kokoro ubuntu migration
See #3886 for ongoing fixes (Corresponding ticket: b/261210577)

emmileaf:
Resolves #3928, with fixes needed for compatibility with kokoro ubuntu migration:

1. Docker container removal issue ([details](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1337722294))

    * When stopping docker containers as part of cleanup, exclude the hosting container that is running the build script. 

2. Javadoc issue ([details](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1337722294))

    * This is resolved by fully qualifying inner class reference in a few places, due to a change of jdk in the environment.

3. Basic authentication to local registry issue ([details](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1496523371))

    * This is resolved by creating the test htpassword file in a location mounted from the VM, as docker login for local registries cannot find the programmatically created htpassword file otherwise.

5. Reaching local registry through localhost issue ([details](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1424543613))

    * This is resolved by fetching the VM IP address and associating it with localhost in `/etc/hosts`. (Alternative considered: using this IP directly to access local registries, but this requires additional `insecure_registries` setup in docker daemon configs)
    
```
export DOCKER_IP_UBUNTU="$(/sbin/ip route|awk '/default/ { print $3 }')"
echo "${DOCKER_IP_UBUNTU} localhost" >> /etc/hosts
```

6. Integration tests that run containerized web applications and make HTTP request to `localhost:8080` (details: [jib-cli](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1425962874), [jib-gradle-plugin](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1496523371), [jib-maven-plugin](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1496612432))

    * This is resolved by using VM IP instead of localhost in tests that make http requests to server, with logic added in `getDockerHostForHttpRequest()`). (The solution in 4 does not work for these tests because `HttpUrlConnection` does not continue trying after the first failed attempt to resolve localhost fails.)


emmileaf:
Docker container removal issue resolved.
Build now fails on javadoc warning:
```
> Task :jib-core:javadoc
...
Starting process 'command '/usr/lib/jvm/java-8-openjdk-amd64/bin/javadoc''. Working directory: /tmpfs/src/github/jib/jib-core Command: /usr/lib/jvm/java-8-openjdk-amd64/bin/javadoc @/tmpfs/src/github/jib/jib-core/build/tmp/javadoc/javadoc.options
Successfully started process 'command '/usr/lib/jvm/java-8-openjdk-amd64/bin/javadoc''
/tmpfs/src/github/jib/jib-core/src/main/java/com/google/cloud/tools/jib/configuration/BuildContext.java:556: warning - Tag @link: reference not found: RegistryClient.Factory
```

emmileaf:
Re-opening to continue with testing.

emmileaf:
In attempt to reach local registry through HTTP through registry container IP instead of localhost, an additional workaround is needed here, e.g. a way to add these ephemeral IPs as insecure registries. Otherwise the following error is encountered:
```
com.google.cloud.tools.jib.registry.ManifestPullerIntegrationTest STANDARD_ERROR
    Feb 09, 2023 3:24:17 PM com.google.cloud.tools.jib.registry.LocalRegistry waitUntilReady
    INFO: DockerHost: 172.17.0.3
    Feb 09, 2023 3:24:17 PM com.google.cloud.tools.jib.registry.LocalRegistry waitUntilReady
    INFO: URL: http://172.17.0.3:5000/v2/_catalog

com.google.cloud.tools.jib.registry.ManifestPullerIntegrationTest > classMethod FAILED
    java.lang.RuntimeException: Command 'docker push 172.17.0.3:5000/busybox' failed: Get "https://172.17.0.3:5000/v2/": http: server gave HTTP response to HTTPS client
```

emmileaf:
Mapping the local registry container IP to localhost in `/etc/hosts` gets past the insecure registries issue, and presubmit is unblocked so far. On continuous, integration tests for Jib CLI also run into this problem when running the jar/war applications locally in docker:
```
com.google.cloud.tools.jib.cli.JarCommandTest > testSpringBootJar_packagedMode FAILED
    value of: getContent(...)
    expected: Hello world
    but was : null
        at com.google.cloud.tools.jib.cli.JarCommandTest.testSpringBootJar_packagedMode(JarCommandTest.java:298)
```

```
com.google.cloud.tools.jib.cli.WarCommandTest > testWar_customJettySpecified FAILED
    value of: getContent(...)
    expected: Hello world
    but was : null
        at com.google.cloud.tools.jib.cli.WarCommandTest.testWar_customJettySpecified(WarCommandTest.java:127)
```

emmileaf:
After splitting up the test suite and testing locally, logging a couple more errors to resolve:

`./gradlew clean build :jib-gradle-plugin:integrationTest --info --stacktrace`

1) A number of docker login issue that could be similar to https://github.com/GoogleContainerTools/jib/pull/3598#issuecomment-1168720770
```
com.google.cloud.tools.jib.gradle.SingleProjectIntegrationTest > testBuild_dockerDaemonBase FAILED
    java.lang.RuntimeException: Command 'docker login localhost:5000 -u testuser --password-stdin' failed: Error response from daemon: login attempt to http://localhost:5000/v2/ failed with status: 401 Unauthorized
        at com.google.cloud.tools.jib.Command.run(Command.java:79)
        at com.google.cloud.tools.jib.registry.LocalRegistry.login(LocalRegistry.java:199)
        at com.google.cloud.tools.jib.registry.LocalRegistry.pullAndPushToLocal(LocalRegistry.java:188)
        at com.google.cloud.tools.jib.gradle.SingleProjectIntegrationTest.setup(SingleProjectIntegrationTest.java:168)

```

2) In [`SpringBootProjectIntegrationTest`](https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/SpringBootProjectIntegrationTest.java) and [`WarProjectIntegrationTest`](https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/integration-test/java/com/google/cloud/tools/jib/gradle/WarProjectIntegrationTest.java), errors likely similar to [ones from jib-cli above](https://github.com/GoogleContainerTools/jib/pull/3886#issuecomment-1425962874):
```
com.google.cloud.tools.jib.gradle.SpringBootProjectIntegrationTest > testBuild_packagedMode FAILED
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at com.google.cloud.tools.jib.gradle.SpringBootProjectIntegrationTest.testBuild_packagedMode(SpringBootProjectIntegrationTest.java:62)

com.google.cloud.tools.jib.gradle.WarProjectIntegrationTest > testBuild_jettyServlet25 FAILED
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at com.google.cloud.tools.jib.gradle.WarProjectIntegrationTest.verifyBuildAndRun(WarProjectIntegrationTest.java:66)
        at com.google.cloud.tools.jib.gradle.WarProjectIntegrationTest.testBuild_jettyServlet25(WarProjectIntegrationTest.java:49)
```

emmileaf:
`./gradlew clean build :jib-maven-plugin:integrationTest --info --stacktrace`

[BuildImageMojoIntegrationTest](https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java) also encounters issue with `docker login` in setup step involving local registry

https://github.com/GoogleContainerTools/jib/blob/fa8cc958f4ddad55a9eba043fdfd7c32eddb49b6/jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java#L337-L339

```
com.google.cloud.tools.jib.maven.BuildImageMojoIntegrationTest > testBuild_dockerDaemonBase FAILED
    java.lang.RuntimeException: Command 'docker login localhost:5000 -u testuser --password-stdin' failed: Error response from daemon: login attempt to http://localhost:5000/v2/ failed with status: 401 Unauthorized
        at com.google.cloud.tools.jib.Command.run(Command.java:79)
        at com.google.cloud.tools.jib.registry.LocalRegistry.login(LocalRegistry.java:199)
        at com.google.cloud.tools.jib.registry.LocalRegistry.pullAndPushToLocal(LocalRegistry.java:188)
        at com.google.cloud.tools.jib.maven.BuildImageMojoIntegrationTest.setUp(BuildImageMojoIntegrationTest.java:382)

```

Though blocked on this early step, I suspect some tests here that make use of `localhost` through `buildAndRunWebApp` (and perhaps `pullAndRunBuiltImage`) will also need host IP workarounds similar to issues above.

jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java
```diff
@@ -119,6 +132,37 @@ public void stop() {
...
+            "'{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}'",
+            containerName);
+    try {
+      String result = new Command(dockerTokens).run();
+      // Remove single quotes and LF from result (e.g. '127.0.0.1'\n)
+      containerIp = result.replaceAll("['\n]", "");
+    } catch (InterruptedException | IOException ex) {
+      throw new RuntimeException("Could not get local registry IP for: " + containerName, ex);
+    }
+
```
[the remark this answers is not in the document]
emmileaf: @mpeddada1 Hmm yeah, the difference here is that `docker inspect` is called on a specific `containerName` to grab its container IP. So for each local registry instance this is not actually available after `docker run` is executed

https://github.com/GoogleContainerTools/jib/blob/dc24c0c3b132cd745e89ecc567ea7cf68d9b4066/jib-core/src/integration-test/java/com/google/cloud/tools/jib/registry/LocalRegistry.java#L76

I think the approach you mention would allow us to get the IP for the container that the build script runs in (correct me if I'm misunderstanding!). I haven't been able to make this work yet, but something like:
```
export DOCKER_IP="$(docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' $(hostname))"
``` 
There may still be nuances between the VM-based / container-based build environments, but I'll experiment more with the two; if there could be a fixed IP replacement, that would definitely make the test refactoring a lot more straightforward!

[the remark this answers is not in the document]
mpeddada1: I see, thanks for the context. Using the docker IP as calculated here was working for other cases but just not for web-based application so this shouldn't be a concern atm. Since the web apps ping localhost:8080, I'm curious if there is an issue with port 8080 being unreachable? I think  `sudo lsof -i :8080`  can be used to verify this 
(1) https://unix.stackexchange.com/questions/62247/how-do-i-know-what-service-is-running-on-a-particular-port-in-linux
(2) https://stackoverflow.com/questions/48352370/curl-7-couldnt-connect-to-host-what-is-wrong-in-my-settings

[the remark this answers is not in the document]
emmileaf: Oh that is a good point! Confirmed that the port is unallocated, but came across something that might help narrow down the problem: when splitting up the steps from `WarCommandTest.testWar_jetty` and executing them individually, I was able to retrieve the contents okay with `curl` command.

https://github.com/GoogleContainerTools/jib/blob/9f2d0ba16dfb02eece1b40f55c6e2f1c42c7e178/jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/WarCommandTest.java#L82

```
cd ~/jib/jib-cli/src/integration-test/resources/warTest && ./gradlew clean build war
cd ~/jib && ./gradlew jib-cli:installDist

./jib-cli/build/install/jib/bin/jib war --target docker://exploded-war jib-cli/src/integration-test/resources/warTest/build/libs/standard-war.war

docker run --rm --detach -p8080:8080 exploded-war --network=host
curl http://localhost:8080/ (curl http://<container-ip>:8080/ also worked)
```
However running the test individually still fails - this leads me to suspect the incompatibility lies in `HttpURLConnection` and the setup in `getContent(URL url)`helpers, rather than earlier build/run steps - going to dig in this direction a bit more!

https://github.com/GoogleContainerTools/jib/blob/9f2d0ba16dfb02eece1b40f55c6e2f1c42c7e178/jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/WarCommandTest.java#L160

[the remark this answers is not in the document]
emmileaf: @mpeddada1 Looping back on the use of container IP, I've reverted most of the code changes from this in [858accc](https://github.com/GoogleContainerTools/jib/pull/3886/commits/858accca9accb1db1f8066d06fcef1474e5fc6a7) given the newly suggested VM IP workaround, hopefully this is looking like a much cleaner solution now!   

Having this VM IP may also help to unblock the web app tests - I suspect the tests’ usage of `java.net.HttpUrlConnection` is failing on the first localhost resolution attempt, and does not continue to the entry added at the end of `/etc/hosts`, unlike when making a CURL request to the running container (which was able to eventually succeed). ([stackoverflow ref]( https://stackoverflow.com/questions/7478424/java-httpurlconnection-dns-resolution-with-multiple-ip-addresses))

```
curl -v localhost:8080/hello
*   Trying 127.0.0.1:8080...
* TCP_NODELAY set
* connect to 127.0.0.1 port 8080 failed: Connection refused
*   Trying 192.168.9.1:8080...
* TCP_NODELAY set
* Connected to localhost (192.168.9.1) port 8080 (#0)

```
I'm not sure why my attempts to use `[container-ip]:8080` also ran into issues earlier, but was just able to get `[vm-ip]:8080` to succeed locally for a few of these tests - going to try applying this more broadly and see if it can resolve the outstanding errors 🤞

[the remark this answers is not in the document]
mpeddada1: That's awesome news @emmileaf! That's great to know that using the VM IP ended up doing the trick. As a quick reminder, before we are ready to merge the PR, it may also be worth verifying this behavior with the continuous job since the presubmits only run a subset of tests.

emmileaf:
Presubmit and continuous seem to be passing in the new environment as of https://github.com/GoogleContainerTools/jib/pull/3886/commits/6e1992ef7d86dc229a3cf31d2ac6d2e3685d511f! Will clean up the debug logging and code changes introduced to get this PR ready for merge.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
```diff
@@ -331,4 +332,16 @@ private static String getContent(URL url) throws InterruptedException {
     }
     return null;
   }
+
+  private String getDockerHostForHttpRequest() {
```
mpeddada1, line 336: The `get` prefix is generally reserved for getters where we have one return statement. Since we have a little bit of logic here, what are your thoughts on this? 

```suggestion
  private String fetchDockerHostForHttpRequest() {
```

mpeddada1, line 336: This method seems like a good candidate for a utility class. https://github.com/GoogleContainerTools/jib/tree/master/jib-plugins-common/src/test/java/com/google/cloud/tools/jib/api looks like a fitting home for this class.

mpeddada1 (commented):
Thanks a lot Emily! Added a review. Mostly LGTM, had one comment about moving the duplicate methods to a test utility class.

jib-cli/src/integration-test/java/com/google/cloud/tools/jib/cli/JarCommandTest.java
emmileaf: Thanks for the pointer @mpeddada1! For the utility class, I adapted and moved the maven plugin's [`HttpGetVerifier`](https://github.com/GoogleContainerTools/jib/blob/master/jib-maven-plugin/src/test/java/com/google/cloud/tools/jib/maven/HttpGetVerifier.java) helper to this location, since it contains the http request logic used by both plugins for integration testing. 

A quick question for jib-cli: I also made a copy of this test helper into https://github.com/GoogleContainerTools/jib/tree/master/jib-cli/src/test/java/com/google/cloud/tools/jib/api since I don't think the `jib-cli` tests should depend on `jib-plugins-common`, but do you have more context on the setup here and whether this refactoring makes sense?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/api/HttpGetVerifier.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Verifies the response body. Repeatedly tries {@code url} at the interval of .5 seconds for up
+   * to 20 seconds until getting OK HTTP response code.
+   */
+  public static void verifyBody(String expectedBody, URL url) throws InterruptedException {
+    Assert.assertEquals(expectedBody, getContent(url));
+  }
+
+  @Nullable
+  public static String getContent(URL url) throws InterruptedException {
```
mpeddada1, line 39: Do we need this to be public?

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/api/HttpGetVerifier.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.cloud.tools.jib.blob.Blobs;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import javax.annotation.Nullable;
+import org.junit.Assert;
+
+/** Verifies the response of HTTP GET. */
+public class HttpGetVerifier {
```
mpeddada1, line 28: nit: since this class is a little more than verifying the body now, perhaps we can rename this to something that encapsulates this new behavior?  Here are some patterns recommended in the java style guide for utility classes like this one:  `FooFactory`, `BarTester` or `BazClient`

mpeddada1 (approved):
Looks great!

jib-plugins-common/src/test/java/com/google/cloud/tools/jib/api/HttpGetVerifier.java
emmileaf: Oh true, this can be kept private now that the individual tests have been updated to use `verifyBody()` for making assertions. Updated!

mpeddada1 (approved):

PR #3892, opened 2022-12-12
Fix Gradle deprecations

Sineaggi:
* Gradle has deprecated `JavaPluginConvention` in favor of `JavaPluginExtension`
* The file `com.google.cloud.tools.jib.properties` is already being generated by the gradle plugin
* The compile configuration has been replaced by implementation


emmileaf:
@Sineaggi Apologies for the number of PRs that fell off our radar - we will take some time to review them and appreciate your patience!

Sineaggi:
@emmileaf Take your time, and thanks for reaching out!

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.9.3-bin.zip
```
emmileaf, line 3: Are we able to exclude these gradle wrapper updates and keep the existing version, or is it tied to another aspect of this PR that I am missing?

build.gradle
```diff
@@ -198,24 +198,16 @@ subprojects {
...
       java.srcDir file('src/integration-test/java')
       resources.srcDir file('src/integration-test/resources')
+      compileClasspath += sourceSets.main.output + sourceSets.test.output
+      runtimeClasspath += sourceSets.main.output + sourceSets.test.output
     }
   }
 
   configurations {
     integrationTestImplementation.extendsFrom testImplementation
-    integrationTestImplementation.setCanBeResolved(true)
```
emmileaf, line 206: Ah, I’m curious to what IDE you are using to develop with? Digging back in history a bit (#2961, #2375), I think some of this `integrationTestImplementation` setup may have been added to work around some Eclipse-specific issues, but I don’t personally use Eclipse for development and don’t have a quick way to confirm. It would be great to know if they no longer causing problems though!

emmileaf (commented):
@Sineaggi Thanks for helping to address these deprecation warnings! I’ve left two quick inline comments below.

build.gradle
Sineaggi: I use Intellij, which didn't need this. If it's required for Eclipse, keeping it seems reasonable.

gradle/wrapper/gradle-wrapper.properties
Sineaggi: Nope, it's just bumping gradle to the latest patch version is one of the first things I do when working on deprecation warnings.

emmileaf: Gotcha. If it's not a blocker, could you exclude the files related to gradle build version upgrade, and keep this PR just for the deprecation-related fixes? Thanks!

build.gradle
emmileaf: Let's keep it in if it doesn't hurt.

gradle/wrapper/gradle-wrapper.properties
Sineaggi: Excluded

build.gradle
Sineaggi: Sounds good

mpeddada1:
Thank you for your contribution @Sineaggi!

PR #3890, opened 2022-12-12
Fix jackson deprecations

Sineaggi:
Fix a few jackson deprecations. `ObjectMapper.configure` has [been deprecated](https://javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/latest/com/fasterxml/jackson/databind/ObjectMapper.html#configure-com.fasterxml.jackson.databind.MapperFeature-boolean-)
```
> Task :jib:jib-core:compileJava
/Users/cwalker/git/jib/jib-core/src/main/java/com/google/cloud/tools/jib/registry/credentials/DockerConfigCredentialRetriever.java:91: warning: [deprecation] configure(MapperFeature,boolean) in ObjectMapper has been deprecated
        new ObjectMapper().configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
                          ^
/Users/cwalker/git/jib/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageSteps.java:221: warning: [deprecation] configure(MapperFeature,boolean) in ObjectMapper has been deprecated
              .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true)
              ^
2 warnings
```
The replacement being
```
Since 2.13 use {@code JsonMapper.builder().configure(...)
```

Issue #926, opened 2018-09-04
Make JibExtension public.


coollog:
From https://github.com/GoogleContainerTools/jib/issues/802#issuecomment-417839721

anuraaga:
To avoid confusion, `JibExtension` is already public but `task.setJibExtension` is not. It might be a nice time to have a common interface to avoid logic that has to be duplicated 4 times for each type of jib task too (for use with `withType`).

coollog:
@anuraaga Right - fixed the title.

coollog:
Hi @anuraaga , we have released version `0.9.11` with `setJibExtension` made public. All the tasks are also of type `JibTask`.

PR #3897, opened 2022-12-16
Use dependency injection in tasks

Sineaggi:
Replaces setters with constructor injection. Allows us to remove nullable annotation, make the private JibExtension fields final, and remove null checks.


Sineaggi:
Also, I noticed that the JibTask interface had a setJibExtension function, but it wasn't implemented by all tasks registered by the jib plugin, so there was a bit of duplication for other tasks. Not sure if that was by design or a bug. Perhaps JibTask was only meant to be implemented by the main 3 tasks? Regardless, JibTask is now empty.

Sineaggi:
Bump

Sineaggi:
@emmileaf Thanks for merging https://github.com/GoogleContainerTools/jib/pull/3890, can you check a look at this as well?

emmileaf:
Will be reviewing this along with #3892 - thanks again for your patience!

emmileaf:
@Sineaggi Is there any way we could keep the public `setJibExtension` along with introducing the constructor injection, or are these mutually exclusive? 

I'm mainly concerned this might be a breaking change for users, given the context that the `JibTask` interface was introduced in #926 to expose the public `setJibExtension`. Though the original workaround that prompted this is no longer relevant, I’m not sure if there may be other use scenarios similar to [this comment](https://github.com/GoogleContainerTools/jib/issues/802#issuecomment-590611413) that we would be breaking here with this removal.

> Also, I noticed that the JibTask interface had a setJibExtension function, but it wasn't implemented by all tasks registered by the jib plugin, so there was a bit of duplication for other tasks.

Hmm yeah, I'm not entirely sure why the skaffold tasks didn’t implement the `JibTask` interface either. But since they don’t have any custom `setJibExtension` logic, I don’t see an issue with adding `implements JibTask` to these as well, to reduce the duplication.

Sineaggi:
> Is there any way we could keep the public setJibExtension along with introducing the constructor injection, or are these mutually exclusive?

We could keep the public setter for backwards compatibility.

Perhaps we could require the extension to be non-null in the setter, but still have it initialized via the constructor?

emmileaf (commented):
> We could keep the public setter for backwards compatibility.
> Perhaps we could require the extension to be non-null in the setter, but still have it initialized via the constructor?

I think that would be a good idea. Could we update this implementation so that the extension is required to be non-null in the private field and getter, gets initialized via the constructor, but has the public setter available? Thanks!

mpeddada1 (commented):
Thanks for opening this PR and helping us improve our code @Sineaggi! One quick question from the UI perspective - Will these changes affect how the jib gradle plugin is defined in a build.gradle file?

PR #3904, opened 2022-12-29
Fixes #3898, log an info instead of o warning when entrypoint makes t…

### LINKED ISSUE #3898: Log info instead of warning when entrypoint is used
Not sure why but on 3.3.0 a warning is issued when using entrypoint instead of mainclass (and other java friends):

```
[WARNING] mainClass, extraClasspath, jvmFlags, and expandClasspathDependencies are ignored when entrypoint is specified
```

Is it possible to lower the level (idea being warning -> requires attention, error -> fails). Here it does not require any intention since it is an explicit user action so it is intentional.

rmannibucau:
…he image to ignore jvm parameters

Before filing a pull request, make sure to do the following:

- [ ] Create a new issue at https://github.com/GoogleContainerTools/jib/issues/new/choose.
- [ ] Verify that integration tests and unit tests are passing after the change.
- [ ] Address all checkstyle issues. Refer to the [style guide](https://github.com/GoogleContainerTools/jib/blob/0ed7dca36864b6b19ff61629fc578018041fa15f/STYLE_GUIDE.md#style-guide).

This helps to reduce the chance of having a pull request rejected.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
@@ -595,8 +595,13 @@ public class PluginConfigurationProcessor {
             || !rawConfiguration.getJvmFlags().isEmpty()
             || !rawExtraClasspath.isEmpty()
             || rawConfiguration.getExpandClasspathDependencies())) {
+      // in a multi-module project it is not rare to configure a ton of jvmflags, sometimes
+      // classpath
+      // but having a submodule overriding entrypoint and ignoring all that is intended when such a
+      // config is done
+      // so just log it for user awareness but as an INFO, it is NOT an error
       projectProperties.log(
-          LogEvent.warn(
+          LogEvent.info(
               "mainClass, extraClasspath, jvmFlags, and expandClasspathDependencies are ignored "
                   + "when entrypoint is specified"));
     }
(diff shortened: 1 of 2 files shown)


jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
```diff
@@ -595,8 +595,13 @@ static List<String> computeEntrypoint(
             || !rawConfiguration.getJvmFlags().isEmpty()
             || !rawExtraClasspath.isEmpty()
             || rawConfiguration.getExpandClasspathDependencies())) {
+      // in a multi-module project it is not rare to configure a ton of jvmflags, sometimes
+      // classpath
+      // but having a submodule overriding entrypoint and ignoring all that is intended when such a
+      // config is done
+      // so just log it for user awareness but as an INFO, it is NOT an error
```
mpeddada1, line 602: nit: Thanks for documenting this! This comment may be better suited as a PR description. 
Comments are usually recommended when it is not entirely clear what the code is doing but in this case, the use of `LogEvent.info` is self-explanatory so it should be safe to omit it.

mpeddada1 (changes requested):
Thank you for your contribution @rmannibucau!

jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/PluginConfigurationProcessor.java
rmannibucau: Intent was to document it shouldnt be switched to warn in code - not sure anyone would use blame to check that but happy to drop it if you think it is better

mpeddada1: > Intent was to document it shouldnt be switched to warn in code -

I see! Increasing the log severity (to warnings or errors) needs to be done thoughtfully regardless so this shouldn't be a major concern. This change would also help us make the style be consistent with other classes in the repo.

rmannibucau: Done!

mpeddada1 (approved):
LGTM! Thank you so much @rmannibucau!

PR #3926, opened 2023-02-08
test: update integration test assertions for docker inspect output

### LINKED ISSUE #3925: Fix integration test assertions related to docker inspect
Opening this issue to investigate and track CI errors recently observed in Kokoro ubuntu and macos. 
Example error log:

```
com.google.cloud.tools.jib.gradle.EmptyProjectIntegrationTest > testDockerDaemon_empty FAILED
   java.lang.AssertionError:
   Expected: a string containing "            \"Labels\": {\n                \"key1\": \"value1\",\n                \"key2\": \"value2\"\n            }"
        but: was "[
       {
           "Id": "sha256:c866f0c122f88871708eafff0d3e305a2ec8de073c2b3e1eeb99040b73624d66",
           "RepoTags": [
               "emptyimage:gradle1045891597653",
               "emptyimage:gradle1070789717143"
           ],
          ...
           "Config": {
               ...
               "Labels": {
                   "key1": "value1",
                   "key2": "value2",
                   "org.opencontainers.image.ref.name": "ubuntu",
                   "org.opencontainers.image.version": "20.04"
               }
```

There is probably an underlying `docker inspect` output format change that requires updating these test assertions.

emmileaf:
Fixes #3925. Integration tests that assert against the entire `docker inspect` output string are observing failures, likely due to a change in output format / whitespace. This PR updates them to fetch more specific fields and specify json format.

- [x] Verified by running continuous build against branch

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
@@ -127,7 +127,7 @@ public class ContainerizerIntegrationTest {
     MatcherAssert.assertThat(
         dockerInspectExposedPorts,
         CoreMatchers.containsString(
-            "\"1000/tcp\":{},\"2000/udp\":{},\"2001/udp\":{},\"2002/udp\":{},\"3000/udp\":{}"));
+            "\"1000/tcp\":{},\"2000/tcp\":{},\"2001/tcp\":{},\"2002/tcp\":{},\"3000/udp\":{}"));
     MatcherAssert.assertThat(
         dockerInspectLabels,
         CoreMatchers.containsString("\"key1\":\"value1\",\"key2\":\"value2\""));
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
@@ -127,7 +127,7 @@ public class ContainerizerIntegrationTest {
     MatcherAssert.assertThat(
         dockerInspectExposedPorts,
         CoreMatchers.containsString(
-            "\"1000/tcp\":{},\"2000/udp\":{},\"2001/udp\":{},\"2002/udp\":{},\"2003/udp\":{}"));
+            "\"1000/tcp\":{},\"2000/udp\":{},\"2001/udp\":{},\"2002/udp\":{},\"3000/udp\":{}"));
     MatcherAssert.assertThat(
         dockerInspectLabels,
         CoreMatchers.containsString("\"key1\":\"value1\",\"key2\":\"value2\""));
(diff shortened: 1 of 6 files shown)


jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
```diff
@@ -266,22 +268,20 @@ private static String pullAndRunBuiltImage(String imageReference)
...
-                + "                \"key2\": \"value2\"\n"
-                + "            }");
+    String dockerInspectExposedPorts =
+        new Command("docker", "inspect", "-f", "'{{json .Config.ExposedPorts}}'", imageReference)
+            .run();
+    String dockerInspectLabels =
+        new Command("docker", "inspect", "-f", "'{{json .Config.Labels}}'", imageReference).run();
     String history = new Command("docker", "history", imageReference).run();
+
+    MatcherAssert.assertThat(
```
mpeddada1, line 278: minor nit: how about we make `MatcherAssert` static?

mpeddada1 (approved):
Thank you @emmileaf!

jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/ContainerizerIntegrationTest.java
```diff
@@ -115,28 +115,24 @@ private static FileEntriesLayer makeLayerConfiguration(
...
-        dockerContainerConfig,
-        CoreMatchers.containsString(
-            "            \"Labels\": {\n"
-                + "                \"key1\": \"value1\",\n"
-                + "                \"key2\": \"value2\"\n"
-                + "            }"));
-    String dockerConfigEnv =
-        new Command("docker", "inspect", "-f", "{{.Config.Env}}", imageReference).run();
+        dockerInspectLabels,
+        CoreMatchers.containsString("\"key1\":\"value1\",\"key2\":\"value2\""));
```
lqiu96, line 133: (Not too familiar with Jib or these ITs) If the whitespace changes were causing the mismatch for String, would it make sense (or be possible) to convert this to something like a JsonObject and match JSON contents instead of String comparisons?

emmileaf: That's a good suggestion! Jib uses jackson instead of gson, but there's probably a similar way to do JSON comparison. I'll merge this PR to unblock CI for now, and keep an issue open to follow-up on the test refactoring.

jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java
emmileaf: Do you mean something like `import static org.hamcrest.MatcherAssert.assertThat`? (I wonder if this will introduce ambiguity with the static import for `com.google.common.truth.Truth.assertThat`, which is also present in this file.)

We might be able to remove this altogether if replaced by JSON comparison as @lqiu96 suggested below. I'll look into this as follow-up.

PR #3940, opened 2023-02-23
Fixing the problem of converting numbers to Persian and Arabic (64 to ۶۴)

sheikhoo:
When the Windows region settings are set to Persian or Arabic, the `.\mvnw compile jib:build` command encounters an error.
The error message is as follows:

```
[ERROR] : Illegal repetition near index 251
[ERROR] : ^(?:((?:[a-zA-Z\d]|(?:[a-zA-Z\d][a-zA-Z\d-]*[a-zA-Z\d]))(?:\.(?:[a-zA-Z\d]|(?:[a-zA-Z\d][a-zA-Z\d-]*[a-zA-Z\d])))*(?::\d+)?)/)?((?:[a-z\d]+(?:(?:[_.]|__|-+)[a-z\d]+)*/)*[a-z\d]+(?:(?:[_.]|__|-+)[a-z\d]+)*)(?::([\w][\w.-]{0,127}))?(?:@(sha256:[a-f0-9]{۶۴}))?$
```
Error because number 64 convert to Persian or Arabic number character "۶۴"  

My change will fix the problem

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/api/DescriptorDigest.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/api/DescriptorDigest.java
@@ -37,7 +37,7 @@ public class DescriptorDigest {
   public static final int HASH_LENGTH = 64;
 
   /** Pattern matches a SHA-256 hash - 32 bytes in lowercase hexadecimal. */
-  private static final String HASH_REGEX = String.format("[a-f0-9]{%d}", HASH_LENGTH);
+  private static final String HASH_REGEX = String.format("[a-f0-9]{%s}", HASH_LENGTH);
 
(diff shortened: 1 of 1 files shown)


mpeddada1:
Thanks for putting up a fix @sheikhoo! However, I believe the use of `%d` is intentional here as we are expecting an integer value (HASH_LENGTH). `%s` on the other hand, expects a pointer to a String and modifying `%d` to `%s` can lead to unintended results.

sheikhoo:
@mpeddada1 What do you think about using the Locale.US parameter in the String.format method? It allows to format an integer value using English characters and also fixes the problem of the regional settings affecting the formatting of numbers.
I have made changes and pushed a new commit to the pull request.

Issue #2755, opened 2020-09-03
Feature: per-platform files for multi-platform image building


chanseokoh:
I think this is a low priority work at the moment.

For multi-platform image building, the user may want to provide different binaries and files for different platforms.

wakingrufus:
I would greatly appreciate this feature

wakingrufus:
Is there currently a way this can be achieved via a custom jib extension?
looking though the API, the build plan still just has a list of platforms, I don't see a way to configure a different extraFiles layer for each platform

MariusVolkhart:
I'm very interested in this issue. My use case is that I have an application that builds using Gradle. I use a Gradle plugin to create a JRE, using jlink, for each OS/Arch the app is available for. I want to containerize the app for different architectures. I'm using `distroless/java-base-debian11` as a base image. Accordingly, I need to provide a unique set of file inputs (the arch-specific JRE) per platform.

@chanseokoh I have a design/prototype of jib changes for this that I'd appreciate feedback on. See #3941 . In short, it adds a `LayerObject`, the `PlatformDependentLayer`, that maps is essentially a `Map<Platform, FileEntriesLayer>`. `jib-core` uses the platform of the base image to select the `FileEntriesLayer` when constructing the image. The [`container-build-plan-spec.md`](https://github.com/GoogleContainerTools/jib/pull/3941/files#diff-a0854039629e9dc1e35cfa8c557a3bc04073f4ea5bd609a05c62f1cfa27653b7) in the draft PR gives a decent representation of how this works.

The prototype does work, and I've used it to make multi-arch images, but it's just a prototype.

SgtSilvio:
Might be a little off-topic: The Gradle OCI Plugin (https://github.com/SgtSilvio/gradle-oci) might be interesting for those wanting fine-grained configuration per platform.

01es:
Our use case involves different versions of a Java agent (YourKit specifically) – one for `amd64` and one for `arm64`. We package a Java agent lib file into the image and it would be very convenient if we could specify different files for different target architectures. We have contemplated mapping the agent lib as a volume, but that seems to be a bit fragile.

Per-platform files for multi-platform images would have solved this. Thanks.

PR #3941, opened 2023-02-25
Prototype having platform-dependent layers

MariusVolkhart:
Despite building Java containers, it's not uncommon for platform-specific dependencies to be needed. Common examples include custom JREs produced via jlink and native libraries used via JNI. The image for each platform will typically still have the same layers, but the contents of one of the layers will differ.

To model this, a new "platform layer" is introduced. It associates various platforms with the file entries to be used for that platform. Importantly, the layer still shows as a single layer in the overall build plan, since each image must contain 1 such layer. To achieve this, a new LayerObject, the PlatformDependentLayer, is added to the API. The jib-core project is updated to understand this new LayerObject and select the platform-dependent entry based on the platform of the base image.

Relates to #2755. **This is a prototype. Do not review for merge.**

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
@@ -19,6 +19,9 @@ package com.google.cloud.tools.jib.builder.steps;
 import com.google.cloud.tools.jib.api.LogEvent;
 import com.google.cloud.tools.jib.api.buildplan.FileEntriesLayer;
 import com.google.cloud.tools.jib.api.buildplan.FileEntry;
+import com.google.cloud.tools.jib.api.buildplan.LayerObject;
+import com.google.cloud.tools.jib.api.buildplan.Platform;
+import com.google.cloud.tools.jib.api.buildplan.PlatformDependentLayer;
 import com.google.cloud.tools.jib.blob.Blob;
 import com.google.cloud.tools.jib.builder.ProgressEventDispatcher;
 import com.google.cloud.tools.jib.builder.TimerEventDispatcher;
@@ -33,6 +36,7 @@ import java.io.IOException;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.Callable;
+import java.util.stream.Stream;
 
 /** Builds and caches application layers. */
 class BuildAndCacheApplicationLayerStep implements Callable<PreparedLayer> {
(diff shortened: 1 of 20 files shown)


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildAndCacheApplicationLayerStep.java
```diff
@@ -46,24 +50,54 @@ class BuildAndCacheApplicationLayerStep implements Callable<PreparedLayer> {
...
+          })
+          .flatMap(layerConfiguration -> {
+            if (layerConfiguration instanceof FileEntriesLayer) {
+              return Stream.of(new BuildAndCacheApplicationLayerStep(
+                  buildContext,
+                  progressEventDispatcher.newChildProducer(),
+                  layerConfiguration.getName(),
+                  (FileEntriesLayer) layerConfiguration,
+                  new Platform("amd64", "linux")));
+            } else if (layerConfiguration instanceof PlatformDependentLayer) {
```
MariusVolkhart, line 87: Something key to understand with this is that a layers gets built for every platform in the  `PlatformDependentLayer`, and only later _might_ that layer be bound to an image. It's possible for layers to be prepared and to never get used. That would happen if the build spec has more platforms than Jib has been asked to produce images for.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PreparedLayer.java
```diff
@@ -55,19 +60,27 @@ Builder setStateInTarget(StateInTarget stateInTarget) {
       return this;
     }
 
+    /** Sets the platform this layer is associated with. */
+    Builder setPlatform(@Nullable Platform platform) {
```
MariusVolkhart, line 64: Platform is `null` if the layer isn't platform-specific, meaning it applies too all platforms.

proposals/container-build-plan-spec.md
```diff
@@ -39,7 +39,30 @@ Although looking similar, the structure and semantics of similarly named propert
 
   "layers": [
     {
-      "type": "fileEntries"
+      "type": "platformDependent",
+      "entries": {
+        "amd64_linux": {
```
MariusVolkhart, line 44: I don't have great thoughts on how to represent the hybrid key `Platform` as a JSON key, so went with `<arch>_<os>`. 🤷‍♂️

tnaroska: a common format used in `Dockerfile` and other tools is `os/arch[/variant] (e.g. linux/amd64)`

PR #3944, opened 2023-02-27
Allow 2023 copyright

MariusVolkhart:


mpeddada1:
Thank you so much for the update @MariusVolkhart! The jobs seemed to be stuck here so we ended up making the edit in a separate PR (https://github.com/GoogleContainerTools/jib/pull/3976).

Issue #3949, opened 2023-03-02
Extend list of default docker clients


g0ddest:
There is a few docker compatible clients (for instance podman) and in the case of using podman instead of docker it would be better if jib will look for podman executable as it acts now with docker.
There are a few workarounds with this issue - setting custom path to executable, or making alias for alternative container software as docker, but it is not so useful.

PR #3950, opened 2023-03-02
feat: look for and use existing podman installation as well as Docker

g0ddest:
### Motivation

As said in #3949 jib works with hardcoded `docker` executable name. But there are alternative software (podman, rancher, ...) and it's good to provide out-of-box for the most popular ones.

### Implementation

I made from `DEFAULT_DOCKER_CLIENT` as list of the most popular container software (docker and podman in this PR).


jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
```diff
@@ -99,21 +99,41 @@ public List<DescriptorDigest> getDiffIds() throws DigestException {
     }
   }
 
-  /** Default path to the docker executable. */
-  public static final Path DEFAULT_DOCKER_CLIENT = Paths.get("docker");
+  /** Default paths to the docker or alternatives executables. */
+  public static final List<Path> DEFAULT_DOCKER_CLIENT = Arrays.asList(
```
sdobrovolschi, line 103: It can be private for now

sdobrovolschi, line 103: I found the CONTAINER_ENGINES name better reflecting the purpose of the variable.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerDaemonImage.java
```diff
@@ -49,7 +49,7 @@ public static DockerDaemonImage named(String imageReference)
   }
 
   private final ImageReference imageReference;
-  private Path dockerExecutable = CliDockerClient.DEFAULT_DOCKER_CLIENT;
+  private Path dockerExecutable = CliDockerClient.getExistingDefaultDocker();
```
sdobrovolschi, line 52: What will happen if a custom path to a Docker executable is set and jib identifies a Podman default installation?
I can see that this variable is also used by the CliDockerClient for command issuance.
I guess the command has to be deducted from the custom path to a container engine as well.

g0ddest: If custom path is set it will be used, this it default path to executable if none is provided.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
g0ddest: Docker name is widely hardcoded everywhere, so I keep naming convention. Otherwise I should replace every "docker" word to container.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
```diff
@@ -99,21 +99,41 @@ public List<DescriptorDigest> getDiffIds() throws DigestException {
...
-      new ProcessBuilder(DEFAULT_DOCKER_CLIENT.toString()).start();
-      return true;
-    } catch (IOException ex) {
-      return false;
+    for (Path dockerClient: DEFAULT_DOCKER_CLIENT) {
+      try {
+        new ProcessBuilder(dockerClient.toString()).start();
+        return true;
+      } catch (IOException ignore) {
+        // wait for all docker clients fails
```
OneCricketeer, line 118: I think it'd be useful to at least log an error for which client was used, and had failed

PR #3954, opened 2023-03-10
[3935] Retry on unsuccessful code too

rmannibucau:
Before filing a pull request, make sure to do the following:

- [X] Create a new issue at https://github.com/GoogleContainerTools/jib/issues/new/choose.
- [X] Verify that integration tests and unit tests are passing after the change.
- [X] Address all checkstyle issues. Refer to the [style guide](https://github.com/GoogleContainerTools/jib/blob/0ed7dca36864b6b19ff61629fc578018041fa15f/STYLE_GUIDE.md#style-guide).

This helps to reduce the chance of having a pull request rejected.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/blob/InputStreamBlob.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/blob/InputStreamBlob.java
@@ -39,16 +39,24 @@ class InputStreamBlob implements Blob {
     if (isWritten) {
       throw new IllegalStateException("Cannot rewrite Blob backed by an InputStream");
     }
+    if (inputStream.markSupported()) {
+      inputStream.mark(Integer.MAX_VALUE);
+    }
     try (InputStream inStream = this.inputStream) {
       return Digests.computeDigest(inStream, outStream);
 
     } finally {
-      isWritten = true;
+      if (inputStream.markSupported()) {
+        isWritten = false;
+        inputStream.reset();
+      } else {
+        isWritten = true;
+      }
     }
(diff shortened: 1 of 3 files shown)


fahedouch:
Hello, thank you for your effort, Is it possible to handle `502` too ? ( [issue](https://github.com/GoogleContainerTools/jib/issues/4266) raised here)

rmannibucau:
Hi @fahedouch , did you try and got and issue with 502/503? It should be in when possible.

fahedouch:
> 502

thanks a lot :D. yes I got an issue with `502`

rmannibucau:
@fahedouch ok, if this patch is not sufficient once applied, maybe clone master, apply this diff, duplicate the test, return a 502 in the test and check the exception thrown cause if with this patch you still have 502 issues it either means the request was not retryable (unlikely but possible in rare cases) or that the `IOExceptionHandler` does not handle 502 properly which shouldn't and must be fixed in the http dependency.

fahedouch:
> @fahedouch ok, if this patch is not sufficient once applied, maybe clone master, apply this diff, duplicate the test, return a 502 in the test and check the exception thrown cause if with this patch you still have 502 issues it either means the request was not retryable (unlikely but possible in rare cases) or that the `IOExceptionHandler` does not handle 502 properly which shouldn't and must be fixed in the http dependency.

Ah Sorry I tried with a tag, I will try with your PR and keep you updated. I hope that jib community move this PR forward.

rmannibucau:
do we keep it running? it starts to be old

PR #4048, opened 2023-06-17
fix the issue that makes jib stuck at reading images from local docke…

### LINKED ISSUE #4046: jib gets stuck at 'processing base image' when the base image comes from local docker daemon
### Description of the issue
When reading the base image from local docker daemon, if the base image is built from a dockerfile consisting of a relatively huge amount of instructions([example here](https://gist.github.com/EternalWind/b177820578a316dbca1bdb099984ad47)), jib will stuck at 'processing base image' forever like below.
image

### Root cause
When inspecting the base image from docker daemon, jib invokes Process.waitFor() immediately without reading from stdout. It only tries to read from stdout after waitFor() returns, i.e. the process terminates. 

If the size of the output is significant enough to overwhelm the buffer, the process would become pending until someone read from its stdout to clear out the buffer. This results in a deadlock that makes waitFor() never returns.

### Proposed fix
Updates the implementation of CliDockerClient.inspect() as below.
image

EternalWind:
Fixes #4046 🛠️

The exact issue description, root cause analysis and proposed fix plan are mentioned in the issue.

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
@@ -32,6 +32,7 @@ import com.google.common.io.ByteStreams;
 import com.google.common.io.CharStreams;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -246,11 +247,21 @@ public class CliDockerClient implements DockerClient {
       throws IOException, InterruptedException {
     Process inspectProcess =
         docker("inspect", "-f", "{{json .}}", "--type", "image", imageReference.toString());
-    if (inspectProcess.waitFor() != 0) {
-      throw new IOException(
-          "'docker inspect' command failed with error: " + getStderrOutput(inspectProcess));
+
+    try (InputStreamReader stdout =
+        new InputStreamReader(inspectProcess.getInputStream(), StandardCharsets.UTF_8)) {
+
+      String output = CharStreams.toString(stdout);
+
+      if (inspectProcess.waitFor() != 0) {
+        throw new IOException(
+            "'docker inspect' command failed with error: " + getStderrOutput(inspectProcess));
+      }
+
+      return JsonTemplateMapper.readJson(
+          new ByteArrayInputStream(output.getBytes(StandardCharsets.UTF_8)),
+          DockerImageDetails.class);
     }
-    return JsonTemplateMapper.readJson(inspectProcess.getInputStream(), DockerImageDetails.class);
   }
 
   /** Runs a {@code docker} command. */
--- a/jib-core/src/test/java/com/google/cloud/tools/jib/docker/CliDockerClientTest.java
+++ b/jib-core/src/test/java/com/google/cloud/tools/jib/docker/CliDockerClientTest.java
@@ -253,6 +253,9 @@ public class CliDockerClientTest {
               Assert.assertEquals("inspect", subcommand.get(0));
               return mockProcessBuilder;
             });
+
+    Mockito.when(mockProcess.getInputStream()).thenReturn(InputStream.nullInputStream());
+
     Mockito.when(mockProcess.waitFor()).thenReturn(1);
 
     Mockito.when(mockProcess.getErrorStream())


jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
```diff
@@ -246,11 +247,21 @@ public DockerImageDetails inspect(ImageReference imageReference)
...
     Process inspectProcess =
         docker("inspect", "-f", "{{json .}}", "--type", "image", imageReference.toString());
-    if (inspectProcess.waitFor() != 0) {
-      throw new IOException(
-          "'docker inspect' command failed with error: " + getStderrOutput(inspectProcess));
+
+    try (InputStreamReader stdout =
+        new InputStreamReader(inspectProcess.getInputStream(), StandardCharsets.UTF_8)) {
+
+      String output = CharStreams.toString(stdout);
```
chanseokoh, line 254: I have a concern. `stdout` is an input stream, and at this point, you may not have reached "EOF" of the stream. That is, the process is still running and writing to the stream, which you will miss onward. And hypothetically, you can run in the same issue again.

EternalWind: Hi. According to the documentation and source code(https://github.com/google/guava/blob/master/guava/src/com/google/common/io/CharStreams.java) of CharStreams.toString(), it reads all data from the readable. Besides, this is how other parts of this class handles outputs from the docker process.

But of course please do suggest if you have better ideas in mind. ;)

chanseokoh: It's a stream, and what the doc says is a different thing.

To be correct, you need to keep consuming the stream while the process is running (i.e., looping), because you never know when the process will start/resume writing data on the other side of the stream. And after you noticed the process is terminated, consume the stream once last time.

However, I'd be happy just to see an improvement. So, what I suggest is to read the stream once more after `process.waitFor()`. It's definitely an improvement.

EternalWind: lol. I do get what you mean but allow me to elaborate a bit more and you can correct me if I am wrong. :)

image

image

Reference for the behavior of InputStream.read() can be found here:
https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read-byte:A-
image

chanseokoh: And what's the rationale for using a Reader?

EternalWind: Oops. My bad. It is indeed a Reader in our case. Thanks for the correction. :)
Below is the case for Reader.
image

As you can see, it will still exhaust the underlying Stream. Reference can be found here. https://docs.oracle.com/javase/8/docs/api/java/io/Reader.html
image

As for the rationale of using a Reader, I was merely following the style of existing code of the same class so that it is consistent.

chanseokoh: Cool then. Please just leave the comment "CharStreams.toString() will block until the end of stream is reached." That'll save people questioning the same concern as I did.

I'll let the repo maintainer to review and approve this. Ping them if necessary.

jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
```diff
@@ -246,11 +247,21 @@ public DockerImageDetails inspect(ImageReference imageReference)
...
+
+      if (inspectProcess.waitFor() != 0) {
+        throw new IOException(
+            "'docker inspect' command failed with error: " + getStderrOutput(inspectProcess));
+      }
+
+      return JsonTemplateMapper.readJson(
+          new ByteArrayInputStream(output.getBytes(StandardCharsets.UTF_8)),
```
chanseokoh, line 262: Perhaps you can use [this](https://github.com/GoogleContainerTools/jib/blob/c2a43c5b786889f490b8d1314b5b0cd86bee3355/jib-core/src/main/java/com/google/cloud/tools/jib/json/JsonTemplateMapper.java#L118).

jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
EternalWind: Good idea! Added the comment as requested. :)

jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
EternalWind: Thanks for the tips. Updated the code to use the overload version you suggested.

pindaroli:
But any news? actually We are working with a local version

jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java
```diff
@@ -246,11 +246,20 @@ public DockerImageDetails inspect(ImageReference imageReference)
...
+
+    try (InputStreamReader stdout =
+        new InputStreamReader(inspectProcess.getInputStream(), StandardCharsets.UTF_8)) {
+
+      // CharStreams.toString() will block until the end of stream is reached.
+      String output = CharStreams.toString(stdout);
+
+      if (inspectProcess.waitFor() != 0) {
+        throw new IOException(
+            "'docker inspect' command failed with error: " + getStderrOutput(inspectProcess));
```
artemik, line 258: @EternalWind  I think it would be safer to also exhaust StdErr stream the same way like StdOut to avoid the similar "being stuck" issue because of StdErr being full?

StdErr stream may have some data regardless whether process exited with success or error, so it should be read regardless of waitFor() result.

PR #4063, opened 2023-07-05
Add from(DockerClient, DockerDaemonImage)

eddumelendez:
New method to set the DockerClient implementation.


eddumelendez:
@ldetmer could you please consider this PR? Thanks in advance

PR #4103, opened 2023-09-13
Follow up: address order permutation in MutuallyExclusiveArgsException

kaiyaok2:
Fixes #<4090> 🛠️

This PR is a follow up of the tentative PR https://github.com/GoogleContainerTools/jib/pull/4093, as the previous one is merged.  It addresses the same problem as described in the issue, but in `BuildTest` and `WarTest`, respectively. Same fix applies.


mpeddada1 (approved):
LGTM!

PR #4069, opened 2023-07-07
Upgrade to gradle 8.4, junit 5. Fix configuration cache issues.

wwadge:
This is a large MR, but mostly touches unit tests.

Fixes #4062 #3132 https://github.com/GoogleContainerTools/jib/issues/4070

PR #4139, opened 2023-11-10
Bump Gradle to 8.4, fix cache.

wwadge:
Updated https://github.com/GoogleContainerTools/jib/pull/4069 to not touch junit.


wwadge:
Note: shift from port 5000 -> 5001 is because macOS is running AirPlay Receiver on the same port (settings -> Airdrop & Handoff -> AirPlay Receiver)

loosebazooka:
Thanks for splitting it up!

loosebazooka (commented):
I understand that maybe you switching 5000 -> 5001 was needed for some local mac development, but it really does pollute this PR with a HUGE number of changed. Once again making it hard to review. I think isolating the logical changes in here from that massive change would make it easier to read and review (and not accidentally miss something).

wwadge:
> I understand that maybe you switching 5000 -> 5001 was needed for some local mac development, but it really does pollute this PR with a HUGE number of changed. Once again making it hard to review. I think isolating the logical changes in here from that massive change would make it easier to read and review (and not accidentally miss something).

Agree: removed the 5000 -> 5001 changes.

build.gradle
```diff
@@ -251,7 +271,7 @@ subprojects {
   /* JAVADOC ENFORCEMENT */
   // Fail build on javadoc warnings
   tasks.withType(Javadoc) {
-    options.addBooleanOption('Xwerror', true)
+//    options.addBooleanOption('Xwerror', true)
```
wwadge, line 274: Note: this causes hundreds of warnings (+ error) of missing javadocs. I can fix, but it will make this PR very large. I suggest I enable as part of a separate PR.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -81,7 +81,10 @@ private static List<String> computeEntrypoint(
       return entrypoint;
     }
     if (commonContainerConfigCliOptions.isJettyBaseimage()) {
-      return ImmutableList.of("java", "-jar", "/usr/local/jetty/start.jar");
+      // jetty 12+ requires explicitly enabling ee10-deploy module. See:
+      // https://eclipse.dev/jetty/documentation/jetty-12/operations-guide/index.html
+      return ImmutableList.of(
```
wwadge, line 86: Note: we're pulling in "jetty" without a version, currently that means it's pulling v12 which starts up without serving content by default. This flag makes it serve content like in versions prior to v12

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageStepsTest.java
```diff
@@ -84,6 +84,22 @@ public void tearDown() {
     tempDirectoryProvider.close();
   }
 
+  public static int getJavaVersion() {
```
wwadge, line 87: Is there a good place to move this common test code somewhere?

build.gradle
loosebazooka: yeah you can ignore javadoc for now

jib-core/src/test/java/com/google/cloud/tools/jib/builder/steps/LocalBaseImageStepsTest.java
loosebazooka: yeah you could probably put it in `jib-core/src/test/java/com/google/cloud/tools/jib/jre/Versions.java` or something

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -187,15 +189,20 @@ public void buildDocker()
...
       projectProperties.waitForLoggingThread();
     }
   }
 
   @Override
   public BuildDockerTask setJibExtension(JibExtension jibExtension) {
     this.jibExtension = jibExtension;
+    this.projectProperties =
+        GradleProjectProperties.getForProject(
+            getProject(),
```
loosebazooka, line 202: I thought using getProject was incompatible with configuration cache?

loosebazooka: oh just kidding, execution time :facepalm:

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
```diff
@@ -18,60 +18,47 @@
...
 import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.InputFiles;
 import org.gradle.api.tasks.Internal;
 
 /** Configuration of an extra directory. */
 public class ExtraDirectoryParameters implements ExtraDirectoriesConfiguration {
 
-  private Project project;
-  private Property<Path> from;
+  ConfigurableFileCollection from;
```
loosebazooka, line 35: Should this type should be `DirectoryProperty` instead of `ConfigurableFileCollection`?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2.java
```diff
@@ -139,10 +139,7 @@ private void addGradleFiles(Project project) {
     skaffoldFilesOutput.addBuild(project.getBuildFile().toPath());
 
     // Add settings.gradle
-    if (project.getGradle().getStartParameter().getSettingsFile() != null) {
-      skaffoldFilesOutput.addBuild(
-          project.getGradle().getStartParameter().getSettingsFile().toPath());
-    } else if (Files.exists(projectPath.resolve(Settings.DEFAULT_SETTINGS_FILE))) {
+    if (Files.exists(projectPath.resolve(Settings.DEFAULT_SETTINGS_FILE))) {
```
loosebazooka, line 142: I think there is some loss in fidelity here, we cannot handle the case where there is a custom settings file. We could just inject this into the task?

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/SkaffoldWatchParameters.java
```diff
@@ -17,35 +17,34 @@
...
   }
 
   /**
    * A set of absolute paths to include with skaffold watching.
    *
    * @return a set of absolute paths
    */
-  @Internal
-  public Set<Path> getBuildIncludes() {
+  @InputFiles
```
loosebazooka, line 46: I don't know how changing these from `@Internal` to `@InputFiles` affects the UP-TO-DATE-ness of these skaffold tasks, but these tasks are required to always execute, so we don't want them to be every up-to-date

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/GradleProjectPropertiesExtensionTest.java
```diff
@@ -219,7 +246,24 @@ public void setUp() {
...
+    Mockito.when(mockExtensionContainer.getByType(JavaPluginExtension.class))
+        .thenReturn(mockJavaPluginExtension);
+    Mockito.when(mockJavaPluginExtension.getSourceSets()).thenReturn(mockSourceSetContainer);
+    Mockito.when(mockJavaPluginExtension.getTargetCompatibility())
+        .thenReturn(JavaVersion.VERSION_1_8);
+    Mockito.when(mockSourceSetContainer.getByName(MAIN_SOURCE_SET_NAME)).thenReturn(mockSourceSet);
+
+    Mockito.when(mockObjectFactory.property(String.class))
+        .thenReturn(new DefaultProperty<>(null, String.class));
+    //    Mockito.when(mockObjectFactory.fileProperty()).thenReturn(new DefaultRegularFileVar());
```
loosebazooka, line 264: if we're not using this, lets remove it.

loosebazooka (commented):
okay cool, I think I got through most of this, just a few questions

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
wwadge: Unfortunately the implementation of DirectoryProperty is a bit of a mess (see: https://github.com/gradle/gradle/issues/23708). Might be able to convert it but it would be pretty messy

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/SkaffoldWatchParameters.java
wwadge: Are you sure? From the docs on @InputFiles: "task to be considered out-of-date when the file paths or **contents have changed**"; so my reasoning here was: if skaffold changed either the paths or the files themselves have changed, then run it, otherwise our inputs are unchanged so no need to run it

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2.java
wwadge: I'm not sure how to handle this, the option has been deprecated in Gradle and will be removed. See: https://docs.gradle.org/8.4/userguide/upgrading_version_8.html#configuring_custom_build_layout

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/SkaffoldWatchParameters.java
loosebazooka: This is a task called by skaffold to understand which paths to watch when using a jib build. It's data scaffold wants to set it's own watcher. Called from the skaffold process.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/FilesTaskV2.java
loosebazooka: Ah I see. Okay this works.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
loosebazooka: Kk.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/skaffold/SkaffoldWatchParameters.java
wwadge: Ok, reverted back to @Internal

wwadge:
Note: there's one more deprecated thing in place:
```> Configure project :jib-build-plan
The StartParameter.settingsFile property has been deprecated. This is scheduled to be removed in Gradle 9.0. Setting custom build file to select the default project has been deprecated. Please use 'projectDir' to specify the directory of the default project instead.
```

This is coming from the researchgate plugin:
```
        at org.gradle.StartParameter.logBuildOrSettingsFileDeprecation(StartParameter.java:507)
        at org.gradle.StartParameter.getSettingsFile(StartParameter.java:497)
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)
        at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:323)
        at net.researchgate.release.ReleasePlugin$_apply_closure5.doCall$original(ReleasePlugin.groovy:107)
        at net.researchgate.release.ReleasePlugin$_apply_closure5.doCall(ReleasePlugin.groovy)
```

There's no update to the version we're using (v3.0.2) at time of writing

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/BuildDockerTask.java
```diff
@@ -104,23 +113,16 @@ public void buildDocker()
...
-            getProject(),
-            getLogger(),
-            tempDirectoryProvider,
-            jibExtension.getConfigurationName().get());
     GlobalConfig globalConfig = GlobalConfig.readConfig();
     Future<Optional<String>> updateCheckFuture =
-        TaskCommon.newUpdateChecker(projectProperties, globalConfig, getLogger());
-    try {
+        TaskCommon.newUpdateChecker(this.projectProperties, globalConfig, getLogger());
+    try (TempDirectoryProvider tempDirectoryProvider = new TempDirectoryProvider()) {
```
mpeddada1, line 120: Nice refactoring!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
```diff
@@ -18,60 +18,47 @@
...
-    return from.get();
+    return from.getSingleFile().toPath();
   }
 
   public void setFrom(Object from) {
-    this.from.set(project.file(from).toPath());
-  }
-
-  public void setFrom(Provider<Object> from) {
-    this.from.set(from.map(obj -> project.file(obj).toPath()));
```
mpeddada1, line 74: Would you provide some context for the removal of this method? It was initially implemented to enable lazy evaluation for the extraDirectories property and is scoped as `public` so I wonder if this will break current behavior for folks.

jib-cli/src/integration-test/resources/jarTest/spring-boot/build-layered.gradle
```diff
@@ -1,12 +1,12 @@
...
-    id 'org.springframework.boot' version '2.3.7.RELEASE'
+    id 'org.springframework.boot' version '2.7.13'
     id 'io.spring.dependency-management' version '1.0.10.RELEASE'
     id 'java'
 }
 
 sourceCompatibility = '1.8'
 
-bootJar {
+tasks.named("bootJar") {
```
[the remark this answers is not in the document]
wwadge: I think bootJar should also work, reverting.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
wwadge: Added it back

wwadge:
Can't see kokoro output, but if it's like last time might be worth waiting for gradle 8.5: https://github.com/gradle/gradle/issues/25752

loosebazooka:
Yeah it's the same issue with the windows cache. One of the fixes was to move `task.create` -> `task.register`? Do we still have some `task.create` lying around?

wwadge:
Huh - just literally now Gradle bumped to 8.5, let me try to update it

loosebazooka:
Here's a few I found
```diff
diff --git a/build.gradle b/build.gradle
index a8ec1fe5..f5e1eaee 100644
--- a/build.gradle
+++ b/build.gradle
@@ -128,7 +128,7 @@ subprojects {
   }

   // Adds NullAway errorprone checks.
-  tasks.withType(JavaCompile) {
+  tasks.withType(JavaCompile).configureEach {
     if (!name.toLowerCase().contains('test')) {
       options.errorprone {
         check('NullAway', CheckSeverity.ERROR)
@@ -250,7 +250,7 @@ subprojects {

   /* JAVADOC ENFORCEMENT */
   // Fail build on javadoc warnings
-  tasks.withType(Javadoc) {
+  tasks.withType(Javadoc).configureEach {
     options.addBooleanOption('Xwerror', true)
   }
   assemble.dependsOn javadoc
```

loosebazooka:
I chatted with gradle team and this fix for windows is not available till gradle 8.6. I think we don't *need* to be doing clean, but we might also just need to disable the cache on windows.

loosebazooka:
@wwadge been a while since I've used any of this tooling, @mpeddada1 can you see what it would require to get the windows build to work here (by removing clean or by disabling caching)?

loosebazooka:
I think just remove "clean" from the windows build. Clean is unusual for gradle anyway.

wwadge:
Removed clean from kokoro (windows) build

mpeddada1:
Apologies for the delay, folks! Thanks for the suggestion @loosebazooka and making the edit @wwadge. Running the kokoro build.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
```diff
@@ -18,60 +18,55 @@
...
   }
 
   public void setFrom(Object from) {
-    this.from.set(project.file(from).toPath());
+    this.from.from(from);
   }
 
   public void setFrom(Provider<Object> from) {
-    this.from.set(from.map(obj -> project.file(obj).toPath()));
+    this.from.setFrom(from.map(obj -> fileOperations.file(obj).toPath()));
```
mpeddada1, line 69: Is there a particular reason why we are choosing to use `org.gradle.api.internal.file.FileOperations` as opposed to `org.gradle.api.Project`? From the package name, it looks like FileOperations is still an internal API which may change in future Gradle versions.

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -157,12 +192,21 @@ private static boolean isProgressFooterEnabled(Project project) {
     }
   }
 
-  private final Project project;
-  private final SingleThreadedExecutor singleThreadedExecutor = new SingleThreadedExecutor();
-  private final ConsoleLogger consoleLogger;
+  @VisibleForTesting
+  @SuppressWarnings("java:S2065") // keep sonar happy. We cannot serialize an executor
+  transient SingleThreadedExecutor singleThreadedExecutor = new SingleThreadedExecutor();
+
```
mpeddada1, line 198: @wwadge I could definitely be missing something about the Gradle 8.x updates. Could you provide a bit of context regarding the use of a transient variable here?

mpeddada1 (commented):
I think we're getting close! Added just a few questions. Thanks again so much for your patience @wwadge! Will take a look at the windows issue.

mpeddada1:
Looks like we're getting a slightly different error message now:

```
com.google.cloud.tools.jib.gradle.GradleProjectPropertiesTest > testGetMajorJavaVersion_jvm11 SKIPPED

com.google.cloud.tools.jib.gradle.JibExtensionTest > testProperties FAILED
    expected: \foo
    but was : C:\tmpfs\src\github\jib\jib-gradle-plugin\build\tmp\test\work\gradle1551860326683086368projectDir\foo
        at com.google.cloud.tools.jib.gradle.JibExtensionTest.testProperties(JibExtensionTest.java:525)
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/ExtraDirectoryParameters.java
wwadge: Removed references to FileOperations

wwadge:
> Looks like we're getting a slightly different error message now:
> 
> ```
> com.google.cloud.tools.jib.gradle.GradleProjectPropertiesTest > testGetMajorJavaVersion_jvm11 SKIPPED
> 
> com.google.cloud.tools.jib.gradle.JibExtensionTest > testProperties FAILED
>     expected: \foo
>     but was : C:\tmpfs\src\github\jib\jib-gradle-plugin\build\tmp\test\work\gradle1551860326683086368projectDir\foo
>         at com.google.cloud.tools.jib.gradle.JibExtensionTest.testProperties(JibExtensionTest.java:525)
> ```
> 
> Hm it looks like the returned file path is resolving to the absolute path instead of the expected `\foo`.

I changed the way from(..) was implemented, let's try again with the new implementation that I just pushed

jib-maven-plugin/src/test/resources/maven/projects/war_servlet25/pom-tomcat.xml
```diff
@@ -30,6 +30,11 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-war-plugin</artifactId>
+        <version>3.4.0</version>
+      </plugin>
```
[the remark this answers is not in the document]
wwadge: Not sure, but I think I had done this earlier when the PR also included junit5 stuff. Removed it.

loosebazooka:
```
com.google.cloud.tools.jib.gradle.JibExtensionTest > testProperties FAILED
    expected: \foo
    but was : C:\tmpfs\src\github\jib\jib-gradle-plugin\build\tmp\test\work\gradle5672696465279302889projectDir\foo
        at com.google.cloud.tools.jib.gradle.JibExtensionTest.testProperties(JibExtensionTest.java:525)
```

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
mpeddada1: Got it, thanks for the explanation!

jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java
```diff
@@ -418,15 +419,27 @@ public void testSkaffold() {
...
+                .map(File::toPath)
+                .collect(Collectors.toSet()))
         .containsExactly(root.resolve("watch3").toAbsolutePath());
-    assertThat(testJibExtension.getSkaffold().getWatch().getExcludes())
+    assertThat(
+            testJibExtension.getSkaffold().getWatch().getExcludes().getFiles().stream()
+                .map(File::toPath)
+                .collect(Collectors.toSet()))
         .containsExactly(root.resolve("watch4").toAbsolutePath());
   }
```
mpeddada1, line 445: It looks like there is some inconsistency in how windows and other systems resolve paths; the former resolves to the absolute path whereas the latter resolves to the relative path. For consistent test behavior, can we always test on the absolute path with the help of `toAbsolutePath()`? 

https://github.com/GoogleContainerTools/jib/blob/440bd4897c6c427cc9a491128aa76c72378d605b/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibExtensionTest.java#L511-L514

Same as what is being done in `testSkaffold()`. cc/ @loosebazooka

wwadge: [These docs](https://docs.gradle.org/current/userguide/working_with_files.html#sec:single_file_paths) have this to say:
``
Never use new File(relative path) unless passed to file() or files() or from() or other methods being defined in terms of file() or files(). Otherwise this creates a path relative to the current working directory (CWD). Gradle can make no guarantees about the location of the CWD, which means builds that rely on it may break at any time.
``

Adding toAbsolutePath() to that test....

mpeddada1: Added a few observations in https://github.com/GoogleContainerTools/jib/pull/4159#issuecomment-1866598854 after trying out a couple of things (including adjusting the tests to verify on the absolute path). The getFrom() method (pasted below) always resolves to the absolute path in both windows and ubuntu) when the file paths are provided through gradle properties but shows discrepancies when the path is provided through a System Property. 

```
@Override
@InputFile
public Path getFrom() {
   return from.getSingleFile().toPath();
}
```

Since this is not a blocker for Gradle 8.x updates, could we revert the logic of extra directories to how it was originally? This annotation is an interesting enhancement to explore but probably belongs in a separate issue due to the change in behavior that is yet to be understood further.

wwadge:
Hopefully this latest commit is close enough to how it was and works - I cannot revert completely because we used "project.file" previously. I will try to obtain a windows box to test if this one fails too for I'm still working a bit blind.

wwadge:
One more kokoru run please - I reverted that .toAbsolutePath() bit we had added to our tests. Now that I reverted how I'm handling paths, I needed to revert that .toAbsolutePath() bit too.

Seems to work on a local windows box.

loosebazooka:
@mpeddada1 I think this is ready for another review

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
```diff
@@ -30,16 +32,16 @@
...
-import org.gradle.api.tasks.SourceSetContainer;
 import org.gradle.api.tasks.TaskContainer;
 import org.gradle.api.tasks.TaskProvider;
 import org.gradle.api.tasks.bundling.Jar;
 import org.gradle.util.GradleVersion;
 
 public class JibPlugin implements Plugin<Project> {
 
-  @VisibleForTesting static final GradleVersion GRADLE_MIN_VERSION = GradleVersion.version("5.1");
+  @VisibleForTesting static final GradleVersion GRADLE_MIN_VERSION = GradleVersion.version("8.0");
```
blakeli0, line 44: Unfortunately this is a breaking change for Jib, we still have customers on Gradle versions < 8, and we don't want to stop the support for them. Since `javaPluginExtension.getSourceSets()` does not exist until Gradle 8, I don't think we have a way to add this feature without stop supporting Gradle 6 and 7 customers.

loosebazooka: Is this the only breaking change? We can make conditional calls to various things.

blakeli0:
@wwadge Thanks for submitting this PR and all the effort! @loosebazooka Thanks for putting the effort reviewing the PR! Unfortunately, as I mentioned in the [comment](https://github.com/GoogleContainerTools/jib/pull/4139#discussion_r1453934489), after careful consideration, we don't want to introduce a breaking change for customers still use Gradle < 8 at this moment. So we can not accept this PR as it is right now, we may re-visit it once we feel comfortable to move the minimum supported version for jib to 8. 

We are really sorry for the back and forth in the past few months, please let us know if you have other suggestions!

jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java
wwadge: It's from 7.1 according to: https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/JavaPluginExtension.html#getSourceSets--

We're using it in very few places so I don't think it's very difficult to support older versions. I can attempt a patch to support as from v6 if you want though I understand if you wish to put this PR on hold till some arbitrary future date.

bendb-instacart:
@blakeli0 may I ask, a year later, whether the situation has changed here enough to think about updating Gradle compatibility for jib?

hfhbd:
You can also use different variants/java features to support multiple Gradle versions.

syh7:
Gradle 5.1 was released the 2nd of january 2019. It is now october 2025. How long are you planning to support a now more than 6 year old version and prevent upgrading?

oldium:
Can you (Google guys) please suggest how to move this forward? Gradle is now 9.4.1 and configuration cache is still not supported by this Jib plugin. This feature was promoted stable in 2023.

Just to illustrate the situation (expect sarkasm — just read it, how it sounds). Jib does not support feature released almost 3 years ago, because of a compatibility with 7 years old version. Yes, seriously.

pschichtel:
@oldium you might want to check if [tiny-jib](https://github.com/pschichtel/tiny-jib) (I'm the author) works for you.

sschuberth:
> @oldium you might want to check if [tiny-jib](https://github.com/pschichtel/tiny-jib) (I'm the author) works for you.

We've taken tiny-jib into use already. I can recommend it.

oldium:
It does not support Skaffold, which is why we added Jib actually.

PR #4156, opened 2023-12-14
Add windows to gha ci

loosebazooka:
This might help with the debugging on external contribs on windows tests.


loosebazooka:
weird, I wonder how kokoro runs these?

PR #4204, opened 2024-03-02
Update apache-commons to 1.26.0 and set PAX headers to address build reproducibility issue

### LINKED ISSUE #4141: Builds not reproducible
**Environment**:

- *Jib version:* 3.4.0
- *Build tool:* Gradle 8.4
- *OS:* MacOS

**Description of the issue**:

Some of the Docker image layers are given new sha256-hashes when invoking jib after a `./gradlew clean` but otherwise do not make any changes.  Locally it appears to be a layer that contains the output of a `Copy` task that changes checksum, but the file itself remains the same, with the same checksum.  Even worse, on CI it's the "bottom" layer with the external libraries that change on every single build.  

**Expected behavior**:

Invoking Jib twice on the same git rev and with the same parameters should yield the same image every time.

**Probable cause**

When inspecting the individual layers from the `jibBuildTar` task it seems that the culprits are the `atime`, `mtime`, `ctime`, and `LIBARCHIVE.creationtime` tar headers.  They change between the builds dependning on whether input files have been deleted and recreated again.  These headers should probably be cleared in `ReproducibleLayerBuilder::setUserAndGroup` where the username/group info is also cleared.

izogfif:
Fixes #4141 🛠️

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
@@ -97,10 +97,10 @@ public class ReproducibleLayerBuilder {
   }
 
   private static void clearPaxTimeHeaders(TarArchiveEntry entry) {
-    entry.addPaxHeader("mtime", "0");
-    entry.addPaxHeader("atime", "0");
-    entry.addPaxHeader("ctime", "0");
-    entry.addPaxHeader("LIBARCHIVE.creationtime", "0");
+    entry.addPaxHeader("mtime", "1"); // EPOCH plus 1 second
+    entry.addPaxHeader("atime", "1");
+    entry.addPaxHeader("ctime", "1");
+    entry.addPaxHeader("LIBARCHIVE.creationtime", "1");
   }
 
   private static void setUserAndGroup(TarArchiveEntry entry, FileEntry layerEntry) {
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
@@ -81,9 +81,6 @@ public class ReproducibleLayerBuilder {
         dir.setGroupId(0);
         dir.setUserName("");
         dir.setGroupName("");
-        // In this particular place we have to use default modification time (1 second past epoch)
-        // instead the modification time user might've set in JIB settings. That's because we have
-        // no way to find out the modification time user might've set in JIB settings here.
         clearTimeHeaders(dir, FileEntriesLayer.DEFAULT_MODIFICATION_TIME);
         add(dir);
       }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
@@ -100,6 +100,10 @@ public class ReproducibleLayerBuilder {
     entry.setGroupId(0);
     entry.setUserName("");
     entry.setGroupName("");
+    entry.addPaxHeader("mtime", "0");
+    entry.addPaxHeader("atime", "0");
+    entry.addPaxHeader("ctime", "0");
+    entry.addPaxHeader("LIBARCHIVE.creationtime", "0");
 
     if (!layerEntry.getOwnership().isEmpty()) {
       // Parse "<user>:<group>" string.
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashSet;
@@ -80,7 +81,10 @@ public class ReproducibleLayerBuilder {
         dir.setGroupId(0);
         dir.setUserName("");
         dir.setGroupName("");
-        clearTimeHeaders(dir); // DEFAULT_MODIFICATION_TIME == EPOCH+1
+        // In this particular place we have to use default modification time (1 second past epoch)
+        // instead the modification time user might've set in JIB settings. That's because we have
+        // no way to find out the modification time user might've set in JIB settings here.
+        clearTimeHeaders(dir, FileEntriesLayer.DEFAULT_MODIFICATION_TIME);
         add(dir);
       }
 
@@ -95,12 +99,18 @@ public class ReproducibleLayerBuilder {
     }
   }
 
-  private static void clearTimeHeaders(TarArchiveEntry entry) {
-    entry.setModTime(FileEntriesLayer.DEFAULT_MODIFICATION_TIME.toEpochMilli());
-    entry.addPaxHeader("mtime", "1");
-    entry.addPaxHeader("atime", "1");
-    entry.addPaxHeader("ctime", "1");
-    entry.addPaxHeader("LIBARCHIVE.creationtime", "1");
+  private static void clearTimeHeaders(TarArchiveEntry entry, Instant modTime) {
+    entry.setModTime(modTime.toEpochMilli());
+
+    String headerTime = Long.toString(modTime.getEpochSecond());
+    final long nanos = modTime.getNano();
+    if (nanos > 0) {
+      headerTime += "." + nanos;
+    }
+    entry.addPaxHeader("mtime", headerTime);
+    entry.addPaxHeader("atime", headerTime);
+    entry.addPaxHeader("ctime", headerTime);
+    entry.addPaxHeader("LIBARCHIVE.creationtime", headerTime);
   }
 
   private static void setUserAndGroup(TarArchiveEntry entry, FileEntry layerEntry) {
@@ -165,7 +175,7 @@ public class ReproducibleLayerBuilder {
       // lowest 9 bits) then using a bitwise OR to set them to the layerEntry's permissions.
       entry.setMode((entry.getMode() & ~0777) | layerEntry.getPermissions().getPermissionBits());
       setUserAndGroup(entry, layerEntry);
-      clearTimeHeaders(entry);
+      clearTimeHeaders(entry, layerEntry.getModificationTime());
 
       uniqueTarArchiveEntries.add(entry);
     }
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
@@ -76,12 +76,11 @@ public class ReproducibleLayerBuilder {
       if (namePath.getParent() != namePath.getRoot()) {
         Path tarArchiveParentDir = Verify.verifyNotNull(namePath.getParent());
         TarArchiveEntry dir = new TarArchiveEntry(DIRECTORY_FILE, tarArchiveParentDir.toString());
-        dir.setModTime(FileEntriesLayer.DEFAULT_MODIFICATION_TIME.toEpochMilli());
         dir.setUserId(0);
         dir.setGroupId(0);
         dir.setUserName("");
         dir.setGroupName("");
-        clearPaxTimeHeaders(dir);
+        clearTimeHeaders(dir); // DEFAULT_MODIFICATION_TIME == EPOCH+1
         add(dir);
       }
 
@@ -96,8 +95,9 @@ public class ReproducibleLayerBuilder {
     }
   }
 
-  private static void clearPaxTimeHeaders(TarArchiveEntry entry) {
-    entry.addPaxHeader("mtime", "1"); // EPOCH plus 1 second
+  private static void clearTimeHeaders(TarArchiveEntry entry) {
+    entry.setModTime(FileEntriesLayer.DEFAULT_MODIFICATION_TIME.toEpochMilli());
+    entry.addPaxHeader("mtime", "1");
     entry.addPaxHeader("atime", "1");
     entry.addPaxHeader("ctime", "1");
     entry.addPaxHeader("LIBARCHIVE.creationtime", "1");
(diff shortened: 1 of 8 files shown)


izogfif:
This is a copy of [this pull request](https://github.com/GoogleContainerTools/jib/pull/4142) with an extra commit that fixes the formatting, test, and modification time set by JIB issues.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -76,11 +77,14 @@ private void add(TarArchiveEntry tarArchiveEntry) throws IOException {
...
         Path tarArchiveParentDir = Verify.verifyNotNull(namePath.getParent());
         TarArchiveEntry dir = new TarArchiveEntry(DIRECTORY_FILE, tarArchiveParentDir.toString());
-        dir.setModTime(FileEntriesLayer.DEFAULT_MODIFICATION_TIME.toEpochMilli());
         dir.setUserId(0);
         dir.setGroupId(0);
         dir.setUserName("");
         dir.setGroupName("");
+        // In this particular place we have to use default modification time (1 second past epoch)
+        // instead the modification time user might've set in JIB settings. That's because we have
+        // no way to find out the modification time user might've set in JIB settings here.
```
chanseokoh, line 86: This is not true. Jib intentionally doesn't allow setting the mod time of directories. Let's just remove the comment.

jib-core/src/test/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilderTest.java
```diff
@@ -84,12 +85,43 @@ private static void verifyNextTarArchiveEntryIsDirectory(
...
+      Path source, AbsoluteUnixPath destination, FilePermissions permissions) throws IOException {
+    return layerEntry(source, destination, permissions, "");
+  }
+
+  private static FileEntry layerEntry(Path source, AbsoluteUnixPath destination)
+      throws IOException {
+    return layerEntry(
+        source,
+        destination,
+        FileEntriesLayer.DEFAULT_FILE_PERMISSIONS_PROVIDER.get(source, destination));
```
chanseokoh, line 124: Just revert this file back to its original state. The original author misunderstood this part. The PR shouldn't touch this file, and the existing tests will verify the correctness of your changes.

chanseokoh (commented):
Also, upgrade the apache-commons-compress (#3964) to verify if it's working correctly.

izogfif:
@chanseokoh done. I had to update TarExtractorTest since it was failing now. Probably, newer version of apache-commons-compress (I used 1.26.0 instead of 1.23.0 from [#3964](https://github.com/GoogleContainerTools/jib/pull/3964)) extracts modification timetamp with milliseconds precision.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
```diff
@@ -95,6 +96,20 @@ private List<TarArchiveEntry> getSortedEntries() {
     }
   }
 
+  private static void clearTimeHeaders(TarArchiveEntry entry, Instant modTime) {
+    entry.setModTime(modTime.toEpochMilli());
+
+    String headerTime = Long.toString(modTime.getEpochSecond());
+    final long nanos = modTime.getNano();
```
chanseokoh, line 103: I'm not familiar with the PAX headers, so just asking: does it have to be the nanosecond precision? Like, what if someone decides to go with milli or microseconds? Is it possible? Not that I care as long as nano works.

chanseokoh (approved):
Normally you should update `jib-core/CHANGELOG`, `'jib-maven-plugin/CHANGELOG` and `jib-gradle-plugin/CHANGELOG` about what is fixed.

BTW, I'm not the maintainer of this repo.

jib-core/src/main/java/com/google/cloud/tools/jib/image/ReproducibleLayerBuilder.java
izogfif: Here is a post I was able to find:

> This is important for example with the tarfile module with the pax tar format. The POSIX tar [standard](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html) mandates storing the mtime in the extended header (if it is not an integer) with as much precision as is available in the underlying file system, and likewise to restore this time properly upon extraction.

So I'd say that it's better to store time with nanosecond precision. File systems that don't support this precision won't fail upon extraction: they'll simply truncate values.

izogfif:
CHANGELOG files were updated.

chanseokoh (approved):

izogfif:
Rebased on master.

izogfif:
Rebased on master (2).

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
```diff
@@ -92,16 +94,20 @@ public void testExtract_modificationTimePreserved() throws URISyntaxException, I
 
     TarExtractor.extract(source, destination);
 
+    // We have to use millisecond precision here. Taken from Instant.parse.
+    final DateTimeFormatter parser =
+        new DateTimeFormatterBuilder().parseCaseInsensitive().appendInstant(3).toFormatter();
+
```
mpeddada1, line 100: @izogfif Is there is a reason we're testing with millisecond precision here? Curious since we opted for nanosecond precision above (ReproducibleLayerBuilder)

mpeddada1:
Pasting stacktrace of failing checks below:

**kokoro-macos:**

```
Successfully started process 'Gradle Test Executor 4'
Picked up JAVA_TOOL_OPTIONS: -Djava.net.preferIPv6Addresses=true

com.google.cloud.tools.jib.registry.DockerRegistryBearerTokenTest > testDecode_invalidToken_actionsArray SKIPPED

com.google.cloud.tools.jib.tar.TarExtractorTest > testExtract_modificationTimePreserved FAILED
    value of: getLastModifiedTime(...)
    expected: 2019-08-01T16:13:09.72Z
    but was : 2019-08-01T16:13:09Z
        at com.google.cloud.tools.jib.tar.TarExtractorTest.testExtract_modificationTimePreserved(TarExtractorTest.java:102)

com.google.cloud.tools.jib.tar.TarExtractorTest > testExtract_reproducibleTimestampsEnabled FAILED
    value of: getLastModifiedTime(...)
    expected: 2021-01-29T21:10:02.78Z
    but was : 2021-01-29T21:10:02Z
        at com.google.cloud.tools.jib.tar.TarExtractorTest.testExtract_reproducibleTimestampsEnabled(TarExtractorTest.java:129)

Gradle Test Executor 4 finished executing tests.

> Task :jib-core:test FAILED
```

**kokoro-ubuntu**
```
com.google.cloud.tools.jib.api.ContainerizerIntegrationTest STANDARD_ERROR
    SLF4J: No SLF4J providers were found.
    SLF4J: Defaulting to no-operation (NOP) logger implementation
    SLF4J: See [https://www.slf4j.org/codes.html#noProviders](https://www.google.com/url?q=https://www.slf4j.org/codes.html%23noProviders&sa=D) for further details.

com.google.cloud.tools.jib.api.ReproducibleImageTest > testConfiguration FAILED
    value of: extractFromTarFileAsString(...)
    expected: ?"diff_ids":["sha256:18e4f44e6d1835bd968339b166057bd17ab7d4cbb56dc7262a5cafea7cf8d405","sha256:13369c34f073f2b9c1fa6431e23d925f1a8eac65b1726c8cc8fcc2596c69b414","sha256:4f92c507112d7880ca0f504ef8272b7fdee107263270125036a260a741565923"]}}
    but was : ?"diff_ids":["sha256:2fcc2157bf42c89195676ef6e973a96d7b018c9d30ba89db95e9e0722e1c8ef3","sha256:21f521f3217067d277af37512a08c72281d90fdd02d7174db632c8c3a34403bd","sha256:6beba018395265af5061864b7f4678e831eb2daebb1045487c641fc8b142e319"]}}
        at com.google.cloud.tools.jib.api.ReproducibleImageTest.testConfiguration(ReproducibleImageTest.java:130)

com.google.cloud.tools.jib.api.ReproducibleImageTest > testTarballStructure FAILED
    missing (3)
    #1      : c46572ef74f58d95e44dd36c1fbdfebd3752e8b56a794a13c11cfed35a1a6e1c.tar.gz
    #2      : 6d2763b0f3940d324ea6b55386429e5b173899608abf7d1bff62e25dd2e4dcea.tar.gz
    #3      : 530c1954a2b087d0b989895ea56435c9dc739a973f2d2b6cb9bb98e55bbea7ac.tar.gz

    unexpected (3)
    #1      : 98682a867906d9d07cf3c51a4fb9e08e9d5baddd1ca5dc7834f58f434c9cb15c.tar.gz
    #2      : 527db49d4e0c4159346119b4971d59016bfedceed874abab2b510ce433f6b15c.tar.gz
    #3      : 16d03883198935b4119896dcea0ea14e1bf105b6ac0a35a88820d08bc0263306.tar.gz
    ---
    expected: [c46572ef74f58d95e44dd36c1fbdfebd3752e8b56a794a13c11cfed35a1a6e1c.tar.gz, 6d2763b0f3940d324ea6b55386429e5b173899608abf7d1bff62e25dd2e4dcea.tar.gz, 530c1954a2b087d0b989895ea56435c9dc739a973f2d2b6cb9bb98e55bbea7ac.tar.gz, config.json, manifest.json]
    but was : [98682a867906d9d07cf3c51a4fb9e08e9d5baddd1ca5dc7834f58f434c9cb15c.tar.gz, 527db49d4e0c4159346119b4971d59016bfedceed874abab2b510ce433f6b15c.tar.gz, 16d03883198935b4119896dcea0ea14e1bf105b6ac0a35a88820d08bc0263306.tar.gz, config.json, manifest.json]
        at com.google.cloud.tools.jib.api.ReproducibleImageTest.testTarballStructure(ReproducibleImageTest.java:104)

com.google.cloud.tools.jib.api.ReproducibleImageTest > testManifest FAILED
    value of: extractFromTarFileAsString(...)
    expected: [{"Config":"config.json","RepoTags":["jib-core/reproducible:latest"],"Layers":["c46572ef74f58d95e44dd36c1fbdfebd3752e8b56a794a13c11cfed35a1a6e1c.tar.gz","6d2763b0f3940d324ea6b55386429e5b173899608abf7d1bff62e25dd2e4dcea.tar.gz","530c1954a2b087d0b989895ea56435c9dc739a973f2d2b6cb9bb98e55bbea7ac.tar.gz"]}]
    but was : [{"Config":"config.json","RepoTags":["jib-core/reproducible:latest"],"Layers":["98682a867906d9d07cf3c51a4fb9e08e9d5baddd1ca5dc7834f58f434c9cb15c.tar.gz","527db49d4e0c4159346119b4971d59016bfedceed874abab2b510ce433f6b15c.tar.gz","16d03883198935b4119896dcea0ea14e1bf105b6ac0a35a88820d08bc0263306.tar.gz"]}]
        at com.google.cloud.tools.jib.api.ReproducibleImageTest.testManifest(ReproducibleImageTest.java:119)

Gradle Test Executor 7 finished executing tests.

> Task :jib-core:integrationTest FAILED
```

jib-core/src/test/java/com/google/cloud/tools/jib/tar/TarExtractorTest.java
izogfif: For some reason, the code below can't extract micro- and nanoseconds, it's limited to milliseconds (at least on Windows machine with NTFS).
In one of the methods below I had to add centiseconds and avoided having to use the same DateTimeFormatter since, conveniently, truncation to milliseconds yielded "0" as the third digit, and Instant.parse can parse centiseconds.

izogfif:
> Pasting stacktrace of failing checks below:
> 
> **kokoro-macos:**
> 
> ```
> com.google.cloud.tools.jib.tar.TarExtractorTest > testExtract_modificationTimePreserved FAILED
>     value of: getLastModifiedTime(...)
>     expected: 2019-08-01T16:13:09.72Z
>     but was : 2019-08-01T16:13:09Z
>         at com.google.cloud.tools.jib.tar.TarExtractorTest.testExtract_modificationTimePreserved(TarExtractorTest.java:102)

How can I check that current OS is macOS? It seems that storing files with millisecond precision on macOS is not supported. It seems that this is what's happening:

* `TarExtractor.extract ` extracts a `.tar` file from resources into a temporary folder on test machine.
* Underlying file system truncates file modifications timestamps (they're stored with nanosecond precision inside tar file). On Windows and Ubuntu, truncation is done to millisecond precision. On macOS, apparently, to seconds.
* Tests pass on Ubuntu and Windows. They fail on macOS.

Possible solutions:
* Make macOS provide file modification timestamps with millisecond precision (like Windows and Ubuntu do).
* Read file modification timestamps with any precision, then truncate them to seconds, and then check these truncated values in tests.
*  Third way.

> **kokoro-ubuntu**

About these failures: if I update values ("expected") with new ones ("but was"), it should fix Ubuntu build. But won't this fail Windows build?

izogfif:
@mpeddada1 I finally found the [contributor guide](https://github.com/GoogleContainerTools/jib/blob/master/CONTRIBUTING.md) and was able to run tests on Ubuntu via:
```
export JIB_INTEGRATION_TESTING_LOCATION=localhost:9990
./gradlew clean goJF build integrationTest
```
but then `testWar_jetty` test fails with:
```
```
Apparently, I'm unable to make Jetty container start in Docker. Please re-run those workflows / kokoro-xxx tests on your end.

chanseokoh:
I also noticed the Jetty failure locally. Don't know why. It is the test in `jib-cli`. Since you are not changing the `jib-cli` code, for now, you can do the following which will cover your changes:
```
./gradlew jib-core:integrationTest jib-maven-plugin:integrationTest jib-gradle-plugin:integrationTest
```

mpeddada1:
> * `TarExtractor.extract ` extracts a `.tar` file from resources into a temporary folder on test machine.
> * Underlying file system truncates file modifications timestamps (they're stored with nanosecond precision inside tar file). On Windows and Ubuntu, truncation is done to millisecond precision. On macOS, apparently, to seconds.
> * Tests pass on Ubuntu and Windows. They fail on macOS.

Thanks for looking into this! It is interesting that macos only goes to seconds precision. It is possible that the job still uses the old HFS+ format drive. That being said, I think verifying seconds precision should be sufficient. 

Re `kokoro-ubuntu`: I would assume that the digest value change is due to the new PAX header logic we added so i would expect it to match in both Ubuntu and windows. But I could also be missing something here, why do we think the windows tests would fail?

izogfif:
> Re `kokoro-ubuntu`: I would assume that the digest value change is due to the new PAX header logic we added so i would expect it to match in both Ubuntu and windows. But I could also be missing something here, why do we think the windows tests would fail?

Yes, I think that PAX header changed checksums. I was expecting it to fail build in all OS (Windows, macOS, Ubuntu), but, apparently, before I fixed checksum checks, kokoro-windows tests were successful. No idea why they did that.

mpeddada1:
> Yes, I think that PAX header changed checksums. I was expecting tests to fail on all OS (Windows, macOS, Ubuntu), but, apparently, before I fixed checksum checks, kokoro-windows tests were successful. No idea why. The tests have passed this time, too.

Ah our coverage on Windows is fairly limited at the moment as we run integration tests only in ubuntu and MacOS. That's probably why we weren't able to see the failure in windows.

izogfif:
So, about this pull request, what are the next steps? Is there something I need to do in order for it to be accepted (and a new build of JIB to be released)?

mpeddada1:
No further action need, I'll go ahead and merge the PR. Thanks for your contribution @izogfif!

glasser:
Thanks for everyone who worked on this! Looking forward to seeing this get into a release!

Issue #3350, opened 2021-07-21
Disable update check by default


rmannibucau:
com.google.cloud.tools.jib.plugins.common.globalconfig.GlobalConfigTemplate#disableUpdateCheck default should be true until user explicitly accept to enable it IMHO.

Setting the field to true is sufficient to solve that issue.

meltsufin:
Many products have update checks on by default. It helps with security and getting bug fixes ASAP.
Closing because it work as intended.
If you think there is a way to offer changing default more easily, we're open to that.

rmannibucau:
Well it only handles latest which will break as soon as you need major handling at least and it creates undesired requests which is impacting on low speed networks (phone).
Also note that most *libraries* (so it concerns jib-core, maven and gradle plugins) disable it by defaults (quartz did it some years ago: https://github.com/quartz-scheduler/quartz/commit/dfe1e5a3cc248e2a46a6ea55567aaa6dc8e15ca5) because it is noise for the consumer and does not integrate in the common dependency/security check flow.
Maven and gradle will use ossndex, blackduck, .... and maven version plugin for example which is already used for security fixes for all dependencies so something specific to JIB is not needed and does not help which the build does not fail (which is not an option right?)

chanseokoh:
Appreciate your feedback. Certainly for jib-core, there's no update check implemented, as it doesn't make sense as a library. It's only for Jib build plugins and Jib CLI, and honestly, we don't consider Jib build plugins as libraries but as plugins for tools, so we think it's perfectly valid to implement an update check and at the same time having it turned on by default.

And this really helped the users upgrade fast. We continue to fix bugs, and it bothers us a lot and makes us sad to see some people running an old version of Jib that has bugs affecting their built images. We want people to build images better. Fortunately so far, we believe the upgrade checks have been working very well in that it really helped people up-to-date fast. To be honest, it is unlikely that we will disable it by default. Hope you understand.

BTW, network or build performance shouldn't be a concern. We've designed it very carefully so that it doesn't affect the main build at all. The update check runs in a completely separate and isolated thread, any kind of errors during the check will not affect the build, and it can even be that the update check is naturally terminated by JVM if the main build finishes earlier.

rmannibucau:
Well I cant say much for gradle but on maven side it is really bothering (pollutes logs, triggers wrong alerts when logs are parsed and duplicates the widely used and standard tooling without an associated handling in general - trivial example is that such tools make the build fial whereas the checker does not).

[...]

PR #4212, opened 2024-03-18
Enable to skip update checks from pom properties

rmannibucau:
Follow up of https://github.com/GoogleContainerTools/jib/issues/3350#issuecomment-885058082 since it is really bothering to not be able to disable update check for project in time - when you don't work on a single "everyday" project and disabling it globally is rarely the desired behavior (it is really a per project thing) but the properties were not read as expected so this PR ensures it is the case.

Thank you for your interest in contributing! For general guidelines, please refer to
the [contributing guide](https://github.com/GoogleContainerTools/jib/blob/master/CONTRIBUTING.md).

Please follow the guidelines below before opening an issue or a PR:
- [ ] Ensure the issue was not already reported. 
- [ ] Create a new issue at https://github.com/GoogleContainerTools/jib/issues/new/choose if you are unable to find an existing issue addressing your problem. Make sure to include a title and clear description, as much relevant information as possible, and a code sample or an executable test case demonstrating the expected behavior that is not occurring.
- [ ] Discuss the priority and potential solutions with the maintainers in the issue. The maintainers would review the issue and add a label "Accepting Contributions" once the issue is ready for accepting contributions.
- [ ] Open a PR only if the issue is labeled with "Accepting Contributions", ensure the PR description clearly describes the problem and solution. Note that an open PR without an issues labeled with "Accepting Contributions" will not be accepted.
- [ ] Verify that integration tests and unit tests are passing after the change.
- [ ] Address all checkstyle issues. Refer to the [style guide](https://github.com/GoogleContainerTools/jib/blob/master/STYLE_GUIDE.md).

Fixes #<issue_number_goes_here> 🛠️

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
@@ -455,6 +455,12 @@ public class GradleProjectProperties implements ProjectProperties {
   }
 
   @Override
+  public boolean isDisableUpdateCheck() {
+    return Boolean.parseBoolean(
+        String.valueOf(project.findProperty(PropertyNames.DISABLE_UPDATE_CHECKS)));
+  }
+
+  @Override
   public JibContainerBuilder runPluginExtensions(
       List<? extends ExtensionConfiguration> extensionConfigs,
       JibContainerBuilder jibContainerBuilder)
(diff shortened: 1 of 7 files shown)


slawekjaranowski:
even more there is the problem to settings environment variable with dots like `jib.disableUpdateChecks` on unix systems 

I have tested with zsh and bash on macOs ...

zsh
```
export jib.disableUpdateChecks=true 
export: not valid in this context: jib.disableUpdateChecks
```

bash
```
export jib.disableUpdateChecks=true
bash: export: `jib.disableUpdateChecks=true': not a valid identifier
```

so I can not disable it at easy way

rmannibucau:
goal was to do it per project in the pom or globally either in settings or MAVEN_OPTS so think it works but not the way you think @slawekjaranowski

PR #4216, opened 2024-03-22
fix: modify default entrypoint for WAR containerization to be compatible with Jetty 12+

### LINKED ISSUE #4211: Jib Ubuntu and MacOS Continuous Builds are failing
As observed by @zhumin8, the jib-ubuntu and jib-macos continious tests have been consistently failing. 

The failure appears to originate from the jib-cli integration tests, particularly:

```
com.google.cloud.tools.jib.cli.WarCommandTest > testWar_jetty FAILED
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at com.google.cloud.tools.jib.api.HttpRequestTester.verifyBody(HttpRequestTester.java:35)
        at com.google.cloud.tools.jib.cli.WarCommandTest.testWar_jetty(WarCommandTest.java:99)
```

[Test Fusion Link](https://fusion2.corp.google.com/ci/kokoro/prod:jib%2Fgcp_ubuntu%2Fcontinuous/activity/092c0fd9-f39d-4569-9651-e7a711dbc78d/summary?q=%20jib%2Fgcp_ubuntu%2Fcontinuous)

mpeddada1:
Fixes #4211 

Verified locally with `./gradlew integrationTest`


jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -77,12 +77,24 @@ private static List<String> computeEntrypoint(
       CommonContainerConfigCliOptions commonContainerConfigCliOptions)
       throws InvalidImageReferenceException {
     List<String> entrypoint = commonContainerConfigCliOptions.getEntrypoint();
+    String baseImage = fetchBaseImage(commonContainerConfigCliOptions);
     if (!entrypoint.isEmpty()) {
       return entrypoint;
     }
     if (commonContainerConfigCliOptions.isJettyBaseimage()) {
+      // If using jetty 12 or later, then specify deploy module. See
```
meltsufin, line 85: Where are we detecting that it's version 12 of Jetty?

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -77,12 +77,24 @@ private static List<String> computeEntrypoint(
       CommonContainerConfigCliOptions commonContainerConfigCliOptions)
       throws InvalidImageReferenceException {
     List<String> entrypoint = commonContainerConfigCliOptions.getEntrypoint();
+    String baseImage = fetchBaseImage(commonContainerConfigCliOptions);
     if (!entrypoint.isEmpty()) {
       return entrypoint;
     }
     if (commonContainerConfigCliOptions.isJettyBaseimage()) {
```
meltsufin, line 84: What if the `baseImage` is jetty but due to the default rather than being specified by CLI options?

mpeddada1: The isJettyBaseImage() method contains some logic to check if the user-specified base image is jetty. If the base  image is not customized then we assume that the default (`jetty:latest`) will be used and return `true`. Reference: https://github.com/GoogleContainerTools/jib/blob/6df635ee51bc466c961cdeea23d84e84e87ba4b0/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/CommonContainerConfigCliOptions.java#L189-L197

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: That's a good question. The current default in Jib is `jetty:latest` and this logic makes the assumption that the [latest docker image](https://hub.docker.com/layers/library/jetty/latest/images/sha256-eceaca283f0cd3a51741801cccecc38b279cf8153ad940b332f0d3ec2b7f9c8c?context=explore) will be v12 or greater now (as seen by the `ENV JETTY_VERSION=12.0.7` in the image's layer) now that jetty 10 and jetty 11 have reached [EOL](https://github.com/jetty/jetty.project/issues/10485). 

The key logic for determining the `tag` lies in [ContainerBuilders#create](https://github.com/GoogleContainerTools/jib/blob/6df635ee51bc466c961cdeea23d84e84e87ba4b0/jib-cli/src/main/java/com/google/cloud/tools/jib/cli/ContainerBuilders.java#L69) which is called by WarFiles#toJibContainerBuilder. This class then calls `ImageReference.parse(baseImage)` which parses the registry, repository and tag from baseImage. If a tag isn't specified (which is the case here) then `latest` is set as the default.

meltsufin: In that case, it might be worthlwhile to clarify the comment. For instance, `// Since we're using Jetty 12 or later...`.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
```diff
@@ -77,12 +77,24 @@ private static List<String> computeEntrypoint(
       CommonContainerConfigCliOptions commonContainerConfigCliOptions)
       throws InvalidImageReferenceException {
     List<String> entrypoint = commonContainerConfigCliOptions.getEntrypoint();
+    String baseImage = fetchBaseImage(commonContainerConfigCliOptions);
```
meltsufin, line 80: Can this line move to inside the if statement on line 84?

meltsufin (approved):

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Good idea, done.

jib-cli/src/main/java/com/google/cloud/tools/jib/cli/war/WarFiles.java
mpeddada1: Thanks! Removed this line as it is no longer needed after [b893521](https://github.com/GoogleContainerTools/jib/pull/4216/commits/b893521a465b91b2cb73d8bb44e8d9454d3c465e).

PR #4249, opened 2024-05-15
feat: determine platform details from local docker installation for jibDockerBuild

mpeddada1:
This code has been tested manually  with `jib/examples/helloworld/`. 

**Notes:**

- If manifest list found then the log picks the image that contains the same os and architecture as the local docker environment (derived through `docker info`). Otherwise, it defaults to publishing the image with the osType and architecture matching the ones in the first <platform></platform> block. 
For example, if the platform configurations are as follows then the image with os of `linux` and architecture of `arm64` will be pushed to the docker engine:

```
<platforms>
                <platform>
                  <architecture>arm64</architecture>
                  <os>linux</os>
                </platform>
                <platform>
                  <architecture>amd64</architecture>
                  <os>linux</os>
                </platform>
 </platforms>
```


jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -647,4 +649,28 @@ private void writeTarFile(
...
+    String osType = dockerInfoDetails.getOsType();
+    String dockerArchitecture = computeArchitecture(dockerInfoDetails.getArchitecture());
+    for (Map.Entry<Image, Future<Image>> imageEntry :
+        results.baseImagesAndBuiltImages.get().entrySet()) {
+      Image image = imageEntry.getValue().get();
+      if (image.getArchitecture().equals(dockerArchitecture) && image.getOs().equals(osType)) {
+        return image;
+      }
+    }
+    return results.baseImagesAndBuiltImages.get().values().iterator().next().get();
```
blakeli0, line 674: Shall we error out if no matching found?

mpeddada1: Yup, good idea, done.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerClient.java
```diff
@@ -69,4 +69,15 @@ void save(ImageReference imageReference, Path outputPath, Consumer<Long> written
    * @throws InterruptedException if the {@code docker inspect} process was interrupted
    */
   ImageDetails inspect(ImageReference imageReference) throws IOException, InterruptedException;
+
+  /**
+   * Gets the os and architecture of the local docker installation.
```
blakeli0, line 74: This could be just docker info details, `os and architecture` are implementation details of `DockerInfoDetails`, and we could add more attributes later.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -647,4 +656,29 @@ private void writeTarFile(
...
+      return "arm64";
+    }
+    return architecture;
+  }
+
+  private Optional<Image> fetchBuiltImageForLocalBuild(String osType, String architecture)
+      throws InterruptedException, ExecutionException {
+    if (results.baseImagesAndBuiltImages.get().size() > 1) {
+      LOGGER.warning(
+          "Detected multi-platform configuration, only building the one that matches the local Docker Engine's os and architecture");
```
blakeli0, line 673: Maybe add `osType` and `architecture` to the log?

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerInfoDetails.java
```diff
@@ -0,0 +1,40 @@
...
+
+package com.google.cloud.tools.jib.api;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.cloud.tools.jib.json.JsonTemplate;
+
+/** Contains the os and architecture from {@code docker info}. */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DockerInfoDetails implements JsonTemplate {
```
blakeli0, line 25: Add unit tests for different scenarios, e.g. `OSType` does not exist etc.

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-multiplatform-build.xml
blakeli0, line 1: The test was passing for this pom that contains multi-platfom?

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-multi-platform-invalid.gradle
blakeli0, line 1: I don't think we need an integration test for invalid scenarios, it should be covered by unit tests.

jib-core/src/main/java/com/google/cloud/tools/jib/api/DockerInfoDetails.java
mpeddada1: Good idea, added unit testing for when OSType and architecture don't exist.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -647,4 +656,29 @@ private void writeTarFile(
   private <E> List<Future<E>> scheduleCallables(ImmutableList<? extends Callable<E>> callables) {
     return callables.stream().map(executorService::submit).collect(Collectors.toList());
   }
+
+  private String computeArchitecture(String architecture) {
+    if (architecture.equals("x86_64")) {
```
blakeli0, line 661: Where do we get the mapping between the architecture from docker info and the architecture needed for docker build? Are `x86_64 ` and `aarch64 ` all we need? Also please add unit tests for all the cases.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
```diff
@@ -647,4 +656,29 @@ private void writeTarFile(
...
+    return architecture;
+  }
+
+  private Optional<Image> fetchBuiltImageForLocalBuild(String osType, String architecture)
+      throws InterruptedException, ExecutionException {
+    if (results.baseImagesAndBuiltImages.get().size() > 1) {
+      LOGGER.warning(
+          "Detected multi-platform configuration, only building the one that matches the local Docker Engine's os and architecture");
+    }
+    for (Map.Entry<Image, Future<Image>> imageEntry :
```
blakeli0, line 675: I see that they are already tested in `JibIntegrationTest`, but ideally I think the different cases should be tested in `StepsRunnerTest` since we don't have to run docker engine. I'll leave it to you to decide if the refactoring it needed.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
mpeddada1: Done, added os and arch info to log statement.

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-multiplatform-build.xml
mpeddada1: Reverted the change to this pom in favor of more unit tests in StepsRunnerTest and integration testing in jib-core.

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-multi-platform-invalid.gradle
mpeddada1: Reverted the change to this gradle file in favor of more coverage through unit testing in StepsRunnerTest and integration testing in jib-core.

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
blakeli0: I see, so the root cause is that the architecture returned from `docker info` is different from the architecture used by `docker build`. If that's the case, maybe we can change the method name to something like `mapDockerInfoArchitechtureToDockerBuildArchitechture`? Or `toDockerBuildArchitecture(String dockerInfoArchitecture)` if we think the previous one is too verbose?

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-multiplatform-build.xml
blakeli0: I think it still makes sense to have a happy path end-to-end test, so that it is easier for us to reproduce any multi-platform issues in the future. Specifically for this file though, I see that it exists for a few years, and why the [test using this file](https://github.com/GoogleContainerTools/jib/blame/8b13f0d1bd4d88d7b3503b570ae065e322daaf47/jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java#L715) was passing?

jib-gradle-plugin/src/integration-test/resources/gradle/projects/simple/build-multi-platform-invalid.gradle
blakeli0: Thanks!

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-multiplatform-build.xml
mpeddada1: That makes sense. I'll bring back the happy path.



> Specifically for this file though, I see that it exists for a few years, and why the [test using this file](https://github.com/GoogleContainerTools/jib/blame/8b13f0d1bd4d88d7b3503b570ae065e322daaf47/jib-maven-plugin/src/integration-test/java/com/google/cloud/tools/jib/maven/BuildImageMojoIntegrationTest.java#L715) was passing?

This test is expected to pass because it is verifying multi-platform registry pushes through the `mvn compile jib:build` command. While we don't have multi-platform support for builds to the local docker daemon, pushing to the registry is supported. It verifies the contents of the manifest list as well as the two platform specific images pushed. The pom defines `busybox` as the base image which contains support for `amd64/linux` and `arm64/linux` which matches the platforms specified under the <platforms> configuration block. This can be verified through `docker manifest inspect busybox@sha256:4f47c01fa91355af2865ac10fef5bf6ec9c7f42ad2321377c21e844427972977` which shows:

```
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 527,
         "digest": "sha256:400ee2ed939df769d4681023810d2e4fb9479b8401d97003c710d0e20f7c49c6",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 527,
         "digest": "sha256:00466b7923ac6cb7bdfd211101c09917e0d416212ca058740954d7adb0aac4f6",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v5"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 527,
         "digest": "sha256:8fcef70caa67710d8cd568d611b1b23a61886ce5ffab267c5ffa34136999b023",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v6"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 527,
         "digest": "sha256:6f84685399b2b834d1bc230d2b7f47bed1169b058a338f1b4d05997bb2293a2d",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v7"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 527,
         "digest": "sha256:759af0b171cc3742032232b5562c1a6d5deb29e487efa2d0e425b2df3a8d5521",
         "platform": {
            "architecture": "arm64",
            "os": "linux",
            "variant": "v8"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 527,
         "digest": "sha256:ab680d33f94b93e4c3383faad99a9d0195a73d50d4bb039acea5ae7f0ae6691d",
         "platform": {
            "architecture": "386",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 527,
         "digest": "sha256:3d96083d0e7c8fee34cd7207880cf838f404d14349cb5e2f78085e05441e8576",
         "platform": {
            "architecture": "mips64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 528,
         "digest": "sha256:ad1ceba3580dfde5e97fd3dc0fdb56b508ab4cb5ea1ff1dd688982c4eff20357",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 528,
         "digest": "sha256:ab8e9b6566a776d442e6d20415a5a73124e2f5bd20dd179fb11ca079de1c13d3",
         "platform": {
            "architecture": "s390x",
            "os": "linux"
         }
      }
   ]
}

```
Let me know if you're looking for any additional details!

jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/StepsRunner.java
mpeddada1: That's right, the naming doesn't seem to be standardized everywhere. While `docker info` results in `x86_64` on my machine, `docker manifest inspect` on images uses `amd64` which is supposed to be equivalent (https://github.com/docker/docs/issues/4295). 
Renaming the method is a good idea. Alternatively, what are your thoughts on `normalizeArchitecture()` or `standardizeArchitecture()`? I found a couple of references that do something similar: 
(1)https://github.com/containerd/platforms/blob/c1438e911ac7596426105350652fe267d0fb8a03/database.go#L76
 (2)https://github.com/openzipkin/zipkin/blob/2f0a26fd4e3f3ea24e9fb30f55a84ebbc194129c/build-bin/docker/docker_arch#L7

blakeli0: Thanks for the suggestions! I like `normalizeArchitecture()`.

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-multiplatform-build.xml
blakeli0: Got it, thanks! I guess we were pushing images to GCR?

blakeli0 (approved):

jib-maven-plugin/src/test/resources/maven/projects/simple/pom-multiplatform-build.xml
mpeddada1: I think in this case (and many other integration tests) is verifying registry pushes against a local registry hosted on localhost:5000 ([reference](https://github.com/GoogleContainerTools/jib/blob/8b13f0d1bd4d88d7b3503b570ae065e322daaf47/CONTRIBUTING.md#integration-tests))

profhenry:
I think this feature does not work with podman, did you test or consider this?

When using podman the `docker info` command differs :-/. It returns
```json
{
  "host": {
    "arch": "amd64",
    "buildahVersion": "1.23.0",
    "cgroupManager": "systemd",
    "cgroupVersion": "v2",
    "cgroupControllers": [],

```
Please also see: https://docs.podman.io/en/latest/markdown/podman-info.1.html

So determining os and architecture fails and you get `"The configured platforms don't match the Docker Engine's OS and architecture (/)`. It would be nice if we could support podman as well.

profhenry:
In addition this feature **is** also enabled when having just **one single** platform specified.

Even when specifying
```json
<platform>
  <architecture>amd64</architecture>
  <os>linux</os>
</platform>
```
there is no need for this new feature. But it kicks in and fails the build. How can i disable this feature?
This leaves Jib is no longer usable with podman :-/

chanseokoh:
@profhenry #4265

Issue #4267, opened 2024-06-06
No progress after "Container entrypoint set to" with jib-maven-plugin 3.4.3


LeovR:
**Environment**:

- *Jib version:* 3.4.3 
- *Build tool:* Maven 3.9.6
- *OS:* Windows 10

**Description of the issue**:
After upgrading to jib-maven-plugin 3.4.3 the progress stops after "Container entrypoint set to". With jib-maven-plugin 3.4.2 the build succeeds.

**Expected behavior**:
The build succeeds.

**Steps to reproduce**:

  1. skaffold run

**`jib-maven-plugin` Configuration**: 
```xml
            <plugin>
                <groupId>com.google.cloud.tools</groupId>
                <artifactId>jib-maven-plugin</artifactId>
                <version>3.4.3</version>
                <configuration>
                    <skip>false</skip>
                    <containerizingMode>packaged</containerizingMode>
                    <container>
                        <creationTime>USE_CURRENT_TIMESTAMP</creationTime>
                        <mainClass>${start-class}</mainClass>
                    </container>
                    <from>
                        <image>eclipse-temurin:21</image>
                    </from>
                    <to>
                        <image>${docker-registry}/application</image>
                        <tags>
                            <tag>${project.version}</tag>
                            <tag>latest</tag>
                            <tag>${env.CI_COMMIT_SHA}</tag>
                        </tags>
                    </to>
                    <allowInsecureRegistries>true</allowInsecureRegistries>
                </configuration>
                <executions>
                    <execution>
                        <phase>deploy</phase>
                        <goals>
                            <goal>build</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

**Log output**: 

```
[DEBUG] TIMING	Pulling base image manifest
[INFO] Getting manifest for base image eclipse-temurin:21...
[DEBUG] TIMING	trying mirrors
[DEBUG] TIMED	trying mirrors : 0.0 ms
[DEBUG] TIMING	Preparing application layer builders
[DEBUG] TIMED	Preparing application layer builders : 12.164 ms
[INFO] Building dependencies layer...
[INFO] Building project dependencies layer...
[INFO] Building extra files layer...
[DEBUG] TIMING	Building dependencies layer
[DEBUG] TIMING	Building project dependencies layer
[INFO] Building jvm arg files layer...
[DEBUG] TIMING	Building extra files layer
[DEBUG] TIMING	Building jvm arg files layer
[...]

PR #4268, opened 2024-06-06
fix: allow pushing images with different arch/os to docker daemon

### LINKED ISSUE #4265: Impossible to build amd-64 docker on Mac M3 machine
Starting 3.4.3 version of `jib-maven-plugin` it is not possible to build amd-64 image on Mac with Mx chip (arm-64)
I didn't find a setting, that fix it. There is no problem with 3.4.2 version.

mpeddada1:
Fixes #4265 and #4267
- Avoid throwing exception when building image with mismatched arch/os to the local docker daemon.
- Follow convention of using event handlers to log warning when building for matching os/arch. Example: https://github.com/GoogleContainerTools/jib/blob/b06b001ed26b6a1fedc1ac738f628513390433e9/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/PullBaseImageStep.java#L225

### DIFF
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
@@ -356,7 +356,7 @@ public class JibIntegrationTest {
     System.out.println("testBasic_jibImageToDockerDaemon_arm64");
 
     // Use arm64v8/busybox as base image.
-    String toImage = "docker-daemon-mismatched-arch";
+    String toImage = dockerHost + ":5000/docker-daemon-mismatched-arch";
     Jib.from(
             RegistryImage.named(
                 "busybox@sha256:eb427d855f82782c110b48b9a398556c629ce4951ae252c6f6751a136e194668"))
--- a/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
+++ b/jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
@@ -321,7 +321,7 @@ public class JibIntegrationTest {
   }
 
   @Test
-  public void testBasic_jibImageToDockerDaemon_arm64()
+  public void testBasic_jibImageToDockerDaemon_arm64BaseImage()
       throws IOException, InterruptedException, InvalidImageReferenceException, ExecutionException,
           RegistryException, CacheDirectoryCreationException {
     // Use arm64v8/busybox as base image.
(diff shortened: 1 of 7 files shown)


jib-core/src/integration-test/java/com/google/cloud/tools/jib/api/JibIntegrationTest.java
```diff
@@ -343,27 +342,38 @@ public void testBasicMultiPlatform_toDockerDaemon()
...
-                    .containerize(
-                        Containerizer.to(
-                                DockerDaemonImage.named(
-                                    dockerHost + ":5000/docker-daemon-multi-platform"))
-                            .setAllowInsecureRegistries(true)));
-    assertThat(exception)
-        .hasCauseThat()
-        .hasMessageThat()
-        .startsWith("The configured platforms don't match the Docker Engine's OS and architecture");
+  public void testBasicMultiPlatform_toDockerDaemon_pickFirstPlatformWhenNoMatchingImage()
```
mpeddada1, line 345: TODO(@mpeddada1): Add test for single image build with different os/arch.

mpeddada1: Done in 9ff90bd4cf8aea53db7bfea31f7aa265550f5cd0

blakeli0 (approved):

mpeddada1:
Notes on `kokoro-ubuntu` timeout:
- The job passed on https://github.com/GoogleContainerTools/jib/pull/4268/commits/f2d886b289a2e0d471f37281ce23322e00e6a288 but started failing after test for arm64 was added.

chanseokoh:
@mpeddada1 just so you know, it sounds like #4267 is another issue introduced by #4249 in 3.4.3.

mpeddada1:
> @mpeddada1 just so you know, it sounds like #4267 is another issue introduced by #4249 in 3.4.3.

Thank you @chanseokoh! Given that it is not reproducible in 3.4.2, the change in logging behavior may connected to some of the modifications made in https://github.com/GoogleContainerTools/jib/pull/4249. Will have to take a deeper look to understand what causes the difference in behavior for windows (perhaps in a separate dedicated PR).

mpeddada1:
Regarding the integration test behavior on the CI:
- Changing test names is causing different behavior on the ubuntu CI. Documenting test names to better understand execution order: https://github.com/GoogleContainerTools/jib/pull/4268/commits/d92123d9675a3ea662c8c1a77e757bf003ffe0c6. According to https://github.com/junit-team/junit4/wiki/Test-execution-order#test-execution-order, Junit 4 uses deterministic but not predictable order.

mpeddada1:
[Update on CI] Splitting the test methods up into separate test classes no longer causes the builds to be stuck. Ready for another round of review.

blakeli0 (approved):

calebAtIspot:
should https://github.com/GoogleContainerTools/jib/blob/master/docs/faq.md#how-do-i-specify-a-platform-in-the-manifest-list-or-oci-index-of-a-base-image be updated? It still says "Does not support pushing to a Docker daemon"

PR #4298, opened 2024-08-22
fix: Podman image inspect digest for local images

### LINKED ISSUE #4134: Slightly different json output causes dockerBuild with podman to fail with "Invalid digest" error for docker://<image> URIs
**Environment**:

- *Jib version:* 3.4.0
- *Build tool:* Maven 3.9.4
- *OS:* Linux

**Description of the issue**:
When using a URI like `docker://ubuntu:latest` for the base image, the build fails with the following exception when using podman (v4.7.1) as `jib.dockerClient.executable`:

```
Caused by: java.security.DigestException: Invalid digest: e03c05f86eda698d40d66ca2c8542ebf597ebc5254cee070dde6c407a132a385
    at com.google.cloud.tools.jib.api.DescriptorDigest.fromDigest (DescriptorDigest.java:74)
    at com.google.cloud.tools.jib.docker.CliDockerClient$DockerImageDetails.getImageId (CliDockerClient.java:83)
    at com.google.cloud.tools.jib.builder.steps.LocalBaseImageSteps.getCachedDockerImage (LocalBaseImageSteps.java:183)
    at com.google.cloud.tools.jib.builder.steps.LocalBaseImageSteps.lambda$retrieveDockerDaemonLayersStep$0 (LocalBaseImageSteps.java:114)
 ```

The problem seems to be that in https://github.com/GoogleContainerTools/jib/blob/440bd4897c6c427cc9a491128aa76c72378d605b/jib-core/src/main/java/com/google/cloud/tools/jib/docker/CliDockerClient.java#L70C17-L70C17 the value of the `"Id"` property is used as the "digest". The json output for "inspect" differs slightly between podman and docker:

`podman inspect -f "{{json .}}" --type image ubuntu:latest`:
```
{"Id":"e4c58958181a5925816faa528ce959e487632f4cfd192f8132f71b32df2744b4","Digest":"sha256:2b7412e6465c3c7fc5bb21d3e6f1917c167358449fecac8176c6e496e5c1f05f", ...
```

`docker inspect -f "{{json .}}" --type image ubuntu:latest`
```
{"Id":"sha256:e4c58958181a5925816faa528ce959e487632f4cfd192f8132f71b32df2744b4",
```

So it seems that in the output of podman we would have to use the value of the "Digest" property instead.

Would you consider adjusting the logic here a bit so that building images from the local cache with podman can also work?

mihalyr:
When using `docker://` prefix to use a base image from the local docker daemon, Jib expects the `{{.Id}}` field from the `docker image inspect` output to contain a valid image digest prefixed with `sha256:`. However, with Podman this is not the case and such builds fail with an "Invalid digest" exception, because Podman returns only the 64-char hash value in the `{{.Id}}` field without the `sha256:` prefix.

`CliDockerClient` already has the functionality to deal with the hashes without the `sha256:` prefix. A new method, `fromDigestOrgHash` has beend added to `DescriptorDigest` which first checks for the digest prefix and then checks for the hash. This allows full backwards compatibility and accepts the Podman version of the image Ids.

Thank you for your interest in contributing! For general guidelines, please refer to
the [contributing guide](https://github.com/GoogleContainerTools/jib/blob/master/CONTRIBUTING.md).

Fixes #4134 🛠️


bdellegrazie:
@mihalyr this looks great!

Maintainers, can we please get a review so it can be merged?
The bug solved here effectively stops people from using Jib with Podman right now.

@ldetmer is there any chance you could review or assign it to the appropriate person to be reviewed please?

maroony:
In my company, we also need this fix. For security reasons, we have to use podman instead of docker.

geoand:
Hey folks, I just ran into this issue as well in Quarkus. Is there any chance we can get a fix?

PR #3125, opened 2021-03-09
Add Integration testing for new jib-gradle-plugin option `configurationName`

mpeddada1:
For #3034

PR #4396, opened 2025-07-01
Configurable Gradle SourceSet

### LINKED ISSUE #4316: Gradle: Allow setting custom sourceset
**Environment**:

- *Jib version:* 3.4.3
- *Build tool:* Gradle 8.7

**Description of the issue**:
Currently, the main sourceset is hardcoded into the gradle plugin. With the default Java plugin this works well, however running anything else makes using Jib difficult.
My use-case is Kotlin Multiplatform - it does generate a JVM target, however it does not create a `main` sourceset. This results in Jib throwing an exception:

```
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':server'.
Caused by: org.gradle.api.UnknownDomainObjectException: SourceSet with name 'main' not found.
	at com.google.cloud.tools.jib.gradle.JibPlugin.lambda$apply$9(JibPlugin.java:189)
```

Having to manually set up the main sourceset as a "mirror" is error-prone and cumbersome. A very simple choice of sourceset in the Jib configuration would solve most issues.

This exact issue was already mentioned in https://github.com/GoogleContainerTools/jib/issues/1778, and even though the issue was closed, only the configuration was allowed to be configurable. The sourceset still isn't.

nathanfallet:
Fixes #4316

This introduces a new configuration option `sourceSetName` that just sets the name of the source set that should be used.
This is required to make the Gradle Jib plugin compatible with the Kotlin Multiplatform Plugin.
See https://youtrack.jetbrains.com/issue/KTOR-8464 for more about it.

The PR is inspired by #3034 which is a very similar change.
If this looks good, I can add integration tests the same way #3125 did.

### DIFF
--- a/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
+++ b/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
@@ -94,9 +94,6 @@ public class GradleProjectProperties implements ProjectProperties {
   /** Used for logging during main class inference. */
   private static final String JAR_PLUGIN_NAME = "'jar' task";
 
-  /** Name of the `main` {@link SourceSet} to use as source files. */
-  private static final String MAIN_SOURCE_SET_NAME = "main";
-
   private static final Duration LOGGING_THREAD_SHUTDOWN_TIMEOUT = Duration.ofSeconds(1);
 
   /**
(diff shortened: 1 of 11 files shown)


jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/GradleProjectProperties.java
```diff
@@ -409,14 +411,17 @@ public boolean isWarProject() {
...
+      String configurationName,
+      String sourceSetName) {
     List<FileCollection> dependencyFileCollections = new ArrayList<>();
     dependencyFileCollections.add(project.getConfigurations().getByName(configurationName));
     // Output directories (classes and resources) from main SourceSet are added
     // so that BuildTarTask picks up changes in these and do not skip task
     SourceSetContainer sourceSetContainer =
         project.getExtensions().getByType(SourceSetContainer.class);
-    SourceSet mainSourceSet = sourceSetContainer.getByName(MAIN_SOURCE_SET_NAME);
+    SourceSet mainSourceSet = sourceSetContainer.getByName(sourceSetName);
```
LiYing2010, line 424: I don't think this code can work

JIB code search source set, and expect a `org.gradle.api.tasks.SourceSet` type

but in Kotlin Multiplatform Project, the source sets are type `org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet`

these 2 types are totally different

and `KotlinSourceSet` CAN NOT be used as a `SourceSet`

in the implementation of `withJava()` function, it actually created a "main" `SourceSet`, and does lot of configuration for it
and then, JIB use this "main" `SourceSet` instead of "jvmMain" `KotlinSourceSet`

I made a workaround here:
https://github.com/GoogleContainerTools/jib/issues/4316#issuecomment-4808284830

which does same thing with `withJava()` function

please take a look

PR #4466, opened 2025-12-07
Fix version comparison logic in update checker

Gautam-aman:
### Summary
This PR fixes the issue where the update check incorrectly reports an older version as a newer one. For example, it reported `3.4.6` as newer even when the currently used version was `3.5.0`.

### Root Cause
Version strings were compared lexicographically instead of numerically (e.g., `"3.10"` < `"3.9"` when using string comparison).

### Fix
- Implemented numeric comparison for version segments.
- Treated shorter version strings (`3.5`) as equivalent to `3.5.0`.
- Added comprehensive unit test `testCompareVersions()` to prevent regressions.

### Tests
All tests pass:
./gradlew :jib-plugins-common:test
./gradlew :jib-maven-plugin:test

This resolves the incorrect user warning which caused confusion as 3.4.6 is older than 3.5.0.

### Notes
No behavior changes other than correcting version comparison.

### DIFF
--- a/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
+++ b/jib-plugins-common/src/main/java/com/google/cloud/tools/jib/plugins/common/UpdateChecker.java
@@ -63,30 +63,53 @@ public class UpdateChecker {
    */
 
   @VisibleForTesting
-  static int compareVersions(String v1, String v2) {
+   static int compareVersions(String v1, String v2) {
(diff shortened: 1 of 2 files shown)


Gautam-aman:
Thanks for the helpful review! 🙌
I've updated the PR to:
- Handle prerelease suffixes such as -SNAPSHOT, -alpha, -beta
- Add test coverage for these cases
- Switch to a manual numeric-prefix parser to avoid regex overhead

Please let me know if you'd like any further improvements.

PR #4475, opened 2026-01-28
Update gradle to 8.14.4

### LINKED ISSUE #4474: Update gradle to 8.14.4
Gradle 6.9.2 is 5 years old (December 2021) and is preventing this project from using Java versions greater than 15, [gradle java runtime compatibility matrix](https://docs.gradle.org/current/userguide/compatibility.html).  The project is currently targeting java 8 as the source and target compatibility ([source code](https://github.com/GoogleContainerTools/jib/blob/master/build.gradle#L92-L93)), so how about updating to gradle 8.14.4 the latest version of gradle that supports Java 8.

Gave it a first go and found some issues below to discuss:
* overlapping jar and bootJar output addressed by spring boot 2.5.0+ (support ended for 2.5.0 2023-08) - logic exists to address this issue in [jib-gradle-plugin](https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/main/java/com/google/cloud/tools/jib/gradle/JibPlugin.java#L160-L181) and [unit tests are failing](https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/src/test/java/com/google/cloud/tools/jib/gradle/JibPluginTest.java#L250-L268).  Should the logic be removed as well as the unit tests?  The logic could be left in and the unit tests updated to keep support for legacy spring boot users.
* [feature registration on main has been deprecated](https://docs.gradle.org/8.14.3/userguide/upgrading_version_8.html#deprecate_register_feature_main_source_set) - Is zstd support still desired.  A new source set can be created in [jib-core](https://github.com/GoogleContainerTools/jib/blob/master/jib-core/build.gradle#L10-L11), but it is unnecessary if the support is not still desired.

Welcome any thoughts.

mluckam:
Fixes #4474  🛠️


austek:
@diegomarquezp / @mpeddada1 — would one of you be able to take a look at this when you have a chance?

Happy to rebase or split this up if that makes review easier. Thanks!

PR #4479, opened 2026-02-02
Gradle9 upgrade

ldetmer:
this is effort to actually upgrade to gradle9 compatability

### DIFF
--- a/jib-gradle-plugin/build.gradle
+++ b/jib-gradle-plugin/build.gradle
@@ -45,7 +45,7 @@ dependencies {
   integrationTestImplementation dependencyStrings.JBCRYPT
 
   // only for testing a concrete Spring Boot example in a test (not for test infrastructure)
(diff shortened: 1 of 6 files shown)


mluckam:
Also made an effort to update gradle.  Chose 8.14.4 to allow for building with Java 1.8 since the source and target compatibility are still 1.8, see [source](https://github.com/GoogleContainerTools/jib/blob/master/build.gradle#L92C3-L93).  Not required to have the same version of java building as are the source and target, but wanted to keep that option available.

PR is #4475.  Using #4475 as a base I did the following:
* upgrade to gradle 9.3.1
* upgrade dependencies to latest
    * `error_prone_core`
    * `nullway`
* upgrade plugin to the latest
    * `net.ltgt.errorprone`

The result was many static analysis findings, as to be expected when updating 5 year old dependencies.  Welcome your thoughts on which version of gradle should be targeted.

PR #4507, opened 2026-05-23
Fix OverlappingFileLockException in Gradle composite builds

### LINKED ISSUE #3347: jibDockerBuild build randomly fails in multi-project setup with parrallel flag
**Environment:**

*Jib version: 3.1.2
*Build tool: Gradle v7.0.2
*OS: Ubuntu-18.04 
*JDK 14,
*Java source/target compatibility 12

**Description of the issue:**

In a multi-project gradle project with the flag org.gradle.parallel=true, the execution of the top level jibDockerBuild target randomly fails for one of the projects.
This seems to be due to the parrallel nature of the build.

**Steps to reproduce:**

1. Multi-project with at least 7 services built as docker containers
2. gradle.properties set with parrallel execution
org.gradle.parallel=true
3. build target jibDockerBuild or the root project
4. Extract of the build log
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':project4:jibDockerBuild'.
> com.google.cloud.tools.jib.plugins.common.BuildStepsExecutionException: (null exception message)

**Expected behavior:**
Images for all projects created

**Mitigation:**
For each project enabling jib plugin:

// Make sure only one image is created at a time
// as it may cause to overload the socket and fail the build.
// This is done by identifying a common output file
jibDockerBuild.configure {
	outputs.file "$rootDir/jib.lock"
}

hczedik:
In Gradle composite builds where multiple included builds have independent buildSrc directories, each included build loads jib-core in its own classloader.
The static lockMap in LockFile becomes multiple separate instances (one per classloader) so the synchronization that prevents concurrent FileChannel.lock() calls on the same file doesn't work, causing OverlappingFileLockException.

My fix stores the shared lock map in System.getProperties(), which returns the same Hashtable instance regardless of which classloader calls it. This is admittedly a hack (there's no clean JVM API for cross-classloader shared state), but it's a pattern used by some other libraries facing classloader isolation (e.g. [RESTHeart](https://github.com/SoftInstigate/restheart/blob/master/mongoclient/src/main/java/org/restheart/mongodb/MongoClientSingleton.java#L206), also based on [this SO answer](https://stackoverflow.com/a/47445573)). Only bootstrap-classloader types (Path, Lock) are stored, so cross-classloader casting is safe.

Reproduction:
A Gradle composite build with 8 included builds, each with its own buildSrc declaring the jib plugin, running jibBuildTar in parallel. With jib 3.5.3, this reliably triggers the exception (and builds hang). With this fix applied, all 8 tasks complete successfully across repeated runs.
[jib-parallel-build-reproducer.zip](https://github.com/user-attachments/files/28172482/jib-parallel-build-reproducer.zip)

Fixes https://github.com/GoogleContainerTools/jib/issues/3347

### DIFF
--- a/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
+++ b/jib-core/src/main/java/com/google/cloud/tools/jib/filesystem/LockFile.java
@@ -31,7 +31,19 @@ import java.util.concurrent.locks.ReentrantLock;
 /** Creates and deletes lock files. */
 public class LockFile implements Closeable {
 
-  private static final ConcurrentHashMap<Path, Lock> lockMap = new ConcurrentHashMap<>();
+  // In Gradle composite builds with independent buildSrc directories, each included build loads
+  // jib-core in its own classloader. A normal static field creates separate lockMap instances per
+  // classloader, failing to prevent concurrent FileChannel.lock() calls on the same file and
+  // causing OverlappingFileLockException (#3347). This is admittedly a hack, but there is no
+  // clean way to share state across classloaders in the same JVM. System.getProperties() returns
+  // the same Hashtable regardless of classloader, so it serves as a JVM-global namespace.
+  // The map's key/value types (Path, Lock) are bootstrap-loaded, so they're safe to cast.
+  @SuppressWarnings("unchecked")
+  private static final ConcurrentHashMap<Path, Lock> lockMap =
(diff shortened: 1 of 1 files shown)


hczedik:
I have signed the Contributor License Agreement (CLA).

@diegomarquezp I would appreciate if you or another of the maintainers could review this. Thank you!

PR #4528, opened 2026-09-01
fix: disambiguate duplicate dependency JARs by content hash instead of file size

sideeffffect:
## Issue

When several dependency JARs share the same filename, jib renames the duplicates so they don't overwrite each other under `libs/` in the image. Today the disambiguating suffix is the **file size** (`Files.size(...)`):

```java
jarName = jarName.replaceFirst("\\.jar$", "-" + Files.size(file)) + ".jar";
```

File size is not a reliable discriminator. Two genuinely different JARs that happen to share a filename can also happen to have the *exact same size*, in which case one still silently overwrites the other — producing `NoClassDefFoundError` at runtime. This is the same class of problem originally reported in #3331; the size-based fix narrowed it but didn't close it.

## Fix

Use a short prefix of the **SHA-256 of the JAR contents** as the suffix instead, computed via jib-core's existing `Digests` utility. A content hash distinguishes distinct JARs regardless of their size, while staying deterministic across builds (so it doesn't hurt reproducibility or layer caching). The digest is truncated to 12 hex characters to keep filenames reasonable — more than enough to tell apart the handful of same-named JARs in a single build.

The renaming exists in two places that are required to stay in sync:
- `jib-core` — `JavaContainerBuilder`
- `jib-plugins-common` — `PluginConfigurationProcessor` (used by both the Maven and Gradle plugins)

Both are updated identically, via a small private `computeContentHash(Path)` helper in each.

## Tests

Updated the two existing duplicate-JAR tests to assert the new content-hash suffixes (`JavaContainerBuilderTest`, `PluginConfigurationProcessorTest`). Verified locally:
- `:jib-core:test --tests JavaContainerBuilderTest` ✅
- `:jib-plugins-common:test --tests PluginConfigurationProcessorTest` ✅
- `verifyGoogleJavaFormat` and errorprone/NullAway (via `compileJava`) ✅

CHANGELOGs updated for `jib-core`, `jib-maven-plugin`, and `jib-gradle-plugin`.

Fixes the remaining collision described in #3331.


sideeffffect:
@diegomarquezp @whowes could you please have a look at this PR? :pray:

